PR #4, opened 2016-05-16
Tx tests invalid

Christewart:
This pull request modifies to the codebase to ensure that all tests inside of tx_invalid.json fail. These tests are specified by bitcoin core to indicate behavior that all bitcoin libraries should follow.


TomMcCabe:
ACK.

PR #5, opened 2016-05-19
Refactoring project into 'core' package

Christewart:
Places the project into a 'core' package. Now the core project resides in this package path 'org.bitcoins.core'


TomMcCabe:
ACK.

PR #6, opened 2016-05-19
Rename marshallers to serializers

Christewart:
Renames anything called 'marshaller' to 'serializer'. Removing some rpc-client tests that do not belong in this project


TomMcCabe:
ACK

PR #8, opened 2016-05-24
Renaming TransactionElement to NetworkElement

Christewart:
This renames our previous trait 'TransactionElement' to 'NetworkElement'. A 'NetworkElement' is any object that be serialized to hex/binary


TomMcCabe:
ACK.

PR #9, opened 2016-05-24
Create hash types

Christewart:
This pull request creates 'HashDigest' types. These types are used to differentiate between different HashDigest - for instance distinguishing between DoubleSHA256 hash & a single SHA256 hash. What we were doing previously was just returning a Seq[Byte] or a String.


Christewart:
This can be merged, I adjusted the coveralls setting so it doesn't fail unless we see a decrease in test coverage greater than 2%

TomMcCabe:
Concept and Coveralls config ACK.

PR #10, opened 2016-05-25
Remove asm arg script types, refactor script types into their companion objects

Christewart:
Previously we had two arguments for both script signature & script pubkey types. The first one was the hexadecimal representation of the scriptSig/scriptPubkey. The second one was the 'asm' representation. The asm representation can be derived from the hexadecimal representation so this pull request removes it from the case class constructor. 

This pull request also moves all script signature / script pubkey types into their corresponding companion object. This is just good software engineering practices to encapsulate the implementation types.


TomMcCabe:
ACK.

PR #11, opened 2016-05-30
Remove bitcoinj as source dependency

Christewart:
Previously we were using some helper functions inside of bitcoinj to help ease our barrier to entry to writing a bitcoin library, this removes the bitcoinj dependency


TomMcCabe:
ACK.

PR #12, opened 2016-05-31
Refactor ec factory

Christewart:
This removes the ECFactory object, and moves the creation of EC types into their corresponding companion objects. For instance, to create a fresh private key we would now call ECPrivateKey.freshPrivateKey


TomMcCabe:
ACK

PR #13, opened 2016-06-02
Moving apply(bytes) & apply(hex) into the Factory trait

Christewart:
This removes the redundant implementations of the two 'apply' functions found in all of our companion object factories to their common parent trait called 'Factory'. This was becoming boilerplate code added to every companion object factory - best to refactor it to its parent trait.


TomMcCabe:
ACK

PR #16, opened 2016-06-22
Base58 json tests

TomMcCabe:
This pull request includes tests for:
- base58_keys_valid.json
- base58_keys_invalid.json
- base58_encode_decode.json

Some improvements to Base58 as well:
-  "isValid" function to determine whether or not a string is a valid Base58 string
- Function to encode private keys into Wallet Import Format (WIF), which is Base58 encoded
- Function to encode a Sha256Hash160Digest to a base58 address
- Fix bug in "encode" function that didn't account for leading byte values of zero.


src/main/scala/org/bitcoins/core/crypto/ECPrivateKey.scala
```diff
@@ -85,9 +85,17 @@ object ECPrivateKey extends Factory[ECPrivateKey] {
     val decodedBase58 : Seq[Byte] = Base58.decode(base58)
     //Drop(1) will drop the network byte. The last 5 bytes are dropped included the checksum (4 bytes), and 0x01 byte that
     //is appended to compressed keys (which we implemented as the default option).
-    val trim = decodedBase58.drop(1).dropRight(5)
+    val trim = {
```
Christewart: Factor this { ... } into a function, whenever you use val trim = {...} it indicates you should be probably create a function for whatever is inside {...}

Christewart: Factor this `{ ... }`into a function, whenever you use `val trim = {...}` it indicates you should be probably create a function for whatever is inside `{...}`

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,6 +60,67 @@ trait Base58 extends BitcoinSLogger {
...
+    * Encodes a Base58 address from a hash
+    * @param hash The result of Sha256(RipeMD-160(public key))
+    * @param addressType string. Either "pubkey" or "script"
+    * @param isTestNet boolean
+    * @return
+    */
+  def encodePubKeyHashToBase58Address(hash: Sha256Hash160Digest,
+                                  addressType : String,
+                                  isTestNet : Boolean) : String = {
+    val versionByte : Byte = {
```
Christewart: See comment above, create a function called `parseVersionByte` or something like that

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,6 +60,67 @@ trait Base58 extends BitcoinSLogger {
...
+        throw new IllegalArgumentException("Address must be of type 'pubkey' or 'script'."))
+      if (!isTestNet) {
+        if (addressType == "pubkey") MainNet.p2pkhNetworkByte
+        else MainNet.p2shNetworkByte
+      }
+      else if (isTestNet) {
+        if (addressType == "pubkey") TestNet3.p2pkhNetworkByte
+        else TestNet3.p2shNetworkByte
+      }
+      else throw new IllegalArgumentException("Something broke -- Check your parameters. There should be a " +
```
Christewart: This case can never happen, since `if(!isTestNet)`, and `else if(isTestNet)` are inverses, so one will always evaluate to true. This is dead code

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -65,10 +131,22 @@ trait Base58 extends BitcoinSLogger {
...
-    val decoded = trim.foldLeft(BigInt(0))((a,b) =>a.*(BigInt(58L)).+(BigInt(base58Pairs(b))))
+    val decoded = trim.foldLeft(BigInt(0))((a,b) =>
+      a.*(BigInt(58L)).+(BigInt(base58Pairs(b))))
     if (trim.isEmpty) zeroes else zeroes ++ decoded.toByteArray.dropWhile(_ == 0)
   }
 
+  def isValid(base58 : String) : Boolean = {
+    val firstByte : Seq[Byte]= if (base58.isEmpty) List() else Seq(decode(base58).head)
+    val length = base58.length
+    val validFirstByteInHex = List("00", "05", "80", "6f", "c4", "ef")
```
Christewart: What are these numbers? These look awfully similar to some parameters I have in ChainParams

src/test/scala/org/bitcoins/core/util/Base58Test.scala
```diff
@@ -53,5 +90,53 @@ class Base58Test extends FlatSpec with MustMatchers {
...
+
+    val source = Source.fromURL(this.getClass.getResource("/base58_keys_invalid.json"))
+    val lines = try source.getLines.filterNot(_.isEmpty).map(_.trim) mkString "\n" finally source.close()
+    val json = lines.parseJson
+    val testCases : Seq[Base58InvalidTestCase] = json.convertTo[Seq[Base58InvalidTestCase]]
+    for {
+      testCase <- testCases
+    } yield {
+      testCase must be (Base58InvalidTestCaseImpl(testCase.base58EncodedString))
+      Base58.isValid(testCase.base58EncodedString) match {
```
Christewart: This is not right, you can replace this with `Base58.isvalid(...) must be (false)` instead of your match statement

src/test/scala/org/bitcoins/core/util/testprotocol/Base58ValidTestCaseProtocol.scala
```diff
@@ -0,0 +1,37 @@
...
+  import ConfigParamsProtocol._
+  implicit object Base58ValidTestCaseFormatter extends RootJsonFormat[Base58ValidTestCase] {
+    override def read(value: JsValue): Base58ValidTestCase = {
+      val jsArray: JsArray = value match {
+        case array: JsArray => array
+        case _: JsValue => throw new RuntimeException("Core test case must be in the format of js array")
+      }
+      val elements: Vector[JsValue] = jsArray.elements
+      val configParams : ConfigParams = elements(2).convertTo[ConfigParams]
+      val addressOrPrivKey : Either[Address, String] = {
```
Christewart: instead of doing `val addressOrPrivKey = { ... }` factor everything in between `{ ... }` into a function

src/test/scala/org/bitcoins/core/util/testprotocol/Base58ValidTestCaseProtocol.scala
```diff
@@ -0,0 +1,37 @@
...
+        if (!configParams.isPrivKey) {
+          Left(Address(elements(0).convertTo[String]))
+        }
+        else Right(elements(0).convertTo[String])
+      }
+      val hashOrPrivKey : Either[Sha256Hash160Digest, ECPrivateKey] = {
```
Christewart: See comment above

src/test/scala/org/bitcoins/core/util/Base58Test.scala
```diff
@@ -53,5 +90,53 @@ class Base58Test extends FlatSpec with MustMatchers {
...
+    val json = lines.parseJson
+    val testCases : Seq[Base58ValidTestCase] = json.convertTo[Seq[Base58ValidTestCase]]
+    for {
+      testCase <- testCases
+    } yield {
+      testCase must be (Base58ValidTestCaseImpl(testCase.addressOrWIFPrivKey, testCase.hashOrPrivKey, testCase.configParams))
+      if (testCase.addressOrWIFPrivKey.isLeft) {
+        Base58.isValid(testCase.addressOrWIFPrivKey.left.get.value) must be (true)
+      }
+      else println("Base58.isValid(testCase.addressOrWIFPrivKey.right.get): " + Base58.isValid(testCase.addressOrWIFPrivKey.right.get))
```
Christewart: Is this even testing anything? Shouldn't there be a `must be` some where here to test the `right` value?

src/test/scala/org/bitcoins/core/util/testprotocol/Base58InvalidTestCaseProtocol.scala
```diff
@@ -0,0 +1,52 @@
...
+    }
+    override def write (testCase : Seq[Base58InvalidTestCase]) : JsValue = ???
+  }
+
+  /**
+    * Function to parse the JSON values into a Seq[Base58InvalidTestCase]
+    * @param elements
+    * @return
+    */
+  def parser(elements : Seq[JsValue]) : Seq[Base58InvalidTestCase] = {
```
Christewart: Probably need a better name for this function, also can be it `private`

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,6 +60,67 @@ trait Base58 extends BitcoinSLogger {
...
+    * @param addressType string. Either "pubkey" or "script"
+    * @param isTestNet boolean
+    * @return
+    */
+  def encodePubKeyHashToBase58Address(hash: Sha256Hash160Digest,
+                                  addressType : String,
+                                  isTestNet : Boolean) : String = {
+    val versionByte : Byte = {
+      require(addressType != "pubkey" || addressType != "script",
+        throw new IllegalArgumentException("Address must be of type 'pubkey' or 'script'."))
```
Christewart: You don't need the `throw new IllegalArgumentException` here, you can just give the string constant. The `require` function will throw the `IllegalArgumentException` for you

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,6 +60,67 @@ trait Base58 extends BitcoinSLogger {
...
+    * @param privateKey
+    * @param isCompressed
+    * @param isTestNet
+    * @return
+    */
+  //TODO: Create WIF PrivateKey Type
+  def encodePrivateKeyToWIF(privateKey : ECPrivateKey,
+                            isCompressed : Boolean,
+                            isTestNet : Boolean) : String = {
+    val versionByte : Byte = isTestNet match {
```
Christewart, line 112: Nice use of match statements.

Christewart:
I'm guessing the build failed because `INFO` is set inside of `logback-test.xml` instead of `WARN`

TomMcCabe:
IsValid function in Base58 trait needs to be cleaner, but everything works. The bug involving those previous invalid keys that were passing as true was due to the fact that I wasn't verifying a compressed public key's 5th-to-last byte as being 0x01.

src/main/scala/org/bitcoins/core/crypto/ECPrivateKey.scala
```diff
@@ -83,11 +83,21 @@ object ECPrivateKey extends Factory[ECPrivateKey] {
     */
   def fromBase58ToPrivateKey(base58 : String) : ECPrivateKey = {
     val decodedBase58 : Seq[Byte] = Base58.decode(base58)
+    //Private keys starting with 'K', 'L', or 'c' correspond to compressed public keys.
+    def isCompressed : Boolean = {
```
Christewart: These two functions (`isCompressed` and `trimFunction`) should be moved out of the inside of `fromBase58ToPrivateKey`, also `isCompressed` should take in a `Seq[Byte]` as an argument. Also add a scaladoc to `trimFunction` explaining exactly what it does

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,20 +64,132 @@ trait Base58 extends BitcoinSLogger {
...
+
+  /**
+    * Encodes a Base58 address from a hash
+    *
+    * @param hash The result of Sha256(RipeMD-160(public key))
+    * @param addressType string. Either "pubkey" or "script"
+    * @param isTestNet boolean
+    * @return
+    */
+  def encodePubKeyHashToBase58Address(hash: Sha256Hash160Digest,
```
Christewart: Removed the commented code in here

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,20 +64,132 @@ trait Base58 extends BitcoinSLogger {
...
+    else {
+      val decoded = decode(base58)
+      val firstByte = decoded.head
+      val validAddressPreFixBytes: Seq[Byte] =
+        MainNetChainParams.base58Prefixes(PubKeyAddress) ++ MainNetChainParams.base58Prefixes(ScriptAddress) ++
+          TestNetChainParams.base58Prefixes(PubKeyAddress) ++ TestNetChainParams.base58Prefixes(ScriptAddress)
+      val validSecretKeyPreFixBytes : Seq[Byte] =
+        MainNetChainParams.base58Prefixes(SecretKey) ++ TestNetChainParams.base58Prefixes(SecretKey)
+      val compressedPubKey = List('K', 'L', 'c').contains(base58.head)
+      def checkCompressedPubKeyValidity : Boolean = {
```
Christewart: Move this outside of the `isValid` function and make it `private` and add a parameter to the function

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,20 +64,132 @@ trait Base58 extends BitcoinSLogger {
...
+        val compressedByte = decoded(decoded.length - 5)
+        compressedByte == 0x01.toByte
+      }
+      try {
```
Christewart: Don't use `try`, use `Try` (with an uppercase `T`). This is better for various reasons, one of which is performance. You can use a match statement on the `Try` afterwards checking if it is a `Success` or `Failure`

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,20 +64,132 @@ trait Base58 extends BitcoinSLogger {
...
+        }
+        else false
+      }
+      catch {
+        case error : IllegalArgumentException => false
+      }
+    }
+  }
+
+/*    def isValid(base58 : String) : Boolean = {
```
Christewart: Remove this code

src/main/scala/org/bitcoins/core/crypto/ECPrivateKey.scala
```diff
@@ -77,17 +82,54 @@ object ECPrivateKey extends Factory[ECPrivateKey] {
...
+    * Takes in WIF private key, as a string, and determines if it corresponds to a compressed public key
+    * If the private key corresponded to a compressed public key, also drop the last byte (it should be 0x01).
+    * If it corresponded to a compressed public key, the WIF string will have started with K or L instead of 5
+    * (or c instead of 9 on testnet).
+    *
+    * @param WIF
+    * @return
+    */
+  def isCompressed(WIF : String) : Boolean = {
+    val decoded = Base58.decode(WIF)
```
Christewart: This is relatively minor, but instead of calling `decode` right away inside of `isCompressed` you should just move this stuff inside of `isCompressed(bytes : Seq[Byte])` therefore you don't have to call the decode

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -55,23 +62,149 @@ trait Base58 extends BitcoinSLogger {
...
+    * @param hash The result of Sha256(RipeMD-160(public key))
+    * @param addressType string. Either "pubkey" or "script"
+    * @param isTestNet boolean
+    * @return
+    */
+    //TODO: add logic to determine pubkey/script from first byte
+  def encodePubKeyHashToBase58Address(hash: Sha256Hash160Digest,
+                                  addressType : String,
+                                  isTestNet : Boolean) : String = {
+    def parseVersionByte(addressType : String, isTestnet : Boolean) : Byte = isTestnet match {
```
Christewart: Move this outside of `encodePubKeyHashToBase58Address`

Christewart:
ACK

PR #18, opened 2016-07-18
Refactoring generating random numbers in Generators to use ScalaCheck…

Christewart:
Refactoring generators to use ScalaCheck number generators instead of creating our own.


Christewart:
ACK

PR #20, opened 2016-07-21
Add functions to create address

Christewart:
Refactoring address types to Algebraic Data Types. Making companion objects for `P2PKHAddress` and `P2SHAddress` and adding helper functions inside of each to make it easier to create their corresponding type.


TomMcCabe:
tACK

PR #21, opened 2016-07-28
Transaction signature creator

Christewart:
This pull request implements `TransactionSignatureCreator`, which is responsible for creating digital signatures for transactions in bitcoin-s. The base functionality is similar to bitcoin core's [TransactionSignatureCreator](https://github.com/bitcoin/bitcoin/blob/master/src/script/sign.h#L34) class. 

Also implemented in this pull request is generators for signed script signature types `P2PK`, `P2PKH`, `MultiSignature` and `P2SH`. With these generators, we test our `ScriptInterpreter` inside of `TransactionSignatureCreatorSpec` to make sure all of the correctly signed scriptSignatures evaluate to true.


Christewart:
ACK

src/test/scala/org/bitcoins/core/util/CryptoTestUtil.scala
```diff
@@ -9,7 +9,7 @@ import org.bitcoins.core.crypto.{ECPrivateKey}
  */
 trait CryptoTestUtil {
   def privateKeyBase58 = "cVLwRLTvz3BxDAWkvS3yzT9pUcTCup7kQnfT2smRjvmmm1wAP6QT"
-  def privateKeyBytes  = Base58.decode(privateKeyBase58)
+  def privateKeyBytes  = Base58.decodeCheck(privateKeyBase58).get
```
TomMcCabe: Wasn't sure about this b/c `decodeCheck` doesn't return the checksum, but I also noticed `privateKeyBytes` and `privateKeyHex` aren't even used in our code as I didn't find any usages. All tests passed when I deleted lines 12/13 so they could be removed.

TomMcCabe:
ACK pending comment review.

PR #24, opened 2016-08-05
Ec priv key compressed

Christewart:
This adds support to actually specify if you want a `ECPrivateKey` to correspond to a compressed or uncompressed public key. This is important because a compressed private key will generate a different bitcoin address.


TomMcCabe:
ACK.

PR #26, opened 2016-08-08
Adding function inside of Merkle to actually build and return the Mer…

Christewart:
…kleTree, changing how endianess is stored inside of the Merkle root from big endian to little endian.


TomMcCabe:
ACK.

Issue #28, opened 2016-08-15
signatures not being properly DER encoded


TomMcCabe:
ECDigitalSignatures aren't returning as strictly DER encoded.

Values for example when building CLTV spend:
- Unsigned transaction hex: `0100000001afecbf9d3729cdfea287538d097bd58e6099a504a5464af9d4a221ad3f6015d7000000002b2a046202b257b1752102ed83fb33da52014f08497b61b73d3deefbc298728fccfcf3e0b6042c8363df51ac4d01000001c0e4022a010000001976a91466c6532c4f35619ca671286d407f2e39fe08fd7488ac2934b257`
- RedeemScript hex: `046202b257b1752102ed83fb33da52014f08497b61b73d3deefbc298728fccfcf3e0b6042c8363df51ac`
- RedeemScript asm: `1471283810 OP_CHECKLOCKTIMEVERIFY OP_DROP 02ed83fb33da52014f08497b61b73d3deefbc298728fccfcf3e0b6042c8363df51 OP_CHECKSIG`
- P2SHScriptPubKey hex: `a9140db4fe237cb30e7ee7a096dbc2033d93242435ed87`
- P2SH asm: `OP_HASH160 0db4fe237cb30e7ee7a096dbc2033d93242435ed OP_EQUAL`
- PrivateKey to spending redeemScript: `7939911e7d96660d8de84d7376f4a85bf72ba1b7e107815aa6c99076a14796e1`
- TransactionSignatureComponent: `TransactionSignatureComponent(unsignedTx, UInt32(0), p2sh, flags)`
- TransactionSignatureSerializer: `hashForSignature(txSigComponent, SIGHASH_ALL.defaultValue))`

What I returned:
- Hash: `a758727808e9e35b7f9388016cfa0fb604d558ccef3b169b469f952266c8e611)`
- ECDigitalSignature with privateKey: `3045022100c16c04aa9923472b0f2d796bc3360b4a4277c6377d1db3758bc06df88cec017f02200f43e0af839e59e68bd34b745ebd7e291a4ba548facf16146d13cf39c70b0717`
- sig.isDEREncoded: `false`

---

Also found this signature from a CLTV spend on bitcointalk that IS strictly DER encoded: 

`30440220429b09433f70c16b26021bd19f71a7d30087c3bbd7e1255cfda7c00fdc126bed02207219634bd0f66eee1b73e60552bbf177f278ba856b99100bdc610b5da720168501`

[...]

PR #29, opened 2016-08-16
Der encoding bug

Christewart:
Fixing #28 , adding some invariants to make sure we are dealing with der encoded signatures and signatures with low s values. Renaming `isStrictDEREncoding` to `isValidSignatureEncoding`, which is the naming Core uses. This is more accurate because bitcoin's signature scheme uses DER encoding along with a hash type appended to the end of the DER encoding -- if we validate this signature using ONLY DER encoding rules it will fail therefore we need to use `isValidSignatureEncoding`

### DIFF
--- a/src/main/scala/org/bitcoins/core/crypto/BaseECKey.scala
+++ b/src/main/scala/org/bitcoins/core/crypto/BaseECKey.scala
@@ -36,8 +36,9 @@ trait BaseECKey extends BitcoinSLogger {
     //https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures
     //bitcoinj implementation
     //https://github.com/bitcoinj/bitcoinj/blob/1e66b9a8e38d9ad425507bf5f34d64c5d3d23bb8/core/src/main/java/org/bitcoinj/core/ECKey.java#L551
-    if (s.compareTo(CryptoParams.halfCurveOrder) <= 0) signature
-    else ECDigitalSignature(r,CryptoParams.curve.getN().subtract(s))
+    val signatureLowS = lowS(signature)
+    require(signatureLowS.isDEREncoded, "We must create DER encoded signatures when signing a peice of data, got: " + signatureLowS)
+    signatureLowS
   }
 
   def sign(hex : String, signingKey : BaseECKey) : ECDigitalSignature = sign(BitcoinSUtil.decodeHex(hex),signingKey)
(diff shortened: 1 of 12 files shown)


src/main/scala/org/bitcoins/core/crypto/BaseECKey.scala
```diff
@@ -51,6 +52,13 @@ trait BaseECKey extends BitcoinSLogger {
   def sign(hash: HashDigest, signingKey: BaseECKey): ECDigitalSignature = sign(hash.bytes,signingKey)
 
 
+  /** Checks if the given digital signature uses a low s value, if it does not it converts it to a low s value and returns it */
+  private def lowS(signature: ECDigitalSignature): ECDigitalSignature = {
+    val sigLowS = if (signature.s.bigInteger.compareTo(CryptoParams.halfCurveOrder) <= 0) signature
```
TomMcCabe: Perhaps use `DERSignatureUtil`'s `isLowS` function inside `sigLowS`

src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
```diff
@@ -80,4 +91,12 @@ object ECDigitalSignature extends Factory[ECDigitalSignature] {
       r.toByteArray.toSeq ++ Seq(0x2.toByte, s.toByteArray.size.toByte) ++ s.toByteArray.toSeq
     fromBytes(bytes)
   }
+
+
+  /** Checks if the given digital signature uses a low s value,
+    * if it does not it converts it to a low s value and returns it */
+  private def lowS(signature: ECDigitalSignature): ECDigitalSignature = {
+    if (signature.s.bigInteger.compareTo(CryptoParams.halfCurveOrder) <= 0) signature
```
TomMcCabe, line 99: Same as comment above. What about just making `lowS` in `DERSignatureUtil` ?

src/main/scala/org/bitcoins/core/crypto/BaseECKey.scala
```diff
@@ -36,8 +36,9 @@ trait BaseECKey extends BitcoinSLogger {
     //https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures
     //bitcoinj implementation
     //https://github.com/bitcoinj/bitcoinj/blob/1e66b9a8e38d9ad425507bf5f34d64c5d3d23bb8/core/src/main/java/org/bitcoinj/core/ECKey.java#L551
-    if (s.compareTo(CryptoParams.halfCurveOrder) <= 0) signature
-    else ECDigitalSignature(r,CryptoParams.curve.getN().subtract(s))
+    val signatureLowS = lowS(signature)
+    require(signatureLowS.isDEREncoded, "We must create DER encoded signatures when signing a peice of data, got: " + signatureLowS)
```
TomMcCabe: Nit: typo

TomMcCabe:
ACK.

PR #31, opened 2016-08-25
Add CLTVScriptPubKey and CSVScriptPubKey

TomMcCabe:
This PR implements two new scriptPubKeys to implement the opCodes, OP_CHECKLOCKTIMEVERIFY (BIP65) and OP_CHECKSEQUENCEVERIFY (BIP112):
- CLTVScriptPubKey
- CSVScriptPubKey

Few nits/typos fixed along the way, notably:
- improved hashTypes function in ScriptSignature
- added a p2shAddress function for the ScriptPubKey trait -- thought it may be useful but perhaps in an inappropriate place
- added apply functions for each ScriptPubKey's fromAsm function


src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -26,14 +29,21 @@ sealed trait ScriptPubKey extends NetworkElement with BitcoinSLogger {
    */
   lazy val asm : Seq[ScriptToken] = ScriptParser.fromBytes(bytes)
 
+  def p2shAddress(network : NetworkParameters) : P2SHAddress = {
```
Christewart: Don't think this is the right place for this, it should be inside of `P2SHScriptPubKey`. Perhaps we could add a `address` function here, and then implement that function inside of all of the sub classes, however we would then have to figure out a way to generate an address for a `NonStandardScriptPubKey` -- i don't think this behavior is defined in Bitcoin Core.

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,20 +334,137 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+
+  def apply (asm: Seq[ScriptToken]) : CLTVScriptPubKey = fromAsm(asm)
+
+  /**
+    * Creates a P2PKH-formatted CLTVScriptPubKey
+    * @param locktime block height or timestamp
+    * @param pubKey public key corresponding to the CLTV output
+    * @return
+    */
+  def apply(locktime : ScriptNumber, pubKey : ECPublicKey) : CLTVScriptPubKey = {
```
Christewart: Not sure if I like the idea of exposing this function to everyone, they can just call the other `apply` function with an already built `P2PKHScriptPubKey`... let me think about it more

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -353,16 +488,17 @@ object ScriptPubKey extends Factory[ScriptPubKey] with BitcoinSLogger {
...
-      P2PKHScriptPubKey.fromAsm(asm)
-    case List(OP_HASH160, x : BytesToPushOntoStack, y : ScriptConstant, OP_EQUAL) =>
-      P2SHScriptPubKey.fromAsm(asm)
-    case List(b : BytesToPushOntoStack, x : ScriptConstant, OP_CHECKSIG) => P2PKScriptPubKey.fromAsm(asm)
-    case _ if (MultiSignatureScriptPubKey.isMultiSignatureScriptPubKey(asm)) =>
-      MultiSignatureScriptPubKey.fromAsm(asm)
-    case _ => NonStandardScriptPubKey.fromAsm(asm)
+    case _ if P2PKHScriptPubKey.isP2PKHScriptPubKey(asm) => P2PKHScriptPubKey(asm)
+    case _ if P2SHScriptPubKey.isP2SHScriptPubKey(asm) => P2SHScriptPubKey(asm)
+    case _ if P2PKScriptPubKey.isP2PKScriptPubKey(asm) => P2PKScriptPubKey(asm)
```
Christewart: Duplicate `P2PKScriptPubKey`

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -39,15 +39,21 @@ sealed trait ScriptSignature extends NetworkElement with BitcoinSLogger {
...
+    * Derives the hash types for every digitalSignature in the ScriptSignature object
     *
-    * @param digitalSignature
     * @return
     */
-  def hashType(digitalSignature: ECDigitalSignature) : HashType = {
-    digitalSignature match {
-      case EmptyDigitalSignature => SIGHASH_ALL(Int32.one)
-      case sig : ECDigitalSignature => HashType(Seq(digitalSignature.bytes.last))
+  def hashTypes : Seq[HashType] = {
```
Christewart: Not sure if this is useful for anything, perhaps we should delete this function `hashType` and `hashTypes` all together? Would this cause any errors in our codebase?

src/main/scala/org/bitcoins/core/util/Base58.scala
```diff
@@ -71,54 +71,24 @@ trait Base58 extends BitcoinSLogger {
...
+    * If p2pkh, encodes the Sha256Hash160 of a public key to a base 58 address.
+    * If p2sh, encodes the Sha256Hash160 of a scriptPubKey to a base 58 address.
     * @param hash the result of Sha256(RipeMD160(pubkey))
-    * @param network the network on which this address is being generated for
+    * @param network the network on which this address is being generated for. (MainNet, TestNet3, RegTest)
     * @return
     */
-  def encodePubKeyHashToAddress(hash: Sha256Hash160Digest, network: NetworkParameters): Address = {
-    val versionByte: Byte = network.p2pkhNetworkByte
+  def encodeHashToAddress(hash: Sha256Hash160Digest, network: NetworkParameters, p2sh : Boolean): Address = {
```
Christewart: We already have this functionality https://github.com/bitcoin-s/bitcoin-s-core/blob/master/src/main/scala/org/bitcoins/core/protocol/Address.scala#L29

src/test/scala/org/bitcoins/core/protocol/script/CLTVScriptPubKeySpec.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.core.protocol.script
+
+import org.bitcoins.core.gen.ScriptGenerators
+import org.scalacheck.{Properties, Prop}
+
+/**
+  * Created by tom on 8/23/16.
+  */
+class CLTVScriptPubKeySpec extends Properties("CLTVScriptPubKeySpec") {
+  property("Serialization symmetry") =
```
Christewart, line 14: Try to think of a way to write a property that says if the locktime is NOT in the past, our CLTV script should fail

src/test/scala/org/bitcoins/core/protocol/script/CSVScriptPubKeySpec.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.core.protocol.script
+
+import org.bitcoins.core.gen.ScriptGenerators
+import org.scalacheck.{Prop, Properties}
+
+/**
+  * Created by tom on 8/23/16.
+  */
+class CSVScriptPubKeySpec extends Properties("CSVScriptPubKeySpec") {
```
Christewart, line 9: Same thing as I said on CLTV except with CSV

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
TomMcCabe: Agree, I don't think it's placed appropriately, or if it's even necessary at all.

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
TomMcCabe: Would not be opposed to removing it. Could be misconstrued as producing a P2PKScriptPubKey if documentation isn't read.

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
TomMcCabe: It's only used in two test files, so we could safely remove it.

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
TomMcCabe: Also FWIW, decoding a scriptPubKey (of any type) in hex in the bitcoin daemon returns the P2SH address that it would generate if that scriptPubKey was put into a P2SHScriptPubKey.

Removing it for time being.

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
Christewart: That was my exact concern, lets remove it for now

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
Christewart: Ok let's remove it then, no point in keeping this for the added complexity. Although I think it would be handy to have a function some where to get the hash type of a digital signature, perhaps `DERSignatureUtil`? I guess we have to be careful of bitcoin digital signatures vs standard digital signatures used in crypto libraries

src/main/scala/org/bitcoins/core/gen/NumberGenerator.scala
```diff
@@ -72,6 +72,8 @@ trait NumberGenerator {
 
   def scriptNumbers: Gen[ScriptNumber] = Gen.choose(Int64.min.underlying, Int64.max.underlying).map(ScriptNumber(_))
 
+  def unspendableLockTimeScriptNumbers : Gen[ScriptNumber] = Gen.choose(4294967296L, Int64.max.underlying).map(ScriptNumber(_))
```
Christewart: Add a comment on top of this function for what it is intended to be used for, is this meant for CSV, CLTV, both etc? Why are these numbers always unspendable?

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -73,36 +103,64 @@ trait ScriptGenerators extends BitcoinSLogger {
...
-    val randomNum = (scala.util.Random.nextInt() % 3).abs
+    val randomNum = (scala.util.Random.nextInt() % 5).abs
     if (randomNum == 0) p2pkScriptPubKey
     else if (randomNum == 1) p2pkhScriptPubKey
+    else if (randomNum == 2) cltvScriptPubKey
+    else if (randomNum == 3) csvScriptPubKey
     else multiSigScriptPubKey
   }
 
+  def pickRandomNonLockTimeScriptPubKey(pubKey : ECPublicKey) : Gen[ScriptPubKey] = {
```
Christewart: `P2SHScriptPubKey` and `MultsigScriptPubKey` isn't included here

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -73,36 +103,64 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    val randomNum = (scala.util.Random.nextInt() % 3).abs
+    if (randomNum == 0) P2PKScriptPubKey(pubKey)
+    else if (randomNum == 1) P2PKHScriptPubKey(pubKey)
+    else P2PKHScriptPubKey(pubKey)
+  }
+
+  def pickRandomNonLockTimeScriptPubKey : Gen[(ScriptPubKey, Seq[ECPrivateKey])] = {
```
Christewart: Why is there two different instances of this function, one that takes a `pubKey: ECPublicKey` and one that doesn't?

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -73,36 +103,64 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    val num = (scala.util.Random.nextInt() % 5).abs
+    val isMulti : Boolean = if (num == 4) true else false
+      if (isMulti)  {
```
Christewart: This should probably just use the `multiSigScriptPubKey` generator shouldn't it instead of calling `CryptoGenerators.privateKeySeqWithRequiredSigs` etc...

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,9 +277,49 @@ trait ScriptGenerators extends BitcoinSLogger {
     (p2SHScriptSignature, p2SHScriptPubKey, privateKeys)
   }
 
+  /**
+    * Generates a signed [[CLTVScriptSignature]] that spends from a [[CLTVScriptPubKey]] correctly
+    *
+    * @return the signed [[CLTVScriptSignature]], the [[CLTVScriptPubKey]] it spends, and the sequences of [[ECPrivateKey]]
+    *         used to sign the scriptSig
+    */
+  def signedCLTVScriptSignature(scriptNumber : ScriptNumber, lockTime : UInt32, sequence : UInt32) : Gen[(ScriptSignature, CLTVScriptPubKey, Seq[ECPrivateKey])] = for {
```
Christewart: I think this should technically return something of type `CLTVScriptSignature` instead of just a generic `ScriptSignature`

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,9 +277,49 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    val signedScriptSig = CLTVScriptSignature(cltv, Seq(txSignature), Seq(privKey.publicKey), None)
+    (signedScriptSig, cltv, Seq(privKey))
+  }
+
+  /**
+    * Generates a signed [[CSVScriptSignature]] that spends from a [[CSVScriptPubKey]] correctly
+    * @return the signed [[CSVScriptSignature]], the [[CSVScriptPubKey]] it spends, and the sequences of [[ECPrivateKey]]
+    *         used to sign the scriptSig
+    */
+  def signedCSVScriptSignature(csvScriptNum : ScriptNumber, sequence : UInt32) : Gen[(ScriptSignature, CSVScriptPubKey, Seq[ECPrivateKey])] = for {
```
Christewart: I think this should technically return a `CSVScriptSignature` instead of a generic `ScriptSignature`

src/main/scala/org/bitcoins/core/gen/TransactionGenerators.scala
```diff
@@ -116,24 +124,143 @@ trait TransactionGenerators {
...
+  } yield {
+    val (creditingTx, outputIndex) = buildCreditingTransaction(UInt32(2),csvScriptPubKey)
+    //Transaction version must not be less than 2 for a CSV transaction
+    val (signedSpendingTx, inputIndex) = buildSpendingTransaction(UInt32(2), creditingTx, signedScriptSig, outputIndex, UInt32.zero, sequence)
+
+    val txSigComponent = TransactionSignatureComponent(signedSpendingTx, inputIndex, csvScriptPubKey, Policy.standardScriptVerifyFlags)
+    (txSigComponent, privateKeys, csvScriptNum, sequence)
+  }
+
+  private def cltvLockTimesOfSameType(generatorComponent : (TransactionSignatureComponent, Seq[ECPrivateKey],  ScriptNumber)) : Boolean = {
```
Christewart: Add comments to this function to explain exactly what it does

src/main/scala/org/bitcoins/core/gen/TransactionGenerators.scala
```diff
@@ -116,24 +124,143 @@ trait TransactionGenerators {
...
+      case positive if positive >= 0 =>
+        if ((tx.lockTime.underlying < threshold && num.underlying < threshold) ||
+          (tx.lockTime.underlying > threshold && num.underlying > threshold)) true
+        else {
+          false
+        }
+    }
+  }
+
+  private def csvValues : Gen[(ScriptNumber, UInt32)] = (for {
```
Christewart: Add comments to explain what this does

src/main/scala/org/bitcoins/core/gen/TransactionGenerators.scala
```diff
@@ -116,24 +124,143 @@ trait TransactionGenerators {
...
+    sequence <- NumberGenerator.uInt32s
+    csvScriptNum <- NumberGenerator.uInt32s.map(x => ScriptNumber(x.underlying))
+  } yield (csvScriptNum, sequence)).suchThat(csvLockTimesOfSameType)
+
+  private def csvLockTimesOfSameType(sequenceNumbers : (ScriptNumber, UInt32)) : Boolean = {
```
Christewart: Same as above

src/main/scala/org/bitcoins/core/gen/TransactionGenerators.scala
```diff
@@ -116,24 +124,143 @@ trait TransactionGenerators {
...
+    //Generate UInt32 values that are less than txLockTime values. UInt32 values are then mapped to ScriptNumbers
+    cltvLockTime <- NumberGenerator.uInt32s.suchThat(num => num < txLockTime).map(x => ScriptNumber(x.underlying))
+    sequence <- NumberGenerator.uInt32s.suchThat(num => num != UInt32.max)
+    (signedScriptSig, cltvScriptPubKey, privateKeys) <- ScriptGenerators.signedCLTVScriptSignature(cltvLockTime, txLockTime, sequence)
+  } yield {
+    val (creditingTx, outputIndex) = buildCreditingTransaction(cltvScriptPubKey)
+    val (signedSpendingTx, inputIndex) = buildSpendingTransaction(TransactionConstants.version, creditingTx, signedScriptSig, outputIndex, txLockTime, sequence)
+    val txSigComponent = TransactionSignatureComponent(signedSpendingTx, inputIndex, cltvScriptPubKey, Policy.standardScriptVerifyFlags)
+    (txSigComponent, privateKeys, cltvLockTime)
+  }).suchThat(cltvLockTimesOfSameType)
```
Christewart: Could you move this `.suchThat(...)` to line 165 so we have something like  

`NumberGenerator.uInt32s.suchThat(num => num < txLockTime && cltvLockTimesOfSameType(..))`?

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,20 +330,126 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+  def scriptPubKeyAfterCLTV : ScriptPubKey = ScriptPubKey(asm.slice(4, asm.length))
+
+  /**
+    * The absolute CLTV-LockTime value (i.e. the output will remain unspendable until this timestamp or block height)
+    * @return
+    */
+  def locktime : UInt32 = {
+    val hex = BitcoinSUtil.flipEndianess(asm(1).hex)
+    UInt32(hex)
+  }}
```
Christewart: Why is there two `}}` here?

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,20 +330,126 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+    * Determines the nested ScriptPubKey inside the CSVScriptPubKey
+    * @return
+    */
+  def scriptPubKeyAfterCSV : ScriptPubKey = ScriptPubKey(asm.slice(4, asm.length))
+
+  /**
+    * The relative CSV-LockTime value (i.e. the amount of time the output should remain unspendable)
+    * @return
+    */
+  def locktime : UInt32 = {
```
Christewart: Shouldn't this technically be of type `ScriptNumber`?

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,20 +330,126 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+    * Determines the nested ScriptPubKey inside the CLTVScriptPubKey
+    * @return
+    */
+  def scriptPubKeyAfterCLTV : ScriptPubKey = ScriptPubKey(asm.slice(4, asm.length))
+
+  /**
+    * The absolute CLTV-LockTime value (i.e. the output will remain unspendable until this timestamp or block height)
+    * @return
+    */
+  def locktime : UInt32 = {
```
Christewart: Shouldn't this technically be of type `ScriptNumber`

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -410,6 +387,72 @@ object P2PKScriptSignature extends Factory[P2PKScriptSignature] {
...
+    case p2pkhScriptPubKey : P2PKHScriptPubKey => CLTVScriptSignatureImpl(P2PKHScriptSignature(sigs.head, pubKeys.head))
+    case multiSigScriptPubKey : MultiSignatureScriptPubKey => CLTVScriptSignatureImpl(MultiSignatureScriptSignature(sigs))
+    case cltvScriptPubKey : CLTVScriptPubKey => apply(cltvScriptPubKey.scriptPubKeyAfterCLTV, sigs, pubKeys, redeemScript)
+    case csvScriptPubKey : CSVScriptPubKey => apply(csvScriptPubKey.scriptPubKeyAfterCSV, sigs, pubKeys, redeemScript)
+    case p2shScriptPubKey : P2SHScriptPubKey =>
+      require(redeemScript.isDefined, "If the underlying scriptSig is a P2SHScriptSignature, a redeemScript must be defined.")
+      val cltvScriptSigBeforeRedeemScript = apply(redeemScript.get, sigs, pubKeys, None)
+      CLTVScriptSignatureImpl(P2SHScriptSignature(cltvScriptSigBeforeRedeemScript, redeemScript.get))
+    case EmptyScriptPubKey => CLTVScriptSignatureImpl(EmptyScriptSignature)
+    case nonstandard : NonStandardScriptPubKey => throw new IllegalArgumentException("A NonStandardScriptSignature cannot be" +
```
Christewart: I'm just leaving a comment here so I can think about this more, not sure if this is the best way to handle this... However we can't really create a program to spend something we don't know how to spend..

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
TomMcCabe: As of latest commit (https://github.com/bitcoin-s/bitcoin-s-core/pull/31/commits/ea2229946e92b46ba815f5ab8485a1e4bda39b82), the respective scriptPubKey generators don't return the ECPrivateKey(s) needed to satisfy the script, rendering the functionality useless for generating signedScriptSignatures. Currently working on this.

src/test/scala/org/bitcoins/core/protocol/script/CLTVScriptPubKeySpec.scala
TomMcCabe: Done in `TransactionSignatureCreatorSpec`, where we generate validly constructed but unspendable CLTV (and CSV) transactions that will fail with either `ScriptErrorUnsatisifedLockTime` or `ScriptErrorPushSize`.

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -45,64 +48,96 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    p2sh = P2SHScriptPubKey(randomScriptPubKey)
+  } yield (p2sh, privKeys)
+
+  def emptyScriptPubKey : Gen [(ScriptPubKey, Seq[ECPrivateKey])] = (EmptyScriptPubKey, Seq())
+
+  def pickRandomNonP2SHScriptPubKey: Gen[(ScriptPubKey, Seq[ECPrivateKey])] = {
+    val randomNum = (scala.util.Random.nextInt() % 5).abs
+    if (randomNum == 0) p2pkScriptPubKey
+    else if (randomNum == 1) p2pkhScriptPubKey
+    else if (randomNum == 2) cltvScriptPubKey.suchThat(!_._1.scriptPubKeyAfterCLTV.isInstanceOf[CSVScriptPubKey])
```
Christewart, line 129: These two lines seem odd, what is the point of doing this?

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -45,64 +48,96 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    else if (randomNum == 3) csvScriptPubKey.suchThat(!_._1.scriptPubKeyAfterCSV.isInstanceOf[CLTVScriptPubKey])
+    else multiSigScriptPubKey
+  }
 
-  def pickRandomNonP2SHScriptPubKey : Gen[ScriptPubKey] = {
+  def pickRandomNonCLTVNonCSVScriptPubKey : Gen[(ScriptPubKey, Seq[ECPrivateKey])] = {
```
Christewart: Could we have a `P2SHScriptPubKey` here?

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -114,19 +149,28 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    *
+    * @param scriptPubKey
+    * @return
+    */
   private def pickCorrespondingScriptSignature(scriptPubKey : ScriptPubKey): Gen[ScriptSignature] = scriptPubKey match {
     case p2pk : P2PKScriptPubKey => p2pkScriptSignature
     case p2pkh : P2PKHScriptPubKey => p2pkhScriptSignature
     case multisig : MultiSignatureScriptPubKey => multiSignatureScriptSignature
     case EmptyScriptPubKey => emptyScriptSignature
+    case cltv : CLTVScriptPubKey => pickCorrespondingScriptSignature(cltv.scriptPubKeyAfterCLTV)
```
Christewart: We need to wrap what is returned from `pickCorrespondingScriptSIgnature` into a `CLTVScriptSignature`, same goes for `CSV`

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -177,8 +223,15 @@ trait ScriptGenerators extends BitcoinSLogger {
   } yield {
     val publicKeys = privateKeys.map(_.publicKey)
     val scriptPubKey = MultiSignatureScriptPubKey(requiredSigs,publicKeys)
+    val (signedScriptSig, _, _) = multiSigHelper(privateKeys, requiredSigs, scriptPubKey)
+    (signedScriptSig, MultiSignatureScriptPubKey(scriptPubKey.hex), privateKeys)
```
Christewart: `MultiSignatureScriptPubKey(scriptPubKey.hex)` is unnecessary, you can just do `scriptPubKey`

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,6 +260,82 @@ trait ScriptGenerators extends BitcoinSLogger {
...
   }
 
+  /**
+    * Generates a signed [[CLTVScriptSignature]] that spends from a [[CLTVScriptPubKey]] correctly
+    *
+    * @return the signed [[CLTVScriptSignature]], the [[CLTVScriptPubKey]] it spends, and the sequences of [[ECPrivateKey]]
+    *         used to sign the scriptSig
+    */
+  def signedCLTVScriptSignature(cltvLockTime : ScriptNumber, lockTime : UInt32, sequence : UInt32) : Gen[(ScriptSignature, CLTVScriptPubKey, Seq[ECPrivateKey])] = for {
+   // (cltv, privKey, num) <- cltvScriptPubKey(scriptNumber)
```
Christewart: Nit: Remove this

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,6 +260,82 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    cltv = CLTVScriptPubKey(cltvLockTime, scriptPubKey)
+  } yield scriptPubKey match {
+    case _ : MultiSignatureScriptPubKey =>
+      val requiredSigs = scriptSig.signatures.size
+      val (cltvScriptSig, _, _) = cltvHelper(lockTime, sequence, cltv, privKeys, Some(requiredSigs))
+      (cltvScriptSig, cltv, privKeys)
+    case _ : P2PKHScriptPubKey | _ : P2PKScriptPubKey =>
+      val (cltvScriptSig, _, _) = cltvHelper(lockTime, sequence, cltv, privKeys, None)
+      (cltvScriptSig, cltv, privKeys)
+    case _ => throw new IllegalArgumentException("We only want to generate P2PK, P2PKH, and MultiSig ScriptSignatures.")
```
Christewart: Don't use a `_`, match on all of the scriptPubKey types. If in the future we could create new valid scriptPubKeys we need to take into consideration for `cltvScriptPubKey`

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,6 +260,82 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+  }
+
+  private def cltvHelper (lockTime : UInt32, sequence : UInt32, cltv: CLTVScriptPubKey, privateKeys : Seq[ECPrivateKey], requiredSigs : Option[Int]) : (CLTVScriptSignature, CLTVScriptPubKey, Seq[ECPrivateKey]) = {
```
Christewart, line 345: Scaladoc on what this function does

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -206,6 +260,82 @@ trait ScriptGenerators extends BitcoinSLogger {
...
+    csv = CSVScriptPubKey(csvScriptNum, scriptPubKey)
+  } yield scriptPubKey match {
+      case _ : MultiSignatureScriptPubKey =>
+          val requiredSigs = scriptSig.signatures.size
+          val (csvScriptSig, _, _) = csvHelper(sequence, csv, privKeys, Some(requiredSigs))
+          (csvScriptSig, csv, privKeys)
+        case _ : P2PKHScriptPubKey | _ : P2PKScriptPubKey =>
+          val (csvScriptSig, _, _) = csvHelper(sequence, csv, privKeys, None)
+          (csvScriptSig, csv, privKeys)
+      case _ => throw new IllegalArgumentException("We only want to generate P2PK, P2PKH, and MultiSig ScriptSignatures.")
```
Christewart: See above comment on using `case _ =>`

src/main/scala/org/bitcoins/core/gen/ScriptGenerators.scala
```diff
@@ -223,6 +353,15 @@ trait ScriptGenerators extends BitcoinSLogger {
     //gets rid of Gen[Gen[...]]
     scriptSig.flatMap(g => g)
   }
+
+
+  private def randomSignedScriptSig : Gen[(ScriptSignature, ScriptPubKey, Seq[ECPrivateKey])] = {
```
Christewart: Can this function include a signed `cltvScriptSignature` and `csvScriptSignature`?

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -410,6 +387,89 @@ object P2PKScriptSignature extends Factory[P2PKScriptSignature] {
...
+    */
+  def apply(scriptPubKey: ScriptPubKey, sigs : Seq[ECDigitalSignature], pubKeys : Seq[ECPublicKey]) : CLTVScriptSignature = scriptPubKey match {
+    case p2pkScriptPubKey : P2PKScriptPubKey => CLTVScriptSignature(P2PKScriptSignature(sigs.head))
+    case p2pkhScriptPubKey : P2PKHScriptPubKey => CLTVScriptSignature(P2PKHScriptSignature(sigs.head, pubKeys.head))
+    case multiSigScriptPubKey : MultiSignatureScriptPubKey => CLTVScriptSignature(MultiSignatureScriptSignature(sigs))
+    case cltvScriptPubKey : CLTVScriptPubKey => apply(cltvScriptPubKey.scriptPubKeyAfterCLTV, sigs, pubKeys)
+    case csvScriptPubKey : CSVScriptPubKey => apply(csvScriptPubKey.scriptPubKeyAfterCSV, sigs, pubKeys)
+    case EmptyScriptPubKey => CLTVScriptSignature(EmptyScriptSignature)
+    case _ : NonStandardScriptPubKey | _ : P2SHScriptPubKey => throw new IllegalArgumentException("A NonStandardScriptSignature or P2SHScriptSignature cannot be" +
+      "the underlying scriptSig in a CLTVScriptSignature. Got: " + this)
```
Christewart: This is an incorrect error message, `this` would return the object we currently in, in this case `CLTVScriptSignature`. You want to return the scriptPubKey we are pattern matching on. You can do something like this 

``` scala

case x @ (_ : NonStandardScriptPubKey | _ : P2SHScriptPubKey) => throw new IllegalArgumentException("A NonStandardScriptSignature or P2SHScriptSignature cannot be" +
 +      "the underlying scriptSig in a CLTVScriptSignature. Got: " + x)
```

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -410,6 +387,89 @@ object P2PKScriptSignature extends Factory[P2PKScriptSignature] {
...
+    * @return
+    */
+  def apply(scriptPubKey: ScriptPubKey, sigs : Seq[ECDigitalSignature], pubKeys : Seq[ECPublicKey]) : CSVScriptSignature = scriptPubKey match {
+    case p2pkScriptPubKey : P2PKScriptPubKey => CSVScriptSignature(P2PKScriptSignature(sigs.head))
+    case p2pkhScriptPubKey : P2PKHScriptPubKey => CSVScriptSignature(P2PKHScriptSignature(sigs.head, pubKeys.head))
+    case multiSigScriptPubKey : MultiSignatureScriptPubKey => CSVScriptSignature(MultiSignatureScriptSignature(sigs))
+    case cltvScriptPubKey : CLTVScriptPubKey => apply(cltvScriptPubKey.scriptPubKeyAfterCLTV, sigs, pubKeys)
+    case csvScriptPubKey : CSVScriptPubKey => apply(csvScriptPubKey.scriptPubKeyAfterCSV, sigs, pubKeys)
+    case EmptyScriptPubKey => CSVScriptSignature(EmptyScriptSignature)
+    case _ : NonStandardScriptPubKey | _ : P2SHScriptPubKey => throw new IllegalArgumentException("A NonStandardScriptSignature or P2SHScriptSignature cannot be" +
```
Christewart: See above comment in `CLTVScriptSignature`

src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
```diff
@@ -60,8 +60,12 @@ trait ScriptInterpreter extends CryptoInterpreter with StackInterpreter with Con
       scriptPubKey match {
         case p2shScriptPubKey : P2SHScriptPubKey if (ScriptFlagUtil.p2shEnabled(program.flags)) =>
           executeP2shScript(scriptSigExecutedProgram, programBeingExecuted, p2shScriptPubKey)
+/*        case csv : CSVScriptPubKey if csv.scriptPubKeyAfterCSV.isInstanceOf[P2SHScriptPubKey] && ScriptFlagUtil.p2shEnabled(program.flags) =>
```
Christewart: Nit: Remove these

src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCreatorSpec.scala
```diff
@@ -37,16 +38,62 @@ class TransactionSignatureCreatorSpec extends Properties("TransactionSignatureCr
...
+    Prop.forAllNoShrink(TransactionGenerators.spendableCLTVTransaction :| "cltv_spendable") {
+      case (txSignatureComponent: TransactionSignatureComponent, _, scriptNumber) =>
+        //run it through the interpreter
+        require(txSignatureComponent.transaction.lockTime.underlying >= scriptNumber.underlying, "TxLocktime must be satisifed so it should be greater than or equal to  " +
+          "the cltv value. Got TxLockTime : " + txSignatureComponent.transaction.lockTime.underlying + " , and cltv Value: " +
+          scriptNumber.underlying)
+        val program = ScriptProgram(txSignatureComponent)
+        val result = ScriptInterpreter.run(program)
+        val cltv = CLTVScriptPubKey(txSignatureComponent.scriptPubKey.hex)
+        Seq(ScriptOk).contains(result)
```
Christewart: Why not just do `result == ScriptOk`?

src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCreatorSpec.scala
```diff
@@ -37,16 +38,62 @@ class TransactionSignatureCreatorSpec extends Properties("TransactionSignatureCr
...
+  property("generate a valid signature for a valid and spendable cltv transaction") =
+    Prop.forAllNoShrink(TransactionGenerators.spendableCLTVTransaction :| "cltv_spendable") {
+      case (txSignatureComponent: TransactionSignatureComponent, _, scriptNumber) =>
+        //run it through the interpreter
+        require(txSignatureComponent.transaction.lockTime.underlying >= scriptNumber.underlying, "TxLocktime must be satisifed so it should be greater than or equal to  " +
+          "the cltv value. Got TxLockTime : " + txSignatureComponent.transaction.lockTime.underlying + " , and cltv Value: " +
+          scriptNumber.underlying)
+        val program = ScriptProgram(txSignatureComponent)
+        val result = ScriptInterpreter.run(program)
+        val cltv = CLTVScriptPubKey(txSignatureComponent.scriptPubKey.hex)
```
Christewart: What is this line for?

src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCreatorSpec.scala
```diff
@@ -37,16 +38,62 @@ class TransactionSignatureCreatorSpec extends Properties("TransactionSignatureCr
...
+  property("generate a valid signature for a validly constructed, but NOT spendable cltv transaction") =
+    Prop.forAllNoShrink(TransactionGenerators.unspendableCLTVTransaction :| "cltv_unspendable") {
+      case (txSignatureComponent: TransactionSignatureComponent, _, scriptNumber) =>
+        //run it through the interpreter
+        require(txSignatureComponent.transaction.lockTime.underlying < scriptNumber.underlying, "TxLocktime must not be satisifed so it should be less than " +
+          "the cltv value. Got TxLockTime : " + txSignatureComponent.transaction.lockTime.underlying + " , and cltv Value: " +
+          scriptNumber.underlying)
+        val program = ScriptProgram(txSignatureComponent)
+        val result = ScriptInterpreter.run(program)
+        val cltv = CLTVScriptPubKey(txSignatureComponent.scriptPubKey.hex)
```
Christewart: See above, why is this here?

src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCreatorSpec.scala
```diff
@@ -37,16 +38,62 @@ class TransactionSignatureCreatorSpec extends Properties("TransactionSignatureCr
...
+  property("generate a valid signature for a valid and spendable csv transaction") =
+    Prop.forAllNoShrink(TransactionGenerators.spendableCSVTransaction :| "spendable csv") {
+      case (txSignatureComponent: TransactionSignatureComponent, keys, scriptNumber, sequence) =>
+        //run it through the interpreter
+        val program = ScriptProgram(txSignatureComponent)
+        val result = ScriptInterpreter.run(program)
+        Seq(ScriptOk, ScriptErrorPushSize).contains(result)
+    }
+
+  property("generate a valid signature for a validly constructed but unspendable csv transaction") =
```
Christewart: We don't get to check the signature in this case, so I would probably just name this property `fail to verify a transaction with a csv value that hasn't passed yet`

src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCreatorSpec.scala
```diff
@@ -37,16 +38,62 @@ class TransactionSignatureCreatorSpec extends Properties("TransactionSignatureCr
...
+        require(txSignatureComponent.transaction.lockTime.underlying >= scriptNumber.underlying, "TxLocktime must be satisifed so it should be greater than or equal to  " +
+          "the cltv value. Got TxLockTime : " + txSignatureComponent.transaction.lockTime.underlying + " , and cltv Value: " +
+          scriptNumber.underlying)
+        val program = ScriptProgram(txSignatureComponent)
+        val result = ScriptInterpreter.run(program)
+        val cltv = CLTVScriptPubKey(txSignatureComponent.scriptPubKey.hex)
+        Seq(ScriptOk).contains(result)
+    }
+
+  property("generate a valid signature for a validly constructed, but NOT spendable cltv transaction") =
```
Christewart: Since we don't check the validity of the signature, we probably shouldn't put it in the name of the test case. Should be something like `fail to verify a transaction whose locktime has not passed yet`

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,24 +327,174 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+  * Adds an absolute/defined locktime condition to any scriptPubKey.
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki]]
+  * Format: <locktime> OP_CLTV OP_DROP <scriptPubKey>
+  */
+sealed trait CLTVScriptPubKey extends ScriptPubKey {
+  /**
+    * Determines the nested ScriptPubKey inside the CLTVScriptPubKey
+    * @return
+    */
+  def scriptPubKeyAfterCLTV : ScriptPubKey = ScriptPubKey(asm.slice(4, asm.length))
```
Christewart: I think there is a bug here in the case that we have an `OP_1` as the leading pushop, or any other `OP_n` operation. You handle that case inside of the `isCLTVScriptPubKey` function, but not here

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -316,24 +327,174 @@ object P2PKScriptPubKey extends Factory[P2PKScriptPubKey] {
...
+  * Adds a relative lockTime condition to any scriptPubKey.
+  * https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki
+  * Format: <locktime> OP_CSV OP_DROP <scriptPubKey>
+  */
+sealed trait CSVScriptPubKey extends ScriptPubKey {
+  /**
+    * Determines the nested ScriptPubKey inside the CSVScriptPubKey
+    * @return
+    */
+  def scriptPubKeyAfterCSV : ScriptPubKey = ScriptPubKey(asm.slice(4, asm.length))
```
Christewart: Same comment as above wrt `scriptPubKeyAfterCLTV`. Add test cases to make sure this stuff actually gets tested correctly.

src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -353,16 +514,24 @@ object ScriptPubKey extends Factory[ScriptPubKey] with BitcoinSLogger {
...
+    case _ if CLTVScriptPubKey.isCLTVScriptPubKey(asm) => CLTVScriptPubKey(asm)
+    case _ if CSVScriptPubKey.isCSVScriptPubKey(asm) => CSVScriptPubKey(asm)
+    case _ => NonStandardScriptPubKey(asm)
   }
 
   def fromBytes(bytes : Seq[Byte]) : ScriptPubKey = RawScriptPubKeyParser.read(bytes)
 
+  def apply(asm : Seq[ScriptToken]) : ScriptPubKey = fromAsm(asm)
+
+  private def isScriptNumberCorrectAsm(num : ScriptNumber) : Boolean = {
```
Christewart: Is this used any where?

src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -462,6 +519,7 @@ object ScriptSignature extends Factory[ScriptSignature] with BitcoinSLogger {
     case s : P2PKScriptPubKey => P2PKScriptSignature.fromAsm(tokens)
     case s : MultiSignatureScriptPubKey => MultiSignatureScriptSignature.fromAsm(tokens)
     case s : NonStandardScriptPubKey => NonStandardScriptSignature.fromAsm(tokens)
+    case s : CLTVScriptPubKey => fromScriptPubKey(tokens, s.scriptPubKeyAfterCLTV)
     case EmptyScriptPubKey if (tokens.size == 0) => EmptyScriptSignature
```
Christewart: This guard here breaks pattern pattern matching exhaustiveness, see #32 . Need to refactor this to be like 

``` scala
case EmptyScriptPubKey => 
  if (tokens.size==0) EmptyScriptSignature else NonStandardScriptSignature(tokens)
```

also we do no match on a `CSVScriptPubKey` in this pattern match

src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -102,57 +98,35 @@ sealed trait ExecutionInProgressScriptProgram extends ScriptProgram {
...
-   * Implentation type for a script program that has not been executed at all
- *
-   * @param txSignatureComponent
-   * @param stack
-   * @param script
-   * @param originalScript
-   * @param altStack
-   * @param flags
+   * Implementation type for a [[PreExecutionScriptProgram]] - a [[ScriptProgram]] that has not yet begun being
+    * evaluated  by the [[org.bitcoins.core.script.interpreter.ScriptInterpreter]].
```
Christewart, line 113: Good idea on adding better docs for these. The old ones sucked.

src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -300,6 +308,12 @@ trait BitcoinScriptUtil {
     }
     return true
   }
+
+  def correctScriptNumberRepresentation(num : ScriptNumber) : Option[ScriptOperation] = {
```
Christewart: This looks a little hairy. I thought about this more and I think we should change the name to `minimalScriptNumberRepresentation` as that is a little less ambiguous compared to `correctScriptNumberRepresentation`. Also, I think we can just return a `ScriptNumber` instead of an `Option[ScriptOperation]`. We can just assume this function returns `num` if it is the minimally encoded number, or else it transforms `num` into the appropriate `ScriptNumberOperation`. [Technically, in our type hierarchy, `ScriptNumberOperation` extends ScriptNumbe](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/src/main/scala/org/bitcoins/core/script/constant/Constants.scala#L211)r, so we can get away with just returning a ScriptNumber.  I can help with this if needed.

src/test/scala/org/bitcoins/core/util/BitcoinScriptUtilTest.scala
```diff
@@ -183,4 +183,17 @@ class BitcoinScriptUtilTest extends FlatSpec with MustMatchers {
...
+    val scriptNum100 = ScriptNumber(100)
+    val scriptNum10 = ScriptNumber(10)
+    val scriptNumZero = ScriptNumber(0)
+    val scriptNum16 = ScriptNumber(16)
+    val scriptNum17 = ScriptNumber(17)
+    BitcoinScriptUtil.correctScriptNumberRepresentation(scriptNum100) must be (None)
+    BitcoinScriptUtil.correctScriptNumberRepresentation(scriptNum10) must be (Some(OP_10))
+    BitcoinScriptUtil.correctScriptNumberRepresentation(scriptNumZero) must be (Some(OP_0))
+    BitcoinScriptUtil.correctScriptNumberRepresentation(scriptNum16) must be (Some(OP_16))
+    BitcoinScriptUtil.correctScriptNumberRepresentation(scriptNum17) must be (None)
```
Christewart: Maybe add a test case for -1?

Christewart:
ACK

PR #33, opened 2016-08-31
Add hash to address

Christewart:
Refactors `Address` types in our codebase to store the `hash` and `networkParameters` they were originally derived from.


TomMcCabe:
Not all the SciptPubKey traits were changed to `sealed`

TomMcCabe:
ACK.

PR #34, opened 2016-09-03
Modifying build.sbt to exclude files from jar

Christewart:
Previously when we would run `sbt assembly` we would include files that would be irrelevant to subsequent projects using our core project, this pull request eliminates some redundant files from being included into our `bitcoin-s-core-assembly-0.0.1.jar` file


TomMcCabe:
ACK.

PR #37, opened 2016-09-22
fix endianness typo

TomMcCabe:
Simple typo fix. "Endianness" has 2 `n`'s

https://en.wikipedia.org/wiki/Endianness


Christewart:
Well that is embarrassing :/

Christewart:
ACK

PR #39, opened 2016-10-07
Moving spray dependency to 'test' scope

Christewart:
This pull request removes the few source files that still depended on `spray-json`, the functionality in those sources files are, what I consider, out of the scope of this project. We do use `spray-json` heavily in our tests though, that is why it is moved to a test dependency instead of being totally removed.


TomMcCabe:
ACK.

PR #41, opened 2016-10-20
Adding helper function to generate a valid chain with a specific star…

Christewart:
Add a simple generator helper function to specify a start header for a valid chain.


TomMcCabe:
ACK.

PR #44, opened 2016-12-19
Segwit sighash tests

Christewart:
Bitcoin Core did not add any test vectors to [`sighash.json`](https://github.com/bitcoin/bitcoin/blob/master/src/test/data/sighash.json), therefore this pull request simply changes our generators from using a constant hash type (`SIGHASH_ALL`) to using a generated hash type ([`CryptoGenerators.hashType`](https://github.com/Christewart/bitcoin-s-core/blob/segwit_sighash_tests/src/main/scala/org/bitcoins/core/gen/CryptoGenerators.scala#L91))


TomMcCabe:
ACK pending successful Travis build

PR #46, opened 2016-12-24
Wtxid and commitment structure

Christewart:
This pull request implements the new wtxid for transactions and a commitment structure inside of the coinbase transaction for segregated witness. More information on the commitment structure can be found inside of [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#commitment-structure)


TomMcCabe:
ACK.

PR #47, opened 2016-12-28
Refactor flag checks to sig checker

Christewart:
The way we check digital signatures in Bitcoin depends on what [ScriptFlags](https://github.com/bitcoin-s/bitcoin-s-core/blob/464964d4a1838ff946c9a8417ee3dfd7afa5980b/src/main/scala/org/bitcoins/core/script/flag/ScriptFlags.scala#L11-L137) are set when executing a [ScriptProgram](https://github.com/bitcoin-s/bitcoin-s-core/blob/464964d4a1838ff946c9a8417ee3dfd7afa5980b/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala#L15-L83). This pull request aims to refactor some of the flag checking logic from [CryptoInterpreter](https://github.com/bitcoin-s/bitcoin-s-core/blob/464964d4a1838ff946c9a8417ee3dfd7afa5980b/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala#L80-L308) to [TransactionSignatureChecker](https://github.com/Christewart/bitcoin-s-core/blob/655e10a9d86769761f8799d8e74da045424530ef/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala#L37-L144) and remove redundant checks that were happening in both places

### DIFF
--- a/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
+++ b/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
@@ -68,9 +68,10 @@ trait TransactionSignatureChecker extends BitcoinSLogger {
           TransactionSignatureSerializer.hashForSignature(w.transaction,w.inputIndex,sigsRemovedScript, hashType, w.amount,w.sigVersion)
       }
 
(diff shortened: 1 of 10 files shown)


TomMcCabe:
ACK.

PR #52, opened 2017-02-13
Refactoring dependencies to only include scalacheck & spongycastle in…

Christewart:
…side a fat jar

Moving dependencies around

This pull request adjusts the scope of various dependencies. For instance, previously we had a scalatest dependency included in the fat jar, we now added the `"test"` scope so it is only used as a dependency when we run test cases.


Christewart:
@TomMcCabe

TomMcCabe:
tACK

Issue #57, opened 2017-02-16
Failure to correctly type P2PKHScriptSignatures


Christewart:
# Problem

Every `P2PKHScriptSignature` should follow the following format: 

```<sig> <pubkey>``` 

Currently we have a naive typing algorithm for classifying `P2PKHScriptSignature`. The algorithm is defined [here](https://github.com/bitcoin-s/bitcoin-s-core/blob/1f1deb3cbace3c154e927eba54f6dbf54c636eb6/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala#L111-L119)

```scala
  /** Determines if the given asm matches a [[P2PKHScriptSignature]] */
  def isP2PKHScriptSig(asm: Seq[ScriptToken]): Boolean = asm match {
    case List(w : BytesToPushOntoStack, x : ScriptConstant, y : BytesToPushOntoStack,
      z : ScriptConstant) =>
      //TODO: We need to change this to use CPubKey::IsFullyValid() when we integrate libsecp256k1
      //this checks to make sure we do not have a redeem script as the 'z' constant
      //for instance, we can have a p2sh scriptsignature that has a redeem script for a p2pkhScriptSig
      //and it will have the same format as the pattern match before
      !P2SHScriptSignature.isRedeemScript(z)
    case _ => false
}
```
as you can see, we pattern match on the given asm tokens and see if we have a script signature of the format `<pushop for sig> <sig> <pushop for pubkey> <pubkey>`

This works in 99% of cases, _except_ when we have a `P2SHScriptSignature` which is required to only be pushops. We can _nest_ a `P2PKScriptPubKey` inside of a `P2SHScriptSignature` to have the _exact same_ format as above. The `P2SHScriptSignature` looks like:

`<pushop for sig> <sig> <pushop for redeemScript> <P2PK redeemScript>`

This causes `isP2PKHScriptSig` to return false, which is the cause for this travis failure: 

https://travis-ci.org/bitcoin-s/bitcoin-s-core/builds/201652191

# Solution

The most obvious solution is after we add libsecp256k1 we add a check inside of `isP2PKHScriptSig` to detect if the constant `z` is a valid public key. If it is a valid public key, we assume that the given asm is _not_ a `P2PKHScriptSignature`

[...]

PR #61, opened 2017-02-22
Implementing fix to p2pkh script sig bug

Christewart:
This pull request implements a fix for #57 

This was being caused by the wrong classification for `CSVScriptPubKey` and a `CLTVScriptPubKey`. Some times our `ScriptParser` got lucky and parsed the script after the locktime op codes. 

We fixed this by implementing a test to see what op codes we have after the lock time op codes. For instance, if we have a `UndefinedOP_NOP` we can assume that the script is invalid. This should fix 99% of these cases.


Christewart:
@TomMcCabe Review this one next

TomMcCabe:
tACK.

PR #67, opened 2017-03-08
Fixing bug in calculating pushops for ScriptNumberOperations

Christewart:
ScriptNumberOperations do not need a push operation for them to be pushed onto the stack, it simply pushes the number itself onto the stack.

For instance, if the interpreter sees an OP_1 it simply pushes the number 1 onto the stack. 

This also fixes a bug where if we have a script of the format: 
`
OP_PUSHDATA1 OP_3 <83 byte constant> `

It will now interpret OP_3 as an 83 byte push operation, instead of pushing the number 3 onto the stack


Christewart:
@TomMcCabe

TomMcCabe:
ACK

Issue #68, opened 2017-03-16
Refactoring ScriptProgram companion object


Christewart:
Currently we have three states inside of a ScriptProgram

`PreExecutionScriptProgram` - which is just a ScriptProgram whose execution hasn't started
`ExecutionInProgressScriptProgram` - a ScriptProgram whose execution is in progress, which just means it is currently being run through the ScriptInterpreter
`ExecutedScriptProgram` - a Script Program whose execution has finished. 

We currently have a mess of functions that is used to modify the state of a `ScriptProgram`. This is not maintainable and is hard to reason about. We need to eventually think about a better way to handle this -- or figure out a way to make this more readable. Here is the offending code: 

https://github.com/bitcoin-s/bitcoin-s-core/blob/a0062e33a97cc4091a2e76370b9c2b4ad40f9585/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala#L107-L339

Probably would be a good idea to break the three states of `PreExecutionScriptProgram`, `ExecutionInProgresScriptProgram` and `ExecutedScriptProgram` into three separate companion objects. Maybe higher order functions would be useful for modifying state in `ExecutionInProgressScriptProgram`?

PR #72, opened 2017-04-12
Refactoring WitnessTransactionSignatureComponent to contain a 'Witnes…

Christewart:
…sTransaction' type instead of a 'Transaction' type

`WitnessV0TransactionSignatureComponent` use to have the type `Transaction` instead `WitnessTransaction` for the tx we are verfiying. Since `WitnessV0TransactionSignatureComponent` is specifically dedicated to verify `WitnessTransaction` we should require that the tx be a `WitnessTransaction`


Christewart:
This starts to address the issues in #68

PR #117, opened 2018-03-05
Adding TransactionSignatureCreator.createSig API that allows for pass…

Christewart:
…ing in a 'sign' function that is an abstraction of the signing process, useful for hardware wallets

As the commit comment says, this pull request exposes a abstraction in `TransactionSignatureCreator` that could allow integrate hardware wallets into bitcoin-s. Now there is a `TransactionSignatureCreator.createSig` method that can take in a `sign` function from `Seq[Byte] -> ECDigitalSignature`. Fundamentally, a hardware wallet takes in a `Seq[Byte]` and spits back out a digital signature. A person looking to integrate Trezor/Ledger support into bitcoin-s could use this to integrate the hardware wallet signing process with the wider bitcoin-s-core codebase.


Christewart:
Note, we used this new API to implement our old `createSig` function

https://github.com/bitcoin-s/bitcoin-s-core/pull/117/files#diff-cc91a80e3d4d4efe5cf741d91d41d65aR20 

This API will be used inside of `TxBuilder` in #113, so instead of passing in just an `ECPrivateKey` -- which you don't have access to if you are using a hardware wallet -- you can pass in a function of type `Seq[Byte] -> ECDigitalSignature` to sign the message.

Something to think about is modifying this API to be `Seq[Byte] -> Future[ECDigitalSignature]` as the signing process may rely on user input which could take a long time.

PR #135, opened 2018-03-28
Adding scalariform formatting preferences

mecampbellsoup:
Inspired by the `sbt-scalariform` plugin
[itself](https://github.com/sbt/sbt-scalariform/blob/master/build.sbt#L39-L48).

@Christewart I think this is a better approach than having things reformatted surprisingly, like 
what happened with https://github.com/bitcoin-s/bitcoin-s-core/pull/134.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,6 +27,20 @@ lazy val appDependencies = Seq(
   "io.spray" %% "spray-json" % sprayV  % "test"
 )
 
+com.typesafe.sbt.SbtScalariform.ScalariformKeys.preferences := {
+  import scalariform.formatter.preferences._
+  FormattingPreferences()
+    .setPreference(FirstParameterOnNewline, Preserve)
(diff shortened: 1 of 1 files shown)


Christewart:
agreed, I don't like the fact that things can be formatted unexpectedly that causes huge diffs

Christewart:
I'm not sure what the best formatting stuff is though. It seems like those settings have weird things like 
```
if () { 
} 
else if () { 

} 
else {

}
```

mecampbellsoup:
Yea, I figure the actual declared formatting preferences will evolve overtime as things go.


On March 28, 2018 at 12:53:48 PM, Chris Stewart (notifications@github.com) wrote:

I'm not sure what the best formatting stuff is though. It seems like those settings have weird things like

if () {  
}  
else if () {  

}  
else {

}

Christewart:
I'll look at this more closely later

PR #156, opened 2018-05-02
Refactor multi-module build

dwhjames:
- separate Java source copied from libsecp256k1 into separate module
- split main module into core core-gen core-test
  - generators can be published separated to core code for use in 3rd-party tests
    `"org.bitcoins" %% "bitcoin-s-core-gen" % "0.0.1-SNAPSHOT" % "test" classifier "tests"`
- code coverage needs to be aggregated, so minor edit to travis file


Christewart:
tack 20c4732

PR #171, opened 2018-06-05
Rpc

nkohen:
An RPC client for bitcoin-s

Implementation is based on bitcoind version 0.16.0

See https://bitcoin.org/en/developer-reference#rpcs

Travis continuous integration still needs configuration.


rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+    bitcoindCall[BitcoinAddress]("addwitnessaddress", List(JsString(address.toString)))
+  }
+
+  def backupWallet(destination: File): Future[Unit] = {
+    bitcoindCall[Unit]("backupwallet", List(JsString(destination.getAbsolutePath)))
+  }
+
+  def bumpFee(txid: DoubleSha256Digest, confTarget: Int = 6, totalFee: Option[Satoshis], replaceable: Boolean = true): Future[BumpFeeResult] = {
+    val options =
+      if (totalFee.nonEmpty)
```
Christewart: Generally I prefer to wrap expressions in if statements in a `{}`

this would be 

```scala
if (totalFee.nonEmpty) {
   //stuff here
} else {
  //else stuff here
}

```

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  def dumpWallet(file: File): Future[Unit] = {
+    bitcoindCall[Unit]("dumpwallet", List(JsString(file.getAbsolutePath)))
+  }
+
+  def encryptWallet(passphrase: String): Future[String] = {
+    bitcoindCall[String]("encryptwallet", List(JsString(passphrase)))
+  }
+
+  def estimateFee(blocks: Int): Future[Satoshis] = {
```
Christewart: Let's return this here: 

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala#L22

EDIT:

Just realized that the `SatoshisPerByte` and `SatoshisPerVirtualByte` do not extend `BitcoinFeeUnit`. That is low hanging fruit if you want to submit a small pr to fix that.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+    bitcoindCall[Bitcoins]("estimatefee", List(JsNumber(blocks)))
+      .map(_.satoshis)
+  }
+
+  def estimatePriority(blocks: Int): Future[BigDecimal] = {
+    bitcoindCall[BigDecimal]("estimatepriority", List(JsNumber(blocks)))
+  }
+
+  // This needs a home?
+  case class FundRawTransactionOptions(
```
Christewart: Maybe create a `RPCOpts.scala` and place all things in there? Might help people using the library if you reference it as `RPCOpts.FundRawTransactionOptions(...)`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def getAccountAddress(account: String = ""): Future[Address] = {
+    bitcoindCall[Address]("getaccountaddress", List(JsString(account)))
+  }
+
+  def getAccount(address: Address): Future[String] = {
+    bitcoindCall[String]("getaccount", List(JsString(address.toString)))
+  }
+
+  def getAddedNodeInfo(details: Boolean, node: Option[InetAddress]): Future[Vector[Node]] = {
+    val params = if (node.isEmpty) List(JsBoolean(details)) else List(JsBoolean(details), JsString(node.get.toString))
```
Christewart: multiline if statements please

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+    subtractFeeFromOutputs: Option[Array[Int]])
+
+  implicit val fundRawTransactionOptionsWrites: Writes[FundRawTransactionOptions] = Json.writes[FundRawTransactionOptions]
+
+  def fundRawTransaction(transaction: Transaction, options: Option[FundRawTransactionOptions]): Future[FundRawTransactionResult] = {
+    bitcoindCall[FundRawTransactionResult]("fundrawtransaction", List(JsString(transaction.hex), Json.toJson(options)))
+  }
+
+  def generateToAddress(blocks: Int, address: Address, maxTries: Int = 1000000): Future[Vector[DoubleSha256Digest]] = {
+    bitcoindCall[Vector[DoubleSha256Digest]]("generatetoaddress", List(JsNumber(blocks), JsString(address.toString), JsNumber(maxTries)))
```
Christewart: I think we should probably have this format for argument passing: 

```scala
bitcoindCall[Vector[DoubleSha256Digest]](
"generatetoaddress", 
List(
  JsNumber(blocks), 
  JsString(address.toString), 
  JsNumber(maxTries))
)
```

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def getBalance(account: String = "*", minConfirmations: Int = 0, includeWatchOnly: Boolean = true): Future[Bitcoins] = {
+    bitcoindCall[Bitcoins]("getbalance", List(JsString(account), JsNumber(minConfirmations), JsBoolean(includeWatchOnly)))
+  }
+
+  def getBlockRaw(headerHash: DoubleSha256Digest): Future[Block] = {
+    bitcoindCall[Block]("getblock", List(JsString(headerHash.hex), JsNumber(0)))
+  }
+
+  def getBlock(headerHash: DoubleSha256Digest): Future[GetBlockResult] = {
+    bitcoindCall[GetBlockResult]("getblock", List(JsString(headerHash.hex), JsNumber(1)))
```
Christewart: let's create a value for this.

```scala
val isJsonObject = JsNumber(1)
```

and then pass it into the rpc call. This is a little more explicit as to what the magic number is.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  def getBlockWithTransactions(headerHash: DoubleSha256Digest): Future[GetBlockWithTransactionsResult] = {
+    bitcoindCall[GetBlockWithTransactionsResult]("getblock", List(JsString(headerHash.hex), JsNumber(2)))
```
Christewart: same as comment above wrt `JsNumber(2)`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  def getMemPoolInfo: Future[GetMemPoolInfoResult] = {
+    bitcoindCall[GetMemPoolInfoResult]("getmempoolinfo")
+  }
+
+  def getNetworkHashPS(blocks: Int = 120, height: Int = -1): Future[Int] = {
+    bitcoindCall[Int]("getnetworkhashps", List(JsNumber(blocks), JsNumber(height)))
+  }
+
+  def getRawChangeAddress: Future[P2PKHAddress] = {
```
Christewart: Type this as a `Future[BitcoinAddress]`. It isn't guaranteed that we will always get a `P2PKHAddress` back.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def getReceivedByAddress(address: BitcoinAddress, minConfirmations: Int): Future[Bitcoins] = {
+    bitcoindCall[Bitcoins]("getreceivedbyaddress", List(JsString(address.toString), JsNumber(minConfirmations)))
+  }
+
+  def getTransaction(txid: DoubleSha256Digest, watchOnly: Boolean = false): Future[GetTransactionResult] = {
+    bitcoindCall[GetTransactionResult]("gettransaction", List(JsString(txid.hex), JsBoolean(watchOnly)))
+  }
+
+  def getTxOutProof(txids: Vector[DoubleSha256Digest], headerHash: Option[DoubleSha256Digest]): Future[MerkleBlock] = {
+    def params =
```
Christewart:
```scala 
  val params = {
    val hashes = txids.map(hash => JsString(hash.hex))
    if (headerHash.isEmpty) {
        List(JsArray(hashes))
      } else {
        List(
             JsArray(hashes),
             JsString(headerHash.get.hex)
         ) 
      } 
  }
```

Christewart: My spacing is off, but i'm sure you get the gist.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // This needs a home
+  case class RpcAddress(
+    address: BitcoinAddress,
+    balance: Bitcoins,
+    account: Option[String])
+
+  implicit object RpcAddressReads extends Reads[RpcAddress] {
+    def reads(json: JsValue): JsResult[RpcAddress] = json match {
```
Christewart: This will need to be cleaned up

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def listSinceBlock(headerHash: DoubleSha256Digest, confirmations: Int = 1, includeWatchOnly: Boolean = false): Future[ListSinceBlockResult] = {
+    bitcoindCall[ListSinceBlockResult]("listsinceblock", List(JsString(headerHash.hex), JsNumber(confirmations), JsBoolean(includeWatchOnly)))
+  }
+
+  def listTransactions(account: String = "*", count: Int = 10, skip: Int = 0, includeWatchOnly: Boolean = false): Future[Vector[ListTransactionsResult]] = {
+    bitcoindCall[Vector[ListTransactionsResult]]("listtransactions", List(JsString(account), JsNumber(count), JsNumber(skip), JsBoolean(includeWatchOnly)))
+  }
+
+  def listUnspent(minConfirmations: Int = 1, maxConfirmations: Int = 9999999, addresses: Vector[BitcoinAddress]): Future[Vector[UnspentOutput]] = {
+    bitcoindCall[Vector[UnspentOutput]]("listunspent", List(JsNumber(minConfirmations), JsNumber(maxConfirmations), JsArray(addresses.map(address => JsString(address.value)))))
```
Christewart: It's best to usually do something like this 

```scala
val addressString = addresses.map(address => JsString(address.value))
```

and then pass that as a parameter to `JsArray`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def pruneBlockChain(height: Int): Future[Int] = {
+    bitcoindCall[Int]("pruneblockchain", List(JsNumber(height)))
+  }
+
+  def removePrunedFunds(txid: DoubleSha256Digest): Future[Unit] = {
+    bitcoindCall[Unit]("removeprunedfunds", List(JsString(txid.hex)))
+  }
+
+  def sendFrom(fromAccount: String, toAddress: BitcoinAddress, amount: Bitcoins, confirmations: Int = 1, comment: String = "", toComment: String = ""): Future[DoubleSha256Digest] = {
+    bitcoindCall[DoubleSha256Digest]("sendfrom", List(JsString(fromAccount), JsString(toAddress.value), JsNumber(amount.toBigDecimal), JsNumber(confirmations), JsString(comment), JsString(toComment)))
```
Christewart: make this nicer to read with one parameter per line

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  def setBan(address: InetAddress, command: String, banTime: Int = 86400, absolute: Boolean = false): Future[Unit] = {
+    bitcoindCall[Unit]("setban", List(JsString(address.getHostAddress), JsString(command), JsNumber(banTime), JsBoolean(absolute)))
+  }
+
+  def setNetworkActive(activate: Boolean): Future[Unit] = {
+    bitcoindCall[Unit]("setnetworkactive", List(JsBoolean(activate)))
+  }
+
+  def setTxFee(feePerKB: Bitcoins): Future[Boolean] = {
```
Christewart: Take in a `BitcoinFeeUnit` here, and then convert what the user passed in to `BitcoinsPerKB`.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  def signMessage(address: P2PKHAddress, message: String): Future[String] = {
+    bitcoindCall[String]("signmessage", List(JsString(address.value), JsString(message)))
+  }
+
+  def signMessageWithPrivKey(key: ECPrivateKey, message: String): Future[String] = {
+    bitcoindCall[String]("signmessagewithprivkey", List(JsString(key.toWIF(network)), JsString(message)))
+  }
+
+  case class SignRawTransactionOutputParameter(
```
Christewart: Perhaps refactor to `RpcOpts` file?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+    txid: DoubleSha256Digest,
+    vout: Int,
+    scriptPubKey: ScriptPubKey,
+    reedemScript: Option[ScriptPubKey])
+
+  implicit val signRawTransactionOutputParameterWrites: Writes[SignRawTransactionOutputParameter] = Json.writes[SignRawTransactionOutputParameter]
+
+  def signRawTransaction(transaction: Transaction, utxoDeps: Vector[SignRawTransactionOutputParameter], keys: Option[Vector[ECPrivateKey]] = None, sigHash: Option[String] = None): Future[SignRawTransactionResult] = {
+    val params =
```
Christewart: We can work on cleaning this up. Try to create intermediate states for values and not just pass them directly into the `JsString,JsArray` objects.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  def stop: Future[String] = {
+    bitcoindCall[String]("stop")
+  }
+
+  def submitBlock(block: Block, parameters: Option[JsValue] = None): Future[Unit] = {
+    val params =
+      if (parameters.isEmpty)
+        List(JsString(block.hex))
+      else
+        List(JsString(block.hex), parameters.get)
```
Christewart: I remember this was a weird rpc call, just making a note to come back to this.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  def verifyChain(level: Int, blocks: Int): Future[Boolean] = {
+    bitcoindCall[Boolean]("verifychain", List(JsNumber(level), JsNumber(blocks)))
+  }
+
+  def verifyMessage(address: P2PKHAddress, signature: String, message: String): Future[Boolean] = {
+    bitcoindCall[Boolean]("verifymessage", List(JsString(address.value), JsString(signature), JsString(message)))
+  }
+
+  def verifyTxOutProof(proof: String): Future[Vector[DoubleSha256Digest]] = {
```
Christewart: I believe the parameter should be `proof: MerkleBlock`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  // Uses string:object pattern - GetMemoryInfo, GetMemPoolAncestors, GetMemPoolDescendants, GetNetTotals, GetPeerInfo, GetRawMemPoolWithTransactions, GetTxOut, ImportMulti, ListAccounts, SendMany
+  // Also Skipped: GetBlockTemplate
+  // TODO: Overload calls with Option inputs?
+  // --------------------------------------------------------------------------------
+  // EVERYTHING BELOW THIS COMMENT HAS TESTS
+
+  // This needs a home
+  implicit object TransactionInputWrites extends Writes[TransactionInput] {
```
Christewart: Do we have issues if we move all of these `implicit object ... ` to another file? Cannot find the implicit definition?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+    result match {
+      case res: JsSuccess[T] => res.value
+      case res: JsError =>
+        logger.error(JsError.toJson(res).toString())
+        throw new IllegalArgumentException(s"Could not parse JsResult: $res")
+    }
+  }
+
+  private def getPayload(response: HttpResponse): Future[JsValue] = {
+    val payloadF = response.entity.dataBytes.runFold(ByteString(""))(_ ++ _)
```
Christewart: We should come back and evaluate the `.runFold` stuff.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  def buildRequest(methodName: String, params: JsArray): HttpRequest = {
+    val m: Map[String, JsValue] = Map(
+      "method" -> JsString(methodName),
+      "params" -> params,
+      "id" -> JsString("")) // java.util.UUID
+    val jsObject = JsObject(m)
+
+    val uri = "http://localhost:18443"
```
Christewart: need to come back and configure this stuff.

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -0,0 +1,49 @@
...
+  walletname: String,
+  walletversion: Int,
+  balance: Bitcoins,
+  unconfirmed_balance: Bitcoins,
+  immature_balance: Bitcoins,
+  txcount: Int,
+  keypoololdest: UInt32,
+  keypoolsize: Int,
+  keypoolsize_hd_internal: Int,
+  paytxfee: Bitcoins,
```
Christewart: This type might need adjusted to `BitcoinFeeUnit`.

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -0,0 +1,49 @@
...
+  hdmasterkeyid: Sha256Hash160Digest,
+  unlocked_until: Option[Int]) extends WalletResult
+case class BumpFeeResult(
+  txid: DoubleSha256Digest,
+  origfee: Bitcoins,
+  fee: Bitcoins,
```
Christewart: `BitcoinFeeUnit` probably

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,201 @@
...
+import org.bitcoins.core.protocol.blockchain.{ Block, BlockHeader, MerkleBlock }
+import org.bitcoins.core.protocol.script.{ ScriptPubKey, ScriptSignature }
+import org.bitcoins.core.protocol.transaction._
+import play.api.libs.json._
+
+import scala.util.{ Failure, Success }
+
+object JsonReaders {
+  implicit object BigIntReads extends Reads[BigInt] {
+    def reads(json: JsValue) = json match {
```
Christewart: Can we add return types and the `override` modifier to all of the `reads` method?

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,201 @@
...
+      case JsNumber(n) => JsSuccess(n.toBigInt())
+      case err => JsError(s"error.expected.jsnumber, got ${Json.toJson(err).toString()}")
```
Christewart: Note: You are doing 

```
case _ => 
```
here which means if another algebraic data type is added to `JsValue` we will not receive pattern match warnings.

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,201 @@
...
+  implicit object Sha256Hash160DigestReads extends Reads[Sha256Hash160Digest] {
+    def reads(json: JsValue) = json match {
+      case JsString(s) => JsSuccess(Sha256Hash160Digest.fromHex(s))
+      case err => JsError(s"error.expected.jsstring, got ${Json.toJson(err).toString()}")
+    }
+  }
+
+  implicit object ECPublicKeyReads extends Reads[ECPublicKey] {
+    def reads(json: JsValue) = json match {
+      case JsString(s) => JsSuccess(ECPublicKey.fromHex(s)) // Should this be WIF or Hex?
```
Christewart: public keys do not have a WIF format, so you have this correct.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,727 @@
...
+    bitcoindCall[Unit]("dumpwallet", List(JsString(file.getAbsolutePath)))
+  }
+
+  def encryptWallet(passphrase: String): Future[String] = {
+    bitcoindCall[String]("encryptwallet", List(JsString(passphrase)))
+  }
+
+  def estimateSmartFee(
+      blocks: Int,
+      mode: String = "CONSERVATIVE"): Future[EstimateSmartFeeResult] = {
```
Christewart: Perhaps create options for this in a file? Or a algebraic data type that wrap the strings

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,727 @@
...
+    bitcoindCall[Vector[Address]]("getaddressesbyaccount",
+                                  List(JsString(account)))
+  }
+
+  def getRawTransactionRaw(txid: DoubleSha256Digest): Future[Transaction] = {
+    bitcoindCall[Transaction]("getrawtransaction",
+                              List(JsString(txid.hex), JsBoolean(false)))
+  }
+
+  // As is, RpcTransaction may not have enough data (add options?), also is RpcTransaction in the right place?
```
Christewart: old comment?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,727 @@
...
+      banTime: Int = 86400,
+      absolute: Boolean = false): Future[Unit] = {
+    bitcoindCall[Unit]("setban",
+                       List(JsString(address.getHostAddress),
+                            JsString(command),
+                            JsNumber(banTime),
+                            JsBoolean(absolute)))
+  }
+
+  // Should be BitcoinFeeUnit
```
Christewart: just commenting to make sure you fix this later

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,727 @@
...
+
+    bitcoindCall[SignRawTransactionResult]("signrawtransactionwithkey", params)
+  }
+
+  def stop: Future[String] = {
+    bitcoindCall[String]("stop")
+  }
+
+  def submitBlock(block: Block, dummy: Option[JsValue] = None): Future[Unit] = {
+    val params =
```
Christewart: This is debatable what is cleaner to read, but I prefer 

```scala
val params = dummy match { 
  case Some(dum) => ...
  case None => ...  
}
```

I don't think you necessarily have to change, but would like to hear your thoughts

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,727 @@
...
+    Http(m.system).singleRequest(req)
+  }
+
+  def buildRequest(methodName: String, params: JsArray): HttpRequest = {
+    val m: Map[String, JsValue] = Map("method" -> JsString(methodName),
+                                      "params" -> params,
+                                      "id" -> JsString("")) // java.util.UUID
+    val jsObject = JsObject(m)
+
+    val uri = "http://localhost:18443"
```
Christewart: Just a reminder that we need to figure out how to configure this stuff eventually.

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/NetworkResult.scala
```diff
@@ -0,0 +1,362 @@
...
+
+case class GetMemPoolEntryResult(
+    size: Int,
+    fee: Bitcoins,
+    modifiedfee: Bitcoins,
+    time: UInt32,
+    height: Int,
+    descendantcount: Int,
+    descendantsize: Int,
+    descendantfees: Bitcoins, // Should be BitcoinFeeUnit
```
Christewart: reminder to change this

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -0,0 +1,68 @@
...
+    keypoolsize_hd_internal: Int,
+    paytxfee: BitcoinFeeUnit,
+    hdmasterkeyid: Sha256Hash160Digest,
+    unlocked_until: Option[Int])
+    extends WalletResult
+
+case class BumpFeeResult(
+    txid: DoubleSha256Digest,
+    origfee: Bitcoins,
+    fee: Bitcoins, // Should be BitcoinFeeUnit
```
Christewart: reminder to change

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,271 @@
...
+import org.bitcoins.core.protocol.script.{ScriptPubKey, ScriptSignature}
+import org.bitcoins.core.protocol.transaction._
+import org.bitcoins.core.wallet.fee.{BitcoinFeeUnit, SatoshisPerByte}
+import org.bitcoins.rpc.jsonmodels.RpcAddress
+import play.api.libs.json._
+
+import scala.util.{Failure, Success, Try}
+
+object JsonReaders {
+  private def processJsNumber[T](numFunc: BigDecimal => T)(
```
Christewart: Nice! Perhaps add a comment about the intentions of this method. It is self evident to me since we talked about it in person, but might be nice for a reader of the library to write a sentence of what it is doing

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,271 @@
...
+        case _                    => return JsError("error.expected.vin.sequence")
+      }
+      (json \ "coinbase").validate[String] match {
+        case s: JsSuccess[String] =>
+          JsSuccess(
+            CoinbaseInput(ScriptSignature.fromAsmHex(s.value), sequence))
+        case _ =>
+          val txid = (json \ "txid").validate[DoubleSha256Digest] match {
+            case id: JsSuccess[DoubleSha256Digest] => id.value
+            case _                                 => return JsError("error.expected.vin.txid")
```
Christewart: Do you need the `return` here? Generally it is considered an anti pattern to return per-emptively unless there is some performance implications

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,271 @@
...
+  // This still needs cleanup
+  implicit object RpcAddressReads extends Reads[RpcAddress] {
+    override def reads(json: JsValue): JsResult[RpcAddress] = json match {
+      case array: JsArray =>
+        val balance = array.value.find(_.isInstanceOf[JsNumber]) match {
+          case Some(JsNumber(n)) => Bitcoins(n)
+          case Some(err) =>
+            return JsError(
+              s"error.expected.jsnumber, got ${Json.toJson(err).toString()}")
+          case None => return JsError("error.expected.balance")
```
Christewart: See comment above about returning per-empitevly. I guess you do have a comment about cleanign this up though.

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala
```diff
@@ -0,0 +1,235 @@
...
+  )(RpcScriptPubKey)
+  implicit val rpcTransactionOutputReads: Reads[RpcTransactionOutput] =
+    Json.reads[RpcTransactionOutput]
+  implicit val rpcTransactionReads: Reads[RpcTransaction] =
+    Json.reads[RpcTransaction]
+  implicit val getBlockWithTransactionsResultReads: Reads[
+    GetBlockWithTransactionsResult] = Json.reads[GetBlockWithTransactionsResult]
+
+  implicit val paymentReads: Reads[Payment] = (
+    (__ \ "involvesWatchonly").readNullable[Boolean] and
```
Christewart: Generally I think we should avoid using scala's method calls via [space] instead of using `.` notation. It is confusing for most people and can lead to ambiguities in parsing if i recall correctly.

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+import org.bitcoins.core.protocol.script.ScriptSignature
+import org.bitcoins.rpc.jsonmodels.GetTransactionResult
+
+import scala.concurrent.Future
+
+class RpcClientTest extends AsyncFlatSpec {
+  implicit val system = ActorSystem()
+  implicit val m = ActorMaterializer()
+  implicit val ec = m.executionContext
+  implicit val regTestNetworkParam = RegTest
```
Christewart: Eventually we want to read this in from a config file to test your configuration stuff to :-)

Christewart: and move all of the authentication stuff into this file. Thinkign about this more, I think the `RpcClient` should probably take in as a parameter some configuration class.

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/NetworkResult.scala
```diff
@@ -0,0 +1,362 @@
...
+
+case class GetTransactionResult(
+    amount: Bitcoins,
+    fee: Option[Bitcoins],
+    confirmations: Int,
+    generated: Option[Boolean],
+    blockhash: Option[DoubleSha256Digest],
+    blockindex: Option[Int],
+    blocktime: Option[UInt32],
+    txid: DoubleSha256Digest,
```
Christewart: This is sort of a nit, but let's use `txId` as that is what is used in the Transaction class in core. Another note, this is a big endian hash right?

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+                        .inputs(0)
+                        .previousOutput
+                        .txId
+                        .flip == input0.txId)
+                    assert(
+                      transaction
+                        .inputs(1)
+                        .previousOutput
+                        .txId
+                        .flip == input1.txId)
```
Christewart: bug here with endianness

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  it should "be able to create a raw transaction" in {
+    client.generate(2).flatMap { blocks =>
+      client.getBlock(blocks(0)).flatMap { block0 =>
+        client.getBlock(blocks(1)).flatMap { block1 =>
+          client.getTransaction(block0.tx(0)).flatMap { transaction0 =>
+            client.getTransaction(block1.tx(0)).flatMap { transaction1 =>
+              val input0 =
+                TransactionOutPoint(transaction0.txid,
```
Christewart: There is a bug here, the txid you are using in the outPoint here is big endian, NOT little endian which is the protocol level serialization

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  private def createRawTransaction: Future[Transaction] = {
+    client.generate(2).flatMap { blocks =>
+      client.getBlock(blocks(0)).flatMap { block0 =>
+        client.getBlock(blocks(1)).flatMap { block1 =>
+          client.getTransaction(block0.tx(0)).flatMap { transaction0 =>
+            client.getTransaction(block1.tx(0)).flatMap { transaction1 =>
+              val input0 =
+                TransactionOutPoint(transaction0.txid,
```
Christewart: the `.txid` here is big endian not little endian. We need to be careful with these. Probably should just add a new hash type to be `BEDoubleSha256Digest` to distinguish the endianness

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+    transactionF.flatMap { transaction =>
+      client.signRawTransactionWithWallet(transaction).flatMap {
+        signedTransaction =>
+          client
+            .generate(100)
+            .flatMap { _ => // Can't spend coinbase until depth 100
+              client.sendRawTransaction(signedTransaction.hex, true).map {
+                transactionHash =>
+                  logger.info(transactionHash.hex)
+                  assert(true)
```
Christewart: can we call  `.getRawTransaction` in here instead of just `assert(true)`?

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+    }
+  }
+
+  it should "be able to send a raw transaction to the mem pool" in {
+    val transactionF = createRawTransaction
+    transactionF.flatMap { transaction =>
+      client.signRawTransactionWithWallet(transaction).flatMap {
+        signedTransaction =>
+          client
+            .generate(100)
```
Christewart: Instead of generating 100 blocks in every test can we just generate 150 blocks in the `beforeAll` section of the test suite?

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+  it should "be able to get the network hash per sec" in {
+    client.getNetworkHashPS().map { hps =>
+      assert(hps > 0)
+    }
+  }
+
+  it should "be able to get an address from bitcoind" in {
+    val addressF = client.getNewAddress()
+    addressF.map { address =>
+      logger.info(address.value)
```
Christewart: Let's not log this.

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+  it should "be able to get a new raw change address" in {
+    client.getRawChangeAddress().map { address =>
+      assert(true)
+    }
+
+    client.getRawChangeAddress(Some("legacy")).map { address =>
+      assert(true)
+    }
+
+    client.getRawChangeAddress(Some("p2sh-segwit")).map { address =>
```
Christewart: We can do a check here that says `address.isInstanceOf[P2SHAddress]`

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+      assert(true)
+    }
+
+    client.getRawChangeAddress(Some("bech32")).map { address =>
+      assert(true)
```
Christewart: `assert(address.isInstanceOf[Bech32Address]`

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+        }
+      }
+    }
+  }
+
+  it should "be able to get the block count" in {
+    val blockCountF = client.getBlockCount
+    blockCountF.map { count =>
+      logger.info(count.toString)
+      assert(count >= 0)
```
Christewart: `assert(count > 0)` ? We should have a block at this point right? Especially if we end up setting up blocks in a `beforeAll`

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+      assert(info.networkactive)
+      assert(info.connections == 0)
+      assert(info.localrelay)
+    }
+  }
+
+  it should "be able to generate blocks" in {
+    val blocksF = client.generate(3)
+    blocksF.map { blocks =>
+      blocks.foreach(block => logger.info(block.toString))
```
Christewart: log statement isn't needed. The best way to test this might be to call `.getBlockCount` and then generate 3 blocks and then call getblockcount again

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+      assert(blocks.length == 3)
+    }
+  }
+
+  it should "be able to generate blocks and then get their serialized headers" in {
+    val blocksF = client.generate(2)
+    blocksF.flatMap { blocks =>
+      blocks.foreach(block => logger.info(block.toString))
```
Christewart: Remove excess logging when done

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+
+  it should "be able to generate blocks and then get their headers" in {
+    val blocksF = client.generate(2)
+    blocksF.flatMap { blocks =>
+      val headerF0 = client.getBlockHeader(blocks(0))
+      headerF0.map { header =>
+        logger.info(header.toString)
+        assert(header.nextblockhash.contains(blocks(1)))
+      }
+      val headerF1 = client.getBlockHeader(blocks(1))
```
Christewart: Note that this Future does not depend on the `headerF0` future. So this one might finish running first and then terminate the test case without testing 

```
assert(header.nextblockhash.contains(blocks(1)))
```

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+    }
+  }
+
+  it should "be able to get block hash by height" in {
+    client.generate(2).flatMap { blocks =>
+      client.getBlockCount.flatMap { count =>
+        client.getBlockHash(count).map { hash =>
+          assert(blocks(1) == hash)
+        }
+        client.getBlockHash(count - 1).map { hash =>
```
Christewart: Note that this Future might be finished running before 

```
assert(blocks(1) == hash)
```

Is done running, which means that that assertion will not get tested.

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+        }
+      }
+    }
+  }
+
+  it should "be able to get help from bitcoind" in {
+    client.help().flatMap { genHelp =>
+      assert(!genHelp.isEmpty)
+      client.help("help").map { helpHelp =>
+        assert(genHelp != helpHelp)
```
Christewart, line 1002: Hah this was pretty entertaining to read :-)

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,517 @@
...
+  val client = new RpcClient
+  val logger = BitcoinSLogger.logger
+
+  behavior of "RpcClient"
+
+  it should "be able to get blockchain info" in {
+    client.getBlockChainInfo.flatMap { info =>
+      logger.info(info.toString)
+      assert(info.chain == "regtest")
+      client.getBestBlockHash.map(bestHash =>
```
Christewart: We are not waiting for this Future to run, so the test might complete without this check getting executed on the thread pool

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+import org.bitcoins.rpc.serializers.JsonSerializers._
+import org.bitcoins.rpc.serializers.JsonWriters.mapWrites
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.sys.process._
+
+class RpcClient(instance: DaemonInstance)(
+    implicit
+    m: ActorMaterializer,
+    ec: ExecutionContext) {
```
Christewart: You do not need to pass in a `ec`. You can just create `implicit val ec = m.system.dispatcher`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+  private val resultKey = "result"
+  private val errorKey = "error"
+  private val logger = BitcoinSLogger.logger
+  private implicit val network = instance.network
+
+  def getDaemon: DaemonInstance = instance
```
Christewart: This is redundant, you can just call `.instance` if you are a client.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+
+  def createMultiSig(
+      minSignatures: Int,
+      keys: Vector[ECPublicKey]): Future[MultiSigResult] = {
+    bitcoindCall[MultiSigResult](
+      "createmultisig",
+      List(JsNumber(minSignatures), Json.toJson(keys.map(_.hex))))
+  }
+
+  implicit val outputMapWrites: Writes[Map[BitcoinAddress, Bitcoins]] =
```
Christewart: Can this be moved into the `JsonWriters` file?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+      "createrawtransaction",
+      List(Json.toJson(inputs), Json.toJson(outputs), JsNumber(locktime)))
+  }
+
+  def decodeRawTransaction(transaction: Transaction): Future[RpcTransaction] = {
+    bitcoindCall[RpcTransaction]("decoderawtransaction",
+                                 List(JsString(transaction.hex)))
+  }
+
+  // TODO: add ScriptPubKey.asmHex
```
Christewart, line 183: ?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+  }
+
+  def getAccountAddress(account: String): Future[BitcoinAddress] = {
+    bitcoindCall[BitcoinAddress]("getaccountaddress", List(JsString(account)))
+  }
+
+  private def getAddedNodeInfo(node: Option[URI]): Future[Vector[Node]] = {
+    val params =
+      if (node.isEmpty) {
+        List()
```
Christewart: Let's use `List.empty` or `Nil` instead of allocating a new instance of List()

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+      includeMemPool: Boolean = true): Future[GetTxOutResult] = {
+    bitcoindCall[GetTxOutResult](
+      "gettxout",
+      List(JsString(txid.hex), JsNumber(vout), JsBoolean(includeMemPool)))
+  }
+
+  private def getTxOutProof(
+      txids: Vector[DoubleSha256Digest],
+      headerHash: Option[DoubleSha256Digest]): Future[MerkleBlock] = {
+    def params = {
```
Christewart: why `def` instead of `val`?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+    Json.writes[RpcOpts.ImportMultiAddress]
+  implicit val importMultiRequestWrites: Writes[RpcOpts.ImportMultiRequest] =
+    Json.writes[RpcOpts.ImportMultiRequest]
+
+  def importMulti(
+      requests: Vector[RpcOpts.ImportMultiRequest],
+      rescan: Boolean = true): Future[Vector[ImportMultiResult]] = {
+    bitcoindCall[Vector[ImportMultiResult]](
+      "importmulti",
+      List(Json.toJson(requests), JsObject(Seq(("rescan", JsBoolean(rescan))))))
```
Christewart: In general avoid `Seq`, use `Vector`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+      confirmations: Int = 1,
+      includeEmpty: Boolean = false,
+      includeWatchOnly: Boolean = false): Future[Vector[ReceivedAddress]] = {
+    bitcoindCall[Vector[ReceivedAddress]]("listreceivedbyaddress",
+                                          List(JsNumber(confirmations),
+                                               JsBoolean(includeEmpty),
+                                               JsBoolean(includeWatchOnly)))
+  }
+
+  // Need to configure default headerHash
```
Christewart: Inside of a `bitcoin.conf` file?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+        List(Json.toJson(include.get), JsArray.empty)
+      } else {
+        List(Json.toJson(include.get), Json.toJson(exclude.get))
+      }
+    bitcoindCall[Map[String, Int]]("logging", params)
+  }
+  def logging: Future[Map[String, Int]] = logging(None, None)
+
+  def logging(
+      include: Vector[String] = Vector(),
```
Christewart: Vector.empty

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+
+  def removePrunedFunds(txid: DoubleSha256Digest): Future[Unit] = {
+    bitcoindCall[Unit]("removeprunedfunds", List(JsString(txid.hex)))
+  }
+
+  private def rescanBlockChain(
+      start: Option[Int],
+      stop: Option[Int]): Future[RescanBlockChainResult] = {
+    val params =
+      if (start.isEmpty) {
```
Christewart, line 851: Is there a missing case here? Where we start scanning from the genesis block, but stop at a given block? Perhaps pattern match on this so you get an exhaustiveness check?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+                            JsString(command),
+                            JsNumber(banTime),
+                            JsBoolean(absolute)))
+  }
+
+  def setNetworkActive(activate: Boolean): Future[Unit] = {
+    bitcoindCall[Unit]("setnetworkactive", List(JsBoolean(activate)))
+  }
+
+  // Should be BitcoinFeeUnit
```
Christewart: Add a TODO to this

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+                         List(JsString(key.toWIF(network)), JsString(message)))
+  }
+
+  private def signRawTransaction(
+      transaction: Transaction,
+      utxoDeps: Option[Vector[RpcOpts.SignRawTransactionOutputParameter]],
+      keys: Option[Vector[ECPrivateKey]],
+      sigHash: Option[String]): Future[SignRawTransactionResult] = {
+    val params =
+      if (utxoDeps.isEmpty) {
```
Christewart: Let's avoid nesting here. Break this out into top level statements. 

``` 
val utxos = utxos.map(u => Json.toJson(u)
.getOrElse(Vector.empty)
val keys = keys.map(k => Json.toJson(k))
.getOrElse(Vectory.empty)
val sighash = sigHash.map(s => Json.toJson(s))
.getOrElse(Vector.empty)
```
unless you can't define `keys` without a `utxo`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+
+  // Should both logging and throwing be happening?
+  private def parseResult[T](result: JsResult[T], json: JsValue): T = {
+    // First catch errors thrown by calls with Unit as the expected return type (which isn't handled by UnitReads)
+    if (result == JsSuccess(())) {
+      (json \ errorKey).validate[RpcError] match {
+        case err: JsSuccess[RpcError] =>
+          logger.error(s"Error ${err.value.code}: ${err.value.message}")
+          throw new RuntimeException(
+            s"Error ${err.value.code}: ${err.value.message}") // More specific type?
```
Christewart: I think we can just leave this as a `RuntimeException` for now.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+    }
+  }
+
+  case class RpcError(code: Int, message: String)
+  implicit val rpcErrorReads: Reads[RpcError] = Json.reads[RpcError]
+
+  // Should both logging and throwing be happening?
+  private def parseResult[T](result: JsResult[T], json: JsValue): T = {
+    // First catch errors thrown by calls with Unit as the expected return type (which isn't handled by UnitReads)
+    if (result == JsSuccess(())) {
```
Christewart: Let's factor this out into it's own method. Call it `checkUnit()` or something? Add documentation to this method.

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
Christewart: I think this is fine, just change to `ByteString.empty`

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+  private def getPayload(response: HttpResponse): Future[JsValue] = {
+    val payloadF = response.entity.dataBytes.runFold(ByteString(""))(_ ++ _)
+
+    payloadF.map { payload =>
+      Json.parse(payload.decodeString(ByteString.UTF_8))
+    }
+  }
+
+  def sendRequest(req: HttpRequest): Future[HttpResponse] = {
+    Http(m.system).singleRequest(req)
```
Christewart, line 1129: After we send the request, we need to make sure the `HttpResponse` "id" field is the exact same UUID as we sent in the request. Else we need to throw an exception (we have the wrong request)

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
```diff
@@ -0,0 +1,1114 @@
...
+  }
+
+  def buildRequest(
+      instance: DaemonInstance,
+      methodName: String,
+      params: JsArray): HttpRequest = {
+    val m: Map[String, JsValue] = Map("method" -> JsString(methodName),
+                                      "params" -> params,
+                                      "id" -> JsString("")) // java.util.UUID
```
Christewart:
```scala 
val uuid = UUID.randomUUID().toString
```
and then

```scala 
"id" -> JsString(uuid)
```

rpc/src/main/scala/org/bitcoins/rpc/client/RpcOpts.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  case class ImportMultiAddress(address: BitcoinAddress)
+
+  case class LockUnspentOutputParameter(txid: DoubleSha256Digest, vout: Int)
+
+  implicit val lockUnspentParameterWrites: Writes[LockUnspentOutputParameter] =
+    Json.writes[LockUnspentOutputParameter]
+
+  object AddressType extends Enumeration {
+    type AddressType = Value
```
Christewart: Don't really understand what is going on here, why not just create an ADT and then case objects for all the address types?

rpc/src/main/scala/org/bitcoins/rpc/config/AuthCredentials.scala
```diff
@@ -0,0 +1,36 @@
+package org.bitcoins.rpc.config
+
+/**
+  * Created by chris on 5/2/17.
+  */
+sealed trait AuthCredentials {
+
+  /** The directory where our bitcoin.conf file is located */
+  def datadir: String
```
Christewart: Perhaps change this to a `File` type?

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RawTransactionResult.scala
```diff
@@ -0,0 +1,92 @@
...
+case class GetRawTransactionVin(
+    txid: Option[DoubleSha256Digest],
+    vout: Option[Int],
+    scriptSig: Option[GetRawTransactionScriptSig],
+    sequence: Option[BigDecimal],
+    txinwitness: Option[Vector[String]] // Should be TransactionWitness?
+) extends RawTransactionResult
+
+case class GetRawTransactionScriptSig(asm: String, hex: ScriptPubKey)
+    extends RawTransactionResult // is hex right???
```
Christewart: Not sure if I understand the question

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,313 @@
...
+  implicit object Int32Reads extends Reads[Int32] {
+    override def reads(json: JsValue): JsResult[Int32] = json match {
+      case JsNumber(n) =>
+        n.toBigIntExact() match {
+          case Some(num) => JsSuccess(Int32(num))
+          case None      => JsError(s"error.expected.Int32, got $n")
+        }
+      case JsString(s) => JsSuccess(Int32.fromHex(s))
+      case err @ (JsNull | _: JsBoolean | _: JsArray | _: JsObject) =>
+        JsError(s"error.expected.jsnumber, got ${Json.toJson(err).toString()}")
```
Christewart: Let's create a `buildErrorMsg` method to do this stuff instead of being redundant

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,313 @@
...
+      val sequence = (json \ "sequence").validate[UInt32] match {
+        case s: JsSuccess[UInt32] => s.value
+        case _                    => return JsError("error.expected.vin.sequence")
+      }
+      (json \ "coinbase").validate[String] match {
+        case s: JsSuccess[String] =>
+          JsSuccess(
+            CoinbaseInput(ScriptSignature.fromAsmHex(s.value), sequence))
+        case _ =>
+          val txid = (json \ "txid").validate[DoubleSha256Digest] match {
```
Christewart: Looking at this further, can't we just 

```scala
(json \ "txid").validate[DoubleSha256Digest].map(_.value)
```
?

rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -0,0 +1,313 @@
...
+      json.validate[OutPoint] match {
+        case JsSuccess(op, _) =>
+          JsSuccess(TransactionOutPoint(op.txid, op.vout))
+        case JsError(err) =>
+          JsError(s"Could not parse TransactionOutPoint, got ${err.toString()}")
+      }
+    }
+  }
+
+  // This still needs cleanup
```
Christewart: Bump

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1695 @@
...
+  val logger = BitcoinSLogger.logger
+
+  var password = "password"
+
+  P2SHScriptSignature(
+    "5221031ef932a3e39f7273534fd51a387b2d3bb2a663cd92739840167da6b5dd07b90e21023ba657cfddf8ce4a78df941d42ddb394c61234cd4b66ce45de223382752194d352ae")
+
+  val redeemScript = ScriptPubKey.fromAsmHex(
+    "5221031ef932a3e39f7273534fd51a387b2d3bb2a663cd92739840167da6b5dd07b90e21023ba657cfddf8ce4a78df941d42ddb394c61234cd4b66ce45de223382752194d352ae")
+  BitcoinSUtil.encodeHex(P2SHScriptPubKey(redeemScript).asmBytes)
```
Christewart: This can be deleted now right?

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
Christewart: Confirmed that this is the case yesterday. This needs to be re-written

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
nkohen: Should it be `m.system.dispatcher` or `m.executionContext`?

rpc/src/main/scala/org/bitcoins/rpc/client/RpcClient.scala
nkohen: This actually isn't possible since rescanblockchain does not take a map as input but rather it takes 0, 1, or 2 integers and so it always interprets a single Int as the start parameter

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RawTransactionResult.scala
nkohen: Should the type of hex be ScriptPubKey? This is what the documentation looks like for this result:
```
"scriptSig": {     (json object) The script
         "asm": "asm",  (string) asm
         "hex": "hex"   (string) hex
       }
```

Christewart:
Should be a ScriptSignature

Christewart:
M.executionContext should be fine

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1713 @@
...
+    }
+  }
+
+  override def afterAll(): Unit = {
+    client.stop().map(println)
+    otherClient.stop().map(println)
+    walletClient.stop().map(println)
+    pruneClient.stop().map(println)
+    if (TestUtil.deleteTmpDir(client.getDaemon.authCredentials.datadir))
+      println("Temp bitcoin directory deleted")
```
Christewart: Should be logs and not `println`

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1713 @@
...
+class RpcClientTest
+    extends AsyncFlatSpec
+    with BeforeAndAfterAll
+    with BeforeAndAfter {
+  implicit val system = ActorSystem()
+  implicit val m = ActorMaterializer()
+  implicit val ec = m.executionContext
+  implicit val networkParam = TestUtil.network
+
+  private var newPort = networkParam.port + 5
```
Christewart: Let's not use `var` here. Just create method that generates a random port like [this](https://github.com/bitcoin-s/bitcoin-s-rpc-client/blob/master/src/test/scala/org/bitcoins/rpc/util/TestUtil.scala#L74)

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1713 @@
...
+  private var newRpcPort = networkParam.rpcPort + 5
```
Christewart: Same as above

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1713 @@
...
+  private def getNewPort(): Int = {
```
Christewart: Move this into a `TestUtil` file.

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
```diff
@@ -0,0 +1,1713 @@
...
+    newPort = newPort + 5
+    newPort
+  }
+  private def getNewRpcPort(): Int = {
```
Christewart: Move this into a `TestUtil` file.

Christewart: If you are trying to minimize LOC do something like 

```scala
val util = TestUtil

```

rpc/src/test/scala/org/bitcoins/rpc/RpcClientTest.scala
Christewart: if you need to save the port, save it in a `val` where ever you are using it. In general we shouldn't use `var` unless absolutely needed. It creates nightmares for concurrency

Issue #174, opened 2018-06-07
Coverage report broken


Christewart:
We recently refactored bitcoin-s into a multi build project and this broke test coverage reporting for coveralls. We need to fix this.

TannerR1776:
Possible information related to this issue?
https://github.com/scoverage/sbt-scoverage#multi-project-reports 

Current build command: https://github.com/bitcoin-s/bitcoin-s-core/blob/61efc93a4e538f3da0583ba7fc33dee07caa9f13/.travis.yml#L33 is executed as a single command, which appears to not be supported for multi-project reports?

Christewart:
I think that is exactly right. Can you tackle this?

TannerR1776:
Yes, I will see if I can get this working, and report back!

TannerR1776:
I have submitted PR #179 to resolve this issue.

PR #179, opened 2018-06-22
Configure Travis-CI for multiproject build

TannerR1776:
This pull request is intended to resolve issue #174. The issue was that Travis-CI requires an exit code of 1 to fail a build. With the previous build configuration, when the coverageReport failed, it was not correctly exiting, which caused the build to pass, even when coverage had fallen below 90%. I have modified the configuration to comply with the Travis-CI multiproject guidelines (see https://github.com/scoverage/sbt-coveralls#travis-ci-integration).

The build is now (correctly) failing when coverage is below 90%: https://travis-ci.org/TannerR1776/bitcoin-s-core/builds/395423421

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,9 @@ install:
   - ./configure --enable-jni --enable-experimental --enable-module-ecdh
   - sudo make install
   - cd ../
-script: sbt -Djava.library.path=secp256k1/.libs clean coverage test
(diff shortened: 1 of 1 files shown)


Christewart:
CI is failing because we are *actually* checking the coverage again. Awesome work @TannerR1776 !

Issue #177, opened 2018-06-18
Supporting lightning network invoice encoding


Christewart:
Lightning network invoice encoding is a variant of bech32. We should support this in bitcoin-s 

https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md

PR #184, opened 2018-08-05
WIP: Add support for Lightning Network Currency Units

TannerR1776:
This PR is to add support for Lightning Network currency units, such as MilliBitcoin and MicroBitcoin.

These new currency units are in support of Issue #177, and will be used in the creation of Lightning network invoices in the future.

This PR is still a WIP.

### DIFF
--- a/core-gen/src/test/scala/org/bitcoins/core/gen/CurrencyUnitGenerator.scala
+++ b/core-gen/src/test/scala/org/bitcoins/core/gen/CurrencyUnitGenerator.scala
@@ -1,7 +1,8 @@
 package org.bitcoins.core.gen
 
 import org.bitcoins.core.currency.{ Bitcoins, CurrencyUnit, CurrencyUnits, Satoshis }
-import org.bitcoins.core.number.Int64
+import org.bitcoins.core.number.{ Int32, Int64 }
+import org.bitcoins.core.protocol.ln._
 import org.scalacheck.Gen
 
 /**
@@ -31,3 +32,28 @@ trait CurrencyUnitGenerator {
 }
 
 object CurrencyUnitGenerator extends CurrencyUnitGenerator
+
+trait LnCurrencyUnitGenerator {
+
+  def milliBitcoin: Gen[MilliBitcoins] = for {
+    amount <- Gen.choose(MilliBitcoins.min.toLong, MilliBitcoins.max.toLong)
+  } yield MilliBitcoins(amount)
+
+  def microBitcoin: Gen[MicroBitcoins] = for {
+    amount <- Gen.choose(MicroBitcoins.min.toLong, MicroBitcoins.max.toLong)
+  } yield MicroBitcoins(amount)
+
+  def nanoBitcoin: Gen[NanoBitcoins] = for {
+    amount <- Gen.choose(NanoBitcoins.min.toLong, NanoBitcoins.max.toLong)
+  } yield NanoBitcoins(amount)
+
+  def picoBitcoin: Gen[PicoBitcoins] = for {
+    amount <- Gen.choose(PicoBitcoins.min.toLong, PicoBitcoins.max.toLong)
+  } yield PicoBitcoins(amount)
+
+  def lnCurrencyUnit: Gen[LnCurrencyUnit] = Gen.oneOf(milliBitcoin, microBitcoin, nanoBitcoin, picoBitcoin)
+
+  def negativeLnCurrencyUnit: Gen[LnCurrencyUnit] = lnCurrencyUnit.suchThat(_ < LnCurrencyUnits.zero)
+}
+
+object LnCurrencyUnitGenerator extends LnCurrencyUnitGenerator
--- a/core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
@@ -1,0 +1,231 @@
+package org.bitcoins.core.protocol.ln
+
+import org.bitcoins.core.currency.{ Bitcoins, Satoshis }
+import org.bitcoins.core.number.{ BaseNumbers, Int64 }
+import org.bitcoins.core.protocol.NetworkElement
+import scodec.bits.ByteVector
+
+import scala.math.BigDecimal.RoundingMode
+
+sealed abstract class LnCurrencyUnit extends NetworkElement {
+  type A
+
+  def character: Char
+
+  def multiplier: BigDecimal
+
+  def >=(ln: LnCurrencyUnit): Boolean = {
+    toPicoBitcoinValue >= ln.toPicoBitcoinValue
+  }
+
+  def >(ln: LnCurrencyUnit): Boolean = {
+    toPicoBitcoinValue > ln.toPicoBitcoinValue
+  }
+
+  def <(ln: LnCurrencyUnit): Boolean = {
+    toPicoBitcoinValue < ln.toPicoBitcoinValue
+  }
+
+  def <=(ln: LnCurrencyUnit): Boolean = {
+    toPicoBitcoinValue <= ln.toPicoBitcoinValue
+  }
+
+  def !=(ln: LnCurrencyUnit): Boolean = !(this == ln)
+
+  def ==(ln: LnCurrencyUnit): Boolean = toPicoBitcoinValue == ln.toPicoBitcoinValue
+
+  def +(ln: LnCurrencyUnit): LnCurrencyUnit = {
+    PicoBitcoins(toPicoBitcoinValue + ln.toPicoBitcoinValue)
+  }
+
+  def -(ln: LnCurrencyUnit): LnCurrencyUnit = {
+    PicoBitcoins(toPicoBitcoinValue - ln.toPicoBitcoinValue)
+  }
+
+  def *(ln: LnCurrencyUnit): LnCurrencyUnit = {
+    PicoBitcoins(toPicoBitcoinValue * ln.toPicoBitcoinValue)
+  }
+
+  def unary_- : LnCurrencyUnit = {
+    PicoBitcoins(-toPicoBitcoinValue)
+  }
+
+  override def bytes: ByteVector = Int64(toPicoBitcoinValue).bytes.reverse
+
+  def toBigInt: BigInt
+
+  def toLong: Long = toBigInt.toLong
+
+  def toInt: Int = {
+    require(this.toBigInt >= Int.MinValue, "Number was too small for Int, got: " + underlying)
+    require(this.toBigInt <= Int.MaxValue, "Number was too big for Int, got: " + underlying)
+    toBigInt.toInt
+  }
+
+  protected def underlying: A
+
+  def toSatoshis: Satoshis
+
+  def toPicoBitcoinValue: BigInt
+
+  def toPicoBitcoinMultiplier: Int
+
+  def toEncodedString: String = this.toBigInt + this.character.toString()
+}
+
+sealed abstract class MilliBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
+
+  override def character: Char = 'm'
+
+  override def multiplier: BigDecimal = LnPolicy.milliMultiplier
+
+  override def toPicoBitcoinMultiplier: Int = 1000000000
+
+  override def toBigInt: A = underlying
+
+  override def toSatoshis: Satoshis = LnCurrencyUnits.toSatoshi(this)
+
+  override def toPicoBitcoinValue: BigInt = LnCurrencyUnits.toPicoBitcoinValue(this)
+}
+
+object MilliBitcoins extends BaseNumbers[MilliBitcoins] {
+  val min = MilliBitcoins(LnPolicy.minMilliBitcoins)
+  val max = MilliBitcoins(LnPolicy.maxMilliBitcoins)
+  val zero = MilliBitcoins(0)
+  val one = MilliBitcoins(1)
+
+  def apply(milliBitcoins: Int64): MilliBitcoins = MilliBitcoins(milliBitcoins.toBigInt)
+
+  def apply(underlying: BigInt): MilliBitcoins = MilliBitcoinsImpl(underlying)
+
+  private case class MilliBitcoinsImpl(underlying: BigInt) extends MilliBitcoins {
+    require(underlying >= LnPolicy.minMilliBitcoins, "Number was too small for MilliBitcoins, got: " + underlying)
+    require(underlying <= LnPolicy.maxMilliBitcoins, "Number was too big for MilliBitcoins, got: " + underlying)
+  }
+}
+
+sealed abstract class MicroBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
+
+  override def character: Char = 'u'
+
+  override def multiplier: BigDecimal = LnPolicy.microMultiplier
+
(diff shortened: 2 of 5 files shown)


core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+    PicoBitcoins(toPicoBitcoins * ln.toPicoBitcoins)
+  }
+
+  def unary_- : LnCurrencyUnit = {
+    currencyUnitFromPicoValue(this, -toPicoBitcoins)
+  }
+
+  override def bytes: Seq[Byte] = toSatoshis.bytes
+
+  def toBigInt: BigInt
```
Christewart: Let's add a `def toLong: Long` to this as a helper method, just make sure we check the bounds that we aren't having a overflow happen. Maybe add a `toInt` as well ?

core-gen/src/test/scala/org/bitcoins/core/gen/CurrencyUnitGenerator.scala
```diff
@@ -31,3 +32,26 @@ trait CurrencyUnitGenerator {
 }
 
 object CurrencyUnitGenerator extends CurrencyUnitGenerator
+
+trait LnCurrencyUnitGenerator {
+
+  def milliBitcoin: Gen[MilliBitcoins] = for {
+    amount <- Gen.choose(MilliBitcoins.min.toBigInt.toLong, MilliBitcoins.max.toBigInt.toLong)
```
Christewart: Let's use `.toLong` here and remove `toBigInt.toLong` when we add a helper method.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+  val PicoMultiplier: BigDecimal = BigDecimal(0.000000000001)
+
+  def apply(lnUnit: LnCurrencyUnit, amount: BigInt): LnCurrencyUnit = lnUnit match {
+    case b: MilliBitcoins => MilliBitcoins(amount)
+    case c: MicroBitcoins => MicroBitcoins(amount)
+    case d: NanoBitcoins => NanoBitcoins(amount)
+    case e: PicoBitcoins => PicoBitcoins(amount)
+  }
+
+  def getZeroUnit(lnUnit: LnCurrencyUnit): LnCurrencyUnit = apply(lnUnit, BigInt(0))
```
Christewart: Let's replace this with 

```
val zero: LnCurrencyUnit = PicoBitcoins.zero
```
It shouldn't matter what the *actual* type is as they are all `LnCurrencyUnit` and the client should be able to use them interchangeably, we are hiding the implementation details from that user. That is why we convert everything to `PicoBitcoins` for all arithemtic operations.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  def toPicoBitcoins(lnCurrencyUnits: LnCurrencyUnit): BigInt = lnCurrencyUnits.toBigInt * (lnCurrencyUnits.multiplier / LnCurrencyUnits.PicoMultiplier).toBigInt()
```
Christewart: First, let's wrap this to multiple lines using 
```
def toPicoBitcoins(): PicoBitcoins = { 
... 
}
```

and this doesn't seem to convert to `PicoBitcoins`? The return type is `BigInt`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  def toSatoshi(lnCurrencyUnits: LnCurrencyUnit): Satoshis = {
```
Christewart: Let's add a link to the official lightning documentation for this.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+  val oneMicroBTC: LnCurrencyUnit = MicroBitcoins.one
+  val oneNanoBTC: LnCurrencyUnit = NanoBitcoins.one
+  val onePicoBTC: LnCurrencyUnit = PicoBitcoins.one
+
+  val MilliMultiplier: BigDecimal = BigDecimal(0.001)
+  val MicroMultiplier: BigDecimal = BigDecimal(0.000001)
+  val NanoMultiplier: BigDecimal = BigDecimal(0.000000001)
+  val PicoMultiplier: BigDecimal = BigDecimal(0.000000000001)
+
+  def apply(lnUnit: LnCurrencyUnit, amount: BigInt): LnCurrencyUnit = lnUnit match {
```
Christewart: This doesn't really make sense to me, and can be really dangerous as you don't know what the actual type of `lnUnit` is. There is a big different between 1 PicoBitcoin and 1 MilliBitcoin

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+  override def toPicoBitcoins: BigInt = LnCurrencyUnits.toPicoBitcoins(this)
+}
+
+object MilliBitcoins extends BaseNumbers[MilliBitcoins] {
+  val min = MilliBitcoins(minimum)
+  val max = MilliBitcoins(maximum)
+  val zero = MilliBitcoins(0)
+  val one = MilliBitcoins(1)
+
+  def minimum: BigInt = -(Consensus.maxMoney.toBigDecimal / (Bitcoins.one.satoshis.toBigDecimal * LnCurrencyUnits.MilliMultiplier)).toBigInt()
```
Christewart: I don't think these belong in here. They belong in something like `LnPolicy`. You cannot currently send more than `0.042` btc on the lightning protocol so this isn't really correct. 

But moreover, just like how [satoshis](https://github.com/bitcoin/bitcoin/blob/08a7316c144f9f2516db8fa62400893f4358c5ae/src/amount.h#L12) are unaware of what the actual protocol limitation is in the bitcoin codebase, we should keep these unaware of what the actual protocol limitations are in the lightning codebase. Satoshis are a int64, even though there is only 21M btc ever possible. I think we should remove these limits

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  def currencyUnitFromPicoValue(lnType: LnCurrencyUnit, unit: BigInt): LnCurrencyUnit = {
+    val amount = unit / (lnType.multiplier / LnCurrencyUnits.PicoMultiplier).toBigInt()
+    lnType match {
+      case b: MilliBitcoins => MilliBitcoins(amount)
+      case c: MicroBitcoins => MicroBitcoins(amount)
+      case d: NanoBitcoins => NanoBitcoins(amount)
+      case e: PicoBitcoins => PicoBitcoins(amount)
+    }
+  }
```
Christewart: Should we add a `toString` method that does the actual serialization?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+  def toBigInt: BigInt
+
+  protected def underlying: A
+
+  def toSatoshis: Satoshis
+
+  def toPicoBitcoins: BigInt
+
+  def currencyUnitFromPicoValue(lnType: LnCurrencyUnit, unit: BigInt): LnCurrencyUnit = {
+    val amount = unit / (lnType.multiplier / LnCurrencyUnits.PicoMultiplier).toBigInt()
```
Christewart: We should always use `toBigIntExact()` rather than `toBigInt`. It checks that we aren't losing precision.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  def *(ln: LnCurrencyUnit): LnCurrencyUnit = {
+    PicoBitcoins(toPicoBitcoins * ln.toPicoBitcoins)
+  }
+
+  def unary_- : LnCurrencyUnit = {
+    currencyUnitFromPicoValue(this, -toPicoBitcoins)
+  }
+
+  override def bytes: Seq[Byte] = toSatoshis.bytes
```
Christewart: Is this actually how these are serialized? I think they might be serialized differently as is you convert to satoshis, you can't convert back if you have something like PicoBitcoins as you lose precision.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  def toBigInt: BigInt
+
+  protected def underlying: A
+
+  def toSatoshis: Satoshis
+
+  def toPicoBitcoins: BigInt
+
+  def currencyUnitFromPicoValue(lnType: LnCurrencyUnit, unit: BigInt): LnCurrencyUnit = {
```
Christewart: I'm going to see if Nadav knows of a better way to do this, but at minimum I think we should make this `private` so no one else can call this

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
TannerR1776: This is correct, and something I missed, as we aren't actually comparing types when using `==`. `LnCurrencyUnit.zero` should work fine.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
TannerR1776: That is because what I really wanted was to get the PicoBitcoin value, not necessarily an actual PicoBitcoin object. I can rename the function, or return an object, either will work. I will think a bit more about which would be better.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
TannerR1776: As I mentioned in Slack, if `Bitcoins(1) + Bitcoins(1)` returns `Satoshis(200000000)`, then `MilliBitcoin(1) + MilliBitcoin(1)` should just return `PicoBitcoins(2000000000)` ? If we did this, I believe that would allow us to get rid of this function and the nasty apply function (above) all together.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
TannerR1776: We are going to have to figure out how we want to set the min/max values for these types then. One issue we run into can be see in the following example. If we start with max limits such as `MilliBitcoins(Int64.max)` and `PicoBitcoins(Int64.max)`, and then do something like `MilliBitcoins(Int64.max).toPicoBitcoins`, its going to produce a value that is larger than the "max" PicoBitcoins.

I see that the way that Satoshis/Bitcoins handle this is that Satoshis are defined as Int64, but Bitcoins are limited to 21m. Should PicosBitcoins similarly be Int64, and Milli/Micro/Nano be limited to `>= Pico.min and <= Pico.max` ?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
Christewart: Let's do that for v1 of this

core/src/main/scala/org/bitcoins/core/protocol/ln/LnPolicy.scala
```diff
@@ -0,0 +1,22 @@
+package org.bitcoins.core.protocol.ln
+
+import org.bitcoins.core.number.Int64
+
+sealed abstract class LnPolicy {
+
+  def MilliMultiplier: BigDecimal = BigDecimal(0.001)
```
Christewart: Probably best to change these to `val` so they only get evaluated once and then stored in memory. 

The difference between a `val` and a `def` is that a `def` is re-evaluated every time, while a `val` is cached in memory. 

Also let's make sure we follow the camel casing convention in Java. `aliceBobCharlie()` for methods not `AliceBobCharlie()`

Christewart (commented):
A couple things,

First, let's add some hard coded test vectors in the format of a test file like this test case example of a Transaction

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core-test/src/test/scala/org/bitcoins/core/protocol/transaction/TransactionTest.scala#L46


I'd also like you to address the nits, and then re-target this branch to `ln`. Let's start putting `ln` specific stuff in that branch for now. 

We can do a big merge into master later when we have other features of the LN included in the `ln` branch.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,219 @@
...
+  def toPicoBitcoinValue(lnCurrencyUnits: LnCurrencyUnit): BigInt = {
+    lnCurrencyUnits.toBigInt * (lnCurrencyUnits.multiplier / LnCurrencyUnits.picoMultiplier).toBigIntExact().get
+  }
+
+  /**
+   * For information regarding the rounding of sub-Satoshi values, please visit:
+   * https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#commitment-transaction-outputs
+   */
+  def toSatoshi(lnCurrencyUnits: LnCurrencyUnit): Satoshis = {
+    val sat = BigDecimal(lnCurrencyUnits.toBigInt) * Bitcoins.one.satoshis.toBigDecimal * lnCurrencyUnits.multiplier
```
Christewart, line 225: Hmm, is this right? It isn't entirely clear to me. Why do we need to use `Bitcoins.one.satoshis.toBigDecimal` ?

TannerR1776: I believe this is right. We need to find the Satoshi value of a single LnCurrencyUnit. Consider the following example: `PicoBitcoins(10000).toSatoshi` that would produce this calculation `10000 * 100,000,000 * 0.000000000001`.

`100,000,000 * 0.000000000001 = 0.0001` In other words, `1 PicoBitcoin = 0.0001 Satoshi`. From there, we can multiply this ratio times our actual number of PicoBitcoins (in this case 10000), and find that `10000 * 0.0001 = 1 Satoshi`

I have not been able to find a much clearer way to express this, outside of doing something like 
```scala
val satoshiValuePerLnUnit = Bitcoins.one.satoshis.toBigDecimal * lnCurrencyUnits.multiplier
val sat = BigDecimal(lnCurrencyUnits.toBigInt) * satoshiValuePerLnUnit
```

PR #190, opened 2018-09-02
WIP: Implement LnHumanReadablePart

TannerR1776:
This PR is a WIP. It is intended to implement the LnHumanReadablePart, which will be required to create LnInvoices in the future.


core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  override def toString: String = network.value + invoiceAmount
+
+  def invoiceAmount: String = {
+    if (amount.isDefined) {
+      amount.get.toBigInt.toString() + amount.get.character
+    } else ""
+  }
+
+  def bytes: Seq[Byte] = network.value.map(c => c.toByte) ++ invoiceAmount.map(c => c.toByte)
```
Christewart: We just modifed the api to have a `NetworkElement` return a `ByteVector` instead of a `Seq[Byte]`. This is because a `ByteVector` is MUCH more performant.

You should always use the `override` modifier to make sure you are following the interfaces specification. So this should look like 

```scala
override def bytes: ByteVector = { ... }
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,67 @@
+package org.bitcoins.core.protocol.ln
+
+import org.bitcoins.core.config.{ MainNet, NetworkParameters, RegTest, TestNet3 }
+import org.bitcoins.core.protocol.ln.LightningNetworkPrefix._
+
+sealed abstract class LnHumanReadablePart {
```
Christewart: Should `extend NetworkElement`

core/src/main/scala/org/bitcoins/core/protocol/ln/LightningNetworkPrefix.scala
```diff
@@ -0,0 +1,36 @@
+package org.bitcoins.core.protocol.ln
+
+import org.bitcoins.core.config.{ MainNet, NetworkParameters, RegTest, TestNet3 }
+
+sealed abstract class LightningNetworkParams {
+  def value: String
+  def network: NetworkParameters
+}
+
+object LightningNetworkPrefix {
```
Christewart: I think this should be renamed to `LIghtningNetworkParams` -- also we need to decide if want to consistently call things `Ln` (abbreviation) or `LightningNetwork` -- i'm in favor of brevity.

Christewart: If you are in favor of brevity, we shoud rename all of this class to `LnParams`, `LnPrefix` etc

TannerR1776: I agree, `LnParams` is better, and that this should be renamed. Not sure how it got to `LightningNetworkPrefix` in the first place.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,84 @@
...
+sealed abstract class LnHumanReadablePart extends NetworkElement {
+  def network: LnParams
+
+  def amount: Option[LnCurrencyUnit]
+
+  override def toString: String = network.value + invoiceAmount
+
+  def invoiceAmount: String = {
+    if (amount.isDefined) {
+      amount.get.toBigInt.toString() + amount.get.character
```
Christewart: Instead of doing this, there should just be a nice `toString()` method on `LnCurrencyUnit` I think. We don't want to have to replicate this piece of code everywhere we want to serialize a `LnCurrencyUnit`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,84 @@
...
+
+case object lnbcrt extends LnHumanReadablePart {
+  def network: LnParams = LnBitcoinRegTest
+  def amount: Option[LnCurrencyUnit] = None
+}
+
+object LnHumanReadablePart {
+  def apply(network: NetworkParameters): Option[LnHumanReadablePart] = {
+    val lnNetwork = LnParams.fromNetworkParameters(network)
+    if (lnNetwork.isDefined) {
```
Christewart: You can also re-write this as 
```scala
lnNetwork.map { params => 
  HumanReadablePart(params)
}
```

What the `.map` call does in this case is 

1. If `lnNetwork` is defined, the inner express gets evaluated
2. If `lnNetwork` is NOT defined, `None` is returned. 

The fancy word for these things is a monad. This pattern is used on a lot of scala types like `Try` and `Future`.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  override def bytes: ByteVector = ByteVector(network.value.map(c => c.toByte) ++ invoiceAmount.map(c => c.toByte)) //TODO: Verify
+}
+
+case class HumanReadablePart(inNetwork: LnParams, amnt: Option[LnCurrencyUnit]) extends LnHumanReadablePart {
+  def network: LnParams = inNetwork
+  def amount: Option[LnCurrencyUnit] = amnt
+}
+
+case object lnbc extends LnHumanReadablePart {
```
Christewart: This isn't quite right, you are missing 

```scala

sealed abstract class LnPrefix 

case object lnbc extends LnPrefix  {
  def network: LnParams = MainNet 
  //note we do not want the `LnCurrencyUnit` here, put that in LnHumanReadablePart
}
```

and then `LnHumanReadablePart` should be

```scala

case class LnHumanReadablePart(prefix: LnPrefix, amount: Option[LnCurrency])

```

TannerR1776: I'm a bit confused on this one. A few questions.
1. Are we going with `LnPrefix` or `LnParams` as the name for the network piece?
2. I dont think the `lnbc`, `lntb`, and `lnbcrt` should extend the `LnPrefix/Params`, as that is what [these are for](https://github.com/bitcoin-s/bitcoin-s-core/blob/09014a223c98244af8f98b2fc423ed722bb3b827/core/src/main/scala/org/bitcoins/core/protocol/ln/LnParams.scala#L10)?
3. These objects exists because similar objects exist in the Bitcoin HRP. However, they are not actually be used at the moment. Should we remove them all together?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
TannerR1776: Very cool! I have have been missing my null coalescing and ternary operators from C#. I guess this will have to be close enough.  :smiley:

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
TannerR1776: Makes total sense. Do you want that functionality as part of this PR, or separately?

TannerR1776: I already forgot, we actually have a `.toEncodedString` defined in `LnCurrencyUnit`, which does essentially this. However, it does currently return a 0 value, which I dont think would be desirable in this case. We would rather just return a `""` in that case.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
Christewart: Null is bad, *never* use null!!! Always encode things like null into the type system with `Option[T]` so we don't get random `NullPointerExceptions`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
Christewart: I'm envisioning `LnParams` is similar to [`ChainParams`](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala#L26) in bitcoin, which has other relevant information about the network (like ports, dns seeds, etc). The `value` field should probably be moved out of `LnParams` IMO.

Christewart: Thinking about this more, a `LnParams` should probably have a `ChainParams` val inside of it.

```scala
sealed abstract class LnParams {
  /** The underlying blockchain the lightning network is using */
  def chain: ChainParams = network.chainParams
  /** The underlying blockchain's p2p network parameters */
  def network: NetworkParameters
  /** The port to connect to our lightning network nodes rpc client */
  def lnRpcPort: Int
  /** The port that our lightning network node is listening on p2p lightning network messages */
  def lnPort : Int 
}
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
Christewart: I'm not sure if I understand the case where it is valid to return just a `""`. I think we should probably encode these methods as `Option[String]` instead and return `None` instead of `""` in the case that we don't have anything too return. In general, we try to encode things at the type level using `Option` rather than doing it at runtime with a `NULL`, `""`, `-1` or whatever your preferred C idiom is :-). The type system is our friend, right @nkohen ?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  def apply(network: LnParams, amount: LnCurrencyUnit): LnHumanReadablePart = {
+    HumanReadablePart(network, Some(amount))
+  }
+
+  def apply(input: String): Try[LnHumanReadablePart] = {
+    val networkPattern: Regex = "^[a-z]*".r //Select all of the letters, until we hit a number, as the network
+    val networkString = networkPattern.findFirstIn(input).getOrElse("")
+    val amountString = input.substring(networkString.length) //Select the remaining part of the string as the amount
```
Christewart: Very important to note here that there is an implicit assumption that the `input` string is already formatted when it is passed in. 

What I mean by this is `input` MUST stop before the separator (`1`) that is specified in Bech32. This isn't necessarily wrong, but it should be noted on a comment on the PR. 

Sometimes when parsing things, you pass in excess information. In the case of the LN Invoice it might be the ENTIRE invoice as `input` rather than just the HumanReadablePart`. If you passed in the entire invoice, this method would fail. 

Some would say the desired behavior would be if you pass in the ENTIRE invoice, it just parses the HumanReadablePart and returns it successfully. 

Would be valid: 
`lnbc2500u`

Would not be valid:
`lnbc2500u1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdpquwpc4curk03c9wlrswe78q4eyqc7d8d0xqzpuyk0sg5g70me25alkluzd2x62aysf2pyy8edtjeevuv4p2d5p76r4zkmneet7uvyakky2zr4cusd45tftc9c5fh0nnqpnl2jfll544esqchsrny`

TannerR1776: This is correct. And, I agree that it would be nice to be able to pass in an entire invoice, and return an HRP. However, we are going to have to think about this a little bit. For example, using the invoice you showed, what if we simply removed the amount (which, in my understanding, is completely valid)? We then get this `lnbc1pvjluez`.

How do we distinguish `1p` as the Bech32 separator and time field, and not read it in as 1 PicoSatoshi?

Perhaps it is not the job of LnHRP to correctly deserialize a LnInvoice? What if LnInvoice has its own deserialization, where it decodes all of the invoice pieces, and correctly identifies `1pvjluez` as the Bech32 separator and time field?

If its able to correctly identify `1pvjluez`, it could correctly pass in `lnbc` to LnHRP.

Christewart: >what if we simply removed the amount

This would invalidate the checksum on the invoice -- but you do have a great point. 

>Perhaps it is not the job of LnHRP to correctly deserialize a LnInvoice? What if LnInvoice has its own deserialization, where it decodes all of the invoice pieces, and correctly identifies 1pvjluez as the Bech32 separator and time field?

Agreed.

PR #191, opened 2018-09-04
WIP: TestUtil Refactor

nkohen:
Changing logic that blocks on a given condition to use the ActorSystem scheduler instead of Thread.sleep

Adding tests for that logic

### DIFF
--- a/rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
+++ b/rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
@@ -1,0 +1,93 @@
+package org.bitcoins.rpc
+
+import akka.actor.ActorSystem
+import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.rpc.client.BitcoindRpcClient
+
+import scala.concurrent.duration.{ DurationInt, FiniteDuration }
+import scala.concurrent.{ Await, Future, Promise }
+
+trait RpcUtil extends BitcoinSLogger {
+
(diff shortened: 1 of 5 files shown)


rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
```diff
@@ -10,13 +10,13 @@ import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.rpc.client.RpcClient
 import org.bitcoins.rpc.config.{ AuthCredentials, DaemonInstance }
 
-import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration.DurationInt
+import scala.concurrent.{ Await, Future, Promise }
 import scala.util.Try
 
-trait TestUtil extends BitcoinSLogger {
+trait RpcUtil extends BitcoinSLogger {
```
Christewart: What i meant here is keeping a `TestUtil`, but moving the timing stuff into a new src file called `RpcUtil`

rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
```diff
@@ -10,13 +10,13 @@ import org.bitcoins.core.util.BitcoinSLogger
...
 
-  implicit val system = ActorSystem()
+  implicit val system = ActorSystem("TestUtil_ActorSystem")
```
Christewart: In general, let's make `ActorSystem` be an implicit parameter to a method rather than creating a val inside of `TestUtil`

PR #172, opened 2018-06-05
Added scalafmt plugin and .scalafmt.conf file.

nkohen:
Run in sbt with the command
> scalafmt
for compile sources, and
> test:scalafmt
for test sources

Check compile sources for good formatting with
> scalafmt::test
and check test sources for good formatting with
>test:scalafmt::test

PR #192, opened 2018-09-06
Add 'Bitcoind' prefix to all bitcoind specific rpc stuff

Christewart:
Adds the `Bitcoind` prefix to all bitcoind specific Rpc stuff. In general, we should prefix what network/rpc server rpc stuff is specific for. I.e. `EclairRpcClient` for an eclair rpc client.


torkelrogstad:
Good idea:-) 

Relevant to my questions on formatting in Slack: The diff has some extra lines in it consisting of just imports moving around or being spread out/collapsed. In the ideal programming environment, every single character's position should be determined deterministically, and your diff would be even easier to inspect at a glance.

Christewart:
@torkelrogstad Yes, this is unfortunate at the moment. We are in between code formatting tools -- we are planning to move to `scalafmt` in #172. We've been waiting on some closed source stuff from another company that is going to be open sourced soon(TM) so we don't have a crazy amount of merge conflicts. But they are taking a long time. So maybe should just bite the bullet and merge 172.

PR #193, opened 2018-09-12
Eclair rpc v2

Christewart:
This builds off of #188 

This PR aims to implement a RPC for [eclair](https://github.com/acinq/eclair) -- a lightning network implementation. 

We are aiming to have a full RPC client with a comprehensive set of unit tests that open create their own eclair instances, run unit tests against those eclair instances, and then destroy them.

Currently we have a unit test implemented for `open`

Here is a full list of Eclair's RPCs

https://github.com/acinq/eclair#json-rpc-api

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: scala
 jdk: oraclejdk8
 
 scala:
-  - 2.11.7
+  - 2.11.12
 
 # These directories are cached to S3 at the end of the build
 cache:
@@ -36,6 +36,9 @@ install:
   - ./configure --enable-jni --enable-experimental --enable-module-ecdh
   - sudo make install
   - cd ../
+  - wget https://github.com/ACINQ/eclair/releases/download/v0.2-beta5/eclair-node-0.2-beta5-8aa51f4.jar
+  - export ECLAIR_PATH=$(pwd)
+
 script:
   - sbt -Djava.library.path=secp256k1/.libs clean coverage test &&
     sbt coverageReport &&
--- a/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
+++ b/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
@@ -1,0 +1,389 @@
+package org.bitcoins.eclair.rpc.client
+
+import java.util.UUID
+
+import akka.http.javadsl.model.headers.HttpCredentials
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model._
+import akka.stream.ActorMaterializer
+import akka.util.ByteString
+import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.eclair.rpc.config.EclairInstance
+import org.bitcoins.eclair.rpc.json._
+import org.bitcoins.rpc.RpcUtil
+import play.api.libs.json._
+
+import scala.concurrent.{ Await, ExecutionContext, Future }
+import scala.concurrent.duration.DurationInt
+import scala.sys.process._
+import scala.util.Try
+
+class EclairRpcClient(instance: EclairInstance)(implicit m: ActorMaterializer) {
+  import JsonReaders._
+
+  private val resultKey = "result"
+  private val errorKey = "error"
+  implicit val ec: ExecutionContext = m.executionContext
+  private val logger = BitcoinSLogger.logger
+
+  def getDaemon: EclairInstance = instance
+
+  def allChannels: Future[Vector[ChannelDesc]] = {
+    eclairCall[Vector[ChannelDesc]]("allchannels")
+  }
+
+  def allNodes: Future[Vector[NodeInfo]] = {
+    eclairCall[Vector[NodeInfo]]("allnodes")
+  }
+
+  private def allUpdates(
+    nodeId: Option[String]): Future[Vector[ChannelUpdate]] = {
+    val params = if (nodeId.isEmpty) List.empty else List(JsString(nodeId.get))
+    eclairCall[Vector[ChannelUpdate]]("allupdates", params)
+  }
+
+  def allUpdates: Future[Vector[ChannelUpdate]] = allUpdates(None)
+
+  def allUpdates(nodeId: String): Future[Vector[ChannelUpdate]] =
+    allUpdates(Some(nodeId))
+
+  def channel(channelId: String): Future[ChannelResult] = {
+    eclairCall[ChannelResult]("channel", List(JsString(channelId)))
+  }
+
+  private def channels(nodeId: Option[String]): Future[Vector[ChannelInfo]] = {
+    val params = if (nodeId.isEmpty) List.empty else List(JsString(nodeId.get))
+
+    eclairCall[Vector[ChannelInfo]]("channels", params)
+  }
+
+  def channels: Future[Vector[ChannelInfo]] = channels(None)
+
+  def channels(nodeId: String): Future[Vector[ChannelInfo]] =
+    channels(Some(nodeId))
+
+  def checkInvoice(invoice: String): Future[PaymentRequest] = {
+    eclairCall[PaymentRequest]("checkinvoice", List(JsString(invoice)))
+  }
+
+  // When types are introduced this can be two different functions
+  def checkPayment(invoiceOrHash: String): Future[Boolean] = {
+    eclairCall[Boolean]("checkpayment", List(JsString(invoiceOrHash)))
+  }
+
+  private def close(
+    channelId: String,
+    scriptPubKey: Option[String]): Future[String] = {
+    val params =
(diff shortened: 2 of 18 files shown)


core-test/src/test/resources/logback-test.xml
```diff
@@ -14,7 +14,7 @@
     </appender>
 
 
-    <root level="INFO">
+    <root level="OFF">
```
nkohen: Not even ERROR for CI purposes?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+  override def beforeAll(): Unit = {
+    logger.info("Temp eclair directory created")
+    logger.info("Temp eclair directory created")
+
+    logger.debug(client.getDaemon.authCredentials.datadir.toString)
+
+    client.start()
+    otherClient.start()
+
+    Thread.sleep(20000)
```
nkohen: We should probably use the RpcUtil stuff

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    val infoF = otherClient.getInfo
+
+    val connected: Future[String] = infoF.flatMap { info =>
+      client.connect(info.nodeId, "localhost", info.port)
+    }
+
+    Await.result(connected, 15.seconds)
```
nkohen: RpcUtil here as well

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+  private def hasChannel(client: EclairRpcClient, chanId: String): Future[Assertion] = {
+    val recognizedOpenChannel: Future[Assertion] = {
+      val chanResultF: Future[ChannelResult] = client.channel(chanId)
+      chanResultF.map(c => assert(c.channelId == chanId))
+    }
+
+    recognizedOpenChannel
+  }
+
+  override def afterAll(): Unit = {
```
nkohen, line 94: Should probably terminate the ActorSystem here as well

.travis.yml
```diff
@@ -36,6 +36,9 @@ install:
   - ./configure --enable-jni --enable-experimental --enable-module-ecdh
   - sudo make install
   - cd ../
+  - wget https://github.com/ACINQ/eclair/releases/download/v0.2-beta5/eclair-node-0.2-beta5-8aa51f4.jar
+  - export ECLAIR_PATH=$(pwd)
```
nkohen, line 40: Document this?

nkohen (changes requested):

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,382 @@
...
+  def start(): Unit = {
+    val p = Process("java -jar -Declair.datadir=" + instance.authCredentials.datadir + s" ${pathToEclairJar} &")
+    val result = p.run()
+    logger.info(s"Starting eclair with datadir ${instance.authCredentials.datadir}")
+
+    process = Some(result)
+    ()
+  }
+
+  def isStarted(): Boolean = {
```
nkohen: We could probably do a check for `process == None` before stalling the program that calls this for 1.second

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -0,0 +1,188 @@
+package org.bitcoins.eclair.rpc.json
+
+import play.api.libs.json.{ JsArray, JsObject }
+
+sealed abstract class EclairModels
```
nkohen, line 5: Is there a reason that this isn't extended by anything in here? Also is there a difference between `abstract class` and `trait` in this instance?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -0,0 +1,100 @@
...
+    logger.info("Temp eclair directory created")
+    logger.info("Temp eclair directory created")
+
+    logger.debug(client.getDaemon.authCredentials.datadir.toString)
+
+    client.start()
+    otherClient.start()
+
+    RpcUtil.awaitCondition(
+      condition = () => client.isStarted(),
```
nkohen: are these`() =>` necessary? I think it's a bit more reader-friendly to drop them

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    recognizedOpenChannel
+  }
+
+  override def afterAll(): Unit = {
+    val s1 = client.stop()
+    val s2 = otherClient.stop()
+    val s3 = bitcoindRpc.stop()
+    system.terminate()
```
nkohen, line 98: This returns a Future we should probably block on

rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
```diff
@@ -68,25 +78,48 @@ trait RpcUtil extends BitcoinSLogger {
    * @param system An ActorSystem to schedule calls to condition
    */
   def awaitCondition(
-    condition: => Boolean,
+    condition: () => Boolean,
+    duration: FiniteDuration = 100.milliseconds,
+    maxTries: Int = 50,
+    overallTimeout: FiniteDuration = 1.hour)(implicit system: ActorSystem): Unit = {
+    implicit val ec = system.dispatcher
+    def h: Boolean = condition()
```
nkohen: why is this needed? I feel like `condition` is a more descriptive name and should just be used where `h` is

rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
```diff
@@ -17,42 +17,52 @@ trait RpcUtil extends BitcoinSLogger {
...
-   * @param condition The condition being waited on
+   * @param conditionF The condition being waited on
    * @param duration The interval between calls to check condition
    * @param maxTries If condition is tried this many times, the Future fails
    * @param system An ActorSystem to schedule calls to condition
    * @return A Future[Unit] that succeeds if condition becomes true and fails otherwise
    */
-  def retryUntilSatisfied(
-    condition: => Boolean,
+  def retryUntilSatisfiedF(
```
nkohen, line 34: I think it wouldn't hurt to also have a `retryUntilSatisfied` that takes in a `=> Boolean` with no `Future` which could literally be

```
def retryUntilSatisfiedF(
  condition: => Boolean,
  maxTries: Int = 50)(implicit system: ActorSystem): Future[Unit] = {
    retryUntilSatisfiedF(Future.successful(conditionF), duration, maxTries)
}

nkohen (changes requested):

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
Christewart: I this case they are not necessary, but in the more general case they are. I.e. if `client.isStarted()` took params you would need to use an explicit anonymous function.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
Christewart: I default to `abstract class` over `trait` because it is "more native" to the JVM.

rpc/src/main/scala/org/bitcoins/rpc/RpcUtil.scala
Christewart: This is a bit of type hackery going on here. Basically we need to transform from a 

`() => Boolean`

to 

`() => Future[Boolean]`

the trick is figuring out how to do this WITHOUT caching the value -- but making it so that the `condition` is re-evaluated everytime you call `retryUntilSatisfied`.

nkohen: Got it, a short comment and maybe the name `conditionCallByName` or `conditionDef` wouldn't hurt?

PR #196, opened 2018-09-15
Add NodeId, NodeUri, ChannelId

Christewart:
This adds three new types that are used in the LN network

1. NodeId - represents a unique identifier for a node
2. NodeUri - a uri you can use to connect to on the LN
3. ChannelId - a unique identifer for a lightning network channel -- has two sub types -- `TempChannelId` and `FundedChannelId`

### DIFF
--- a/rpc/src/main/scala/org/bitcoins/rpc/client/BitcoindRpcClient.scala
+++ b/rpc/src/main/scala/org/bitcoins/rpc/client/BitcoindRpcClient.scala
@@ -3,6 +3,7 @@ package org.bitcoins.rpc.client
 import java.net.URI
 import java.util.UUID
 
+import akka.actor.ActorSystem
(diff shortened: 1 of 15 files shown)


rpc/src/main/scala/org/bitcoins/rpc/client/BitcoindRpcClient.scala
```diff
@@ -29,30 +31,41 @@ import scala.util.Try
...
-    result.getOrElse(false)
+      result
+    }
+
+    val await = Try(
+      RpcUtil.awaitConditionF(
+        conditionF = isConnected,
+        duration = 1.seconds))
+
+    if (await.isSuccess) true else false
```
nkohen, line 68: Why not just `await.isSuccess`?

PR #197, opened 2018-09-16
Use Gen.delay instead of Gen.const for generating privateKey

boatkorachal:
When using `Gen.listOf` for generating list of private keys or other cryptos that derived from private keys, it generates all the same objects in list.


Christewart:
Awesome catch! I really appreciate it. Not sure if you are going to continue doing dev with bitcoin-s but the most active channel is currently on our slack: 

https://join.slack.com/t/suredbits/shared_invite/enQtNDEyMjY3MTg1MTg3LTYyYjkwOGUzMDQ4NDAwZjE1M2I3MmQyNWNlZjNlYjg4OGRjYTRjNWUwNjRjNjg4Y2NjZjAxYjU1N2JjMTU1YWM

Feel free to join or just continue discussing things here :-)

PR #200, opened 2018-09-20
Two way eclair transactions sanity test

nkohen:
Added a test that currently fails in which two nodes attempt to send each other payments.


nkohen:
Changed which client does the call to `checkPayment`, it works now :)

PR #215, opened 2018-10-03
[WIP] PSBT RPC calls

torkelrogstad:
This compiles, but it doesn't work. Making this PR just some code can be shared.

Question on formatting: The format functionality in Intellij changed some whitespace here and there. Is there any configs I/we can tweak to make it work the same for everybody? I guess #172 is trying to tackle exactly that...


torkelrogstad:
There's also some `String`s that can be turned into more specific types here.

PR #217, opened 2018-10-13
Add serialization symmetry property for LnInvoice, fixing various bug…

Christewart:
This incorporates property based testing features for our `LnInvoice` implementation. This also fixes a few bugs that were found in our LnInvoice implementation that was previous merged.

### DIFF
--- a/core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
+++ b/core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
@@ -1,0 +1,129 @@
+package org.bitcoins.core.gen.ln
+
+import org.bitcoins.core.gen._
+import org.bitcoins.core.number.{ UInt64, UInt8 }
+import org.bitcoins.core.protocol.ln._
+import org.bitcoins.core.protocol.ln.routing.LnRoute
+import org.bitcoins.core.util.NumberUtil
+import org.scalacheck.Gen
+
+sealed abstract class LnInvoiceGen {
+
+  /**
+   * Generates a [[org.bitcoins.core.protocol.ln.LnHumanReadablePart]]
+   * that does not contain a amount
+   * @return
+   */
+  def lnHrpNoAmt: Gen[LnHumanReadablePart] = {
+    ChainParamsGenerator.lnNetworkParams.flatMap {
+      lnParam =>
+        LnHumanReadablePart.fromLnParams(lnParam)
+    }
+  }
+
+  /**
+   * Generates a [[org.bitcoins.core.protocol.ln.LnHumanReadablePart]]
+   * with an amount encoded
+   */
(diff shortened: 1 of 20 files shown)


core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
```diff
@@ -0,0 +1,129 @@
...
+
+sealed abstract class LnInvoiceGen {
+
+  /**
+   * Generates a [[org.bitcoins.core.protocol.ln.LnHumanReadablePart]]
+   * that does not contain a amount
+   * @return
+   */
+  def lnHrpNoAmt: Gen[LnHumanReadablePart] = {
+    ChainParamsGenerator.lnNetworkParams.flatMap {
```
torkelrogstad, line 18: Is there any difference in doing `flatMap` vs `map` here?

core-test/src/test/scala/org/bitcoins/core/protocol/blockchain/BlockTest.scala
```diff
@@ -37,13 +37,13 @@ class BlockTest extends FlatSpec with MustMatchers {
     val fileName = "/00000000000000000008513c860373da0484f065983aeb063ebf81c172e81d48.txt"
     val lines = Source.fromURL(getClass.getResource(fileName)).mkString
     val time = timeBlockParsing(Block.fromHex(lines))
-    assert(time <= 15000)
+    assert(time <= 20000)
```
torkelrogstad, line 40: What's the motivation for increasing this threshold?

core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
Christewart: Yes. If we did not use `.flatMap` here the resulting type would be `Gen[Gen[LnHumanReadablePart]]`. 

The use of `.flatMap` here collapses the `Gen[Gen[]]` into just one `Gen[]`. 

You will see this pattern used consitently across Scala code bases, for instance the same applies to `Future[Future[T]]`. 

```scala
//since we use `.map` here get nested futures
val x: Future[Future[Int]] = Future.successful(1).map(_ => Future.successful(2))

val y: Future[Int] = Future.successful(1).flatMap(_ => Future.successful(2))
```

core-test/src/test/scala/org/bitcoins/core/protocol/blockchain/BlockTest.scala
Christewart: i was getting spurious failures on my machine with this (probably because I was running too much stuff). We've also experience failures on travis due to this as well. Frankly we need to keep improving performance on bitcoin-s, but it is of lower priority.

PR #218, opened 2018-10-13
Reworking AuthCredentials and Instances so that we can read from conf…

Christewart:
This PR makes it easier to read config files for `EclairInstance`, also makes it easier to get `BitcoinRpcClient` from `EclairInstance` in test functionality. This makes it easier to do things to the underlying blockchain when testing.


Christewart:
This also fixes #216

PR #222, opened 2018-10-20
compiler opts

ghost:
https://github.com/bitcoin-s/bitcoin-s-core/issues/195

Need to remove more warnings. Most of them are just numeric widening

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,22 @@
 
-
-lazy val compilerOpts =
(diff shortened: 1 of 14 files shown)


Christewart:
Not that you should add more of these flags, but i just wanted to add this link here for future documentation: https://docs.scala-lang.org/overviews/compiler-options/index.html

torkelrogstad:
I think it makes sense to add some more warnings for unused stuff, for example `-Ywarn-unused:params`, `-Ywarn-unused:locals` and `-Ywarn-unused:privates`

PR #230, opened 2018-11-06
Derive nodeId from ln invoice signature, move nodeid case class into …

Christewart:
…the core project

BOLT11 specifies that if a `NodeIdTag` is not explicitly stated in the `LnInvoice` you need to derive it from the `LnInvoiceSignature` using some elliptic curve math. 

This also moves the `NodeId` into the `core` project.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/util/CryptoUtilTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/util/CryptoUtilTest.scala
@@ -1,12 +1,16 @@
 package org.bitcoins.core.util
 
+import org.bitcoins.core.gen.{ CryptoGenerators, NumberGenerator }
+import org.scalatest.prop.PropertyChecks
 import org.scalatest.{ FlatSpec, MustMatchers }
+import org.slf4j.LoggerFactory
+import scodec.bits.ByteVector
 
 /**
  * Created by chris on 1/26/16.
  */
 class CryptoUtilTest extends FlatSpec with MustMatchers {
-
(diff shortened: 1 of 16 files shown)


core-test/src/test/scala/org/bitcoins/core/util/CryptoUtilTest.scala
```diff
@@ -55,4 +59,14 @@ class CryptoUtilTest extends FlatSpec with MustMatchers {
     CryptoUtil.sha256Hash160(hex).hex must be(expected)
     CryptoUtil.sha256Hash160(hex).flip.flip.hex must be(expected)
   }
-}
+
+  it must "recover the two public keys used to sign a message" in {
+    PropertyChecks.forAll(CryptoGenerators.privateKey, CryptoGenerators.sha256Digest) {
+      case (privKey, hash) =>
```
torkelrogstad, line 65: What does the `case` keyword mean here?

Christewart: The `case` keyword used to enumerate a condition that your code should handle. 

You can use case inside of pattern matches like 

```scala
val x = "Hello World" 
x match {
  case "Hello World" => println("matched found hello world") 
  case _: String => println("Did not match hello world")
}
```
You can also use `case` inside of lambdas like this 

```scala
//define an anonymous function that prints the given string
{ case x : String => println(x) }
```
but you could also write this

```scala
{ x: String => println(x) }
```

In the specific instance you are talking about, the `case` keyword is not strictly necessary, and I could have removed it. One of the good things (and bad things) about Scala is you can be extremely flexible with syntax.

PR #231, opened 2018-11-07
[WIP] Rpc client refactor

torkelrogstad:
Here's the changes in 0.16 => 0.17, from official [docs](https://bitcoin.org/en/release/v0.17.0#label-and-account-apis-for-wallet).

There was big changes in how accounts are handled. Here's the changes relating to that:

Deprecated Method | New Method | Notes | Status 
-- | -- | -- | --
getaccount | getaddressinfo | getaddressinfo returns a json object with address information instead of just the name of the account as a string. | *Almost done*. `DateTime` chrases horribly 
getaccountaddress | n/a | There is no replacement for getaccountaddress since labels do not have an associated receive address. | **Done**
getaddressesbyaccount | getaddressesbylabel | getaddressesbylabel returns a json object with the addresses as keys, instead of a list of strings. | Not done 
getreceivedbyaccount | getreceivedbylabel | no change in behavior | **done** 
listaccounts | listlabels | listlabels does not return a balance or accept minconf and watchonly arguments. | Not done 
listreceivedbyaccount | listreceivedbylabel | Both methods return new labelfields, along with account fields for backward compatibility. | Not done 
move | n/a | no replacement | **Done**
sendfrom | n/a | no replacement | **Done**
setaccount | setlabel | Both methods now: <ul><li>allow assigning labels to any address, instead of raising an error if the address is not receiving address.<li>delete the previous label associated with an address when the final address using that label is reassigned to a different label, instead of making an implicit getaccountaddress call to ensure the previous label still has a receiving address. | Not done 

There was also minor changes to other RPC methods:

Changed Method | Notes | Status
-- | -- | -- 
addmultisigaddress  | Renamed account named parameter to label. Still accepts account for backward compatibility if running with ‘-deprecatedrpc=accounts’.  | Not done 
getnewaddress | Renamed account named parameter to label. Still accepts account for backward compatibility. if running with ‘-deprecatedrpc=accounts’ | Not done 
listunspent | Returns new label fields. account field will be returned for backward compatibility if running with ‘-deprecatedrpc=accounts’ | Not done 
sendmany | The account named parameter has been renamed to dummy. If provided, the dummy parameter must be set to the empty string, unless running with the -deprecatedrpc=accounts argument (in which case functionality is unchanged). | Not done 
listtransactions  | The account named parameter has been renamed to dummy. If provided, the dummy parameter must be set to the string *, unless running with the -deprecatedrpc=accounts argument (in which case functionality is unchanged). | Not done 
getbalance | account, minconf and include_watchonly parameters are deprecated, and can only be used if running with ‘-deprecatedrpc=accounts’ | Not done


rpc/src/main/scala/org/bitcoins/rpc/client/Client.scala
```diff
@@ -0,0 +1,168 @@
...
+  /**
+    * This is here (and not in BitcoindJsonWrriters)
+    * so that the implicit network val is accessible
+     */
+  implicit object ECPrivateKeyWrites extends Writes[ECPrivateKey] {
+    override def writes(o: ECPrivateKey): JsValue = JsString(o.toWIF(network))
+  }
+
+  implicit val eCPrivateKeyWrites: Writes[ECPrivateKey] = ECPrivateKeyWrites
+  implicit val importMultiAddressWrites: Writes[RpcOpts.ImportMultiAddress] =
```
torkelrogstad, line 40: If I move this implicit and the one below it anywhere else (like `RpcOpts` or `BitcoindJsonWriters`), I get an error. I don't understand why that is. Is it possible to get an explanation for that?

rpc/src/main/scala/org/bitcoins/rpc/client/Client.scala
```diff
@@ -0,0 +1,168 @@
...
+import java.util.UUID
+
+import akka.http.javadsl.model.headers.HttpCredentials
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model._
+import akka.stream.ActorMaterializer
+import akka.util.ByteString
+import org.bitcoins.core.config.NetworkParameters
+import org.bitcoins.core.crypto.ECPrivateKey
+import org.bitcoins.rpc.serializers.BitcoindJsonWriters._
```
torkelrogstad: This line show's as unused in Intellij (and is removed if I use the optimize imports), but I need the implicits in there to get it to compile. Am I doing something wrong?

nkohen: Nope, you’re just fine, IntelliJ just isn’t smart enough to see some implicit that are called within others

rpc/src/main/scala/org/bitcoins/rpc/client/Client.scala
nkohen: These implicits depend on Writes[ECPrivateKey] since the case classes contain an ECPrivateKey and so that Writes must be in scope where these are defined

rpc/src/main/scala/org/bitcoins/rpc/client/Client.scala
torkelrogstad: Is it possible to prevent Intellij from removing that line on import optimization?

torkelrogstad:
The only failing tests seem to be a timeout issue.

rpc/src/main/scala/org/bitcoins/rpc/client/BlockchainRpc.scala
```diff
@@ -0,0 +1,196 @@
...
+import org.bitcoins.rpc.serializers.BitcoindJsonReaders._
+import org.bitcoins.rpc.serializers.BitcoindJsonSerializers._
+import play.api.libs.json.{ JsBoolean, JsNumber, JsString }
+
+import scala.concurrent.Future
+
+/**
+ * RPC calls related to querying the state of the blockchain
+ */
+protected trait BlockchainRpc extends Client {
```
nkohen: In general we should leave these top level traits public and using protected/private for lower level `def`s and `val`s

rpc/src/main/scala/org/bitcoins/rpc/client/Client.scala
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  implicit val eCPrivateKeyWrites: Writes[ECPrivateKey] = ECPrivateKeyWrites
+  implicit val importMultiAddressWrites: Writes[RpcOpts.ImportMultiAddress] =
+    Json.writes[RpcOpts.ImportMultiAddress]
+  implicit val importMultiRequestWrites: Writes[RpcOpts.ImportMultiRequest] =
+    Json.writes[RpcOpts.ImportMultiRequest]
+
+  protected val resultKey: String = "result"
+  protected val errorKey: String = "error"
```
nkohen: Could these be private?

nkohen (commented):
Looking pretty good! I think it would also be a good idea to split up the giant test into multiple files mirroring what has been done here (i.e. each test tests a call so calls in the same file should have tests in the same file but not just one big test file). Don't bother cleaning up the giant ugly tests full of lambdas if you don't want to (I wouldn't :P) but just copy and paste them over to new files

rpc/src/main/scala/org/bitcoins/rpc/client/BlockchainRpc.scala
torkelrogstad: My reasoning was that these traits will only be used by `BitcoindRpcClient`. If they're public they will "pollute" the global import scope (what Intellij suggests as imports). Or is there something I have misunderstood?

Christewart: I think you have a good understanding of how they work, but it is fairly uncommon in Scala to use access modifiers on `trait` or `class`. The most common access modifier at the trait level is `sealed` which is used to create an algebraic data type. 

I think we can have a discussion later about doing access modifiers on these traits, but let's table it for now.

rpc/src/test/scala/org/bitcoins/rpc/TestUtil.scala
```diff
@@ -73,7 +73,11 @@ trait TestUtil extends BitcoinSLogger {
     } else firstAttempt
   }
 
-  def startServers(servers: Vector[BitcoindRpcClient])(implicit system: ActorSystem): Unit = {
+  /**
+   * Starts the given servers, and waits for them (in a blocking manner)
+   * until they are started.
+   */
+  def startServers(servers: BitcoindRpcClient*)(implicit system: ActorSystem): Unit = {
```
Christewart, line 83: A word of caution here with `*` (var args) is performance considerations. Scala actually has a very complicated story with collections and performance. `Vector` is usually the best performing general data structure. 

I think this is ok to be using in test code, but would push back against using var args in source code. 

Here is an article that talks more about scala collections performance:

http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html

https://docs.scala-lang.org/overviews/collections/performance-characteristics.html

rpc/src/test/scala/org/bitcoins/rpc/TestUtil.scala
```diff
@@ -92,7 +96,7 @@ trait TestUtil extends BitcoinSLogger {
     to: BitcoindRpcClient,
     duration: FiniteDuration = 100.milliseconds,
     maxTries: Int = 50)(implicit system: ActorSystem): Unit = {
-    implicit val ec = system.dispatcher
+    implicit val ec: ExecutionContextExecutor = system.dispatcher
```
Christewart: This should probably be typed as 

```scala
implicit val ec: ExecutionContext = system.dispatcher
```

The `ExecutionContext` is what is required by `.map` or `flatMap` to execute

rpc/src/test/scala/org/bitcoins/rpc/TestUtil.scala
torkelrogstad: Allright. Var args are cool so I wanted to use them here, didn't know they had performance implications!

rpc/src/main/scala/org/bitcoins/rpc/client/BlockchainRpc.scala
torkelrogstad: Does that mean I should keep the `protected` modifiers or remove them? I have no strong opinion on that

rpc/src/test/scala/org/bitcoins/rpc/TestUtil.scala
Christewart: It should be noted that they *can* -- but not necessarily -- have them. A bigger goal of mine is to improve performance in bitcoin-s, which means limiting what is possible in certain areas of the codebase. 

Somehthing like varargs if hard to reason about because you can pass in a lot of different things! :-). This is just test code though, so it isn't a huge concern here.

torkelrogstad: That totally makes sense!

rpc/src/main/scala/org/bitcoins/rpc/client/BlockchainRpc.scala
nkohen: I'd say remove them for now from the top-level traits and classes if you have them

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -108,6 +110,18 @@ case class ReceivedAccount(
   lable: Option[String])
   extends WalletResult
 
+case class ReceivedLabel(
+  involvesWatchonly: Option[Boolean],
+  /**
```
Christewart, line 115: This was deprecated in `0.17` right? Technically I think we should use the `@deprecated` annotation on the method for now

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -172,3 +186,34 @@ case class UnspentOutput(
...
+  scriptPubKey: ScriptPubKey,
+  ismine: Boolean,
+  iswatchonly: Boolean,
+  isscript: Boolean,
+  iswitness: Boolean,
+  // witness_version: Option[WitnessVersion],
+  // TODO something better here
+  witness_program: Option[String],
+  // TODO Option[Script]
+  script: Option[String],
```
Christewart: Probably `Option[ScriptPubKey]`

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -172,3 +186,34 @@ case class UnspentOutput(
...
+case class AddressInfoResult(
+  address: BitcoinAddress,
+  scriptPubKey: ScriptPubKey,
+  ismine: Boolean,
+  iswatchonly: Boolean,
+  isscript: Boolean,
+  iswitness: Boolean,
+  // witness_version: Option[WitnessVersion],
+  // TODO something better here
+  witness_program: Option[String],
```
Christewart, line 200: `Option[ScriptWitness]`

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -172,3 +186,34 @@ case class UnspentOutput(
...
+  // TODO Option[Script]
+  script: Option[String],
+  // TODO something better here
+  hex: String,
+  // TODO something better here
+  pubkeys: Option[Vector[ECPublicKey]],
```
Christewart: This looks ok?

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
torkelrogstad: It's actually accounts that's deprecated, not labels. I was confused previously

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
torkelrogstad: Possible values are:  `nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash, witness_unknown`. `ScriptPubKey` is not that, if I've understood it correctly?

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
torkelrogstad: There's no `Reads` for `ScriptWitness`

Issue #236, opened 2018-11-08
Add test case to eclair / bitcoind rpc to make sure we CANNOT authenticate if we provide the wrong credentials


Christewart:
Basically if we provide the wrong password, we shouldn't able to authenticate. These are sanity test cases

PR #237, opened 2018-11-08
Cleaned up eclair conf

nkohen:
Removed username/alias from eclair credentials, added default ports if they are not specified


Christewart:
This resolves half of the issue in #236

PR #242, opened 2018-11-12
[WIP] Some more async-await-ify of tests

torkelrogstad:


rpc/src/test/scala/org/bitcoins/rpc/RpcUtilTest.scala
```diff
@@ -103,35 +102,45 @@ class RpcUtilTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
               assert(!client2.getDaemon.authCredentials.datadir.exists)
             }
-          }
-        }
-    }
-  }
 
-  override def afterAll(): Unit = {
-    Await.result(system.terminate(), 10.seconds)
+  override def afterAll(): Unit = async {
```
Christewart, line 143: I think maybe we should just use 

```scala
await(system.terminate())
```

with no async wrapper here?

rpc/src/test/scala/org/bitcoins/rpc/RawTransactionRpcTest.scala
```diff
@@ -0,0 +1,176 @@
...
+  val logger: Logger = BitcoinSLogger.logger
+
+  override protected def beforeAll(): Unit = {
+    logger.info("Starting RawTransactionRpcTest")
+    logger.info("Bitcoin servers starting")
+    TestUtil.startServers(client, otherClient)
+    logger.info("Bitcoin servers started")
+  }
+
+  override protected def afterAll(): Unit = async {
```
Christewart, line 36: If I remember `scala-async` correctly you always need to wrap a `await` inside of a `async`. I don't think it makes sense to have this wrapped in `async`/`await`. We don't want `beforeAll`/`afterAll` executed asynchronously. Also, does this type check? I believe this returns a `Future[Unit]` not `Unit`?

Christewart (commented):
I think the `async/await` stuff is the right direction to go, and will make our test code much more legible.

PR #246, opened 2018-11-18
2018 11 18 rework script data structure

Christewart:
This PR does two things

1. Creates a `Script` trait that is extended by both `ScriptSignature` and `ScriptPubKey`
2. Changes the internal storage type of a Script from `ByteVector` -> `Vector[ScriptToken]`.

During parsing -- since we aim to type scripts in bitcoin-s -- we MUST deserialize the `ByteVector` representation of a Script to `Vector[ScriptToken]`. Previously we were discarding the parsed format after we correctly typed the Script. This is wasteful since parsing is expensive, so now we cache the `Vector[ScriptToken]` so we can use that for other things internally.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/Script.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/Script.scala
@@ -1,0 +1,37 @@
+
+package org.bitcoins.core.protocol.script
+
+import org.bitcoins.core.protocol.{ CompactSizeUInt, NetworkElement }
+import org.bitcoins.core.script.constant.ScriptToken
+import scodec.bits.ByteVector
+
+
+/** This is meant to be a super type for
+  * scripts in the bitcoin protocol. This gives us
+  * access to the asm representation, and how to serialize the script
+  */
+abstract class Script extends NetworkElement {
+
+  /**
+    * Representation of a script in a parsed assembly format
+    * this data structure can be run through the script interpreter to
+    * see if a script evaluates to true
+    * used to represent the size of the script serialization
+    */
+  def asm: Seq[ScriptToken]
+
+  /**
+    * The byte representation of [[asm]], this does NOT have the bytes
+    * for the [[org.bitcoins.core.protocol.CompactSizeUInt]] in the script
+    */
+  val asmBytes: ByteVector = {
+    asm.foldLeft(ByteVector.empty)(_ ++ _.bytes)
+  }
+
+  /** The size of the script, this is used for network serialization */
+  val compactSizeUInt: CompactSizeUInt = CompactSizeUInt.calc(asmBytes)
+
+  /** The full byte serialization for a script on the network */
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/Script.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.bitcoins.core.script.constant.ScriptToken
+import scodec.bits.ByteVector
+
+abstract class Script extends NetworkElement {
+
+  /**
+   * Representation of a script in a parsed assembly format
+   * this data structure can be run through the script interpreter to
+   * see if a script evaluates to true
+   * Note: The first byte(s) inside the byte array is the [[CompactSizeUInt]]
```
Christewart: This part of the comment is incorrect

core/src/main/scala/org/bitcoins/core/protocol/script/Script.scala
```diff
@@ -0,0 +1,32 @@
+package org.bitcoins.core.protocol.script
+
+import org.bitcoins.core.protocol.{ CompactSizeUInt, NetworkElement }
+import org.bitcoins.core.script.constant.ScriptToken
+import scodec.bits.ByteVector
+
+abstract class Script extends NetworkElement {
```
Christewart, line 13: Add scaladoc explaining the intention of `Script`

core-test/src/test/scala/org/bitcoins/core/protocol/blockchain/ChainParamsTest.scala
```diff
@@ -60,6 +60,7 @@ class ChainParamsTest extends FlatSpec with MustMatchers {
   it must "generate the correct txid for the genesis transaction" in {
     genesisTransaction.version must be(TransactionConstants.version)
     genesisTransaction.lockTime must be(TransactionConstants.lockTime)
+    genesisTransaction.inputs.head.scriptSignature.asm must be(expectedGenesisInput.scriptSignature.asm)
```
Christewart: remove

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -338,14 +337,23 @@ sealed trait LockTimeScriptSignature extends ScriptSignature {
...
 
 sealed trait CLTVScriptSignature extends LockTimeScriptSignature {
-  override def toString = "CLTVScriptSignature(" + hex + ")"
+  override def toString: String = s"CLTVScriptSignature($hex)"
 }
 
-object CLTVScriptSignature extends Factory[CLTVScriptSignature] {
-  private case class CLTVScriptSignatureImpl(bytes: ByteVector) extends CLTVScriptSignature
+/**
+ * Note that extend [[org.bitcoins.core.protocol.script.ScriptFactory]] here
```
Christewart: Add this comment to `CSVScriptSignature` as well

PR #253, opened 2018-11-25
fix new compiler warnings for scalac 2.12.x on ln

Christewart:
Fixes new compiler warnings that were introduced with bumping the scala compiler version

### DIFF
--- a/core-gen/src/test/scala/org/bitcoins/core/gen/CreditingTxGen.scala
+++ b/core-gen/src/test/scala/org/bitcoins/core/gen/CreditingTxGen.scala
@@ -1,12 +1,12 @@
 package org.bitcoins.core.gen
 
 import org.bitcoins.core.crypto.Sign
+import org.bitcoins.core.number.UInt32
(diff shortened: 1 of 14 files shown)


core-gen/src/test/scala/org/bitcoins/core/gen/CreditingTxGen.scala
```diff
@@ -20,7 +20,7 @@ sealed abstract class CreditingTxGen {
   }
 
   def rawOutput: Gen[BitcoinUTXOSpendingInfo] = {
-    Gen.oneOf(p2pkOutput, p2pkhOutput, multiSigOutput, /*cltvOutput,*/ csvOutput, p2wpkhOutput)
+    Gen.oneOf(p2pkOutput, p2pkhOutput, multiSigOutput, /*cltvOutput,*/ csvOutput)
```
Christewart: this is wrong and needs to fixed to add `p2wpkhOutput` back

PR #256, opened 2018-12-01
Lightning Network

Christewart:
This pull request merges lightning specific application data structures into bitcoin-s

This PR contains the following

- A [BOLT11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md) compliant [Lightning network invoice](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala) implementation 

- An [eclair rpc client](https://github.com/bitcoin-s/bitcoin-s-core/tree/ln/eclair-rpc)

- Basic lightning data structures ([`ChannelId`](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/channel/ChannelId.scala),  [`NodeId`](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/node/NodeId.scala), [`MilliSatoshis`](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala), [`ShortChannelId`](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala) etc)

- [Parameters](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/protocol/ln/LnParams.scala) for the bitcoin lightning network

- Reworking of [Bech32 code](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/util/Bech32.scala) including adding a [`UInt5`](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/core/src/main/scala/org/bitcoins/core/number/NumberType.scala#L95) to make it easier to distinguish bech32 data
## Bitcoin-s testkit

This PR also creates the [`bitcoin-s-testkit`](https://github.com/bitcoin-s/bitcoin-s-core/tree/ln/testkit). The goal of this testkit is to create a jar that is for application developers to leverage for testing their bitcoin/lightning applications. We are using it internally at suredbits and thought it would be useful for other people to use. The goal is to provide useful utilties similar to how the [akka-testkit](https://doc.akka.io/docs/akka/current/stream/stream-testkit.html#testkit) and [slick-testkit](http://slick.lightbend.com/doc/3.2.1/testkit.html) do for their respective projects. 

As it currently stands, you can arbitrarily create/destroy bitcoind and eclair instances with the testkit. We use this functionality in the `rpc` and `eclair-rpc` project to test bitcoind and eclair respectively. 

[bitcoind's testkit example](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/rpc/src/test/scala/org/bitcoins/rpc/BitcoindRpcClientTest.scala#L31-L40)

[eclair's teskit  example](https://github.com/bitcoin-s/bitcoin-s-core/blob/ln/eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala#L30)

## Thanks

Thanks to @TannerR1776 and @nkohen for their contributions on the `ln` branch!


core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
```diff
@@ -0,0 +1,128 @@
...
+    paymentHash = paymentHash,
+    descriptionOrHash = descOrHashTag,
+    expiryTime = expiryTime,
+    cltvExpiry = cltvExpiry,
+    fallbackAddress = fallbackAddress,
+    routingInfo = routes)
+
+  def lnInvoiceSignature: Gen[LnInvoiceSignature] = for {
+    sig <- CryptoGenerators.digitalSignature
+    version <- Gen.choose(0, 3).map(UInt8(_))
```
Christewart: Consider breaking this out into it's own generator

core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnInvoiceGen.scala
```diff
@@ -0,0 +1,128 @@
...
+  } yield LnInvoiceSignature(version, sig)
+
+  def lnInvoice: Gen[LnInvoice] = for {
+    hrp <- lnHrp
+    //timestamp is 35 bits according to BOLT11
+    timestamp <- Gen.choose(0, NumberUtil.pow2(35).toLong).map(UInt64(_))
```
Christewart: private helper method?

core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnRouteGen.scala
```diff
@@ -0,0 +1,46 @@
...
+trait LnRouteGen {
+  def shortChannelId: Gen[ShortChannelId] = {
+    NumberGenerator.uInt64s.map { u64 =>
+      ShortChannelId(u64)
+    }
+  }
+
+  def feeBaseMSat: Gen[FeeBaseMSat] = for {
+    //note that the feebase msat is only 4 bytes
+    p <- Gen.choose(0, UInt32.max.toLong).map(MilliSatoshis(_))
```
Christewart: Is there no `uint32` generator? If so we should use that...

core-gen/src/test/scala/org/bitcoins/core/gen/ln/LnRouteGen.scala
```diff
@@ -0,0 +1,46 @@
...
+  def feeProportionalMillionths: Gen[FeeProportionalMillionths] = for {
+    fee <- NumberGenerator.uInt32s
+  } yield FeeProportionalMillionths(fee)
+
+  def route: Gen[LnRoute] = for {
+    pubKey <- CryptoGenerators.publicKey
+    id <- shortChannelId
+    baseFee <- feeBaseMSat
+    feeProp <- feeProportionalMillionths
+    cltvExpiryDelta <- Gen.chooseNum[Short](0, Short.MaxValue)
```
Christewart: This smells, if it needs to be this should probably create a gen in `NumberGenerator` for `Short`

core-test/src/test/scala/org/bitcoins/core/protocol/blockchain/BlockTest.scala
```diff
@@ -37,13 +32,13 @@ class BlockTest extends FlatSpec with MustMatchers {
     val fileName = "/00000000000000000008513c860373da0484f065983aeb063ebf81c172e81d48.txt"
     val lines = Source.fromURL(getClass.getResource(fileName)).mkString
     val time = timeBlockParsing(Block.fromHex(lines))
-    assert(time <= 15000)
+    assert(time <= 20000)
```
Christewart: Revert these changes since optimization improvements were integrated into master

core-test/src/test/scala/org/bitcoins/core/protocol/ln/LnInvoiceSpec.scala
```diff
@@ -0,0 +1,5 @@
+package org.bitcoins.core.protocol.ln
+
+class LnInvoiceSpec {
```
Christewart: remove

core-test/src/test/scala/org/bitcoins/core/protocol/ln/LnInvoiceTagsTest.scala
```diff
@@ -0,0 +1,159 @@
...
+      paymentHash = paymentTag,
+      descriptionOrHash = descriptionHashTagE)
+
+    lnTags.toString must be(expected)
+
+  }
+
+  it must "serialize and deserialize the example 5 tags" in {
+
+    //weird ordering for serialization... ignore this test case for now
```
Christewart: I think this can be deleted. Unfortunately it doesn't seem like there is a cannonical ordering of tags described in BOLT11. You can place them in any order. So when we go to re-serialize the invoice it will not serialize to the "correct" string. It will have the tags in a different order -- which leads to a different checksum and digital signature.

core-test/src/test/scala/org/bitcoins/core/protocol/ln/LnInvoiceTagsTest.scala
```diff
@@ -0,0 +1,159 @@
...
+import org.bitcoins.core.protocol.ln.currency.MilliSatoshis
+import org.bitcoins.core.protocol.ln.fee.{ FeeBaseMSat, FeeProportionalMillionths }
+import org.bitcoins.core.protocol.ln.routing.LnRoute
+import org.bitcoins.core.protocol.ln.util.LnUtil
+import org.bitcoins.core.util.{ Bech32, CryptoUtil }
+import org.scalatest.{ FlatSpec, MustMatchers }
+import org.slf4j.LoggerFactory
+import scodec.bits.ByteVector
+
+class LnTagsTest extends FlatSpec with MustMatchers {
```
Christewart: I think we should just delete these tests, they are entirely a subset of the unit tests in `LnInvoiceUnitTest`

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    MilliSatoshis.fromPico(PicoBitcoins(20)) must be(MilliSatoshis(2))
+
+    MilliSatoshis.fromPico(PicoBitcoins(101)) must be(MilliSatoshis(10))
+
+    MilliSatoshis.fromPico(PicoBitcoins(110)) must be(MilliSatoshis(11))
+  }
+
+  it must "covert from a ln currency unit -> millisatoshis -> lnCurrencyUnit" in {
+    def gen(): Gen[Long] = {
```
Christewart: Why is this gen here?

core/src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
```diff
@@ -36,10 +36,29 @@ sealed abstract class ECDigitalSignature {
...
+
+  /**
+   * Creates a ByteVector with only
+   * the 32byte r value and 32 byte s value
+   * in the vector
+   */
+  def toRawRS: ByteVector = {
+
+    //comeback and look at this, i think this will fail if r/s is a small bigint
+    val rBytes = r.toByteArray.takeRight(32)
```
Christewart, line 51: ping

core/src/main/scala/org/bitcoins/core/crypto/ECKey.scala
```diff
@@ -286,6 +286,10 @@ sealed abstract class ECPublicKey extends BaseECKey {
       ECPublicKey.fromBytes(ByteVector(decompressed))
     } else this
   }
+
+  def toPoint: ECPoint = {
```
Christewart, line 295: scaladoc

core/src/main/scala/org/bitcoins/core/crypto/ECKey.scala
```diff
@@ -320,4 +324,9 @@ object ECPublicKey extends Factory[ECPublicKey] {
    * [[https://github.com/bitcoin/bitcoin/blob/27765b6403cece54320374b37afb01a0cfe571c3/src/pubkey.h#L158]]
    */
   def isValid(bytes: ByteVector): Boolean = bytes.nonEmpty
+
+  def fromPoint(p: ECPoint, isCompressed: Boolean = true): ECPublicKey = {
```
Christewart, line 340: scaladoc

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -92,12 +92,29 @@ sealed abstract class SignedNumber[T <: Number[T]] extends Number[T]
  */
 sealed abstract class UnsignedNumber[T <: Number[T]] extends Number[T]
 
+sealed abstract class UInt5 extends UnsignedNumber[UInt5] {
```
Christewart, line 100: Scaladoc for information on what this is useful for -- if you are unaware of bech32 this probably doesn't make sense to have a `UInt5`

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -65,155 +65,104 @@ sealed abstract class Bech32Address extends BitcoinAddress {
...
 
   override def toString = "Bech32Address(" + value + ")"
 
 }
 
 object Bech32Address extends AddressFactory[Bech32Address] {
-  private case class Bech32AddressImpl(hrp: HumanReadablePart, data: Seq[UInt8]) extends Bech32Address {
-    verifyChecksum(hrp, UInt8.toBytes(data))
+  private case class Bech32AddressImpl(hrp: HumanReadablePart, data: Vector[UInt5]) extends Bech32Address {
+    //require(verifyChecksum(hrp, data), "checksum did not pass")
```
Christewart, line 103: Should this be uncommented?

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -65,155 +65,104 @@ sealed abstract class Bech32Address extends BitcoinAddress {
...
-    val encoded = Bech32Address.encode(prog)
+    val encoded = Bech32.from8bitTo5bit(prog)
     val hrp = networkParameters match {
       case _: MainNet => bc
       case _: TestNet3 | _: RegTest => tb
     }
-    val witVersion = witSPK.witnessVersion.version.toLong.toShort
-    encoded.map(e => Bech32Address(hrp, Seq(UInt8(witVersion)) ++ e))
+    val witVersion = witSPK.witnessVersion.version.toInt.toByte
+    Try(Bech32Address(hrp, Vector(UInt5(witVersion)) ++ encoded))
```
Christewart: Think about just returning `Bech32Address` here rather than `Try[Bech32Address]`?

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -65,155 +65,104 @@ sealed abstract class Bech32Address extends BitcoinAddress {
...
+      case bech32Addr =>
+        val bytes = bech32Addr.data
+        val (v, _) = (bytes.head, bytes.tail)
+        val convertedProg = NumberUtil.convertUInt5sToUInt8(bytes.tail)
+        val progBytes = UInt8.toBytes(convertedProg)
+        val witVersion = WitnessVersion(v.toInt)
+        val pushOp = BitcoinScriptUtil.calculatePushOp(progBytes)
+        witVersion match {
+          case Some(v) =>
+            WitnessScriptPubKey(List(v.version) ++ pushOp ++ List(ScriptConstant(progBytes))) match {
```
Christewart: intermediate val, this is nasty

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,138 @@
...
+   * [[https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Specification]]
+   * @param input
+   * @return
+   */
+  def fromString(input: String): Try[LnHumanReadablePart] = {
+    val hrpIsValidT = Bech32.checkHrpValidity(input, parse)
+    hrpIsValidT
+  }
+
+  private def parse(input: String): Try[LnHumanReadablePart] = {
```
Christewart, line 109: This is rather nasty, we should look for a cleaner alternative

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,294 @@
...
+import org.bitcoins.core.protocol.ln.node.NodeId
+import org.bitcoins.core.protocol.ln.util.LnUtil
+import org.bitcoins.core.util._
+import scodec.bits.ByteVector
+
+import scala.util.{ Failure, Success, Try }
+
+sealed abstract class LnInvoice {
+  require(
+    timestamp < UInt64(NumberUtil.pow2(35)),
```
Christewart: consider creating 

```scala

object LnInvoice { 

val MAX_TIMESTAMP = UInt64(NumberUtil.pow2(35))

}
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,294 @@
...
+
+  def fromString(bech32String: String): Try[LnInvoice] = {
+    val sepIndexes = {
+      bech32String.zipWithIndex.filter(_._1 == Bech32.separator)
+    }
+    if (sepIndexes.isEmpty) {
+      Failure(new IllegalArgumentException("LnInvoice did not have the correct separator"))
+    } else {
+      val sepIndex = sepIndexes.last._2
+      val (hrp, data) = (bech32String.take(sepIndex), bech32String.splitAt(sepIndex + 1)._2)
```
Christewart:
```scala
val hrp = bech32String.take(sepIndex)
val data = bech32String.splitAt(sepIndex + 1)._2
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,294 @@
...
+      if (hrp.size < 1 || data.size < 6) {
+        Failure(new IllegalArgumentException("Hrp/data too short"))
+      } else {
+
+        val hrpValid = LnHumanReadablePart.fromString(hrp)
+
+        //is checksum returned here?
```
Christewart: no, remove comment

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,294 @@
...
+        val dataValid = Bech32.checkDataValidity(data)
+
+        val isChecksumValid: Try[Vector[UInt5]] = hrpValid.flatMap { h: LnHumanReadablePart =>
+          dataValid.flatMap { d: Vector[UInt5] =>
+            if (verifyChecksum(h, d)) {
```
Christewart: Probably should break this inner lambda into a helper method

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,294 @@
...
+    val lnInvoiceSignature = buildLnInvoiceSignature(hrp, timestamp, lnTags, privateKey)
+
+    LnInvoice(
+      hrp = hrp,
+      timestamp = timestamp,
+      lnTags = lnTags,
+      signature = lnInvoiceSignature)
+  }
+
+  def uInt64ToBase32(input: UInt64): Vector[UInt5] = {
```
Christewart: private?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnPolicy.scala
```diff
@@ -0,0 +1,31 @@
...
+
+sealed abstract class LnPolicy {
+
+  /**
+   * The "amount_msat" field has been artificially limited to a UInt32. This means that the current maximum transaction that can be completed
+   * over the lightning network is 4294967295 MilliSatoshi.
+   * This is a self imposed limit, and is subject to change.
+   * Please see BOLT #2: https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#adding-an-htlc-update_add_htlc
+   */
+  lazy val maxAmountMSat: MilliBitcoins = MilliBitcoins(4294967295L)
```
Christewart: We've added a new `MilliSatoshis` type since this code was created, this should be typed as `lazy val maxAmountMSat: MilliSatoshis = ...`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnPolicy.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  val milliMultiplier: BigDecimal = BigDecimal(0.001)
+  val microMultiplier: BigDecimal = BigDecimal(0.000001)
+  val nanoMultiplier: BigDecimal = BigDecimal(0.000000001)
+  val picoMultiplier: BigDecimal = BigDecimal(0.000000000001)
+
+  val maxPicoBitcoins: BigInt = Int64.max.toBigInt
+  val minPicoBitcoins: BigInt = Int64.min.toBigInt
+  val maxMilliBitcoins: BigInt = maxPicoBitcoins / (milliMultiplier / picoMultiplier).toBigIntExact().get
```
Christewart: Are these used? If not, I think we should just get rid of this as it is confusing

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTagPrefix.scala
```diff
@@ -0,0 +1,75 @@
+package org.bitcoins.core.protocol.ln
+
+import org.bitcoins.core.number.UInt5
+import org.bitcoins.core.util.Bech32
+
+sealed abstract class LnTagPrefix {
+  def value: Char
```
Christewart: I think this can be changed to `val value: Char` -- in all of the implementations we should definitely be using `val`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,199 @@
...
+  private case class InvoiceTagImpl(
+    paymentHash: LnTag.PaymentHashTag,
+    description: Option[LnTag.DescriptionTag],
+    nodeId: Option[LnTag.NodeIdTag],
+    descriptionHash: Option[LnTag.DescriptionHashTag],
+    expiryTime: Option[LnTag.ExpiryTimeTag],
+    cltvExpiry: Option[LnTag.MinFinalCltvExpiry],
+    fallbackAddress: Option[LnTag.FallbackAddressTag],
+    routingInfo: Option[LnTag.RoutingInfo]) extends LnTaggedFields {
+    require(
```
Christewart: Move this invariant into the `sealed abstract class `LnInvoiceTag`

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,199 @@
...
+          throw new IllegalArgumentException("Failed to parse LnTaggedFields, needs 15bits of meta data to be able to parse")
+        case Nil =>
+          fields
+      }
+    }
+
+    val tags = loop(u5s.toList, Vector.empty)
+
+    val paymentHashTag = tags.find(_.isInstanceOf[LnTag.PaymentHashTag])
+      .get.asInstanceOf[LnTag.PaymentHashTag]
```
Christewart: `PaymentHashTag` is required, but we should probably throw a nice exception or return a `Try` if it is not found

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * All of the different invoice tags that are currently defined
+ * Refer to BOLT11 for a full list
+ * [[https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields]]
+ */
+object LnTag {
+
+  /** Fallback address versions */
+  object FallbackAddressV {
+    val p2pkh = UInt8(17)
```
Christewart: ADT?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -0,0 +1,279 @@
...
+          val witSPK = {
+            if (bytes.size == 32) {
+              val hash = Sha256Digest.fromBytes(bytes)
+              P2WSHWitnessSPKV0.fromHash(hash)
+            } else if (bytes.size == 20) {
+
+              val hash = Sha256Hash160Digest.fromBytes(bytes)
+              P2WPKHWitnessSPKV0.fromHash(hash)
+            } else {
+              throw new IllegalArgumentException(s"Can only create witness spk out of a 32 byte or 20 byte hash")
```
Christewart: Add what we got in this error message

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,238 @@
...
+    PicoBitcoins(toPicoBitcoinValue * ln.toPicoBitcoinValue)
+  }
+
+  def unary_- : LnCurrencyUnit = {
+    PicoBitcoins(-toPicoBitcoinValue)
+  }
+
+  override def bytes: ByteVector = Int64(toPicoBitcoinValue).bytes.reverse
+
+  def fiveBitEncoding: Vector[UInt5] = {
```
Christewart: `toUInt5s`?

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  def toPicoBitcoins: PicoBitcoins = PicoBitcoins(toPicoBitcoinValue)
+
+  def encodedBytes: ByteVector = {
+    ByteVector(toEncodedString.map(_.toByte))
+  }
+
+  def toMSat: MilliSatoshis = MilliSatoshis.fromPico(toPicoBitcoins)
+
+  def toEncodedString: String = {
```
Christewart, line 93: Add scaladoc saying this is defined in BOLT11, and looks something like `100p` for 100 picobitcoins

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,238 @@
...
+    val currency = input.splitAt(input.length - 1)
+    val amount = Try(BigInt(currency._1))
+    if (amount.isSuccess) {
+      val unit = currency._2
+      unit match {
+        case "m" => Try(MilliBitcoins(amount.get))
+        case "u" => Try(MicroBitcoins(amount.get))
+        case "n" => Try(NanoBitcoins(amount.get))
+        case "p" => Try(PicoBitcoins(amount.get))
+        case _ => Failure(new IllegalArgumentException(s"LnCurrencyUnit not found. Expected MilliBitcoins (m), MicroBitcoins (u), NanoBitcoins (n), or PicoBitcoins (p), got: $unit"))
```
Christewart:
```
case _ : String => ...
```

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,238 @@
...
+      }
+    } else { Failure(new IllegalArgumentException(s"Could not convert amount to valid number, got: $amount")) }
```
Christewart: new lin

core/src/main/scala/org/bitcoins/core/protocol/ln/fee/FeeBaseMSat.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Represents the fee we charge for forwarding an HTLC on the Lightning network
+ * This is used in the ChannelUpdate and Routing information a [[org.bitcoins.core.protocol.ln.LnInvoice]]
+ * [[https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#the-channel_update-message]]
+ * [[https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#on-mainnet-with-fallback-address-1rustyrx2oai4eyydpqgwvel62bbgqn9t-with-extra-routing-info-to-go-via-nodes-029e03a901b85534ff1e92c43c74431f7ce72046060fcf7a95c37e148f78c77255-then-039e03a901b85534ff1e92c43c74431f7ce72046060fcf7a95c37e148f78c77255]]
+ */
+case class FeeBaseMSat(msat: MilliSatoshis) extends NetworkElement {
+  require(msat.toLong <= UInt32.max.toLong, s"Value too large for FeeBaseMSat ${msat}")
+  require(msat.toLong >= 0, s"Value cannot be negative for FeeBaseMSat ${msat}")
```
Christewart: redundant, already required by `MilliSatoshis`

core/src/main/scala/org/bitcoins/core/protocol/ln/routing/LnRoute.scala
```diff
@@ -0,0 +1,65 @@
...
+
+object LnRoute {
+
+  def fromBytes(bytes: ByteVector): LnRoute = {
+
+    val pubKey = ECPublicKey.fromBytes(bytes.take(33))
+
+    val shortChannelId = ShortChannelId.fromBytes(bytes.slice(33, 41))
+
+    //type hacking here, fee base is a uint32, but PicoBitcoins needs a i64
```
Christewart: This comment is out of date since we are using `MilliSatoshis` now

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -0,0 +1,219 @@
...
+      case Nil => Success(accum.reverse)
+      case h :: t =>
+        if (!Bech32.charset.contains(h.toLower)) {
+          Failure(new IllegalArgumentException("Invalid character in data of bech32 address, got: " + h))
+        } else {
+          if ((h.isUpper && hasLower) || (h.isLower && hasUpper)) {
+            Failure(new IllegalArgumentException("Cannot have mixed case for bech32 address"))
+          } else {
+            val byte = Bech32.charset.indexOf(h.toLower).toByte
+            require(byte >= 0 && byte < 32, "Not in valid range, got: " + byte)
```
Christewart: This shouldn't be just a raw `require`, should be returned as a `Failure(...)`

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -60,33 +60,35 @@ trait NumberUtil extends BitcoinSLogger {
   def toLong(hex: String): Long = toLong(BitcoinSUtil.decodeHex(hex))
 
   /** Converts a sequence uint8 'from' base to 'to' base */
-  def convertUInt8s(data: Seq[UInt8], from: UInt32, to: UInt32, pad: Boolean): Try[Seq[UInt8]] = {
+  def convert[To <: Number[To]](data: Vector[UInt8], from: UInt32, to: UInt32, pad: Boolean, f: UInt8 => To): Try[Vector[To]] = {
     var acc: UInt32 = UInt32.zero
     var bits: UInt32 = UInt32.zero
-    var ret: Seq[UInt8] = Nil
+    var ret: Vector[To] = Vector.empty
```
Christewart: Should really be using a `Vector.newBuilder[To]` here..

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.0.1-SNAPSHOT"
+version in ThisBuild := "0.0.5-LN-SNAPSHOT"
```
Christewart: reset

Christewart (commented):
This is only a partial review, mainly of the new lightning related stuff. This does not look too closely at the eclair rpc stuff or the testkit stuff

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
Christewart: Filed an issue for this on #257

core/src/main/scala/org/bitcoins/core/protocol/ln/LnPolicy.scala
Christewart: Unfortunately they are used

core-test/src/test/scala/org/bitcoins/core/number/UInt5Test.scala
```diff
@@ -0,0 +1,53 @@
...
+    UInt5(31.toByte) must be(UInt5.max)
+  }
+
+  it must "not allow negative numbers" in {
+    intercept[IllegalArgumentException] {
+      UInt5(-1)
+    }
+  }
+
+  it must "not allow numbers more than 32" in {
```
torkelrogstad: More than 31?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -92,12 +92,34 @@ sealed abstract class SignedNumber[T <: Number[T]] extends Number[T]
...
+
 sealed abstract class UInt8 extends UnsignedNumber[UInt8] {
   override def apply: A => UInt8 = UInt8(_)
 
   override def hex = BitcoinSUtil.encodeHex(toInt.toShort).slice(2, 4)
 
   override def andMask = 0xff
+
+  def toUInt5: UInt5 = {
+    //this will throw if not in range of a UInt5, come back and look later
```
torkelrogstad, line 120: What's meant by this comment?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -166,6 +188,47 @@ trait BaseNumbers[T] {
...
+  lazy val min = zero
+  lazy val max = UInt5(31.toByte)
+
+  def apply(byte: Byte): UInt5 = fromByte(byte)
+
+  def apply(bigInt: BigInt): UInt5 = {
+
+    require(
+      bigInt.toByteArray.size == 1,
+      s"To create a uint5 from a ByteVector it must be of size one ${bigInt}")
```
torkelrogstad:
```suggestion
      s"To create a uint5 from a BigInt it must be less than 32. Got: ${bigInt.toString}")
```

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -166,6 +188,47 @@ trait BaseNumbers[T] {
...
+
+    UInt5.fromByte(bigInt.toByteArray.head)
+  }
+
+  override def fromBytes(bytes: ByteVector): UInt5 = {
+    require(bytes.size == 1, s"To create a uint5 from a ByteVector it must be of size one ${bytes.length}")
```
torkelrogstad, line 215:
```suggestion
require(bytes.size == 1,
      s"To create a uint5 from a ByteVector it must be of size one. Got a ByteVector of length ${bytes.length}")
```

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -200,8 +263,8 @@ object UInt8 extends Factory[UInt8] with BaseNumbers[UInt8] {
     ByteVector(us.map(toByte(_)))
   }
 
-  def toUInt8s(bytes: ByteVector): Seq[UInt8] = {
-    bytes.toSeq.map { b: Byte => toUInt8(b) }
+  def toUInt8s(bytes: ByteVector): Vector[UInt8] = {
+    bytes.toArray.map(toUInt8(_)).toVector
```
torkelrogstad:
```suggestion
    bytes.toArray.map(toUInt8).toVector
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.bitcoins.core.util.Bech32
+import scodec.bits.ByteVector
+
+import scala.util.matching.Regex
+import scala.util.{ Failure, Success, Try }
+
+sealed abstract class LnHumanReadablePart {
+  require(
+    amount.isEmpty || amount.get.toBigInt > 0,
+    s"Invoice amount must be greater then 0, got $amount")
```
torkelrogstad: This might not matter at all, feel free to disregard:

The wording here implies that an amount must be specified when creating an invoice. Would it be better to say "cannot be negative"?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,138 @@
...
+  require(
+    amount.isEmpty || amount.get.toMSat <= LnPolicy.maxAmountMSat,
+    s"Invoice amount must be less than ${LnPolicy.maxAmountMSat}, got ${amount.get.toMSat}")
```
torkelrogstad: Same as above.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,138 @@
...
+  def bytes: ByteVector = {
+    network.invoicePrefix ++ amount.map(_.encodedBytes).getOrElse(ByteVector.empty)
+  }
+
+  override def toString: String = {
+    val b = StringBuilder.newBuilder
+    val prefix = network.invoicePrefix.toArray.map(_.toChar).mkString
+    b.append(prefix)
+
+    val amt = amount.map(_.toEncodedString).getOrElse("")
```
torkelrogstad, line 35: Use the `StringBuilder` here as well?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /** Prefix for genearting a LN invoice on the Bitcoin RegTest */
+  case class lnbcrt(override val amount: Option[LnCurrencyUnit]) extends LnHumanReadablePart {
+    def network: LnParams = LnBitcoinRegTest
+  }
+
+  def apply(network: NetworkParameters): Option[LnHumanReadablePart] = {
+    val lnNetworkOpt = LnParams.fromNetworkParameters(network)
+    lnNetworkOpt.map(LnHumanReadablePart.fromLnParams(_))
```
torkelrogstad:
```suggestion
    lnNetworkOpt.map(LnHumanReadablePart.fromLnParams)
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,319 @@
...
+import org.bitcoins.core.util._
+import scodec.bits.ByteVector
+
+import scala.util.{Failure, Success, Try}
+
+sealed abstract class LnInvoice {
+
+  require(
+    timestamp < LnInvoice.MAX_TIMESTAMP_U64,
+    s"timestamp ${timestamp.toBigInt} < ${LnInvoice.MAX_TIMESTAMP}")
```
torkelrogstad, line 17:
```suggestion
    s"Timestamp must be less than ${LnInvoice.MAX_TIMESTAMP}. Got: ${timestamp.toBigInt} ")
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,319 @@
...
+  def amountPicoBitcoins: Option[PicoBitcoins] = {
+    amount.map(_.toPicoBitcoins)
+  }
+
+  /**
+   * The [[NodeId]] that we are paying this invoice too
+   * We can either recover this with public key recovery from
+   * the [[LnInvoiceSignature]] or if [[LnTag.NodeIdTag]] is
+   * defined we MUST use that NodeId.
+   * [[https://github.com/lightningnetwork/lihtning-rfc/blob/master/11-payment-encoding.md#requirements-3]]
```
torkelrogstad:
```suggestion
   * [[https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#requirements-3]]
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,319 @@
...
+    u5s
+  }
+
+  def verifyChecksum(hrp: LnHumanReadablePart, u5s: Seq[UInt5]): Boolean = {
+    val data = hrpExpand(hrp) ++ u5s
+    val checksum = Bech32.polyMod(data)
+    checksum == 1
+  }
+
+  def apply(hrp: LnHumanReadablePart, data: Vector[UInt5]): LnInvoice = {
```
torkelrogstad, line 144: This is currently going to throw very generic and hard to interpret errors about index mismatches etc, if too little data is provided. I'm not sure what the best way to handle that is, but one solution is to at least do some basic sanity testing at the beginning of function, to check for length of `data`.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -0,0 +1,319 @@
...
+    } else {
+      val sepIndex = sepIndexes.last._2
+
+      val hrp = bech32String.take(sepIndex)
+
+      val data = bech32String.splitAt(sepIndex + 1)._2
+
+      if (hrp.size < 1 || data.size < 6) {
+        Failure(new IllegalArgumentException("Hrp/data too short"))
+      } else {
```
torkelrogstad, line 194:
```scala
    if (hrp.length < 1) {
        Failure(new IllegalArgumentException("HumanReadablePart is too short"))
      } else if (data.length < 6) {
        Failure(new IllegalArgumentException("Data part is too short"))
      } else {
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
Christewart: It is used on the next line

torkelrogstad: I was thinking line 30-31 could be replaced by
```scala
network.invoicePrefix.toArray.map(_.toChar).foreach(b.append)
```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnParams.scala
```diff
@@ -0,0 +1,88 @@
...
+    override def lnPort = 9735
+
+    override val invoicePrefix: ByteVector = {
+      ByteVector('l', 'n', 'b', 'c', 'r', 't')
+    }
+  }
+
+  private val allNetworks: Vector[LnParams] = Vector(LnBitcoinMainNet, LnBitcoinTestNet, LnBitcoinRegTest)
+
+  def fromNetworkParameters(np: NetworkParameters): Option[LnParams] = {
```
torkelrogstad: `NetworkParameters` is a sealed abstract class. Could we replace the `.find` with a switch and return `LnParams` instead of `Option[LnParams]`?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTagPrefix.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  case object FallbackAddress extends LnTagPrefix {
+    override val value: Char = 'f'
+  }
+
+  case object RoutingInfo extends LnTagPrefix {
+    override val value: Char = 'r'
+  }
+
+  private val all: List[LnTagPrefix] = List(
```
torkelrogstad: Would `Map[String, LnTagPrefix]` be more performant?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTagPrefix.scala
```diff
@@ -0,0 +1,75 @@
...
+    PaymentHash, Description, NodeId,
+    DescriptionHash, ExpiryTime, CltvExpiry,
+    FallbackAddress, RoutingInfo)
+
+  def fromString(str: String): Option[LnTagPrefix] = {
```
torkelrogstad, line 62: Add `fromChar` method

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  def expiryTime: Option[LnTag.ExpiryTimeTag]
+
+  def cltvExpiry: Option[LnTag.MinFinalCltvExpiry]
+
+  def fallbackAddress: Option[LnTag.FallbackAddressTag]
+
+  def routingInfo: Option[LnTag.RoutingInfo]
+
+  def data: Vector[UInt5] = {
```
torkelrogstad: `lazy val` with less repetition

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,207 @@
...
+  def apply(
+    paymentHash: LnTag.PaymentHashTag,
+    descriptionOrHash: Either[LnTag.DescriptionTag, LnTag.DescriptionHashTag],
+    nodeId: Option[LnTag.NodeIdTag] = None,
+    expiryTime: Option[LnTag.ExpiryTimeTag] = None,
+    cltvExpiry: Option[LnTag.MinFinalCltvExpiry] = None,
+    fallbackAddress: Option[LnTag.FallbackAddressTag] = None,
+    routingInfo: Option[LnTag.RoutingInfo] = None): LnTaggedFields = {
+
+    if (descriptionOrHash.isLeft) {
```
torkelrogstad: Most of the lines here are redundant

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,207 @@
...
+          //t is the actual possible payload
+          val payload: Vector[UInt5] = t.take(dataLength.toInt).toVector
+
+          val tag = LnTag.fromLnTagPrefix(prefix.get, payload)
+
+          val newRemaining = t.slice(payload.size, t.size)
+
+          loop(newRemaining, fields.:+(tag))
+
+        case _ :: _ | _ :: _ :: _ =>
```
torkelrogstad: I believe this would also work
```scala
case Nil =>
    fields
case _ =>
    throw new IllegalArgumentException("Failed to parse LnTaggedFields, needs 15bits of meta data to be able to parse")
```
I'd say this makes it clearer what the base case (`Nil`) is, and what an illegal state (everything else) is

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,207 @@
...
+      val fOpt = tags.find(_.isInstanceOf[LnTag.PaymentHashTag])
+      if (fOpt.isDefined) {
+        fOpt.get.asInstanceOf[LnTag.PaymentHashTag]
+      } else {
+        throw new IllegalArgumentException(s"Payment hash must be defined in a LnInvoice")
+      }
+
+    }
+
+    val description = tags.find(_.isInstanceOf[LnTag.DescriptionTag])
```
torkelrogstad: helper method?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,207 @@
...
+    val cltvExpiry = tags.find(_.isInstanceOf[LnTag.MinFinalCltvExpiry])
+      .map(_.asInstanceOf[LnTag.MinFinalCltvExpiry])
+
+    val fallbackAddress = tags.find(_.isInstanceOf[LnTag.FallbackAddressTag])
+      .map(_.asInstanceOf[LnTag.FallbackAddressTag])
+
+    val routingInfo = tags.find(_.isInstanceOf[LnTag.RoutingInfo])
+      .map(_.asInstanceOf[LnTag.RoutingInfo])
+
+    val d: Either[LnTag.DescriptionTag, LnTag.DescriptionHashTag] = {
```
torkelrogstad, line 184: The same check is performed in a `require` clause in the abstract class. Would it be an idea to only do this in one place?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -0,0 +1,298 @@
...
+
+    case object P2SH extends FallbackAddressV {
+      val u8: UInt8 = UInt8(18)
+    }
+
+    case object WitSPK extends FallbackAddressV {
+      val u8 = UInt8.zero
+    }
+
+    def fromU8(version: UInt8, bytes: ByteVector, np: NetworkParameters): FallbackAddressTag = {
```
torkelrogstad, line 92: refactor into `match`

core/src/main/scala/org/bitcoins/core/protocol/ln/channel/ChannelState.scala
```diff
@@ -0,0 +1,40 @@
...
+  private val all = List(
+    WAIT_FOR_INIT_INTERNAL, WAIT_FOR_OPEN_CHANNEL, WAIT_FOR_ACCEPT_CHANNEL,
+    WAIT_FOR_FUNDING_INTERNAL, WAIT_FOR_FUNDING_CREATED, WAIT_FOR_FUNDING_SIGNED,
+    WAIT_FOR_FUNDING_CONFIRMED, WAIT_FOR_FUNDING_LOCKED, NORMAL,
+    SHUTDOWN, NEGOTIATING, CLOSING,
+    CLOSED, OFFLINE, SYNCING,
+    WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT, ERR_FUNDING_LOST, ERR_FUNDING_TIMEOUT,
+    ERR_INFORMATION_LEAK)
+
+  def fromString(str: String): Option[ChannelState] = {
```
torkelrogstad, line 51: `Map[String, ChannelState]`?

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+import org.bitcoins.core.number.{ BaseNumbers, Int64, UInt5 }
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.protocol.ln._
+import org.bitcoins.core.util.Bech32
+import scodec.bits.ByteVector
+
+import scala.util.{ Failure, Try }
+
+sealed abstract class LnCurrencyUnit extends NetworkElement {
+  type A
```
torkelrogstad: Is there any reason for this type alias being used? IMO it hinders readability, especially when hovering over symbols in an IDE.

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+    u5s
+  }
+
+  def toBigInt: BigInt
+
+  def toLong: Long = toBigInt.bigInteger.longValueExact()
+
+  def toInt: Int = toBigInt.bigInteger.intValueExact()
+
+  protected def underlying: A
```
torkelrogstad: Same as above comment on type alias.

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+    * For instance, 100 [[PicoBitcoins]] would appear as "100p"
+    * @return
+    */
+  def toEncodedString: String = {
+    toBigInt + character.toString()
+  }
+}
+
+sealed abstract class MilliBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
```
torkelrogstad: Above

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+}
+
+sealed abstract class MicroBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
+
+  override def character: Char = 'u'
+
+  override def toPicoBitcoinMultiplier: Int = 1000000
+
+  override def toBigInt: A = underlying
```
torkelrogstad: Above

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+  def apply(underlying: BigInt): MicroBitcoins = MicroBitcoinsImpl(underlying)
+
+  private case class MicroBitcoinsImpl(underlying: BigInt) extends MicroBitcoins {
+    require(underlying >= LnPolicy.minMicroBitcoins, "Number was too small for MicroBitcoins, got: " + underlying)
+    require(underlying <= LnPolicy.maxMicroBitcoins, "Number was too big for MicroBitcoins, got: " + underlying)
+  }
+}
+
+sealed abstract class NanoBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
```
torkelrogstad: Above

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+  def apply(underlying: BigInt): NanoBitcoins = NanoBitcoinsImpl(underlying)
+
+  private case class NanoBitcoinsImpl(underlying: BigInt) extends NanoBitcoins {
+    require(underlying >= LnPolicy.minNanoBitcoins, "Number was too small for NanoBitcoins, got: " + underlying)
+    require(underlying <= LnPolicy.maxNanoBitcoins, "Number was too big for NanoBitcoins, got: " + underlying)
+  }
+}
+
+sealed abstract class PicoBitcoins extends LnCurrencyUnit {
+  override type A = BigInt
```
torkelrogstad: Above

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+  def apply(underlying: BigInt): PicoBitcoins = PicoBitcoinsImpl(underlying)
+
+  private case class PicoBitcoinsImpl(underlying: BigInt) extends PicoBitcoins {
+    require(underlying >= LnPolicy.minPicoBitcoins, "Number was too small for PicoBitcoins, got: " + underlying)
+    require(underlying <= LnPolicy.maxPicoBitcoins, "Number was too big for PicoBitcoins, got: " + underlying)
+  }
+}
+
+object LnCurrencyUnits {
+  val PICO_TO_SATOSHIS = 10000
```
torkelrogstad: `private`

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+  val MSAT_TO_PICO = 10
```
torkelrogstad: `private`

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -0,0 +1,244 @@
...
+
+  def fromMSat(msat: MilliSatoshis): PicoBitcoins = {
+    //msat are technically 10^-11
+    //while pico are 10^-12, so we need to convert
+    val underlying = msat.toBigInt * MSAT_TO_PICO
+    PicoBitcoins(underlying)
+  }
+
+  def fromEncodedString(input: String): Try[LnCurrencyUnit] = {
+    val currency = input.splitAt(input.length - 1)
```
torkelrogstad: Would destructuring be more readable?

core/src/main/scala/org/bitcoins/core/protocol/ln/routing/LnRoute.scala
```diff
@@ -0,0 +1,62 @@
...
+      shortChannelID.bytes ++
+      feeBaseMsat.bytes ++
+      feePropMilli.bytes ++
+      ByteVector.fromValidHex(cltvExpiryDeltaHex)
+  }
+}
+
+object LnRoute {
+
+  def fromBytes(bytes: ByteVector): LnRoute = {
```
torkelrogstad, line 42: Slicing at magic numbers without length checking is probably something we should avoid.

core/src/main/scala/org/bitcoins/core/protocol/ln/util/LnUtil.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    val encoded = {
+      if (encodedNoPadding.size == 1) {
+        UInt5.zero +: encodedNoPadding
+      } else {
+        encodedNoPadding
+      }
+    }
+
+    require(encoded.size == 2, s"data_length must be 2 uint5s")
```
torkelrogstad: add message of how long was provided

core/src/main/scala/org/bitcoins/core/protocol/ln/LnParams.scala
torkelrogstad: @Christewart

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -0,0 +1,224 @@
...
+      case h :: t =>
+        if (!Bech32.charset.contains(h.toLower)) {
+          Failure(new IllegalArgumentException("Invalid character in data of bech32 address, got: " + h))
+        } else {
+          if ((h.isUpper && hasLower) || (h.isLower && hasUpper)) {
+            Failure(new IllegalArgumentException("Cannot have mixed case for bech32 address"))
+          } else {
+            val byte = Bech32.charset.indexOf(h.toLower).toByte
+
+            if (byte >= 0 && byte < 32) {
```
torkelrogstad, line 130: Isn't this already covered by line 111?

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -60,43 +60,62 @@ trait NumberUtil extends BitcoinSLogger {
...
         }
       }
 
       if (pad) {
         if (bits > UInt32.zero) {
           val r: Long = ((acc << (to - bits) & maxv)).toLong
-          ret = ret ++ Seq(UInt8(r.toShort))
+          ret.+=(f(UInt8(r.toShort)))
         }
       } else if (bits >= from || ((acc << (to - bits)) & maxv) != UInt8.zero) {
```
torkelrogstad, line 98: Intellij warns me we're comparing unrelated types, `UInt32` vs `UInt8`. I'm kind of afraid to touch this code though, with kind of messy bit operations all over the place...

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,522 @@
...
+    val uri = NodeUri(nodeId, host, port)
+    connect(uri)
+  }
+
+  override def connect(uri: NodeUri): Future[String] = {
+    logger.info(s"Connecting to $uri")
+    eclairCall[String]("connect", List(JsString(uri.toString)))
+  }
+
+  // When types are introduced this can be two different functions
```
torkelrogstad: Types are introduced now

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,522 @@
...
+      uri,
+      entity =
+        HttpEntity(ContentTypes.`application/json`, obj.toString))
+      .addCredentials(
+        HttpCredentials.createBasicHttpCredentials(username, password))
+  }
+
+  // TODO: THIS IS ALL HACKY
+
+  private def pathToEclairJar: String = {
```
torkelrogstad, line 490: should really throw here

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,522 @@
...
+    val path = System.getenv("ECLAIR_PATH")
+    val eclairV = "/eclair-node-0.2-beta8-52821b8.jar"
```
torkelrogstad, line 498: Should also probably check for jar exists

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+  }
+
+  def isStarted(): Boolean = {
+    val t = Try(Await.result(getInfo, 1.second))
+    t.isSuccess
+  }
+
+  def stop(): Option[Unit] = {
```
torkelrogstad, line 532: This might just be me being to inexperienced in Scala, but I feel having the `Option[Unit]` signature is somewhat quirky. Would it be better to return a boolean, signifying whether or not we stopped the process?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/config/AuthCredentials.scala
```diff
@@ -0,0 +1,102 @@
+package org.bitcoins.eclair.rpc.config
```
torkelrogstad: Rename file to match trait/object name?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/config/EclairInstance.scala
```diff
@@ -0,0 +1,82 @@
...
+  def apply(
+    network: NetworkParameters,
+    uri: URI,
+    rpcUri: URI,
+    authCredentials: EclairAuthCredentials): EclairInstance = {
+    EclairInstanceImpl(network, uri, rpcUri, authCredentials)
+  }
+
+  def fromDatadir(datadir: File): EclairInstance = {
+    val eclairConf = new File(datadir.getAbsolutePath + "/eclair.conf")
```
torkelrogstad, line 41: check for file existence?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/JsonReaders.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  implicit val channelUpdateReads: Reads[ChannelUpdate] = {
+    Json.reads[ChannelUpdate]
+  }
+
+  implicit val paymentRequestReads: Reads[PaymentRequest] = {
+    Json.reads[PaymentRequest]
+  }
+
+  /*  private def parsePaymentReq(obj: JsObject): JsResult[PaymentRequest] = {
```
torkelrogstad: Any reason to keep the commented stuff around?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/network/NodeUri.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private val defaultPort = ":9735"
+
+  def fromString(uri: String): Try[NodeUri] = {
+    val patternWithPort = ("""(\w+)@([\w\.]+(\w+)):(\d{2,5})""").r
+
+    val isUriWithPort = patternWithPort.findFirstIn(uri)
+
+    val nodeUriT = isUriWithPort match {
+      case Some(uri) =>
```
torkelrogstad: Shadowed variable name

rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -8,26 +8,32 @@ import org.bitcoins.core.config.NetworkParameters
...
 
 object BitcoindInstance {
   private case class BitcoindInstanceImpl(
     network: NetworkParameters,
     uri: URI,
     rpcUri: URI,
-    authCredentials: BitcoindAuthCredentials)
-    extends BitcoindInstance
+    authCredentials: BitcoindAuthCredentials,
+    zmqPortOpt: Option[Int]) extends BitcoindInstance
```
torkelrogstad, line 29: Should this have default value? `None` or `Some(28332)` (28332 is used in the Bitcoin Core ZMQ doc as an example, I don't know if this is standard)

core-test/src/test/scala/org/bitcoins/core/protocol/transaction/TransactionTest.scala
```diff
@@ -20,7 +21,7 @@ import scala.io.Source
  * Created by chris on 7/14/15.
  */
 class TransactionTest extends FlatSpec with MustMatchers {
-  private val logger = BitcoinSLogger.logger
+  private val logger = LoggerFactory.getLogger(this.getClass.getSimpleName)
```
nkohen, line 24: Do we have a trait for this in Bitcoin-s? Wasn't there some issue we've run into in the past using this pattern?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnPolicy.scala
```diff
@@ -0,0 +1,32 @@
...
+   * This is a self imposed limit, and is subject to change.
+   * Please see [[https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#adding-an-htlc-update_add_htlc BOLT2]]
+    * for more info.
+   */
+  val maxAmountMSat: MilliSatoshis = MilliSatoshis(4294967295L)
+
+  val milliMultiplier: BigDecimal = BigDecimal(0.001)
+  val microMultiplier: BigDecimal = BigDecimal(0.000001)
+  val nanoMultiplier: BigDecimal = BigDecimal(0.000000001)
+  val picoMultiplier: BigDecimal = BigDecimal(0.000000000001)
```
nkohen: Perhaps this could be an ADT `sealed abstract class LnMultiplier { def multiplier: BigDecimal }`?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -0,0 +1,199 @@
...
+      expiryTime = expiryTime,
+      cltvExpiry = cltvExpiry,
+      fallbackAddress = fallbackAddress,
+      routingInfo = routingInfo)
+  }
+
+
+  def fromUInt5s(u5s: Vector[UInt5]): LnTaggedFields = {
+    @tailrec
+    def loop(remaining: List[UInt5], fields: Vector[LnTag]): Vector[LnTag] = {
```
nkohen, line 129: A comment would be nice

eclair-rpc/README.md
```diff
@@ -0,0 +1,27 @@
+# Bitcoin-s Eclair RPC client
```
nkohen, line 1: It is probably worth mentioning somewhere in this readme that a bitcoind instance has to be running to run the eclair jar (and probably link to the bitcoind rpc client)

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,529 @@
...
+import org.bitcoins.eclair.rpc.network.{NodeUri, PeerState}
+import org.slf4j.LoggerFactory
+import play.api.libs.json._
+
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext, Future, Promise}
+import scala.sys.process._
+import scala.util.{Failure, Properties, Success, Try}
+
+class EclairRpcClient(val instance: EclairInstance)(implicit system: ActorSystem) extends EclairApi {
```
nkohen, line 33: scaladoc?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,529 @@
...
+      pushMsat = None,
+      Some(feerateSatPerByte),
+      Some(channelFlags))
+  }
+
+  override def getPeers: Future[Vector[PeerInfo]] = {
+    eclairCall[Vector[PeerInfo]]("peers")
+  }
+
+  override def receive(
```
nkohen, line 254: scaladoc?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -0,0 +1,529 @@
...
+    HttpRequest(
+      method = HttpMethods.POST,
+      uri,
+      entity =
+        HttpEntity(ContentTypes.`application/json`, obj.toString))
+      .addCredentials(
+        HttpCredentials.createBasicHttpCredentials(username, password))
+  }
+
+  // TODO: THIS IS ALL HACKY
```
nkohen: Probably get rid of this comment

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/config/EclairInstance.scala
```diff
@@ -0,0 +1,82 @@
...
+    * [[org.bitcoins.eclair.rpc.config.EclairInstance EclairInstance]]
+    */
+  def fromConfig(config: Config): EclairInstance = {
+    val chain = config.getString("eclair.chain")
+
+    val serverBindingIp = config.getString("eclair.server.binding-ip")
+    val serverPort = ConfigUtil.getIntOrElse(config, "eclair.server.port", 9735)
+
+    val rpcHost = config.getString("eclair.api.binding-ip")
+    val rpcPort = ConfigUtil.getIntOrElse(config, "eclair.api.port", 8080)
```
nkohen: Names for magic numbers

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/JsonReaders.scala
nkohen: ^ same for EclairModels?

Oops this didn't turn out like I thought it would, I'm referencing @torkelrogstad's comment about getting rid of commented out stuff

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/network/NodeUri.scala
```diff
@@ -0,0 +1,46 @@
+package org.bitcoins.eclair.rpc.network
+
+import org.bitcoins.core.protocol.ln.node.NodeId
+
+import scala.util.{ Failure, Success, Try }
+
+case class NodeUri(nodeId: NodeId, host: String, port: Int) {
```
nkohen, line 7: Any chance there's an easy non-string type we can use for host here?

nkohen (commented):
Here is my code review, I didn't look at any tests and I also didn't include any of my nits that I didn't think were important enough to slow down this branch being merged.

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
Christewart: The valid range of UInt8 values are from 0 - 255, while the valid range of UInt5 values are 0 - 31. If the number is between 31 - 255 it cannot be a UInt5

rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
Christewart: I ended up defaulting to `None` since you need to explicitly specify the address in your `bitcoin.conf`. We also have a bad limitation here in the fact that we can only use one port, when in reality you want to have different ports for different zmq subscriber queues. 

https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md

I think we will have to address this in another PR. I've filed an issue to rectify this #262

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/network/NodeUri.scala
Christewart: Not that I can find from a cursory search

PR #260, opened 2018-12-05
Addresses some review on #256

torkelrogstad:
With the first two commits I had the intention of keeping the number of commits relatively small and mergeable as is, but there's going to be more commits (with most likely a bugfix commit or two on top of that). I'll therefore just commit away, and squash later.


core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -94,31 +96,27 @@ object LnTaggedFields {
...
-      InvoiceTagImpl(
-        paymentHash = paymentHash,
-        description = None,
-        nodeId = nodeId,
-        descriptionHash = descriptionOrHash.right.toOption,
-        expiryTime = expiryTime,
-        cltvExpiry = cltvExpiry,
-        fallbackAddress = fallbackAddress,
-        routingInfo = routingInfo)
+    val (description, descriptionHash): (Option[LnTag.DescriptionTag],
```
Christewart, line 99: Probably would want to use `Either[LnTag.DescriptionTag, LnTag.DescriptionHashTag]` since I believe BOLT11 says that one of the two has to be defined

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -141,46 +139,40 @@ object LnTaggedFields {
 
           loop(newRemaining, fields.:+(tag))
 
-        case _ :: _ | _ :: _ :: _ =>
-          throw new IllegalArgumentException("Failed to parse LnTaggedFields, needs 15bits of meta data to be able to parse")
         case Nil =>
           fields
+        case _ =>
```
Christewart: This is bad, the previous 
```scala

case _ :: _ | _ :: _ :: _ => ... 

```

Is creating pattern match exhaustiveness

```scala

case _ => 

```

is dangerous because `_` can be any type.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
```diff
@@ -141,46 +139,40 @@ object LnTaggedFields {
...
 
-      val fOpt = tags.find(_.isInstanceOf[LnTag.PaymentHashTag])
-      if (fOpt.isDefined) {
-        fOpt.get.asInstanceOf[LnTag.PaymentHashTag]
-      } else {
-        throw new IllegalArgumentException(s"Payment hash must be defined in a LnInvoice")
-      }
+    val tags = loop(u5s.toList, Vector.empty)
 
+    def getTag[T <: LnTag : ClassTag]: Option[T] = {
```
Christewart, line 152: private

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -75,36 +75,33 @@ object LnTag {
       val u8 = UInt8.zero
     }
 
+    private def witnessFromU8(bytes: ByteVector, np: NetworkParameters) = {
+      val witSPK = bytes.size match {
```
Christewart, line 79: .length for intellij

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -75,36 +75,33 @@ object LnTag {
...
+        case 32 =>
+          val hash = Sha256Digest.fromBytes(bytes)
+          P2WSHWitnessSPKV0.fromHash(hash)
+        case 20 =>
+          val hash = Sha256Hash160Digest.fromBytes(bytes)
+          P2WPKHWitnessSPKV0.fromHash(hash)
+        case _ =>
```
Christewart:
```
case _ : Int => 

```

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
```diff
@@ -75,36 +75,33 @@ object LnTag {
...
-              P2WPKHWitnessSPKV0.fromHash(hash)
-            } else {
-              throw new IllegalArgumentException(s"Can only create witness spk out of a 32 byte or 20 byte hash, got ${bytes.length}")
-            }
-          }
-
-          Bech32Address(witSPK, np)
-        } else {
+        case WitSPK.u8 => witnessFromU8(bytes, np)
+        case _ =>
```
Christewart:
```scala

case _: UInt8 => ... 

```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -33,7 +33,9 @@ trait EclairApi {
 
   def close(id: ChannelId, spk: ScriptPubKey): Future[String]
 
-  def findRoute(nodeIdOrInvoice: Either[NodeId, LnInvoice]): Future[Vector[String]]
+  def findRoute(nodeId: NodeId ): Future[Vector[String]]
+
+  def findRoute(invoice: LnInvoice): Future[Vector[String]]
```
Christewart: Let's implement this in the `EclairApi` file. You should be able to do 

```scala

def findRoute(invoice: LnInvoice): Future[Vector[String]] = {
  findRoute(nodeId = invoice.nodeId)
}

```

The return type for this should probably also change to 

```scala 

def findRoute(): Future[Vector[LnRoute]]

```

but we can change that on a separate follow up PR

Christewart (commented):
This was a huge improvement on the existing `ln` branch! Great work.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
torkelrogstad: That's what's being used in the signature of the containing method. My change here is to just unwrap the `Either` into a tuple, to avoid calling `InvoiceTagImpl()` twice with almost identical arguments.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala
torkelrogstad: I don't think that should matter here? `getTag` is defined inside `fromUInt5s`, so shouldn't be accessible from the outside anyway. Same as with `loop` a few lines up

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
torkelrogstad: There's no change in semantics here, as `version` is an `UInt8` anyway. Unless you want the change for readability, which is fine.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
torkelrogstad: from `ByteVector.scala` implementation: 
```scala
  /**
   * Alias for [[size]].
   * @group collection
   */
  final def length: Long = size
```

So here it doesn't matter.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
torkelrogstad: Same as convo below

core/src/main/scala/org/bitcoins/core/protocol/ln/LnTags.scala
Christewart: Let's change it for readability, and for the case that if we change the type of `version` in the future we will get a compiler error here :-)

PR #266, opened 2018-12-10
Splits V16 tests into separate file, other tests into traits based on what kind of RPC call

torkelrogstad:
Problematic tests: 
- [x] Entire test suites times out altogether every now and then in `beforeAll` - we could just up the timeout value
- [x] `should be able to abandon a transaction` - this has been failing consistently
- [x] `should be able to get network statistics`
- [x] `should be able to create a connected node pair with 100 blocks and then delete them` - failing somewhat often
- [x] `should be able to clear banned subnets`


testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -213,62 +243,193 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+    }
+  }
+
+  def fundMemPoolTransaction(
+      sender: BitcoindRpcClient,
+      address: BitcoinAddress,
+      amount: Bitcoins)(
+      implicit ec: ExecutionContext): Future[DoubleSha256Digest] = {
+    sender
+      .sendToAddress(address, amount)
```
torkelrogstad: This method was meant to be in the second commit, because it is a hack which shouldn't be there. I was having issues with transactions not being propagated/accepted into the mempool by doing just `sendToAddress`, so I had to go this weird route of `sendToAddress` -> `getTransaction` -> `sendRawTransaction`

rpc/src/main/scala/org/bitcoins/rpc/client/v16/BitcoindV16RpcClient.scala
```diff
@@ -0,0 +1,88 @@
...
+  */
+class BitcoindV16RpcClient(override val instance: BitcoindInstance)(
+    implicit
+    m: ActorMaterializer)
+    extends BitcoindRpcClient(instance)
+    with V16AccountRpc
+    with V16SendRpc {
+
+  override val version: BitcoindVersion = BitcoindVersion.V16
+  require(version == instance.getVersion, "bitcoind version must be 0.16")
```
torkelrogstad: `instance.getVersion` does `bitcoind --version` e.g. an IO operation. Is that problematic?

rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -23,6 +25,21 @@ sealed trait BitcoindInstance {
   def zmqConfig: ZmqConfig
 
   def rpcPort: Int = authCredentials.rpcPort
+
+  def getVersion: BitcoindVersion = {
```
torkelrogstad: See other comment about potentially problematic IO operation

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -213,62 +243,193 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
 
-        if (try4.isFailure) {
-          deleteNodePair(client1, client2)
-          throw try4.failed.get
+  def createRawCoinbaseTransaction(
+      sender: BitcoindRpcClient,
+      receiver: BitcoindRpcClient,
+      amount: Bitcoins = Bitcoins(1))(
+      implicit executionContext: ExecutionContext): Future[Transaction] = {
+    sender.generate(2).flatMap { blocks =>
```
torkelrogstad: This nested `flatMap` is a good spot for a `for`-comprehension

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -213,62 +243,193 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+              }
+            }
+          }
         }
-        (client1, client2)
       }
     }
   }
 
+  def signRawTransaction(
```
torkelrogstad, line 431: This util method is needed because V16 and V17 has different APIs for signing raw transactions (V17 deprecated the `signrawtransaction` RPC and split it into `signrawtransactionwithwallet` and `signrawtransactionwithkey`)

Quite a few tests sign raw transactions, and I figured it was best to use this util methods in that case, and then implement tests specifically for the V16 and V17 RPC calls

torkelrogstad:
For some reason there's a lot more test failures due to timeouts on macOS...

rpc/src/main/scala/org/bitcoins/rpc/client/v16/BitcoindV16RpcClient.scala
Christewart: I don't think so

rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
Christewart: What is the main concern here? Is this going to be getting called a lot?

rpc/src/test/scala/org/bitcoins/rpc/common/P2PRpcTest.scala
```diff
@@ -0,0 +1,234 @@
...
+    assert(info.connections == 0)
+    freshClient.addNode(otherFreshClient.getDaemon.uri, AddNodeArgument.Add)
+    BitcoindRpcTestUtil.awaitConnection(freshClient, otherFreshClient)
+    val newInfo = await(freshClient.getNetworkInfo)
+    assert(newInfo.connections == 1)
+
+  }
+
+  it should "be able to get network statistics" in {
+    BitcoindRpcTestUtil.createNodePair() flatMap {
```
Christewart: Not a fan of infix notation as previously talked about

rpc/src/test/scala/org/bitcoins/rpc/v17/BitcoindV17RpcClientTest.scala
```diff
@@ -131,7 +134,7 @@ class BitcoindV17RpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
     val l = "setLabel"
     val btc = Bitcoins(1)
     val addr = await(client.getNewAddress)
-    await(TestUtil.fundBlockChainTransaction(otherClient, addr, btc))
+    await(BitcoindRpcTestUtil.fundBlockChainTransaction(otherClient, addr, btc))
```
Christewart: I think it is better to do something like 

```scala
val _ = await(BitcoindRpcTestUtil.fundBlockChainTransaction(otherClient,addr,btc))
```

This is says to the reader of this code that `fundBlockchainTransaction` returns something, but we do not care about it in this context. Thus the `val _ = ...`

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
Christewart: Let's add this comment to the source code

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -213,62 +243,193 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+      transaction: Transaction,
+      utxoDeps: Vector[RpcOpts.SignRawTransactionOutputParameter] = Vector.empty,
+  )(implicit actorMaterializer: ActorMaterializer): Future[
+    SignRawTransactionResult] = signer match {
+    case v17: BitcoindV17RpcClient =>
+      v17.signRawTransactionWithWallet(transaction, utxoDeps)
+    case v16: BitcoindV16RpcClient =>
+      v16.signRawTransaction(transaction, utxoDeps)
+    case unknown: BitcoindRpcClient =>
+      val v16T = BitcoindV16RpcClient.fromUnknownVersion(unknown)
```
Christewart: Why would we need to pattern match try and construct a `V16` or `V17` client when we already know that it isn't one of those. I think this code is uncessary.

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
Christewart: That is strange. Add this comment as a scaladoc on this method

torkelrogstad: I was thinking I should try and remove it first, and see if the same errors are still there

rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
torkelrogstad: It's a IO operation (although a very modest one) that's done by blocking. It's called on startup (to check we have the correct version), and when using utility methods such as `signRawTransaction` where we create a versioned RPC client from an unversioned one

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
torkelrogstad: V16 or V17 clients are only obtained by explicitly requesting one. A general `BitcoindRpcClient` can have both versions in it's underlying `bitcoind`. So `case unknown: BitcoindRpcClient` doesn't mean it's not V16 or V17, it means we're not sure what it is.

We figure out what version we're dealing with by trying to construct V16 and V17 clients, throwing an error if we're unable to construct both of them

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -122,12 +136,28 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
 
+  /**
+    * Stops the given servers, and waits for them (in a blocking manner)
+    * until they are stopped. Deletes their temp dir.
+    */
+  def stopServers(servers: BitcoindRpcClient*)(
+      implicit system: ActorSystem): Unit =
+    servers.foreach(server => {
+      server.stop()
+      RpcUtil.awaitServerShutdown(server)
```
nkohen: If you made this two separate calls (one that calls stop for each server, and the other that waits for it to stop and then deletes its TmpDir ), then this would be much faster for longer lists of servers where currently it is waiting to completely take down one server before starting on the next

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -338,19 +339,19 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
     val txsF = for {
       blocks <- sender.generate(2)
       firstBlock <- sender.getBlock(blocks(0))
-      firstTransaction <- sender.getTransaction(firstBlock.tx(0))
+      firstTransaction <- sender.getRawTransaction(firstBlock.tx.head)
```
torkelrogstad, line 396: We were having issues with `gettransaction` sporadically failing, either by missing the `blockindex` field or getting `Invalid or non-wallet transaction id (code -5)`

Is this problematic? From what I can tell it should be fine

rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RawTransactionResult.scala
```diff
@@ -60,10 +60,10 @@ case class GetRawTransactionResult(
     locktime: UInt32,
     vin: Vector[GetRawTransactionVin],
     vout: Vector[RpcTransactionOutput],
-    blockhash: DoubleSha256Digest,
+    blockhash: Option[DoubleSha256Digest],
```
torkelrogstad: We had some errors in our RPC definitions. 
Bitcoin Core developer reference: https://bitcoin.org/en/developer-reference#getrawtransaction

PR #268, opened 2018-12-13
WIP: Add ability to publish to bintray, take on the org.scijava native-lib…

Christewart:
…-loader dependenency so we can bundle secp256k1 natively in a jar file and publish. This dependency will allow us to link the correct secp256k1 version dependending on the os and arch in the future

This PR does a few things: 

1. Takes on a new dependency https://github.com/scijava/native-lib-loader . This allows us to add secp256k1 files to our classpath and then load them at runtime based on the env we are on. Currently this PR only has linux x86_64 binaries for libsecp256k1.

2. Integrates the ability to publish to [bintray](https://bintray.com/bitcoin-s/bitcoin-s-core). The only thing that has been published so for is the `secp256k1jni` project. This will allow users to consume bitcoin-s in a way that is more convenient than manually building it.

3. Integrates the [sbt-pgp](https://www.scala-sbt.org/sbt-pgp/) plugin to produce signed binaries.


torkelrogstad:
I'm getting the same result as Travis, @Christewart.

torkelrogstad:
This is the contents of `secp256k1jni/natives`: 

```
secp256k1jni/natives/
├── linux_32
│   ├── bench_sign
│   ├── bench_verify
│   ├── libsecp256k1.a
│   ├── libsecp256k1.la -> ../libsecp256k1.la
│   ├── libsecp256k1.lai
│   ├── libsecp256k1.so -> libsecp256k1.so.0.0.0
│   ├── libsecp256k1.so.0 -> libsecp256k1.so.0.0.0
│   └── libsecp256k1.so.0.0.0
└── linux_64
    ├── libsecp256k1.lai
    ├── libsecp256k1.so.0 -> libsecp256k1.so.0.0.0
    └── libsecp256k1.so.0.0.0
```

`linux_64` looks suspiciously empty compared to `linux_32`. But you're also using 64 bit Ubuntu, @Christewart? So should be same as me and Travis

Also, `linux_32/libsecp256k1.la` symlinks to a file which doesn't exist on my system: 

```bash
$ file secp256k1jni/natives/linux_32/libsecp256k1.la
secp256k1jni/natives/linux_32/libsecp256k1.la: broken symbolic link to ../libsecp256k1.la
```

PR #274, opened 2018-12-18
RpcRetryException

nkohen:
Created RpcRetryException for better error messages on RpcUtil timeouts


Christewart (commented):
So a couple things:

1. There is two `RpcUtil` classes in the `bitcoin-s-bitcoind-rpc` project. I must have duplicated them. Can you delete one of them?
2. There is an `RpcUtil` in `testkit`. Now that I think about this, we could probably delete this file too. However this will couple the `testkit` to `RpcUtil` in the `bitcoin-s-bitcoind-rpc` project pretty strongly. I think we might want to consider renaming `RpcUtil` -> `AsyncUtil` in the `teskit` project and then delegate calls to the `RpcUtil` in the `bitcoin-s-bitcoind-rpc` project. This means if we need `testkit` specific functionality in the future, we don't have to add it to the `RpcUtil` in the rpc project.
3. Maybe we should consider renaming `RpcUtil` all together? At least in `testkit` since i think this is used for more generic async stuff? At least it could be used anytime we have a `Future` and we want to see if it has completed in a given interval of time in a non blocking manner right?

Christewart:
Looks like there are `ClassNotFoundExceptions` happening

PR #280, opened 2019-01-02
WIP: 2018 12 22 node project

Christewart:
This ports code over from the legacy [`spv node`](https://github.com/bitcoin-s/bitcoin-s-spv-node) project to be included in bitcoin-s. 

- [x] compile
- [x] test:compile
- [ ] tests pass
- [ ] block headers sync

## Testing this PR

To use this PR, you will need to publish a the `testkit` project locally with version `0.0.3-SNAPSHOT`. To do this you need to revert the `bitcoinsV` to `0.0.1` in the [`project/Deps.scala`](https://github.com/Christewart/bitcoin-s-core/blob/2018-12-22-node-project/project/Deps.scala#L23). 

After doing this, you will need to run 
```scala 
$ sbt clean
$ sbt compile
$ sbt publishLocal
``` 

This will create local bitcoin-s dependencies with version `0.0.3-SNAPSHOT`. You can check if this exists by doing 

```bash
$ ls ~/.ivy2/local/org.bitcoins/bitcoin-s-testkit_2.12/
0.0.3-SNAPSHOT
```

After creating the local dependencies, you need to change the [`bitcoinsV`](https://github.com/Christewart/bitcoin-s-core/blob/2018-12-22-node-project/project/Deps.scala#L23) BACK TO `0.0.3-SNAPSHOT`. 

Now you should be able to compile the project successfully with 

```bash
sbt test:compile
```


node/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
```
torkelrogstad, line 1: This file should be removed?

torkelrogstad:
Files that should be looked at: 
`bitcoin-s-core/node/src/main/scala/org/bitcoins/node/networking/PaymentActor.scala`
`bitcoin-s-core/node/src/main/scala/org/bitcoins/node/networking/PeerMessageHandler.scala`
`bitcoin-s-core/node/src/main/scala/org/bitcoins/node/networking/Client.scala`

PR #281, opened 2019-01-03
Add ability to publish snapshots for bitcoin-s

Christewart:
This allows us to publish bitcoin-s snapshots to oss.jfrog.org. You can see the snapshots here:

https://oss.jfrog.org/webapp/#/artifacts/browse/tree/General/oss-snapshot-local/org/bitcoins

I've bumped the version of bitcoin-s to 0.0.2-SNAPSHOT, because we get a weird dependency resolution error if if we depend on the same version as we have specified in version.sbt. 

I.e. if we need `0.0.1-SNAPSHOT` for `bitcoinsV` in `Deps.scala`, we cannot have the version be `0.0.1-SNAPSHOT` in version.sbt.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,6 @@
+import sbt.Credentials
+import sbt.Keys.publishTo
+
 cancelable in Global := true
 
 lazy val commonCompilerOpts = {
@@ -26,6 +29,8 @@ lazy val compilerOpts = Seq(
 
 lazy val testCompilerOpts = commonCompilerOpts
 
+
+
 lazy val commonSettings = List(
   scalacOptions in Compile := compilerOpts,
   scalacOptions in Test := testCompilerOpts,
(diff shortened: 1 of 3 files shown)


torkelrogstad:
Does this solve the use case where the interface of `testkit` is changed, and something in for example `rpc` or `eclair-rcp` needs those changes? 

A possible solution (haven't tested, though): 
1) `sbt compile`
2) Publish snapshots some way, either `sbt publishLocal` or push to Bintray/JFrog
3) Read in `version.sbt` value, replace `bitcoinsV` value in `Deps.scala`
3) `sbt test` - we want this to pull down the latest changes in `testkit`, either pulling from local `.ivy2` or Bintray/JFrog

I tried doing the local way of doing this by goofing around with Travis config, with no success: 
91e6d185a7582eeaeb0e72fcab7fe9b7909924c0
3596d49233f3403772acfbcb1d6801f19cb6b338
7e1efede1e7d5cbcd55be9e8feda3b5c7b206e63
c3ed9fc360eab93679dc1b651bc85573945bcea5

Christewart:
@torkelrogstad Your comment is addressed in #282 right?

Christewart:
For future reference on publishing: http://szimano.org/automatic-deployments-to-jfrog-oss-and-bintrayjcentermaven-central-via-travis-ci-from-sbt/

PR #285, opened 2019-01-06
Fix BOLT11 incompatibility with byte boundaries and sigdata when recovering public key

### LINKED ISSUE #277: Decoding ln invoice with bitcoin-s returns different results than eclair
```
lntb100n1pwz34mzpp5dwqtndejp0qjqdf5u78cdd4r99zuxk45vnj8tmgqayk8hxr4t7wsd890v3xgatjv96xjmmwygarzvpsxqczcgnrdpskumn9ds3r5gn5wfskgetnygkzyetkv4h8gg36yfeh2cnnvdexjcn9ygkzyat4d9jzyw3zxqcrzvfjxgenxtf5xs6n2tfkxcmnwtfc8qunjttpv93xycmrv3jx2etxvc3zcgn90p3ksctwvajjyw3zvf5hgenfdejhsg3vyfehjmtzdakzyw3zgf2yx42ngs386xqrrssqr6xn7dtkyxk0rhl98k3esksst578uwhud5glp9svq24ddwlgqwz6v9uf7mqljrj07xl87ufrn4yfplrsz2vpmc9xwv44634h54dq3sq257hh4
```

the node id returned by bitcoin-s is 
```
02b6a252ee5d895e87540e1bdbca4c33d26ef87ec9a6069ccf6a6275a754510e59
```

while eclair returns 

```
$node1 checkinvoice lntb100n1pwz34mzpp5dwqtndejp0qjqdf5u78cdd4r99zuxk45vnj8tmgqayk8hxr4t7wsd890v3xgatjv96xjmmwygarzvpsxqczcgnrdpskumn9ds3r5gn5wfskgetnygkzyetkv4h8gg36yfeh2cnnvdexjcn9ygkzyat4d9jzyw3zxqcrzvfjxgenxtf5xs6n2tfkxcmnwtfc8qunjttpv93xycmrv3jx2etxvc3zcgn90p3ksctwvajjyw3zvf5hgenfdejhsg3vyfehjmtzdakzyw3zgf2yx42ngs386xqrrssqr6xn7dtkyxk0rhl98k3esksst578uwhud5glp9svq24ddwlgqwz6v9uf7mqljrj07xl87ufrn4yfplrsz2vpmc9xwv44634h54dq3sq257hh4
{
  "prefix": "lntb",
  "amount": 10000,
  "timestamp": 1546180450,
  "nodeId": "03fad6c016f998e85d03ce0b7358b3b6a38ebc7fd60030340d0245fea0d95c8c12",
  "tags": [
    {
      "hash": "6b80b9b7320bc1203534e78f86b6a32945c35ab464e475ed00e92c7b98755f9d"
    },
    {
      "description": "{\"duration\":10000,\"channel\":\"trades\",\"event\":\"subscribe\",\"uuid\":\"00112233-4455-6677-8899-aabbccddeeff\",\"exchange\":\"bitfinex\",\"symbol\":\"BTCUSD\"}"
    },
    {
      "seconds": 3600
    }
  ],
  "signature": "00f469f9abb10d678eff29ed1cc2d082e9e3f1d7e3688f84b0601556b5df401c2d30bc4fb60fc8727f8df3fb891cea4487e38094c0ef0533995aea35bd2ad04600"
}
```

Christewart:
WIP2

Fix bug where we weren't padding byte boundary on signature data

This fixes #277 

There is a very peculiar requirement in BOLT11 related to signature data. When you are converting signature data from `u5s: Vector[UInt5]` -> `u8s: Vector[UInt8]` you need to pad to the nearest byte boundary. This makes sense since the `u5s` is not guaranteed to line up on a byte boundary. 

However the [BOLT11 standard](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#requirements-2) says in the case where it DOES line up on a byte boundary, you need to pad the signature data with ANOTHER byte. 

> ... represented as UTF-8 bytes, concatenated with the data part (excluding the signature) with zero bits appended to pad the data to the next byte boundary, with a trailing byte containing the recovery ID (0, 1, 2 or 3)

This commit checks if the `u5s` line up at a byte boundary, and if they do it appends another zero byte. This should make us compatible with other lightning implementations.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
@@ -1,12 +1,12 @@
 package org.bitcoins.core.crypto
 
-import org.bitcoins.core.util.{BitcoinSUtil, Factory}
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
 import scodec.bits.ByteVector
 
 /**
   * Created by chris on 2/26/16.
   */
-sealed abstract class ECDigitalSignature {
+sealed abstract class ECDigitalSignature extends BitcoinSLogger {
 
   def hex: String = BitcoinSUtil.encodeHex(bytes)
 
@@ -37,27 +37,42 @@ sealed abstract class ECDigitalSignature {
   def decodeSignature: (BigInt, BigInt) = DERSignatureUtil.decodeSignature(this)
 
   /** Represents the r value found in a elliptic curve digital signature */
-  def r: BigInt = decodeSignature._1
+  def r: BigInt = {
+    decodeSignature._1
+  }
 
-  /** Represents the s value found in a elliptic curve digital signature */
-  def s: BigInt = decodeSignature._2
+  /** If we need to do serialization with the
+    * r value, you should use this. It will pad
+    * the byte vector so we have exactly 32 bytes
+    * @return
+    */
+  def rBytes: ByteVector = {
+    val bytes = r.bigInteger.toByteArray.takeRight(32)
+    val padded = ByteVector(bytes).padLeft(32)
+    padded
+  }
 
+  /** If we need to do serialization with the
+    * s value, you should use this. It will pad
+    * the byte vector so we have exactly 32 bytes
+    * @return
+    */
+  def s: BigInt = {
+    decodeSignature._2
+  }
+
+  def sBytes:ByteVector = {
+    val bytes = s.bigInteger.toByteArray.takeRight(32)
+    val padded = ByteVector(bytes).padLeft(32)
+    padded
+  }
   /**
     * Creates a ByteVector with only
     * the 32byte r value and 32 byte s value
     * in the vector
     */
   def toRawRS: ByteVector = {
-
-    val rBytes = r.toByteArray.takeRight(32)
-    val sBytes = s.toByteArray.takeRight(32)
-
-    val rPadded = ByteVector(rBytes).padLeft(32)
-    val sPadded = ByteVector(sBytes).padLeft(32)
-
-    require(rPadded.size == 32, s"rPadded.size ${rPadded.size}")
-    require(sPadded.size == 32, s"sPadded.size ${sPadded.size}")
-    rPadded ++ sPadded
+    rBytes ++ sBytes
   }
 
 }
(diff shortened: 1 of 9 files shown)


Christewart:
Conversation with Alex Bosworth about this: http://gnusha.org/lightning-dev/2019-01-06.log

Many thanks to Alex for being extremely patient and helpful

core/src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
```diff
@@ -37,27 +37,44 @@ sealed abstract class ECDigitalSignature {
...
+
+  /** If we need to do serialization with the
+    * r value, you should use this. It will pad
+    * the byte vector so we have exactly 32 bytes
+    * @return
+    */
+  def rBytes: ByteVector = {
+    val bytes = r.bigInteger.toByteArray.takeRight(32)
+    val padded = ByteVector(bytes).padLeft(32)
+    require(padded.length == 32, s"r value not 32 bytes in size, got ${padded.length}")
```
torkelrogstad: I'm a bit confused as to what this `require` does. Is there any scenario where `padded.length` is not 32? `ByteVector.padLeft` throws if the length of vector we're padding is greater than the specified length

core/src/main/scala/org/bitcoins/core/crypto/ECDigitalSignature.scala
```diff
@@ -37,27 +37,44 @@ sealed abstract class ECDigitalSignature {
...
+    * @return
+    */
+  def s: BigInt = {
+    decodeSignature._2
+  }
 
+  def sBytes:ByteVector = {
+    val bytes = s.bigInteger.toByteArray.takeRight(32)
+    val padded = ByteVector(bytes).padLeft(32)
+    require(padded.length == 32, s"s value not 32 bytes in size, got ${padded.length}")
```
torkelrogstad: same as above

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -233,7 +239,20 @@ object LnInvoice {
...
+    val allBytes = hrp.bytes ++ payload
+
+
+    //for an explanation of why this is needed see
+    //https://github.com/bitcoin-s/bitcoin-s-core/issues/277
+    //https://github.com/bitcoin-s/bitcoin-s-core/pull/285
+    val allBytesU5s = Bech32.from8bitTo5bit(allBytes)
+
+    if (allBytesU5s.length % 8 == 0) {
+      allBytes ++ ByteVector(List(0.toByte))
```
torkelrogstad: This is equivalent to just `ByteVector(0)`

Issue #283, opened 2019-01-05
Fix scaladoc warnings


Christewart:
There are a ton of scaladoc warnings when we publish bitcoin-s. Fix these

https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/475679754#L725

PR #286, opened 2019-01-07
Fixes some Scaladocs

torkelrogstad:
addresses #283

Does not fix all warnings, but a bunch of them

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilderError.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilderError.scala
@@ -3,7 +3,7 @@ package org.bitcoins.core.wallet.builder
 import scala.util.Failure
 
 /**
-  * Represents an error that can be returned by the [[org.bitcoins.core.wallet.builder.TxBuilder]]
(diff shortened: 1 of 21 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilderError.scala
```diff
@@ -127,15 +139,19 @@ object TxBuilderError {
 
   /**
     * Means that the signed version of this transaction has MORE outputs than what was specified
-    * when building the [[TxBuilder]]. [[TxBuilder.destinations]] && [[TxBuilder.changeOutput]] should
+    * when building the [[org.bitcoins.core.wallet.builder.TxBuilder TxBuilder]].
+    * [[org.bitcoins.core.wallet.builder.TxBuilder.destinations TxBuilder.destinations]] &&
+    * [[org.bitcoins.core.wallet.builder.TxBuilder.changeSPK TxBuilder.changeSPK]] should
```
torkelrogstad, line 144: Changed `changeOutput` to `changeSPK`, as `changeOutput` does not exist on `TxBuilder`. Was this correct?

Christewart: Yes, we since we do not know the amount we cannot have a change output.

PR #290, opened 2019-01-08
version 0.0.3

Christewart:


Christewart:
So this PR ended up fixing an issue to do with cross compiling scala versions as well. If you look inside of [inThisBuild.sbt](https://github.com/bitcoin-s/bitcoin-s-core/pull/290/files#diff-19d5a2c32570a5cb6aa1b9aacdd59e49R3) we omitted the postfix `in ThisBuild`. This means we did not actually compile bitcoin-s against the second compiler series -- `2.11.x`.

With this fix, we can now compile bitcoin-s against both scala `2.11` and `2.12` (with a few fixes for the `2.11` series in this PR). You can do this by 

```scala
 +test:compile
```

If you want to read more on cross compiling and cross publishing for different scala versions please see these links

- https://www.scala-sbt.org/1.x/docs/Publishing.html#Cross-publishing
- https://www.scala-sbt.org/1.x/docs/Cross-Build.html#Cross+building+a+project

I've published binaries for both scala 2.11 and 2.12 on bintray for our dependencies: 

https://bintray.com/bitcoin-s/bitcoin-s-core

rpc/src/test/scala/org/bitcoins/rpc/BitcoindInstanceTest.scala
```diff
@@ -18,8 +18,13 @@ class BitcoindInstanceTest extends AsyncFlatSpec with BeforeAndAfterAll {
   private implicit val actorSystem: ActorSystem = ActorSystem(
     "BitcoindInstanceTest")
 
-  private val sampleConf: Seq[String] =
-    Source.fromResource("sample-bitcoin.conf").mkString.split("\n")
+  private val source =
+    Source.fromURL(getClass.getResource("/sample-bitcoin.conf"))
+
+  private val sampleConf: Seq[String] = {
+    source.getLines.toSeq
```
torkelrogstad, line 25: Is this tested locally? I believe I did something similar to this, and it didn't work

Christewart: It does pass for me locally. Check for yourself in the morning

torkelrogstad: Works for me as well. I'm not sure what I did the other day, then

A small (not important, feel free to ignore) simplifIcation I found: 

```scala
private val source = Source.fromResource("sample-bitcoin.conf")
private val sampleConf = source.getLines
```

PR #298, opened 2019-01-17
Improves types for Eclair RPC

torkelrogstad:


eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -24,13 +28,33 @@ case class GetInfoResult(
 case class PeerInfo(
     nodeId: NodeId,
     state: PeerState,
-    //address: String,
+    address: Option[String],
     channels: Int)
 
+sealed trait ChannelInfoT {
```
torkelrogstad: I'm not sure what to call this...
The issue here is that the result of querying for channel info varies based on what state the channel is in. Here I've only implemented the `ChannelState.NORMAL` and the default cases, haven't looked into the details of the other cases.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -19,9 +19,16 @@ trait EclairApi {
 
   def allNodes(): Future[Vector[NodeInfo]]
 
-  def allUpdates(nodeIdOpt: Option[NodeId]): Future[Vector[ChannelUpdate]]
+  def allUpdates(): Future[Vector[ChannelUpdate]]
 
-  def channels(nodeId: NodeId): Future[Vector[ChannelInfo]]
+  def allUpdates(nodeId: NodeId): Future[Vector[ChannelUpdate]]
+
+  def allUpdates(nodeIdOpt: Option[NodeId]): Future[Vector[ChannelUpdate]] = nodeIdOpt match {
```
Christewart: Perhaps make a default param here to be `nodeIdOpt: Option[NodeId] = None`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
Christewart: It seems to me the abstract type should be `ChannelInfo` and then the thing you have named `ChannelInfo` now should be renamed to `NormalChannelInfo`? As in the channel is in a normal state? Maybe i am misunderstaning what is going on here. You could also do 

```
sealed abstract class ChannelInfo {
  def nodeId: NodeId
  def channelId: FundedChannelId
  def localMsat: MilliSatoshis
  def remoteMsat: MilliSatoshis
  def state: ChannelState
}

case class BaseChannelInfo(
    nodeId: NodeId,
    channelId: FundedChannelId,
    localMsat: MilliSatoshis,
    remoteMsat: MilliSatoshis,
    state: ChannelState
) extends ChannelInfo

case class OpenChannelInfo(
    nodeId: NodeId,
    shortChannelId: ShortChannelId,
    channelId: FundedChannelId,
    localMsat: MilliSatoshis,
    remoteMsat: MilliSatoshis
) extends ChannelInfo {
  override val state: ChannelState.NORMAL.type = ChannelState.NORMAL
}
```

I'm not sure if I understand what the point of `BaseChannelInfo` is though.

Christewart: Side note: let's make sure we use abstract classes where possible for older java compatability. If we wan't use scala 2.12 (which android doesn't support) I dont think we get a very strong guarantee that a trait is going to work on android

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/JsonReaders.scala
```diff
@@ -79,12 +80,74 @@ object JsonReaders {
...
+    }
+  }
+
   implicit val channelDescReads: Reads[ChannelDesc] = {
     Json.reads[ChannelDesc]
   }
 
-  implicit val channelInfoReads: Reads[ChannelInfo] = {
-    Json.reads[ChannelInfo]
+  implicit val normalChannelStateReads: Reads[ChannelState.NORMAL.type] =
```
Christewart: Maybe we should just make a `Reads[ChannelState]` and then use the `fromString` method in the `ChannelState` companion object to serialize.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -456,7 +451,7 @@ class EclairRpcClient(val instance: EclairInstance)(
   case class RpcError(code: Int, message: String)
   implicit val rpcErrorReads: Reads[RpcError] = Json.reads[RpcError]
 
-  private def parseResult[T](result: JsResult[T], json: JsValue): T = {
+  private def parseResult[T: TypeTag](result: JsResult[T], json: JsValue): T = {
```
nkohen: Why is a TypeTag necessary here? Is it just for the log message? (if so why not a ClassTag?)

torkelrogstad: Oh shoot that wasn't supposed to get in the commit...

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/JsonReaders.scala
torkelrogstad: That's already in the file further up, this is for validating a specific value

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
torkelrogstad: The point of `BaseChannelInfo` is to catch all responses where `state != "NORMAL"`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -24,13 +28,49 @@ case class GetInfoResult(
...
+sealed abstract class ChannelInfo {
+  def nodeId: NodeId
+  def channelId: FundedChannelId
+  def localMsat: MilliSatoshis
+  def remoteMsat: MilliSatoshis
+  def state: ChannelState
+}
+
+/**
+  * This represents the case where the channel is
```
Christewart:
```
/**
This represents the case where the [[ChannelState]] 
is undetermined
*/
```
right?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -310,6 +312,84 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+        EclairTestUtil
+          .openChannel(c1, c2, Satoshis(Int64(500000)), MilliSatoshis(500000)))
+    }
+
+    openChannel(firstFreshClient, secondFreshClient)
+    openChannel(thirdFreshClient, fourthFreshClient)
+
+    await(bitcoindRpcClient.generate(10))
+
+    def sendPayments(c1: EclairRpcClient, c2: EclairRpcClient) = {
```
Christewart: type annotation for return type please

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -310,6 +312,84 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+          c1.receive("foobar")
+            .flatMap(invoice => c2.send(invoice, MilliBitcoins(n))))
+
+      await(Future.sequence(payments))
+    }
+
+    sendPayments(firstFreshClient, secondFreshClient)
+    sendPayments(thirdFreshClient, fourthFreshClient)
+
+    def updateIsInChannels(channels: Seq[OpenChannelInfo])(
```
Christewart, line 351: could these me util methods that are moved to `EclairTestUtil`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
torkelrogstad: Unfortunately we have to do `[[org.bitcoins.core.protocol.ln.channel.ChannelState ChannelState]]`...
Adding it

PR #301, opened 2019-01-18
findroute RPC call tests/types

torkelrogstad:


core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -7,6 +7,17 @@ import scodec.bits.ByteVector
 
 case class ShortChannelId(u64: UInt64) extends NetworkElement {
   override def bytes: ByteVector = u64.bytes
+
+  def toHex: String = bytes.toHex
```
Christewart: Can't we just use `.hex` inside of `NetworkElement`?

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -9,14 +9,24 @@ import scala.math.BigDecimal.RoundingMode
...
   protected def underlying: BigInt
 
+  /**
+    * Output example:
+    * {{{
+    * > MilliSatoshis(10)
+    * 10 MilliSatoshis
+    * }}}
+    */
+  override def toString: String = s"$toLong MilliSatoshis"
```
Christewart: nit: i like `msat` better for the toString

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -320,6 +335,29 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+  it should "get a route to a node ID" in {
+    val (first, second, third, fourth) =
+      EclairRpcTestUtil.createNodeLink(bitcoindRpcClient)
+    clients ++= List(first, second, third, fourth)
+
+    for {
+      fourthInfo <- fourth.getInfo
+      route <- first.findRoute(fourthInfo.nodeId)
+    } yield assert(route.length === 4)
+    succeed
```
Christewart, line 356: the `succeed` at the end makes this unit test trivially true, is this just here for wip purposes now?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtilTest.scala
```diff
@@ -0,0 +1,91 @@
...
+import scala.collection.immutable.VectorBuilder
+
+class EclairRpcTestUtilTest extends AsyncFlatSpec with BeforeAndAfterAll {
+
+  private implicit val actorSystem: ActorSystem =
+    ActorSystem.create("EclairRpcTestUtilTest")
+
+  private val bitcoindRpc = BitcoindRpcTestUtil.startedBitcoindRpcClient()
+  private val clients: VectorBuilder[EclairRpcClient] =
+    new VectorBuilder[EclairRpcClient]
```
Christewart: `Vector.newBuilder[EclairRpcClient]` is the recommended way to create a new builder for any collection

testkit/src/main/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -215,8 +226,100 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
...
+      BitcoindRpcTestUtil.startedBitcoindRpcClient()
+    )
+
+    val (first, second) = createNodePair(Some(internalBitcoind))
+    val (third, fourth) = createNodePair(Some(internalBitcoind))
+
+    def open(
+        c1: EclairRpcClient,
+        c2: EclairRpcClient): Future[FundedChannelId] = {
+      openChannel(c1,
```
Christewart: nit: use named parameters

testkit/src/main/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -215,8 +226,100 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
...
+                  c2,
+                  channelAmount.toSatoshis,
+                  MilliSatoshis(channelAmount.toLong / 2))
+    }
+
+    EclairRpcTestUtil.connectLNNodes(second, third)
```
Christewart, line 249: nit: I would like to start using this syntax for things where we don't care about the return type

```scala

val _ = EclairRpcTestUtil.connectLNNodes(second, third)

```

testkit/src/main/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -215,8 +226,100 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
...
+
+    val openF = Future.sequence(
+      List(open(first, second), open(second, third), open(third, fourth)))
+
+    AsyncUtil.awaitCondition(() => openF.isCompleted)
```
Christewart: nit: 

```scala

val _ = AsyncUtil.awaitCondition(() => openF.isCompleted)

val _ = internalBitcoind.generate(3)

```

testkit/src/main/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -215,8 +226,100 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
...
+
+  /**
+    * $nodeLinkDoc
+    * @note Blocks the current thread
+    * @return A 4-tuple of the created nodes' respective
+    *         [[org.bitcoins.eclair.rpc.client.EclairRpcClient EclairRpcClient]]
+    */
+  def createNodeLink(
+      bitcoindRpcClient: BitcoindRpcClient,
+  )(implicit actorSystem: ActorSystem): (
```
Christewart: Perhaps make a case class instead of this 4 tuple?

```scala
case class EclairNodes4(n1: EclairRpcClient, n2: EclairRpcClient, n3: EclairRpcClient, n4: EclairRpcClient)
```

This also allows us to add helper methods to the case class if we need to in the future

Christewart: It might even be worth genericizing this to be `n` nodes

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
Christewart: If you remove the succeeds from the new unit tests they seem to fail

testkit/src/main/scala/org/bitcoins/eclair/rpc/EclairRpcTestUtil.scala
torkelrogstad: It's not possible to do multiple `val _ =  foo` in the same block :-( And if we do `val __ =  foo` we get warnings about unused params

PR #318, opened 2019-02-01
Fixes conversion bug when sending with Eclair RPC 

torkelrogstad:
Eclair expects msat integers in the send method, we were providing pBTC

### DIFF
--- a/project/Deps.scala
+++ b/project/Deps.scala
@@ -17,7 +17,7 @@ object Deps {
     val nativeLoaderV = "2.3.2"
     val typesafeConfigV = "1.3.3"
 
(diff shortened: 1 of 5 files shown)


project/Deps.scala
```diff
@@ -17,7 +17,7 @@ object Deps {
     val nativeLoaderV = "2.3.2"
     val typesafeConfigV = "1.3.3"
 
-    val bitcoinsV = "0.0.4.1-SNAPSHOT"
+    val bitcoinsV = "0a3bfb-1549032221832-SNAPSHOT"
```
torkelrogstad: This is published to JFrog

PR #323, opened 2019-02-04
Adds Ammonite REPL in test scope

torkelrogstad:
See changes in README.md for information on how to use.


Christewart:
Can we rebase this and perhaps create a `commonDeps` val in `project/Deps.scala` so we don't have to specify that dep everywhere?

PR #328, opened 2019-02-06
2019 02 04 testkit future refactor

Christewart:
This PR does a few things

1. Refactors the `start()` method on `BitcoindRpcClient` and `EclairRpcClient` to return a `Future[Unit]`. This future is completed when the RpcClient is fully started and you can send requests and expect responses in a timely manner.
2. Refactors `EclairRpcTestUtil` to return Futures than using hard blocking `Await`. This should theoretically boost performance of tests.
3. Remove 2 instances of `EclairRpcClient` that were being spun up in the eclair tests. It is very expensive to spin up new eclair nodes and we should avoid doing so unless there is a good reason. In most cases we can re-use already running eclair nodes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,6 +40,8 @@ lazy val commonSettings = List(
 
   scalacOptions in Test := testCompilerOpts,
 
+  testOptions in Test += Tests.Argument("-oF"),
+
   assemblyOption in assembly := (assemblyOption in assembly).value
     .copy(includeScala = false),
 
--- a/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
+++ b/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
@@ -11,10 +11,10 @@ import akka.stream.ActorMaterializer
 import akka.util.ByteString
 import org.bitcoins.core.crypto.Sha256Digest
 import org.bitcoins.core.currency.CurrencyUnit
-import org.bitcoins.core.protocol.ln.{LnInvoice, LnParams, ShortChannelId}
 import org.bitcoins.core.protocol.ln.channel.{ChannelId, FundedChannelId}
 import org.bitcoins.core.protocol.ln.currency.{LnCurrencyUnit, MilliSatoshis}
 import org.bitcoins.core.protocol.ln.node.NodeId
+import org.bitcoins.core.protocol.ln.{LnInvoice, LnParams, ShortChannelId}
 import org.bitcoins.core.protocol.script.ScriptPubKey
 import org.bitcoins.core.util.BitcoinSUtil
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
@@ -23,13 +23,14 @@ import org.bitcoins.eclair.rpc.config.EclairInstance
 import org.bitcoins.eclair.rpc.json._
 import org.bitcoins.eclair.rpc.network.{NodeUri, PeerState}
 import org.bitcoins.rpc.serializers.JsonReaders._
+import org.bitcoins.rpc.util.AsyncUtil
 import org.slf4j.LoggerFactory
 import play.api.libs.json._
 
 import scala.concurrent.duration.DurationInt
-import scala.concurrent.{Await, ExecutionContext, Future, Promise}
+import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.sys.process._
-import scala.util.{Failure, Properties, Success, Try}
+import scala.util.{Failure, Properties, Success}
 
 class EclairRpcClient(val instance: EclairInstance)(
     implicit system: ActorSystem)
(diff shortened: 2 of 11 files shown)


eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -560,22 +561,42 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-  def start(): Unit = {
+  def start(): Future[Unit] = {
 
-    if (process.isEmpty) {
-      val p = Process(
-        s"java -jar -Declair.datadir=${instance.authCredentials.datadir.get} $pathToEclairJar &")
-      val result = p.run()
-      logger.info(
-        s"Starting eclair with datadir ${instance.authCredentials.datadir.get}")
+    val _ = {
```
nkohen, line 572: Is there a reason this is wrapped in a block and not just on the top level?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -560,22 +561,42 @@ class EclairRpcClient(val instance: EclairInstance)(
 
   private var process: Option[Process] = None
 
-  def start(): Unit = {
+  def start(): Future[Unit] = {
```
nkohen, line 570: scaladoc? Should probably mention that the timeout is 1 minute

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -32,89 +32,150 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
-  val bitcoindRpcClient: BitcoindRpcClient =
+  val bitcoindRpcClientF: Future[BitcoindRpcClient] =
     BitcoindRpcTestUtil.startedBitcoindRpcClient()
 
-  lazy val EclairNodes4(firstClient, secondClient, thirdClient, fourthClient) = {
-    val EclairNodes4(first, second, third, fourth) =
-      EclairRpcTestUtil.createNodeLink(bitcoindRpcClient)
-    clients ++= List(first, second, third, fourth)
-    EclairNodes4(first, second, third, fourth)
+  lazy val eclairNodesF: Future[EclairNodes4] = {
```
nkohen: is there any case where this is never called or can we speed things up by making this eager since its wrapped in a `Future` now

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -32,89 +32,150 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+
+  lazy val secondClientF = eclairNodesF.map(_.c2)
+
+  lazy val thirdClientF = eclairNodesF.map(_.c3)
+
+  lazy val fourthClientF = eclairNodesF.map(_.c4)
+
+
+
+  lazy val clientOtherClientF = {
```
nkohen, line 70: comment as to what this is used for would be nice

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -32,89 +32,150 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+  lazy val otherClientF = clientOtherClientF.map(_._2)
+
   private val clients =
     Vector.newBuilder[EclairRpcClient]
 
   override def beforeAll(): Unit = {
     // make sure we have enough money open channels
-    bitcoindRpcClient.generate(200)
+    //not async safe
+    bitcoindRpcClientF.flatMap(_.generate(200))
```
nkohen: perhaps we want do define the `bitcoindRpcClientF` in such a way that the future only completes when the `generate` call is done? I.e. move the `flatMap` to `generate` up to the definition of this val?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -32,89 +32,150 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+  }
+
+  /** Executes a test with the default clients defined at the
+    * top of this file
+    * @param test
+    * @tparam T
+    * @return
+    */
+  def executeWithClientOtherClient[T](test: (EclairRpcClient,EclairRpcClient) => Future[T]): Future[T] = {
```
nkohen, line 89: I really like this setup!

nkohen: Do you think it might be a cleaner thing to do to extend AsyncFlatSpec to take in tests like this and have this be done by the test runner? I'm not saying we need to but just a thought

rpc/src/main/scala/org/bitcoins/rpc/client/BitcoindRpcClient.scala
```diff
@@ -1190,13 +1188,32 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
         HttpCredentials.createBasicHttpCredentials(username, password))
   }
 
-  def start(): String = {
+  def start(): Future[Unit] = {
```
nkohen, line 1196: scaladoc?

build.sbt
```diff
@@ -152,6 +154,7 @@ lazy val rpc = project
   .in(file("rpc"))
   .enablePlugins()
   .settings(commonSettings: _*)
+  .settings(testOptions in Test += Tests.Argument("-oF"))
```
torkelrogstad: Isn't this line redundant, because of line 43?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -560,22 +561,42 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-    if (process.isEmpty) {
-      val p = Process(
-        s"java -jar -Declair.datadir=${instance.authCredentials.datadir.get} $pathToEclairJar &")
-      val result = p.run()
-      logger.info(
-        s"Starting eclair with datadir ${instance.authCredentials.datadir.get}")
+    val _ = {
+      if (process.isEmpty) {
+        val p = Process(
+          s"java -jar -Declair.datadir=${instance.authCredentials.datadir.get} $pathToEclairJar &")
```
torkelrogstad, line 578: The error message that potentially triggers on `.get` is somewhat cryptic. Would it be better to  add a `require` statement or something similar on the line above?

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -32,89 +32,150 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    }
   }
 
   behavior of "RpcClient"
 
   it should "be able to open and close a channel" in {
 
-    val changeAddrF = bitcoindRpcClient.getNewAddress()
+    val changeAddrF = bitcoindRpcClientF.flatMap(_.getNewAddress())
     val result: Future[Assertion] = {
```
torkelrogstad, line 113: This intermediate val can be deleted as it is just one big block which is returned at the end. IMO it just adds unnecessary nesting

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
Christewart: I think it logically groups all of these `Unit` together, as that is the return type of `_`

eclair-rpc/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
Christewart: That would probably be a good idea -- we need to really create a common trait for all tests in bitcoin-s (or at least the various sub projects).

PR #329, opened 2019-02-07
Add BasicArithmetic trait

torkelrogstad:


core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -62,27 +63,63 @@ sealed abstract class MilliSatoshis extends NetworkElement {
...
+    minus(ms)
+  }
+
+  def minus(ms: MilliSatoshis): Try[MilliSatoshis] = {
+    Try {
+      minusExc(ms)
+    }
+  }
+
+  def -!(ms: MilliSatoshis): MilliSatoshis = {
```
torkelrogstad: What do you think of this syntax? It's meant to be an unsafe version of `-`

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -62,27 +63,63 @@ sealed abstract class MilliSatoshis extends NetworkElement {
   }
 
   def ==(ms: MilliSatoshis): Boolean = {
-    toLnCurrencyUnit == ms.toLnCurrencyUnit
+    toBigInt == ms.toBigInt
```
torkelrogstad, line 67: Errors were appearing because there's `require` statements in `PicoBitcoin` etc. IMO there's no real reason to do these comparisons via `toLnCurrencyUnit`, as it's all `BigInt`s all the way down anyways

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
Christewart: I don't really like it.

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -62,27 +63,63 @@ sealed abstract class MilliSatoshis extends NetworkElement {
...
+    MilliSatoshis(toBigInt - ms.toBigInt)
+  }
+
+  /**
+    * Acts similarly to how integer division works. I.e.:
+    * {{{
+    *   13 msat / 4 = 3 msat
+    * }}}
+    */
+  def /(divisor: BigInt): MilliSatoshis = {
```
Christewart: I think this is dangerous and we shouldn't support it in the library. It is too easy for library users to screw this up.

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -62,27 +63,63 @@ sealed abstract class MilliSatoshis extends NetworkElement {
...
   def <=(ms: MilliSatoshis): Boolean = {
-    toLnCurrencyUnit <= ms.toLnCurrencyUnit
+    toBigInt <= ms.toBigInt
+  }
+
+  def +(ms: MilliSatoshis): MilliSatoshis = {
+    MilliSatoshis(toBigInt + ms.toBigInt)
+  }
+
+  def -(ms: MilliSatoshis): Try[MilliSatoshis] = {
```
Christewart: we aren't being consistent here with the return types, should be `MilliSatoshis` rather than `Try[MilliSatoshis]`

Christewart:
I think an important thing here is that we stay with the convention we have already layed out in `CurrencyUnit`, i.e. the operator returns a type `T` rather than a `Try[T]`. We could make a parallel API that returns a `Try[T]`, but I don't think we should change convention for now on the `+`,`-`,`*` operators etc

torkelrogstad:
The vibe I'm getting here: 

- Remove divison operator
- Make `-` operator unsafe

I think it would make sense to have a safe subtraction operator/method that returns a `Try[MilliSatoshi]`. `minusSafe`, or something like that. What's your opinion on that?

Also, something I've been thinking about, while we're discussing change to `MilliSatoshi`. I really like `DurationInt` and `DurationLong` from `scala.concurrent.duration` AKA being able to write `1.millsecond`. Are you opposed to adding something similar to Bitcoin-S currency units?

Christewart:
I'm not opposed to that, I just think we should make sure we can use the library without using those postfix methods. 

```scala

abstract class BasicArithmetic[+N] {

def +(n: N): N 

def +t(n: N) : Try[N] = Try(this + n)

def *(n : N): N

def *t(n: N): Try[N] = Try(this * n)

}
```

core/src/main/scala/org/bitcoins/core/number/BasicArithmetic.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def +(n: N): N
+
+  /**
+    * Some classes have restrictions on upper bounds
+    * for it's underlying value. This might cause the `+`
+    * operator to throw. This method wraps it in a `Try`
+    * block.
+    */
+  def safeAdd(n: N): Try[N] = Try { this + n }
```
Christewart: I think this is important, we should probably do `addSafe` rather than `safeAdd`

If i am using autocomplete on my text editor I will type `a` for `add` and not see the `safeAdd` method. I think it is a little thing that is important when designing a library.

Christewart: Same thing with others

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
```diff
@@ -28,6 +25,50 @@ class MilliSatoshisTest extends FlatSpec with MustMatchers {
...
+        val added = first + second
+        val subtracted = added - second
+        assert(subtracted == first)
+    }
+  }
+
+  it must "subtract msats" in {
+    PropertyChecks.forAll(LnCurrencyUnitGen.milliSatoshisPair) {
+      case (first, second) =>
+        val subtracted = first safeSubtract second
```
Christewart: use `.` notation when dealing with named methods.

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
```diff
@@ -28,6 +25,50 @@ class MilliSatoshisTest extends FlatSpec with MustMatchers {
...
+        val isPositive = (first.toBigInt - second.toBigInt) >= 0
+
+        assert(subtracted.isSuccess == isPositive)
+        if (subtracted.isSuccess) {
+          val underlyingCalc = first.toBigInt - second.toBigInt
+          assert(subtracted.get.toBigInt == underlyingCalc)
+        }
+    }
+  }
+
```
Christewart, line 84: nit: test coverage is going to take a hit since safe stuff isn't tested

Christewart (commented):
Is it trivially to extend this on `NumberType`, `CurrencyUnit`, `LnCurrencyUnit`? If so, we should probably do this in this PR.

core-test/src/test/scala/org/bitcoins/core/number/BasicArithmeticSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+      val unsafe = num * i
+      val safe = num.multiplySafe(i)
+      assert(safe.toOption.contains(unsafe))
+    }
+  }
+
+  it must "multiply safely and unsafely with itself" in {
+    PropertyChecks.forAll(numWrapperGen, numWrapperGen) { (first, second) =>
+      val unsafe = first * second
+      val safe = first
```
Christewart: no check here?

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnitSpec.scala
```diff
@@ -58,6 +63,20 @@ class LnCurrencyUnitSpec extends Properties("LnCurrencyUnitSpec") {
         else Try(num1 * num2).isFailure
     }
 
+  property("Multiply a LnCurrencyUnit value with an int") =
+    Prop.forAll(lnCurrWithInt) {
+      case (ln, int) =>
+        val safeProduct = ln.multiplySafe(int)
+        val underlyingProduct = ln.toBigInt * int
+        if (underlyingProduct <= PicoBitcoins.min.toBigInt && underlyingProduct >= PicoBitcoins.max.toBigInt) {
+          safeProduct.toOption.exists {
```
Christewart: I think 
```scala
safeProduct.get.toBigInt == underlyingProduct
```

is better

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
```diff
@@ -28,6 +25,63 @@ class MilliSatoshisTest extends FlatSpec with MustMatchers {
...
+  private val msatWithNum = for {
+    msat <- LnCurrencyUnitGen.milliSatoshis
+    num <- NumberGenerator.bigIntsUInt64Range.filter(_ > 0)
+  } yield (msat, num)
+
+  it must "multiply millisatoshis with an int" in {
+    PropertyChecks.forAll(msatWithNum) {
+      case (msat, bigint) =>
+        val underlyingCalc = msat.toBigInt * bigint
+        assert((msat * bigint).toBigInt == underlyingCalc)
```
Christewart, line 45: shouldn't this hit an assertion if `msat * bigInt` results in too big of a number? I forget if `msat` is bound to a protocol type (UInt32, UInt64 etc)

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
```diff
@@ -28,6 +25,63 @@ class MilliSatoshisTest extends FlatSpec with MustMatchers {
...
+    }
+  }
+
+  it must "multiply millisatoshis with itself" in {
+    PropertyChecks.forAll(LnCurrencyUnitGen.milliSatoshisPair) {
+      case (first, second) =>
+        val safe = first.multiplySafe(second)
+        val unsafe = first * second
```
Christewart, line 53: Same comment as above

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshisTest.scala
torkelrogstad: msats are only lower bound, with bigint as underlying

core-test/src/test/scala/org/bitcoins/core/currency/CurrencyUnitSpec.scala
```diff
@@ -67,6 +72,21 @@ class CurrencyUnitSpec extends Properties("CurrencyUnitSpec") {
...
   }
 
+  property("Multiply a satoshi value with an int") = Prop.forAll(
+    satoshiWithInt
+  ) {
+    case (sat, int) =>
+      val safeProduct = sat.multiplySafe(int)
+      val underlyingProduct = sat.toBigInt * int
+      if (underlyingProduct < Satoshis.max.toBigInt && underlyingProduct > Satoshis.min.toBigInt) {
+        assert(safeProduct.isSuccess)
```
Christewart, line 82: I don't think this line is necessary, and isn't the `assert` you are familiar with in scalatest. 

Just a note, these tests are run using scalacheck not scalatest. Scalatest has a dependency on scalacheck, which means you can do this stuff in scalatest too -- which is what I am trying to move towards in #335

core-test/src/test/scala/org/bitcoins/core/number/BasicArithmeticSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+    int <- NumberGenerator.bigInts
+  } yield NumWrapper(int)
+
+  behavior of "BasicArithmetic"
+
+  it must "multiply safely and unsafely with an int" in {
+    PropertyChecks.forAll(NumberGenerator.bigInts, numWrapperGen) { (i, num) =>
+      val unsafe = num * i
+      val safe = num.multiplySafe(i)
+      assert(safe.toOption.contains(unsafe))
```
Christewart, line 36: I prefer 

```scala
safe.get == unsafe
```
It is just tests after all, so i'm not as concerned about being type safe. Worst case is it throws a `NoSuchElementException`.

PR #335, opened 2019-02-14
Add bitcoin-s unit test class, all unit tests going forward should ex…

Christewart:
…tend this

Creates a common test abstract class for bitcoin unit tests. This PR does not go through the code base and extend this abstract class, it just extends it in one called `LnInvoiceUnitTest`. 

One interesting thing about property based testing is that the longer the code lives, the surer you are that it is correct. This PR also introduces the concept of "old code" and "new code". New code should be executed more often in property based tests, while old code can be executed less times. 

See this talk by Bill Venners (creator of ScalaTest) for more info on this:

https://youtu.be/lKtg-CDVDsI?t=2293

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/util/BitcoinSUnitTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/util/BitcoinSUnitTest.scala
@@ -1,0 +1,57 @@
+package org.bitcoins.core.util
+
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{FlatSpec, MustMatchers}
+import org.slf4j.LoggerFactory
+
+/** A wrapper for boiler plate testing procesures in bitcoin-s */
+abstract class BitcoinSUnitTest extends FlatSpec with MustMatchers with PropertyChecks {
+
+  lazy val logger = LoggerFactory.getLogger(getClass)
+
+  /** The configuration for property based tests in our testing suite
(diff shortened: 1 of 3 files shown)


torkelrogstad:
From first CI job: `[info] Coverage is above minimum [90.01% > 90.0%]` :sunglasses: 

LGTM!

core-test/src/test/scala/org/bitcoins/core/util/BitcoinSUnitTest.scala
```diff
@@ -0,0 +1,57 @@
+package org.bitcoins.core.util
+
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{FlatSpec, MustMatchers}
+import org.slf4j.LoggerFactory
+
+/** A wrapper for boiler plate testing procesures in bitcoin-s */
+abstract class BitcoinSUnitTest extends FlatSpec with MustMatchers with PropertyChecks {
+
+  lazy val logger = LoggerFactory.getLogger(getClass)
```
nkohen, line 10: Don't we have a trait in Bitcoin-S for logging?

nkohen (commented):
Other than the one comment, LGTM

core-test/src/test/scala/org/bitcoins/core/util/BitcoinSUnitTest.scala
Christewart: It is an abstract class, so i'm just going to create a val here for now.

PR #337, opened 2019-02-15
2019 02 14 fix digital signature bug ln invoice

### LINKED ISSUE #297: Invalid digital signature on invoice returned from eclair
```
lnbcrtpwrlzq9pp5phdlf0s83aw96rymm03d4awyjehtlj5yfkhml79ytv8yv8c3qlasdqh2fjkvatwvssxjmnkda5kxegxqrrssqr4xl2lk0325hkxy03pvsrq3fjma7elqm2lczc5mqef38vjw4ruyz57p392xn6ezry2v460yukz675z45km7evqzjvkl06sthgvgjssq3ztaep
```

Christewart:
This fixes #297 

The root problem was making sure we were dealing with unsigned numbers inside of `BigInt` when reading a reading a `ECDigitalSignature` from a raw `r,s` value. Depending on how large the number was, previously it could be interpreted as a negative number which would produce an invalid digital signature when reading the invoice from another source.

https://github.com/bitcoin-s/bitcoin-s-core/pull/337/files#diff-ce7c7a075ef8c7f783017f3a133b7cd7R154

This PR also adds more tests for `LnUtil`, `LnInvoiceSignature`, and `LnInvoice`.


nkohen (commented):
I'm not going to pretend I really know what's going on here but assuming tests pass I didn't see any code things that looked wrong

core-test/src/test/scala/org/bitcoins/core/protocol/Bech32Test.scala
```diff
@@ -134,4 +134,16 @@ class Bech32Test extends FlatSpec with MustMatchers {
...
   }
+
+
+  it must "encode from 8 bit to 5 bit and back" in {
+    forAll(NumberGenerator.uInt8s) { u8s =>
+      val u5s = Bech32.from8bitTo5bit(u8s.toVector)
+
+      val u8sAgain = Bech32.from5bitTo8bit(u5s)
+
+      u8s == u8sAgain
```
Christewart: `assert(u8s == u8sAgain)`

core-test/src/test/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnitSpec.scala
```diff
@@ -58,11 +58,6 @@ class LnCurrencyUnitSpec extends Properties("LnCurrencyUnitSpec") {
         else Try(num1 * num2).isFailure
     }
 
-  property("Convert negative LnCurrencyUnit value to Satoshis") =
```
Christewart, line 79: removed this because it doesn't really make sense to have a negative ln currency unit

core/src/main/scala/org/bitcoins/core/protocol/ln/LnInvoice.scala
```diff
@@ -99,7 +99,7 @@ sealed abstract class LnInvoice {
   override def toString: String = {
     val b = new StringBuilder
     b.append(hrp.toString)
-    b.append(Bech32.separator)
+    b.append('1')
```
Christewart, line 102: I don't know why i had to use the hard coded `1` but the initilization problem persisted if I didn't add it. When invoices were being serialized in some cases the separator would be missing

core/src/main/scala/org/bitcoins/core/protocol/ln/util/LnUtil.scala
```diff
@@ -16,19 +17,19 @@ trait LnUtil {
     * [[https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#examples Bolt11]]
     * for examples.
     */
-  def createDataLength(bech32String: String): Vector[UInt5] = {
+  def createDataLength(bech32String: String): List[UInt5] = {
```
Christewart, line 20: Most changes in this file are because `List` has way better performance for construction and destruction of collections one by one. 

http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#lists-vs-vectors

```
Constructing a List item-by-item is 5-15x (!) faster than constructing a Vector item-by-item.

De-constructing a List item-by-item using .tail is 50-60x (!) faster than de-constructing a Vector item-by-item using .tail
```

core/src/main/scala/org/bitcoins/core/protocol/ln/util/LnUtil.scala
```diff
@@ -40,42 +41,48 @@ trait LnUtil {
...
+  final def decodeNumber(list: List[UInt5], accum: BigInt = 0): BigInt = {
 
-    if (vector.isEmpty) accum
-    else if (vector.size == 1) {
-      decodeNumber(vector.tail, vector.head.toInt + accum)
-    } else {
-      val n = BigInt(32).pow(vector.size - 1)
-      val newAccum = vector.head.toBigInt * n + accum
-      decodeNumber(vector.tail, newAccum.toLong)
+    val b32 = BigInt(32)
```
Christewart: Should be moved back inside the ` h :: t` case otherwise it is pointless construction every time we decode

PR #343, opened 2019-02-18
Reduce number of requests we send to eclair in testkits, add commandN…

Christewart:
…ame param to parseResult in EclairRpcClient

Reduces the number of requests we make to eclair during tests, adds command name param to parseResult in EclairRpcClient to allow you to see what command failed,


eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -476,24 +476,24 @@ class EclairRpcClient(val instance: EclairInstance)(
...
     result match {
       case res: JsSuccess[T] =>
         res.value
       case res: JsError =>
         (json \ errorKey).validate[RpcError] match {
           case err: JsSuccess[RpcError] =>
-            logger.error(s"Error ${err.value.code}: ${err.value.message}")
-            throw new RuntimeException(
-              s"Error ${err.value.code}: ${err.value.message}")
+            val errMsg = s"Error for command=${commandName} ${instance.authCredentials.datadir.map(d => s"datadir=${d}").getOrElse("")}, ${err.value.code}=${err.value.message}"
```
torkelrogstad: Would something like this be better?

```suggestion
 val datadirMsg = instance.authCredentials.datadir
              .map(d => s"datadir=$d")
              .getOrElse("")
            val eclairErrCodeMsg = s"code=${err.value.code}"
            val eclairErrMsg =  s"message=${err.value.message}"
            val errMsg =
              s"Error for command=$commandName $datadirMsg, $eclairErrCodeMsg, $eclairErrMsg"
```

.jvmopts
```diff
@@ -1,2 +1,2 @@
--Xmx1024m
+-Xmx2048m
```
torkelrogstad: When I pulled down this PR and opened it in IntelliJ, my laptop grinded to an absolute halt. Had to reboot. I'm no expert in Java memory allocation, but I once tweaked the default IntelliJ memory parameters, causing IntelliJ to several times a day hog every single byte available on my machine. Do you think this option change might behave badly in some situations? 

It might be that my laptop crashed for some other reason as well.

torkelrogstad:
It looks like you forgot to run Scalafmt on this PR, @Christewart. We've briefly talked about this before, but it is surprisingly tricky to get Scalafmt to run on save through Intellij. On my machine I have to open up the Scalafmt settings in Intellij before it "remembers" to format on save. One solution to this is adding Scalafmt as a pre-commit hook. 

Installation instructions for [Scalafmt](https://scalameta.org/scalafmt/docs/installation.html)  
Installation instructions for [Couriser](https://get-coursier.io/docs/cli-overview), the recommended tool to install Scalafmt with  

The Scalafmt CLI could then be invoked with `scalafmt --diff-branch=$(current_branch)` (assuming you have a bash function called `current_branch` that prints the current branch). `--diff-branch` causes it to only format changed files. 

What do you think of this approach? If you think it makes sense we could add something about it in `CONTRIBUTING.md`. 

Edit: 

Here's a suggestion for `current_branch`, by the way: 

```bash
git symbolic-ref HEAD | cut --delimiter '/' --fields 3,4
```

torkelrogstad:
Speaking of Scalafmt, you should read the section on rewrite rules! https://scalameta.org/scalafmt/docs/configuration.html#rewrite-rules. Lots of rules there that seems nice:-)

PR #353, opened 2019-02-20
BIP39 support

torkelrogstad:
Implements: 
 - [x] constructing entropy for creating mnemonic
 - [x] getting mnemonic from entropy
 - [x] getting entropy from mnemonic
 - [x] create seed from mnemonic - I have tried to implement it, but tests are failing
 - [x] import all test vectors from Trezor tests, see link in `MnemonicCodeTest`

Design question: Right now all `fromXXX` or `toXXX` methods on mnemonic codes and BIP39 seeds are unsafe, returning `X` instead of `Option[X]`/`Try[X]`. `fromBytes` has to be unsafe, as that's the signature in `Factory`. The other methods can be as safe as we want them, though. Any thoughts on this?


Christewart:
Link to BIP39: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,951 @@
...
+import scodec.bits.{BitVector, ByteVector, HexStringSyntax, BinStringSyntax}
+
+import scala.util.Try
+
+/**
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#test-vectors Trezor]]
+  * test vectors
+  */
+class MnemonicCodeTest extends BitcoinSUnitTest {
+  behavior of "MnemonicCode"
```
Christewart, line 17: Since this is new code, let's override the generatorDriveConfig to get more iterations on the property based tests. The old version only does 10 iterations, while the "new code" config does 100 iterations. You can do this like this:

```scala
  override implicit val generatorDrivenConfig: PropertyCheckConfiguration = {
    generatorDrivenConfigNewCode
  }
```

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,951 @@
...
+      // is "satoshis", in that case we woulnd't produce a
+      // bad phrase
+      val newLastWord = if (lastWord == "satoshi") "absurd" else "satoshi"
+      val fromWordsT = Try {
+        MnemonicCode.fromWords(words.dropRight(1) :+ newLastWord)
+      }
+      fromWordsT.isFailure
+    }
+  }
+
```
Christewart, line 108: I would like to see a property based test for the entropy generators. Something of the flavor 

```
If I call `getEntropyXBits` twice I do NOT get the same value back. This should be tested for all of the entropy generators. 

```

Here is the equivalent for `ECPrivateKey` 

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core-test/src/test/scala/org/bitcoins/core/crypto/ECPrivateKeySpec.scala#L29

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,951 @@
...
+      val entropy = code.toEntropy
+      val newCode = MnemonicCode.fromEntropy(entropy)
+      code == newCode
+    }
+  }
+
+  it must "fail on a seed phrase with bad checksum" in {
+    Prop.forAll(CryptoGenerators.mnemonicPhrase) { words =>
+      val lastWord = words.reverse.head
+      // there is a chance the last word in the seed phrase
```
Christewart: I don't understand what is happening here. Why would this matter?

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,951 @@
...
+      // bad phrase
+      val newLastWord = if (lastWord == "satoshi") "absurd" else "satoshi"
+      val fromWordsT = Try {
+        MnemonicCode.fromWords(words.dropRight(1) :+ newLastWord)
+      }
+      fromWordsT.isFailure
+    }
+  }
+
+  private def testTrezorVector(
```
Christewart: Scaladoc with link to where these test vectors are coming from

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
```diff
@@ -0,0 +1,69 @@
...
+    * seed from a sequence of bytes. Must be between 16 and 64 bytes.
+    */
+  override def fromBytes(bytes: ByteVector): BIP39Seed =
+    BIP39SeedImpl(bytes)
+
+  private val MIN_SEED_LENGTH_BYTES = 16
+
+  private val MAX_SEED_LENGTH_BYTES = 64
+
+  private case class BIP39SeedImpl(bytes: ByteVector) extends BIP39Seed {
```
Christewart: nit: I usually like to have the `Impl` at the top of the companion object

core/src/main/scala/org/bitcoins/core/crypto/ExtKeyVersion.scala
```diff
@@ -31,4 +15,20 @@ object ExtKeyVersion {
 
   def apply(bytes: ByteVector): Option[ExtKeyVersion] =
     all.find(_.bytes == bytes)
+
+  case object MainNetPub extends ExtKeyVersion {
+    override def bytes = ByteVector(0x04, 0x88, 0xb2, 0x1E)
```
Christewart: Nit: I think we should change these to 

```
override val bytes = ...
```

This means the byte vector doesn't get re-computed every time.

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+import java.security.SecureRandom
+
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.{CryptoUtil, Factory}
+import scodec.bits.{BitVector, ByteVector}
+
+import scala.annotation.tailrec
+import scala.io.Source
+
+sealed abstract class MnemonicCode extends NetworkElement {
```
Christewart: Scaladoc

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  private val VALID_LENGTHS = Vector(12, 15, 18, 21, 24)
```
Christewart: Scaladoc indicating what these valid lengths mean, I"m assuming they corresponding to the amount of words in a BIP39 seed?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+
+    val hashedEntropy = CryptoUtil.sha256(entropyNoChecksum).bytes.toBitVector
+    val checksum = hashedEntropy.take(codeInfo.checksumLength)
+
+    entropyNoChecksum ++ checksum == entropyWithChecksum
+
+  }
+
+  def words: Vector[String]
+  override def bytes: ByteVector = ByteVector(words.mkString(" ").getBytes)
```
Christewart: Is this right? Is the `words.mkString(" ").getBytes` going to be compatible with 

```
UTF-8 using Normalization Form Compatibility Decomposition (NFKD). 
```

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+    * Generates a mnemonic code from the given entropy bytes
+    */
+  def fromEntropy(entropy: ByteVector): MnemonicCode =
+    fromEntropy(entropy.bits)
+
+  /**
+    * Generates a mnemonic code from the given entropy bits
+    */
+  def fromEntropy(entropy: BitVector): MnemonicCode = {
+    val MIN_BITS = 128
```
Christewart: move this out of this method perhaps?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+    require(
+      entropy.length >= MIN_BITS,
+      // $COVERAGE-OFF$
```
Christewart: I don't like the 
```scala
// $COVERAGE-OFF 
```
here. We should have test cases that explicitly test this.

Christewart (commented):
Partial review

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,951 @@
...
+
+    val bits256 = MnemonicCode.getEntropy256Bits
+    MnemonicCode.fromEntropy(bits256).words.length must be(24)
+  }
+
+  it must "fail to generate mnemonics from entropy of bad length" in {
+    def testEntropyFailure(entropy: BitVector): Assertion = {
+      val short = entropy.drop(1)
+      val long = entropy ++ bin"1"
+      Try(MnemonicCode.fromEntropy(short)).isFailure must be(true)
```
nkohen: I think using `assertThrows[ExceptionTypeHere]` is nice for things like this because then it will fail the test if you don't get the expected failure

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
nkohen: I think this is resolved if you do something like your bad length test and just alter the correct checksum?

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
nkohen: It also might be nice if you put the data for these tests (the long Vectors) in a separate file in this package just so that the tests stay nice

core/src/main/scala/org/bitcoins/core/crypto/ExtKeyVersion.scala
Christewart: Another nit: I believe we should mark these as `final` according to https://nrinaudo.github.io/scala-best-practices/adts/final_case_objects.html

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -200,21 +204,23 @@ object ExtPrivateKey extends Factory[ExtPrivateKey] {
...
-      seedOpt: Option[ByteVector] = None): ExtPrivateKey = {
-    val seed = seedOpt match {
-      case Some(bytes) => bytes
-      case None        => ECPrivateKey().bytes
+      seedOpt: Option[BIP39Seed] = None): ExtPrivateKey = {
+    val seed: ByteVector = seedOpt match {
+      case Some(s) => s.bytes
+      case None    => ECPrivateKey().bytes
     }
+    val key = ByteVector.encodeAscii("Bitcoin seed").right.get
```
nkohen: Can this string be anything? If it is specified someplace a link would be nice

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+    */
+  def toEntropyWithChecksum: BitVector = {
+
+    def bitsForWord(word: String): BitVector = {
+
+      // tail-recursive binary search for word in word list
+      // stolen from https://codereview.stackexchange.com/questions/161887/re-inventing-the-wheel-binary-search-in-scala
+      import MnemonicCode.{ENGLISH_WORDS => wordlist}
+      @tailrec
+      def search(
```
nkohen, line 77: Any chance we have some place for utility functions like this in core?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  def words: Vector[String]
+  override def bytes: ByteVector = ByteVector(words.mkString(" ").getBytes)
+
+  /**
+    * Returns the entropy initially provided to construct
+    * this mnemonic code
+    */
+  def toEntropy: BitVector = {
```
Christewart: Should we make this method publicly available?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  private lazy val randomGenerator: SecureRandom = new SecureRandom
+
+  /**
+    * Generates the specified bits of entropy
+    */
+  private def getEntropy(bits: Int): BitVector = {
+    val byteArray: Array[Byte] = new Array[Byte](bits / 8)
+    randomGenerator.nextBytes(byteArray)
+    val bitVector = BitVector(byteArray)
```
Christewart, line 207:
```scala
require(bits == bitVector.length, s"Did not generate enough bits of entropy! Exepcted=$bits, actual=${bitVector.length}")
```

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+    val bitGroups = entropyWithChecksum.grouped(BIT_GROUP_LENGTH)
+    val words = bitGroups.map { group =>
+      val index = group.toInt(signed = false)
+      ENGLISH_WORDS(index)
+    }.toVector
+
+    MnemonicCodeImpl(words)
+  }
+
+  private lazy val randomGenerator: SecureRandom = new SecureRandom
```
Christewart: I believe it is ok to cache this from reading the javadoc on `SecureRandom`. But I don't really like the idea of this sitting in memory for a long amount of time. In practice, it shouldn't hurt to make this a `def` no? I would be even more in favor or just moving it into `getEntropy` so it is in the smallest scope possible.

Christewart: I want our code that is generating entropy to have the smallest scope possible to make it harder for people that are modifying the code later to do anything weird with the PRNG

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  private lazy val ENGLISH_WORDS: Vector[String] = {
+    val ENGLISH_WORDS_FILE = "/bip39-wordlists/english.txt"
+    val resourceStream = getClass.getResourceAsStream(ENGLISH_WORDS_FILE)
+    val source = Source.fromInputStream(resourceStream)
+
+    val lines = source.getLines
+    val linesVec = lines.toVector
+    linesVec
+  } ensuring (words =>
```
Christewart: Nice!

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+  /**
+    * Creates a mnemonic code from the provided words
+    */
+  def fromWords(words: Vector[String]): MnemonicCode = MnemonicCodeImpl(words)
+
+  /**
+    * Creates a mnemonic code from the provided bytes
+    */
+  override def fromBytes(bytes: ByteVector): MnemonicCode = {
+    val words = bytes.toHex.split(" ").toVector
```
Christewart: This seems weird to me. When would we ever have the words in hex?

core/src/main/scala/org/bitcoins/core/crypto/ExtKeyVersion.scala
```diff
@@ -6,23 +6,7 @@ import scodec.bits.ByteVector
...
-  override def bytes = ByteVector(0x04, 0x88, 0xAD, 0xE4)
-}
-
-case object TestNet3Pub extends ExtKeyVersion {
-  override def bytes = ByteVector(0x04, 0x35, 0x87, 0xCF)
-}
-
-case object TestNet3Priv extends ExtKeyVersion {
-  override def bytes = ByteVector(0x04, 0x35, 0x83, 0x94)
+  override def hex: String = BitcoinSUtil.encodeHex(bytes)
```
Christewart: I think this can be deleted

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
```diff
@@ -0,0 +1,69 @@
...
+    * Generates a [[https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki BIP32]]
+    * seed from a mnemonic code. An optional password can be supplied.
+    * @param password Defaults to the empty string
+    */
+  def fromMnemonic(mnemonic: MnemonicCode, password: String = ""): BIP39Seed = {
+    val salt = s"mnemonic$password"
+
+    val words = mnemonic.words.mkString(" ")
+
+    val keySpec = new PBEKeySpec(
```
Christewart, line 53: named arguments here please

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
torkelrogstad: The last N bits (depending on how long your seed phrase is) is a checksum. In the case of a 12 word phrase, the last word is a checksum. So tweaking the last word should result in a failed phrase

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
torkelrogstad: A use case is that you have a English phrase, and you want to convert it to a Spanish. Which I think is a legit use case

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
torkelrogstad: Moved it into `getEntropy`

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
torkelrogstad: It's a java method, cannot have named arguments :-(

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -200,21 +204,23 @@ object ExtPrivateKey extends Factory[ExtPrivateKey] {
     */
   def apply(
       version: ExtKeyVersion,
-      seedOpt: Option[ByteVector] = None): ExtPrivateKey = {
-    val seed = seedOpt match {
-      case Some(bytes) => bytes
-      case None        => ECPrivateKey().bytes
+      seedOpt: Option[BIP39Seed] = None): ExtPrivateKey = {
```
Christewart: let's not change this method and add a `fromBIP39Seed` method

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,247 @@
...
+      case 21 => MnemonicCodeInfo(224, 7, 231, 21)
+      case 24 => MnemonicCodeInfo(256, 8, 264, 24)
+      case _: Int =>
+        throw new RuntimeException(s"Bad mnemonic code length: ${words.length}")
+    }
+
+  private lazy val ENGLISH_WORDS: Vector[String] = {
+    val ENGLISH_WORDS_FILE = "/bip39-wordlists/english.txt"
+    val resourceStream = getClass.getResourceAsStream(ENGLISH_WORDS_FILE)
+    val source = Source.fromInputStream(resourceStream)
```
torkelrogstad, line 266: Do I need to close this source?

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,289 @@
...
+
+  it must "have serialization symmetry - entropy" in {
+    forAll(CryptoGenerators.mnemonicCode) { code =>
+      val entropy = code.toEntropy
+      val newCode = MnemonicCode.fromEntropy(entropy)
+      assert(code == newCode)
+    }
+  }
+
+  it must "fail on a seed phrase with bad checksum" in {
```
torkelrogstad, line 128: I can't figure this one out... For some reason I can't seem to construct the correct invalid word to add at the end. Note that I've tested all the falsifying cases against https://iancoleman.io/bip39/, and the phrases the implementation claims are correct are correct according to that website. So I believe this is an issue with the test, not the implementation

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,289 @@
...
+    implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny
+
+    /**
+      * There is a chance that a new last word would still produce
+      * a valid word phrase. Here we make sure we get an invalid
+      * phrase. This is done by checking how many bits of checksum
+      * there are, and then recursing until we find a word that
+      * doesn't match the initial checksum
+      */
+    def getInvalidLastWord(words: Vector[String]) = {
```
nkohen: Return types on the def here and the ones defined inside would be nice :)

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
nkohen: It can't hurt, I would do it

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
```diff
@@ -0,0 +1,289 @@
...
+      val entropy = code.toEntropy
+      val newCode = MnemonicCode.fromEntropy(entropy)
+      assert(code == newCode)
+    }
+  }
+
+  it must "fail on a seed phrase with bad checksum" in {
+    // scalacheck shrinks our twelve words without considering
+    // that not any twelve words will do
+    implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny
```
Christewart: I think there is a method called `forAllNoShink()` you can use instead of this

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,257 @@
...
+      entropy.length % ENTROPY_MULTIPLE == 0,
+      s"Entropy must be a multiple of $ENTROPY_MULTIPLE bits!"
+    )
+
+    val ENTROPY_CHECKSUM_DIVISOR = 32
+    val checksumLength = entropy.length / ENTROPY_CHECKSUM_DIVISOR
+
+    val hashedEntropy = CryptoUtil.sha256(entropy)
+    val hashedEntropyBits = hashedEntropy.bytes.toBitVector
+    val checkSum = hashedEntropyBits.take(checksumLength)
```
Christewart, line 179: Shouldn't we be validating the checksum here?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+    * Generates the specified bits of entropy
+    */
+  private def getEntropy(bits: Int): BitVector = {
+    val randomGenerator: SecureRandom = new SecureRandom
+
+    val byteArray: Array[Byte] = new Array[Byte](bits / 8)
+    randomGenerator.nextBytes(byteArray)
+    val bitVector = BitVector(byteArray)
```
Christewart, line 207: It might be wise to have some sort of invariant that the `bits % 8 == 0`. The ensuring clause will catch this invariant, but it will be baffling to the user why they did not get enough entropy.

testkit/src/main/scala/org/bitcoins/core/gen/CryptoGenerators.scala
```diff
@@ -4,12 +4,109 @@ import org.bitcoins.core.crypto._
...
 import org.bitcoins.core.util.CryptoUtil
 import org.scalacheck.Gen
+import scodec.bits.BitVector
 
 /**
   * Created by chris on 6/22/16.
   */
 sealed abstract class CryptoGenerators {
 
+  object entropy {
```
Christewart: Seems weird to me to have a `object` inside of a class. Any reason we don't just move this into it's own file?

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
Christewart: Maybe i am missing something here, but why can't you just generate a random bit vector of the appropriate size and check that it fails the checksum test?

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
```diff
@@ -0,0 +1,65 @@
...
+import javax.crypto.spec.PBEKeySpec
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.Factory
+import scodec.bits.ByteVector
+
+sealed abstract class BIP39Seed extends NetworkElement {
+  private val MIN_SEED_LENGTH_BYTES = 16
+
+  private val MAX_SEED_LENGTH_BYTES = 64
+  require(
```
Christewart, line 10: Please put the invariant as the first thing inside of the class, easier to see for a reader.

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -171,6 +175,19 @@ object ExtPrivateKey extends Factory[ExtPrivateKey] {
             "Fingerprint must be 4 bytes in size, got: " + fingerprint)
   }
 
+  def fromString(hex: String): Try[ExtPrivateKey] =
```
Christewart: Why do we have this here? Is this just a safe version of `fromHex`?

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -194,29 +211,41 @@ object ExtPrivateKey extends Factory[ExtPrivateKey] {
     ExtPrivateKeyImpl(version, depth, fingerprint, child, chainCode, privateKey)
   }
 
+  /**
+    * Hard coded value according to
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation BIP32]]
+    */
+  private val BIP39_KEY: ByteVector =
```
Christewart: Wrong val name? Shouldn't this be `BIP32_KEY`

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,257 @@
...
+  * HD wallet.
+  */
+sealed abstract class MnemonicCode {
+
+  /**
+    * The valid lengths a BIP39 mnemonic code phrase can be, according to
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki BIP39]]
+    */
+  private val VALID_LENGTHS = Vector(12, 15, 18, 21, 24)
+  require(
```
Christewart, line 20: invariants at the top of the class definition please

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
Christewart: I think we should at least do `private[crypto] def toEntropy = ...`

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,257 @@
...
+  /**
+    * The entropy for a mnemonic code has to be a multiple of 32
+    */
+  private[crypto] val ENTROPY_MULTIPLE = 32
+
+  /**
+    * Generates a mnemonic code from the given entropy bits
+    */
+  def fromEntropy(entropy: BitVector): MnemonicCode = {
+    require(
```
Christewart, line 159: Can't we move this invariants into the abstract class?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,257 @@
...
+
+  private[crypto] lazy val ENGLISH_WORDS: Vector[String] = {
+    val ENGLISH_WORDS_FILE = "/bip39-wordlists/english.txt"
+    val resourceStream = getClass.getResourceAsStream(ENGLISH_WORDS_FILE)
+    val source = Source.fromInputStream(resourceStream)
+
+    val lines = source.getLines
+    val linesVec = lines.toVector
+    linesVec
+  } ensuring (words =>
```
Christewart: Can we use dot notation here though, i.e. 

```}.ensuring(words => ...```

File: core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala (hunk 26 above)
Comment (torkelrogstad) [reply core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala]: It seems like that is a Scalacheck think, not Scalatest

File: core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala (hunk 28 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala:207]: It is a  private method, though. I think that makes a `require` a bit overkill. If you disagree I can add it in

File: testkit/src/main/scala/org/bitcoins/core/gen/CryptoGenerators.scala (hunk 29 above)
Comment (torkelrogstad) [reply testkit/src/main/scala/org/bitcoins/core/gen/CryptoGenerators.scala]: Wanted to keep all crypto related generators within one file, and avoid a bunch of name repetition.

File: core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala (hunk 31 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala]: 1) variable name is confusing from by side, should be `base58` and not `hex`. 
2) method already exists on `ExtKey`, but the user has to perform a match/cast to get a private/public key. This (and the correpsonding method in `ExtPublicKey`) does that match/cast for them

File: core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala (hunk 34 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala:159]: These `require` statements are checking the validity of the provided arguments to the function, though. I'd say it makes more sense to have them in the function. Fail fast, and all that. Open to changing it

File: core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala (hunk 31 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala]: 1) variable name is confusing from by side, should be `base58` and not `hex`. 
2) method already exists on `ExtKey`, but the user has to perform a match/cast to get a private/public key. This (and the corresponding method in `ExtPublicKey`) does that match/cast for them

File: core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala (hunk 27 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala:179]: The checksum isn't provided by the user, it's calculated by us and then used to get the final word in our phrase. Then later on (at the top of the abstract class) , we validate it.

File: core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala (hunk 14 above)
Comment (torkelrogstad) [reply core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala:77]: Haven't found one. Can add it to `BitcoinSUtil`? What would the signature look like, in that case? 

What I tried: 

```scala
def binarySearch[
      T <: Ordered[T],
  ](sortedCollection: Vector[T], elem: T): Option[Int] 
```

the method itself compiles, but not passing it a `Vector[String]` and a `String`

core-test/src/test/scala/org/bitcoins/core/crypto/MnemonicCodeTest.scala
torkelrogstad, line 128: I ended up not doing this as a property check. Tried two approaches: 

1) Replace last word in valid phrase
2) Generate a random checksum

The issue was that the checksum is so short that we end up with accidentally producing valid seeds.

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
```diff
@@ -0,0 +1,270 @@
...
+    val checksum = hashedEntropy.take(codeInfo.checksumLength)
+
+    entropyNoChecksum ++ checksum == entropyWithChecksum
+
+  }
+
+  /**
+    * The mnemonic code itself
+    */
+  def words: Vector[String]
```
Christewart, line 50: Should this be `private[crypto]` as well?

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
Christewart: I think it is probably better to error on the side of overkill with code like this

core/src/main/scala/org/bitcoins/core/crypto/MnemonicCode.scala
torkelrogstad: I ended up doing `private[bitcoins]` because a generator in `core.gen` uses it. Is that fine by you?

core/README.md
```diff
@@ -53,6 +53,47 @@ hexAgain: String = 0100000001ccf318f0cbac588a680bbad075aebdda1f211c94ba28125b0f6
...
+Here's an example: 
+
+```scala
+import scodec.bits._
+import org.bitcoins.core.crypto._
+
+// the length of the entropy bit vector determine
+// how long our phrase ends up being
+// 128 bits of entropy results in 12 words
+val entropy: BitVector = MnemonicCode.getEntropy128Bits
```
Christewart: I think in our examples we should use more entropy, people are likely to just copy and paste code from examples without thinking...

PR #355, opened 2019-02-20
testkit refactor (rename and repackage)

### LINKED ISSUE #352: Rename AsyncUtil in testkit to something else
It's extremely confusing these two files are named the same thing. We should probably have a teskit package that it goes into. 

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/testkit/src/main/scala/org/bitcoins/util/AsyncUtil.scala

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/AsyncUtil.scala

nkohen:
Refactored testkit to have a package called testkit

Renamed `AsyncUtil` -> `TestAsyncUtil` and `RpcUtil` -> `TestRpcUtil`

Ran scalafmt

Fixes #352


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/AsyncUtil.scala
```diff
@@ -55,12 +55,14 @@ abstract class AsyncUtil extends BitcoinSLogger {
...
+     *
      * This trims the top of the stack trace to exclude these internal calls.
      */
     private val relevantStackTrace = caller.tail
-      .dropWhile(elem => elem.getFileName == "AsyncUtil.scala"
-        || elem.getFileName == "RpcUtil.scala")
+      .dropWhile(
+        elem =>
+          elem.getFileName == "AsyncUtil.scala"
+            || elem.getFileName == "RpcUtil.scala")
```
torkelrogstad: Does this filename need to change?

nkohen: Yup

PR #360, opened 2019-02-25
Rework Bech32

torkelrogstad:
Refactor commonalities between LN and BTC
Add support for RegTest

There's a distinction in the Bech32 standard between a Bech32 _string_ and a Bech32 _address_. That wasn't handled correctly in the code as is. `Bech32.splitToHrpAndData` performs this operation on Bech32 _strings_, and is used in the test vectors grabbed from Core/Sipa.


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -66,11 +65,11 @@ sealed abstract class P2SHAddress extends BitcoinAddress {
   */
 sealed abstract class Bech32Address extends BitcoinAddress {
 
-  def hrp: HumanReadablePart
+  def hrp: BtcHumanReadablePart
```
Christewart, line 68: This should stay as a `HumanReadablePart` -- and we should create a 

```scala

sealed abstract class BtcBech32Address extends Bech32Address { 
  override def hrp: BtcHumanReadablePart
  override def networkParameters: BitcoinNetwork 
  //etc 
}
```

core/src/main/scala/org/bitcoins/core/protocol/BtcHumanReadablePart.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.bitcoins.core.util.Bech32HumanReadablePart
+
+import scala.util.{Failure, Success, Try}
+
+/**
+  * Represents the HumanReadablePart of a Bech32 address
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki]]
+  */
+sealed abstract class BtcHumanReadablePart extends Bech32HumanReadablePart {
+  def network: Option[NetworkParameters]
```
Christewart: I think network should be a required field and we should disregard what Pieter Wuille said, it really doesn't make sense to have a hrp that isn't associated with a network IMO.

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
```diff
@@ -3,13 +3,13 @@ package org.bitcoins.core.protocol.ln
...
 import org.bitcoins.core.protocol.ln.currency.{LnCurrencyUnit, LnCurrencyUnits}
-import org.bitcoins.core.util.Bech32
+import org.bitcoins.core.util.{Bech32, Bech32HumanReadablePart}
 import scodec.bits.ByteVector
 
 import scala.util.matching.Regex
 import scala.util.{Failure, Success, Try}
 
-sealed abstract class LnHumanReadablePart {
+sealed abstract class LnHumanReadablePart extends Bech32HumanReadablePart {
```
Christewart, line 12: I think i would rather move all of these `HumanReadablePart`s to one file and create a proper ADT

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -248,3 +311,22 @@ object Bech32 extends Bech32 {
...
+  require(chars.forall(Bech32.isInHrpRange),
+          s"Some characters in $chars were not in valid HRP range ([33-126])")
+
+  def chars: String
+
+  /** Expands this HRP into a vector of UInt5s, in accordance with the Bech32 spec */
+  def expand: Vector[UInt5] = Bech32.hrpExpand(this)
+}
+
+object Bech32HumanReadablePart {
```
Christewart: I'm against this, it doesn't make sense since there isn't a "concrete" Bech32HumanReadablePart. This needs to be an abstract type, and then we pattern match during the parsing phase to figure what network it is associated with

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -42,18 +39,18 @@ sealed abstract class Bech32 {
...
+    val x: Vector[UInt5] = lowerchars.map { c =>
+      UInt5(c >> 5)
+    }.toVector
+
+    val y: Vector[UInt5] = lowerchars.map { c =>
+      UInt5(c & 0x1f)
+    }.toVector
 
-    UInt5.toUInt5s(result)
+    x ++ Vector(UInt5(0)) ++ y
```
Christewart: `UInt5.zero`

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
torkelrogstad: This would require `Bech32Address` to extend something else though, and a general rework of how the different addresses work. All the other addresses extend `BitcoinAddress`, after all. Perhaps this would be better to do at a later point, when more coins are added to Bitcoin-S?

core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala
torkelrogstad: I'm not completely sure what you mean here. Do you want `HumanReadablePart`, `LnHumanReadablePart` and `BtcHumanReadablePart` to live in one file? And `LnHumanReadablePart` is an ADT right now?

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
Christewart: Hmm, so thinking about this more, `Bech32Address` should probably be a trait and we should do 

```scala

sealed trait Bech32Address { ... }

sealed abstract class BtcBech32Address extends BitcoinAddress with Bech32Address { ... }

``` 

I guess `Bech32Address` isn't really specific to a network, so we shouldn't make it network specific probably. I could be convinced that we should rework this on a different PR, but we should definitely file an issue.

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
torkelrogstad: I went down this road today, but realized this breaks tests from Sipa: https://github.com/sipa/bech32/blob/master/ref/python/tests.py#L34

It is possible for a `Bech32HumanReadablePart` to exist independently, as the standard isn't necessarily tied to Bitcoin

PR #361, opened 2019-02-25
Add flyway plugin to manage database schemas

Christewart:
Adds the [sbt-flyway plugin](https://davidmweber.github.io/flyway-sbt-docs/) to the node project. This will be using to manage our database schema changes in the future.

https://flywaydb.org/

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -268,18 +268,70 @@ lazy val doc = project
 
 lazy val nodeFlywaySettings = {
   lazy val DB_HOST = "localhost"
-  lazy val DB_NAME = "nodedb"
-  lazy val unittest = List(
-    Test / flywayUrl := s"jdbc:sqlite:${System.getenv("HOME")}/.bitcoin-s/unittest/nodedb.sqlite",
+  lazy val DB_NAME = "nodedb.sqlite"
+  lazy val network = "unittest" //mainnet, testnet3, regtest, unittest
+
+  lazy val mainnetDir = s"${System.getenv("HOME")}/.bitcoin-s/mainnet/"
+  lazy val testnetDir = s"${System.getenv("HOME")}/.bitcoin-s/testnet3/"
+  lazy val regtestDir = s"${System.getenv("HOME")}/.bitcoin-s/regtest/"
+  lazy val unittestDir = s"${System.getenv("HOME")}/.bitcoin-s/unittest/"
+
+  lazy val dirs = List(mainnetDir,testnetDir,regtestDir,unittestDir)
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -261,6 +266,22 @@ lazy val doc = project
...
+  lazy val DB_HOST = "localhost"
+  lazy val DB_NAME = "nodedb"
+  lazy val unittest = List(
+    Test / flywayUrl := s"jdbc:sqlite:${System.getenv("HOME")}/.bitcoin-s/unittest/nodedb.sqlite",
+    Test / flywayLocations := List("nodedb/migration"),
+    Test / flywayUser := "nodedb",
+    Test / flywayPassword := "",
+    flywayUrl := s"jdbc:sqlite:${System.getenv("HOME")}/.bitcoin-s/unittest/nodedb.sqlite",
+    flywayUser := "nodedb",
+    flywayPassword := ""
```
torkelrogstad, line 296: Would it be an idea to do a env var here?

Christewart: We probably want to eventually, although with sqlite as the driver apparently it doesnt have users so it isn't really necessary now. I believe we could even remove the `nodedb` username setting.

build.sbt
```diff
@@ -261,6 +266,22 @@ lazy val doc = project
...
+  )
+
+  unittest
```
nkohen: I assume in the future we'll have other options for this with configurable url etc?

nkohen: Currently migrations are only going to affect this one database and it looks like there are others?

build.sbt
```diff
@@ -261,6 +266,74 @@ lazy val doc = project
...
+
+  //create directies if they DNE
+  dirs.foreach { d =>
+    val file = new File(d)
+    file.mkdirs()
+    val db = new File(d + DB_NAME)
+    db.createNewFile()
+  }
+
+  lazy val mainnet = List(
```
torkelrogstad: To cut down on the repetition we could do 

```scala
  def makeNetworkSettings(directory: File) = List(
    Test / flywayUrl := s"jdbc:sqlite:$directory$DB_NAME",
    Test / flywayLocations := List("nodedb/migration"),
    Test / flywayUser := "nodedb",
    Test / flywayPassword := "",
    flywayUrl := s"jdbc:sqlite:$directory$DB_NAME",
    flywayUser := "nodedb",
    flywayPassword := ""
  )

lazy val testnet3 = makeNetworkSettings(testnetDir)
// etc
```

PR #362, opened 2019-02-26
Test v16 and v17 at the same time

torkelrogstad:
TODO:
 - [x] Add Travis config

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala
@@ -7,10 +7,23 @@ import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.ln.currency.MilliSatoshis
 import org.bitcoins.core.protocol.script.ScriptPubKey
 import org.bitcoins.core.protocol.transaction.{Transaction, TransactionInput}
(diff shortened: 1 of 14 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala
```diff
@@ -7,10 +7,23 @@ import org.bitcoins.core.protocol.BitcoinAddress
...
 object JsonWriters {
+  implicit object HashTypeWrites extends Writes[HashType] {
+    override def writes(hash: HashType): JsValue = hash match {
+      case _: SIGHASH_ALL                 => JsString("ALL")
+      case _: SIGHASH_NONE                => JsString("NONE")
+      case _: SIGHASH_SINGLE              => JsString("SINGLE")
+      case _: SIGHASH_ALL_ANYONECANPAY    => JsString("ALL|ANYONECANPAY")
+      case _: SIGHASH_NONE_ANYONECANPAY   => JsString("NONE|ANYONECANPAY")
+      case _: SIGHASH_SINGLE_ANYONECANPAY => JsString("SINGLE|ANYONECANPAY")
+      case _: SIGHASH_ANYONECANPAY        => ???
```
torkelrogstad: What's the correct way of handling this? AFAIK `SIGHASH_ANYONECANPAY` is not a valid RPC argument to bitcoind

PR #369, opened 2019-03-01
Add PSBT RPC calls

torkelrogstad:


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/RpcUtilTest.scala
```diff
@@ -138,7 +150,44 @@ class RpcUtilTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      secondBlock <- second.getBestBlockHash
+    } yield assert(firstBlock == secondBlock)
+  }
+
+  it should "be able to find outputs of previous transactions" in {
+    for {
+      (first, second) <- clientsF
+      address <- second.getNewAddress
+      txid <- first.sendToAddress(address, Bitcoins.one)
+      firstBlockhash +: _ <- BitcoindRpcTestUtil.generateAndSync(
```
Christewart: Not a big fan of using pattern matching this way, i would rather have `hashes` and then use `hashes.head` somewhere else.

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v17/PsbtRpcTest.scala
```diff
@@ -0,0 +1,223 @@
...
+    "BitcoindV17RpcClientTest_ActorSystem")
+  implicit val m: ActorMaterializer = ActorMaterializer()
+  implicit val ec: ExecutionContext = m.executionContext
+  implicit val networkParam: NetworkParameters = BitcoindRpcTestUtil.network
+
+  val client = new BitcoindV17RpcClient(BitcoindRpcTestUtil.v17Instance())
+  val otherClient = new BitcoindV17RpcClient(BitcoindRpcTestUtil.v17Instance())
+  val thirdClient = new BitcoindV17RpcClient(BitcoindRpcTestUtil.v17Instance())
+
+  override protected def beforeAll(): Unit = {
```
Christewart: So here is a starting point for refactoring this

```scala
  val startF =
    BitcoindRpcTestUtil.startServers(Vector(client, otherClient, thirdClient))

  val pairsF = startF.map { _ =>
    List((client, otherClient), (client, thirdClient))
  }

  val addNodesF: Future[List[Unit]] = {
    pairsF.flatMap { pairs =>
      val addedF = pairs.map {
        case (first, second) =>
          first.addNode(second.getDaemon.uri, AddNodeArgument.Add)
      }
      Future.sequence(addedF)
    }
  }

  val connectedF: Future[List[(BitcoindV17RpcClient, BitcoindV17RpcClient)]] = {
    pairsF.flatMap { pairs =>
      val connectedPairsF = {
        pairs.map {
          case (first, second) =>
            val connF =
              BitcoindRpcTestUtil.awaitConnectionF(first,
                                                   second,
                                                   duration = 1.second)
            connF.map(_ => (first, second))
        }
      }
      Future.sequence(connectedPairsF)
    }
  }
```

and i think this should be moved outside of the `beforeAll`. I also took the liberty of `BitcoindRpcTestUtil.awaitConnection` to these two methods now which i use above:

```scala
  def awaitConnection(
      from: BitcoindRpcClient,
      to: BitcoindRpcClient,
      maxTries: Int = 50)(implicit system: ActorSystem): Unit = {
    implicit val ec: ExecutionContextExecutor = system.dispatcher
    val connF = awaitConnectionF(from, to, duration, maxTries)
    Await.result(connF, 1.hour)
  }

  def awaitConnectionF(
      from: BitcoindRpcClient,
      to: BitcoindRpcClient,
      maxTries: Int = 50)(implicit system: ActorSystem): Future[Unit] = {
    import system.dispatcher
    def isConnected(): Future[Boolean] = {
      from
        .getAddedNodeInfo(to.getDaemon.uri)
        .map { info =>
          info.nonEmpty && info.head.connected.contains(true)
        }
    }

    AsyncUtil.retryUntilSatisfiedF(conditionF = isConnected,
                                   duration = duration,
                                   maxTries = maxTries)
  }
```

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v17/PsbtRpcTest.scala
```diff
@@ -0,0 +1,223 @@
...
+    })
+
+    Await.result(addNodesF, DEFAULT_TIMEOUT)
+
+    pairs.foreach {
+      case (first, second) =>
+        BitcoindRpcTestUtil.awaitConnection(first, second, duration = 1.second)
+    }
+
+    Await.result(client.generate(200), DEFAULT_TIMEOUT)
```
Christewart: In general it's a bad idea to do `Await.result` like this, we should aggregate the futures and then Await if we must. We probably don't need to await here though, we should just have 

```scala
val clientF = client.generate(200).map(_ => client)
```

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RpcOpts.scala
```diff
@@ -10,14 +10,41 @@ import play.api.libs.json.{Json, Writes}
...
       changeAddress: Option[BitcoinAddress] = None,
       changePosition: Option[Int] = None,
       includeWatching: Boolean = false,
       lockUnspents: Boolean = false,
       reverseChangeKey: Boolean = true,
       feeRate: Option[Bitcoins] = None,
-      subtractFeeFromOutputs: Option[Array[Int]])
+      subtractFeeFromOutputs: Option[Vector[Int]])
+
+  sealed trait FeeEstimationMode
```
Christewart: abstract class please

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RpcPsbtResult.scala
```diff
@@ -0,0 +1,70 @@
...
+final case class RpcPsbtInput(
+    nonWitnessUtxo: Option[RpcTransaction],
+    witnessUtxo: Option[PsbtWitnessUtxoInput],
+    partialSignatures: Option[Map[ECPublicKey, ECDigitalSignature]],
+    sighash: Option[HashType],
+    redeemScript: Option[RpcPsbtScript],
+    witnessScript: Option[RpcPsbtScript],
+    bip32Derivs: Option[Vector[PsbtBIP32Deriv]],
+    finalScriptSig: Option[RpcPsbtScript],
+    finalScriptwitness: Option[Vector[String]], // todo(torkelrogstad)
```
Christewart: ping for todo

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RpcPsbtResult.scala
```diff
@@ -0,0 +1,70 @@
...
+    unknown: Option[Map[String, String]] // The unknown global fields
+) extends RpcPsbtResult
+
+final case class RpcPsbtScript(
+    asm: String, // todo(torkelrogstad) split into Vector[ScriptToken]?
+    hex: ScriptPubKey,
+    scriptType: Option[String], // todo(torkelrogstad) found values: witness_v0_keyhash, scripthash, multisig, ???
```
Christewart: Hmm, if we haven't created a ADT for this we probably should. It is used else where in the rpc client

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -292,6 +292,53 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+    require(clients.length > 1, "Can't sync less than 2 nodes")
+
+    implicit val ec: ExecutionContextExecutor = system.getDispatcher
+
+    for {
+      hashes <- clients.head.generate(blocks)
+      _ = {
+        clients.sliding(2).foreach {
+          case first +: second +: _ =>
+            awaitSynced(first, second)
```
Christewart: Don't like this, we should create `awaitSyncedF` and just start using that

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RpcPsbtResult.scala
torkelrogstad: I actually haven't found any example PSBTs that produce this field when testing. This is from the `bitcoind` documentation description of this field: `hex-encoded witness data (if any)`. Do you know what the correct type is here?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/RpcPsbtResult.scala
torkelrogstad: I agree. Do you know what the possible values are? The ones in that comment are those I've come across

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -272,31 +272,117 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+    */
+  def generateAllAndSync(clients: Vector[BitcoindRpcClient], blocks: Int = 6)(
+      implicit system: ActorSystem): Future[
+    Vector[Vector[DoubleSha256Digest]]] = {
+    import system.dispatcher
+
+    val sliding = ListUtil.slideHead(clients)
+
+    sliding.foldLeft(
+      Future.successful(Vector.empty[Vector[DoubleSha256Digest]])) {
```
Christewart: I think this is much more readable: 

```scala
  def generateAllAndSync(clients: Vector[BitcoindRpcClient], blocks: Int = 6)(
      implicit system: ActorSystem): Future[
    Vector[Vector[DoubleSha256Digest]]] = {
    import system.dispatcher

    val sliding = ListUtil.slideHead(clients)

    val initF = Future.successful(List.empty[Vector[DoubleSha256Digest]])
    
    val generatedHashesF = {
      sliding.foldLeft(initF) { case (accumHashesF, clients) =>
        accumHashesF.flatMap { accumHashes =>
          val hashesF = generateAndSync(clients, blocks)
          hashesF.map(hashes => hashes :: accumHashes)
        }
      }
    }
    
    generatedHashesF.map(_.reverse.toVector)
  }
```

testkit/src/main/scala/org/bitcoins/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -272,31 +272,117 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
+      }
+    } yield hashes
+  }
+
+  def awaitSyncedF(
       client1: BitcoindRpcClient,
       client2: BitcoindRpcClient,
       duration: FiniteDuration = 1.second,
-      maxTries: Int = 50)(implicit system: ActorSystem): Unit = {
+      maxTries: Int = 50)(implicit system: ActorSystem): Future[Unit] = {
```
Christewart, line 389: :heart_eyes:

testkit/src/main/scala/org/bitcoins/util/ListUtil.scala
```diff
@@ -0,0 +1,32 @@
...
+  /**
+    * Generates a vector of vectors "sliding" the head element
+    * over `xs`.
+    *
+    * {{{
+    *   > slideFirst(Vector(1, 2, 3))
+    *    Vector(Vector(1, 2, 3), Vector(2, 3, 1), Vector(3, 1, 2))
+    * }}}
+    */
+  def slideHead[T](xs: Vector[T]): Vector[Vector[T]] = {
```
Christewart: I think `rotateHead` might be a better method name here?

Issue #252, opened 2018-11-25
Investigate occasional reuse of r values failure in ECDigitalSignatureSpec


Christewart:

Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/493215099#L2374

Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/494541569

Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/494929893#L1299

Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/501662492#L2437

PR #371, opened 2019-03-04
Make the doublesha256digest generator more random

Christewart:
This makes the doublesha256digest generator more random, previously it was just hashing hex chars from `StringGenerators.hexString` which really wasn't that random at all


Christewart:
This should fix #252

PR #373, opened 2019-03-06
Move BlockHeaderDAO to chain project, create BlockHeaderDb type to re…

Christewart:
…prsent a header in the database

This PR creates a sqlite database called `block_headers` to store block headers for our blockchain. This also adds a lot more structure to the chain project for things like `ChainDbManagement`, `ChainConfig`, slick related column mappers etc. 

This PR does NOT integrate the node project and the chain project, i.e. we can't take headers that are received via the p2p network and store them in this database. This PR should be viewed as stand alone functionality.

PR #374, opened 2019-03-06
WIP: 2019 03 06 chain handling

Christewart:
This is built on #373 

This implements data structures related to managing and storing a blockchain. 

- `ChainApi` - this intended to be the interface that external data providers call to add blockchain data to the bitcoin-s chain project. The most important method right now is `ChainApi.processHeader` which will do all our internal work needed to update our blockchain with the header.
- `ChainHandler` - this manages instances of `Blockchain` 
- `Blockchain` - this stores the chain of headers in memory, where the current blockchain tip is at index 0, and the 2nd to last header seen is at index 1 etc. 
- `TipValidation` - this does all of the checks to make sure we can connect one `BlockHeader` to another. `TipValidation.check()` returns a `TipUpdateResult` ADT indicating if we can or cannot connect the tip. In the case that we can `TipUpdateResult.Success` contains a `BlockHeaderDb` instance which can be inserted into the database.


chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  /** Helper method to ensure our blockchain is properly connected at all times */
+  private def connectionInvariant: Boolean = {
+    val noTip = headers.tail
+    @tailrec
+    def loop(
+        remaining: List[BlockHeaderDb],
+        nextHeader: BlockHeaderDb): Boolean = {
```
torkelrogstad, line 39: This param is unused?

chain/src/main/scala/org/bitcoins/chain/validation/TipValidation.scala
```diff
@@ -0,0 +1,47 @@
...
+    logger.debug(
+      s"Checking header=${header.hashBE.hex} to try to connect to currentTip=${currentTip.hashBE.hex}")
+    if (header.previousBlockHashBE != currentTip.hashBE) {
+      logger.warn(
+        s"Failed to connect tip=${header.hashBE.hex} to current chain")
+      TipUpdateResult.BadPreviousBlockHash(newPotentialTip)
+    } else if (header.nBits != currentTip.nBits) {
+      //TODO: THis is a bug, this should only occurr only 2016 blocks
+      //also this doesn't apply on testnet/regtest
+      TipUpdateResult.BadPOW(newPotentialTip)
```
torkelrogstad, line 29: Is the intention of `BadPOW` to also check if sufficient work is performed? Or is that done elsewhere? 

Anyway I think it would be an idea to do the same you did with `isBadNonce` and move it into a method that can be clearly marked as WIP

torkelrogstad:
I love `BlockchainUpdate` and `UpdateTipResult`! I think a similar approach is valuable in multiple places. Would also work well if we return `Either[SealedFooFailure, Foo]` (either in this code with blockchain updates or more generally in unsafe operations)

PR #377, opened 2019-03-08
Implement difficulty calculation from nBits

Christewart:
useful links:

- https://bitcoin.org/en/developer-reference#target-nbits

- https://github.com/bitcoin/bitcoin/blob/master/src/pow.cpp#L74

- https://github.com/bitcoin/bitcoin/blob/2068f089c8b7b90eb4557d3f67ea0f0ed2059a23/src/arith_uint256.cpp#L206

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
@@ -144,6 +144,49 @@ trait NumberUtil extends BitcoinSLogger {
     //should always be able to convert from uint5 => uint8
     u8sTry.get
   }
+
+  /** Expands the [[org.bitcoins.core.protocol.blockchain.BlockHeader.nBits nBits]]
+    * field given in a block header to the _actual_ target difficulty.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -144,6 +144,49 @@ trait NumberUtil extends BitcoinSLogger {
...
+    val mantissaBytes = {
+      val withSignBit = nBits.bytes.takeRight(3)
+      val noSignBit = withSignBit.head & 0x7f
+      noSignBit.toByte +: withSignBit.tail
+    }
+
+    val significand = nBits.bytes.head
+
+    //if the most significant bit is set, we have a negative number
+    val signum = if ((0x80 & nBits.bytes(1)) != 0) {
```
torkelrogstad: Is it cleaner to do interact with the bits directly rather than using bit masks? `if (nBits.bytes.bits(1)) ...`

I guess same applies for line 162

PR #379, opened 2019-03-12
Add BIP44 and BIP32 path support

torkelrogstad:
Implements: 

- General BIP32 paths
- BIP44 paths (special case of BIP32 paths)
- Key derivation from xprivs and xpubs according to paths

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/crypto/ExtKeyTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/crypto/ExtKeyTest.scala
@@ -1,17 +1,87 @@
 package org.bitcoins.core.crypto
 
 import org.bitcoins.core.crypto.ExtKeyVersion._
+import org.bitcoins.core.crypto.bip32.BIP32Path
 import org.bitcoins.core.number.UInt32
-import org.scalatest.{FlatSpec, MustMatchers}
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, NumberGenerator}
+import org.bitcoins.testkit.util.BitcoinSUnitTest
 import scodec.bits.HexStringSyntax
 
-class ExtKeyTest extends FlatSpec with MustMatchers {
+import scala.util.{Failure, Success, Try}
+
+class ExtKeyTest extends BitcoinSUnitTest {
+
+  behavior of "ExtKey"
+
+  it must "derive similar keys for UInt32s and primitive numbers" in {
+    forAll(CryptoGenerators.extPrivateKey, NumberGenerator.uInt32s) {
+      (priv, index) =>
+        val derivedPrivUInt32 = priv.deriveChildPrivKey(index)
+        val derivedPrivPrimitive = priv.deriveChildPrivKey(index.toLong)
+        assert(Success(derivedPrivUInt32) == derivedPrivPrimitive)
+
+        val pub = priv.extPublicKey
+        val derivedPubUInt32: Try[ExtPublicKey] = pub.deriveChildPubKey(index)
+        val derivedPubPrimitive: Try[ExtPublicKey] =
+          pub.deriveChildPubKey(index.toLong)
+        (derivedPubUInt32, derivedPubPrimitive) match {
+          case (Success(_), Success(_)) => succeed
+          case (Failure(exc1), Failure(exc2)) =>
+            assert(exc1.getMessage == exc2.getMessage)
+          case _: (Try[ExtPublicKey], Try[ExtPublicKey]) => fail
+        }
+    }
+  }
+
+  it must "fail to make a private key out of a public key" in {
+    forAll(CryptoGenerators.extPublicKey) { pub =>
+      val attempt = ExtPrivateKey.fromString(pub.toString)
+      attempt match {
+        case Success(_)   => fail
+        case Failure(exc) => assert(exc.getMessage.contains("expected private"))
+      }
+    }
+  }
+
+  it must "derive private keys from a BIP32 path and an xpriv" in {
+    forAll(CryptoGenerators.extPrivateKey, CryptoGenerators.bip32Path) {
+      (priv, path) =>
+        priv.deriveChildPrivKey(path)
+        succeed
+    }
+  }
+
+  it must "derive public keys from a BIP32 path and an xpriv" in {
+    forAll(CryptoGenerators.extPrivateKey, CryptoGenerators.bip32Path) {
+      (priv, path) =>
+        val pub = priv.deriveChildPubKey(path)
+        pub match {
+          case Failure(exc) => fail(exc.getMessage)
+          case Success(_)   => succeed
+        }
+    }
+  }
+
+  it must "fail to derive public keys from a hardened public key" in {
+    forAll(CryptoGenerators.extPrivateKey, CryptoGenerators.hardBip32Child) {
+      (priv, child) =>
+        val pub = priv.extPublicKey
+        val derivedPub = pub.deriveChildPubKey(child.toUInt32)
+        derivedPub match {
+          case Success(_)   => fail
+          case Failure(exc) => assert(exc.getMessage.contains("hardened"))
+        }
+    }
+  }
 
   //https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vectors
-  "ExtKey" must "pass the test vectors in BIP32" in {
+  it must "pass the test vectors in BIP32" in {
     //master key
     val seedBytes = hex"000102030405060708090a0b0c0d0e0f"
-    val masterPriv = ExtPrivateKey(MainNetPriv, Some(seedBytes))
+
+    val path = BIP32Path.empty
+
+    val masterPriv = ExtPrivateKey(MainNetPriv, Some(seedBytes), path)
     masterPriv.toString must be(
       "xprv9s21ZrQH143K3QTDL4LXw2F7HEK3wJUD2nW2nRk4stbPy6cq3jPPqjiChkVvvNKmPGJxWUtg6LnF5kejMRNNU3TGtRBeJgk33yuGBxrMPHi")
 
@@ -21,8 +91,8 @@ class ExtKeyTest extends FlatSpec with MustMatchers {
       "xpub661MyMwAqRbcFtXgS5sYJABqqG9YLmC4Q1Rdap9gSE8NqtwybGhePY2gZ29ESFjqJoCu1Rupje8YtGqsefD265TMg7usUDFdp6W1EGMcet8")
 
     //derive child
-    val hidx = ExtKey.hardenedIdx
-    val m0h = masterPriv.deriveChildPrivKey(hidx)
+    val m0hPath = BIP32Path.fromString("m/0'")
+    val m0h = masterPriv.deriveChildPrivKey(m0hPath)
     m0h.toString must be(
       "xprv9uHRZZhk6KAJC1avXpDAp4MDc3sQKNxDiPvvkX8Br5ngLNv1TxvUxt4cV1rGL5hj6KCesnDYUhd7oWgT11eZG7XnxHrnYeSvkzY7d2bhkJ7")
 
@@ -30,7 +100,8 @@ class ExtKeyTest extends FlatSpec with MustMatchers {
     m0hPub.toString must be(
       "xpub68Gmy5EdvgibQVfPdqkBBCHxA5htiqg55crXYuXoQRKfDBFA1WEjWgP6LHhwBZeNK1VTsfTFUHCdrfp1bgwQ9xv5ski8PX9rL2dZXvgGDnw")
 
-    val m0h1 = m0h.deriveChildPrivKey(UInt32.one)
+    val m0h1Path = BIP32Path.fromString("m/0'/1")
+    val m0h1 = masterPriv.deriveChildPrivKey(m0h1Path)
     m0h1.toString must be(
       "xprv9wTYmMFdV23N2TdNG573QoEsfRrWKQgWeibmLntzniatZvR9BmLnvSxqu53Kw1UmYPxLgboyZQaXwTCg8MSY3H2EU4pWcQDnRnrVA1xe8fs")
 
@@ -38,7 +109,8 @@ class ExtKeyTest extends FlatSpec with MustMatchers {
     m0h1Pub.toString must be(
       "xpub6ASuArnXKPbfEwhqN6e3mwBcDTgzisQN1wXN9BJcM47sSikHjJf3UFHKkNAWbWMiGj7Wf5uMash7SyYq527Hqck2AxYysAA7xmALppuCkwQ")
 
-    val m0h12h = m0h1.deriveChildPrivKey(UInt32(2) + ExtKey.hardenedIdx)
+    val m0h1P2hath = BIP32Path.fromString("m/0'/1/2'")
+    val m0h12h = masterPriv.deriveChildPrivKey(m0h1P2hath)
     m0h12h.toString must be(
       "xprv9z4pot5VBttmtdRTWfWQmoH1taj2axGVzFqSb8C9xaxKymcFzXBDptWmT7FwuEzG3ryjH4ktypQSAewRiNMjANTtpgP4mLTj34bhnZX7UiM")
 
@@ -46,7 +118,8 @@ class ExtKeyTest extends FlatSpec with MustMatchers {
     m0h12hPub.toString must be(
       "xpub6D4BDPcP2GT577Vvch3R8wDkScZWzQzMMUm3PWbmWvVJrZwQY4VUNgqFJPMM3No2dFDFGTsxxpG5uJh7n7epu4trkrX7x7DogT5Uv6fcLW5")
 
-    val m0h12h2 = m0h12h.deriveChildPrivKey(UInt32(2))
+    val m0h12h2Path = BIP32Path.fromString("m/0'/1/2'/2")
+    val m0h12h2 = masterPriv.deriveChildPrivKey(m0h12h2Path)
     m0h12h2.toString must be(
       "xprvA2JDeKCSNNZky6uBCviVfJSKyQ1mDYahRjijr5idH2WwLsEd4Hsb2Tyh8RfQMuPh7f7RtyzTtdrbdqqsunu5Mm3wDvUAKRHSC34sJ7in334")
 
@@ -54,7 +127,8 @@ class ExtKeyTest extends FlatSpec with MustMatchers {
     m0h12h2Pub.toString must be(
       "xpub6FHa3pjLCk84BayeJxFW2SP4XRrFd1JYnxeLeU8EqN3vDfZmbqBqaGJAyiLjTAwm6ZLRQUMv1ZACTj37sR62cfN7fe5JnJ7dh8zL4fiyLHV")
 
-    val m0h12h21000000000 = m0h12h2.deriveChildPrivKey(UInt32(1000000000))
+    val m0h12h21000000000Path = BIP32Path.fromString("m/0'/1/2'/2/1000000000")
+    val m0h12h21000000000 = masterPriv.deriveChildPrivKey(m0h12h21000000000Path)
     m0h12h21000000000.toString must be(
       "xprvA41z7zogVVwxVSgdKUHDy1SKmdb533PjDz7J6N6mV6uS3ze1ai8FHa8kmHScGpWmj4WggLyQjgPie1rFSruoUihUZREPSL39UNdE3BBDu76")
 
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -119,10 +121,15 @@ sealed abstract class ExtPrivateKey extends ExtKey {
 
   override def key: ECPrivateKey
 
+  def deriveChildPrivKey(path: BIP32Path): ExtPrivateKey = {
```
Christewart, line 173: Scaladoc explaining how this can be used

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
+package org.bitcoins.core.crypto.bip32
+
+import org.bitcoins.core.number.UInt32
+
+abstract class BIP32Path {
```
Christewart, line 6: is there a reason this isn't sealed?

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+      .fold("m")((accum, curr) => accum + "/" + curr)
+
+}
+
+object BIP32Path {
+
+  /**
+    * The empty BIP32 path "m"
+    */
+  def empty: BIP32Path = BIP32PathImpl(Vector.empty)
```
Christewart: Link to BIP32 to what this means to be path 'm' and also 

```scala
val empty: Bip32Path = ...
```

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  private case class BIP32PathImpl(children: Vector[BIP32Child])
```
Christewart: move this to the top

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
Christewart: Also is there any invariants we can add in here?

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+      extends BIP32Path
+
+  def apply(children: Vector[BIP32Child]): BIP32Path = BIP32PathImpl(children)
+
+  def fromString(string: String): BIP32Path = {
```
Christewart, line 51: Scaladoc link to where this string format is coming from

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+    val parts = string
+      .split("/")
+      .toVector
+      .map(_.trim)
```
Christewart, line 57: why did you add in `.trim` and `.toLowerCase`? Is this in the BIP somewhere that this is specified? 

If not, is there any way this can cause incompatabilities where users of the wallet infrastructure are storing _invalid_ strings some where that we are deriving seemingly valid keys? 

I think we need to be careful with sanitizing input, it is probably best in situations like this to throw. Can you link to any other implementations that do this as well?

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    BIP32PathImpl(children)
+  }
+}
+
+case class BIP32Child(index: Int, hardened: Boolean) {
+  require(index >= 0, "BIP32 child index must be positive!")
+
+  private val HARDENED_KEY_OFFSET = 2147483647 // pow(2, 31)
```
Christewart: i think there was a `ExtKey.hardenedIdx`, let's just reuse that val

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+    * Converts this child to a BIP32 notation
+    * unsigned 32 bit integer
+    */
+  def toUInt32: UInt32 =
+    if (hardened) UInt32(HARDENED_KEY_OFFSET + index.toLong + 1)
```
Christewart: the `+1` scares me, where is that coming from ?

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Account.scala
```diff
@@ -0,0 +1,21 @@
+package org.bitcoins.core.crypto.bip44
+import org.bitcoins.core.crypto.bip32.BIP32ChainType
+
+sealed abstract class BIP44Account {
```
Christewart: Link to BIP in scala doc please

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32ChainType.scala
```diff
@@ -0,0 +1,31 @@
+package org.bitcoins.core.crypto.bip32
+import scala.annotation.switch
+
+sealed abstract class BIP32ChainType
```
Christewart: Scaladoc, and 

```scala
sealed abstract class Bip32ChainType { 
  def index: Int
}

case object External extends Bip32ChainType { 
  override val index: Int = 0 
}
```

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32ChainType.scala
```diff
@@ -0,0 +1,31 @@
...
+  /**
+    * External chain is used for addresses that
+    * are meant to be visible outside of the
+    * wallet (e.g. for receiving payments).
+    */
+  case object Change extends BIP32ChainType
+
+  def fromInt(int: Int): BIP32ChainType =
+    (int: @switch) match {
+      case 0 => BIP32ChainType.External
```
Christewart:
```scala 
int match { 
  case BIP32ChainType.External.index => BIP32ChainType.External
  case BIP32ChainType.Change.index => BIP32ChainType.Change
  case _: Int => ... 
}
```

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32ChainType.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object BIP32ChainType {
+
+  /**
+    * Internal chain is used for addresses which
+    * are not meant to be visible outside of the
+    * wallet and is used for return transaction
+    * change
+    */
+  case object External extends BIP32ChainType
```
Christewart: i think the scaladocs here are switched

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32ChainType.scala
Christewart: Is this specified in BIP32 or is this BIP44? Let's make sure we get the names right with links to documentation

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
torkelrogstad: BIP44 paths are an extension to BIP32 paths (certain path fields have specific meanings they lack in BIP32). So that makes sealing difficult. 

Don't think there are any invariants. The path can be empty, and I don't believe it has a upper bound for long it can continue

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
torkelrogstad: You're right about upper/lower case, removed that. There are many examples of people having/lacking whitespace between their segments. Linked to one example for each in comment above.

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32ChainType.scala
torkelrogstad: you're right good catch

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
torkelrogstad: Was because of wonky arithmetics, solved it by adding UInt32s directly

core-test/src/test/scala/org/bitcoins/core/crypto/ExtKeyTest.scala
```diff
@@ -1,17 +1,87 @@
...
+import org.bitcoins.testkit.util.BitcoinSUnitTest
 import scodec.bits.HexStringSyntax
 
-class ExtKeyTest extends FlatSpec with MustMatchers {
+import scala.util.{Failure, Success, Try}
+
+class ExtKeyTest extends BitcoinSUnitTest {
+
+  behavior of "ExtKey"
+
```
Christewart, line 18: since these seem like newish tests, you may want to consider overriding 
```scala
  override implicit val generatorDrivenConfig: PropertyCheckConfiguration   = generatorDrivenConfigNewCode
```

this will mean we have more iterations on the property based tests

core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, NumberGenerator}
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.scalacheck.{Gen, Shrink}
+import org.scalatest.path
+
+import scala.util.{Success, Try}
+
+class BIP32PathTest extends BitcoinSUnitTest {
+
+  override implicit val generatorDrivenConfig: PropertyCheckConfiguration =
```
Christewart, line 13: :+1:

core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  it must "have varargs and vector constructors what work the same way" in {
+    forAll(CryptoGenerators.bip32Path) { path =>
+      assert(BIP32Path(path.children) == BIP32Path(path.children: _*))
+    }
+  }
+
+  it must "derive public keys" in {}
```
Christewart: ping

core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  it must "parse a hardened path" in {
+    val fromString = BIP32Path.fromString("m/0'")
+    assert(fromString.children.length == 1)
+    assert(fromString.children.head.toUInt32 == ExtKey.hardenedIdx)
+  }
+
+  it must "parse the paths from the BIP32 test vectors" in {
+    assert(BIP32Path.fromString("m/0'/1") == BIP32Path(
```
Christewart: nits: Would be nicer to read with all of the following test cases if you made a intermediate val

```scala
val expectedPath = BIP32Path(
      Vector(BIP32Child(0, hardened = true), BIP32Child(1, hardened = false)))

assert(BIP32Path.fromString("m/0'/1") == expectedPath
```

core-test/src/test/scala/org/bitcoins/core/crypto/bip44/BIP44PathTest.scala
```diff
@@ -0,0 +1,272 @@
...
+
+import org.bitcoins.core.crypto.bip32.BIP32Path
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, NumberGenerator}
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import os.copy
+
+import scala.util.{Failure, Success, Try}
+
+class BIP44PathTest extends BitcoinSUnitTest {
+  behavior of "BIP44Account"
```
Christewart, line 13: new code so override the generator here so we have more test cases

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -42,16 +44,38 @@ sealed abstract class ExtKey extends NetworkElement {
...
 
+  /**
+    * Derives the child pubkey at the specified path
+    */
+  def deriveChildPubKey(path: BIP32Path): Try[ExtPublicKey] = {
+    this match {
+      case priv: ExtPrivateKey =>
+        Success(priv.deriveChildPrivKey(path).extPublicKey)
+      case pub: ExtPublicKey =>
+        Try {
```
Christewart: An alternative that doesn't throw and preserves the original error message, but is longer

```scala
        @tailrec
        def loop(
            remainingPath: List[BIP32Child],
            accum: ExtPublicKey): Try[ExtPublicKey] = {
          remainingPath match {
            case h :: t =>
              val derivedPubT = accum.deriveChildPubKey(h.toUInt32)
              if (derivedPubT.isSuccess) {
                loop(t, derivedPubT.get)
              } else {
                derivedPubT
              }
            case Nil => Success(accum)
          }
        }
        loop(path.children.toList, pub)
```

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44ChainType.scala
```diff
@@ -0,0 +1,39 @@
...
+}
+
+object BIP44ChainType {
+
+  /**
+    * External chain is used for addresses that
+    * are meant to be visible outside of the
+    * wallet (e.g. for receiving payments).
+    */
+  case object External extends BIP44ChainType {
```
Christewart: mark these as final

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44ChainType.scala
```diff
@@ -0,0 +1,39 @@
...
+    */
+  case object Change extends BIP44ChainType {
+    override final val index: Int = 1
+  }
+
+  def fromInt(int: Int): BIP44ChainType =
+    int match {
+      case External.index => BIP44ChainType.External
+      case Change.index   => BIP44ChainType.Change
+      case _ =>
```
Christewart:
```scala
case _: Int => 
```

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Coin.scala
```diff
@@ -0,0 +1,42 @@
...
+    */
+  case object Testnet extends BIP44Coin {
+    override final val toInt: Int = 1
+  }
+
+  def fromInt(int: Int): BIP44Coin =
+    int match {
+      case Bitcoin.toInt => Bitcoin
+      case Testnet.toInt => Testnet
+      case _ =>
```
Christewart:
```scala
case _ : Int => ...
```

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Path.scala
```diff
@@ -0,0 +1,112 @@
+package org.bitcoins.core.crypto.bip44
+
+import org.bitcoins.core.crypto.bip32.{BIP32Child, BIP32Path}
+
+import scala.util.Try
+
+sealed abstract class BIP44Path extends BIP32Path {
```
Christewart, line 7: This is really nice work!

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CryptoGenerators.scala
```diff
@@ -223,6 +225,91 @@ sealed abstract class CryptoGenerators {
 
   def extKey: Gen[ExtKey] = Gen.oneOf(extPrivateKey, extPublicKey)
 
+  def bip32Child: Gen[BIP32Child] =
+    for {
```
Christewart:
```scala
Gen.oneOf(softBip32Child,hardBip32Child) 
```
?

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Path.scala
```diff
@@ -0,0 +1,112 @@
...
+    */
+  val ACCOUNT_INDEX: Int = 2
+
+  /**
+    * The index of the chain segement of a BIP44 path
+    */
+  val CHAIN_INDEX: Int = 3
+
+  /**
+    * The index of the coin segement of a BIP44 path
```
torkelrogstad: ping

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Path.scala
```diff
@@ -0,0 +1,112 @@
...
+              chainType = address.chain.chainType,
+              addressIndex = address.index)
+  }
+
+  /**
+    * Tries to generate a BIP44 path from the given path segments
+    */
+  def apply(children: Vector[BIP32Child]): Try[BIP44Path] = Try {
+    val bip32 = BIP32Path(children)
+    BIP44Path.fromString(bip32.toString)
```
torkelrogstad: ping use `apply`

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,88 @@
+package org.bitcoins.core.crypto.bip32
+
+import org.bitcoins.core.crypto.ExtKey
+import org.bitcoins.core.number.UInt32
+
+abstract class BIP32Path {
+  def children: Vector[BIP32Child]
```
Christewart: So I think i would prefer to name this `path` rather than `children`. From my understanding, `Vector[BIP32Child]` does _not_ contain only children. 

In the case of `BIP44Coin`, it actually appears that it is the opposite. It appears that it contains itself, and all of it's ancestors? 

https://github.com/bitcoin-s/bitcoin-s-core/pull/379/files#diff-eaf823968efd2e0e07cc92b04ae905cbR10

core/src/main/scala/org/bitcoins/core/crypto/bip32/BIP32Path.scala
```diff
@@ -0,0 +1,88 @@
...
+          (str, false)
+        }
+      BIP32Child(index.toInt, hardened)
+    }
+
+    BIP32PathImpl(children)
+  }
+}
+
+case class BIP32Child(index: Int, hardened: Boolean) {
```
Christewart: I think this should be renamed to something like `BIP32Entry`/`Bip32Node`/`BIP32Level` (level is what it is called in BIP44).

core/src/main/scala/org/bitcoins/core/crypto/bip44/BIP44Coin.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.bitcoins.core.crypto.bip32.{BIP32Child, BIP32Path}
+
+/**
+  * Represents a
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#change BIP44]]
+  * coin type.
+  */
+sealed abstract class BIP44Coin extends BIP32Path {
+  override def children: Vector[BIP32Child] =
+    Vector(BIP44Path.purposeChild, BIP32Child(toInt, hardened = true))
```
Christewart: I think we should have 

```scala
def node: Bip32Node = BIP44Node(toInt)
override def ancestors: Vector[BIP32Node] = Vector(BIP44Path.purposeChild)
override def path = ancectors.+=(node)
```

Christewart: and something like this for all other BIP44* implementations

torkelrogstad: I don't really see the benefit of having `ancestors` in there. For it to be overridable it would have to be in the `BIP32Path` abstract class. But that's supposed to represent a complete path, and has no ancestors. It could be added to `BIP32Node`, but I don't really see the benefit there either

PR #385, opened 2019-03-19
Reduce number of payments sent from 5 to 2 in gossiping channel updat…

Christewart:
…es messages

Attempt to fix our annoying eclair test case that seems to fail on CI but not locally


Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s-core/jobs/505203992#L848

PR #388, opened 2019-03-20
2019 03 20 chainparams add mining consensus rules

Christewart:
This adds mining related consensus parameters to `ChainParams`. These parameters can be found here

https://github.com/bitcoin/bitcoin/blob/eb7daf4d600eeb631427c018a984a77a34aca66e/src/consensus/params.h#L70-L76

The implementation of this struct is found in this file

https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp

This will be needed for block header validation in the node project.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
@@ -11,8 +11,8 @@ import org.bitcoins.core.protocol.script.{ScriptPubKey, ScriptSignature}
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.script.constant.{BytesToPushOntoStack, ScriptConstant}
 import org.bitcoins.core.script.crypto.OP_CHECKSIG
-import org.bitcoins.core.util.{BitcoinSUtil, BitcoinScriptUtil}
-import scodec.bits.ByteVector
+import org.bitcoins.core.util.BitcoinScriptUtil
+import scodec.bits.{ByteVector, _}
 
 import scala.concurrent.duration.{Duration, DurationInt}
 
@@ -164,45 +164,37 @@ sealed abstract class ChainParams {
   def powTargetTimeSpan: Duration
 
   /**
-    * Not sure what this value is right now, but it is constant across bitcoin
+    * The targeted interval between blocks
     * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L74 mainnet]]
     * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L191 testnet]]
     * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L286 regtest]]
     * @return
     */
-  def powTargetSpacing: Int
+  def powTargetSpacing: Duration
 
   /** In bitcoin [[MainNetChainParams mainnet]], the network recalculates the difficulty for the network every 2016 blocks
     * [[https://github.com/bitcoin/bitcoin/blob/eb7daf4d600eeb631427c018a984a77a34aca66e/src/consensus/params.h#L75 bitcoin core implementation]]
     * */
   def difficultyChangeInterval: Long = {
-    powTargetTimeSpan.toSeconds / powTargetSpacing
+    powTargetTimeSpan.toSeconds / powTargetSpacing.toSeconds
   }
 }
 
 sealed abstract class BitcoinChainParams extends ChainParams {
 
-  /** The targetted timespan between difficulty adjustments
-    * As of this implementation, all of these are the same in bitcoin core
-    *
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L73 mainnet]]
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L190 testnet]]
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L285 regtest]]
+  /**
+    * @inheritdoc
     * */
   override lazy val powTargetTimeSpan: Duration = {
-    val time = 14 * 24 * 60 * 60 //two weeks
-    time.seconds
+    14.days
   }
 
   /**
-    * Not sure what this value is right now, but it is constant across bitcoin
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L74 mainnet]]
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L191 testnet]]
-    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L286 regtest]]
-    * @return
+    * @inheritdoc
     */
-  override lazy val powTargetSpacing: Int = {
-    10 * 60 //what are these magic numbers?
+  override lazy val powTargetSpacing: Duration = {
+    val time = 10 * 60 //10 minutes * 60 seconds
+    time.seconds
   }
 
   /** The best chain should have this amount of work */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+  def powTargetTimeSpan: Duration
+
+  /**
+    * Not sure what this value is right now, but it is constant across bitcoin
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L74 mainnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L191 testnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L286 regtest]]
+    * @return
+    */
+  def powTargetSpacing: Int
```
torkelrogstad: Could this be how often we're supposed to have blocks? 10  * 60 seconds = 10 minutes

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+
+  /** In bitcoin [[MainNetChainParams mainnet]], the network recalculates the difficulty for the network every 2016 blocks
+    * [[https://github.com/bitcoin/bitcoin/blob/eb7daf4d600eeb631427c018a984a77a34aca66e/src/consensus/params.h#L75 bitcoin core implementation]]
+    * */
+  def difficultyChangeInterval: Long = {
```
torkelrogstad, line 178: Is this `nMinerConfirmationWindow`? In that case it'd be nice to mention that

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+sealed abstract class BitcoinChainParams extends ChainParams {
+
+  /** The targetted timespan between difficulty adjustments
+    * As of this implementation, all of these are the same in bitcoin core
+    *
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L73 mainnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L190 testnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L285 regtest]]
+    * */
+  override lazy val powTargetTimeSpan: Duration = {
```
torkelrogstad, line 188: Could just do `14.days` instead? No need to calculate it every time

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+    * */
+  def difficultyChangeInterval: Long = {
+    powTargetTimeSpan.toSeconds / powTargetSpacing
+  }
 }
 
-sealed abstract class BitcoinChainParams extends ChainParams
+sealed abstract class BitcoinChainParams extends ChainParams {
+
+  /** The targetted timespan between difficulty adjustments
```
torkelrogstad: If you do 
```
  /**
    * @inheritdoc
    */
```

you avoid copy pasting

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L73 mainnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L190 testnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L285 regtest]]
+    * */
+  override lazy val powTargetTimeSpan: Duration = {
+    val time = 14 * 24 * 60 * 60 //two weeks
+    time.seconds
+  }
+
+  /**
```
torkelrogstad, line 192: Same as scaladoc comment above

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
     createGenesisBlock(UInt32(1231006505),
                        UInt32(2083236893),
                        UInt32(0x1d00ffff),
                        Int32.one,
                        Satoshis(Int64(5000000000L)))
 
-  override def base58Prefixes: Map[Base58Type, ByteVector] =
+  override lazy val base58Prefixes: Map[Base58Type, ByteVector] =
     Map(
       Base58Type.PubKeyAddress -> BitcoinSUtil.decodeHex("00"),
```
torkelrogstad:
```scala
import scodec.bits.HexStringSyntax
Base58Type.PubKeyAddress -> hex"00" 
```
?

Hex is checked for validity at compile time

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -173,20 +251,55 @@ object MainNetChainParams extends BitcoinChainParams {
...
+    val limit = new BigInteger(1, bytes)
+    limit
+  }
+
+  /**
+    * The minimum amount of chain work on mainnet as of 2019/03/20
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L94 mainnet chain work]]
+    */
+  override lazy val minimumChainWork: BigInteger = {
+    val bytes = ByteVector.fromValidHex(
```
torkelrogstad: `hex"0000000000000000000000000000000000000000051dc8b82f450202ecb3d471"`

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -173,20 +251,55 @@ object MainNetChainParams extends BitcoinChainParams {
...
     createGenesisBlock(UInt32(1296688602),
                        UInt32(414098458),
                        UInt32(0x1d00ffff),
                        Int32.one,
                        Satoshis(Int64(5000000000L)))
 
-  override def base58Prefixes: Map[Base58Type, ByteVector] =
+  override lazy val base58Prefixes: Map[Base58Type, ByteVector] =
     Map(
       Base58Type.PubKeyAddress -> BitcoinSUtil.decodeHex("6f"),
```
torkelrogstad: `hex` syntax from above?

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -200,18 +313,75 @@ object TestNetChainParams extends BitcoinChainParams {
...
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L189 testnet pow limit]]
+    */
+  override lazy val powLimit: BigInteger = MainNetChainParams.powLimit
+
+  /** Minimum amount of chain work on the test network
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L211 testnet min chain work]]
+    * */
+  override lazy val minimumChainWork: BigInteger = {
+    val bytes = ByteVector.fromValidHex(
+      "00000000000000000000000000000000000000000000007dbe94253893cbd463")
```
torkelrogstad: `hex` syntax from above?

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -143,23 +146,98 @@ sealed abstract class ChainParams {
...
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L74 mainnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L191 testnet]]
+    * [[https://github.com/bitcoin/bitcoin/blob/a083f75ba79d465f15fddba7b00ca02e31bb3d40/src/chainparams.cpp#L286 regtest]]
+    * @return
+    */
+  override lazy val powTargetSpacing: Int = {
+    10 * 60 //what are these magic numbers?
+  }
+
+  /** The best chain should have this amount of work */
```
torkelrogstad, line 200: Any reason for this to be a `java.math.BigInteger` and not `scala.math.BigInt`? BigInt has implicit conversions from Int

torkelrogstad (commented):
General comment on overridden vals/defs: If you add `@inheritdoc` in the Scaladoc you get the doc from the parent value, in addition to whatever you write in the child member

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
Christewart: This is actually difficulty adjustment interval as defined here: 

https://github.com/bitcoin/bitcoin/blob/eb7daf4d600eeb631427c018a984a77a34aca66e/src/consensus/params.h#L75

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
Christewart: Not really, just thought it would be more friendly to java people.

PR #389, opened 2019-03-21
Refactor bench module

torkelrogstad:
I originally planned on creating benchmarks for derivation of private and public keys. After spending an awful long time in the Scalatest weeds trying to figure out what a nice benchmark framework would like, it turned out that deriving keys had negligible cost, as in a couple of millis  for deriving hundreds of thousands of levels deep. I therefore ended up not including any benchmarks for key derivation, as it seemed somewhat pointless.


torkelrogstad:
Closing this, no real point in pursuing this until #442 is resolved.

PR #391, opened 2019-03-25
Initial wallet: import UTXO and spend it

torkelrogstad:
> Note: the last commit here isn't really part of the node/wallet project, but belongs in the bitcoind RPC client. Opening a PR for that as well, it can be removed once we get things merged in/rebased. 

Functionality: 

- Initialize wallet - with or without custom entropy
- Generate P2WPKH addresses
- Add UTXO to wallet DB
- Send money to an address by spending imported UTXOs - caveat: very naive implementation with no real coin selection (we fail if we don't have a single UTXO large enough), also doesn't add change UTXO to DB. Does the TX builder API provide a way of figuring out which output is change? Shoulnd't be hard to implement anyways. 

Other lacking areas: 
 - Test fixtures. Not really sure how to handle this. There is a commented-out attempt in the wallet test trait. The issue is that we need multiple DAOs in the test, as well as an initialized wallet
- The `AddressDb` case class in particular is also very lacking. Right now it works because we only handle P2WPKH addresses, but it should be turned into an ADT with different addresses having well-defined case classes (without room for erroring by forgetting to supply a reedem script, for example). 
- A ton of other things, some of which are documented by adding `todo` comments.


wallet-test/src/test/scala/org/bitcoins/wallet/WalletTest.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  val accountDAO = AccountDAO(dbConfig, chainParams)
+  val addressDAO = AddressDAO(dbConfig, chainParams)
+  val mnemonicDAO = MnemonicCodeDAO(dbConfig)
+
+  it should "create a new wallet" in {
+    val passphrase = "foobar"
+
+    for {
+      _ <- Wallet.initialize(chainParams, passphrase)
```
torkelrogstad: This test crashes. There's some weirdness going in different databases being connected to in setup (`beforeAll` of BitcoinSWalletTest`) and the actual tests being run

wallet-test/src/test/scala/org/bitcoins/wallet/models/AddressDAOTest.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.bitcoins.core.crypto.ECPublicKey
+import org.bitcoins.core.crypto.bip44.{BIP44ChainType, BIP44Coin, BIP44Path}
+import org.bitcoins.core.protocol.P2SHAddress
+import org.bitcoins.core.util.CryptoUtil
+import org.bitcoins.wallet.util.BitcoinSWalletTest
+
+class AddressDAOTest extends BitcoinSWalletTest {
+  private val addressDAO = AddressDAO(dbConfig, chainParams)
+
+  it should "insert and read an address into the database" in {
```
torkelrogstad: This test should crash, but doesn't... It should give an error message that the foreign key reference isn't found in the accounts table

Christewart:
As @nkohen said in slack, it would be a good idea to start creating types for `EncryptedPayload`, `DecryptedPayload`, `Password` etc

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -17,6 +17,11 @@ sealed abstract class Address {
   /** The string representation of this address */
   def value: String
 
+  override def equals(obj: Any): Boolean = obj match {
+    case addr: Address => value == addr.value
+    case _: Any        => false
```
Christewart, line 22: What's this?

db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
```diff
@@ -17,18 +17,19 @@ abstract class DbManagement {
...
-  def dropTable(
-      table: TableQuery[_ <: Table[_]],
-      dbConfig: DbConfig): Future[Unit] = {
+  def dropTable(table: TableQuery[_ <: Table[_]], dbConfig: DbConfig)(
+      implicit ec: ExecutionContext): Future[Unit] = {
     val database = dbConfig.database
-    val result = database.run(table.schema.drop)
-    result
+    val tableName = table.baseTableRow.tableName
+    val query = sqlu"DROP TABLE IF EXISTS #$tableName;"
```
Christewart: i upgraded to slick 3.3.0, which gives us a `dropIfExists` method

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
```diff
@@ -0,0 +1,40 @@
...
+  /** Timeout for async operations */
+  protected val timeout: FiniteDuration = 10.seconds
+
+  protected val networkParam: RegTest.type = WalletTestUtil.networkParam
+
+  override protected def beforeAll(): Unit = {
+    Await.result(WalletDbManagement.dropAll(dbConfig), timeout)
+    Await.result(WalletDbManagement.createAll(dbConfig), timeout)
+  }
+
```
Christewart, line 80: I think we should create a `withWallet` test fixture similar to how i created fixtures in the chain project https://github.com/bitcoin-s/bitcoin-s-core/blob/node/chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala#L51

wallet/src/main/resources/application.conf
```diff
@@ -0,0 +1,40 @@
...
+  db {
+    url="jdbc:sqlite:"${regtestDb.dbPath}${regtestDb.dbName}
+  }
+}
+
+unittestDb = ${commonNodeDbSettings}
+unittestDb {
+  dbPath = ${HOME}"/.bitcoin-s/unittest/"
+  db {
+    url="jdbc:sqlite:"${unittestDb.dbPath}${unittestDb.dbName}
```
Christewart, line 36: Let's consider changing this to an in memory sqlite database like i did in https://github.com/bitcoin-s/bitcoin-s-core/blob/node/chain/src/main/resources/reference.conf#L37

core/src/main/scala/org/bitcoins/core/protocol/Address.scala
nkohen: Java `Object`'s `equals` method takes an `Object` as a param and I think you're supposed to return false if the types don't match

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
torkelrogstad: Should the fixture pass in a 3-tuple of DAOs needed? That would cause DAO's that aren't used in the test to be passed in/created/dropped. Not necessarily an issue, but might be

core/src/main/scala/org/bitcoins/core/util/EitherUtil.scala
```diff
@@ -0,0 +1,15 @@
+package org.bitcoins.core.util
+
+import scala.concurrent.Future
+
+object EitherUtil {
+
+  def flattenFutureE[L, R](
```
nkohen, line 17: A short scaladoc would be nice

wallet-test/src/test/scala/org/bitcoins/wallet/WalletTest.scala
```diff
@@ -0,0 +1,41 @@
...
+      wallet <- Wallet.initialize(chainParams = chainParams,
+                                  dbConfig = dbConfig,
+                                  passphrase = passphrase)
+      addr <- wallet.getNewAddress()
+      otherAddr <- wallet.getNewAddress()
+      allAddrs <- addressDAO.findAll()
+    } yield {
+      logger.info(s"addr: $addr")
+      logger.info(s"otherAddr: $otherAddr")
+      assert(allAddrs.length == 2)
```
nkohen: shall we also check that
```
allAddrs.head == addr
allAddrs.last == otherAddr
```

wallet-test/src/test/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAOTest.scala
```diff
@@ -0,0 +1,27 @@
...
+        val outpoint = TransactionOutPoint(WalletTestUtil.sampleTxid,
+                                           WalletTestUtil.sampleVout)
+        val output = TransactionOutput(Bitcoins.one, WalletTestUtil.sampleSPK)
+        val privkeyPath = WalletTestUtil.sampleBip44Path
+        val utxo =
+          UTXOSpendingInfoDb(None, outpoint, output, privkeyPath, None, None)
+        utxoDAO.create(utxo)
+      }
+      read <- utxoDAO.read(created.id.get)
+    } yield assert(read.contains(created))
```
nkohen, line 30: I think this is the kind of situation where I prefer reading futures

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -44,23 +61,82 @@ sealed abstract class Wallet extends UnlockedWalletApi with BitcoinSLogger {
...
+          outPoint <- outPointE
+          address <- addressE
+        } yield {
+          val utxo =
+            UTXOSpendingInfoDb(None, outPoint, output, address.path, None, None)
+          utxoDAO.create(utxo).map(_ => Right(this.asInstanceOf[WalletApi]))
+        }
+      EitherUtil.flattenFutureE(nestedWalletE)
+    }
+  }
```
nkohen, line 138: I think this code becomes much cleaner if you change `Either[AddUtxoError, T] -> Try[T]` and `Future[Either[AddUtxoError, T]] -> Future[T]` since `Future`s can fail (They're little more than (mutable) wrappers for `Try[Option[_]]`)

torkelrogstad: Two issues with that:
1) `Try[T]` lacks well defined errors. With `AddUtxoError` being an ADT, consumers can be sure they handle (or at least are aware of) every error
2) `Future` doesn't force users to handle the error case, and the error is also untyped (like `Try`)

Although I agree that handling nested `Either`s and `Future`s isn't ideal. Something like `EitherT` from Cats would handle this better: https://typelevel.org/cats/datatypes/eithert.html

So eventually we should implement some of that functionality, for cases like this

wallet-test/src/test/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAOTest.scala
torkelrogstad: Sorry, what do you mean?

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
```diff
@@ -0,0 +1,73 @@
...
+    val valueFromBitcoind = Bitcoins.one
+    val walletF =
+      Wallet.initialize(chainParams = chainParams,
+                        dbConfig = dbConfig,
+                        passphrase = passphrase)
+    val bitcoindF = BitcoindRpcTestUtil.startedBitcoindRpcClient()
+
+    val addUtxoF: Future[Unit] = for {
+      bitcoind <- bitcoindF
+      _ <- bitcoind.generate(101)
```
Christewart: `startedBitcoindRpcClient` already has blocks generated, so I don't think this is needed

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
```diff
@@ -0,0 +1,73 @@
...
+        val vout = voutOpt.getOrElse(
+          throw new IllegalArgumentException(
+            "Could not find ouput that spent to our address!"))
+
+        wallet.addUtxo(tx.hex, UInt32(vout.n))
+      }
+    } yield {
+      addUtxoRes match {
+        case Left(err: AddUtxoError) => fail(err)
+        case Right(_: WalletApi)     =>
```
Christewart: nothing being returned here?

wallet-test/src/test/scala/org/bitcoins/wallet/models/AccountDAOTest.scala
```diff
@@ -0,0 +1,22 @@
+package org.bitcoins.wallet.models
+
+import org.bitcoins.testkit.core.gen.CryptoGenerators
+import org.bitcoins.wallet.util.{BitcoinSWalletTest, WalletTestUtil}
+
+class AccountDAOTest extends BitcoinSWalletTest {
+
+  it should "insert and read an account into the database" in {
+    for {
```
Christewart, line 10: could change this into a property based test using 

```scala
forAll(CryptoGeneators.extPublicKey) { case xpub => ... }
```

wallet-test/src/test/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAOTest.scala
```diff
@@ -0,0 +1,32 @@
...
+  TransactionOutPoint,
+  TransactionOutput
+}
+import org.bitcoins.wallet.util.{BitcoinSWalletTest, WalletTestUtil}
+
+class UTXOSpendingInfoDAOTest extends BitcoinSWalletTest {
+  behavior of "UTXOSpendingInfoDAO"
+
+  it should "insert a UTXO and read it" in {
+    for {
```
Christewart, line 27: can use property based testing here

```scala 

forAll(TransactionGenerators.outPoint, TransactionGenerators.output) { ... } 

```

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+
+  def chainParams: ChainParams
+
+  val feeProvider: FeeProvider = _ =>
+    Future.successful(SatoshisPerByte(Satoshis.one))
+
+  /**
+    * @inheritdoc
+    */
+  override def lock: Future[WalletApi] = ???
```
Christewart: Should return `LockedWalletApi`

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+  /** Generates a new change address */
+  private def getNewChangeAddress(
+      accountIndex: Int = 0): Future[BitcoinAddress] = {
+    getNewAddressHelper(accountIndex, BIP44ChainType.Change)
+  }
+
+  override def sendToAddress(
+      address: BitcoinAddress,
+      amount: CurrencyUnit,
+      accountIndex: Int = 0): Future[Transaction] = {
```
Christewart, line 197: I think we should pass in `FeeUnit` as a parameter here

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+    for {
+      change <- getNewChangeAddress(accountIndex)
+      fee <- feeProvider.getFee(1)
+      walletUtxos <- listUtxos()
+      txBuilder <- {
+        val destinations: Seq[TransactionOutput] = List(
+          TransactionOutput(amount, address.scriptPubKey))
+
+        // currencly just grabs one utxos, throws if can't find big enough
+        // todo: implement coin selection
```
Christewart, line 206: we need to implement some sort of abstraction for this

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+      } */
+    } yield {
+      signed
+    }
+  }
+
+  /**
+    * @inheritdoc
+    */
+  override def unlock(passphrase: String): Future[Try[UnlockedWalletApi]] = ???
```
Christewart: Future[Try[]] is a bit redundant, just use `Future.failed` instead

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+
+// todo: create multiple wallets, need to maintain multiple databases
+object Wallet extends CreateWalletApi {
+  private case class WalletImpl(
+      mnemonicCode: MnemonicCode,
+      passphrase: String,
+      dbConfig: DbConfig,
+      chainParams: ChainParams)(implicit executionContext: ExecutionContext)
+      extends Wallet {
+    override implicit def ec: ExecutionContext = executionContext
```
Christewart: any reason we don't just do 

```scala
 private case class WalletImpl(
      mnemonicCode: MnemonicCode,
      passphrase: String,
      dbConfig: DbConfig,
      chainParams: ChainParams)(override implicit val ec: ExecutionContext)
      extends Wallet { ... } 
```

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+
+  def apply(
+      mnemonicCode: MnemonicCode,
+      passphrase: String = BIP39Seed.EMPTY_PASSWORD,
+      dbConfig: DbConfig,
+      chainParams: ChainParams)(implicit ec: ExecutionContext): Wallet =
+    WalletImpl(mnemonicCode, passphrase, dbConfig, chainParams)(ec)
+
+  override protected def initializeWithEntropy(
+      entropy: BitVector,
```
Christewart, line 267: need scaladoc here saying it is paramount that this is not empty, can we check some how to see if this isn't just uninitialized as a sanity check?

wallet/src/main/scala/org/bitcoins/wallet/api/AddUtxoError.scala
```diff
@@ -0,0 +1,25 @@
...
+
+/** Represents an error that might occur when adding an UTXO to the wallet */
+sealed trait AddUtxoError extends Error
+
+object AddUtxoError {
+
+  /**
+    * The provided vout index does not exist in the given transaction
+    */
+  case object VoutIndexOutOfBounds
```
Christewart: `final` on all of these

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+    * @inheritdoc
+    */
+  // override def createNewAccount: Future[Try[WalletApi]] = ???
+
+}
+
+// todo: create multiple wallets, need to maintain multiple databases
+object Wallet extends CreateWalletApi {
+  private case class WalletImpl(
+      mnemonicCode: MnemonicCode,
```
Christewart, line 247: So if I am understanding our current design correctly, we need to keep the mnemonic code in memory the entire time? 

We want to make sure this isn't kept around in memory longer than needed. Should we have this be an `xpriv` instead?

wallet/src/main/scala/org/bitcoins/wallet/api/FeeProvider.scala
```diff
@@ -0,0 +1,16 @@
+package org.bitcoins.wallet.api
+
+import org.bitcoins.core.wallet.fee.FeeUnit
+
+import scala.concurrent.Future
+
+trait FeeProvider {
```
Christewart: As i have said in slack, I think we should require external fees be provided via methods for apis that require fees for now. I think we should delete this

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -0,0 +1,132 @@
...
+    * times will return the same address, until it has
+    * received funds.
+    */
+  def getNewAddress(accountIndex: Int = 0): Future[BitcoinAddress]
+
+  /**
+    * Unlocks the wallet with the provided passphrase,
+    * making it possible to send transactions.
+    */
+  def unlock(passphrase: String): Future[Try[UnlockedWalletApi]]
```
Christewart: `Future[UnlockedWalletApi]` , the `Try` is redundant here

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -0,0 +1,132 @@
...
+    *
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account BIP44 account section]]
+    */
+  // def createNewAccount: Future[Try[WalletApi]]
+
+}
+
+trait UnlockedWalletApi extends WalletApi with LockedWalletApi {
+
+  def mnemonicCode: MnemonicCode
```
Christewart, line 102: Shouldn't this be a `ExtPrivateKey` instead of Mnemonic?

Christewart: I guess I now see that this later down in this file, why is this needed at all? We are storing redundant information here and we may forget this is hanging around in memory too.

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -0,0 +1,132 @@
...
+    val privVersion = ExtKeyPrivVersion.fromNetworkParameters(networkParameters)
+    seed.toExtPrivateKey(privVersion)
+  }
+
+  /**
+    * Locks the wallet. After this operation is called,
+    * all sensitive material in the wallet should be
+    * encrypted and unaccessible
+    */
+  def lock: Future[WalletApi]
```
Christewart: `LockedWalletApi`

wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.bitcoins.db.{CRUD, DbConfig, SlickUtil}
+import slick.dbio.Effect
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.TableQuery
+import slick.sql.SqlAction
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class AddressDAO(dbConfig: DbConfig)(
+    implicit executionContext: ExecutionContext)
```
Christewart:
```
override implicit val ec: ExecutionContext)
```

and then remove `ec` below

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTable.scala
```diff
@@ -0,0 +1,147 @@
...
+      xpriv: ExtPrivateKey,
+      path: BIP44Path,
+      np: NetworkParameters): AddressDb = {
+
+    val xprivAtPath: ExtPrivateKey = xpriv.deriveChildPrivKey(path)
+    val pub = xprivAtPath.key.publicKey
+    val witnessSpk = P2WPKHWitnessSPKV0(pub)
+    val scriptWitness = P2WPKHWitnessV0(pub)
+    val addr = Bech32Address(witnessSpk, np)
+    AddressDb(path,
```
Christewart: named args

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTable.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  def accountChainType: Rep[BIP44ChainType] =
+    column[BIP44ChainType]("bip44_chain_type")
+
+  def addressIndex: Rep[Int] = column[Int]("address_index")
+
+  def address: Rep[BitcoinAddress] =
+    column[BitcoinAddress]("address", O.PrimaryKey)
+
+  def ecPublicKey: Rep[ECPublicKey] = column[ECPublicKey]("pubkey")
```
Christewart, line 73: There isn't a one to one relationship between public keys and addresses, for instance multisig. This design needs work

wallet/src/main/scala/org/bitcoins/wallet/models/MnemonicCodeDAO.scala
```diff
@@ -0,0 +1,50 @@
...
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.TableQuery
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * @note This DAO does not extend [[org.bitcoins.db.CRUD CRUD]]
+  *       because it's not the intention to create multiple mnenonics
+  *
+  *       todo: would this be better as a flat file, and not a DB?
```
Christewart: I think so, we want to make it as easy as possible to backup this file

wallet/src/main/scala/org/bitcoins/wallet/models/MnemonicCodeTable.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import org.bitcoins.core.crypto.MnemonicCode
+import scodec.bits.ByteVector
+import slick.jdbc.SQLiteProfile.api._
+
+import scala.util.Try
+
+case class EncryptedMnemonic(value: ByteVector) {
+
+  // todo: use AES merged into master
```
Christewart: this is now merged

wallet/src/main/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAO.scala
```diff
@@ -0,0 +1,19 @@
+package org.bitcoins.wallet.models
+
+import org.bitcoins.db.{CRUDAutoInc, DbConfig}
+import slick.jdbc.SQLiteProfile.api._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class UTXOSpendingInfoDAO(dbConfig: DbConfig)(
+    implicit executionContext: ExecutionContext)
```
Christewart: just override `ec` here

wallet-test/src/test/scala/org/bitcoins/wallet/models/AddressDAOTest.scala
```diff
@@ -0,0 +1,52 @@
...
+    AddressDb(path, pubkey, hashedPubkey, address, None, ScriptType.SCRIPTHASH)
+  }
+
+  behavior of "AddressDAO"
+
+  it should "fail to insert and read an address into the database without a corresponding account" in {
+    val readF = {
+      val addressDb = getAddressDb(WalletTestUtil.firstAccountDb)
+      addressDAO.create(addressDb)
+    }.map(_ => succeed)
```
nkohen: is this map necessary?

wallet-test/src/test/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAOTest.scala
nkohen: I think this would read nicer as
```
val outpoint = TransactionOutPoint(WalletTestUtil.sampleTxid, WalletTestUtil.sampleVout)
val output = TransactionOutput(Bitcoins.one, WalletTestUtil.sampleSPK)
val privkeyPath = WalletTestUtil.sampleBip44Path
val utxo = UTXOSpendingInfoDb(id = None,
                              outPoint = outpoint,
                              output = output,
                              privKeyPath = privkeyPath,
                              redeemScriptOpt = None, // todo test this properly
                              scriptWitnessOpt = None) // todo test this properly

val createdF = utxoDAO.create(utxo)
val readF = createdF.flatMap(utxoDAO.read(_.id.get))

readF.map(assert(_.contains(created)))
```
or something closer to this

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
nkohen: So I think what you should try for this fixutre business is having an ADT of case classes that wrap the fixtures that are needed and then use `FixtureTag`s to specify which test should receive which fixture (feel free to ping me about this)

nkohen: Basically `withFixture` should look at the tag and determine which fixture to provide, and then for the test you do a normal `in {` but then you have two cases rather than a normal lambda, one for the case class you expect, and then a catch all for anything else that leads to a fail case.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
nkohen: In this case, what do you think of creating a wrapper type specific to our use case? I.e. instead of using `Try`, creating an ADT like it with a success case and a failure case where the failure wraps an `AddUtxoError`? I always get super confused when looking at maps on `Either` because it isn't obvious to me which one is the "success" case.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -0,0 +1,279 @@
...
+
+  /**
+    * @inheritdoc
+    */
+  override def lock: Future[WalletApi] = ???
+
+  /**
+    * @inheritdoc
+    */
+  override def addUtxo(
```
nkohen, line 89: I think you've done a nice job of sectioning off this method into logical pieces, I think it would be nice to make them separate defs to make this easier to reason about.

wallet/src/main/scala/org/bitcoins/wallet/api/CreateWalletApi.scala
```diff
@@ -0,0 +1,93 @@
...
+      passphrase: String
+  )(implicit executionContext: ExecutionContext): Future[UnlockedWalletApi] =
+    initializeWithEntropy(entropy = entropy,
+                          chainParams = chainParams,
+                          passphrase = passphrase,
+                          dbConfig = None)
+
+  /**
+    * $initializeWithEnt
+    */
```
nkohen, line 75: the scaladocs in this file seem to be missing

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -0,0 +1,132 @@
...
+    * most recent account has no transaction history, as per
+    * BIP44
+    *
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account BIP44 account section]]
+    */
+  // def createNewAccount: Future[Try[WalletApi]]
+
+}
+
+trait UnlockedWalletApi extends WalletApi with LockedWalletApi {
```
nkohen: I think that since `LockedWalletApi extends WalletApi`, extending `WalletApi` here is redundant

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
torkelrogstad: This is similar to returning `()`. `fail(...)` fails the test even if it's not at the end, similar to throwing an error (it's got type `Nothing`). so `fail` + `()` is type `Unit`. Adding a comment and returning `()` explicitly

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: We cannot have the `override` modifier. That turns `WalletImpl` into `ExecutionContext => Wallet.WalletImpl`. But I agree it's cleaner to give up the override word to save some lines

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
torkelrogstad: What do you mean by storing redundant information? The xpriv is derivable from the mnemonic, so we only ask users to provide the mnemonic

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: 1) Keeping an mnemonic and a xpriv in memory is equivalent security wise. If they are leaked the user is screwed regardless:-)
2) The use case for having the mnemonic readily available is displaying it to the user of an app.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: I ended up doing `Either` internally, and ADT for public facing API. What do you think about that?

wallet/src/main/scala/org/bitcoins/wallet/api/CreateWalletApi.scala
torkelrogstad: If you hover over them in Intellij they should be there:-) I've defined two scaladoc macros at the top of the file

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: Also did this in `initializeWithEntropy`

wallet/src/main/scala/org/bitcoins/wallet/api/CreateWalletApi.scala
Christewart: FWIW this doesn't work for me either, i'm guessing we have some different settings in intellij

testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
```diff
@@ -0,0 +1,133 @@
+package org
+
```
Christewart: Looks like the packaging got messed up here

testkit/src/main/scala/org/bitcoins/testkit/fixtures/EmptyFixture.scala
```diff
@@ -0,0 +1,13 @@
+package org
```
Christewart: same packaging comment here

PR #395, opened 2019-03-28
Add AES encryption and decryption

torkelrogstad:
Test suite might be a bit on the light side, and I'd also prefer to flesh out error cases in both encrypting and decrypting. But TBH I'm not quite sure how to do that. 

Edit: 

Looking at Bitcoin Core [source](https://github.com/bitcoin/bitcoin/blob/fa6180188b8ab89af97860e6497716405a48bab6/src/crypto/aes.h) it looks like they also use AES with CBC block cipher


core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,165 @@
...
+                                           salt = salt,
+                                           iterationCount = ITERATIONS,
+                                           derivedKeyLength = KEY_SIZE)
+    new SecretKeySpec(tmp.getEncoded, "AES")
+  }
+}
+
+sealed trait AesEncryptionException extends Exception
+
+sealed trait AesDecryptionException extends Exception
```
nkohen, line 164: `RuntimeException`?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,165 @@
...
+                new IvParameterSpec(initializationVector.toArray))
+    cipher
+  }
+
+  /**
+    * Decrypts the provided data
+    */
+  def decrypt(
+      encrypted: AesEncryptedData,
+      password: String): Either[AesDecryptionException, ByteVector] = {
```
nkohen: We should really consider making a type for passwords that wraps a String so that we never accidentally pass a password anywhere it isn't expected.

core/src/main/scala/org/bitcoins/core/crypto/PBKDF2.scala
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+  * @define keyStretch Derives the provided value and salt to a secret key
+  *                    using the PBKDF2 key derivation function
+  *
+  * Utilities related to the PBKDF2 key derivation function
+  */
+object PBKDF2 {
+
+  private val PSEUDO_RANDOM_FUNCTION = "PBKDF2WithHmacSHA512"
```
nkohen, line 20: Could you add a comment with a link to where this comes from?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,165 @@
...
+                                           derivedKeyLength = KEY_SIZE)
+    new SecretKeySpec(tmp.getEncoded, "AES")
+  }
+}
+
+sealed trait AesEncryptionException extends Exception
+
+sealed trait AesDecryptionException extends Exception
+
+case object BadPasswordException
```
Christewart: Let's create a wrapper object for these? I.e. 

```scala
object AESException { 
   final case object BadPasswordException ...
   
   final case object EmptyPasswordException ...
}

File: core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+    forAll(NumberGenerator.bytevector) { bytes =>
+      val encrypted = AesCrypt.encryptExc(bytes, password)
+      val decryptedE = AesCrypt.decrypt(encrypted, badPassword)
+      decryptedE match {
+        case Right(decrypted) =>
+          assert(decrypted != bytes)
+        case Left(exc) => assert(exc == BadPasswordException)
+
+      }
+      succeed
```
Christewart: why is this needed?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  private def encryptionCipher(secret: SecretKey): Cipher = {
+    val cipher = getCipher
+    cipher.init(Cipher.ENCRYPT_MODE, secret)
+    cipher
+  }
+
+  lazy private val secureRandom: SecureRandom = new SecureRandom
```
Christewart: i'm always leary of caching PRNG, let's not do this.

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+    * Generates a random hex string of length 30
+    */
+  private def getSalt: String = {
+    val stringBuilder = new StringBuilder
+    (0 until 30).foreach { _ =>
```
Christewart: i'd prefer this 

```scala
    val array = Array[Byte](32)
    val rand = new SecureRandom(array)
    rand.nextBytes(array)
    require(array.exists(_ != 0), s"Array for salt must be initialized")
    ByteVector(array).toHex
```

I'm not sure what the optimal salt size is, the SO question you link to says it needs to be 8 bytes. The question is almost 10 years old though, so those security practices may be out dated.

torkelrogstad: Is there any reason we want to provide `array` as seed to the `SecureRandom`?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
torkelrogstad:
```
illegal inheritance; superclass BadPaddingException
[error]  is not a subclass of the superclass RuntimeException
[error]  of the mixin trait AesDecryptionException
[error]       with AesDecryptionException
```

Do you know how to fix that?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -0,0 +1,173 @@
...
+  private def encryptionCipher(secret: SecretKey): Cipher = {
+    val cipher = getCipher
+    cipher.init(Cipher.ENCRYPT_MODE, secret)
+    cipher
+  }
+
+  private def getSecureRandom: SecureRandom = new SecureRandom
+
+  /**
+    * Generates a random hex string of length 30
```
Christewart: commment fix

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import scodec.bits.HexStringSyntax
+
+class AesCryptTest extends BitcoinSUnitTest {
+  behavior of "AesEncrypt"
+
+  val password = AesPassword("PASSWORD")
+  val badPassword = AesPassword("BAD_PASSWORD")
+  val emptyPassword = AesPassword("")
+
```
Christewart, line 14: We are missing a couple of hard coded test vectors with expected ecrypted output and then the corresponding decrypted output. These are really useful for sanity tests. Property tests can be misleading some times as you can implement the encrypt and decrypt functionality in just the right way so that your tests pass :-)

PR #396, opened 2019-03-29
Fix issue where it returns a false positive

ghost:
seems like i have some funky commits from pulling commits upstream. i think merge squash would remove them


Christewart:
Do you mind just checking out a fresh copy of master and fixing this?

PR #400, opened 2019-04-08
Populated Header Fixture

nkohen:
Downloaded real mainnet headers into file that is read by a new test fixture that populates the header table.

Currently populating the database takes like 2 minutes :/


chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -59,6 +67,65 @@ trait ChainUnitTest
...
+        logger.error(s"Failed to parse headers from block_headers.json: $err")
+        Future.failed(new RuntimeException(err.toString))
+      case JsSuccess(headers, _) =>
+        val dbHeaders = headers.zipWithIndex.map {
+          case (header, height) =>
+            BlockHeaderDbHelper.fromBlockHeader(height, header)
+        }
+
+        @tailrec
+        def splitIntoBatches(
```
Christewart, line 132: so just an FYI, when we are doing initial block download (IBD) we do get headers in batches of 2,000 headers. We should probably make sure that it is isn't terrible for the database to handle 2,000 header batches, and if it is, we might want to consider moving this into the `BlockHeaderDAO` createAll method to batch things

https://bitcoin.org/en/developer-reference#getheaders

PR #401, opened 2019-04-09
Add integration test between bitcoind <-> zmq <-> bitcoin-s-chain pro…

Christewart:
…ject. Test that we can relay a header from bitcoind over zmq into the bitcoin-s chain project. Redo ZmqConfig to use InetSocketAddress

This PR

- Adds a fixture for `bitcoindZmqChainHandler`
- Changes ZmqConfig to use InetSocketAddress

This makes it much easier to test our integration with bitcoind and our zmq project. This PR should be looked as a fundamental PR for making sure we correctly implement chain handling logic in bitcoin-s

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -1,11 +1,19 @@
 package org.bitcoins.chain.blockchain
 
+import akka.actor.ActorSystem
 import org.bitcoins.chain.util.ChainUnitTest
+import org.bitcoins.rpc.util.RpcUtil
 import org.bitcoins.testkit.chain.BlockHeaderHelper
+
+import scala.concurrent.Future
 
 class ChainHandlerTest extends ChainUnitTest {
 
   behavior of "ChainHandler"
+
+  implicit val system = ActorSystem("ChainHandlerTest")
+
+  import system.dispatcher
 
   it must "process a new valid block header, and then be able to fetch that header" in withChainHandler {
     case chainHandler: ChainHandler =>
@@ -17,6 +25,34 @@ class ChainHandlerTest extends ChainUnitTest {
         processedHeaderF.flatMap(_.getHeader(newValidHeader.hashBE))
 
       foundHeaderF.map(found => assert(found.get == newValidHeader))
+  }
+
+  it must "peer with bitcoind via zmq and have blockchain info relayed" in bitcoindZmqChainHandler {
+    case bitcoindChainHandler: BitcoindChainHandler =>
+      val bitcoind = bitcoindChainHandler.bitcoindRpc
+
+      val chainHandler = bitcoindChainHandler.chainHandler
+
+      val assert1F = chainHandler.getBlockCount
+        .map(count => assert(count == 0))
+
+      //mine a block on bitcoind
+      val generatedF = assert1F.flatMap(_ => bitcoind.generate(1))
+
+      generatedF.flatMap { headers =>
+        val hash = headers.head
+        val foundHeaderF: Future[Unit] = {
+          //test case is totally async since we
+          //can't monitor processing flow for zmq
+          //so we just need to await until we
+          //have fully processed the header
+          RpcUtil.awaitConditionF(() =>
+            chainHandler.getHeader(hash).map(_.isDefined))
+        }
+        val bitcoinsHeader =
+          foundHeaderF.flatMap(_ => chainHandler.getHeader(hash))
+        bitcoinsHeader.map(header => assert(header.get.hashBE == hash))
+      }
   }
 
 }
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -27,7 +27,7 @@ class ChainHandlerTest extends ChainUnitTest {
       foundHeaderF.map(found => assert(found.get == newValidHeader))
   }
 
-  it must "peer with bitcoind via zmq and have blockchain info relayed" in bitcoindZmqChainHandler {
+  it must "peer with bitcoind via zmq and have blockchain info relayed" in withBitcoindZmqChainHandler {
     case bitcoindChainHandler: BitcoindChainHandler =>
       val bitcoind = bitcoindChainHandler.bitcoindRpc
 
@@ -49,9 +49,11 @@ class ChainHandlerTest extends ChainUnitTest {
           RpcUtil.awaitConditionF(() =>
             chainHandler.getHeader(hash).map(_.isDefined))
         }
-        val bitcoinsHeader =
-          foundHeaderF.flatMap(_ => chainHandler.getHeader(hash))
-        bitcoinsHeader.map(header => assert(header.get.hashBE == hash))
+
+        for {
+          _ <- foundHeaderF
+          header <- chainHandler.getHeader(hash)
+        } yield assert(header.get.hashBE == hash)
       }
   }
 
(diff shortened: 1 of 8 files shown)


chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
 
     dropTableP.future.flatMap(_ => testExecutionF)
   }
+
+  /** Represents a bitcoind instance paired with a chain handler via zmq */
+  case class BitcoindChainHandler(
+      bitcoindRpc: BitcoindRpcClient,
+      chainHandler: ChainHandler)
+
+  def bitcoindZmqChainHandler(test: BitcoindChainHandler => Future[Assertion])(
```
nkohen: I like having the convention of having fixture methods start with `with`

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -19,4 +27,32 @@ class ChainHandlerTest extends ChainUnitTest {
...
+          //so we just need to await until we
+          //have fully processed the header
+          RpcUtil.awaitConditionF(() =>
+            chainHandler.getHeader(hash).map(_.isDefined))
+        }
+        val bitcoinsHeader =
+          foundHeaderF.flatMap(_ => chainHandler.getHeader(hash))
+        bitcoinsHeader.map(header => assert(header.get.hashBE == hash))
+      }
+  }
```
nkohen, line 58: I think this test might be nicer with a `for`

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+      implicit system: ActorSystem): Future[Assertion] = {
+    val instance = BitcoindRpcTestUtil.instance()
+    val bitcoindF = {
+      val bitcoind = new BitcoindRpcClient(instance)
+      bitcoind.start().map(_ => bitcoind)
```
nkohen: If we don't have something like this already, we should add a method to the rpc client's companion object that does this

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+        _ <- zmqSubscriberF
+        bitcoind <- bitcoindF
+      } yield BitcoindChainHandler(bitcoind, chainHandler)
+
+      val testExecutionF = bitcoindChainHandlerF
+        .flatMap(test(_))
+
+      testExecutionF.onComplete { _ =>
+        //kill bitcoind
+        bitcoindChainHandlerF.flatMap { bch =>
```
nkohen, line 160: should this `flatMap` be an `onComplete`?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+
+  def bitcoindZmqChainHandler(test: BitcoindChainHandler => Future[Assertion])(
+      implicit system: ActorSystem): Future[Assertion] = {
+    val instance = BitcoindRpcTestUtil.instance()
+    val bitcoindF = {
+      val bitcoind = new BitcoindRpcClient(instance)
+      bitcoind.start().map(_ => bitcoind)
+    }
+    val zmqRawBlockUriF = bitcoindF.map(_ => instance.zmqConfig.rawBlock)
+    val f: ChainHandler => Future[Assertion] = { chainHandler: ChainHandler =>
```
nkohen: I think it might be nice if this block were made a `def` above this one

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+
+  /** Represents a bitcoind instance paired with a chain handler via zmq */
+  case class BitcoindChainHandler(
+      bitcoindRpc: BitcoindRpcClient,
+      chainHandler: ChainHandler)
+
+  def bitcoindZmqChainHandler(test: BitcoindChainHandler => Future[Assertion])(
+      implicit system: ActorSystem): Future[Assertion] = {
+    val instance = BitcoindRpcTestUtil.instance()
+    val bitcoindF = {
```
torkelrogstad: could do 
```scala
val bitcoindF =  BitcoindRpcTestUtil.startedBitcoindRpcClient()
```

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+    val f: ChainHandler => Future[Assertion] = { chainHandler: ChainHandler =>
+      val handleRawBlock: ByteVector => Unit = { bytes: ByteVector =>
+        val block = Block.fromBytes(bytes)
+        chainHandler.processHeader(block.blockHeader)
+
+        ()
+      }
+
+      val zmqSubscriberF = zmqRawBlockUriF.map { uriOpt =>
+        val socket = uriOpt.get
```
torkelrogstad, line 138: Looking at what config `BitcoindRpcTestUtil.instance()` uses this is safe, but kinda smelly that we have to do a `.get` here. Not necessarily in the scope of this PR, though

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+        val z =
+          new ZMQSubscriber(socket, None, None, None, Some(handleRawBlock))
```
torkelrogstad: Named args?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -94,4 +99,59 @@ trait ChainUnitTest
...
+        z.start()
+        Thread.sleep(1000)
```
torkelrogstad, line 146: Makes me think that `z.start()` should return a `Future[Unit]` instead of `Unit`

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
Christewart: This is not the case, as we need to construct from the gensis block. `startedBitcoindRpcClient()` generates 100 blocks for us before returning so that the client has money in the wallet.

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
Christewart: Filed an issue: https://github.com/bitcoin-s/bitcoin-s-core/issues/402

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
Christewart: No, this isn't dependent on the `testExecutionF`, only on starting the bitcoind instance so this should be fine

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -19,4 +27,34 @@ class ChainHandlerTest extends ChainUnitTest {
...
+
+      generatedF.flatMap { headers =>
+        val hash = headers.head
+        val foundHeaderF: Future[Unit] = {
+          //test case is totally async since we
+          //can't monitor processing flow for zmq
+          //so we just need to await until we
+          //have fully processed the header
+          RpcUtil.awaitConditionF(() =>
+            chainHandler.getHeader(hash).map(_.isDefined))
```
nkohen, line 50: nit: Intermediate val here and a more descriptive name than `assert1F` such as `noBlocksF` would be nice :)

PR #403, opened 2019-04-10
Test Fixtures

nkohen:
Made all tests extending ChainUnitTest use test fixtures rather than an ad-hoc method.

This scheme will become slightly more complicated when one test file requires multiple types of test fixtures but this is achievable using tags (and we'll deal with it when it comes).

The first commit implements fixtures using casting while the second commit replaces casting with a single `ClassTag` context bound on `ChainUnitTest` (which allows for a nice test failed error message rather than having `withFixture` fail on a `ClassCastException`), I'm open to either approach.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
@@ -3,11 +3,7 @@ package org.bitcoins.chain.util
 import org.bitcoins.chain.api.ChainApi
 import org.bitcoins.chain.blockchain.{Blockchain, ChainHandler}
 import org.bitcoins.chain.db.ChainDbManagement
-import org.bitcoins.chain.models.{
-  BlockHeaderDAO,
-  BlockHeaderDb,
-  ChainTestFixture
-}
+import org.bitcoins.chain.models.{BlockHeaderDAO, BlockHeaderDb}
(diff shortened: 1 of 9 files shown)


torkelrogstad (approved):

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -48,24 +55,30 @@ trait ChainUnitTest
...
-  def withBlockHeaderDAO(
-      test: BlockHeaderDAO => Future[Assertion]): Future[Assertion] = {
+  def withBlockHeaderDAO(test: OneArgAsyncTest): FutureOutcome = {
 
     val genesisHeaderF = setupHeaderTableWithGenesisHeader()
 
     val blockHeaderDAOF = genesisHeaderF.map(_ => blockHeaderDAO)
-    val testExecutionF = blockHeaderDAOF.flatMap(test(_))
+    val testExecutionF = blockHeaderDAOF.flatMap {
+      case dao: FixtureType => test(dao).toFuture
```
Christewart: this is really frustrating that this doesn't just work :tm:

Christewart: So basically for every fixture type we will need something like this, and this PR seems to add more LOC. I'm skeptical

PR #419, opened 2019-04-15
Multi fixture file

nkohen:
Implements a framework for having as many different fixtures used as needed in a single test file.

In order to use this, one must `override type FIxtureParam = ChainFixture` and then tag each test with the correct `FixtureTag` and begin each test with `inFixtured { case ExpectedFixture(...) =>`

Optionally, one can also `override def defaultTag` if the plurality of tests use some fixture other than `Empty`.

I have not implemented this but we could make using this framework mandatory by adding `override type FixtureParam = ChainFixture` in `ChainUnitTest.scala`. Doing this would get rid of all casting and all `with...` defs other than `withChainFixture`. On the other hand it would add a tiny bit to test files that are only using one fixture. Thoughts?

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
@@ -99,7 +99,7 @@ trait ChainUnitTest
     * All untagged tests will be given this tag. Override this if you are using
     * ChainFixture and the plurality of tests use some fixture other than Empty.
     */
-  def defaultTag: FixtureTag = FixtureTag.Empty
+  val defaultTag: FixtureTag = FixtureTag.Empty
 
   /**
     * This ADT represents all Chain test fixtures. If you set this type to be your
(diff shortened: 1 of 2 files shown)


chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -59,6 +59,150 @@ trait ChainUnitTest
...
+        case GenisisChainHandler.name     => GenisisChainHandler
+        case BitcoindZmqChainHandlerWithBlock.name =>
+          BitcoindZmqChainHandlerWithBlock
+        case _: String =>
+          throw new IllegalArgumentException(s"$tag is not a valid tag")
+      }
+    }
+  }
+
+  def defaultTag: FixtureTag = FixtureTag.Empty
```
Christewart: should be val

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -59,6 +59,150 @@ trait ChainUnitTest
...
+      val destroyedF =
+        fixtureF.flatMap(fixture => ChainFixture.destroy(fixture))
+
+      destroyedF.map(_ => outcome)
+    }
+
+    new FutureOutcome(fixtureTakeDownF)
+  }
+
+  final class SugaryItVerbStringTaggedAs(
```
Christewart, line 184: Can we get a comment on the methods and on this class? I have no idea what it is meant to do. This seems to be where the test case actually gets run?

PR #425, opened 2019-04-16
Conf files

torkelrogstad:
1) Adds logback conf for wallet-test
2) Adds DB conf resolution that works across projects. Prior to this PR all projects loaded the same conf file, varying between `nodedb.sqlite` and `chaindb.sqlite`.


nkohen (approved):

torkelrogstad:
Superseded by #434

Issue #428, opened 2019-04-17
Blockchain builder


Christewart:
The current [`Blockchain`](https://github.com/bitcoin-s/bitcoin-s-core/blob/node/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala#L22) implementation is extremely bare bones. We should expand this to conform to normal builder apis in Scala like 

```scala

val builder = Blockchain.newBuilder

builder.+=(header1)
builder.+=(header2)
.
.
.
builder.result()
```

There is a question where validation should occur, it can happen either on `+=()` or `.result()`. What i mean by this is where show we indicate that you building a blockchain that is not valid (for instance, a bad previous block hash).

torkelrogstad:
An alternative to the collections-like builder is to  make the blockchain builder and ADT that includes both the error and success case, and then add `map`/`flatMap` to the ADT, making it possible to use it in a for expression.

If we're doing the builder way, I'm thinking it makes sense for `result` to return an ADT. Easier to handle the error in just one place, IMO

PR #429, opened 2019-04-17
Populated Tests

nkohen:
Wrote tests for POW difficulty change calculation and header processing

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -1,10 +1,14 @@
 package org.bitcoins.chain.blockchain
 
 import akka.actor.ActorSystem
-import org.bitcoins.chain.util.ChainUnitTest
-import org.bitcoins.rpc.util.RpcUtil
-import org.bitcoins.testkit.chain.BlockHeaderHelper
-import org.scalatest.FutureOutcome
+import org.bitcoins.chain.models.BlockHeaderDbHelper
+import org.bitcoins.chain.util.{ChainFixtureTag, ChainUnitTest}
+import org.bitcoins.core.protocol.blockchain.BlockHeader
+import org.bitcoins.core.util.FileUtil
+import org.bitcoins.db.NetworkDb
+import org.bitcoins.testkit.chain.{BlockHeaderHelper, ChainTestUtil}
+import org.scalatest.{Assertion, FutureOutcome}
+import play.api.libs.json.Json
 
 import scala.concurrent.Future
 
@@ -12,10 +16,14 @@ class ChainHandlerTest extends ChainUnitTest {
 
   override type FixtureParam = ChainHandler
 
+  override implicit val system = ActorSystem("ChainUnitTest")
+
+  override val defaultTag: ChainFixtureTag = ChainFixtureTag.GenisisChainHandler
+
+  override lazy val networkDb: NetworkDb = NetworkDb.MainNetDbConfig
+
   override def withFixture(test: OneArgAsyncTest): FutureOutcome =
     withChainHandler(test)
-
-  override implicit val system = ActorSystem("ChainUnitTest")
 
   behavior of "ChainHandler"
 
@@ -31,4 +39,95 @@ class ChainHandlerTest extends ChainUnitTest {
       foundHeaderF.map(found => assert(found.get == newValidHeader))
   }
 
+  it must "have an in-order seed" in { _ =>
+    val source = FileUtil.getFileAsSource("block_headers.json")
+    val arrStr = source.getLines.next
+    source.close()
+
+    import org.bitcoins.rpc.serializers.JsonReaders.BlockHeaderReads
+    val headersResult = Json.parse(arrStr).validate[Vector[BlockHeader]]
+    if (headersResult.isError) {
+      fail(headersResult.toString)
+    }
+
+    val blockHeaders = headersResult.get
+
+    blockHeaders.reduce[BlockHeader] {
+      case (prev, next) =>
+        assert(next.previousBlockHashBE == prev.hashBE)
+        next
+    }
+
+    succeed
+  }
+
+  it must "be able to process and fetch real headers from mainnet" in {
+    chainHandler: ChainHandler =>
+      val source = FileUtil.getFileAsSource("block_headers.json")
+      val arrStr = source.getLines.next
+      source.close()
+
+      import org.bitcoins.rpc.serializers.JsonReaders.BlockHeaderReads
+      val headersResult = Json.parse(arrStr).validate[Vector[BlockHeader]]
+      if (headersResult.isError) {
+        fail(headersResult.toString)
+      }
+
+      val blockHeaders =
+        headersResult.get.drop(FIRST_POW_CHANGE - FIRST_BLOCK_HEIGHT)
+
+      val firstBlockHeaderDb =
+        BlockHeaderDbHelper.fromBlockHeader(FIRST_POW_CHANGE - 2,
+                                            ChainTestUtil.blockHeader562462)
+
+      val secondBlockHeaderDb =
+        BlockHeaderDbHelper.fromBlockHeader(FIRST_POW_CHANGE - 1,
+                                            ChainTestUtil.blockHeader562463)
+
+      val thirdBlockHeaderDb =
+        BlockHeaderDbHelper.fromBlockHeader(FIRST_POW_CHANGE,
+                                            ChainTestUtil.blockHeader562464)
+
+      /*
+       * We need to insert one block before the first POW check because it is used on the next
+       * POW check. We then need to insert the next to blocks to circumvent a POW check since
+       * that would require we have an old block in the Blockchain that we don't have.
+       */
+      val firstThreeBlocks =
+        Vector(firstBlockHeaderDb, secondBlockHeaderDb, thirdBlockHeaderDb)
+
+      chainHandler.blockchain.blockHeaderDAO
+        .createAll(firstThreeBlocks)
+        .flatMap { _ =>
+          val processorF = Future.successful(
+            chainHandler.copy(blockchain =
+              chainHandler.blockchain.copy(headers = firstThreeBlocks.reverse)))
+
+          // Takes way too long to do all blocks
+          val blockHeadersToTest = blockHeaders.tail
+            .take(
+              (2 * chainHandler.chainParams.difficultyChangeInterval + 1).toInt)
+            .toList
+
+          processHeaders(processorF, blockHeadersToTest, FIRST_POW_CHANGE + 1)
+        }
+  }
+
+  final def processHeaders(
+      processorF: Future[ChainHandler],
+      remainingHeaders: List[BlockHeader],
+      height: Long): Future[Assertion] = {
+    remainingHeaders match {
+      case header :: headersTail =>
+        val newProcessorF = processorF.flatMap(_.processHeader(header))
+        val getHeaderF = newProcessorF.flatMap(_.getHeader(header.hashBE))
+        val expectedBlockHeaderDb =
+          BlockHeaderDbHelper.fromBlockHeader(height, header)
+        val assertionF =
+          getHeaderF.map(tips => assert(tips.contains(expectedBlockHeaderDb)))
+        assertionF.flatMap(_ =>
+          processHeaders(newProcessorF, headersTail, height = height + 1))
+      case Nil => succeed
+    }
+  }
 }
--- a/chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
@@ -10,12 +10,16 @@ import org.bitcoins.db.NetworkDb
 import org.bitcoins.testkit.chain.ChainTestUtil
 import org.scalatest.FutureOutcome
 
+import scala.concurrent.Future
+
 class BitcoinPowTest extends ChainUnitTest {
 
   override type FixtureParam = ChainFixture
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome =
     withChainFixture(test)
+
+  override val networkDb: NetworkDb = NetworkDb.MainNetDbConfig
 
   override implicit val system: ActorSystem = ActorSystem("BitcoinPowTest")
 
@@ -29,9 +33,8 @@ class BitcoinPowTest extends ChainUnitTest {
       val header1 = ChainTestUtil.ValidPOWChange.blockHeaderDb566494
       val header2 = ChainTestUtil.ValidPOWChange.blockHeaderDb566495
 
-      val nextWorkF = Pow.getNetworkWorkRequired(header1,
-                                                 header2.blockHeader,
-                                                 blockHeaderDAO)
+      val nextWorkF =
+        Pow.getNetworkWorkRequired(header1, header2.blockHeader, blockHeaderDAO)
 
       nextWorkF.map(nextWork => assert(nextWork == header1.nBits))
   }
@@ -51,7 +54,26 @@ class BitcoinPowTest extends ChainUnitTest {
         assert(calculatedWork == expectedNextWork))
   }
 
-  it must "calculate a GetNextWorkRequired correctly" taggedAs ChainFixtureTag.PopulatedBlockHeaderDAO inFixtured {
-    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) => succeed
+  it must "GetNextWorkRequired correctly" taggedAs ChainFixtureTag.PopulatedBlockHeaderDAO inFixtured {
+    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) =>
+      val iterations = 4200
+
+      // We must start after the first POW change to avoid looking for a block we don't have
+      val assertionFs =
+        (FIRST_POW_CHANGE + 1 until FIRST_POW_CHANGE + 1 + iterations).map {
+          height =>
+            val blockF = blockHeaderDAO.getAtHeight(height).map(_.head)
+            val nextBlockF = blockHeaderDAO.getAtHeight(height + 1).map(_.head)
+
+            for {
+              currentTip <- blockF
+              nextTip <- nextBlockF
+              nextNBits <- Pow.getNetworkWorkRequired(currentTip,
+                                                      nextTip.blockHeader,
+                                                      blockHeaderDAO)
+            } yield assert(nextNBits == nextTip.nBits)
+        }
+
+      Future.sequence(assertionFs).map(_ => succeed)
   }
 }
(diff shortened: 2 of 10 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -87,6 +87,18 @@ sealed abstract class BlockHeaderDAO
     table.filter(_.height === height).result
   }
 
+  def getBetweenHeights(from: Long, to: Long): Future[Vector[BlockHeaderDb]] = {
```
Christewart, line 129: probably should add a scaladoc saying there is no ordering guarantee here and it does not account for multiple headers at the same height

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
```diff
@@ -50,6 +50,25 @@ class BitcoinPowTest extends ChainUnitTest {
   }
 
   it must "calculate a GetNextWorkRequired correctly" taggedAs FixtureTag.PopulatedBlockHeaderDAO inFixtured {
-    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) => succeed
+    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) =>
+      val firstAfterAdjustmentF = blockHeaderDAO.getAtHeight(562464)
```
Christewart: What i had in mind here was iterating through each of the block's and verify that `getNextWorkRequired()` produces the correct output for the block that we have created in the fixture

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +33,38 @@ class ChainHandlerTest extends ChainUnitTest {
       foundHeaderF.map(found => assert(found.get == newValidHeader))
   }
 
+  it must "be able to process and fetch real headers from mainnet" taggedAs FixtureTag.PopulatedChainHandler inFixtured {
+    case ChainFixture.PopulatedChainHandler(chainHandler) =>
```
Christewart: So if I understand this correctly, this presumes a already populated `BlockHeaderDAO`, i would like to popuplate the `BlockHeaderDAO` in this test case by calling `ChainHandler.processHeader()`

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +41,96 @@ class ChainHandlerTest extends ChainUnitTest {
       foundHeaderF.map(found => assert(found.get == newValidHeader))
   }
 
+  it must "have an in-order seed" in { _ =>
```
Christewart, line 42: so is this just testing that the json file parses ok?

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +41,96 @@ class ChainHandlerTest extends ChainUnitTest {
...
+        assert(next.previousBlockHashBE == prev.hashBE)
+        next
+    }
+
+    succeed
+  }
+
+  it must "be able to process and fetch real headers from mainnet" in {
+    chainHandler: ChainHandler =>
+      val source =
```
Christewart: maybe factored out the parsing to a util method? Some what expensive and seems to be useful else where.

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -249,10 +273,12 @@ trait ChainUnitTest
     makeFixture(createBlockHeaderDAO, destroyHeaderTable)(test)
   }
 
+  val FIRST_BLOCK_HEIGHT: Int = 562375
```
Christewart, line 184: Add a scaladoc to what this corresponds to, seems like a random number otherwise

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +41,96 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      val secondBlockHeaderDb =
+        BlockHeaderDbHelper.fromBlockHeader(firstPowChange - 1,
+                                            ChainTestUtil.blockHeader562463)
+
+      val thirdBlockHeaderDb =
+        BlockHeaderDbHelper.fromBlockHeader(firstPowChange,
+                                            ChainTestUtil.blockHeader562464)
+
+      chainHandler.blockchain.blockHeaderDAO
+        .createAll(Vector(firstBlockHeaderDb, secondBlockHeaderDb, thirdBlockHeaderDb))
```
Christewart: intermediate `val` please

Christewart: why do we insert the first 3, can't we just do the first one?

Christewart: This is also an illustration why #428 needs to happen. Working with the blockchain api is pretty difficult right now

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +41,96 @@ class ChainHandlerTest extends ChainUnitTest {
...
+        .flatMap { _ =>
+          var processorF = Future.successful(
+            chainHandler.copy(blockchain = chainHandler.blockchain.copy(
+              headers = Vector(thirdBlockHeaderDb, secondBlockHeaderDb, firstBlockHeaderDb))))
+          var blockCount = firstPowChange
+
+          // Takes way too long to do all blocks
+          val blockHeadersToTest = blockHeaders.tail.take((2*chainHandler.chainParams.difficultyChangeInterval + 1).toInt)
```
Christewart: will this get better with the async version?

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
```diff
@@ -49,7 +48,21 @@ class BitcoinPowTest extends ChainUnitTest {
         assert(calculatedWork == expectedNextWork))
   }
 
-  it must "calculate a GetNextWorkRequired correctly" taggedAs FixtureTag.PopulatedBlockHeaderDAO inFixtured {
-    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) => succeed
+  it must "GetNextWorkRequired correctly" taggedAs FixtureTag.PopulatedBlockHeaderDAO inFixtured {
+    case ChainFixture.PopulatedBlockHeaderDAO(blockHeaderDAO) =>
+      (FIRST_BLOCK_HEIGHT until FIRST_BLOCK_HEIGHT + 3000).foreach { height =>
```
Christewart: is `3,000` a magic number? Any reason we aren't doing at least 2 POW changes?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -296,6 +322,11 @@ trait ChainUnitTest
                                                 dbHeaders = dbHeaders,
                                                 batchesSoFar = Vector.empty)
 
+        val chainHandler = makeChainHandler(
```
Christewart: named params, use intermediate val for `BlockHeaderDbHelper.fromBlockHeader...`

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -314,27 +345,45 @@ trait ChainUnitTest
...
-    val genesisHeaderF = setupHeaderTableWithGenesisHeader()
+    val (chainHandler, genesisHeaderF) = setupHeaderTableWithGenesisHeader()
     genesisHeaderF.map(_ => chainHandler)
   }
 
   def withChainHandler(test: OneArgAsyncTest): FutureOutcome = {
     makeFixture(createChainHandler, destroyHeaderTable)(test)
   }
 
+  // Creates and populates BlockHeaderTable with block headers 562375 to 571375
```
Christewart: scaladoc, not a scala comment

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -314,27 +345,45 @@ trait ChainUnitTest
...
   /** Creates the [[org.bitcoins.chain.models.BlockHeaderTable]] */
   private def setupHeaderTable(): Future[Unit] = {
     ChainDbManagement.createHeaderTable(dbConfig = dbConfig,
                                         createIfNotExists = true)
   }
 
   /** Creates the [[org.bitcoins.chain.models.BlockHeaderTable]] and inserts the genesis header */
-  private def setupHeaderTableWithGenesisHeader(): Future[BlockHeaderDb] = {
+  private def setupHeaderTableWithGenesisHeader(): (
+      ChainHandler,
```
Christewart, line 287: do you see this tuple expanding? If so we should just bite the bullet now and make a case class

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -57,21 +58,40 @@ sealed abstract class BlockHeaderDAO
...
-        if (header.height == height) {
-          Future.successful(Some(header))
-        } else if (height <= header.height) {
-          //need to keep traversing backwards to get to the right height
-          getAncestorAtHeight(header, height)
+    val headersF = getBetweenHeights(from = height, to = child.height - 1)
+
+    val headersByHeight: Array[Vector[BlockHeaderDb]] =
+      new Array[Vector[BlockHeaderDb]](_length = (child.height - height).toInt)
+    headersByHeight.indices.foreach(headersByHeight(_) = Vector.empty)
```
Christewart: this seems like magic ?

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
Christewart: Is the `to` inclusive or exclusive? It should be stated in the scaladoc. We should probably follow the convention of scala collections. 

I.e. 
```scala
1.until(3)
```

```scala
1.to(3)

```

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -57,21 +58,40 @@ sealed abstract class BlockHeaderDAO
...
+      @tailrec
+      def loop(
+          currentHeader: BlockHeaderDb,
+          headersByDescHeight: Vector[Vector[BlockHeaderDb]]): Option[
+        BlockHeaderDb] = {
+        if (currentHeader.height == height) {
+          Some(currentHeader)
         } else {
-          Future.successful(None)
+          val prevHeaderOpt = headersByDescHeight.headOption.flatMap(
```
Christewart, line 101: if you are worried about performance, a List might be faster here http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#lists-vs-vectors

specifically: 

```


Constructing a List item-by-item is 5-15x (!) faster than constructing a Vector item-by-item.

De-constructing a List item-by-item using .tail is 50-60x (!) faster than de-constructing a Vector item-by-item using .tail

Looking things up by index in a Vector works; looking things up by index in a List is O(n), and expensive for non-trivial lists.
```

Since we aren't looking things up by index here, probably should be using `List`. I guess this will incur more memory, but I think that is ok in this context

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
nkohen: Yes, and that the data in the json file is correct

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
nkohen: What async version?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
nkohen: I don't but if any field ever gets added I'll change it to a case class

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
```diff
@@ -50,17 +50,19 @@ class BitcoinPowTest extends ChainUnitTest {
...
-                                                  nextTip.blockHeader,
-                                                  blockHeaderDAO)
-        } assert(nextNBits == nextTip.nBits)
+          for {
+            currentTip <- blockF
+            nextTip <- nextBlockF
+            nextNBits <- Pow.getNetworkWorkRequired(currentTip,
+                                                    nextTip.blockHeader,
+                                                    blockHeaderDAO)
+          } assert(nextNBits == nextTip.nBits)
```
Christewart: don't we need to `yield assert(nextNBits == nextTip.nBits)`

Christewart:
So I think we can implement this unit test on this PR too? https://github.com/bitcoin-s/bitcoin-s-core/blob/node/chain-test/src/test/scala/org/bitcoins/chain/validation/TipValidationTest.scala#L38

That test is currently ignored, i think we can either delete the entire test case if it is redundant or implement it with the populated block header dao

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +42,105 @@ class ChainHandlerTest extends ChainUnitTest {
...
+       * POW check. We then need to insert the next to blocks to circumvent a POW check since
+       * that would require we have an old block in the Blockchain that we don't have.
+       */
+      val firstThreeBlocks =
+        Vector(firstBlockHeaderDb, secondBlockHeaderDb, thirdBlockHeaderDb)
+
+      chainHandler.blockchain.blockHeaderDAO
+        .createAll(firstThreeBlocks)
+        .flatMap { _ =>
+          var processorF = Future.successful(
```
torkelrogstad: It's a bit hard to understand what's going on here with the `var`. Is this equivalent to making this 
```scala
val processor = chainHandler.copy(
            blockchain =
              chainHandler.blockchain.copy(headers = firstThreeBlocks.reverse))
```

And then change line 118-122 to 
```scala
val newProcessorF = processor.processHeader(blockHeader)

newProcessorF.flatMap { processor =>
```
?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -296,6 +323,12 @@ trait ChainUnitTest
                                                 dbHeaders = dbHeaders,
                                                 batchesSoFar = Vector.empty)
 
+        val firstHeader =
+          BlockHeaderDbHelper.fromBlockHeader(562375,
```
torkelrogstad: use the val from above?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -314,27 +347,45 @@ trait ChainUnitTest
...
-    val genesisHeaderF = setupHeaderTableWithGenesisHeader()
+    val (chainHandler, genesisHeaderF) = setupHeaderTableWithGenesisHeader()
     genesisHeaderF.map(_ => chainHandler)
   }
 
   def withChainHandler(test: OneArgAsyncTest): FutureOutcome = {
     makeFixture(createChainHandler, destroyHeaderTable)(test)
   }
 
+  // Creates and populates BlockHeaderTable with block headers 562375 to 571375
```
torkelrogstad: Make this a Scaladoc?

chain-test/src/test/scala/org/bitcoins/chain/util/ChainUnitTest.scala
```diff
@@ -314,27 +347,45 @@ trait ChainUnitTest
...
+  def createPopulatedChainHandler(): Future[ChainHandler] = {
```
torkelrogstad, line 268: IMO this is cleaner, but not a big deal at all:

```scala
  def createPopulatedChainHandler(): Future[ChainHandler] =
    for {
      blockHeaderDAO <- createPopulatedBlockHeaderDAO()
      blockHeaderVec <- blockHeaderDAO.getAtHeight(FIRST_BLOCK_HEIGHT)
    } yield ChainHandler(Blockchain(blockHeaderVec, blockHeaderDAO))
```

core/src/main/scala/org/bitcoins/core/util/FileUtil.scala
```diff
@@ -0,0 +1,8 @@
+package org.bitcoins.core.util
+
+object FileUtil {
+
+  def getFileAsSource(fileName: String): scala.io.BufferedSource = {
```
torkelrogstad, line 6: Scaladoc? Might be nice to explain what sort of files you can get with this (I assume it's files on the classpath?)

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -31,4 +42,105 @@ class ChainHandlerTest extends ChainUnitTest {
...
+            chainHandler.copy(blockchain =
+              chainHandler.blockchain.copy(headers = firstThreeBlocks.reverse)))
```
Christewart, line 104: Maybe something like this instead as a helper method? 

```scala
  final def processHeaders(lastHeaderF: Future[ChainHandler], 
                           remainingHeaders: List[BlockHeader], 
                           height: Long = FIRST_BLOCK_HEIGHT + 3): Future[ChainHandler] = {
    remainingHeaders match {
      case h :: t =>
        val newLastHeaderF = lastHeaderF.flatMap(_.processHeader(h))
        val getHeaderF = newLastHeaderF.flatMap(_.getHeader(h.hashBE))
        val expectedBlockHeaderDb = BlockHeaderDbHelper.fromBlockHeader(height, h)
        val assertionF = getHeaderF.map(tips => assert(tips.contains(expectedBlockHeaderDb)))
        assertionF.flatMap(_ => processHeaders(newLastHeaderF, t, height = height + 1))
      case Nil => lastHeaderF
    }
  }
```

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
torkelrogstad: `assert` throws if it encounters a false value, so that's not necessary

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
nkohen: So we loop down below and need the new chain handler each time, so I initialize this var with the starting processor, and update it with the new ones. I could accomplish this with vals if I made a tail recursive def in the middle here, but I thought this was less clunky

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
nkohen: I did end up having to do this because although they threw, they did not come back to the test which still passed (albeit with a ton of error messages and stack traces)

PR #430, opened 2019-04-18
WIP: 2019 04 18 wallet ammonite scripts

Christewart:
The goal of this PR is to create a ammonite script that

1. Creates a bitcoind instance
2. Connects our `chain` project to that bitcoind instance via zmq
3. Notifies our wallet when money is sent to an address that the wallet generated
4. Spend from our bitcoin-s wallet back to the bitcoind instance.

This PR is a wip and there are two major problems that are being encountered right now. 

1. The `ChainHandler.processBlockHeader()` implementation is currently not async safe. This means that we can have blocks generated by bitcoind and then stored in BlockHeaderDAO out of order, which violates database constraints saying that a previous block hash must exist in the database before we can insert a new block header.

2. I am experiencing some weird jeromq behavior. It doesn't seem we are reliably starting and stopping zmq. In the case of not starting zmq correctly, we are missing messages from bitcoind. When I try to stop zmq in the clean up process of this script, i hang infinitely when I call `ZmqSubscriber.stop()`. I've filed an issue on jeromq's github asking about this https://github.com/zeromq/jeromq/issues/721

This script can be run with 

```bashrc
$ sbt project/doc 
sbt:bitcoin-s-doc> run doc/src/main/scala/org/bitcoins/doc/wallet/create-wallet.sc
```


project/Deps.scala
```diff
@@ -202,7 +202,8 @@ object Deps {
   )
 
   val doc = List(
-    Test.ammonite,
+    "com.lihaoyi" %% "ammonite" % V.ammoniteV cross CrossVersion.full,
```
torkelrogstad, line 206: Suggestion for cleaning this up:

Add to `Compile`: 
```scala
    val ammonite = "com.lihaoyi" %% "ammonite" % V.ammoniteV cross CrossVersion.full
```

Add to `Test`: 
```scala
    val logback = Compile.logback % "test"
    val ammonite = Compile.ammonite % "test"
```

And then change `doc` to:
```scala
  val doc = List(
    Compile.ammonite,
    Compile.logback,
    Test.scalaTest
  )
```

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -117,7 +117,7 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
     * Assumes the `config` object has a `datadir` string. Returns the written
     * file.
     */
-  def writeConfigToFile(config: Config): File = {
+  def writeConfigToFile(config: Config, datadir: File): File = {
```
torkelrogstad, line 120: Scaladoc above should be changed

torkelrogstad:
I'm getting these two lines ad infinitum when running `create-wallet.sc`:
```
zmq.poll.Poller@37dc1654 rebuilding selector
zmq.poll.Poller@7030d57c rebuilding selector
```

I cannot find where these lines are coming from, but perhaps you know that @Christewart ?

PR #431, opened 2019-04-18
Basic Coin Selection

### LINKED ISSUE #406: Implement basic coin selection
Related reading/projects: 
 - Murch's [master thesis](http://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf) on coin selection (used by Core, overkill for first round in Bitcoin-S)
- Murch's [Scala (:tada:) implementation](https://github.com/Xekyo/CoinSelectionSimulator) of his algo
- bitcoinjs [`coinselect`](https://github.com/bitcoinjs/coinselect)
- Kalle Alms [short talk on coin selection](http://diyhpl.us/wiki/transcripts/scalingbitcoin/tokyo-2018/edgedevplusplus/coin-selection/)
- Talk by Murch on coin selection: http://diyhpl.us/wiki/transcripts/scalingbitcoin/milan/coin-selection/
- BitcoinJ [KeyTimeCoinSelector](https://github.com/bitcoinj/bitcoinj/blob/fffa4c06fe73c1fd1f1ef8688415ccd9f73ac060/core/src/main/java/org/bitcoinj/wallet/KeyTimeCoinSelector.java#L38)
- BitcoinJ [FilteringCoinSelector](https://github.com/bitcoinj/bitcoinj/blob/3177bd52a2bfa491c5902e95b8840030e1a31159/core/src/main/java/org/bitcoinj/wallet/FilteringCoinSelector.java#L28)

I'm thinking for an initial implementation a greedy algorithm with a slight bias to pick small UTXOs if the feerate is low would be sufficient. 

Other things we should take into consideration: 

 - Privacy gotchas: See for example [this TX](https://blockstream.info/tx/4d7da86b695b4f9a0864dfff3cceffc662e5bd33b31ecc1c9911352487e6d725) on Blocksteram.info. 
- other privacy aspects of coin selection: See [privacy article on Bitcoin Wiki](https://en.bitcoin.it/wiki/Privacy) and CTRL-F "selection"

nkohen:
Implements greedy coin selection (taking smallest or largest UTXOs first).

Replaces current "fail if there is no big enough UTXO" with greedy large UTXOs first coin selection in `Wallet.sendToAddress`.

Closes #406

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
@@ -1,0 +1,94 @@
+package org.bitcoins.wallet.api
+
+import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
+import org.bitcoins.core.number.Int64
+import org.bitcoins.core.protocol.script.ScriptPubKey
+import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerByte}
+import org.bitcoins.testkit.core.gen.{TransactionGenerators, WitnessGenerators}
+import org.bitcoins.wallet.models.{
+  NativeV0UTXOSpendingInfoDb,
+  UTXOSpendingInfoDb
+}
+import org.bitcoins.wallet.util.{BitcoinSWalletTest, WalletTestUtil}
+import org.scalatest.FutureOutcome
+
+class CoinSelectorTest extends BitcoinSWalletTest {
+  case class CoinSelectionFixture(
+      output: TransactionOutput,
+      feeRate: FeeUnit,
+      utxo1: UTXOSpendingInfoDb,
+      utxo2: UTXOSpendingInfoDb,
+      utxo3: UTXOSpendingInfoDb) {
+    val utxoSet: Vector[UTXOSpendingInfoDb] = Vector(utxo1, utxo2, utxo3)
+  }
+
+  override type FixtureParam = CoinSelectionFixture
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val output = TransactionOutput(Satoshis(Int64(99L)), ScriptPubKey.empty)
+    val feeRate = SatoshisPerByte(CurrencyUnits.zero)
+
+    val utxo1 = NativeV0UTXOSpendingInfoDb(
+      id = Some(1),
+      outPoint = TransactionGenerators.outPoint.sample.get,
+      output = TransactionOutput(Satoshis(Int64(10)), ScriptPubKey.empty),
+      privKeyPath = WalletTestUtil.sampleSegwitPath,
+      scriptWitness = WitnessGenerators.scriptWitness.sample.get
+    )
+    val utxo2 = NativeV0UTXOSpendingInfoDb(
+      id = Some(2),
+      outPoint = TransactionGenerators.outPoint.sample.get,
+      output = TransactionOutput(Satoshis(Int64(90)), ScriptPubKey.empty),
+      privKeyPath = WalletTestUtil.sampleSegwitPath,
+      scriptWitness = WitnessGenerators.scriptWitness.sample.get
+    )
+    val utxo3 = NativeV0UTXOSpendingInfoDb(
+      id = Some(3),
+      outPoint = TransactionGenerators.outPoint.sample.get,
+      output = TransactionOutput(Satoshis(Int64(20)), ScriptPubKey.empty),
+      privKeyPath = WalletTestUtil.sampleSegwitPath,
+      scriptWitness = WitnessGenerators.scriptWitness.sample.get
+    )
+
+    test(CoinSelectionFixture(output, feeRate, utxo1, utxo2, utxo3))
+  }
+
+  behavior of "CoinSelector"
+
+  it must "accumulate largest outputs" in { fixture =>
+    val selection =
+      CoinSelector.accumulateLargest(walletUtxos = fixture.utxoSet,
+                                     outputs = Vector(fixture.output),
+                                     feeRate = fixture.feeRate)
+
+    assert(selection == Vector(fixture.utxo2, fixture.utxo3))
+  }
+
+  it must "accumulate smallest outputs" in { fixture =>
+    val selection =
+      CoinSelector.accumulateSmallestViable(walletUtxos = fixture.utxoSet,
+                                            outputs = Vector(fixture.output),
+                                            feeRate = fixture.feeRate)
+
+    assert(selection == Vector(fixture.utxo1, fixture.utxo3, fixture.utxo2))
+  }
+
+  it must "accumulate outputs in order" in { fixture =>
+    val selection = CoinSelector.accumulate(walletUtxos = fixture.utxoSet,
+                                            outputs = Vector(fixture.output),
+                                            feeRate = fixture.feeRate)
+
+    assert(selection == Vector(fixture.utxo1, fixture.utxo2))
+  }
+
+  it must "correctly approximate transaction input size" in { fixture =>
+    val expected1 = 32 + 4 + 1 + 4 + fixture.utxo1.scriptWitnessOpt.get.bytes.length
+    val expected2 = 32 + 4 + 1 + 4 + fixture.utxo2.scriptWitnessOpt.get.bytes.length
+    val expected3 = 32 + 4 + 1 + 4 + fixture.utxo3.scriptWitnessOpt.get.bytes.length
+
+    assert(CoinSelector.approximateUtxoSize(fixture.utxo1) == expected1)
+    assert(CoinSelector.approximateUtxoSize(fixture.utxo2) == expected2)
+    assert(CoinSelector.approximateUtxoSize(fixture.utxo3) == expected3)
+  }
+}
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
@@ -1,6 +1,6 @@
 package org.bitcoins.wallet.api
 
-import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
+import org.bitcoins.core.currency._
 import org.bitcoins.core.number.Int64
 import org.bitcoins.core.protocol.script.ScriptPubKey
 import org.bitcoins.core.protocol.transaction.TransactionOutput
@@ -10,7 +10,7 @@ import org.bitcoins.wallet.models.{
   NativeV0UTXOSpendingInfoDb,
   UTXOSpendingInfoDb
 }
-import org.bitcoins.wallet.util.{BitcoinSWalletTest, WalletTestUtil}
+import org.bitcoins.testkit.wallet.{BitcoinSWalletTest, WalletTestUtil}
 import org.scalatest.FutureOutcome
 
 class CoinSelectorTest extends BitcoinSWalletTest {
@@ -26,27 +26,27 @@ class CoinSelectorTest extends BitcoinSWalletTest {
   override type FixtureParam = CoinSelectionFixture
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    val output = TransactionOutput(Satoshis(Int64(99L)), ScriptPubKey.empty)
+    val output = TransactionOutput(99.sats, ScriptPubKey.empty)
     val feeRate = SatoshisPerByte(CurrencyUnits.zero)
 
     val utxo1 = NativeV0UTXOSpendingInfoDb(
       id = Some(1),
       outPoint = TransactionGenerators.outPoint.sample.get,
-      output = TransactionOutput(Satoshis(Int64(10)), ScriptPubKey.empty),
+      output = TransactionOutput(10.sats, ScriptPubKey.empty),
       privKeyPath = WalletTestUtil.sampleSegwitPath,
       scriptWitness = WitnessGenerators.scriptWitness.sample.get
     )
     val utxo2 = NativeV0UTXOSpendingInfoDb(
       id = Some(2),
       outPoint = TransactionGenerators.outPoint.sample.get,
-      output = TransactionOutput(Satoshis(Int64(90)), ScriptPubKey.empty),
+      output = TransactionOutput(90.sats, ScriptPubKey.empty),
       privKeyPath = WalletTestUtil.sampleSegwitPath,
       scriptWitness = WitnessGenerators.scriptWitness.sample.get
     )
     val utxo3 = NativeV0UTXOSpendingInfoDb(
       id = Some(3),
       outPoint = TransactionGenerators.outPoint.sample.get,
-      output = TransactionOutput(Satoshis(Int64(20)), ScriptPubKey.empty),
+      output = TransactionOutput(20.sats, ScriptPubKey.empty),
       privKeyPath = WalletTestUtil.sampleSegwitPath,
       scriptWitness = WitnessGenerators.scriptWitness.sample.get
     )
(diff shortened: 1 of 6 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerByte}
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, TransactionGenerators}
+import org.bitcoins.wallet.models.UTXOSpendingInfoDb
+import org.bitcoins.wallet.util.BitcoinSWalletTest
+import org.scalatest.FutureOutcome
+
+class CoinSelectorTest extends BitcoinSWalletTest {
+  override type FixtureParam = (
+      TransactionOutput,
```
Christewart: I would prefer us to make a case class for this

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.bitcoins.wallet.models.UTXOSpendingInfoDb
+
+import scala.annotation.tailrec
+
+/** Implements algorithms for selecting from a UTXO set to spend to an output set at a given fee rate. */
+trait CoinSelector {
+
+  /** Greedily selects from walletUtxos starting with the largest outputs, skipping detrimental outputs */
+  def accumulateLargest(
+      walletUtxos: Seq[UTXOSpendingInfoDb],
```
Christewart: Don't use `Seq`, use `Vector`

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.wallet.fee.FeeUnit
+import org.bitcoins.wallet.models.UTXOSpendingInfoDb
+
+import scala.annotation.tailrec
+
+/** Implements algorithms for selecting from a UTXO set to spend to an output set at a given fee rate. */
+trait CoinSelector {
+
+  /** Greedily selects from walletUtxos starting with the largest outputs, skipping detrimental outputs */
```
Christewart: I'm not sure what "deterimental" means in this context? Also it might be nice to add in the comment that this will produce a transaction the lowest transaction fee now in the short term (assuming none of these utxos have EXTREMELY large scriptSigs) with the downside of producing lots of utxos long term which will need to be consolidated. Maybe recommend that this should be used in high fee environments?

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+      walletUtxos: Seq[UTXOSpendingInfoDb],
+      outputs: Seq[TransactionOutput],
+      feeRate: FeeUnit): Vector[UTXOSpendingInfoDb] = {
+    val sortedUtxos =
+      walletUtxos.toVector.sortBy(_.value.satoshis.toLong).reverse
+
+    accumulate(sortedUtxos, outputs, feeRate)
+  }
+
+  /** Greedily selects from walletUtxos starting with the smallest outputs, skipping detrimental outputs */
```
Christewart: So I think the inverse of the comment I have above applies here. This scheme might be best used in low fee environments, and another tradeoff with this coin selection is that you are reducing your privacy by linking `n` utxos into a single transaction. Not that that is necessarily a bad thing -- it has to be done eventually -- but it is good to note the trade off.

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+    val totalValue = outputs.foldLeft(CurrencyUnits.zero) {
+      case (totVal, output) => totVal + output.value
+    }
+
+    @tailrec
+    def addUtxos(
+        alreadyAdded: Vector[UTXOSpendingInfoDb],
+        valueSoFar: CurrencyUnit,
+        utxosLeft: Vector[UTXOSpendingInfoDb]): Vector[UTXOSpendingInfoDb] = {
+      val fee = CurrencyUnits.zero // TODO: Calculate this
```
Christewart: This is tricky, and I was going to perhaps we should assume that the user calculates this _before_ calling anything in the coin selection algorithm. However this would create a recursive dependency between the [`TxBuilder.creditingOutputs`](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala#L74) and the coin selection algorithm. 

I think the right path is 
1. Call coin selection with a `target: CurrencyUnit` 
2. Pass in a desired `FeeUnit` 
3. Call `TxBuilder` with utxo combinations and use [`TxBuilder.unsignedTx`](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala#L103-L110) to get the _actual fee_ (not just the sat/byte!) **
4. If we have enough money with this utxo combination, return it. Else iterate again.

** there is a complication here, there isn't an _easy_ way to get the actual fee from the `unsignedTx` returned by `TxBuilder.unsignedTx`. We should consider modifying `TxBuilder` to provide this information to us some how.  I guess if we also want to construct a `TxBuilder`, we need to pass in all relevant information needed to the construct the TxBuilder, which seems unfortunate.

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import org.bitcoins.core.currency.{CurrencyUnit, CurrencyUnits}
+import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.wallet.fee.FeeUnit
+import org.bitcoins.wallet.models.UTXOSpendingInfoDb
+
+import scala.annotation.tailrec
+
+/** Implements algorithms for selecting from a UTXO set to spend to an output set at a given fee rate. */
+trait CoinSelector {
```
Christewart, line 11: Can we come with a interface that works for _any_ coin selection scheme? So we can plug and play later on?

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,69 @@
...
+        alreadyAdded: Vector[UTXOSpendingInfoDb],
+        valueSoFar: CurrencyUnit,
+        utxosLeft: Vector[UTXOSpendingInfoDb]): Vector[UTXOSpendingInfoDb] = {
+      val fee = CurrencyUnits.zero // TODO: Calculate this
+      if (valueSoFar > totalValue + fee) {
+        alreadyAdded
+      } else {
+        val nextUtxo = utxosLeft.head
+        val nextUtxoFee = CurrencyUnits.zero // TODO: Calculate this
+        if (nextUtxo.value < nextUtxoFee) {
```
Christewart, line 67: Does this mean that the fee to spend this utxo is greater than the utxo value itself? Also we should probably be checking `Policy.dustThreshold`

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
nkohen: I mean we could have an abstract `def select(walletUtxos: Vector[UTXOSpendingInfoDb], outputs: Vector[TransactionOutput], feeRate: FeeUnit): Vector[UTXOSpendingInfoDb]` as the only member of `CoinSelector` and then implement a bunch of different options as objects but I don't see any reason not just put them in one place so that its nice in an IDE to just type `CoinSelector.` and see what comes up, there's no real benefit to making an interface for this one method. I don't expect people to be passing in lambdas where coin selection algorithms are required (in fact that would be shady in my eyes)

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
nkohen: Yup, and can you elaborate on the `Policy.dustThreshold` part?

wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  override type FixtureParam = CoinSelectionFixture
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val output = TransactionOutput(Satoshis(Int64(99L)), ScriptPubKey.empty)
+    val feeRate = SatoshisPerByte(CurrencyUnits.zero)
+
+    val utxo1 = UTXOSpendingInfoDb(
+      Some(1),
```
torkelrogstad: Named params

wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -0,0 +1,83 @@
...
+      TransactionGenerators.outPoint.sample.get,
+      TransactionOutput(Satoshis(Int64(10)), ScriptPubKey.empty),
+      CryptoGenerators.bip44Path.sample.get,
+      None,
+      None
+    )
+    val utxo2 = UTXOSpendingInfoDb(
```
torkelrogstad: Named params

wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -0,0 +1,83 @@
...
+      Some(2),
+      TransactionGenerators.outPoint.sample.get,
+      TransactionOutput(Satoshis(Int64(90)), ScriptPubKey.empty),
+      CryptoGenerators.bip44Path.sample.get,
+      None,
+      None
+    )
+    val utxo3 = UTXOSpendingInfoDb(
```
torkelrogstad: Named params

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
torkelrogstad: https://github.com/bitcoin-s/bitcoin-s-core/blob/2dc798d8569c08241198416b238718cbf64c4128/core/src/main/scala/org/bitcoins/core/policy/Policy.scala

This file contains information that is _not_ part of the consensus rules, but still affect the validity of a transaction/UTXO in Bitcoin Core. For example `standardFlags`: if a node receives a transaction (just a transaction, not in a block) that doesn't satisfy these requirements it will mark it as "nonstandard" and refuse to propagate them further. It would however be accepted if received in a block. More reading in the [Bitcoin developer guide](https://bitcoin.org/en/developer-guide#standard-transactions). 

The dust threshold refers to how small a UTXO can be before it's uneconomical to spend it. I.e. if the fee required to spend it is greater than the value of the UTXO its best to just ignore it.

Right now those values are hard coded (and I'm not sure if they're up to date), but IMO they should be configurable somehow. Both default fee, max fee and dust threshold. That's out of scope for this PR though

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
torkelrogstad: That would mean the change output needs to be provided in this method, then? 

I agree on the last point. Would also make sense for the tx builder to inform which output is the change output, IMO. Changing the tx builder should wait until a certain closed source project is merged into upstream, however...

wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -0,0 +1,81 @@
...
+      outPoint = TransactionGenerators.outPoint.sample.get,
+      output = TransactionOutput(Satoshis(Int64(10)), ScriptPubKey.empty),
+      privKeyPath = ???,
+      scriptWitness = ???
+    )
+    val utxo2 = SegWitUTOXSpendingInfodb(
+      id = Some(2),
+      outPoint = TransactionGenerators.outPoint.sample.get,
+      output = TransactionOutput(Satoshis(Int64(90)), ScriptPubKey.empty),
+      privKeyPath = ???,
```
torkelrogstad: you can use `WalletTestUtil.sampleSegwitPath` here and below + above

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,102 @@
...
+      case (totVal, output) => totVal + output.value
+    }
+
+    @tailrec
+    def addUtxos(
+        alreadyAdded: Vector[UTXOSpendingInfoDb],
+        valueSoFar: CurrencyUnit,
+        bytesSoFar: Long,
+        utxosLeft: Vector[UTXOSpendingInfoDb]): Vector[UTXOSpendingInfoDb] = {
+      val fee = Satoshis(Int64(feeRate.toLong * bytesSoFar))
```
torkelrogstad: `val fee = feeRate.currencyUnit * bytesSoFar`

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,102 @@
...
+      if (valueSoFar > totalValue + fee) {
+        alreadyAdded
+      } else if (utxosLeft.isEmpty) {
+        throw new RuntimeException(
+          s"Not enough value in wallet ($valueSoFar) to make transaction of size $totalValue")
```
torkelrogstad: Nitty comment: would an error message not referencing the word `wallet` be better? Any UTXOs can be passed in, not just ones from our wallet

Not so nitty: What's your guys' opinion on throwing here? I'm thinking it might be better to return a`Try`/`Option` or an ADT

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,102 @@
...
+      } else {
+        val nextUtxo = utxosLeft.head
+        val approxUtxoSize = CoinSelector.approximateUtxoSize(nextUtxo)
+        val nextUtxoFee = Satoshis(Int64(feeRate.toLong * approxUtxoSize))
```
torkelrogstad: `val nextUtxoFee = feeRate.currencyUnit * approxUtxoSize`

wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    addUtxos(Vector.empty, CurrencyUnits.zero, bytesSoFar = 0L, walletUtxos)
+  }
+}
+
+object CoinSelector extends CoinSelector {
+
+  /** Cribbed from [[https://github.com/bitcoinjs/coinselect/blob/master/utils.js]] */
+  def approximateUtxoSize(utxo: UTXOSpendingInfoDb): Long = {
```
torkelrogstad, line 88: Would be nice to have some hard coded unit tests for this method

torkelrogstad (commented):
Could coin selection functionality be moved into `core`? Would have to use `UtxoSpendingInfo` instead of `UtxoSpendingInfoDb`, but otherwise it should work the same

Christewart:
Looks like we have a compiler error https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/538480009#L1424

torkelrogstad:
There's `MatchError`s on CI, looks related to this PR

PR #434, opened 2019-04-23
2019 04 23 app config per project db config per project

Christewart:
Closes #425 
This is is a super set of #432 and #425 

This creates `AppConfigs` and `DbConfigs` for each of our sub projects

- In the `chain` project we now have `ChainAppConfig` and `ChainDbConfig`
- In the `wallet` project we now have `WalletAppConfig` and `WalletDbConfig`
- In the `node` project we now have `NodeAppConfig` and `NodeDbConfig` 

The idea is to start consolidating configuration stuff for these various projects into one place so we don't have an ever expanding parameter list every where.


chain/src/main/scala/org/bitcoins/chain/db/ChainDbConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+
+object ChainMainNetDbConfig extends ChainDbConfig {
+  override val networkDb = NetworkDb.MainNetDbConfig
+}
+
+object ChainTestNet3DbConfig extends ChainDbConfig {
+  override val networkDb = NetworkDb.TestNet3DbConfig
+}
+
+object ChainRegTestDbConfig extends ChainDbConfig  {
```
nkohen: Any reason these aren't `case object`s?

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -3,6 +3,11 @@ package org.bitcoins.db
...
-  val network: NetworkParameters = chain.network
-}
+
+trait AppConfig[C <: DbConfig] {
+  def dbConfig: C
+
+  def chain: ChainParams = dbConfig.networkDb.chain
+
+  def network: NetworkParameters = chain.network
+}
```
nkohen, line 13: I think it would be super nice to have an `EnvParams` and then construct all `AppConfig`s by taking the `EnvParams` and a `ChainParams` rather than forcing the user to construct a `dbConfig`

db-commons/src/main/scala/org/bitcoins/db/CRUD.scala
```diff
@@ -23,7 +23,7 @@ abstract class CRUD[T, PrimaryKeyType] extends BitcoinSLogger {
   def dbConfig: DbConfig
 
   /** Binding to the actual database itself, this is what is used to run querys */
-  def database: SafeDatabase = SafeDatabase(dbConfig)
+  def database: SafeDatabase[_] = SafeDatabase(dbConfig)
```
nkohen: Can you use `dbConfig.type` here? If not then I still think `[DbConfig]` is nicer than `[_]`

chain/src/main/scala/org/bitcoins/chain/db/ChainDbConfig.scala
nkohen: Same for other DbConfig files

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
Christewart: I think this is out of scope of this PR, we can do this later via a config file.

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -33,15 +23,13 @@ import scala.concurrent.Future
...
+  def dbConfig: WalletDbConfig = walletAppConfig.dbConfig
 
-  def networkParameters: BitcoinNetwork = chainParams match {
-    case MainNetChainParams    => MainNet
-    case TestNetChainParams    => TestNet3
-    case RegTestNetChainParams => RegTest
-  }
+  def chainParams: ChainParams = walletAppConfig.chain
+
+  def networkParameters: NetworkParameters = walletAppConfig.network
```
torkelrogstad, line 32:
```suggestion
def networkParameters: BitcoinNetwork = walletAppConfig.network
```

That way we avoid the cast in Wallet.scala

Christewart: I actually want to try and keep this generic for when we support multiple cryptocurrency networks, it is important to keep the `Api` level stuff agnostic to the cryptocurrency network

Christewart: I took the easy way out by casting for now, I should really go fix the cast and pattern match

PR #435, opened 2019-04-23
[fix] Transaction can reference two different outputs of the same previous transaction

piu130:
Currently, the code does not let you create a transaction that references two output indexes of a previous transaction.

This should be allowed, as long as the previous txId AND index is not the same.

Example:

PrevTx
```ts
outputs: [
  "utxo0",
  "utxo1"
]
```

CurrTx
```ts
inputs: [
  "Outpoint(PrevTx.id, 0)", // utxo0
  "Outpoint(PrevTx.id, 1)"  // utxo1
]


piu130:
Looks like there is a `request timed out` error from somewhere else. Could you rerun the tests?

Christewart:
Thank you! Nice catch.

PR #438, opened 2019-04-25
Fix CI tests hanging

torkelrogstad:
SQLite has issues with concurrent connections. 

This line: 
https://github.com/bitcoin-s/bitcoin-s-core/blob/50565ee4b1a0203f3d319a188041b80d9f001710/build.sbt#L344
Forces tests to run sequentially, fixing the problem.

UPDATE: Pushed two commits 

First:  refactor logging config to
a) reduce repetition
b) provide more granular logging config for Slick benchmarking and execution + parts of bitcoin-s

Second: use in memory SQLite database for unit tests


nkohen:
Is there any reason this can't be fixed by just using in-memory dbs alongside persistent read-only dbs where needed?

build.sbt
```diff
@@ -340,6 +340,8 @@ lazy val walletTest = project
   .settings(
     name := "bitcoin-s-wallet-test",
     libraryDependencies ++= Deps.walletTest,
+    // To avoid deadlock issues with SQLite
+    Test / parallelExecution := false,
```
Christewart: Isn't this already applied here? https://github.com/bitcoin-s/bitcoin-s-core/blob/91b0de7320bdfe9833dc12c375c1937a7a70b4a0/build.sbt#L112

torkelrogstad: That's only if we're on CI and macOS, as we had problems with RPC tests on macOS.

torkelrogstad:
@nkohen I swapped out the unit test DBs for in-memory databases. We got a nice speed improvement, but interestingly the deadlock issue still persisted. I'm guessing that the in-memory version of SQLite is a rather dumb implementation, so it probably has some limitations left over from file-based DBs that aren't changed.

PR #439, opened 2019-04-25
BlockchainBuilder

### LINKED ISSUE #428: Blockchain builder

Christewart:
Closes #428 

This creates a class called `BlockchainBuilder` and implements the `mutable.ReusableBuilder` api that is provided as a part of scala collections. 

This PR also moves the `connectTip` method into the companion object for Blockchain and forces the `connectTip` method to take a `Blockchain` as an argument.

I think there is an open question here about how much validation we want `BlockchainBuilder` to do. Do we want to do any validation at all? Do we want to do a full `connectTip`? I'm not sure.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -2,7 +2,7 @@ package org.bitcoins.chain.blockchain
 
 import org.bitcoins.chain.models.{BlockHeaderDAO, BlockHeaderDb}
 import org.bitcoins.chain.validation.{TipUpdateResult, TipValidation}
-import org.bitcoins.core.protocol.blockchain.{BlockHeader, ChainParams}
+import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.util.BitcoinSLogger
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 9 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -166,6 +167,20 @@ sealed abstract class BlockHeaderDAO
 
     database.runVec(aggregate)
   }
+
+  def getBlockchains()(implicit ec: ExecutionContext): Future[Vector[Blockchain]] = {
```
nkohen, line 180: scaladoc (this doesn't return the whole blockchain right?)

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -20,41 +20,80 @@ import scala.concurrent.{ExecutionContext, Future}
...
+
+  def fromHeaders(
+      headers: Vector[BlockHeaderDb]): Blockchain = {
+    Blockchain(headers)
+  }
+
+
+  /**
+    * Attempts to connect the given block header with the given blockchain
+    * @param header
```
nkohen: Some extra detail would be nice

torkelrogstad:
Did we conclude on what we should do with a memory-only blockchain?

To freshen your memory: a `BlockheaderDAO` is required to connect a header to the chain tip, and that's a somewhat weird requirement. It'd probably be a good idea to validate in memory, and then leave the option to persist to disk. 

Not necessarily something we need to do now, but good to get on the issue tracker?

PR #440, opened 2019-04-29
Natural language syntax for currencies

torkelrogstad:
In this PR we:

1) Refactor two test suites from Scalacheck to Scalatest
2) Introduce Int and Long syntax to currencies, similar to `scala.concurrent.duration.{DurationInt, DurationLong}`, allowing you to write:

```scala
import org.bitcoins.core.currency._

> 1.satoshi
> 5.bitcoins

import org.bitcoins.core.protocol.ln.currency._

> 5.picobitcoins
> 1.nBTC
```

I also tweak a few `toString` methods, fixing a bug in the LnCurrencyUnit `toString`s.


Christewart:
It's unfortunate that we need the redundant `Int` and `Long` stuff, I guess I never looked into why there is `DurationInt`/`DurationLong`

torkelrogstad:
WYM by redundant `Int`/`Long`? As far as I can tell, there's no common integer type that covers both `Int` and `Long`, so there needs to be one import for `Int` syntax and one import for `Long` syntax

torkelrogstad:
Theres `Numeric` (see [this SO question](https://stackoverflow.com/a/17999746/10359642) for more), but that also includes non-integer numbers

PR #444, opened 2019-04-29
BIP84 and BIP49

### LINKED ISSUE #408: Investigate BIP84 and BIP49
I was not aware of these BIPs until I poked around in Samourai today. Looks like Segwit addresses should use their own account in the BIP44 derivation scheme. Should look into this, so we stay compatible with the rest of the ecosystem.

Link to BIP84 (Derivation scheme for P2WPKH based accounts): https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki
Link to BIP49 (Derivation scheme for P2WPKH-nested-in-P2SH based accounts): https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki

torkelrogstad:
Closes #408 

In this PR we refactor a lot of code related to HD paths, to accommodate segwit and nested segwit paths from [BIP84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) and [BIP49](https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki).

 A new package `hd` is created in `core`. Everything `BIP32*` and `BIP44*` is moved into this package. The four things intended for public use in `hd` is: 
* `BIP32Path` - a generic BIP32 path
* `LegacyHDPath` - this was previously `BIP44Path`. Non-segwit addresses are derived with this path type
* `NestedSegWitHDPath` - derives P2WPKH-nested-in-P2SH addresses. BIP49
* `SegWitHDPath` - derives P2WPKH  adddresses. BIP84

There's also `HDPath`. This is intended to be the super type for BIP44, BIP49 and BIP84 paths. These BIPs are closely related, as they differ only in the constant used in the first level of key derivation. `HDPath` extends `BIP32Path`. Technically speaking, every BIP32 path is also a HD path. I'm using the terms `HDPath`, `HDAccount` etc in lack of a better word. I'm open to suggestions here, if you guys want to change this. 

`HDPath` is private to bitcoin-s, so it's not exposed to the user. 

This PR has implications for the `node` branch, but I'm targeting master to make it easier to review. 

UPDATE: I'm having issues with getting BIP49 correct. See [this SO](https://bitcoin.stackexchange.com/questions/87396/whats-the-magic-key-version-bytes-for-bip49) question for more

Todo: 
- [x] Test vectors from BIP84
- [x] Test vectors from BIP49
- [x] ExtKeyVersion for both BIPs
- [x] Add examples


core-test/src/test/scala/org/bitcoins/core/hd/HDPathTest.scala
```diff
@@ -0,0 +1,673 @@
...
+    forAll(HDGenerators.hdPath) { path =>
+      val pathFromString = HDPath.fromString(path.toString)
+      pathFromString match {
+        case Some(value: LegacyHDPath) =>
+          assert(value == path.asInstanceOf[LegacyHDPath])
+        case Some(value: SegWitHDPath) =>
+          assert(value == path.asInstanceOf[SegWitHDPath])
+        case Some(value: NestedSegWitHDPath) =>
+          assert(value == path.asInstanceOf[NestedSegWitHDPath])
+        case Some(other) => fail(s"$other is unknown HD path type!")
```
Christewart: I wish the type checker would work for things like this :cry:

core-test/src/test/scala/org/bitcoins/core/hd/HDPathTest.scala
```diff
@@ -0,0 +1,673 @@
...
+      assert(account.coin == coin)
+    }
+  }
+
+  behavior of "HDPath"
+
+  it must "have toString/fromString symmetry" in {
+    forAll(HDGenerators.hdPath) { path =>
+      val pathFromString = HDPath.fromString(path.toString)
+      pathFromString match {
```
Christewart: Should be re-written as this to preserve types 

```scala
    forAll(HDGenerators.hdPath) { path =>
      val pathFromString = HDPath.fromString(path.toString)
      val resultOpt = pathFromString.map {
        case value: LegacyHDPath =>
          assert(value == path.asInstanceOf[LegacyHDPath])
        case value: SegWitHDPath =>
          assert(value == path.asInstanceOf[SegWitHDPath])
        case value: NestedSegWitHDPath =>
          assert(value == path.asInstanceOf[NestedSegWitHDPath])
      }
      resultOpt.getOrElse(fail(s"$path did not have toString/fromString symmetry"))
    }
```

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -19,8 +19,10 @@ sealed abstract class ExtKey extends NetworkElement {
   require(bytes.size == 78,
           "ExtKey must be 78 bytes in size, got: " + bytes.size)
 
+  protected type VersionType <: ExtKeyVersion
```
Christewart, line 22: Any reason not to parameterize both this and `ExtKey` with `ExtKeyVersion`? 

i.e. 

```scala
sealed abstract class ExtPrivateKey[VersionType <: ExtKeyPrivVersion] extends ExtKey[VersionType]```

File: core/src/main/scala/org/bitcoins/core/hd/HDAccount.scala [hunk 4]
```diff
@@ -0,0 +1,25 @@
...
+) extends BIP32Path {
+  require(index >= 0, s"Account index ($index) must be positive!")
+
+  override val path: Vector[BIP32Node] = {
+    coin.path :+ BIP32Node(index, hardened = true)
+  }
+
+  def purpose: HDPurpose = coin.purpose
+
+  def toChain(chainType: HDChainType) =
```
Christewart: return type parameter

core/src/main/scala/org/bitcoins/core/hd/HDChainType.scala
```diff
@@ -0,0 +1,45 @@
...
+  * Address chain (external vs. change) used by
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#change BIP44]],
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki BIP84]]
+  * and
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki BIP49]]
+  *
+  * @see
+  */
+sealed abstract class HDChainType {
+  def index: Int
```
Christewart, line 15: can we put an invariant on this index? I.e it is positive?

core/src/main/scala/org/bitcoins/core/hd/HDPath.scala
```diff
@@ -0,0 +1,36 @@
+package org.bitcoins.core.hd
+import scala.util.Try
+
+private[bitcoins] trait HDPath[T <: HDPath[_]] extends BIP32Path {
```
Christewart: Should be able to do this ```private[bitcoins] trait HDPath[T <: HDPath[T]] extends BIP32Path```

core/src/main/scala/org/bitcoins/core/hd/HDPath.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+    * Increments the address index and returns the
+    * new path that can be passed into a
+    * [[org.bitcoins.core.crypto.ExtKey ExtKey]]
+    */
+  // TODO check out this cast
```
Christewart, line 11: ping

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
torkelrogstad: One downside of that is that it would cause a bunch of type parameters to be needed, with no real clear benefit compared to what's there now.

core/src/main/scala/org/bitcoins/core/hd/HDPath.scala
torkelrogstad: This is not ideal. We cannot match on `T`. In the same vein we cannot introduce a type param to `toPath` and match on that either. However in `toPath` we match on the `purpose` field (which is the level that defines the type of path we have, legacy/segwit/nested-segwit). It's impossible to construct a segwit path with a non-segwit purpose field, and similarly with the other types of paths. Therefore I _think_ this cast is always safe. Are you fine with leaving it in, if I also add a comment something along the line of what I'm writing here? 

I'm not really sure how to handle it without the cast.

PR #447, opened 2019-04-30
DB: Add utility method for listing tables in a DB

torkelrogstad:


db-commons/src/main/scala/org/bitcoins/db/SQLiteTableInfo.scala
```diff
@@ -0,0 +1,7 @@
+package org.bitcoins.db
+
+/**
+  * @param name The name of the table
+  * @param sql The SQL executed to create the table
+  */
+case class SQLiteTableInfo(name: String, sql: String)
```
Christewart, line 7: Does this have to be specific to sql lite?

torkelrogstad: Depends on what the output format looks like for other DBs. Or I guess we could have a generic `DBTableInfo` with different implicit readers:
https://github.com/bitcoin-s/bitcoin-s-core/blob/11020b6d3d82426f34232f551bb0078734aae37d/db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala#L41-L48

If I understand Slick correctly that's generally how their API works, by having a unified syntax API but different implicits (depending on `import slick.jdbc.SQLiteProfile.api._` vs `import slick.jdbc.PostgresProfile.api._` etc), and then the implicits do all the nitty gritty work of mapping the output of the DB in question to the Slick model

torkelrogstad: But this sounds like a premature optimization to me, would make more sense to come back to this when/if we add support for multiple DB systems

PR #450, opened 2019-04-30
WIP: Create ChainSync, BitcoindChainHandlerViaRpc, add simple ChainSyncTes…

Christewart:
This PR is creating a way to sync our blockchain with a third party data source. This can be a normal REST api, a bitcoind rpc client, or by directly hooking into the bitcoin p2p network. 

The most important file in this PR is [`ChainSync.scala`](https://github.com/Christewart/bitcoin-s-core/blob/54f1f759eae73fe70a8019570bc03cac4e79d66c/chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala) this trait current has one method called `sync`. The parameters for this method are 

- `chainHandler: ChainHandler` -- we use this to figure out the current chain state we have
- `getBlockHeaderFunc: DoubleSha256DigestBe => Future[BlockHeader]` -- this function allows us to arbitrarily query block headers. In this PR this is implemented as [`getBlockHeader`](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BlockchainRpc.scala#L42) on the rpc client
- `getBestBlockHashFunc: () => Future[DoubleSha256DigestBE]` -- this function allows us to query a 3rd party data source and figure what their best block hash is. In this PR this is implemented as [`getBestBlockHash`](https://github.com/bitcoin-s/bitcoin-s-core/blob/master/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BlockchainRpc.scala#L16) on our rpc client

The basic idea then is that if we are _not_ in sync with what is returned by `getBestBlockHash`, we walk the chain backwards with our `getBlockHeaderFunc` until we get to one of our chain tips. 

TODO:

- [ ] Add more complex test cases like reorgs
- [x] Consider the situtation where our third party data source has some how fallen behind us
- [ ] make traits for the function signatures above and move them to core so they are accessible to all projects

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
@@ -1,8 +1,9 @@
 package org.bitcoins.rpc.jsonmodels
 
-import org.bitcoins.core.crypto.{DoubleSha256DigestBE}
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
 import org.bitcoins.core.currency.Bitcoins
 import org.bitcoins.core.number.{Int32, UInt32}
+import org.bitcoins.core.protocol.blockchain.{BlockHeader, ChainParams}
 import org.bitcoins.core.wallet.fee.BitcoinFeeUnit
 
 sealed abstract class BlockchainResult
@@ -104,7 +105,25 @@ case class GetBlockHeaderResult(
     chainwork: String,
     previousblockhash: Option[DoubleSha256DigestBE],
     nextblockhash: Option[DoubleSha256DigestBE])
-    extends BlockchainResult
+    extends BlockchainResult {
+  def blockHeader(implicit cp: ChainParams): BlockHeader = {
+
+    val prevHash = {
+      if (height == 0 && previousblockhash.isEmpty) {
+        cp.genesisBlock.blockHeader.previousBlockHashBE
+      } else {
+        previousblockhash.get
+      }
+    }
+    BlockHeader(version = Int32(version),
+      //i think the only exception here is the genesis block? Why is this a option?
+      previousBlockHash = prevHash.flip,
+      merkleRootHash = merkleroot.flip,
+      time = time,
+      nBits = bits,
+      nonce = nonce)
+  }
+}
 
 case class ChainTip(
     height: Int,
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
@@ -3,7 +3,7 @@ package org.bitcoins.rpc.jsonmodels
 import org.bitcoins.core.crypto.DoubleSha256DigestBE
 import org.bitcoins.core.currency.Bitcoins
 import org.bitcoins.core.number.{Int32, UInt32}
-import org.bitcoins.core.protocol.blockchain.{BlockHeader, ChainParams}
+import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.wallet.fee.BitcoinFeeUnit
 
 sealed abstract class BlockchainResult
@@ -106,17 +106,18 @@ case class GetBlockHeaderResult(
     previousblockhash: Option[DoubleSha256DigestBE],
     nextblockhash: Option[DoubleSha256DigestBE])
     extends BlockchainResult {
-  def blockHeader(implicit cp: ChainParams): BlockHeader = {
+  def blockHeader: BlockHeader = {
 
+    //prevblockhash is only empty if we have the genesis block
+    //we assume the prevhash of the gensis block is the empty hash
     val prevHash = {
       if (height == 0 && previousblockhash.isEmpty) {
-        cp.genesisBlock.blockHeader.previousBlockHashBE
+        DoubleSha256DigestBE.empty
       } else {
         previousblockhash.get
       }
     }
     BlockHeader(version = Int32(version),
-      //i think the only exception here is the genesis block? Why is this a option?
       previousBlockHash = prevHash.flip,
       merkleRootHash = merkleroot.flip,
       time = time,
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/sync/ChainSyncTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/sync/ChainSyncTest.scala
@@ -1,0 +1,49 @@
+package org.bitcoins.chain.blockchain.sync
+
+import akka.actor.ActorSystem
+import org.bitcoins.chain.api.ChainApi
+import org.bitcoins.chain.util.{BitcoindChainHandlerViaRpc, ChainUnitTest}
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+import org.scalatest.FutureOutcome
+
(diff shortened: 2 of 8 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
```diff
@@ -104,7 +105,25 @@ case class GetBlockHeaderResult(
...
+
+    val prevHash = {
+      if (height == 0 && previousblockhash.isEmpty) {
+        cp.genesisBlock.blockHeader.previousBlockHashBE
+      } else {
+        previousblockhash.get
+      }
+    }
+    BlockHeader(version = Int32(version),
+      //i think the only exception here is the genesis block? Why is this a option?
```
Christewart: This is the case, on another note i think we should get our `ChainApi` to return something similar that bitcoin core does (mimic the bitcoin core api) https://bitcoin.org/en/developer-reference#getblockheader

torkelrogstad: Does this comment refer to line 106/115? Makes sense to me. Did some basic testing with CLI, and it appears so. It would be nice to get rid of the option, somehow... Could we do something _slightly bad_ like special-casing it to the empty block hash, to get rid of the option?

chain-test/src/test/scala/org/bitcoins/chain/blockchain/sync/ChainSyncTest.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.bitcoins.chain.util.{BitcoindChainHandlerViaRpc, ChainUnitTest}
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+import org.scalatest.FutureOutcome
+
+import scala.concurrent.Future
+
+class ChainSyncTest extends ChainUnitTest {
+  override type FixtureParam = BitcoindChainHandlerViaRpc
+
+  override implicit val system = ActorSystem(s"chain-sync-test-${System.currentTimeMillis()}")
```
torkelrogstad, line 14: Not strictly related to this PR, but we could move this into `ChainUnitTest` and do `ActorSystem(s"${getClass.getSimpleName}-${System.currentTimeMillis()}")`

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
```diff
@@ -0,0 +1,110 @@
...
+    * If we do not have the same chain, we sync our chain handler until we are at the same best block hash
+    * @param chainHandler our internal chain handler
+    * @param getBlockHeaderFunc a function that we can call to retrieve a block
+    * @param getBestBlockHashFunc a function that can call a third party source (bitcoind, block explorer etc)
+    *                             to retrieve what the best block is on the blockchain
+    * @param ec
+    * @return
+    */
+  def sync(chainHandler: ChainHandler,
+           getBlockHeaderFunc: DoubleSha256DigestBE => Future[BlockHeader],
```
torkelrogstad, line 24: Nit: I think we should remove the `Func` part of the param names, makes it more awkward to type it out

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
```diff
@@ -0,0 +1,110 @@
...
+           getBestBlockHashFunc: () => Future[DoubleSha256DigestBE])(implicit ec: ExecutionContext): Future[ChainApi] = {
```
torkelrogstad, line 25: Would it work to change the definition to `getBestBlockHashFunc: => Future[DoubleSha256DigestBE])`? Cleaner API, IMO

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
```diff
@@ -0,0 +1,110 @@
...
+    val currentTipsF: Future[Vector[BlockHeaderDb]] = {
+      chainHandler.blockchain.blockHeaderDAO.chainTips
+    }
+
+    //TODO: We are implicitly trusting whatever
```
torkelrogstad, line 30: File an issue for it?

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
```diff
@@ -0,0 +1,110 @@
...
+    * @param tips the best block header we know about
+    * @param bestBlockHash the best block header seen by our third party data source
+    * @param getBlockHeaderFunc how we can retrieve block headers
+    * @param ec
+    * @return
+    */
+  private def syncTips(chainApi: ChainApi,
+                       tips: Vector[BlockHeaderDb],
+                       bestBlockHash: DoubleSha256DigestBE,
+                       getBlockHeaderFunc: DoubleSha256DigestBE => Future[BlockHeader])(implicit ec: ExecutionContext): Future[ChainApi] = {
```
torkelrogstad, line 67: same nit here as above, wrt to the param name

chain-test/src/test/scala/org/bitcoins/chain/blockchain/sync/ChainSyncTest.scala
```diff
@@ -46,4 +46,30 @@ class ChainSyncTest extends ChainUnitTest {
 
       }
   }
+
+
+  it must "not fail when syncing a chain handler that is synced with it's external data source" in {
```
torkelrogstad, line 51: This test text is a bit unclear to me, WYM?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
Christewart: The problem is figuring out what network that empty block hash is associated with, which is why I included the implicit chain param. But i think that is a bad solution to. I'm open to ideas.

torkelrogstad: Looks like the prev block has is `DoubleSha256Digest.empty` regardless of the network? Looking at https://github.com/bitcoin-s/bitcoin-s-core/blob/100221e94321f96fdad8042a9d7ca2e6e9890688/core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala#L142-L150

chain-test/src/test/scala/org/bitcoins/chain/blockchain/sync/ChainSyncTest.scala
Christewart: The intention here is that if we are in sync with our third party data source (we have the same blockchain tip) we shouldn't walk back on the chain.

This was actually covered before this commit, but there was an edge case where if our third party data source was _behind us on the blockchain_ we would still walk all the way back on our chain until we got to the genesis block. This was because we were only checking the tips, _not to see if the header existed already in the chain database_. If the header exists, we are making the assumption we have properly synced after that header.

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
Christewart: I am going to leave it for now, but I do think we should follow conventions like we do for `Opt` (Option) `F` (Fututre) and one for function. I could be convinced to use `Fn` instead of `Func`

PR #452, opened 2019-05-01
2019 05 01 wallet ammonite scripts pt2

Christewart:
This is built on top of #450 and supersedes #430 

If you look at #430, we have had issues with jeromq when trying to link a bitcoind instance to our chain project via zmq. This PR links a bitcoind instance to our chain project using `ChainSync` which is introduced in #450. Now we sync our internal chain state in bitcoin-s via rpc instead of zmq. 

To run this script you need to do 

```
$ sbt "doc/run doc/src/main/scala/org/bitcoins/doc/wallet/create-wallet.sc"
```

I did have issues running the script _multiple_ (a single run works) times in the sbt console. I believe this has something to do with sharing the same jvm as sbt, but I tried using [`fork in run := true`](https://www.scala-sbt.org/1.x/docs/Forking.html) in our doc project and that did not work. This is probably because I don't understand the relationship between sbt, ammonite, and the script.


doc/src/main/scala/org/bitcoins/doc/wallet/create-wallet.sc
```diff
@@ -0,0 +1,230 @@
...
+  chainApi.getBlockCount.map(count => logger.info(s"bitcoin-s blockcount=${count}"))
+}
+
+val walletF = bitcoinsLogF.flatMap { _ =>
+  //create tables
+  val dropTablesF = WalletDbManagement.dropAll(walletDbConfig)
+  val createTablesF = dropTablesF.flatMap(_ => WalletDbManagement.createAll(walletDbConfig))
+  createTablesF.flatMap { _ =>
+    Wallet.initialize(walletAppConfig)
+      .map(_.asInstanceOf[InitializeWalletSuccess].wallet)
```
torkelrogstad:
```suggestion
      .collect{ case success: InitializeWalletSuccess => success.wallet }
```

doc/src/main/scala/org/bitcoins/doc/wallet/create-wallet.sc
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+
+val bitcoinsAddrF = walletF.flatMap(_.getNewAddress())
+
+//send money to our wallet with bitcoind
+val amt = Bitcoins.one
```
torkelrogstad, line 125: `1.bitcoin` ;)

doc/src/main/scala/org/bitcoins/doc/wallet/create-wallet.sc
```diff
@@ -0,0 +1,230 @@
...
+}
+val createdGenHeaderF = blockHeaderTableF.flatMap(_ => blockHeaderDAO.create(genesisHeader))
+
+val chainF = createdGenHeaderF.map(h => Vector(h))
+
+val blockchainF = chainF.map(chain => Blockchain(chain))
+
+val chainHandlerF = blockchainF.map(blockchain => ChainHandler(blockHeaderDAO, chainAppConfig))
+
+val chainApi101BlocksF = sync(chainHandlerF, 101)
```
torkelrogstad, line 106: I'm getting an error "Wrong forward reference" for `sync` here. It's in Intellij only so probably not a big deal, but might make sense to move it to before its used?

torkelrogstad:
It looks like you deleted `secp256k1jni/natives/linux_64/libsecp256k1.la`. I take it this was not on purpose?

Christewart:
d36a2e0 attempts to fix the problems i'm having with closing the zmq subscriber. 

Related reading I used to fix this problem 

https://github.com/zeromq/jeromq/issues/116

https://stackoverflow.com/questions/38215839/jeromq-subscriber-in-a-runnable

I _think_ my issue was a combination of two things. 

1. Not using `setLinger()`. This setting on the socket gives us a deadline for our subscribers to consume a message before it is discarded. See the `setLinger()` documentation for more information.
2. Using `subscriber.close()` to close the underlying network socket. I think I got bad advice from someone about trusting `context.term()` to release the underlying resources.

PR #458, opened 2019-05-03
Node rebase

torkelrogstad:
This is a rebase PR of node onto master. The reason this rebase became so hairy was poor initial design on my side. The wallet had just one xpub for the entire wallet, but to be BIP-compliant we need to have one xpub per account, and that complicate things a bunch. 

To fix this, I added the concept of _BIP32 path diffing_ in 8791033399a563536b0c2f8f68e54296dd093692. See that commit's tests and Scaladoc for more information. I'm not `diff` is the optimal method/concept name, so I'm open to suggestions here. Also, what do you think of adding an operator that's aliased to this method? I'm a sucker for sleek syntax in Scala, and I was thinking that being able to write something like this would be very cool:
```scala
BIP32Path.fromString("m/44'/0'/0'") ~ BIP32Path.fromString("m/44'/0'/0'/0/2") == BIP32Path.fromString("m/0/2")
```

I've also been thinking of implementing Scala syntax sugar for BIP32 paths, so we down the road we could potentially even do this:
```scala
(m / 44.^ / 0.^ / 0.^)  ~ (m / 44.^ / 0.^ / 0.^ / 0 / 2) == m / 0 / 2
```
If you're not too keen on adding (potentially confusing) operators I'm cool with that:-)

## Why do we use path diffing?

Look at the following code, where we generate a new address within the wallet: 
https://github.com/bitcoin-s/bitcoin-s-core/blob/7274e5ed9b9f7d1ae5d48e9ea7c2969d647e8f2d/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L201-L216

We have three things, at this point in the method: 
1) The account xpub
2) The account path (`m/84'/0'/0'` for the first segwit account in the wallet)
3) The address path (`addrPath`/segwitPath, `m/84'/0'/0'/0/0` for the first external address in the account)

If we now do `pubkeyForAddr = xpub derive addrPath` we end up with deriving all fields of the address path. This is wrong, as the xpub is already derived three levels deep (segwit/legacy/nested, mainnet/testnet, account index) from the root wallet seed. What we want to do is take the last two elements (change/external, address index) of the address path (as those are _not_ in the account path), and _only_ derive our public key using those elements. 

More visually:

```
m/84'/0'/0        // account path 
m/84'/0'/0'/0/0   // address path
      ^ our xpub is derived from the root seed at this level 

m/0/0      // diff of the two paths
^ we must use this to derive the address public key from the account xpub
```

Previously I (wrongly) believed that the xpub was at the root level (not the third), so I just did `xpub derive addresspath`. 

I've manually verified the public keys generated using [Ian Colemans BIP39 site](https://iancoleman.io/bip39/). At a later point I plan to generate more rigourous test vectors we can verify against.


core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
```diff
@@ -127,4 +128,67 @@ class BIP32PathTest extends BitcoinSUnitTest {
...
+  it must "do path diffing" in {
+    {
+      val first = BIP32Path.fromString("m/44'/1'")
+      assert(first.diff(first).contains(BIP32Path.empty))
+    }
+
+    {
+      val first = BIP32Path.fromString("m/44'/0'/0'")
+      val second = BIP32Path.fromString("m/44'/0'/0'/0/2")
+      val expected = BIP32Path.fromString("m/0/2")
```
Christewart, line 141: So this seems weird to me, why do we keep the `m` prefix? Isn't this going to confuse people to believe that this is an _absolute_ path and not a _relative_ one?

Christewart: Same comment applies to the rest of these test vectors i believe

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -6,6 +6,76 @@ import org.bitcoins.core.number.UInt32
...
+  /**
+    * BIP32 paths can be subsets/superset of each other.
+    * If all elements in a path `p` is included in a path
+    * `P`, (i.e. `p` is a subset of `P`), `p.diff(P)`
+    * is the elements from `P` that is not in `p`.
+    *
+    * @example
+    * {{{
+    *  // equal paths
+    * m/44'/1' diff m/44'/1' == Some(BIP32Path.empty)
```
Christewart, line 18: Maybe should return `None` here?

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -6,6 +6,76 @@ import org.bitcoins.core.number.UInt32
...
+        .foldLeft(Option(BIP32Path.empty)) {
+          // we encountered an error along the way, return
+          // none
+          case (None, _) => None
+
+          // we've reached the end of our path, append
+          // the element from their path but don't
+          // include the previous one (as
+          // that's shared)
+          case (Some(_), ((None, their), index)) if index == path.length =>
```
Christewart, line 59: Just a general note: `if` guards inside of a pattern match break pattern match exhaustiveness unfortunately.

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -6,6 +6,76 @@ import org.bitcoins.core.number.UInt32
...
+    if (path.length > otherPath.length) {
+      None
+    } else if (path == otherPath) {
+      Some(BIP32Path.empty)
+    } else {
+      val lengthDiff = otherPath.length - path.length
+
+      val extendedPath: Vector[Option[BIP32Node]] = path.map(Some(_)) ++
+        Vector.fill[Option[BIP32Node]](lengthDiff)(None)
+
```
Christewart, line 44: For readability 

```scala
      val pathsWithIndices: Vector[((Option[BIP32Node],BIP32Node),Int)] = extendedPath
        .zip(otherPath)
        .zipWithIndex
      val calculatedDiff: Option[BIP32Path] = pathsWithIndices...
```

core/src/main/scala/org/bitcoins/core/hd/HDPath.scala
```diff
@@ -1,16 +1,18 @@
 package org.bitcoins.core.hd
 import scala.util.Try
 
-private[bitcoins] trait HDPath[T <: HDPath[T]] extends BIP32Path {
+private[bitcoins] trait HDPath extends BIP32Path {
+
+  protected type NextPath <: HDPath
```
Christewart, line 33: Comment on the intention of this?

Christewart: Is there any way we can do something like `this.type` for this? Seems like we just override it immediately in the sub types

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -22,31 +23,35 @@ import scodec.bits.BitVector
...
 
-  lazy val hdCoin: HDCoin = {
+  /** The default HD coin */
+  lazy private val DEFAULT_HD_COIN: HDCoin = {
     val coinType = chainParams match {
       case MainNetChainParams                         => HDCoinType.Bitcoin
       case RegTestNetChainParams | TestNetChainParams => HDCoinType.Testnet
     }
-    HDCoin(HDPurposes.SegWit, coinType)
+    HDCoin(Wallet.DEFAULT_HD_PURPOSE, coinType)
```
Christewart, line 54: Is there a reason for changing this?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -78,18 +83,25 @@ sealed abstract class Wallet extends UnlockedWalletApi with BitcoinSLogger {
...
+    val utxo: UTXOSpendingInfoDb = addressDb match {
+      case segwitAddr: SegWitAddressDb =>
+        SegWitUTOXSpendingInfodb(
+          id = None,
+          outPoint = outPoint,
+          output = output,
+          privKeyPath = segwitAddr.path,
+          scriptWitness = segwitAddr.witnessScript
+        )
+      case badCombination =>
```
Christewart: Is this just unimplemented for now? If so, i'd rather just explicitly match and use `???`

core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
torkelrogstad: That's just because all paths must begin with `m/at-least-one-level`. 

All paths are relative in HD schemes anyway, as it all depends on what your vantage point (your xpriv) and direction (your path) are.

Something we could add though (although I'd do it a later point), is a wrapper around an xpriv/xpub and a path, allowing us to go back and forth, having a notion of relativeness/absoluteness

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
torkelrogstad: `m/44'/1' derive BIP32Path.empty` is a completely valid operation though
Any path is a subset of itself, and if diffing is defined for subsets, then diffing something with itself should also be defined

Not that big of a point though, I'm fine with both. Do you have strong feelings on it?

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
torkelrogstad: :-(

core/src/main/scala/org/bitcoins/core/hd/HDPath.scala
torkelrogstad: Added a explanatory comment in `HDPath.scala`

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: You'll notice that it's set to be `HDPurposes.SegWit` in Wallet.scala on line 336. The reason for changing it is making it clear it can be any `HDPurpose`, which at a later point in time can be set by the user either through configuration or by some method call. It's mentioned in the comments above `DEFAULT_HD_PURPOSE`

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: explicitly matched a couple of places

Issue #392, opened 2019-03-26
Support new config format for networks in bitcoind RPC


torkelrogstad:
https://github.com/bitcoin/bitcoin/blob/master/doc/bitcoin-conf.md#network-specific-options

PR #460, opened 2019-05-05
2019 05 05 sync chain

Christewart:
This builds ontop of #452 

The creates a new file called `sync-chain.sc` in the doc project which is intended to be a script to show you can sync the chain from a external data source -- such as bitcoind -- to our own local chain state. 

The second time you run the script only _new_ blocks will be synced to our local chain state. 

This is a convenient script to run if your bitcoin-s instance does not always have direct communicate with a bitcoind node, this allows you to quickly sync the block headers that you missed in the time that the chain project was offline.

EDIT: 

This script assumes that you have a `bitcoind` instance running in the background and that you have `~/.bitcoin/bitcoin.conf` setup. You can pass in an alternative datadir if you wish by construct a `new java.io.File()` that points to that datadir [here](https://github.com/bitcoin-s/bitcoin-s-core/pull/460/files#diff-ee0d37fbe3c0a832b861f81c99bb061bR42)

How to run the script
```
$  sbt "doc/run doc/src/main/scala/org/bitcoins/doc/chain/sync-chain.sc"
```

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
@@ -1,7 +1,24 @@
 package org.bitcoins.chain.config
 
 import org.bitcoins.chain.db.ChainDbConfig
+import org.bitcoins.chain.models.BlockHeaderDAO
 import org.bitcoins.db.AppConfig
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success}
 
-case class ChainAppConfig(dbConfig: ChainDbConfig)
-    extends AppConfig[ChainDbConfig]
+case class ChainAppConfig(dbConfig: ChainDbConfig) extends AppConfig[ChainDbConfig] {
+
+  def isDbInitialized()(implicit ec: ExecutionContext): Future[Boolean] = {
+    val bhDAO = BlockHeaderDAO(this)
+    val p = Promise[Boolean]()
+    val isDefinedOptF = {
+      bhDAO.read(chain.genesisBlock.blockHeader.hashBE).map(_.isDefined)
+    }
+    isDefinedOptF.onComplete {
+      case Success(bool) => p.success(bool)
+      case Failure(err) => p.success(false)
+    }
+
+    p.future
+  }
+}
--- a/chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
@@ -16,7 +16,9 @@ case class ChainAppConfig(dbConfig: ChainDbConfig) extends AppConfig[ChainDbConf
     }
     isDefinedOptF.onComplete {
       case Success(bool) => p.success(bool)
-      case Failure(err) => p.success(false)
+      case Failure(err) =>
+        logger.info(s"Failed to init chain app err=${err.getMessage}")
+        p.success(false)
     }
 
     p.future
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -56,7 +56,7 @@ object Blockchain extends BitcoinSLogger {
       val nested: Vector[Future[BlockchainUpdate]] = blockchains.map { blockchain =>
         val tip = blockchain.tip
         logger.debug(
-          s"Attempting to add new tip=${header.hashBE.hex} to chain with current tips=${tip.hashBE.hex}")
+          s"Attempting to add new tip=${header.hashBE.hex} with prevhash=${header.previousBlockHashBE.hex} to chain with current tips=${tip.hashBE.hex}")
         val tipResultF = TipValidation.checkNewTip(newPotentialTip = header,
           currentTip = tip,
           blockHeaderDAO = blockHeaderDAO)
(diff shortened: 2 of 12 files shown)


chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
```diff
@@ -1,6 +1,25 @@
...
+import org.bitcoins.chain.models.BlockHeaderDAO
 import org.bitcoins.db.AppConfig
 
-case class ChainAppConfig(dbConfig: ChainDbConfig) extends AppConfig[ChainDbConfig]
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success}
+
+case class ChainAppConfig(dbConfig: ChainDbConfig) extends AppConfig[ChainDbConfig] {
+
+  def isDbInitialized()(implicit ec: ExecutionContext): Future[Boolean] = {
```
torkelrogstad, line 11: This might just be personal preference, but I find something like 
```scala
   bhDAO
      .read(chain.genesisBlock.blockHeader.hashBE)
      .map(_.isDefined)
      .recover {
        case _: Exception => false
}
```

to be more readable and less imperative

torkelrogstad (changes requested):
So I fiddled quite a bit with this, getting it to run. I kept getting SQLite errors: `Error on reading genesis block: java.sql.SQLException: regtestDb.db - Interrupted during connection acquisition`. SQLite is really sensitive about concurrent connections, so this should be avoided where possible when using SQLite. It seems Hikari is a big source of these interruptions, so setting `commonDbSettings.db.connectionPool = disabled` made the script work for me. I'd suggest adding that to this pr, in `db.conf`. In general there's no point in doing connection pools with SQLite, as it locks on all writes anyway.

Also #392 should probably be made a priority. I'm using the new configuration format for my `bitcoin.conf`, and it's a big PITA that our RPC client doesn't support that

Christewart:
So strangely enough this ended up being a problem for me as well after i restarted my laptop so I applied the fix to the database. 

faabcb8 fixes a bug in the POW.getNextWorkRequired method where we are calculating tcommhe first block header height of a POW interval. The bug revolved around the `-` operator not being commutative (duh!) which meant we were subtracting things out of order, which produced a negative number. We cannot have negative block heights.

Christewart:
When running this locally you sould see something like this as output, depending how long ago you synced, you have a lot of block header hashes show up as being synced

```
$ sbt "doc/run doc/src/main/scala/org/bitcoins/doc/chain/sync-chain.sc"
Compiling (synthetic)/ammonite/predef/interpBridge.sc
Compiling (synthetic)/ammonite/predef/DefaultPredef.sc
Compiling /home/chris/dev/bitcoin-s-core/doc/src/main/scala/org/bitcoins/doc/chain/sync-chain.sc
2019-05-07 11:15:06.612 [run-main-0] INFO  o.b.d.c.SyncChain.<init>(45) - Done configuring rpc client
2019-05-07 11:15:06.616 [run-main-0] INFO  o.b.d.c.SyncChain.<init>(54) - Creating chain tables

Exception: sbt.TrapExitSecurityException thrown from the UncaughtExceptionHandler in thread "run-main-0"
2019-05-07 11:15:07.241 [sync-chain-1557245705986-akka.actor.default-dispatcher-5] INFO  o.b.d.c.SyncChain.$anonfun$syncedChainApiF$1(63) - Beginning sync to bitcoin-s chain state
2019-05-07 11:15:08.288 [sync-chain-1557245705986-akka.actor.default-dispatcher-3] INFO  o.b.c.b.s.ChainSync$.$anonfun$syncTips$4(94) - Best tip from third party=70ae52baf5c49ddc98e7485bfcee3380aa93c61d812aabc76b500f06a5c28e2d currentTips=Vector(25b1f8f2e833d5f4b5cfb6ef7d44a2966a3334336894a91497c910289608a202)
2019-05-07 11:15:08.294 [sync-chain-1557245705986-akka.actor.default-dispatcher-2] INFO  o.b.c.b.s.ChainSync$.$anonfun$syncTips$7(113) - Attempting to sync 1 blockheader to our chainstate
2019-05-07 11:15:08.481 [sync-chain-1557245705986-akka.actor.default-dispatcher-2] INFO  o.b.d.c.SyncChain.$anonfun$syncResultF$2(68) - chain api blockcount=3100
2019-05-07 11:15:08.484 [sync-chain-1557245705986-akka.actor.default-dispatcher-5] INFO  o.b.d.c.SyncChain.$anonfun$syncResultF$3(70) - bitcoind blockcount=3100
2019-05-07 11:15:08.485 [sync-chain-1557245705986-akka.actor.default-dispatcher-5] INFO  o.b.d.c.SyncChain.$anonfun$res_34$1(75) - Sync result=Success(())
[ERROR] [05/07/2019 11:15:08.503] [sync-chain-1557245705986-akka.actor.default-dispatcher-3] [akka://sync-chain-1557245705986/system/pool-master] connection pool for PoolGateway(hcps = HostConnectionPoolSetup(localhost,18332,ConnectionPoolSetup(ConnectionPoolSettings(4,0,5,32,1,100 milliseconds,2 minutes,30 seconds,ClientConnectionSettings(Some(User-Agent: akka-http/10.1.7),10 seconds,1 minute,512,None,WebSocketSettings(<function0>,ping,Duration.Inf,akka.http.impl.settings.WebSocketSettingsImpl$$$Lambda$6476/49403691@6b9e58b8),List(),ParserSettings(2048,16,64,64,8192,64,8388608,8388608,256,1048576,Strict,RFC6265,true,Set(),Full,Error,Map(If-Range -> 0, If-Modified-Since -> 0, If-Unmodified-Since -> 0, default -> 12, Content-MD5 -> 0, Date -> 0, If-Match -> 0, If-None-Match -> 0, User-Agent -> 32),false,true,akka.util.ConstantFun$$$Lambda$6466/355448769@6213e87e,akka.util.ConstantFun$$$Lambda$6466/355448769@6213e87e,akka.util.ConstantFun$$$Lambda$6467/1313618816@7bd1271c),None,TCPTransport),New,1 second),akka.http.scaladsl.HttpConnectionContext$@5d7208ad,akka.event.MarkerLoggingAdapter@20b32ec0))) has shut down unexpectedly
```

You can verify that the sync worked with by inspecting the sqlite3 database

```
chris@chris:~$ sqlite3 ~/.bitcoin-s/regtest/chaindb.sqlite
SQLite version 3.22.0 2018-01-22 18:45:57
Enter ".help" for usage hints.
sqlite> select COUNT(*) from block_headers ;
3101
```

PR #462, opened 2019-05-06
Store encrypted mnemonic to disk

### LINKED ISSUE #409: Implement storage of mnemonic/root privkey, locking/unlocking of wallet
We'd like to store sensitive key material in **only one place**. A flat file for the encrypted mnemonic would make sense. I have this memory of seeing AES encrypted data in a JSON file with keys `iv`, `payload` and some cipher specs. There's probably a standard for this, we should try and conform with that.

When [`unlock`](https://github.com/torkelrogstad/bitcoin-s-core/blob/80061bf6f2f0cb25b787e90a374097e70750e906/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L78) is called the mnemonic should be read from file and decrypted

When [`lock`](https://github.com/torkelrogstad/bitcoin-s-core/blob/80061bf6f2f0cb25b787e90a374097e70750e906/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L115`) is called we should do a best effort to delete the mnemonic and xpriv from memory. Should we also verify that the content on disk matches our memory copy?

There should also be a method for updating the wallet encryption pasphrase.

torkelrogstad:
Depends on #463 
Closes #409 (save for updating passphrase, should be done in a later PR with separate issue)

This PR introduces a proper separation between a locked wallet an an unlocked wallet. 

It also add uJson as a JSON parsing dependency in the wallet project. [uJson docs](http://www.lihaoyi.com/upickle/#uJson)

Todo: 
 - [x] Finish up the `lock` method
- [x] Implement the `unlock` method
- [x] Get rid of the `isTest` param in `WalletStorage`


wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -0,0 +1,277 @@
...
+import scala.util.Success
+import scala.util.Failure
+import org.bitcoins.db.AppConfig
+import scala.concurrent.ExecutionContext
+import org.bitcoins.wallet.ReadMnemonicError.DecryptionError
+import org.bitcoins.wallet.ReadMnemonicError.JsonParsingError
+
+abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
+
+  val addressDAO: AddressDAO = AddressDAO()
```
Christewart: private for these?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -0,0 +1,277 @@
...
+  val accountDAO: AccountDAO = AccountDAO()
+  val utxoDAO: UTXOSpendingInfoDAO = UTXOSpendingInfoDAO()
+
+  override def getBalance(): Future[CurrencyUnit] = listUtxos().map { utxos =>
+    utxos.map(_.value).fold(0.bitcoin)(_ + _)
+  }
+
+  /** The default HD coin */
+  lazy private val DEFAULT_HD_COIN: HDCoin = {
```
Christewart: nit: I usually like to put the access modifier first, i.e.

```private lazy val DEFAULT_HD_COIN = ...```

File: wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala [hunk 3]
```diff
@@ -0,0 +1,277 @@
...
+          privKeyPath = segwitAddr.path,
+          scriptWitness = segwitAddr.witnessScript
+        )
+      case otherAddr @ (_: LegacyAddressDb | _: NestedSegWitAddressDb) =>
+        throw new IllegalArgumentException(
+          s"Bad utxo $otherAddr. Note: Only Segwit is implemented")
+    }
+
+    utxoDAO.create(utxo).map { written =>
+      import written.{outPoint => writtenOut}
```
Christewart: What is the point of this?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -0,0 +1,277 @@
...
+      logger.info(
+        s"Successfully inserted UTXO ${writtenOut.txId.hex}:${writtenOut.vout.toInt} into DB")
+      logger.info(s"UTXO details: ${written.output}")
+      utxo
```
Christewart: We aren't returning `written` here? That means what we are not returning the auto incremented id right?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -0,0 +1,277 @@
...
+    val voutIndexOutOfBounds: Boolean = {
+      val voutLength = transaction.outputs.length
+      val outOfBunds = voutLength <= vout.toInt
+
+      if (outOfBunds)
+        logger.error(
+          s"TX with TXID ${transaction.txId.hex} only has $voutLength, got request to add vout ${vout.toInt}!")
+      outOfBunds
+    }
+
```
Christewart, line 140: I think this ends up being much cleaner

```scala
    if (voutIndexOutOfBounds) {
      Future.successful(VoutIndexOutOfBounds)
    } else {
      val output = transaction.outputs(vout.toInt)
      val outPoint = TransactionOutPoint(transaction.txId, vout)
      // second check: do we have an address associated with the provided
      // output in our DB?
      val addressDbEitherF = findAddress(output.scriptPubKey)

      // insert the UTXO into the DB
      addressDbEitherF.flatMap { addressDbE =>
        val biasedE: Either[AddUtxoError, Future[UTXOSpendingInfoDb]] = for {
          addressDb <- addressDbE
        } yield writeUtxo(output, outPoint, addressDb)

        EitherUtil.liftRightBiasedFutureE(biasedE)

      }.map {
        case Right(_) => AddUtxoSuccess(this)
        case Left(e)  => e
      }
    }
```

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -3,255 +3,41 @@ package org.bitcoins.wallet
...
-  /** @inheritdoc */
-  override protected[wallet] def getDefaultAccount(): Future[AccountDb] = {
-    for {
-      account <- accountDAO.read((DEFAULT_HD_COIN, 0))
-    } yield
-      account.getOrElse(
-        throw new RuntimeException(
-          s"Could not find account with ${DEFAULT_HD_COIN.purpose.constant} " +
-            s"purpose field and ${DEFAULT_HD_COIN.coinType.toInt} coin field"))
+    WalletStorage.writeMnemonicToDisk(encrypted)
```
Christewart, line 38: can this fail? If it can, it doesn't look like it is represented in the type returned here? If lock fails that is problem it seems. I don't have any suggestions at the moment, but it does seem important.

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
```diff
@@ -0,0 +1,231 @@
...
+import org.bitcoins.core.crypto.AesEncryptedData
+import org.bitcoins.core.crypto.AesSalt
+import scala.util.Failure
+import scala.util.Success
+import java.nio.file.Paths
+import java.nio.file.Path
+import scala.util.Try
+import org.bitcoins.db.AppConfig
+
+// what do we do if seed exists? error if they aren't equal?
```
Christewart, line 18: This is a good question. @nkohen any opinions here?

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+    * Writes the encrypted mnemonic to disk.
+    * If we encounter a file in the place we're about
+    * to write to, we move it to a backup location
+    * with the current epoch timestamp as part of
+    * the file name.
+    */
+  def writeMnemonicToDisk(mnemonic: EncryptedMnemonic)(
+      implicit config: AppConfig): Path = {
```
Christewart, line 38: I don't really like the return type here, can we return like a `Try[]` or something like that to indicate that this can fail?

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
```diff
@@ -0,0 +1,231 @@
...
+        CIPHER_TEXT -> encrypted.cipherText.toHex,
+        SALT -> encrypted.salt.value.toHex
+      )
+    }
+
+    val path = config.datadir.resolve(ENCRYPTED_SEED_FILE_NAME)
+
+    logger.debug(s"Writing mnemonic to $path")
+
+    val writtenJs = write(jsObject)
```
Christewart: I would prefer if we used things like `ujson.write` rather than importing everything in the `ujson` package object. It makes it easier to read code as `write` is very generic.

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
```diff
@@ -0,0 +1,231 @@
...
+
+    def writeJsToDisk() = {
+      val writtenPath = Files.write(path, writtenJs.getBytes)
```
Christewart: any reason we are writing bytes here rather than just the plain text?

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
```diff
@@ -0,0 +1,231 @@
...
+              Left(JsonParsingError("JSON contents was not hex strings")))
+      }
+    encryptedEither
+  }
+
+  /**
+    * Reads the wallet mmemonic from disk and tries to parse and
+    * decrypt it
+    */
+  def readMnemonicFromDisk(passphrase: AesPassword)(
```
Christewart: Perhaps this should be named to `decryptMnemonicFromDisk()`?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: I assume that importing a value doesn't allocate a new pointer/object, so this is instead of doing `val writtenOut = written.outPoint`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: You're right, good catch

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: If we encounter a file where we're supposed to write we move it to a backup location. So the only failing I can see it doing is if things start going haywire with the OS' file system. I'm not sure if that's something we should encode into the return type, as everything starts breaking down at that point.

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
torkelrogstad: See my response to your other comment on this

wallet/src/main/scala/org/bitcoins/wallet/WalletStorage.scala
torkelrogstad: `Files.write(path, writtenJs)` results in 
```
overloaded method value write with alternatives:
  (x$1: java.nio.file.Path,x$2: Iterable[_ <: CharSequence],x$3: java.nio.file.OpenOption*)java.nio.file.Path <and>
  (x$1: java.nio.file.Path,x$2: Array[Byte],x$3: java.nio.file.OpenOption*)java.nio.file.Path
 cannot be applied to (java.nio.file.Path, String)
```

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
Christewart: Yeah, I guess I think we should usually default to doing "the dumb thing" unless there is a good reason. I don't understand the performance benefits here and it reduces readability

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -0,0 +1,271 @@
...
+    } yield
+      account.getOrElse(
+        throw new RuntimeException(
+          s"Could not find account with ${DEFAULT_HD_COIN.purpose.constant} " +
+            s"purpose field and ${DEFAULT_HD_COIN.coinType.toInt} coin field"))
+  }
+
+}
+
+object LockedWallet {
```
Christewart, line 263: nit: We can probably just make the above a `case class` right? and remove this companion object?

Christewart:
@torkelrogstad i think we also talked about adding a small document talking about how we store the encrypted mnemonic, where we store it, what it is encrypted with etc.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: Then `Wallet` can't extend `LockedWallet`. So we need it

PR #463, opened 2019-05-07
Refactor chain, node, wallet config

torkelrogstad:
Get rid of NetworkDb, DbConfig
Add proper structure to conf system,
moving everything under the bitcoin-s
root key.

IMO one goal here should be that everything is configurable through `.conf` files. We attempt this in this PR by having `application.conf` for tests (`application.conf` resolves over `reference.conf`), a `reference.conf` for base settings and `.conf` files for each module.

UPDATE: Just pushed a commit that removes Scalacheck from `node-test`. I had issues with tests hanging there, and this is something we've been wanting to do anyways. 

Todo: 
- [x] Markdown doc on config


db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -2,13 +2,133 @@ package org.bitcoins.db
...
+    val pathStr = config.getString("database.dbPath")
+    val path = Paths.get(pathStr)
+    logger.debug(s"DB path: $path")
+    path
+  }
+
+  private def createDbFileIfDNE(): Unit = {
+    //should add a check in here that we are using sqlite
+    if (!Files.exists(dbPath)) {
+      logger.info(s"Creating database directory=$dbPath")
```
torkelrogstad: this should be debug

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -2,13 +2,133 @@ package org.bitcoins.db
...
+
+    // loads reference.conf as well as application.conf,
+    // if the user has made one
+    val unresolvedConfig =
+      ConfigFactory
+        .load()
+        .withFallback(moduleConfig)
+        .withFallback(dbConfig)
+
+    logger.info(s"Unresolved bitcoin-s config:")
```
torkelrogstad: .debug

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -2,13 +2,133 @@ package org.bitcoins.db
...
+    logger.info(unresolvedConfig.getConfig("bitcoin-s").asReadableJson)
+
+    val config = unresolvedConfig
+      .resolve()
+      .getConfig("bitcoin-s")
+
+    logger.info(s"Resolved bitcoin-s config:")
```
torkelrogstad: .debug

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -378,6 +378,7 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
       mnemonicE.flatMap { mnemonic =>
         val encryptedT = EncryptedMnemonicHelper
           .encrypt(mnemonic, badPassphrase)
+        encryptedT.foreach(_ => logger.info(s"Encrypted mnemonic"))
```
torkelrogstad: remove

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -397,14 +398,19 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
         mnemonic <- mnemonicE
         encrypted <- encryptedMnemonicE
       } yield {
-        val wallet = WalletImpl(mnemonic, appConfig)
+        val wallet = WalletImpl(mnemonic)
+        logger.info(1.toString())
```
torkelrogstad: remove, leftover from testing

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -397,14 +398,19 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
...
         val coin =
-          HDCoin(DEFAULT_HD_PURPOSE, HDUtil.getCoinType(chainParams.network))
+          HDCoin(DEFAULT_HD_PURPOSE, HDUtil.getCoinType(config.network))
+        logger.info(2.toString())
```
torkelrogstad: remove

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -397,14 +398,19 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
...
         val account = HDAccount(coin, 0)
+        logger.info(3.toString())
```
torkelrogstad: remove

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -397,14 +398,19 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
...
         val xpriv = wallet.xprivForPurpose(DEFAULT_HD_PURPOSE)
+        logger.info(4.toString())
```
torkelrogstad: remove

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -397,14 +398,19 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
...
 
         // safe since we're deriving from a priv
         val xpub = xpriv.deriveChildPubKey(account).get
+        logger.info(5.toString())
```
torkelrogstad: remove

chain-test/src/test/scala/org/bitcoins/chain/validation/TipValidationTest.scala
```diff
@@ -29,7 +28,6 @@ class TipValidationTest extends ChainUnitTest {
   //blocks 566,092 and 566,093
   val newValidTip = BlockHeaderHelper.header1
   val currentTipDb = BlockHeaderHelper.header2Db
-  override val networkDb = NetworkDb.MainNetDbConfig
```
Christewart, line 37: This is why unit tests are failing in `TipValidationTest`, the network needs to be mainnet to be able to run the mainnet test vectors we have in `TipValidationTest`, is there a way to work around this with your configuration framework?

torkelrogstad: Ah thanks, that's doable

db-commons/src/main/resources/reference.conf
```diff
@@ -0,0 +1,4 @@
+bitcoin-s  {
```
Christewart, line 1: So looking at this, can a user override the `bitcoin-s.database` key in a config file they specify? I'm trying to understand how our setup would work with this section in the typeconfig library: 

I just want to make sure it is possible for the user to easily provide their own database config if they wish.

Christewart: I see you left a comment about this in your [configuration.md](https://github.com/bitcoin-s/bitcoin-s-core/pull/463/commits/6b8932493c643ff365fd61106435b4809eff71f3?short_path=ff00d23#diff-ff00d236def7fefcf44ba3d8dfca5971).

I disagree, there are useful database configurations that should be exposed to the end user. For instance whether you want to enable/disable connection pools. Maybe you want to switch to a postgres backend instead of sqlite. If you aren't using sqlite, you most definitely want to specify a username/password

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -2,13 +2,133 @@ package org.bitcoins.db
...
+    networkStr match {
+      case "mainnet"  => MainNetChainParams
+      case "testnet3" => TestNetChainParams
+      case "regtest"  => RegTestNetChainParams
+    }
+  }
+
+  lazy val network: NetworkParameters = chain.network
+
+  lazy protected val config: Config = {
```
Christewart, line 161: This is what the [typesafe config readme](https://github.com/lightbend/config#standard-behavior) says on this stuff

"Libraries and frameworks should default to ConfigFactory.load() if the application does not provide a custom Config object. This way, libraries will see configuration from application.conf and users can configure the whole app, with its libraries, in a single application.conf file."

So it looks to me like we are doing this, but I just wanted to double check

node/src/main/scala/org/bitcoins/node/constant/Constants.scala
```diff
@@ -43,7 +40,5 @@ sealed abstract class Constants {
     * we are currently connected to inside of the [[networkParameters]] function
     * @return
     */
-  def database: Database = dbConfig.database
+  def database: Database = ??? // dbConfig.database
```
Christewart: ping

wallet-test/src/test/scala/org/bitcoins/wallet/fixtures/DAOFixture.scala
```diff
@@ -21,13 +21,13 @@ private[fixtures] trait DAOFixture
     val tables = daoAccumulator.result()
 
     val dropTablesF =
-      Future.sequence(tables.map((dao: HasTable) =>
-        WalletDbManagement.dropTable(dao.table, dbConfig)))
+      Future.sequence(
+        tables.map((dao: HasTable) => WalletDbManagement.dropTable(dao.table)))
     Await.result(dropTablesF, timeout)
```
Christewart: Can we take out these two separate `Await.result()` and combine them into one `Await.result()`, slightly more efficient and less blocky

db-commons/src/main/resources/reference.conf
torkelrogstad: Everything in `application.conf` overrides our config, so this would typically be supplied by the user of the lib/app. We could also add support for a `bitcoin-s.conf` file later, that we special-case. They can also use the newly added method `AppConfig.withOverrides` that takes precedence over all found files (including `application.conf`). 

Wrt to overriding database conf: You have a point there. However, how is that going to work with Slick? Right now we use `SQLiteProfile` in all Slick imports, I'm assuming that wont work with for example postgres

Christewart: I guess that is the direction I think we should go. We don't have to do it in this PR, but I think we will want to do that in the nearish future. 

You are right that Scala source files need to be modified to support a different driver

torkelrogstad: How would we handle that, then? If the Scala sources need to be modified

PR #467, opened 2019-05-13
update readmes

Christewart:
This updates some READMEs on bitcoin-s and also fixes the version of the `sbt-api-mappings` plugin. Using `latest.release` as the version requires a network connection to resolve what the latest version is. This means we cannot develop offline

### DIFF
--- a/bitcoind-rpc/README.md
+++ b/bitcoind-rpc/README.md
@@ -1,2 +1,21 @@
 See the `bitcoind`/Bitcoin Core section on the 
 Bitcoin-S [website](https://bitcoin-s.org/docs/rpc/rpc-bitcoind).
+
+## Testing
+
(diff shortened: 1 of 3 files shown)


torkelrogstad:
WYM by the `latest.release` part? 

Also, I'm changing the READMEs pretty heavily in #465 . So perhaps it'd be better to make this PR against #465?

Christewart:
>WYM by the `latest.release` part?

To resolve this plugin requires a network call https://github.com/bitcoin-s/bitcoin-s-core/blob/master/project/plugins.sbt#L24 . If you are on a airplane you can't start sbt because you need to make a network call to try and resolve this plugin. It won't be able to detect that the plugin you have cached locally _is_ the latest release

torkelrogstad:
Ah, good point

torkelrogstad:
Ah, good point

PR #478, opened 2019-05-24
Add proper parsing of bitcoin.conf, rework auth credentials

torkelrogstad:
Somewhat recently a new config format was introduced in Bitcoin Core. Options can now be specified under section headers (`[regtest]`, `[main]`, `[test]`), as well as with network prefixes (`regtest.rpcport`,
`main.prune`) and the traditional format. In this PR we introduce a new type `BitcoindConfig` that parses this format. We also make the necessary changes in the bitcoind/Eclair RPC clients to accomodate this new type.

We also remove the `datadir` field from `BitcoindAuthCredentials`. This is not strictly necessary to connect to a bitcoind, so this is a bad separation of concerns. The `datadir` field is instead moved into the `BitcoindInstance`.

Finally we go over some places in tests and test utils and assert that we are operating in the user temporary directory. This is a safety measure so that other developers/users won't do the same mistake I did while working on this and accidentally blow up their `$HOME/.bitcoin` directory.

TODO: 
- [x] Make all tests pass (currently there's at least one failing test in `MempoolRpcTest`)
- [x] Implement `rpcauth` based auth


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -106,19 +114,46 @@ trait Client extends BitcoinSLogger {
       started.future
     }
 
-    val started = AsyncUtil.retryUntilSatisfiedF(() => isStartedF,
-                                                 duration = 1.seconds,
-                                                 maxTries = 60)
+    val started = {
+      for {
+        _ <- instance.authCredentials match {
```
Christewart: nit: Can we do matches like this outside of the for expression? It really clutters things up when we do this

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -106,19 +114,46 @@ trait Client extends BitcoinSLogger {
...
+          case cookie: CookieBased =>
+            val cookieExistsF =
+              AsyncUtil.retryUntilSatisfied(Files.exists(cookie.cookiePath))
+            cookieExistsF.onComplete {
+              case Failure(exception) =>
+                logger.error(s"Cookie filed was never created! $exception")
+              case _: Success[_] =>
+            }
+            cookieExistsF
+          case _: PasswordBased => Future.successful(())
```
Christewart: nit: `FutureUtil.unit` instead of `Future.successful(())`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -106,19 +114,46 @@ trait Client extends BitcoinSLogger {
...
     started.onComplete {
       case Success(_) => logger.debug(s"started bitcoind")
       case Failure(exc) =>
         if (instance.network != MainNet) {
           logger.info(
             s"Could not start bitcoind instance! Message: ${exc.getMessage}")
-          logger.info(s"Log lines:")
-          val logLines = Source.fromFile(logFile).getLines()
-          logLines.foreach(logger.info)
+          if (network != MainNet) {
```
Christewart: Isn't this a redundant check? `instance.network` is already not mainnet?

Christewart: Also it would be nice if there was a couple comments saying why we are dumping this stuff? This is basically creating of a copy of `debug.log` right? Why is this necessary?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindAuthCredentials.scala
```diff
@@ -1,45 +1,102 @@
...
+    */
+  case class PasswordBased(
       username: String,
-      password: String,
-      rpcPort: Int): BitcoindAuthCredentials = {
-    val defaultDataDir = new File(System.getProperty("user.home") + "/.bitcoin")
-    BitcoindAuthCredentials(username, password, rpcPort, defaultDataDir)
+      password: String
+  ) extends BitcoindAuthCredentials {
+    def config: BitcoindConfig = ???
```
Christewart: ping

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindAuthCredentials.scala
```diff
@@ -1,45 +1,102 @@
...
+    def cookie: String = {
+      if (Files.exists(cookiePath)) {
+        val cookieLines = Files.readAllLines(cookiePath).asScala
+        cookieLines.head
+      } else {
+        throw new RuntimeException(s"Could not find $cookiePath!")
+      }
+    }
+
+    def username: String = cookie.split(":").head
```
Christewart, line 84: so username should always be `__cookie__`?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    splittedPairs.collect(collect)
+  }
+
+  /**
+    * Applies the given partial function to all key/value pairs
+    * found in this config
+    */
+  private val collectAllLines = collectFrom(lines)(_)
```
Christewart: nit: Type signature on this would be nice

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
```diff
@@ -0,0 +1,323 @@
...
+  lazy val rpcUri: URI = new URI(s"$rpcbind:$rpcport")
+
+  /** Creates a new config with the given keys and values appended */
+  def withOption(key: String, value: String): BitcoindConfig = {
+    val ourLines = this.lines
+    new BitcoindConfig {
+
+      def lines: Seq[String] = {
+        val newLine = s"$key=$value"
+        Seq(newLine) ++ ourLines
```
Christewart: just a note, you don't need to wrap in `Seq(newLine)` you can do `newLine +: ourLines`.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
torkelrogstad: `FutureUtil` is not in master:-)

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
torkelrogstad: Left a comment in the code, LMK what you think

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindAuthCredentials.scala
torkelrogstad: Yes, assuming the cookie format doesn't change

PR #479, opened 2019-05-24
2019 05 15 spv sync headers

Christewart:
This PR implements the [first test case](https://github.com/bitcoin-s/bitcoin-s/compare/node...Christewart:2019-05-15-spv-sync-headers?expand=1#diff-89502ad43dd4d320a776487ddf2ccd7cR15) for syncing a `SpvNode`.

The test case connects a spv node to bitcoind. Then we generate a block with bitcoind, and verify that the block was relayed over the p2p network to our spv node and stored in our chain project.

This is a lot of auxiliary infrastructure in this PR like test fixtures.


node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
```diff
@@ -0,0 +1,60 @@
...
+class SpvNodeTest extends NodeUnitTest {
+
+  override type FixtureParam = SpvNodeConnectedWithBitcoind
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome =
+    withSpvNodeConnectedToBitcoind(test)
+
+  behavior of "SpvNode"
+
+  override implicit val system: ActorSystem = ActorSystem(
```
torkelrogstad: To avoid doing this you can change the actor system in `NodeUnitTest` to this
```scala
lazy implicit val system: ActorSystem = ActorSystem(s"${getClass.getSimpleName}-${System.currentTimeMillis}")
```

That way there's no need to write this boilerplate in every test

torkelrogstad: You could also avoid the `currentTimeMillis`, from my basic testing in the REPL it seems like it's ok to have multiple actor systems with the same name

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -0,0 +1,77 @@
...
+  private val client: Client =
+    Client(context = system, peer = peer, peerMessageReceiver = peerMsgRecv)
+
+  private val peerMsgSender: PeerMessageSender = {
+    PeerMessageSender(client, appConfig.network)
+  }
+
+  /** Starts our spv node */
+  def start(): Future[SpvNode] = {
+    peerMsgSender.connect()
```
torkelrogstad, line 38: I'm not really familiar with Akka actors. But this immediately seems a little fishy. Is there no way this can go wrong?

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    val isInitializedF = AsyncUtil.retryUntilSatisfied(isInit)
+
+    isInitializedF.map(_ => this)
+  }
+
+  /** Stops our spv node */
+  def stop(): Future[SpvNode] = {
+    peerMsgSender.disconnect()
```
torkelrogstad, line 57: Same as above

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /** Starts our spv node */
+  def start(): Future[SpvNode] = {
+    peerMsgSender.connect()
+
+    def isInit(): Boolean = {
+      peerMsgRecv.isInitialized
+    }
+
+    val isInitializedF = AsyncUtil.retryUntilSatisfied(isInit)
```
torkelrogstad: IMO it's cleaner to just do 
```scala
AsyncUtil.retryUntilSatisfied(peerMsgRecv.isInitialized)
```

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /** Stops our spv node */
+  def stop(): Future[SpvNode] = {
+    peerMsgSender.disconnect()
+
+    def isStopped(): Boolean = {
+      peerMsgRecv.isDisconnected
+    }
+
+    val isStoppedF = AsyncUtil.retryUntilSatisfied(isStopped)
```
torkelrogstad: Same as above

node/src/main/scala/org/bitcoins/node/messages/data/GetHeadersMessage.scala
```diff
@@ -52,12 +52,16 @@ object GetHeadersMessage extends Factory[GetHeadersMessage] {
   }
 
   /** Creates a [[GetHeadersMessage]] with no hash stop set, this requests all possible blocks
-    * if we need more than 500 block headers, we will have to send another [[GetHeadersMessage]]
+    * if we need more than 2000 block headers, we will have to send another [[GetHeadersMessage]]
     * [[https://bitcoin.org/en/developer-reference#getheaders]] */
```
torkelrogstad, line 56: Scaladoc nit: add `@see` before the link

node/src/main/scala/org/bitcoins/node/messages/data/HeadersMessage.scala
```diff
@@ -15,16 +15,18 @@ import scodec.bits.ByteVector
 object HeadersMessage extends Factory[HeadersMessage] {
   private case class HeadersMessageImpl(
       count: CompactSizeUInt,
-      headers: List[BlockHeader])
+      headers: Vector[BlockHeader])
```
torkelrogstad, line 18: We've discussed this before, but I still think its a good idea to not be as picky as to what data structure we expect. This change causes a ripple effect to other places in the code, and although it may be a good thing in this specific case it might not always be that way. Wouldn't it be better to take in a `Seq`/`Iterable` in places like this? We could continue being picky about what data structures we prefer at _call site_, without sacrificing API flexibility.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -16,23 +22,41 @@ import org.bitcoins.db.AppConfig
...
+      handleInventoryMsg(invMsg = invMsg, peerMsgSender = peerMsgSender)
   }
-}
 
-object DataMessageHandler {
+  private def handleInventoryMsg(
+      invMsg: InventoryMessage,
+      peerMsgSender: PeerMessageSender): Future[Unit] = {
+    logger.info(s"Received inv=${invMsg}")
+    /*    val invs: Seq[Future[Unit]] = invMsg.inventories.map { inv =>
```
torkelrogstad: delete?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -222,14 +227,18 @@ object PeerMessageReceiver {
   case class NetworkMessageReceived(msg: NetworkMessage, client: Client)
       extends PeerMessageReceiverMsg
 
-  def apply(state: PeerMessageReceiverState, appConfig: AppConfig)(
+  def apply(
+      state: PeerMessageReceiverState,
+      nodeAppConfig: NodeAppConfig,
```
torkelrogstad, line 232: What's your thoughts on making the configs implicit, as well as the actor ref factory?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -222,14 +227,18 @@ object PeerMessageReceiver {
...
+      chainAppConfig: ChainAppConfig)(
       implicit ref: ActorRefFactory): PeerMessageReceiver = {
-    new PeerMessageReceiver(state, appConfig)(ref)
+    new PeerMessageReceiver(state, nodeAppConfig, chainAppConfig)(ref)
   }
 
-  def newReceiver(appConfig: AppConfig)(
+  def newReceiver(nodeAppConfig: NodeAppConfig, chainAppConfig: ChainAppConfig)(
```
torkelrogstad, line 238: Same as above

node/src/main/scala/org/bitcoins/node/SpvNode.scala
Christewart: akka implies the "send and forget" message passing style. This means that we do not know what happens after we call `client.actor ! Tcp.Connect(socket)` 

You need to look at [PeerMessageReceiver.isConnected](https://github.com/bitcoin-s/bitcoin-s/blob/node/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala#L94) to see if the connection was successful.

Christewart: Although looking at this closer, it does look like we just looking to see if the [underlying promise](https://github.com/bitcoin-s/bitcoin-s/blob/node/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiverState.scala#L38) is completed, not what it is actually completed with. This probably should be changed in the future

Christewart: Issue filed here to address this in the future https://github.com/bitcoin-s/bitcoin-s/issues/482

PR #481, opened 2019-05-28
BIP 158 Golomb-Rice Coded Sets and block filters

nkohen:
Implements BIP 158 Golomb-Rice Coded Sets and their corresponding block filters


core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,281 @@
...
+  behavior of "GCS"
+
+  //https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding
+  it must "encode and decode Golomb Coded Set example 1" in {
+    val p = UInt8(2)
+    val original = UInt64.zero
+
+    val encoding = GCS.golombEncode(item = original, p = p)
+
+    assert(encoding == BitVector.bits(Vector(false, false, false)))
```
torkelrogstad: this is down to preference, but Scodec has nice syntax for this:

```scala
import scodec.bits.BinStringSyntax
val bits = bin"000"
val otherBits = BitVector.bits(Vector(false, false, false))
bits == otherBits // true
```

torkelrogstad: this is down to preference, but Scodec has nice syntax for this:

```scala
import scodec.bits.BinStringSyntax
val bits = bin"000"
val otherBits = BitVector.bits(Vector(false, false, false))
bits == otherBits // true
```

torkelrogstad: this is down to preference, but Scodec has nice syntax for this:

```scala
import scodec.bits.BinStringSyntax
val bits = bin"000"
val otherBits = BitVector.bits(Vector(false, false, false))
bits == otherBits // true
```

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  it must "encode and decode arbitrary sets of elements for arbitrary p" in {
+
+    /**
+      * Bit parameter for GCS, cannot be more than 32 as we will
+      * have a number too large for a uint64.
+      * [[https://github.com/Roasbeef/btcutil/blob/b5d74480bb5b02a15a9266cbeae37ecf9dd6ffca/gcs/gcs.go#L67]]
+      */
+    def genP: Gen[UInt8] = {
```
torkelrogstad: Move this into `NumberGenerator`? Seeing as it's used multiple times just in this test

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  it must "encode and decode arbitrary ByteVectors for arbitrary p" in {
+
+    /**
+      * Bit parameter for GCS, cannot be more than 32 as we will
+      * have a number too large for a uint64.
+      * [[https://github.com/Roasbeef/btcutil/blob/b5d74480bb5b02a15a9266cbeae37ecf9dd6ffca/gcs/gcs.go#L67]]
+      */
+    def genP: Gen[UInt8] = {
```
torkelrogstad, line 220: same as above

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+import org.bouncycastle.crypto.macs.SipHash
+import org.bouncycastle.crypto.params.KeyParameter
+import scodec.bits.{BitVector, ByteVector}
+
+import scala.annotation.tailrec
+
+// TODO: Replace ByteVector with a type for keys
+/**
+  * Defines all functionality dealing with Golomb-Coded Sets as specified in
+  * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#GolombCoded_Sets]]
```
torkelrogstad: I think a better Scaladoc would be 

```scala
/**
  * Defines all functionality dealing with Golomb-Coded Sets 
  * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#GolombCoded_Sets BIP158, GCS]]
  */

File: core/src/main/scala/org/bitcoins/core/gcs/GCS.scala [hunk 5]
```diff
@@ -0,0 +1,254 @@
...
+  */
+object GCS {
+
+  /**
+    * Given parameters and data, golomb-encodes the data as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#set-construction]]
```
torkelrogstad: Same as above, when using the `@see` notation we get nice links in the Scaladoc. It's nice to follow convention:-)

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+    * Given data, constructs a GolombFilter for that data using parameters specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters]]
+    * @param data
+    * @param key
+    * @return
+    */
+  def buildBasicBlockFilter(
+      data: Vector[ByteVector],
+      key: ByteVector): GolombFilter = {
+    buildGolombFilter(data, key, p = UInt8(19), m = UInt64(784931))
```
torkelrogstad: Could we move the magic numbers into constants?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  private def sipHash(item: ByteVector, key: ByteVector): UInt64 = {
+    val sh = new SipHash(2, 4)
```
torkelrogstad: Same here, where's the magic numbers 2 and 4 coming from?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+    val digest = new Array[Byte](8)
+
+    sh.doFinal(digest, 0)
+
+    UInt64.fromBytes(ByteVector(digest))
+  }
+
+  /**
+    * Hashes the item to the range [0, f) as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#hashing-data-objects]]
```
torkelrogstad: Same Scaladoc comment as above

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+    val hash = sipHash(item, key)
+
+    val bigInt = (hash.toBigInt * f.toBigInt) >> 64
+
+    UInt64(bigInt)
+  }
+
+  /**
+    * Hashes the items of a set of items as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#hashing-data-objects]]
```
torkelrogstad: Same Scaladoc comment as abvoe

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+      val setValue = hashToRange(item, f, key)
+      hashedItemsBuilder.+=(setValue)
+    }
+
+    hashedItemsBuilder.result()
+  }
+
+  /**
+    * Converts num to unary (6 = 1111110) as required by
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
```
torkelrogstad: Same Scaladoc comment as above

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+      val leftPadded = BitVector(binUnary.toByteArray)
+      val noPadding = dropLeftPadding(leftPadded)
+
+      noPadding.:+(false)
+    }
+  }
+
+  /**
+    * Encodes a hash into a unary prefix and binary suffix as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
```
torkelrogstad: Scala Scaladoc comment

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+    val prefix = toUnary(q)
+
+    val pBits = item.bytes.toBitVector.takeRight(p.toInt)
+
+    prefix ++ pBits
+  }
+
+  /**
+    * Decodes an item off of the front of a BitVector by reversing [[GCS.golombEncode]] as in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
```
torkelrogstad: Scaladoc

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,254 @@
...
+        loop(encodedLeft, decoded.:+(hash), hash)
+      }
+    }
+
+    loop(encodedData, Vector.empty)
+  }
+
+  /**
+    * Given a set of ascending hashes, golomb-encodes them as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#set-construction]]
```
torkelrogstad: Scaladoc

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+  TransactionOutPoint,
+  TransactionOutput
+}
+import org.bitcoins.core.script.control.OP_RETURN
+import scodec.bits.{BitVector, ByteVector}
+
+import scala.annotation.tailrec
+
+// TODO: Replace ByteVector with a type for keys
+case class GolombFilter(
```
torkelrogstad, line 26: A short Scaladocs with a link or two to other resources would be great

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,93 @@
...
+    binarySearch(from = 0, to = n - 1, hash, decodedHashes)
+  }
+}
+
+object BlockFilter {
+
+  /**
+    * Given a Block and access to the UTXO set, constructs a Block Filter
+    * for that block as specified in
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters]]
```
torkelrogstad: Scaladoc

core/src/main/scala/org/bitcoins/core/gcs/TempUtxoProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+  */
+abstract class TempUtxoProvider {
+
+  def getUtxo(
+      transactionOutPoint: TransactionOutPoint): Option[TransactionOutput]
+}
+
+class TempUtxoProviderImpl extends TempUtxoProvider {
+
+  val utxos: mutable.Map[TransactionOutPoint, TransactionOutput] =
```
torkelrogstad, line 21: Maybe I'm missing something here, but it doesn't appear like  we're ever adding to these utxos?

nkohen: The current plan is to use this in tests which will construct this map by adding things, but no, we don't use this anywhere yet.

core-test/src/test/scala/org/bitcoins/core/util/BitcoinScriptUtilTest.scala
```diff
@@ -32,6 +33,38 @@ class BitcoinScriptUtilTest extends FlatSpec with MustMatchers {
       .isEmpty must be(true)
   }
 
+  it must "filter out all but the data tokens" in {
+    val (p2pkhScript, _) = ScriptGenerators.p2pkhScriptPubKey.sample.get
```
Christewart: Please use property based testing correctly by segregating these test cases and using a `forAll()`

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -66,6 +66,23 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
           || op == OP_PUSHDATA4)
   }
 
+  def getDataTokens(asm: Seq[ScriptToken]): Seq[ScriptToken] = {
+    var prevWasPush: Boolean = false
+
+    asm.filter { token =>
```
Christewart: This looks really weird, why not just
```scala
val accum = Vector.newBuilder
asm.zipWithIndex.foreach { case (token,index) => 
  token match { 
    case OP_PUSHDATA1 | OP_PUSHDATA2 | OP_PUSHDATA4 => 
      //op push data 'n' says the _next_  value is n bytes which indicates how big the constant is to push onto the stack. So if n=257 we need to use OP_PUSHDATA2 (can't represent 257 with 1 byte) and it means take the next 257 bytes _after_ the 2 bytes that indicate the constant is of size 257
      // so it is is not the _next_ value, it is the value after that
      builder.+=(accum(index+2))
     case b: BytesToPushOntoStack => 
       builder.+=(accum(index+1))
     case _ => 

        //skip, make sure we get exhaustive pattern match

build.result()
```

core-test/src/test/scala/org/bitcoins/core/gcs/BlockFilterTest.scala
```diff
@@ -0,0 +1,87 @@
...
+      // TODO header: BlockFilterHeader,
+      notes: String
+  ) {
+
+    def runTest(): org.scalatest.Assertion = {
+      val constructedFilter = BlockFilter(block, prevOutputScripts)
+
+      val script = block.transactions.head.outputs.head.scriptPubKey
+
+      println(
```
Christewart: Use ` extends BitcoinSLogger` on the case class intead of println

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,226 @@
...
+import scodec.bits.{BitVector, ByteVector}
+
+import scala.annotation.tailrec
+
+// TODO: Replace ByteVector with a type for keys
+/**
+  * Defines all functionality dealing with Golomb-Coded Sets
+  * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#GolombCoded_Sets]]
+  */
+object GCS {
```
AlexITC, line 16: Just in case, I did a similar implementation in Scala:
- https://github.com/X9Developers/block-explorer/blob/develop/server/app/com/xsn/explorer/gcs/GolombEncoding.scala
- https://github.com/X9Developers/block-explorer/blob/develop/server/test/com/xsn/explorer/gcs/GolombEncodingSpec.scala

nkohen: Hey Alex! Sweet :) Any chance you've written the code to serialize script outputs (i.e. generate block filters from blocks)? I think I have all my GCS stuff in good shape but my serialization isn't passing tests :(

AlexITC: No, I may end up working on that someday.

Is good to know I'm not the only one building those kind of pieces in Scala.

nkohen: Ah I've found it: https://github.com/X9Developers/block-explorer/commit/37cb7ceb81a9ce8d2a098b3b319b5fca616c2d2b

Any chance you've tested this logic against the bip 158 test vectors?
https://github.com/bitcoin/bips/blob/master/bip-0158/testnet-19.json

AlexITC: That's not the serialization you are likely talking about, it basically collects all the addresses involved in a block and produces the filter, it does involve only the addresses from the `addresses` field from transaction inputs/outputs, I understand that serializing the whole script requires other details.

nkohen: Gotchya, well I just finally got serialization working today! So should you ever do it feel free to see how we did it :)

AlexITC: Will do, thanks!

core-test/src/test/scala/org/bitcoins/core/gcs/BlockFilterTest.scala
```diff
@@ -0,0 +1,77 @@
...
+class BlockFilterTest extends BitcoinSUnitTest {
+  behavior of "BlockFilter"
+
+  // https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#appendix-c-test-vectors
+  case class Bip158TestCase(
+      blockHeight: Int,
+      blockHash: DoubleSha256Digest,
+      block: Block,
+      prevOutputScripts: Vector[ScriptPubKey],
+      // TODO prevHeader: BlockFilterHeader,
```
torkelrogstad, line 20: Are these TODOs something for a later PR? Would be nice to get an issue for it, in that case

core-test/src/test/scala/org/bitcoins/core/gcs/BlockFilterTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      filter: GolombFilter,
+      // TODO header: BlockFilterHeader,
```
torkelrogstad, line 22: Same as above

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,258 @@
...
+  it must "encode and decode a set of elements already tested" in {
+    val p = UInt8(2)
+
+    // Diffs are 1, 2, 3, 4, 5
+    val sortedItems =
+      Vector(UInt64(0), UInt64(1), UInt64(3), UInt64(6), UInt64(10), UInt64(15))
+
+    val codedSet = GCS.encodeSortedSet(sortedItems, p)
+
+    val coded0 = Vector(false, false, false)
```
torkelrogstad: Bin string syntax? Makes it a lot clearer what's going on IMO

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,258 @@
...
+
+    def items: Gen[(Vector[UInt64], UInt8)] = {
+      NumberGenerator.genP.flatMap { p =>
+        Gen.choose(1, 1000).flatMap { size =>
+          // If hash's quotient when divided by 2^p is too large, we hang converting to unary
+          val upperBound: Long = 1L << (p.toInt * 1.75).toInt
+
+          val hashGen = Gen
+            .chooseNum(0L, upperBound)
+            .map(num => UInt64(BigInt(num)))
```
torkelrogstad: I think you can remove the `BigInt` call here?

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,258 @@
...
+
+          Gen.listOfN(size, hashGen).map(_.toVector).map { vec =>
```
torkelrogstad, line 200: Is there really any point in doing `.toVector` here?

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,258 @@
...
+      Gen.choose(8, 32).map(UInt8(_))
+    }
+
+    def genPM: Gen[(UInt8, UInt64)] = genP.flatMap { p =>
+      // If hash's quotient when divided by 2^p is too large, we hang converting to unary
+      val upperBound: Long = p.toInt * 1000
+
+      val mGen = Gen
+        .chooseNum(0L, upperBound)
+        .map(num => UInt64(BigInt(num)))
```
torkelrogstad: Think you can remove the `BigInt` part here

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
```diff
@@ -0,0 +1,258 @@
...
+
+      mGen.map(m => (p, m))
+    }
+
+    def genItems: Gen[Vector[ByteVector]] = {
+      Gen.choose(1, 1000).flatMap { size =>
+        Gen.listOfN(size, NumberGenerator.bytevector).map(_.toVector)
```
torkelrogstad, line 238: Same comment as above wrt to `.toVector`. 

I don't feel strongly at all about that, so feel free to disregard

core-test/src/test/scala/org/bitcoins/core/gcs/GolombFilterTest.scala
```diff
@@ -0,0 +1,75 @@
...
+
+        assert(!filter.matchesHash(rand))
+        assert(filter.matchesHash(data1))
+        assert(filter.matchesHash(data2))
+    }
+  }
+
+  it must "match arbitrary encoded data for bip 158 GCS parameters" in {
+    val genKey: Gen[ByteVector] =
+      Gen.listOfN(16, NumberGenerator.byte).map(ByteVector(_))
```
torkelrogstad, line 52: `NumberGenerator.byteVector`?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,224 @@
...
+
+import org.bitcoins.core.number.{UInt64, UInt8}
+import org.bitcoins.core.protocol.CompactSizeUInt
+import org.bouncycastle.crypto.macs.SipHash
+import org.bouncycastle.crypto.params.KeyParameter
+import scodec.bits.{BitVector, ByteVector}
+
+import scala.annotation.tailrec
+
+// TODO: Replace ByteVector with a type for keys
```
torkelrogstad, line 11: File an issue for this?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,224 @@
...
+    val sipHashCParam = 2
+    val sipHashDParam = 4
+
+    val sh = new SipHash(sipHashCParam, sipHashDParam)
+
+    val keyParam = new KeyParameter(key.toArray)
+
+    sh.init(keyParam)
+
+    sh.update(item.toArray, 0, item.length.toInt)
```
torkelrogstad: Could we give a name to the `0` value here?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  /**
+    * Converts num to unary (6 = 1111110)
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
+    *
+    * TODO: protect against large inputs which cause OutOfMemoryErrors and even larger ones which fail on toInt
+    */
+  def toUnary(num: UInt64): BitVector = {
+    if (num == UInt64.zero) {
+      BitVector.bits(Vector(false))
```
torkelrogstad: `bin"0"`?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,224 @@
...
+  @tailrec
+  private def dropLeftPadding(padded: BitVector): BitVector = {
+    if (padded.isEmpty || padded.head) {
+      padded
+    } else {
+      dropLeftPadding(padded.tail)
+    }
+  }
+
+  private def golombDecodeItemFromSet(
```
torkelrogstad, line 180: Looks like the `BitVector` returned is surplus bits. Could we add a one line scaladoc?

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,147 @@
...
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters]]
+    */
+  def apply(block: Block, utxoProvider: TempUtxoProvider): GolombFilter = {
+    val key = block.blockHeader.hash.bytes.take(16)
+
+    val transactions: Vector[Transaction] = block.transactions.toVector
+
+    val noCoinbase: Vector[Transaction] = transactions.tail
+    val newOutputs: Vector[TransactionOutput] = transactions.flatMap(_.outputs)
+    val newScriptPubKeys: Vector[ByteVector] = newOutputs.flatMap { output =>
```
torkelrogstad: This line took me a little while to figure out. But you're filtering here, right? If I understand this correctly, `None` and `Some` have implicit conversions available to `Vector`, which is why this work. What's your thoughts on using `.collect` instead? 

Also, I don't really see how the first if branch can be triggered. If we're dealing with an ampty SPK, it cannot contain OP_RETURN?

torkelrogstad: Looking at your code in the method below, I'm suspecting that you meant to use `||` instead of `&&`

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -66,6 +66,26 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
           || op == OP_PUSHDATA4)
   }
 
+  def getDataTokens(asm: Seq[ScriptToken]): Seq[ScriptToken] = {
```
torkelrogstad, line 70: Scaladoc?

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
nkohen: It has to happen at some point due to the type signatures of `GCS` functions and so might as well happen asap to get performance benifits

core-test/src/test/scala/org/bitcoins/core/gcs/GCSTest.scala
nkohen: same reason as above

core-test/src/test/scala/org/bitcoins/core/gcs/GolombFilterTest.scala
nkohen: We need a 16 byte `ByteVector` which is not what that generates

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  /**
+    * Hashes the item to the range [0, f)
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#hashing-data-objects]]
+    */
+  def hashToRange(item: ByteVector, f: UInt64, key: ByteVector): UInt64 = {
+    val hash = sipHash(item, key)
+
+    val bigInt = (hash.toBigInt * f.toBigInt) >> 64
```
Christewart, line 82: @nkohen Is this shift right going to be a problem with something that has the upper most sign bit set? I think this is going to be a bug later on, do we have any test vectors where the data element starts with 

`0x8....` (sign bit set!)

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    hashedItemsBuilder.result()
+  }
+
+  /**
+    * Converts num to unary (6 = 1111110)
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
+    *
+    * TODO: protect against large inputs which cause OutOfMemoryErrors and even larger ones which fail on toInt
```
Christewart, line 112: Has this been addressed? If not at minimum we should file an issue

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,227 @@
...
+      noPadding.:+(false)
+    }
+  }
+
+  /**
+    * Encodes a hash into a unary prefix and binary suffix
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-rice-coding]]
+    */
+  def golombEncode(item: UInt64, p: UInt8): BitVector = {
+    val q = item >> p.toInt
```
Christewart, line 134: Again i have this concern with the shift right

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+
+  /** Returns the first hash gcs-encoded at the front of a BitVector, as well as the remaining BitVector */
+  private def golombDecodeItemFromSet(
+      encodedData: BitVector,
+      p: UInt8): (UInt64, BitVector) = {
+    val head = golombDecode(encodedData, p)
+
+    val prefixSize = (head >> p.toInt).toInt + 1
```
Christewart, line 185: Sign bit concern

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -0,0 +1,227 @@
...
+  /**
+    * Decodes all hashes from golomb-encoded data, reversing [[GCS.encodeSortedSet]]
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#set-queryingdecompression]]
+    */
+  def golombDecodeSet(encodedData: BitVector, p: UInt8): Vector[UInt64] = {
+    @tailrec
+    def loop(
+        encoded: BitVector,
+        decoded: Vector[UInt64],
+        lastHash: UInt64 = UInt64.zero): Vector[UInt64] = {
```
Christewart: This default param seems like an easy foot gun, it doesn't seem to be that useful?

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,168 @@
...
+  */
+case class GolombFilter(
+    key: ByteVector,
+    m: UInt64,
+    p: UInt8,
+    n: CompactSizeUInt,
+    encodedData: BitVector) {
+  lazy val decodedHashes: Vector[UInt64] = GCS.golombDecodeSet(encodedData, p)
+
+  // TODO: Offer alternative that stops decoding when it finds out if data is there
```
Christewart, line 34: ping

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,168 @@
...
+    val f = n.num * m
+    val hash = GCS.hashToRange(data, f, key)
+
+    matchesHash(hash)
+  }
+}
+
+object BlockFilter {
+
+  val M: UInt64 = UInt64(784931)
```
Christewart, line 72: scaladoc for what these magic values are

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -0,0 +1,168 @@
...
+    prevOutputs
+      .filterNot(_.scriptPubKey == EmptyScriptPubKey)
+      .map(_.scriptPubKey)
+  }
+
+  /**
+    * Given a Block and access to the UTXO set, constructs a Block Filter for that block
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters]]
+    */
+  def apply(block: Block, utxoProvider: TempUtxoProvider): GolombFilter = {
```
Christewart, line 117: There appears to be redundancy between the apply methods? Can these be consolidated to reduce LOC?

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
AlexITC: Guess what? now I need to implement the filter for the script, I'll try to reuse this work.

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
nkohen: Already have an issue :) #523

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
nkohen: #528

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
nkohen: So `0x8000` is `32768` and `0x800` is `2048` and here is my repl session:
```
scala> BigInt(32768)
res0: scala.math.BigInt = 32768

scala> res0 >> 4
res1: scala.math.BigInt = 2048
```

The reason that this works is because the `>>` defined on `BigInt` respects the arbitrary-precision abstraction and doesn't take into account something like most significant byte or anything like that.

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
nkohen: `UInt64`'s `>>` is `BigInt`'s

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
nkohen: `UInt64`'s `>>` is `BigInt`'s

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
nkohen: Good luck :) feel free to reach out with any questions! (Best way to reach us is probably through the Suredbits slack https://join.slack.com/t/suredbits/shared_invite/enQtNDEyMjY3MTg1MTg3LTYyYjkwOGUzMDQ4NDAwZjE1M2I3MmQyNWNlZjNlYjg4OGRjYTRjNWUwNjRjNjg4Y2NjZjAxYjU1N2JjMTU1YWM where there's a bitcoin-s channel)

PR #484, opened 2019-05-30
Initial SPV node and wallet integration

torkelrogstad:
In this PR we:

- Add a table of interesting public keys to our SPV node, and a method for external users of the node to add keys to that table. The contents of the table is used to construct a bloom filter we send to our peer. In the future we want to expand the range of "interesting things" you can add. Transactions come to mind, probably also scripts. 
- Add a set of callbacks to our SPV node that get executed on certain events. First of is when receiving blocks, merkle blocks and transactions. 
- Add a new method to our wallet API, `getAddressInfo`. It's similar to the `getaddressinfo` Bitcoin Core RPC call, although less ambitious in scope. It is used to get the public key we insert into our bloom filter. 

We also add a WIP test for integrating our SPV node and wallet.

~~We are not receiving `tx` messages after receiving our merkle block, which we can expect to do, according to the Bitcoin.org dev guide:~~

> ~~The `merkleblock` message is a reply to a `getdata` message which requested a block using the inventory type `MSG_MERKLEBLOCK`. It is only part of the reply: if any matching transactions are found, they will be sent separately as `tx` messages.~~

~~I have a memory of seeing `tx` messages in `debug.log` that we were not picking up on our node, but that might be me misremembering. So will have to do some digging there, on why we're not getting the messages.~~

## Update June 3.: 

We also implement legacy address derivation in this PR, as well as legacy UTXO storing and spening. This had to be added to this PR because we do not have Segwit P2P messages implemented in `node`.

## Further update: 

In an effort to further debug this, I add an option to dump all raw P2P messages to file, at `$HOME/.bitcoin-s/NETWORK/p2p.dump`

I also add a file `node-test/src/main/scala/org/bitcoins/node/Bloom.scala` that's a runnable Main method, which has some functionality for spinning up a node and add pubkeys/load bloom filters. This is intended to be left out before merging, leaving it here so you guys can see what I've tried.

## Latest update June 3 before signing off: 

We're now receiving `tx` messages. This PR is balooning out to be really big though, so I'm not sure how much more it makes sense to add to this. 

The current functionality added in this PR:

- Make default account type in wallet configurable
- Create all three account kinds on wallet creation
- Add derivation and spending of legacy addresses in wallet
- Add `getAddressInfo` to wallet API, giving us information about what oublic keys belong to an address (this should be expanded in the future, similar to how the `getaddressinfo` RPC call for Bitcoin Core works)
- Add functionality in SPV node to mark pubkeys as interesting, this causes bloom filters to be generated and sent to our peer
- Add callbacks in SPV node that get triggered on receiving transactions, blocks and merkle blocks (TODO: these callbacks should probably only be triggered if verification of sent data passes. I.e. don't call TX callback if we're sent an invalid transaction)
- Add configuration for dumping all raw P2P messages to file, for debugging purposes

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
@@ -99,6 +99,17 @@ object Sha256DigestBE extends Factory[Sha256DigestBE] {
   * Represents the result of SHA256(SHA256())
   */
 sealed abstract class DoubleSha256Digest extends HashDigest {
+
+  /**
+    * @note The string representation of this is given in
+    *       big endian style. This is in order to make it
+    *       easier to scan logs for mentions of the hash.
+    *       (Bitcoin Core exposes all human facing hashes
+    *       in big endian style).
+    */
+  override def toString(): String =
+    s"DoubleSha256Digest(${flip.hex})"
+
   def flip: DoubleSha256DigestBE = DoubleSha256DigestBE(bytes.reverse)
 }
 
@@ -108,7 +119,6 @@ object DoubleSha256Digest extends Factory[DoubleSha256Digest] {
     require(bytes.length == 32,
             // $COVERAGE-OFF$
             "DoubleSha256Digest must always be 32 bytes, got: " + bytes.length)
-    override def toString = s"DoubleSha256DigestImpl($hex)"
     // $COVERAGE-ON$
   }
   override def fromBytes(bytes: ByteVector): DoubleSha256Digest =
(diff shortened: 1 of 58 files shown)


core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
```diff
@@ -99,6 +99,17 @@ object Sha256DigestBE extends Factory[Sha256DigestBE] {
...
 sealed abstract class DoubleSha256Digest extends HashDigest {
+
+  /**
+    * @note The string representation of this is given in
+    *       big endian style. This is in order to make it
+    *       easier to scan logs for mentions of the hash.
+    *       (Bitcoin Core exposes all human facing hashes
+    *       in big endian style).
+    */
+  override def toString(): String =
```
torkelrogstad, line 110: This is something you guys should look at. Is this a good idea? IMO yes (kinda obvious, otherwise I wouldn't have implemented it), but I'm open to other opinions

Christewart:
@torkelrogstad can you turn down logging and run CI again? It seems Ci is failing because too many logs

torkelrogstad:
@Christewart Tuned down logging and pushed a couple more commits

torkelrogstad:
Closing in favor of #492

PR #490, opened 2019-06-03
Node

Christewart:
This PR is going to merge the node branch into master. The node branch is a _heavy_ WIP, but now that we have tagged `v0.1` we want to move it into master to avoid rebase hell.

This PR introduces 4 new subprojects 

- `db-commons` - basic database configuration shared across projects
- `chain` - managing the blockchain chain state
- `wallet` - creating a native wallet for bitcoin-s 
- `node` - creating infrastructure for a spv node with bitcoin-s

All of these projects are in the alpha stages, and should not be used unless you know what you are doing.


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
```diff
@@ -321,4 +340,22 @@ object BitcoindConfig {
...
+    * exist. Returns the written file.
+    */
+  def writeConfigToFile(config: BitcoindConfig, datadir: File): Path = {
+
+    val confStr = config.lines.mkString("\n")
+
+    Files.createDirectories(datadir.toPath)
+    val confFile = datadir.toPath.resolve("bitcoin.conf")
+
+    if (!Files.exists(confFile)) {
```
torkelrogstad: I think this screws with `withOption`. Updating the config doesn't result in a updated file, which doesn't seem to be what we want

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
```diff
@@ -23,8 +24,25 @@ import java.nio.file.Path
   *
   * @see https://github.com/bitcoin/bitcoin/blob/master/doc/bitcoin-conf.md
   */
-abstract class BitcoindConfig extends BitcoinSLogger {
-  private[bitcoins] def lines: Seq[String]
+case class BitcoindConfig(lines: Seq[String], datadir: File)
```
torkelrogstad: Might make sense to make `lines` `private[bitcoins]`, as that's not really a field it makes sense to expose to end users

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
```diff
@@ -224,13 +242,14 @@ abstract class BitcoindConfig extends BitcoinSLogger {
...
-        val newLine = s"$key=$value"
-        newLine +: ourLines
-      }
-    }
+    val newLine = s"$key=$value"
+    val lines = newLine +: ourLines
+    val newConfig = BitcoindConfig(lines, datadir)
+    logger.debug(
+      s"Appending new config with $key=$value to datadir=${datadir.getAbsolutePath}")
+    BitcoindConfig.writeConfigToFile(newConfig, datadir)
```
torkelrogstad, line 252: Not sure how I feel about a seemingly innocent function performing a IO side effect, without that being explicitly communicated. Same thing applies to the places above we're also calling `BitcoindConfig.writeConfigToFile`. Would be bad if you're playing around with the config class, and without knowing it you're overwriting your `$HOME/.bitcoin` directory

Christewart: So I agree that is a downside, but the other downside is having divergence between what is on disk and what is represented in memory. This is a prepend only operation, so it does preserve the contents of the file before this operation is called. 

I guess this is more theoretical, but do we want to have the ability to diverge from what is on disk and what is actually contained in this case class? I would tend to say no.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
Christewart: I guess the alternative here is risking that we overrwrite a bitcoin.conf with important information in it. I'm not sure what the best way to balance this is? We could check if it is the `DEFAULT_DATADIR`, but this still leaves the possibility of users losing money because their rpc credentials were in a non traditional datadir location.

PR #492, opened 2019-06-04
 Initial SPV node and wallet integration

torkelrogstad:
In this PR we:

- Add a table of interesting public keys to our SPV node, and a method for external users of the node to add keys to that table. The contents of the table is used to construct a bloom filter we send to our peer. In the future we want to expand the range of "interesting things" you can add. Transactions come to mind, probably also scripts. 
- Add a set of callbacks to our SPV node that get executed on certain events. First of is when receiving blocks, merkle blocks and transactions. 
- Add a new method to our wallet API, `getAddressInfo`. It's similar to the `getaddressinfo` Bitcoin Core RPC call, although less ambitious in scope. It is used to get the public key we insert into our bloom filter. 

We also add a WIP test for integrating our SPV node and wallet.

~~We are not receiving `tx` messages after receiving our merkle block, which we can expect to do, according to the Bitcoin.org dev guide:~~

> ~~The `merkleblock` message is a reply to a `getdata` message which requested a block using the inventory type `MSG_MERKLEBLOCK`. It is only part of the reply: if any matching transactions are found, they will be sent separately as `tx` messages.~~

~~I have a memory of seeing `tx` messages in `debug.log` that we were not picking up on our node, but that might be me misremembering. So will have to do some digging there, on why we're not getting the messages.~~

## Update June 3.: 

We also implement legacy address derivation in this PR, as well as legacy UTXO storing and spening. This had to be added to this PR because we do not have Segwit P2P messages implemented in `node`.

## Further update: 

In an effort to further debug this, I add an option to dump all raw P2P messages to file, at `$HOME/.bitcoin-s/NETWORK/p2p.dump`

I also add a file `node-test/src/main/scala/org/bitcoins/node/Bloom.scala` that's a runnable Main method, which has some functionality for spinning up a node and add pubkeys/load bloom filters. This is intended to be left out before merging, leaving it here so you guys can see what I've tried.

## Latest update June 3 before signing off: 

We're now receiving `tx` messages. This PR is balooning out to be really big though, so I'm not sure how much more it makes sense to add to this. 

The current functionality added in this PR:

- Make default account type in wallet configurable
- Create all three account kinds on wallet creation
- Add derivation and spending of legacy addresses in wallet
- Add `getAddressInfo` to wallet API, giving us information about what oublic keys belong to an address (this should be expanded in the future, similar to how the `getaddressinfo` RPC call for Bitcoin Core works)
- Add functionality in SPV node to mark pubkeys as interesting, this causes bloom filters to be generated and sent to our peer
- Add callbacks in SPV node that get triggered on receiving transactions, blocks and merkle blocks (TODO: these callbacks should probably only be triggered if verification of sent data passes. I.e. don't call TX callback if we're sent an invalid transaction)
- Add configuration for dumping all raw P2P messages to file, for debugging purposes

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
@@ -99,6 +99,17 @@ object Sha256DigestBE extends Factory[Sha256DigestBE] {
   * Represents the result of SHA256(SHA256())
   */
 sealed abstract class DoubleSha256Digest extends HashDigest {
+
+  /**
+    * @note The string representation of this is given in
+    *       big endian style. This is in order to make it
+    *       easier to scan logs for mentions of the hash.
+    *       (Bitcoin Core exposes all human facing hashes
+    *       in big endian style).
+    */
+  override def toString(): String =
+    s"DoubleSha256Digest(${flip.hex})"
+
   def flip: DoubleSha256DigestBE = DoubleSha256DigestBE(bytes.reverse)
 }
 
@@ -108,7 +119,6 @@ object DoubleSha256Digest extends Factory[DoubleSha256Digest] {
     require(bytes.length == 32,
             // $COVERAGE-OFF$
             "DoubleSha256Digest must always be 32 bytes, got: " + bytes.length)
-    override def toString = s"DoubleSha256DigestImpl($hex)"
     // $COVERAGE-ON$
   }
   override def fromBytes(bytes: ByteVector): DoubleSha256Digest =
@@ -121,6 +131,8 @@ object DoubleSha256Digest extends Factory[DoubleSha256Digest] {
 
 /** The big endian version of [[org.bitcoins.core.crypto.DoubleSha256Digest DoubleSha256Digest]] */
 sealed abstract class DoubleSha256DigestBE extends HashDigest {
+
+  override def toString(): String = s"DoubleSha256DigestBE($hex)"
   def flip: DoubleSha256Digest = DoubleSha256Digest.fromBytes(bytes.reverse)
 }
 
@@ -130,7 +142,6 @@ object DoubleSha256DigestBE extends Factory[DoubleSha256DigestBE] {
     require(bytes.length == 32,
             // $COVERAGE-OFF$
             "DoubleSha256Digest must always be 32 bytes, got: " + bytes.length)
-    override def toString = s"DoubleSha256BDigestBEImpl($hex)"
     // $COVERAGE-ON$
   }
   override def fromBytes(bytes: ByteVector): DoubleSha256DigestBE =
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -8,5 +8,13 @@ bitcoin-s {
         # todo(torkelrogstad): I have no idea what a fitting 
         # default value here is 
         bloomFalsePositiveRate = 0.01
+
+
+        # settings for how we interact with the P2P network
+        p2p {
+            # if set to true, all raw P2P messages are 
+            # dumped to file
+            dumpRawBytes = true # todo: change me to false
+        }
     }
 }
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -1,4 +1,12 @@
 bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
+
+    # settings for node module
+    node {
+        # False positive rate for our Bloom filters
+        # todo(torkelrogstad): I have no idea what a fitting 
+        # default value here is 
+        bloomFalsePositiveRate = 0.01
+    }
 }
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -2,6 +2,11 @@ bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
 
+    # settings for wallet module
+    wallet {
+        defaultAccountType = legacy # legacy,, segwit, nested-segwit
+    }
+
     # settings for node module
     node {
         # False positive rate for our Bloom filters
--- a/node-test/src/main/scala/org/bitcoins/node/Bloom.scala
+++ b/node-test/src/main/scala/org/bitcoins/node/Bloom.scala
@@ -1,0 +1,90 @@
+package org.bitcoins.node
+
+import java.net.{InetAddress, InetSocketAddress}
+
+import org.bitcoins.chain.blockchain.ChainHandler
+import org.bitcoins.chain.config.ChainAppConfig
+import org.bitcoins.chain.models.{BlockHeaderDAO, BlockHeaderTable}
+import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.node.config.NodeAppConfig
+import org.bitcoins.node.constant.Constants
+import org.bitcoins.node.models.Peer
+import slick.jdbc.SQLiteProfile.api._
+
+import scala.concurrent._
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+import org.bitcoins.wallet.LockedWallet
+import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
+import org.bitcoins.rpc.client.common.BitcoindRpcClient
+import org.bitcoins.rpc.config.BitcoindInstance
+import org.bitcoins.wallet.config.WalletAppConfig
+import org.bitcoins.wallet.Wallet
+import org.bitcoins.core.crypto.ECPublicKey
+import scodec.bits._
+import org.bitcoins.wallet.api.CreateWalletApi
+import org.bitcoins.wallet.api.InitializeWalletSuccess
+import org.bitcoins.wallet.api.InitializeWalletError
+import org.bitcoins.core.bloom.BloomFilter
+import org.bitcoins.core.protocol.CompactSizeUInt
+import org.bitcoins.core.number.UInt32
+
+object Bloom extends App with BitcoinSLogger {
+  implicit val system = Constants.actorSystem
+  import system.dispatcher
+
+  implicit val nodeConfig = NodeAppConfig()
+  implicit val chainAppConfig = ChainAppConfig()
+  implicit val walletConf = WalletAppConfig()
+
+  logger.info(s"Chain config: ${chainAppConfig.dbConfig.config}")
+
+  val bhDAO = BlockHeaderDAO(chainAppConfig)
+  val chainApi = ChainHandler(bhDAO, chainAppConfig)
+  val table = TableQuery[BlockHeaderTable]
+
+  val chainInitF = chainAppConfig.initialize
+  Await.result(chainInitF, 3.seconds)
+
+  val instance = BitcoindInstance.fromDatadir()
+  val rpc = new BitcoindRpcClient(instance)
+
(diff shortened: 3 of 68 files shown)


core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
```diff
@@ -99,6 +99,17 @@ object Sha256DigestBE extends Factory[Sha256DigestBE] {
...
 sealed abstract class DoubleSha256Digest extends HashDigest {
+
+  /**
+    * @note The string representation of this is given in
+    *       big endian style. This is in order to make it
+    *       easier to scan logs for mentions of the hash.
+    *       (Bitcoin Core exposes all human facing hashes
+    *       in big endian style).
+    */
+  override def toString(): String =
```
torkelrogstad, line 110: I think this is a good idea, but open to other views. What do you think?

node-test/src/main/scala/org/bitcoins/node/Bloom.scala
```diff
@@ -0,0 +1,90 @@
+package org.bitcoins.node
```
torkelrogstad, line 1: This file can be disregarded until the PR is merge-ready, this is a nice main method to have available when developing

core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
Christewart: I don't think we should flip the hex here, that is super misleading to have it be `DoubleSha256Digest([hash])` but the hash is BE!

db-commons/src/main/resources/reference.conf
```diff
@@ -1,4 +1,25 @@
...
+    wallet {
+        defaultAccountType = legacy # legacy,, segwit, nested-segwit
+    }
+
+    # settings for node module
+    node {
+        # False positive rate for our Bloom filters
+        # todo(torkelrogstad): I have no idea what a fitting 
+        # default value here is 
+        bloomFalsePositiveRate = 0.01
```
Christewart, line 15: @nkohen do you have any input here?

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -15,16 +15,50 @@ import org.bitcoins.node.networking.peer.{
...
+  *
+  * @param bloomFilter The current bloom filter of the node, if any
+  * @param callbacks Set of optional callbacks that get's executed
+  *        at certain events in the node lifecycle.
+  */
+case class SpvNode(
+    peer: Peer,
+    chainApi: ChainApi,
+    bloomFilter: Option[BloomFilter] = None,
+    callbacks: SpvNodeCallbacks = SpvNodeCallbacks())(
```
Christewart: nit: Something like `SpvNodeCallbacks.empty` rather than constructing a new object every time

node/src/main/scala/org/bitcoins/node/models/InterestingPubKeyDAO.scala
```diff
@@ -0,0 +1,28 @@
...
+import scala.concurrent.Future
+
+case class InterestingPubKeyDAO()(
+    implicit val appConfig: NodeAppConfig,
+    val ec: ExecutionContext)
+    extends CRUD[ECPublicKey, ECPublicKey] {
+  import org.bitcoins.db.DbCommonsColumnMappers._
+
+  def createAll(ts: Vector[ECPublicKey]): Future[Vector[ECPublicKey]] = {
+    val actions = ts.map(table += _)
```
Christewart, line 17: Can we use `SlickUtil.createAllNoAutoInc`?

node/src/main/scala/org/bitcoins/node/models/InterestingPubKeyDAO.scala
```diff
@@ -0,0 +1,28 @@
+package org.bitcoins.node.models
+
+import slick.jdbc.SQLiteProfile.api._
+import org.bitcoins.db.CRUD
+import org.bitcoins.node.config.NodeAppConfig
+import scala.concurrent.ExecutionContext
+import org.bitcoins.core.crypto.ECPublicKey
+import scala.concurrent.Future
+
+case class InterestingPubKeyDAO()(
```
Christewart, line 10: Ok, i'm not convinced that we need this. It seems like it should be the wallet's responsibility for telling us what public keys we are interested in? 

What is the difference between this and `AddressDAO` in the wallet project?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -37,17 +56,93 @@ class DataMessageHandler(appConfig: ChainAppConfig)(
...
+
+    val inventories: Seq[Inventory] = {
+      val listOfOpts = for {
+        inventory <- invMsg.inventories
+      } yield {
+        inventory.typeIdentifier match {
+          // changing the inventory type might be naughty,
+          // but we're not interested
+          // in the complete block
+          case MsgBlock =>
```
Christewart, line 94: Yes this does seem bad, why are we doing this?

node/src/main/scala/org/bitcoins/node/SpvNodeCallbacks.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import org.bitcoins.node.networking.peer.DataMessageHandler._
+
+/**
+  * Callbacks for responding to events in the SPV node.
+  * The approriate callback is executed whenver the node receives
+  * a `getdata` message matching it.
+  *
+  */
+case class SpvNodeCallbacks(
```
Christewart, line 11: So I really like this idea! This really provides loose coupling of our various modules.

node/src/main/scala/org/bitcoins/node/SpvNode.scala
Christewart: Also i think this should be a `Vector[SpvNodeCallbacks]`? I don't think it is that far stretched that there are multiple callbacks that can be provided for various messages? For instance in the wallet project we may want to save the block hash that a utxo is contained in, and in the chain project we would want to add that block's blockheader to our chainstate

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -33,27 +67,105 @@ case class SpvNode(peer: Peer, chainApi: ChainApi)(
...
 
-    isInitializedF.failed.foreach { err =>
-      logger.error(s"Failed to conenct with peer=$peer with err=${err}")
+  /**
+    * Constructs and sends a bloom filter based on
+    * our table of pubkeys that's marked as interesting
+    *
+    * @return An SPV node with the newly set bloom filter
+    */
+  private def initBloomFilter(): Future[SpvNode] = {
```
Christewart, line 128: So I think I disagree with putting this here. I think the bloom filter should be constructed in the wallet project and available via the `WalletApi`. I think we should then be able to call something like `SpvNode.fromBloomFilter()` or something like that. I don't think it should be the SpvNodes responsibility to construct the filter. Only monitor the p2p messages if we have chosen to use bloom filters. 

Also we need to be able to support updating the filter, this may be out of scope for this PR but we need the ability update the filter without disconnecting the underlying peer.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -37,17 +56,93 @@ class DataMessageHandler(appConfig: ChainAppConfig)(
...
 
     FutureUtil.unit
 
   }
 }
+
+object DataMessageHandler extends BitcoinSLogger {
+
+  /** Callback for handling a received block */
+  type OnBlockReceived = Block => Unit
```
Christewart, line 122: I have a feeling that these signatures should be returning `Future[Unit]` so we don't block on whatever they are doing in this function?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -229,16 +231,21 @@ object PeerMessageReceiver {
 
   def apply(
       state: PeerMessageReceiverState,
+      callbacks: SpvNodeCallbacks = SpvNodeCallbacks()
```
Christewart: SpvNodeCallbacks.empty

testkit/src/main/scala/org/bitcoins/testkit/BitcoinSAppConfig.scala
```diff
@@ -0,0 +1,75 @@
...
+  implicit def implicitToNodeConf(
+      implicit conf: BitcoinSAppConfig): NodeAppConfig =
+    conf.nodeConf
+
+  /** Converts the given config to a node config */
+  implicit def toNodeConf(conf: BitcoinSAppConfig): NodeAppConfig =
+    conf.nodeConf
+
+  /** App configuration with data directory to user temp directory */
+  lazy val configWithTmpDatadir = {
```
Christewart: Is this going to cause issues with running tests across all suites at the same time? They may interfere with each other's configs?

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -104,6 +109,33 @@ trait NodeUnitTest
...
   }
 
+  /**
+    * Gives a triple containing a SPV node, a `bitcoind` instance
+    * and a wallet. They are not connected in any way, but
+    * the `bitcoind` instance has some money in it
+    */
+  def withNodeAndBitcoindAndWallet(test: OneArgAsyncTest)(
+      implicit system: ActorSystem): FutureOutcome = {
+    type Triple = (SpvNode, BitcoindRpcClient, UnlockedWalletApi)
```
Christewart, line 119: I would much prefer we just create a case class rather than use tuples

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -86,13 +103,22 @@ trait BitcoinSWalletTest
 
   def withNewWalletAndBitcoind(test: OneArgAsyncTest): FutureOutcome = {
     val builder: () => Future[WalletWithBitcoind] = composeBuildersAndWrap(
-      createNewWallet,
+      createNewWallet _,
```
Christewart, line 106: Is this kind of syntax going to be ok with scala 2.11?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
torkelrogstad: I'm not sure that's needed. Consider this:
```scala
def process(func: () => Any) = {
  println("before in process")
  func()
  println("after in process")
} 

val futureFunc: () => Future[Unit] = { () => 
  println("before the future")
  Future { 
    Thread.sleep(3000)
    println("Just slept")
  }
} 
```

Calling `process(futureFunc)` gives the following output: 
```scala
@ process(futureFunc) 
before in process
before the future
after in process
Just slept
```

So it's really up to the user at call site to block or not to block. But that's always the case, we can't prevent people from blocking if they really want to.

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
torkelrogstad: It compiles fine with sbt and Scala version set to 2.11.12

torkelrogstad:
Closing in favor of #514

PR #494, opened 2019-06-05
Multi module configuration

torkelrogstad:
In this PR we:
    
  1) Introduce new database settings for all submodules,
    and at the same time remove the singular datbase setting
    for the entire project
  2) Introduce a new method `.initialize()` on all configs
    that is responsible for creating needed directories,
    databases and files
  3) Introduce a new class `BitcoinSAppConfig` that wraps
    configs for all three other subproject config. We
    also provide implicit conversions that enable this
    super-config to be passed in wherever a specialized
    config is required.
  4) Add more tests for our configuration setup.


chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
```diff
@@ -53,9 +58,14 @@ case class ChainAppConfig(val confs: Config*) extends AppConfig {
           BlockHeaderDbHelper.fromBlockHeader(height = 0,
                                               bh =
                                                 chain.genesisBlock.blockHeader)
+        val blockHeaderDAO =
+          BlockHeaderDAO()(ec = implicitly[ExecutionContext], appConfig = this)
```
Christewart, line 62: why use `implicitly`? Can't we just use `ec = ec`?

testkit/src/main/scala/org/bitcoins/testkit/BitcoinSAppConfig.scala
```diff
@@ -0,0 +1,103 @@
...
+import com.typesafe.config.ConfigFactory
+
+/**
+  * A unified config class for all submodules of Bitcoin-S
+  * that accepts configuration. Thanks to implicit definitions
+  * in this case class' companion object an instance
+  * of this class can be passed in anywhere a wallet,
+  * chain or node config is required.
+  */
+case class BitcoinSAppConfig(confs: Config*) {
```
Christewart, line 19: So what is the implication here if you are only using a portion of bitcoin-s (i.e. wallet with no node, or chain with no node etc)? 

Maybe it doesn't matter and can be safely ignored?

I guess looking at this closer this is occurring in the testkit, so maybe it doesn't matter. But i would be nice to have something in src as well

Christewart:
It also looks like there are two bitcoind instances that are not being killed when tests run in this PR. 

EDIT: This is wrong, it is happening in master

testkit/src/main/scala/org/bitcoins/testkit/BitcoinSAppConfig.scala
torkelrogstad: This is a very valid objection, but that's also why I placed it in testkit. Once we get some things going that are _not_ just tests I think we can come back and revisit this point, and then we hopefully know more about what requirements and project layout looks like

chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
torkelrogstad: Just because I wasn't sure what the current execution context in scope was called. `implictly` works at compile level, so shoudln't make a difference. You could also argue that it's more readable, because it operates on the type level, and not on the "whatever I've called my constants"-level. If you disagree I can change it, not a big deal for me.

PR #495, opened 2019-06-05
Reshuffle package location of P2P messages

torkelrogstad:
Previously the companion objects for P2P messages
and the actual messages were in different pacakges.
This made them more awkward to use than necessary.

We also clean up some Scaladocs in this PR.

Update: All P2P messages and tests are now moved into core. I believe there's a few test failures in `node` now, haven't had time to look into that yet.


core/src/main/scala/org/bitcoins/core/p2p/NetworkIpAddress.scala
```diff
@@ -48,6 +47,29 @@ sealed abstract class NetworkIpAddress extends NetworkElement {
...
 
 object NetworkIpAddress extends Factory[NetworkIpAddress] {
+  /**
+    * Writes an IP address to the representation that the p2p network requires
+    * An IPv6 address is in big endian byte order
+    * An IPv4 address has to be mapped to an IPv6 address
+    *
+    * @see https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses
+    */
+  def writeAddress(iNetAddress: InetAddress): ByteVector = {
```
torkelrogstad, line 58: Not sure if this is the best location of this method. Moved it from `BitcoinSpvNodeUtil`, and I feel it's not really ideal to create a vague util object for this one method

PR #497, opened 2019-06-05
Legacy addresses

torkelrogstad:
In this PR we add the ability to generate
legacy addresses in the wallet. We also make
the default accound kind (legacy, segwit,
nested segwit) configurable.

We also fix (and add a test for) a bug
where overriding a value in our configuration
reset the rest of the configuration to the
default values. This is in the second commit.

### DIFF
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -1,4 +1,9 @@
 bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
+
+    # settings for wallet module
+    wallet {
+        defaultAccountType = legacy # legacy, segwit, nested-segwit
+    }
 }
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
@@ -10,26 +10,23 @@ import org.bitcoins.core.crypto.AesPassword
 import org.bitcoins.wallet.api.UnlockWalletError.MnemonicNotFound
 import com.typesafe.config.ConfigFactory
 import org.bitcoins.core.protocol.P2PKHAddress
+import org.bitcoins.core.hd.HDPurposes
 
 class LegacyWalletTest extends BitcoinSWalletTest {
 
   override type FixtureParam = UnlockedWalletApi
 
-  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    val confOverride =
-      ConfigFactory.parseString("bitcoin-s.wallet.defaultAccountType = legacy")
-    withNewConfiguredWallet(confOverride)(test)
-  }
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome =
+    withLegacyWallet(test)
 
-  // eventually this test should NOT succeed, as BIP44
-  // requires a limit to addresses being generated when
-  // they haven't received any funds
   it should "generate legacy addresses" in { wallet: UnlockedWalletApi =>
     for {
       addr <- wallet.getNewAddress()
+      account <- wallet.getDefaultAccount()
       otherAddr <- wallet.getNewAddress()
       allAddrs <- wallet.listAddresses()
     } yield {
+      assert(account.hdAccount.purpose == HDPurposes.Legacy)
       assert(allAddrs.forall(_.address.isInstanceOf[P2PKHAddress]))
       assert(allAddrs.length == 2)
       assert(allAddrs.exists(_.address == addr))
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
@@ -1,0 +1,39 @@
+package org.bitcoins.wallet
+
+import org.bitcoins.wallet.api.UnlockedWalletApi
+import org.bitcoins.wallet.util.BitcoinSWalletTest
+import org.scalatest.FutureOutcome
+import org.bitcoins.wallet.api.UnlockWalletError.BadPassword
+import org.bitcoins.wallet.api.UnlockWalletError.JsonParsingError
+import org.bitcoins.wallet.api.UnlockWalletSuccess
+import org.bitcoins.core.crypto.AesPassword
+import org.bitcoins.wallet.api.UnlockWalletError.MnemonicNotFound
+import com.typesafe.config.ConfigFactory
+import org.bitcoins.core.protocol.P2PKHAddress
+
+class LegacyWalletTest extends BitcoinSWalletTest {
+
+  override type FixtureParam = UnlockedWalletApi
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val confOverride =
+      ConfigFactory.parseString("bitcoin-s.wallet.defaultAccountType = legacy")
+    withNewConfiguredWallet(confOverride)(test)
+  }
+
+  // eventually this test should NOT succeed, as BIP44
+  // requires a limit to addresses being generated when
+  // they haven't received any funds
+  it should "generate legacy addresses" in { wallet: UnlockedWalletApi =>
+    for {
+      addr <- wallet.getNewAddress()
+      otherAddr <- wallet.getNewAddress()
+      allAddrs <- wallet.listAddresses()
+    } yield {
+      assert(allAddrs.forall(_.address.isInstanceOf[P2PKHAddress]))
+      assert(allAddrs.length == 2)
+      assert(allAddrs.exists(_.address == addr))
+      assert(allAddrs.exists(_.address == otherAddr))
+    }
+  }
+}
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/SegwitWalletTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/SegwitWalletTest.scala
@@ -1,0 +1,41 @@
+package org.bitcoins.wallet
+
+import org.bitcoins.wallet.api.UnlockedWalletApi
+import org.bitcoins.wallet.util.BitcoinSWalletTest
+import org.scalatest.FutureOutcome
+import org.bitcoins.wallet.api.UnlockWalletError.BadPassword
+import org.bitcoins.wallet.api.UnlockWalletError.JsonParsingError
+import org.bitcoins.wallet.api.UnlockWalletSuccess
+import org.bitcoins.core.crypto.AesPassword
+import org.bitcoins.wallet.api.UnlockWalletError.MnemonicNotFound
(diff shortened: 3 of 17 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.typesafe.config.ConfigFactory
+import org.bitcoins.core.protocol.P2PKHAddress
+
+class LegacyWalletTest extends BitcoinSWalletTest {
+
+  override type FixtureParam = UnlockedWalletApi
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val confOverride =
+      ConfigFactory.parseString("bitcoin-s.wallet.defaultAccountType = legacy")
```
Christewart: Would it be better to just create a fixture type called `LegacyUnlockedWalletApi`? I don't think it is a good idea to be modifying things in configs without having types to represent them.

wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
```diff
@@ -0,0 +1,39 @@
...
+    withNewConfiguredWallet(confOverride)(test)
+  }
+
+  // eventually this test should NOT succeed, as BIP44
+  // requires a limit to addresses being generated when
+  // they haven't received any funds
```
Christewart: 20 is the limit right? Perhaps add that. 

But I thought our wallets for the purposes of unit tests we have in memory databases? So every new run we create a new table which means we don't actually store our state to disk? Which means we shouldn't run into this limit?

wallet-test/src/test/scala/org/bitcoins/wallet/SegwitWalletTest.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.bitcoins.core.protocol.P2PKHAddress
+import org.bitcoins.core.protocol.Bech32Address
+
+class SegwitWalletTest extends BitcoinSWalletTest {
+
+  override type FixtureParam = UnlockedWalletApi
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val confOverride =
+      ConfigFactory.parseString("bitcoin-s.wallet.defaultAccountType = segwit")
```
Christewart: Again, i think this is just asking for a footgun. Let's use types! :-)

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
```diff
@@ -57,21 +58,44 @@ trait BitcoinSWalletTest
...
-      _ <- config.initialize()
-      wallet <- Wallet.initialize().map {
-        case InitializeWalletSuccess(wallet) => wallet
-        case err: InitializeWalletError =>
-          logger.error(s"Could not initialize wallet: $err")
-          fail(err)
-      }
-    } yield wallet
-  }
+  private val createNewWallet: Option[Config] => () => Future[
```
Christewart: scaladoc would be nice indicating the purpose of this. My understanding is it returns a function that can create a wallet with the given config

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
```diff
@@ -57,21 +58,44 @@ trait BitcoinSWalletTest
...
+        }
+
+        AppConfig.throwIfDefaultDatadir(walletConfig)
+
+        for {
+          _ <- walletConfig.initialize()
+          wallet <- Wallet
+            .initialize()(implicitly[ExecutionContext], walletConfig)
+            .map {
+              case InitializeWalletSuccess(wallet) => wallet
```
Christewart: There is where i think for expressions get really tough. I think we should be brekaing this out so it is more readable. Either create a function of use `.map` IMO

wallet-test/src/test/scala/org/bitcoins/wallet/util/BitcoinSWalletTest.scala
```diff
@@ -57,21 +58,44 @@ trait BitcoinSWalletTest
...
-          logger.error(s"Could not initialize wallet: $err")
-          fail(err)
-      }
-    } yield wallet
-  }
+  private val createNewWallet: Option[Config] => () => Future[
+    UnlockedWalletApi] =
+    extraConfig =>
+      () => {
+        val defaultConf = implicitly[WalletAppConfig]
```
Christewart: I don't really understand why this is here? Isn't this reading the default config in this trait? In general it seems like we wanted to use implicits, but now are passing them all explicitly which kind of defeats the point IMO.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -168,24 +166,26 @@ object Wallet extends CreateWalletApi with BitcoinSLogger {
...
+          _ <- {
+            // We want to make sure all level 0 accounts are created,
+            // so the user can change the default account kind later
+            // and still have their wallet work
+            val createAccountFutures =
+              HDPurposes.all.map(createRootAccount(wallet, _))
+
+            Future
+              .sequence(createAccountFutures)
+              .map(_ => logger.debug(s"Created root level accounts for wallet"))
```
Christewart: I think we want to move this log outside of the for expression and make sure we log on the failure case. This currently only logs on the success case

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -7,14 +7,26 @@ import org.bitcoins.wallet.db.WalletDbManagement
...
-case class WalletAppConfig(conf: Config*) extends AppConfig {
+case class WalletAppConfig(private val conf: Config*) extends AppConfig {
   override val configOverrides: List[Config] = conf.toList
   override def moduleName: String = "wallet"
   override type ConfigType = WalletAppConfig
-  override def newConfigOfType(configs: List[Config]): WalletAppConfig =
+  override def newConfigOfType(configs: Seq[Config]): WalletAppConfig =
     WalletAppConfig(configs: _*)
 
+  lazy val defaultAccountKind: HDPurpose =
```
Christewart, line 20: so since configs can be changed dynamically, don't we want this to be a `def` so the `defaultAccountKind` can be changed as well? 

i.e. if someone changes the config to change the default account, this value with be cached and not changed. In general we should be using `def` always with config stuff so everything gets modified at runtime

wallet/src/main/scala/org/bitcoins/wallet/models/UTXOSpendingInfoTable.scala
```diff
@@ -18,26 +18,39 @@ import org.bitcoins.core.hd.SegWitHDPath
...
+case class SegWitUTOXSpendingInfoDb(
     id: Option[Long],
     outPoint: TransactionOutPoint,
     output: TransactionOutput,
     privKeyPath: SegWitHDPath,
     scriptWitness: ScriptWitness
 ) extends UTXOSpendingInfoDb {
-  override def redeemScriptOpt: Option[ScriptPubKey] = None
-  override def scriptWitnessOpt: Option[ScriptWitness] = Some(scriptWitness)
+  override val redeemScriptOpt: Option[ScriptPubKey] = None
```
Christewart, line 33: Nit: But we should probably change the name of this case class to `NativeV0UTXOSpendingInfoDb` as per our discussion this morning about `v1` segwit coming out within the next year

wallet/src/main/scala/org/bitcoins/wallet/models/UTXOSpendingInfoTable.scala
```diff
@@ -114,20 +127,22 @@ case class UTXOSpendingInfoTable(tag: Tag)
           outpoint,
           output,
           path: SegWitHDPath,
-          None,
+          None, // ScriptPubKey
```
Christewart: This is the redeemScript right? Not the scriptPubKey?

db-commons/src/main/resources/reference.conf
```diff
@@ -1,4 +1,9 @@
 bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
+
+    # settings for wallet module
+    wallet {
+        defaultAccountType = legacy # legacy, segwit, nested-segwit
```
Christewart, line 7: not necessarily related to this PR, but we should really change the segwit key to `segwitv0` or something like that

torkelrogstad: Not sure I agree on that point, at least for the time being. I'm a big believer in the ["principle of least surprise"](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). In that spirit I think following convention for naming things makes it a lot easier for users. We could optionally make both `segwit` and `segwitv0` map to the same thing, etc

torkelrogstad: Once a taproot wallet proposal comes out, we could call that `taproot` (or whatever the canonical name becomes) + `segwitv1`

wallet/src/main/scala/org/bitcoins/wallet/models/UTXOSpendingInfoTable.scala
torkelrogstad: good point, fixing

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
torkelrogstad: That's a good point. Right now I haven't put much though into what configuration is dynamic and what is set at initialization. Another question apart from what it is right now is what is _desired_

Pros:
- it's flexible

Cons: 
- it's global, mutable state on steroids

I think I lean more towards an approach where you have to explicitly call a method `updateConfiguration`, or something along those lines.

wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
torkelrogstad: The `eventually` here refers to "once we get around to implementing that part of the BIP". But you're also right about using in-memory databases. I think I'll remove that comment, as it is slightly misleading.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: Added a log on the failure case as well. But not sure about moving it out of the `for`. That would mean we log on the failure of the entire process, not just the account creation

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
Christewart: Let's file an issue and worry about it later?

PR #498, opened 2019-06-06
Add debug prints to bitcoindCall when not on mainnet

torkelrogstad:
We're seeing very weird errors where we try and parse HTML every now and then: https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/541883223#L863

This is an attempt at debugging this. No functionality is changed, except for adding more logs if we're **not** on mainnet

More examples of this error: 
https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/542171259#L533
https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/542171260#L598
https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/542171260#L734

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
@@ -25,6 +25,7 @@ import org.bitcoins.rpc.config.BitcoindAuthCredentials.CookieBased
 import org.bitcoins.rpc.config.BitcoindAuthCredentials.PasswordBased
 import java.nio.file.Path
 import org.bitcoins.rpc.config.BitcoindAuthCredentials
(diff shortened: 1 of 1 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -219,26 +220,37 @@ trait Client extends BitcoinSLogger {
...
 
+        }
+      }
+      .recover {
+        // this can contain sensitive information, so only log this
+        // if not on mainnet
+        case err: JsonParseException if network != MainNet =>
+          logger.error(s"Error when parsing result of command: $command")
+          logger.error(s"Parameters: ${Json.stringify(JsArray(parameters))}")
+          logger.error(s"Sent HTML request: $request")
```
torkelrogstad: http

PR #499, opened 2019-06-06
Refactor logback configuration to reduce repetition

torkelrogstad:
In this PR we delete most Logback files, and are left with three files:

- `common-logback.xml` - This contains the actual logging configuration we're going to use our subprojects. Note that we change the default console output to make it more readable (shortening the timestamp and remove the calling thread name)
- `logback.xml` in `core` - Extend `common-logback.xml`, that's it. All projects depends on `core`, so this gets propagated to all projects
- `logback-test.xml` in `core-test` - Turn off all logging. This only gets applied to `core-test`, as no projects depend on `core-test`


Christewart:
Just a note, putting logging configuration in your library is usually bad practice. It leads to confusing logging settings  for users of the library and makes it hard to override the logging we have pre-configured. I think testkit is the best place for this to live as all of our testing projects depend on testkit

core/src/main/resources/common-logback.xml
```diff
@@ -1,4 +1,8 @@
 <included>
```
Christewart, line 1: So just to be clear, logback.xml does not pick this up by default right? We just need to include it in the sub projectgs we want to get these default logging settings? 

If so, my comment that I left is wrong. This seems to be a sane way to do this for our own internal development purposes while not externally showing this to a user of the library

PR #501, opened 2019-06-06
Generate bloom filters from wallet

torkelrogstad:
In this PR we add support for generating a bloom filter from our wallet. 

We also refactor some bloom tests in `core`, and add logging of SQL errors if we're not on mainnet. 

We also fix a bug in HDPath in `core`


wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
```diff
@@ -40,6 +49,89 @@ class WalletUnitTest extends BitcoinSWalletTest {
     }
   }
 
+  it should "know what the last address index is" in { walletApi =>
+    val wallet = walletApi.asInstanceOf[Wallet]
+
+    def getMostRecent(
+        chain: HDChainType,
+        acctIndex: Int
+    ): Future[AddressDb] =
```
Christewart: As code style i would prefer to wrap things like this in `{}` and use dot notation on the `.map`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -73,6 +75,26 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
   override def listAddresses(): Future[Vector[AddressDb]] =
     addressDAO.findAll()
 
+  /** Enumerates the public keys in this wallet */
+  private[wallet] def listPubkeys(): Future[Vector[ECPublicKey]] =
+    addressDAO.findAllPubkeys().map(_.toVector)
+
+  // todo: insert TXIDs? need to track which txids we should
+  // ask for, somehow
```
Christewart, line 96: So basically we need a table for transactions we have created and signed? 

An side note here:

We should _always_ treat a signed transaction as broadcasted, I don't think we should model the state transitition as 

1. Unsigned tx
2. Signed but not broadcasted
3. Signed and broadcasted
4. Confirmed

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -73,6 +75,26 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+  override def getBloomFilter(): Future[BloomFilter] = {
+    listPubkeys().map { pubkeys =>
+      // todo: Is this the best flag to use?
+      val bloomFlag = BloomUpdateAll
+
+      val baseBloom =
+        BloomFilter(numElements = pubkeys.length * 2,
```
Christewart: What is the `*2` here?

wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
```diff
@@ -39,12 +40,21 @@ case class AddressDAO()(
...
+    * Finds the most recent change address in the wallet, if any
+    */
   def findMostRecentChange(accountIndex: Int): Future[Option[AddressDb]] = {
     val query = findMostRecentForChain(accountIndex, HDChainType.Change)
 
     database.run(query)
   }
 
+  /** Finds all public keys in the wallet */
+  def findAllPubkeys(): Future[Seq[ECPublicKey]] = {
```
Christewart: `Vector[]` rather than `Seq[]`

core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala
```diff
@@ -1,8 +1,26 @@
...
 object FutureUtil {
 
+  /**
+    * Executes a series of futures sequentially
+    *
+    * @param items The elements we want to transform into futures
+    * @param fun A function that transforms each element into a future
+    * @return The processed elements
+    */
+  def sequentially[T, U](items: Seq[T])(fun: T => Future[U])(
```
Christewart, line 15: Is this something that exists in 2.12 and not 2.11? Maybe add a comment about it if that is the case

torkelrogstad: It does not exist in 2.12. There's `Future.sequence` that runs a sequence of futures in parallell, but no method for running them one after another

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: Yeah so this is more for future reference, once that gets added. Thinking we do the stupid thing for now, which is just scan for money on pubkeys we own. Otherwise I agree with your comment

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -73,6 +75,43 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+    }
+
+  }
+
+  // todo: insert TXIDs? need to track which txids we should
+  // ask for, somehow
+  override def getBloomFilter(): Future[BloomFilter] = {
+    for {
+      pubkeys <- listPubkeys()
+      filterSize <- getBloomFilterSize()
```
Christewart, line 100: So this takes care of the case of us _receiving_ money, but doesn't do anything for the case where our utxos are spent on the network. We need to iterate through our `UTXOSpendingInfoTable` and insert the hashes of the script.

PR #511, opened 2019-06-11
Website guides on HD key and address generation

torkelrogstad:
I also slightly clean up TransactionTest, as I stumbled on that while writing the guides.


docs/core/addresses.md
```diff
@@ -0,0 +1,62 @@
...
+val segwitAddress = {
+    // see https://bitcoin.org/en/glossary/pubkey-script
+    // for reading resources on the details of scriptPubKeys
+    // pay-to-witness-pubkey-hash scriptPubKey V0
+    val scriptPubKey = P2WPKHWitnessSPKV0(pubkey)
+    Bech32Address(scriptPubKey, TestNet3)
+}
+```
+
+## Generating legacy (base58) addresses
```
nkohen, line 49: Just because it's so much simpler I'd argue this section should go above segwit

docs/core/addresses.md
```diff
@@ -0,0 +1,62 @@
...
+val privkey = ECPrivateKey()
+val pubkey = privkey.publicKey
+
+val segwitAddress = {
+    // see https://bitcoin.org/en/glossary/pubkey-script
+    // for reading resources on the details of scriptPubKeys
+    // pay-to-witness-pubkey-hash scriptPubKey V0
+    val scriptPubKey = P2WPKHWitnessSPKV0(pubkey)
+    Bech32Address(scriptPubKey, TestNet3)
+}
```
nkohen, line 46: Any reason not to have this be
```
// see https://bitcoin.org/en/glossary/pubkey-script
// for reading resources on the details of scriptPubKeys
// pay-to-witness-pubkey-hash scriptPubKey V0
val scriptPubKey = P2WPKHWitnessSPKV0(pubkey)

val segwitAddress = Bech32Address(scriptPubKey, TestNet3)
```

docs/core/hd-keys.md
```diff
@@ -0,0 +1,131 @@
...
+id: hd-keys
+title: HD key generation
+---
+
+In modern Bitcoin wallets, users only need to write down
+a sequence of words, and that sequence is a complete backup
+of their wallet. This is thanks to what's called Hierarchical
+Deterministic key generation. In short, every wallet using HD
+key generation has a root seed for each wallet, and this
+seed can be used to generate an infinite amount of later
```
nkohen: nit: replace infinite with arbitrary ;)

docs/core/hd-keys.md
```diff
@@ -0,0 +1,131 @@
...
+
+> If you want to jump into the details of how this work,
+> you should check out
+> [BIP 32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki).
+
+Bitcoin-S supports generating keys in this fashion. Here's a
+full example of how to obtain a wallet seed, and then
+use that to generate further private and public keys:
+
+Here's an example:
```
nkohen: You say here's an example twice (once in each paragraph)

docs/core/hd-keys.md
```diff
@@ -0,0 +1,131 @@
...
+val segwitPath = SegWitHDPath.fromString("m/84'/0'/0'/0/0")
+
+// alternatively:
+val otherSegwitPath =
+SegWitHDPath(HDCoinType.Bitcoin,
+             accountIndex = 0,
+             HDChainType.External,
+             addressIndex = 0)
+
+segwitPath == otherSegwitPath
```
nkohen, line 63: either wrap in `assert` or comment `// true` for clarity

docs/rpc/rpc-clients-intro.md
```diff
@@ -3,4 +3,27 @@ id: rpc-clients-intro
...
+have a custom type that represents a Bitcoin transaction,
+how do you get that to play nicely with the result that
+Bitcoin Core gives you after signing a transaction? A
+random hexadecimal string in a HTTP response could be
+anything from a public key, a transaction or a block
+header.
+
+We've done all the mundane work of wiring requests and
+responses from Bitcoin Core to the powerful and safe types
+found in Bitcoin-S. You'll know for sure that you're sending
```
nkohen: mention tests too?

docs/core/addresses.md
torkelrogstad: I was thinking I'd have a variable named `scriptPubKey` beneath here as well, but that didn't end up happening. do you think it's confusing with a block with vals inside them?

docs/core/addresses.md
torkelrogstad: We want to push people towards segwit though! And That's also one of the differentiating things with Bitcoin-S, how easy segwit addresses are. Notice that the code in the block below reuses the values from the block above, so the only lines that are different are 
```scala
val scriptPubKey = P2WPKHWitnessSPKV0(pubkey)
    Bech32Address(scriptPubKey, TestNet3)
```
and 
```scala
 val legacyAddress = P2PKHAddress(pubkey, TestNet3)
```

docs/core/hd-keys.md
torkelrogstad: So this is one of the cool things about mdoc (https://github.com/scalameta/mdoc), it typechecks and even runs the code before publishing. So `// true` gets added at the bottom by the compiler

docs/rpc/rpc-clients-intro.md
torkelrogstad: good point

docs/core/addresses.md
nkohen: Ah, good points :)

docs/core/addresses.md
nkohen: I think it's marginally more work looking back and forth to see what is what, not a big deal obviously

docs/core/hd-keys.md
nkohen: gotchya, very cool!

PR #514, opened 2019-06-11
Pass bloom filter to SPV node

torkelrogstad:
In this PR we verify that passing a bloom filter from the wallet to the node actually triggers something with our Bitcoin Core peer. No further action is taken in the wallet upon receiving data.


node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,143 @@
...
+        bloom <- wallet.getBloomFilter()
+
+        spv <- {
+          val spv =
+            SpvNode(peer,
+                    chainHandler,
+                    bloomFilter = Some(bloom),
+                    callbacks = Vector(callbacks))
+          spv.start()
+        }
```
Christewart, line 115:
```scala
_ <- spv.sync()
```
is needed to actually start the sync

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,136 @@
...
+class NodeWithWalletTest extends BitcoinSWalletTest {
+
+  override type FixtureParam = WalletWithBitcoind
+
+  def withFixture(test: OneArgAsyncTest): FutureOutcome =
+    withNewWalletAndBitcoind(test)
+
+  it must "load a bloom filter and receive information about received payments" in {
+    param =>
+      val WalletWithBitcoind(wallet, rpc) = param
```
nkohen, line 36: any reason to not just do `case WalletWithBitcoind(wallet, rpc) => ...` ?

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,136 @@
...
+
+      /**
+        * This is not ideal, how do we get one implicit value (`config`)
+        * to resolve to multiple implicit parameters?
+        */
+      implicit val nodeConfig: NodeAppConfig = config
+      implicit val chainConfig: ChainAppConfig = config
```
nkohen, line 43: Not sure this would work but perhaps `implicit val config: NodeAppConfig with ChainAppConfig`?

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,136 @@
...
+        onMerkleBlockReceived = block => (),
+        onTxReceived = { tx =>
+          if (expectedTxId.contains(tx.txId)) {
+            cancellable.map(_.cancel())
+            completionP.success(succeed)
+          } else if (unexpectedTxId.contains(tx.txId)) {
+            completionP.failure(
+              new TestFailedException(
+                s"Got ${tx.txId}, which is a TX we expect to NOT get notified about!",
+                failedCodeStackDepth = 0))
```
nkohen, line 67: I think this should be
```
fail(s"Got ${tx.txId}, which is a TX we expect to NOT get notified about!")
```

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,136 @@
...
+        ourTxid <- rpc.sendToAddress(address, 1.bitcoin).map { tx =>
+          expectedTxId = Some(tx.flip)
+          // how long we're waiting for a tx notify before failing the test
+          val delay = 15.seconds
+          val runnable: Runnable = new Runnable {
+            override def run = {
+              val msg =
+                s"Did not receive sent transaction within $delay"
+              logger.error(msg)
+              completionP.failure(new TestFailedException(msg, 0))
```
nkohen: same as above

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -15,16 +15,24 @@ import org.bitcoins.node.networking.peer.{
...
-case class SpvNode(peer: Peer, chainApi: ChainApi)(
+import org.bitcoins.core.bloom.BloomFilter
+import org.bitcoins.core.p2p.FilterLoadMessage
+import org.bitcoins.core.p2p.NetworkPayload
+
+case class SpvNode(
+    peer: Peer,
+    chainApi: ChainApi,
+    callbacks: Vector[SpvNodeCallbacks] = SpvNodeCallbacks.empty,
+    bloomFilter: Option[BloomFilter] = None
```
nkohen: In what situation would this ever be `None`?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -47,8 +68,31 @@ class DataMessageHandler()(
...
+  type OnMerkleBlockReceived = MerkleBlock => Unit
+
+  /** Does nothing with the received Merkle block */
+  val noopMerkleBlockReceived: OnMerkleBlockReceived = _ => ()
+
+  /** Callback for handling a received transaction */
+  type OnTxReceived = Transaction => Unit
+
+  /** Does nothing with the received transaction */
+  val noopTxReceived: OnTxReceived = _ => ()
```
nkohen: If you wanted you could just make a `def noop[T]: T => Unit = _ => ()` to avoid having all these vals?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -5,28 +5,40 @@ import org.bitcoins.chain.blockchain.ChainHandler
...
+import org.bitcoins.core.p2p.MerkleBlockMessage
+import org.bitcoins.node.SpvNodeCallbacks
+import org.bitcoins.core.p2p.GetDataMessage
 
 /** This actor is meant to handle a [[org.bitcoins.node.messages.DataPayload]]
   * that a peer to sent to us on the p2p network, for instance, if we a receive a
   * [[HeadersMessage]] we should store those headers in our database
   */
-class DataMessageHandler()(
+class DataMessageHandler(callbacks: Vector[SpvNodeCallbacks])(
```
nkohen: Rather than using `Vector[SpvNodeCallbacks]` everywhere which requires we have the same number of each kind of callback specified (resulting in noops), is there any reason we don't instead let `SpvNodeCallbacks` have three `Vector`s instead of three vals?

nkohen: The abstraction just seems like an odd module to me

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -38,6 +50,15 @@ class DataMessageHandler()(
...
         }
+      case msg: BlockMessage =>
+        callbacks.foreach(_.onBlockReceived(msg.block))
+        FutureUtil.unit
+      case msg: TransactionMessage =>
+        callbacks.foreach(_.onTxReceived(msg.transaction))
+        FutureUtil.unit
+      case msg: MerkleBlockMessage =>
+        callbacks.foreach(_.onMerkleBlockReceived(msg.merkleBlock))
+        FutureUtil.unit
```
nkohen: If you wrap the callback lines in `Future` then failures will nicely result in failed `Future`s rather than `throw`s

testkit/src/main/scala/org/bitcoins/testkit/node/NodeTestUtil.scala
```diff
@@ -95,6 +99,27 @@ abstract class NodeTestUtil {
...
+    } yield {
+      spvBestHash == hash
+    }
+  }
+
+  /** Awaits sync between the given SPV node and bitcoind client */
+  def awaitSync(node: SpvNode, rpc: BitcoindRpcClient)(
+      implicit sys: ActorSystem): Future[Unit] = {
+    import sys.dispatcher
+    RpcUtil
```
nkohen: I think that using `AsyncTestUtil` rather than `RpcUtil` results in nice `TestFailedException`s on failure

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -314,6 +314,19 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
     }
   }
 
+  override def getAddressInfo(
+      address: BitcoinAddress): Future[Option[AddressInfo]] = {
```
nkohen: Thoughts on flattening this to just a `Future`? How explicitly do you expect users to need to deal with the `None` case very differently from the `Future.failed` case?

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
torkelrogstad: That makes the compiler complain about expecting a function (as compared to a partial function) as the parameter to the test :-(

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
torkelrogstad: `fail` is similar to `???` in that when you try to evaluate it it throws: https://github.com/scalatest/scalatest/blob/441df125737543ce7e1776f087ca57e79ee17159/scalatest/src/main/scala/org/scalatest/Assertions.scala#L1101

So for the failure to propagate correctly I _think_ this is the way to do it

node/src/main/scala/org/bitcoins/node/SpvNode.scala
torkelrogstad: You may want to sync with block filters, or maybe use this node to actually download everything from the network

But you're right that right now there's probably no use cases where this would be none. Can change it

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
torkelrogstad: Good point

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: My reasoning here is twofold: 

1) A failed future isn't encoded into the type, which is bad. In that way its similar to `Option.get`
2) A failed future might be triggered by a DB problem, a network problem, a file system problem, or a bunch of things. Whereas a `None` cleanly indicates that the address isn't in the wallet

Christewart: I do agree with this reasoning, but it does make the api harder to work with with `for` expressions

node/src/main/scala/org/bitcoins/node/SpvNode.scala
Christewart: Yeah, especially with the name of the class being `SpvNode` which implies bloom filters IMO.

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
torkelrogstad: Doesn't work :-(

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -0,0 +1,136 @@
...
+            SpvNode(peer,
+                    chainHandler,
+                    bloomFilter = Some(bloom),
+                    callbacks = Vector(callbacks))
+          spv.start()
+        }
+        _ <- spv.sync()
+        _ <- NodeTestUtil.awaitSync(spv, rpc)
+
+        ourTxid <- rpc.sendToAddress(address, 1.bitcoin).map { tx =>
```
Christewart: Would prefer to break this `rpc.sendToAddress` lambda into a separate method.

core/src/main/scala/org/bitcoins/core/bloom/BloomFilter.scala
```diff
@@ -142,16 +142,18 @@ sealed abstract class BloomFilter extends NetworkElement {
...
     */
   def isRelevant(transaction: Transaction): Boolean = {
     val scriptPubKeys = transaction.outputs.map(_.scriptPubKey)
     //pull out all of the constants in the scriptPubKey's
     val constantsWithOutputIndex = scriptPubKeys.zipWithIndex.flatMap {
       case (scriptPubKey, index) =>
-        val constants = scriptPubKey.asm.filter(_.isInstanceOf[ScriptConstant])
-        constants.map(c => (c, index))
+        scriptPubKey.asm.collect {
+          case c: ScriptConstant => (c, index)
```
Christewart, line 155: Just a heads up, I think i have a bug here implementing this. This won't be _only_ data pushes, but will include things like `ScriptNumber`, `OP_0`,`OP_1`, etc...

Christewart: There are quite a few things that inherit from `ScriptConstant`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
nkohen: I would argue that `Future[T]` is just a wrapper for `Option[Try[T]]` but that point aside, I think that if the future is failed with a descriptive error then the issue of distinguishing a missing address from a db problem isn't an issue

torkelrogstad: Use case for having this return `Future[Option[..]]`: 

A user wants to detect if an address is _not_ in the wallet. With your approach the user would then either match on the type of failure in the future (which he gets no help from the compiler doing, as futures aren't parameterized on the error type), or check if the error message on the exception matches a certain substring (which is bad)

nkohen: I would say there's calls on Futures like `transform` and `onComplete` and `recover` that would let you handle this nicely if the error used in failure has its own type specific to a missing address but at the same time I don't think this is a huge deal so if you still think that `Future[Option]` is the way to go I'm okay with going with it and having the discussion if need be when it gets used in the future

PR #516, opened 2019-06-12
Generate test vectors from Trezor

torkelrogstad:
This PR includes:

1) Test vectors for deriving xpubs and addresses for legacy, segwit and nested segwit accounts, taken from a Trezor emulator on my machine
2) A program that given a running and initialized Trezor emulator generates these vectors
3) A test that parses the vectors into case classes we can use

Todo: 
- [x] Write the actual tests that generate addresses in our wallet and verify them against the test vectors
- [x] Add some documentation on how to reproduce this

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/TrezorAddressTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/TrezorAddressTest.scala
@@ -163,6 +163,22 @@ class TrezorAddressTest extends BitcoinSWalletTest with EmptyFixture {
       addrs: Seq[AddressDb],
       vector: TestVector)
 
+  /** Asserts that the given addresses are gthe same as in the given vector */
+  private def assertSameAddresses(
+      addrs: Seq[AddressDb],
+      vector: TestVector): Seq[Assertion] = {
+    assert(vector.addresses.length == addrs.length)
+
+    val sortedAddresses = addrs.sortBy(_.path.toString)
+    val sortedVectors = vector.addresses.sortBy(_.path.toString)
+    sortedAddresses
+      .zip(sortedVectors)
+      .map {
+        case (foundAddress, expectedAddress) =>
+          assert(foundAddress.address == expectedAddress.address)
+      }
+  }
+
   private def testAccountType(purpose: HDPurpose): Future[Assertion] = {
     val confOverride = configForPurpose(purpose)
     implicit val conf: WalletAppConfig =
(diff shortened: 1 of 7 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/TrezorAddressTest.scala
```diff
@@ -0,0 +1,266 @@
...
+      BitcoinSAppConfig.getTestConfig(confOverride)
+
+    val vectors = purpose match {
+      case HDPurposes.Legacy       => legacyVectors
+      case HDPurposes.SegWit       => segwitVectors
+      case HDPurposes.NestedSegWit => nestedVectors
+      case other                   => fail(s"unknown purpose: $other")
+    }
+
+    for {
```
Christewart: This for expression is massive, can we break it into smaller expressions?

wallet-test/src/test/scala/org/bitcoins/wallet/TrezorAddressTest.scala
```diff
@@ -0,0 +1,266 @@
...
+          assert(vec.addresses.length == addresses.length)
+
+          val sortedAddresses = addresses.sortBy(_.path.toString)
+          val sortedVectors = vec.addresses.sortBy(_.path.toString)
+          sortedAddresses.zip(sortedVectors).foreach {
+            case (foundAddress, expectedAddress) =>
+              assert(foundAddress.address == expectedAddress.address)
+          }
+      }
+      succeed
```
Christewart: I don't really like passing in `suceed` at the end, is there no way to actually get a `Future[Vector[Assertion]]` or something like that? @rorp pointed out yesterday we tests that were trivially passing in the eclair project because we were doing this succeed at the end which is obviously very bad.

wallet-test/src/test/scala/org/bitcoins/wallet/util/GetAddresses.scala
```diff
@@ -0,0 +1,102 @@
+package org.bitcoins.wallet.util
+
+import org.bitcoins.core.hd._
+import play.api.libs.json._
+import scala.sys.process._
+
+/** This program connects to a running Trezor, and gets
```
Christewart, line 7: A running trezor, or a running emulator? Is there no difference?

wallet-test/src/test/scala/org/bitcoins/wallet/util/GetAddresses.scala
```diff
@@ -0,0 +1,102 @@
...
+        case HDPurposes.SegWit       => "segwit"
+        case _                       => ???
+      }
+
+    val trezorPathType =
+      constant match {
+        case HDPurposes.Legacy       => "address"
+        case HDPurposes.NestedSegWit => "p2shsegwit"
+        case HDPurposes.SegWit       => "segwit"
+        case _                       => ???
```
Christewart: is this needed?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -98,6 +98,45 @@ sealed abstract class Wallet
...
+
+  // todo: check if there's addresses in the most recent
+  // account before creating new
+  override def createNewAccount(purpose: HDPurpose): Future[Wallet] = {
+
+    accountDAO
+      .findAll()
+      .map(_.filter(_.hdAccount.purpose == purpose))
+      .map(_.sortBy(_.hdAccount.index))
+      .map(_.lastOption)
```
Christewart, line 115: why do we want `lastOption`?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -98,6 +98,45 @@ sealed abstract class Wallet
...
+      .flatMap { mostRecentOpt =>
+        val accountIndex = mostRecentOpt match {
+          case None          => 0 // no accounts present in wallet
+          case Some(account) => account.hdAccount.index + 1
+        }
+        logger.debug(
```
Christewart: This is probably worth a info log no?

wallet-test/src/test/scala/org/bitcoins/wallet/util/GetAddresses.scala
torkelrogstad: There's no difference from `trezorctl`s point of view

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: Added a explanatory comment

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
torkelrogstad: It gets pretty noisy in tests, but I guess the better solution there is to turn log levels down in tests

rorp (approved):

Christewart (approved):

PR #517, opened 2019-06-13
Default to file based databases in tests

torkelrogstad:
This commit is extracted from #514. Without it CI is going to fail on both #514 and #516, so this should go in first, and then the other PRs should be rebased. 

---

This is due to issues with deadlocks with
in-memory based databases. We need DBs
to stay alive between connections, but
not across tests.

In DB intensive chain validation tests we
use in-memory databases instead. This
seems like a reasonable tradeoff between
simplicity and speed.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -15,15 +15,20 @@ import org.scalatest.{Assertion, FutureOutcome}
 import play.api.libs.json.Json
 
 import scala.concurrent.Future
+import org.bitcoins.testkit.BitcoinSAppConfig
 
 class ChainHandlerTest extends ChainUnitTest {
 
   override type FixtureParam = ChainHandler
 
-  override implicit val system = ActorSystem("ChainUnitTest")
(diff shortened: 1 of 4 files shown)


Christewart:
@torkelrogstad elaborated on why in memory database are not sufficient enough, specifically at the jdbc connection level in slack. Pasting here for completeness.

I asked the question 


>so is this an issue because we use [`lazy val`](https://github.com/bitcoin-s/bitcoin-s/blob/7a4c3c6b71db2affc7760290cab8839f430f8a50/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala#L130) to define db configs instead of `def`? Because I assumed each unit test would have it's own _unique_ in memory db spun up? Or is it dbs inside the specific test suite it self that are harming each other?


he replied
```
`lazy val` or `def` would make no difference.
This is my understanding of the problem:
- By default all in-memory DBs gets wiped on connection close
- We don't want that, as that would make it impossible to persist anything intra-test.
- We workaround this by adding `cache=shared` at the end of our JDBC URI, which makes the DB live across connections

I'm not exactly sure when a DB is wiped. It might be between JVMs, but I don't think that's the case (I don't think Bloop/sbt spins up a complete JVM for just one test). It might also be between tests, which is what it seems to do most of the time. But then I've been running into issues when testing the wallet integration tests, where different tests access the same DB
```

PR #522, opened 2019-06-14
Service identifier and node cleanup

torkelrogstad:
In this PR we

1) Add proper parsing of service identifier bit fields. The old implementation didn't catch the notion
that service identifiers was a bit field where
multiple services could be set at the same time.
2) Make `Peer` take an `InetSocketAddress`. Previously it took `NetworkIpAddress`. This doesn't make any sense, as we need to know the services bitfield to construct a `NetworkIpAddress`.
3) Clean up documentation in `node/README.md`
4) Delete unused files in `node` from when `node` was it's own separate repo
5) Make the main method in `node` runnable, and have it peer to a locally running `bitcoind` (from default datadir)
6) Clean up logging configuration and logging messages in `chain` and `node`
7) Tweak `build.sbt` to skip publishing of `node`, `chain`, and `wallet`


Christewart:
CI failure on scala 2.11.x for a compiler issue https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/546681047#L520

PR #532, opened 2019-06-17
BIP 157 Block Filter Headers

nkohen:
Implements and tests Headers (as specified in https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#filter-headers) for BIP 158 Filters (https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters)


rorp (approved):
LGTM

Christewart (approved):

Issue #530, opened 2019-06-17
MVP node + wallet


torkelrogstad:
This is an umbrella issue for all functionality needed for our SPV node + wallet combo to be a shippable software package (at least for expert users). Please don't submit any PRs that close this issue, rather create sub issues for each item on these lists, and then PRs that close those. Feel free to comment/edit this issue with further things that need to be added.

## Node

- [ ] Connect to multiple peers at the same time
- [ ] Bootstrap with DNS seeds, without explicitly passing in a peer
- [ ] Verify Merkle blocks/proofs
- [x] Handle reorgs (#453)
- [ ] Segwit/newer than 24 months P2P messages

## Wallet

- [x] Add API for importing transactions (#534)
- [x] Extend `processTransaction` to also handle outgoing TXs (#555)
- [ ] Sweeping private keys/seeds
- [ ] Handling reorgs
- [ ] Coin selection (partially fixed by #431)
- [ ] Managing change UTXOs correctly when spending (#555)
- [x] Manage UTXOs in DB correctly when spending (#538)
- [ ] Handle double-spends

## CLI/external communication

Open questions:

- [ ] How do we serve requests in the wallet/node? #536 
- [ ] How do we handle authentication?
- [x] In what language do we write the CLI? Scala has slow JVM startup, but if we can
      use Scala Native/GraalVM AOT compilation we could get it to work  
       **UPDATE**: I wrote a very basic PoC that exposes a nice CLI interface
      and AOT-compiles with GraalVM `native-image`. Link:
      https://github.com/torkelrogstad/bitcoin-s/commit/8898320e25e712b5b91c50801d990a974b6d7882

PR #534, opened 2019-06-18
Process incoming transactions

torkelrogstad:
In this PR we

1) Add new tables for outgoing and incoming transaction outputs, and refactor what was previously called the UTXO table to be a spending info table:
    - incoming outputs have foreign keys to our addresses, on the SPK
    - incoming outputs have foreign keys to spending info rows, which contain information such as the output value, etc
    - outgoing outputs have foreign keys to incoming outputs 
2) Add SQL joins to find incoming outputs with their corresponding addresses
3) Make table creation on test startup be one SQL transaction, instead of many roundtrips
4) Add a method `HDChain.toAddress` in `core`
5) Remove all existing DAO fixtures for the wallet, and instead create one big fixture that creates all tables, drops all tables and provides all DAOs. The different DAO fixtures all did the same thing anyway, so there was no loss in removing them, really.
6) Replace `addUtxo` with `processTransaction` in the wallet API. `addUtxo` is still kept around as an internal utility, but not exposed to the user. `processTransaction` takes in a TX and a number of confirmations, and updates internal state accordingly. 
7) Add a method `getUnconfirmedBalance` to the wallet API. `getBalance` is also updated, taking advantage of our new DB tables. 

TODO: 
  - [x] Add foreign key to UTXOs in incoming TXs
  - [x] Make all DAO tests pass
 -  [x] Add `processTransaction` to wallet
 -  [x] Refactor incoming/outgong transaction model to work with incoming/outgoing TXOs


wallet/src/main/scala/org/bitcoins/wallet/models/AddressTable.scala
```diff
@@ -118,29 +136,27 @@ object AddressDbHelper {
...
 
-  def addressIndex: Rep[Int] = column[Int]("address_index")
+  def address: Rep[BitcoinAddress] = column("address", O.PrimaryKey)
 
-  def address: Rep[BitcoinAddress] =
-    column[BitcoinAddress]("address", O.PrimaryKey)
+  def ecPublicKey: Rep[ECPublicKey] = column("pubkey")
 
-  def ecPublicKey: Rep[ECPublicKey] = column[ECPublicKey]("pubkey")
+  def hashedPubKey: Rep[Sha256Hash160Digest] = column("hashed_pubkey")
```
Christewart, line 152: nit: but this probably should be changed to `hash` and be of type `Rep[HashDigest]`

wallet/src/main/scala/org/bitcoins/wallet/models/TransactionTable.scala
```diff
@@ -0,0 +1,92 @@
...
+  val transaction: Transaction
+  lazy val txid: DoubleSha256DigestBE = transaction.txIdBE
+  val confirmations: Int
+
+}
+
+/** Transactions our wallet has received */
+final case class IncomingTransaction(
+    transaction: Transaction,
+    scriptPubKey: ScriptPubKey,
```
Christewart: i think output index is better than spk here, and we can add a method that gives amount, scriptPubKey, etc in the body of the case class

torkelrogstad: The problem there is that we need to create a foreign key on `scriptPubKey` into address table. I don't know how to do that without having the actual SPK as a field in the case class...

Christewart:
```scala
require(transaction.outputs.exists(_.scriptPubKey == scriptPubKey))
```

core/src/main/scala/org/bitcoins/core/hd/HDChain.scala
```diff
@@ -20,6 +20,9 @@ sealed abstract class HDChain extends BIP32Path {
 
   def toInt: Int = chainType.index
 
+  /** Given a index, creates a HD address */
+  def toAddress(index: Int): HDAddress = HDAddress(this, index = index)
```
Christewart: suggestion: `toHDAddress`

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
```diff
@@ -58,8 +60,16 @@ class NodeWithWalletTest extends BitcoinSWalletTest {
...
           if (expectedTxId.contains(tx.txId)) {
+            logger.debug(s"Cancelling timeout we set earlier")
             cancellable.map(_.cancel())
-            completionP.success(succeed)
+
+            for {
+              prevBalance <- wallet.getUnconfirmedBalance()
+              _ <- wallet.processTransaction(tx, confirmations = 0)
+              balance <- wallet.getUnconfirmedBalance()
+            } completionP.success(
```
Christewart: This seems like weird syntax, why no yield here?

Christewart: Also I can't get this test case to fail if I switch the predicate to be

```scala
assert(balance != prevBalance + amountFromBitcoind))
```

testkit/src/main/scala/org/bitcoins/testkit/wallet/WalletTestUtil.scala
```diff
@@ -64,5 +72,64 @@ object WalletTestUtil {
...
+    * This method also does some asserts on the result, to make sure what
+    * we're writing and reading matches up
+    */
+  def insertIncomingTx(daos: WalletDAOs)(implicit ec: ExecutionContext): Future[
+    (IncomingTransaction, AddressDb)] = {
+    val WalletDAOs(accountDAO, addressDAO, txDAO, _, _) = daos
+
+    /** Get a TX with outputs */
+    def getTx: Transaction =
+      TransactionGenerators.transaction
```
Christewart: It would probably be worth making a `TransactionGenerators.nonEmptyOutputs` generator

wallet-test/src/test/scala/org/bitcoins/wallet/ProcessTransactionTest.scala
```diff
@@ -0,0 +1,94 @@
...
+  private def checkUtxosAndBalance(
+      wallet: UnlockedWalletApi,
+      bitcoind: BitcoindRpcClient)(action: => Future[_]): Future[Assertion] =
+    for {
+      oldUtxos <- wallet.listUtxos()
+      oldUnconfirmed <- wallet.getUnconfirmedBalance()
+      oldConfirmed <- wallet.getBalance()
+      _ <- action // by name
+      newUtxos <- wallet.listUtxos()
+      newUnconfirmed <- wallet.getUnconfirmedBalance()
```
Christewart, line 34: not used?

wallet-test/src/test/scala/org/bitcoins/wallet/ProcessTransactionTest.scala
```diff
@@ -0,0 +1,94 @@
...
+        assert(newUnconfirmed != oldUnconfirmed)
+      }
+  }
+
+  it must "not change state when processing an unrelated transaction" in {
+    walletAndBitcoind =>
+      val WalletWithBitcoind(wallet, bitcoind) = walletAndBitcoind
+
+      for {
+        unrelated <- bitcoind.getNewAddress
```
Christewart: suggestion: Maybe used generators here so we don't need a full fledge bitcoind instance? It can be any arbitrary tx right?

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
```diff
@@ -27,40 +27,45 @@ class WalletIntegrationTest extends BitcoinSWalletTest {
...
+      _ <- wallet.getBalance().map(confirmed => assert(confirmed == 0.bitcoin))
+      _ <- wallet
+        .getUnconfirmedBalance()
+        .map(unconfirmed => assert(unconfirmed == 0.bitcoin))
+
+      // after this, tx is unconfirmed in wallet
+      _ <- wallet.processTransaction(tx, confirmations = 0)
+
+      utxosPostAdd <- wallet.listUtxos()
+      _ = assert(utxosPostAdd.nonEmpty)
```
Christewart, line 47:
```scala
assert(utxosPostAdd.length == 1)
```

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -26,16 +26,34 @@ import org.bitcoins.wallet.ReadMnemonicError.JsonParsingError
...
   private[wallet] val addressDAO: AddressDAO = AddressDAO()
   private[wallet] val accountDAO: AccountDAO = AccountDAO()
   private[wallet] val utxoDAO: UTXOSpendingInfoDAO = UTXOSpendingInfoDAO()
+  private[wallet] val incomingTxDAO = IncomingTransactionDAO(SQLiteProfile)
+  private[wallet] val outgoingTxDAO = OutgoingTransactionDAO(SQLiteProfile)
+
+  /** Sums up the value of all incoming
+    * TXs in the wallet, filtered by the given predicate */
+  // TODO account for outgoing TXs
+  private def sumOfIncomingGiven(
```
Christewart: `filterThenSum` as a name?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,11 +182,160 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
   /**
-    * @inheritdoc
+    * Processes an incoming transaction that's new to us
+    * @return `Some(tx)` if we found anything relevant to us
+    *         in the given transaction, `None` if nothing was
+    *         of relevance
+    */
+  private def processNewIncomingTx(
+      transaction: Transaction,
+      confirmations: Int): Future[Option[IncomingTransaction]] = {
```
Christewart: This return type should probably be `Future[Vector[IncomingTransaction]]`

wallet/src/main/scala/org/bitcoins/wallet/models/TransactionTable.scala
```diff
@@ -0,0 +1,106 @@
...
+  */
+sealed trait TransactionDb[T <: TransactionDb[_]] extends DbRowAutoInc[T] {
+  val transaction: Transaction
+  lazy val txid: DoubleSha256DigestBE = transaction.txIdBE
+  val confirmations: Int
+
+}
+
+/** Transactions our wallet has received */
+final case class IncomingTransaction(
```
Christewart: So I think this should be renamed to `IncomingUtxo`, as this has information relevant to the utxo not necessarily the tx itself. A `IncomingTransaction` shouldn't really be concerned about a `sriptPubKey`, `outputIdx` etc

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,11 +182,160 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+            xs.map { elem =>
+                s"${transaction.txIdBE.hex}:${elem.index}"
+              }
+              .mkString(", ")
+          }
+          logger.trace(
+            s"Found $count relevant output(s) in transaction=${transaction.txIdBE}: $outputStr")
+          if (xs.length > 1) {
+            logger.warn(
+              s"${xs.length} SPKs were relevant to transaction=${transaction.txIdBE}, but we aren't able to handle more than 1 for the time being")
```
Christewart, line 223: This needs a issue at minimum, it is not uncommon for people to re-use addresses (unfortunately) and this seems to break this.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,11 +182,160 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+        val withIndex =
+          transaction.outputs.zipWithIndex
+        withIndex.collect {
+          case (out, idx)
+              if addrs.map(_.scriptPubKey).contains(out.scriptPubKey) =>
+            OutputWithIndex(out, idx)
+        }
+      }
+
+      relevantOutsWithIdx match {
```
Christewart, line 205: Overally I think this should be moved into a function that processes each relevantOuts and returns a `Vector[Future[IncomingUtxo]]` and then sequences those Futures to return a single `Future[Vector[IncomingUtxo]]`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,11 +182,160 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+      updateF.foreach(tx =>
+        logger.debug(
+          s"Updated confirmation count=${tx.confirmations} of transaction=${tx.txid}"))
+      updateF.failed.foreach(err =>
+        logger.error(
+          s"Failed to update confirmation count of transaction=${transaction.txIdBE}",
+          err))
+
+      updateF
+    } else if (foundTx.confirmations > foundTx.confirmations) {
```
Christewart: This isn't satisfiable is it?

testkit/src/main/scala/org/bitcoins/testkit/fixtures/WalletDAOFixture.scala
```diff
@@ -0,0 +1,35 @@
...
+    WalletDAOs(account, address, inTx, outTx, utxo)
+  }
+
+  final override type FixtureParam = WalletDAOs
+
+  implicit private val walletConfig: WalletAppConfig = config
+
+  def withFixture(test: OneArgAsyncTest): FutureOutcome =
+    makeFixture(build = () => WalletDbManagement.createAll().map(_ => daos),
+                destroy = () => WalletDbManagement.dropAll())(test)
```
nkohen, line 34: I think it might be cleaner to decouple this from depending on `WalletDbManagement.allTables`, can we not pass in something like a `WalletDAOs.allTables` to be created so that we never have to worry about inconsistent states between the fixture and what is done by `WalletDbManagement`?

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -52,7 +52,8 @@ trait BitcoinSWalletTest
 
   def destroyWallet(wallet: UnlockedWalletApi): Future[Unit] = {
     WalletDbManagement
-      .dropAll()(config = config.walletConf, ec = implicitly[ExecutionContext])
+      .dropAll()(config = wallet.walletConfig,
+                 ec = implicitly[ExecutionContext])
```
nkohen, line 56: If this gets resolved by `ExecutionContext.global` I think I'd prefer being explicit

wallet-test/src/test/scala/org/bitcoins/wallet/models/UTXOSpendingInfoDAOTest.scala
```diff
@@ -5,45 +5,57 @@ import org.bitcoins.core.protocol.transaction.{
...
+      (tx, _) <- WalletTestUtil.insertIncomingTx(daos)
+      created <- {
+        val utxo = {
+          val outpoint =
+            TransactionOutPoint(WalletTestUtil.sampleTxid,
+                                WalletTestUtil.sampleVout)
+          val output = TransactionOutput(1.bitcoin, WalletTestUtil.sampleSPK)
+          val scriptWitness = WalletTestUtil.sampleScriptWitness
+          val privkeyPath = WalletTestUtil.sampleSegwitPath
+          NativeV0UTXOSpendingInfoDb(outPoint = outpoint,
```
nkohen: any reason not to just put these outside of the block they're in and have this be `val utxo`?

nkohen: same in other places, obvs a nit

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -26,16 +26,34 @@ import org.bitcoins.wallet.ReadMnemonicError.JsonParsingError
...
 
   private[wallet] val addressDAO: AddressDAO = AddressDAO()
   private[wallet] val accountDAO: AccountDAO = AccountDAO()
   private[wallet] val utxoDAO: UTXOSpendingInfoDAO = UTXOSpendingInfoDAO()
+  private[wallet] val incomingTxDAO = IncomingTransactionDAO(SQLiteProfile)
+  private[wallet] val outgoingTxDAO = OutgoingTransactionDAO(SQLiteProfile)
+
+  /** Sums up the value of all incoming
+    * TXs in the wallet, filtered by the given predicate */
+  // TODO account for outgoing TXs
```
nkohen, line 42: issue for this?

wallet/src/main/scala/org/bitcoins/wallet/models/TransactionTable.scala
```diff
@@ -0,0 +1,106 @@
...
+
+}
+
+final case class IncomingTransactionTable(tag: Tag)
+    extends TransactionTable[IncomingTransaction](tag, "incoming_transactions") {
+  import org.bitcoins.db.DbCommonsColumnMappers._
+
+  // TODO: What happens if we get paid to multiple SPKs in the same
+  // transaction? Need to make a table of SPKs, and map IDs in that
+  // table to TXs in this table...
```
nkohen: issue

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
torkelrogstad: This doesn't resolve to `ExecutionContext.global` unless that is explicitly imported. 

For all def with implicit params:

```scala
def foo(bar: Int)(implicit s: Something): Unit = ???
```

All of the below snippets are equivalent: 

```scala
implicit val s: Something = ???
foo(2) // implicitly passed
```

```scala
val s: Something = ???
foo(2)(s) // explicitly passed
```

```scala
implicit val s: Something = ???
val otherS: Something = implicitly[Something] // resolves to s above
foo(2)(otherS)
```

```scala
implicit  val s: Something = ???
foo(2)(implicitly[Something] // this resolves to s above
```

There's no point in the last snippet (because `foo` only takes one implicit param). However on the line you commented on, there's two implicit params. So we're basically saying: "use this specific wallet config, but conjure up the `ExecutionContext` following normal implicit resolution rules"

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: Added it to the umbrella issue: https://github.com/bitcoin-s/bitcoin-s/issues/530

testkit/src/main/scala/org/bitcoins/testkit/fixtures/WalletDAOFixture.scala
torkelrogstad: Sorry, I don't get what you mean. Won't we have to have some sort of list of all tables regardless of what we end up doing? That's what `createAll` does, iterating over `allTables` and creating them

node-test/src/test/scala/org/bitcoins/node/NodeWithWalletTest.scala
torkelrogstad: Nice catch! You're right that this test was not falsifiable. This was because `assert` doesn't return anything if passed `false`, but throws a `TestFailedException`. The `.success(...)` method was therefore never evaluated. 

`completetionP.complete { Try { ... } } ` solved it!

Re. the syntax question: A `for` expression without a `yield` does a `foreach` instead on the final value instead of `map`. We don't use the end result for anything except for a side effect, so IMO not having a `yield` communicates intent better.

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
```diff
@@ -0,0 +1,23 @@
...
+  * aid in testing.
+  */
+object Implicits {
+
+  /** Extension methods for Scalacheck generatos */
+  implicit class GeneratorOps[T](private val gen: Gen[T]) extends AnyVal {
+
+    @tailrec
+    /** Gets a sample from this generator that's not `None` */
+    def sampleSome: T = gen.sample match {
```
Christewart: Just a note, this can really slow down tests if there is a complex generator being used

testkit/src/main/scala/org/bitcoins/testkit/core/gen/TransactionGenerators.scala
```diff
@@ -132,14 +147,23 @@ trait TransactionGenerators extends BitcoinSLogger {
       lockTime <- NumberGenerator.uInt32s
     } yield BaseTransaction(version, is, os, lockTime)
 
-  /** Generates a random [[org.bitcoins.core.protocol.transaction.WitnessTransaction WitnessTransaction]] */
-  def witnessTransaction: Gen[WitnessTransaction] =
+  /** Generates a legacy transaction with at least one output paying to the given SPK */
+  def baseTransactionTo(spk: ScriptPubKey) =
```
Christewart: nit: Add return type parameter

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
Christewart: I don't really understand what the benefit is of the extra complexity rather than just adding a implicit param on the `destroyWallet()` method. KISS

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
Christewart: This seems like the worse of both worlds, the type `Gen` is around for a reason, is there a reason why we aren't doing things properly and just writing a property based test instead of unit tests with one random generated value?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,10 +182,186 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+      val msg =
+        List(
+          s"Incoming transaction=${transaction.txIdBE} has fewer confirmations=$confirmations",
+          s"than what we already have registered=${foundTxo.confirmations}! I don't know how",
+          s"to handle this."
+        ).mkString(" ")
+      logger.warn(msg)
+      Future.failed(new RuntimeException(msg))
+    } else {
+      logger.debug(
```
Christewart, line 324: This could be bad in the case of competing re-orgs. Do we really want this to be our behavior?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,10 +182,186 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+      confirmations: Int): Future[LockedWallet] = {
+    logger.info(
+      s"Processing transaction=${transaction.txIdBE} with confirmations=$confirmations")
+
+    val incomingTxoFut: Future[Vector[IncomingWalletTXO]] =
+      incomingTxoDAO
+        .findTx(transaction)
+        .flatMap {
+          // no existing elements found
+          case Vector() =>
```
Christewart, line 341: `Vector.empty`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,10 +182,186 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+            processNewIncomingTx(transaction, confirmations)
+
+          case txos: Vector[IncomingWalletTXO] =>
+            Future
+              .sequence {
+                txos
```
Christewart: This is pretty hairy

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: `Vector.empty` is a `def` so it cannot be used in pattern-matching :-(

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
torkelrogstad: We want to control precisely which wallet is being destroyed (which is achieved here by making sure we're operating on the correct datadir). If we're adding an implicit parameter to the method we're not getting any guarantees about what wallet we're destroying because it's just grabbing whatever implicit configuration is closest in scope.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: Reorg isn't handled any where in the wallet currently. I can add a comment here, saying that this will need to change once reorg handling is implemented.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,10 +182,186 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+              .mkString(", ")
+          }
+          logger.trace(
+            s"Found $count relevant output(s) in transaction=${transaction.txIdBE}: $outputStr")
+          if (xs.length > 1) {
+            logger.warn(
+              s"${xs.length} SPKs were relevant to transaction=${transaction.txIdBE}, but we aren't able to handle more than 1 for the time being")
+          }
+
+          def processUtxo(index: Int): Future[SpendingInfoDb] =
```
Christewart: I find methods like these very hard to read when they are nested inside of a very large method _and_ they don't use `{}` braces explicitly. The same thing applies with `createIncomingTxo`

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
torkelrogstad: This is meant as a better alternative than `FooGenerator.sample.get`. In many tests in the code base we don't want to have a full blown property-based test (for example while doing wallet tests, running the same test 10 times with marginally different inputs would slow things down _a lot_), but we need testing data that is actually meaningful (like for example a transaction paying to a specific SPK)

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -161,10 +182,186 @@ abstract class LockedWallet extends LockedWalletApi with BitcoinSLogger {
...
+    */
+  private def processExistingIncomingTxo(
+      transaction: Transaction,
+      confirmations: Int,
+      foundTxo: IncomingWalletTXO): Future[Option[IncomingWalletTXO]] = {
+    if (foundTxo.confirmations < confirmations) {
+      logger.debug(
+        s"Increasing confirmation count of txo=${transaction.txIdBE}, old=${foundTxo.confirmations} new=${confirmations}")
+      val updateF =
+        incomingTxoDAO.update(foundTxo.copy(confirmations = confirmations))
```
Christewart, line 301: I don't think we should be making the assumption that this is still in the same transaction. Someone could be trying to double spend. If I understand this code correctly, we are assuming everything else is still the same about the tx and it is a simple reorg?

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
Christewart: This can also slow down the code significantly because it is guaranteed that we generate a value. This can essentially end up in an infinite loop trying to satisfy the generator. That is why the `Option` type is returned.

Christewart: I think we should just cache these values in a `*TestUtil.scala` file if you don't want to do the full blow property based testing.

torkelrogstad: If we have a generator that leads to a significant slowdown/infinite loop I'd say that's a problem on its own, as that's going to make any tests using that generator equally slow. The solution in that case is fixing the generator. 

For a bunch of things, having a `"TestUtil` object works fine, but what about for example an address that pays to a specific SPK, and that SPK is generated by the wallet in a test? I'm thinking `sampleSome` works fine in that case. 

If you'd like, I could add a constraint to the method that it's recursing maximum 10 levels deep, is that something you could live with?

Christewart: If you feel strongly about it I will let it go, but I think this is a classic example of circumventing the type system safety to be more convenient. Circumventing the type system is more acceptable in test code, but I do think it will bite us in the long term since the library designer designed the API like that for a reason.

PR #535, opened 2019-06-18
[WIP] New Eclair RPC client

rorp:


core-test/src/test/scala/org/bitcoins/core/protocol/ln/ShortChannelIdTest.scala
```diff
@@ -0,0 +1,24 @@
...
+    // min value
+    ShortChannelId.fromHumanReadableString("0x0x0") must be (ShortChannelId.fromHex("0"))
+    ShortChannelId.fromHex("0").toHumanReadableString must be ("0x0x0")
+
+    // max value
+    ShortChannelId.fromHumanReadableString("16777215x16777215x65535") must be (ShortChannelId.fromHex("ffffffffffffffff"))
+    ShortChannelId.fromHex("ffffffffffffffff").toHumanReadableString must be ("16777215x16777215x65535")
+
+    // overflow value
+    ShortChannelId.fromHumanReadableString("16777216x16777216x65536") must be (ShortChannelId.fromHex("0"))
```
Christewart: Is this defined in the spec? Would it be better to throw in this case?

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,35 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
+
+  /**
+    * Output example:
+    * {{{
+    * > ShortChannelId.fromHex("db0000010000")
+    * 219x1x0
+    * }}}
+    */
+  def toHumanReadableString: String = {
+    val (blockHeight, txIndex, outputIndex) =  ((u64 >> 40) & UInt64(0xFFFFFF), (u64 >> 16) & UInt64(0xFFFFFF), u64 & UInt64(0xFFFF))
```
Christewart: nit: 

This is more readable for me 

```scala
    val blockHeight = (u64 >> 40) & UInt64(0xFFFFFF)
    val txIndex = (u64 >> 16) & UInt64(0xFFFFFF)
    val outputIndex = u64 & UInt64(0xFFFF)
```

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,35 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
 
   override def fromBytes(byteVector: ByteVector): ShortChannelId = {
     new ShortChannelId(UInt64.fromBytes(byteVector))
   }
+
+  def fromHumanReadableString(str: String): ShortChannelId = str.split("x") match {
+    case Array(blockHeight, txIndex, outputIndex) =>
+      val u64 = UInt64(((BigInt(blockHeight) & 0xFFFFFFL) << 40) | ((txIndex.toInt & 0xFFFFFFL) << 16) | (outputIndex.toInt & 0xFFFFL))
+      ShortChannelId(u64)
+    case _ => fromHex(str)
```
Christewart: I don't like to use `case _ =>` which breaks pattern matching exhaustiveness, do 
```scala 
case _: Array[String] => ...

```

Also I think we should throw in this case? Since the human readable string must have 3 values and it is delimited by `x`?

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,35 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
+
 }
 
 object ShortChannelId extends Factory[ShortChannelId] {
 
   override def fromBytes(byteVector: ByteVector): ShortChannelId = {
     new ShortChannelId(UInt64.fromBytes(byteVector))
   }
+
+  def fromHumanReadableString(str: String): ShortChannelId = str.split("x") match {
```
Christewart, line 45: scala doc explaining what this is, and linking to the BOLT that defines this format.

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,35 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
+    case Array(blockHeight, txIndex, outputIndex) =>
+      val u64 = UInt64(((BigInt(blockHeight) & 0xFFFFFFL) << 40) | ((txIndex.toInt & 0xFFFFFFL) << 16) | (outputIndex.toInt & 0xFFFFL))
```
Christewart: break these out into intermediate vals to make it more readable

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +114,16 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
 
   behavior of "RpcClient"
 
+  it should "be able to create an invoice" in {
+    for {
+      c <- clientF
+      invoice <- c.createInvoice("test", None, None, None, None)
+    } yield {
+      println(invoice)
```
Christewart: I would mixin the [`BitcoinSLogger`](https://github.com/bitcoin-s/bitcoin-s/blob/839d520206023fa1075a65e19392d9f9b72e3a62/core/src/main/scala/org/bitcoins/core/util/BitcoinSLogger.scala) trait and use `logger` rather than `println`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -98,6 +103,8 @@ trait EclairApi {
     getNodeURI.map(_.nodeId)
   }
 
+  def createInvoice(description: String, amountMsat: Option[MilliSatoshis], expireIn: Option[Long], fallbackAddress: Option[Address], paymentPreimage: Option[String]): Future[LnInvoice]
```
Christewart: I think `expireIn` should be of type `Option[FiniteDuration]`, and I think we should create a new type called `PaymentPreimage` in the `core/` project to represent a pre-image. It probably should be located here: 

https://github.com/bitcoin-s/bitcoin-s/tree/master/core/src/main/scala/org/bitcoins/core/protocol/ln

and I think it should be

```scala

/** Scaladoc blah blah blah... */ 
case class PaymentPreimage(bytes: ByteVector) extends NetworkElement { 
//can we put any invariants here, like does `bytes` need to be 32 bytes in size? 

lazy val hash: Sha256Digest = CryptoUtil.sha256Digest(bytes)
}

object PaymentPreimage extends Factory[PaymentPreimage] { 
  override def fromBytes(bytes: ByteVector): PaymentPreimage = { 
    new PaymentPreimage(bytes)  
  }

}

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala [hunk 8]
```diff
@@ -131,29 +129,45 @@ class EclairRpcClient(val instance: EclairInstance)(
...
     val uri = NodeUri(nodeId, host, port)
     connect(uri)
   }
 
   override def connect(uri: NodeUri): Future[String] = {
-    eclairCall[String]("connect", List(JsString(uri.toString)))
+    eclairCallNew[String]("connect", "uri" -> uri.toString)
+  }
+
+  def createInvoice(description: String, amountMsat: Option[MilliSatoshis], expireIn: Option[Long], fallbackAddress: Option[Address], paymentPreimage: Option[String]): Future[LnInvoice] = {
```
Christewart: override

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -131,29 +129,45 @@ class EclairRpcClient(val instance: EclairInstance)(
...
+    val params = Seq(
+      Some("description" -> description),
+      amountMsat.map(x => "amountMsat" -> x.toBigDecimal.toString),
+      expireIn.map(x => "expireIn" -> x.toString),
+      fallbackAddress.map(x => "fallbackAddress" -> x.toString),
+      paymentPreimage.map(x => "paymentPreimage" -> x)
+    ).flatten
+    eclairCallNew[CreateInvoiceResult]("createinvoice", params: _*)
```
Christewart:
```scala
    val responseF = eclairCallNew[CreateInvoiceResult]("createinvoice", params: _*)
    
    responseF.flatMap(res => Future.fromTry(LnInvoice.fromString(res.serialized)))
```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -246,10 +237,22 @@ case class ChannelResult(
...
+    data: JsObject) {
+  import JsonReaders._
+  def shortChannelId: ShortChannelId = (data \ "shortChannelId").validate[ShortChannelId].get
+}
 
 // ChannelResult ends here
 
+case class CreateInvoiceResult(
+  prefix: String,
+  timestamp: Long,
```
Christewart: can we add a better type to this?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -246,10 +237,22 @@ case class ChannelResult(
...
-    data: JsObject)
+    data: JsObject) {
+  import JsonReaders._
+  def shortChannelId: ShortChannelId = (data \ "shortChannelId").validate[ShortChannelId].get
+}
 
 // ChannelResult ends here
 
+case class CreateInvoiceResult(
+  prefix: String,
```
Christewart: This can be typed a [`LnHumanReadablePart`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala) i think

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -246,10 +237,22 @@ case class ChannelResult(
...
+  timestamp: Long,
+  nodeId: NodeId,
+  serialized: String,
+  description: String,
+  paymentHash: Sha256Digest,
+  expiry: Long)
```
Christewart: better type?

Christewart (commented):
In general, it would be nice if we could at least assert something about the `allUpdates`, `allChannels` etc rpc calls rather than just making sure they didn't error

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
Christewart: You should be able to tinker with logging levels in the [`common-logback.xml`](https://github.com/bitcoin-s/bitcoin-s/blob/796492a6781ee285fdbae17f207aae0f432e42f0/core/src/test/resources/common-logback.xml) file

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -53,13 +54,17 @@ trait EclairApi {
...
+
   def close(id: ChannelId, spk: ScriptPubKey): Future[String]
 
-  def findRoute(nodeId: NodeId): Future[Vector[NodeId]]
+  def findRoute(nodeId: NodeId, amountMsat: MilliSatoshis): Future[Vector[NodeId]]
+
+  def findRoute(invoice: LnInvoice, amountMsat: Option[MilliSatoshis]): Future[Vector[NodeId]]
 
-  def findRoute(invoice: LnInvoice): Future[Vector[NodeId]]
+  def forceClose(channelId: ChannelId): Future[String]
```
torkelrogstad: Return type here would be better as `Future[DoubleSha256DigestBE]`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -53,13 +54,17 @@ trait EclairApi {
...
 
-  def forceClose(id: ChannelId): Future[String]
+  def forceClose(shortChannelId: ShortChannelId): Future[String]
```
torkelrogstad: Same comment as above

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    }
+  }
+
+  it should "update the relay fee of a channel with short channel id" in {
+    val channelAndFeeF = for {
+      channelId <- openAndConfirmChannel(clientF, otherClientF)
+      client <- clientF
+      channel <- client.channel(channelId)
+    } yield {
+      assert(channel.feeBaseMsat.isDefined)
```
torkelrogstad, line 798: Could do `assert(channel.feeBaseMsat.exists(_ > MilliSatoshis.zero)`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      assert(channel.feeBaseMsat.get > MilliSatoshis.zero)
+      (channel.channelId, channel.shortChannelId, channel.feeBaseMsat.get)
+    }
+
+    for {
+      (channelId, shortChannelId, oldFee) <- channelAndFeeF
+      _ <- clientF.flatMap(
+        _.updateRelayFee(shortChannelId, MilliSatoshis(oldFee.toLong * 4), 1))
```
torkelrogstad: Can just do `oldFee * 4`, instead of `(shortChannelId, MilliSatoshis(oldFee.toLong * 4)`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+  }
+
+  it should "update the relay fee of a channel with short channel id" in {
+    val channelAndFeeF = for {
+      channelId <- openAndConfirmChannel(clientF, otherClientF)
+      client <- clientF
+      channel <- client.channel(channelId)
+    } yield {
+      assert(channel.feeBaseMsat.isDefined)
+      assert(channel.feeBaseMsat.get > MilliSatoshis.zero)
```
torkelrogstad, line 799: Also, this is pretty nitty, but I think this reads really well: 
```scala
 val feeBaseMsat = channel.feeBaseMsat match {
        case None => fail(s"feeBaseMsat was undefined")
        case Some(fee) =>
          assert(fee > 0.msat)
          fee
      }
      (channel.channelId, channel.shortChannelId, feeBaseMsat)
```

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    }
+
+    for {
+      (channelId, shortChannelId, oldFee) <- channelAndFeeF
+      _ <- clientF.flatMap(
+        _.updateRelayFee(shortChannelId, MilliSatoshis(oldFee.toLong * 4), 1))
+      newFeeOpt <- clientF.flatMap(_.channel(channelId).map(_.feeBaseMsat))
+    } yield {
+      assert(newFeeOpt.isDefined)
+      assert(newFeeOpt.get == MilliSatoshis(oldFee.toLong * 4))
```
torkelrogstad: Could do `oldFee * 4` here as well

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    })
+  }
+
+  it should "get updates for a single node" in {
+    for {
+      client <- clientF
+      nodeInfo <- client.getInfo
+      _ <- client.allUpdates(nodeInfo.nodeId)
+    } yield {
+      succeed
```
torkelrogstad: Could we assert that the result of `allUpdates` is not empty?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+
+  it should "get all nodes" in {
+    clientF.flatMap(_.allNodes().flatMap(_ => succeed))
+  }
+
+  it should "get a route to a node ID" in {
+    val hasRoute = () => {
+      fourthClientF
+        .flatMap(_.getInfo)
+        .flatMap(info => firstClientF.flatMap(_.findRoute(info.nodeId, MilliSatoshis(100))))
```
torkelrogstad: So this is not a big deal at all, just thought I should say it so you're aware of it: It's possible to write money like this:

```scala
100.msat == MilliSatoshis(100)
```

The extension method is provided by the package object in `org.bitcoins.core.protocol.ln.currency`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +679,120 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+        .map(route => route.length == 4)
+        .recover {
+          case err: RuntimeException
+              if err.getMessage.contains("route not found") =>
+            println(err)
```
torkelrogstad: `logger.debug` instead of `println`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -442,25 +458,49 @@ class EclairRpcClient(val instance: EclairInstance)(
...
+    eclairCallNew[Unit]("updaterelayfee",
+      "shortChannelId" -> shortChannelId.toHumanReadableString,
+      "feeBaseMsat" -> feeBaseMsat.toLong.toString,
+      "feeProportionalMillionths" -> feeProportionalMillionths.toString)
   }
 
   // TODO: channelstats, audit, networkfees?
   // TODO: Add types
 
+  private def eclairCallNew[T](command: String, parameters: (String, String)*)(
```
torkelrogstad: So one thing we could do here to make this a bit safer, is replace the type signature here and add an implicit conversion: 

```scala
 import scala.language.implicitConversions
  implicit private def obj2Json[T](obj: T)(
      implicit writes: Writes[T]): JsValue = Json.toJson(obj)

  private def eclairCallNew[T](command: String, parameters: (String, JsValue)*)
```

Then in the rest of the code, instead of passing in strings that are serialized ad hoc we could use all the JSON serializers we have defined already in Bitcoin-S. 

If you add this import at the top:

```
import org.bitcoins.rpc.serializers.JsonSerializers._
```

You get serialization to and from JSON for most data structures for free:-)

rorp: New eclair API doesn't use JSON for encoding API parameters anymore, it uses plain HTTP forms instead.

torkelrogstad: Hm. Still, we'd like to avoid ad-hoc serialization of things in each RPC call. For example if doing `.toString` on an address, I'm pretty sure it'll look like `P2SHBitcoindAddress(actualAddressHere)` instead of just `actualAddressHere`.

The JSON serializers we have already handle correct conversion of data structures to string representation, so it'd be nice to use those somehow

rorp: In order to create an HTTP form in akka-http I need a Seq[(String, String)]. Do you want me to create Seq[(String, JValue]) using the implictis and than to convert the seq into Seq[(String, String)] somehow?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
rorp: Maybe. Which one? All other Eclair models use Long for timestamps.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
rorp: Again, which one? It's always easy to create any type of date/time object from a Long.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
rorp: Why?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
rorp: I this case you can't tell if it failed because there's no data or the data is invalid

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
rorp: It does look like an assert to me. It'll take some time to figure out what it does, when I need to change it.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
Christewart: I'm not picky, but just something that is more structured than `Long`. As you heard this morning during standup we have used `Long` in the past and been burned.

Christewart: I guess is the timestamp in milliseconds or seconds?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
Christewart: So the interesting thing about this field is i believe it is in seconds not milliseconds.

https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields

So it is sort of a foot gun for people using the api if we just pass around `Long` instead of something more structured that captures the fact that this field is in seconds. My go to for things like this is `FiniteDuration` but can be convinced otherwise

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
torkelrogstad: I think that's a good idea. For example here, you can see the JSON writer for addresses: https://github.com/bitcoin-s/bitcoin-s/blob/839d520206023fa1075a65e19392d9f9b72e3a62/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala#L42-L44

As you can see, its calling `.value`, instead of `.toString`. IMO its cleaner to have string serialization happen in one place (the JSON writers) instead of ad hoc in each RPC call.

But again, feel free to disagree on this! I'm open to it being a bad idea:-)

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
torkelrogstad: I believe it returns the TXID of the channel closing transaction

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentId.scala
```diff
@@ -0,0 +1,10 @@
+package org.bitcoins.core.protocol.ln
+
+import java.util.UUID
+
+case class PaymentId(value: String) {
```
torkelrogstad: I'm thinking it would make sense to have a `require` statement here, that checks that the given string is a valid UUID

torkelrogstad: Could make `toUUID` a lazy val and just verify that it doesn't throw

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentPreimage.scala
```diff
@@ -0,0 +1,23 @@
...
+/**
+  * Payment preimage for generating LN invoices.
+  */
+sealed case class PaymentPreimage(bytes: ByteVector) extends NetworkElement {
+  require(bytes.size == 32, s"Payment preimage size must be 32 bytes")
+
+  lazy val hash: Sha256Digest = CryptoUtil.sha256(bytes)
+}
+
+object PaymentPreimage extends Factory[PaymentPreimage] {
```
torkelrogstad, line 17: I'm thinking it'd be nice to have a method on the companion object for generating random preimages. Either an empty `apply` method or something like `PaymentPreimage.random()`

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,48 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
     * > ShortChannelId.fromHex("db0000010000")
-    * ShortChannelId(db0000010000)
+    * 219x1x0
     * }}}
     */
-  override def toString: String = s"ShortChannelId(${hex.drop(4)})"
+  override def toString: String = toHumanReadableString
+
+  /**
+    * Converts the short channel id into the human readable form defined in https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md
```
torkelrogstad: If you add this line in the Scaladoc, it ends up formatting nicely: 

```
* @see [[https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md  BOLT7]]
```

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    } yield {
+      assert(sent.sent.nonEmpty)
+      assert(received.received.nonEmpty)
+      assert(relayed.relayed.nonEmpty)
+    }
+  }
+
+  it should "be able to send payments in both directions" ignore {
+    val paymentAmount = MilliSatoshis(50)
+    //    val paymentAmount = NanoBitcoins(100000000)
```
torkelrogstad: Delete this line?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+          succeed
+        }
+    }
+
+    executeWithClientOtherClient(sendInBothDiercions)
+  }
+
+  it should "check a payment" in {
+    val paymentAmount = MilliSatoshis(50)
+    //    val paymentAmount = NanoBitcoins(100000)
```
torkelrogstad: Delete line?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      c3 <- thirdClientF
+      c4 <- fourthClientF
+      info <- c.getInfo
+      info1 <- c1.getInfo
+      info2 <- c2.getInfo
+      info3 <- c3.getInfo
+      info4 <- c4.getInfo
+      invoice <- c.createInvoice("test", None, None, None, None)
+    } yield {
+//      assert(invoice.nodeId == info.nodeId)
```
torkelrogstad: Does this fail?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    }
+  }
+
+  it should "be able to create an invoice with amount" in {
+    for {
+      c <- clientF
+      info <- c.getInfo
+      invoice <- c.createInvoice("test", Some(12345.msats.toLnCurrencyUnit), None, None, None)
+    } yield {
+//      assert(invoice.nodeId == info.nodeId)
```
torkelrogstad: Same about this line

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      assert(invoice.timestamp > UInt64(1561063731)) // this is when I wrote this code
+      assert(invoice.timestamp <= UInt64(System.currentTimeMillis() / 1000))
+    }
+  }
+
+  it should "be able to create an invoice with amount" in {
+    for {
+      c <- clientF
+      info <- c.getInfo
+      invoice <- c.createInvoice("test", Some(12345.msats.toLnCurrencyUnit), None, None, None)
```
torkelrogstad: Not a big deal, but just so you know it: LN currency units also have syntax extensions. So you can do `123.pBTC`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      assert(invoice.timestamp > UInt64(1561063731)) // this is when I wrote this code
+      assert(invoice.timestamp <= UInt64(System.currentTimeMillis() / 1000))
+    }
+  }
+
+  it should "be able to create an invoice with amount and expiry time" in {
+    for {
+      c <- clientF
+      info <- c.getInfo
+      invoice <- c.createInvoice("test", Some(12345.msats.toLnCurrencyUnit), Some(FiniteDuration(67890, TimeUnit.SECONDS)), None, None)
```
torkelrogstad: Could we write `67890.seconds`? Makes it much easier to understand. The syntax is exposed in `scala.concurrent.duration.DurationInt`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +115,169 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    } yield {
+//      assert(invoice.nodeId == info.nodeId)
```
torkelrogstad: What about this line?

torkelrogstad: Could probably also extract these assertion lines to a method, seeing as they are copy pasted several times

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +756,118 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    } yield {
+      assert(feeOpt.isDefined)
+      assert(feeOpt.get > MilliSatoshis.zero)
+      (channel, feeOpt.get)
+    }
+
+    for {
+      (channel, oldFee) <- channelAndFeeF
+      _ <- clientF.flatMap(
+        _.updateRelayFee(channel, MilliSatoshis(oldFee.toLong * 2), 1))
```
torkelrogstad: `oldFee * 2`?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +756,118 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+      newFeeOpt <- clientF.flatMap(_.channel(channel).map(_.feeBaseMsat))
+    } yield {
+      assert(newFeeOpt.isDefined)
+      assert(newFeeOpt.get == MilliSatoshis(oldFee.toLong * 2))
```
torkelrogstad: `oldFee * 2`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -46,20 +48,19 @@ trait EclairApi {
...
 
+  def connect(nodeId: NodeId, host: String, port: Int): Future[String]
+
   def close(id: ChannelId, spk: ScriptPubKey): Future[String]
 
-  def findRoute(nodeId: NodeId): Future[Vector[NodeId]]
+  def findRoute(nodeId: NodeId, amountMsat: MilliSatoshis): Future[Vector[NodeId]]
 
-  def findRoute(invoice: LnInvoice): Future[Vector[NodeId]]
+  def findRoute(invoice: LnInvoice, amountMsat: Option[MilliSatoshis]): Future[Vector[NodeId]]
```
torkelrogstad: Seeing as there's a method with the `amountMsat` argument as non-optional, what's the point of having this method?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -81,7 +75,6 @@ case class NodeInfo(
     nodeId: NodeId,
     rgbColor: String,
     alias: String,
-    shortChannelId: ShortChannelId,
     addresses: Vector[String])
```
torkelrogstad, line 79: Could parse these into a `java.net.URL`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -1,30 +1,24 @@
...
 sealed abstract class EclairModels
 
 case class GetInfoResult(
     nodeId: NodeId,
     alias: String,
-    port: Int,
     chainHash: DoubleSha256Digest,
-    blockHeight: Long)
+    blockHeight: Long,
+    publicAddresses: Seq[String])
```
torkelrogstad, line 22: Could parse these into a `java.net.URL`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -258,16 +263,48 @@ case class PaymentRequest(
...
+  createdAt: FiniteDuration,
+  completedAt: Option[FiniteDuration],
+  status: PaymentStatus)
+
+case class ReceivedPaymentResult(
+  paymentHash: Sha256Digest,
+  amountMsat: MilliSatoshis,
+  receivedAt: FiniteDuration)
+
+//object PaymentStatus extends Enumeration {
```
torkelrogstad: Delete?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -258,16 +263,48 @@ case class PaymentRequest(
...
+sealed trait PaymentStatus
+object PaymentStatus {
+  case object PENDING extends PaymentStatus
+  case object SUCCEEDED extends PaymentStatus
+  case object FAILED extends PaymentStatus
+
+  def apply(s: String): PaymentStatus = s match {
+    case "PENDING" => PENDING
+    case "SUCCEEDED" => SUCCEEDED
+    case "FAILED" => FAILED
```
torkelrogstad, line 313: Add a catch-all case at the end that gives a better error message than `MatchError`?

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentPreimage.scala
Christewart: And this can just be a wrapper around `ECPrivateKey.freshPrivateKey()`

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -742,6 +756,118 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
 
   }
 
+  it should "get a route to an invoice" in {
+    val hasRoute = () => {
+      fourthClientF.flatMap { fourthClient =>
+        fourthClient
+          .createInvoice("foo", Some(100.msats.toLnCurrencyUnit), None, None, None)
```
Christewart: named arguments really help readable with param lists like this

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/JsonReaders.scala
```diff
@@ -102,6 +107,28 @@ object JsonReaders {
...
+        paymentHash <- (jsValue \ "paymentHash").validate[Sha256Digest]
+        expiry <- (jsValue \ "expiry").validate[Long]
+      } yield
+        InvoiceResult(
+          prefix,
+          FiniteDuration(timestamp, TimeUnit.SECONDS),
+          nodeId,
+          serialized,
+          description,
+          paymentHash: Sha256Digest,
```
Christewart: this seems weird, this is a type annotation?

Christewart:
Looks like we need to setup the new jar for eclair for CI to find as well 

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/549965942#L1257

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
rorp: java.net.URL can't parse onion addresses

rorp: java.net.URL can't parse onion addresses

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
rorp: There's a test case to make sure that the addresses get serialized in the way Eclair can parse them.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
rorp: For some reason it returns "ok": https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1622

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
nkohen: I'm assuming you mean that this one should just be `findRoute(invoice: LnInvoice)` right @torkelrogstad ?

torkelrogstad: yes

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,48 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
 
 object ShortChannelId extends Factory[ShortChannelId] {
 
   override def fromBytes(byteVector: ByteVector): ShortChannelId = {
     new ShortChannelId(UInt64.fromBytes(byteVector))
   }
+
+  def fromHumanReadableString(str: String): ShortChannelId = str.split("x") match {
+    case Array(_blockHeight, _txIndex, _outputIndex) =>
+      val blockHeight = BigInt(_blockHeight)
```
Christewart, line 47: nit: I would add the invalid number to all of these things, for instance

```scala
require(blockHeight >= 0 && blockHeight <= 0xffffff, "ShortChannelId: invalid block height=${blockHeight}")
```

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
```diff
@@ -12,15 +12,48 @@ case class ShortChannelId(u64: UInt64) extends NetworkElement {
...
+      require(blockHeight >= 0 && blockHeight <= 0xffffff, "ShortChannelId: invalid block height")
+
+      val txIndex = _txIndex.toInt
+      require(txIndex >= 0 && txIndex <= 0xffffff, "ShortChannelId:invalid tx index")
+
+      val outputIndex = _outputIndex.toInt
+      require(outputIndex >= 0 && outputIndex <= 0xffff, "ShortChannelId: invalid output index")
+
+      val u64 = UInt64(((blockHeight & 0xffffffL) << 40) | ((txIndex & 0xffffffL) << 16) | (outputIndex & 0xffffL))
```
Christewart, line 56: nit: Can we make this cleaner? It is a pretty long nasty line of code

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -85,6 +86,29 @@ case class AuditResult(
     received: Vector[ReceivedPayment]
 )
 
+case class NetworkFeesResult(
+    remoteNodeId: NodeId,
+    channelId: FundedChannelId,
+    txId: String,
```
torkelrogstad: Better type would be `DoubleShaDigest256BE`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -317,3 +341,40 @@ implicit val sendResultReads: Reads[SendResult] = Reads[SendResult] { json =>
     }
   }
 }*/
+
+object WebSocketEvents {
+
+  sealed trait Event
```
torkelrogstad: For readability, I'd say it makes more sense to move the trait outside of `WebSocketEvents`, and rename it to match it. 

I.e.

```scala
sealed trait WebSocketEvents
object WebSocketEvents {
  case class PaymentRelayed(
    ...
```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -48,11 +48,15 @@ trait EclairApi {
...
 
-  def connect(nodeId: NodeId, host: String, port: Int): Future[String]
+  def connect(nodeURI: NodeUri): Future[Unit]
 
-  def close(id: ChannelId, spk: ScriptPubKey): Future[String]
+  def connect(nodeId: NodeId, host: String, port: Int): Future[Unit]
+
+  def disconnect(nodeId: NodeId): Future[Unit]
+
+  def close(id: ChannelId, spk: ScriptPubKey): Future[Unit]
```
torkelrogstad, line 64: I'm pretty sure `close` returns the transaction ID of the closing transaction, which would be `DoubleSha256DigtestBE`

rorp: "ok" is not valid TxId

Christewart:
@rorp Looks like there is a compiler issue on 2.11.x https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/553041828#L605

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentId.scala
```diff
@@ -0,0 +1,7 @@
+package org.bitcoins.core.protocol.ln
+
+
+case class PaymentId(value: String) {
```
Christewart: What exactly is this? Is this the same thing as payment hash? If it is something _not_ native to the Lightning protocol -- and is created by eclair -- it should live in the `eclair-rpc` project.

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentPreimage.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import org.bitcoins.core.crypto.{ECPrivateKey, Sha256Digest}
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.{CryptoUtil, Factory}
+import scodec.bits.ByteVector
+
+/**
+  * Payment preimage for generating LN invoices.
+  */
+sealed case class PaymentPreimage(bytes: ByteVector) extends NetworkElement {
```
Christewart: I think we probably want to mark this as `final` rather than `sealed` https://stackoverflow.com/a/31133245/967713

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -113,6 +116,177 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+    } yield {
+      assert(sent.sent.nonEmpty)
+      assert(received.received.nonEmpty)
+      assert(relayed.relayed.nonEmpty)
+    }
+  }
+
+  it should "be able to send payments in both directions" ignore {
+    val paymentAmount = MilliSatoshis(100000000)
+    val paymentAmount2 = MilliSatoshis(100000000)
```
Christewart: nit: maybe make this two different amounts for more certainty that the actual amounts transacted are correct? This might just be paranoia on my part so fee free to ignore

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -267,7 +440,44 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
...
+          paymentId <- client.sendToRoute(route, amt, invoice.lnTags.paymentHash.hash, 10000)
+          _ <- EclairRpcTestUtil.awaitUntilPaymentSucceeded(client, paymentId)
+          succeeded <- client.getSentInfo(invoice.lnTags.paymentHash.hash)
+          _ <- client.close(channelId)
+          _ <- bitcoindRpcClientF.flatMap(_.generate(6))
+        } yield {
+          assert(succeeded.nonEmpty)
+
+          val succeededPayment = succeeded.head
+          assert(succeededPayment.status == PaymentStatus.SUCCEEDED)
```
Christewart, line 428: is there any way we can actually verify that the payment was sent along the expected route?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -794,9 +930,77 @@ class EclairRpcClientTest extends AsyncFlatSpec with BeforeAndAfterAll {
     clientF.map(c => assert(c.network == LnBitcoinRegTest))
   }
 
+  it should "force close a channel" in {
+    for {
+      c <- clientF
+      channels <- c.channels()
+      _ <- c.forceClose(channels.head.channelId)
+    } yield succeed
```
Christewart: Can we assert that the channel state is CLOSING here?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -1,16 +1,18 @@
...
 import org.bitcoins.core.protocol.ln.node.NodeId
 import org.bitcoins.core.protocol.script.ScriptPubKey
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.eclair.rpc.json._
 import org.bitcoins.eclair.rpc.network.NodeUri
 
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
 
 trait EclairApi {
```
Christewart, line 23: Can we link to the latest eclair api docs at the top of this trait with a proper scaladoc?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -54,102 +55,96 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-    eclairCall[ChannelResult]("channel", List(JsString(channelId.hex)))
+    eclairCall[ChannelResult]("channel", "channelId" -> channelId.hex)
   }
 
   private def channels(nodeId: Option[NodeId]): Future[Vector[ChannelInfo]] = {
-    val params =
-      if (nodeId.isEmpty) List.empty else List(JsString(nodeId.get.toString))
-
-    eclairCall[Vector[ChannelInfo]]("channels", params)
+    val params = Seq().flatMap(_ => nodeId.map(id => "nodeId" -> id.toString))
```
Christewart: This seems to be cleaner to me ```nodeId.map(id => "nodeId" -> id.toString).toVector```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -54,102 +55,96 @@ class EclairRpcClient(val instance: EclairInstance)(
...
+    channels(Option(nodeId))
 
   private def close(
       channelId: ChannelId,
-      scriptPubKey: Option[ScriptPubKey]): Future[String] = {
+      scriptPubKey: Option[ScriptPubKey]): Future[Unit] = {
     val params =
       if (scriptPubKey.isEmpty) {
-        List(JsString(channelId.hex))
+        Seq("channelId" -> channelId.hex)
```
Christewart, line 93: nit: In general we like to reach for `Vector` as our general purpose data structure.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -185,30 +180,31 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-             JsString(channelFlags.toString))
+        Seq("nodeId" -> nodeId.toString,
+          "fundingSatoshis" -> _fundingSatoshis,
+          "_pushMsat" -> _pushMsat,
+          "fundingFeerateSatByte" -> feerateSatPerByte.get.toLong.toString,
+          "channelFlags" -> channelFlags.get.toString)
       }
     }
 
     //this is unfortunately returned in this format
```
Christewart, line 202: is this comment old now? Delete if so

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -258,25 +287,68 @@ case class PaymentRequest(
...
+object PaymentStatus {
+  case object PENDING extends PaymentStatus
+  case object SUCCEEDED extends PaymentStatus
+  case object FAILED extends PaymentStatus
+
+  def apply(s: String): PaymentStatus = s match {
+    case "PENDING" => PENDING
+    case "SUCCEEDED" => SUCCEEDED
+    case "FAILED" => FAILED
+    case _ => throw new IllegalArgumentException("Invalid payment status code")
```
Christewart: do this do the error is helpful as to what string we didn't match if we miss a type

```scala
case err: String => throw new IllegalArgumentException("Unknown payment status=${err}")
```

testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -64,25 +61,25 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
...
-              logger.error(
-                List(
-                  "Eclair requires Bitcoin Core 0.16.",
-                  "You can set the environment variable BITCOIND_V16_PATH to override",
-                  "the default bitcoind executable on your PATH."
-                ).mkString(" "))
-              throw exception
-          }
-        } else {
+//        if (cli.instance.getVersion == BitcoindVersion.V17) {
```
Christewart: can this stuff be removed?

testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -91,9 +88,12 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
       port: Int = RpcUtil.randomPort,
       rpcPort: Int = RpcUtil.randomPort,
       zmqPort: Int = RpcUtil.randomPort): BitcoindInstance = {
-    BitcoindRpcTestUtil.instance(port = port,
-                                 rpcPort = rpcPort,
-                                 zmqPort = zmqPort)
+//    BitcoindRpcTestUtil.instance(port = port,
```
Christewart: same here, can this be removed?

Christewart (commented):
This is really high quality stuff. Great job @rorp. Code review is mostly nits :+1:

core/src/main/scala/org/bitcoins/core/protocol/ln/PaymentId.scala
Christewart: Either way, we should add a scaladoc explaining what this is

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
Christewart: Also if they are invariants of `ShortChannelId` itself, we should move the invariants inside of the case class liek this 

```scala
case class ShortChannelId(u64: UInt64) extends NetworkElement {
  require(blockHeight >= 0 && blockHeight <= 0xffffff, "ShortChannelId: invalid block height=${blockHeight}")
//add the others too...
}
```

That way all of the invariants are in a central location and we we don't have to verify them other places. This allows for easier verification

rorp: They are invariants of the parser.

core/src/main/scala/org/bitcoins/core/protocol/ln/ShortChannelId.scala
rorp: Any ideas how? It's clean enough for me.

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
rorp: For some reason this API is implemented in Eclair but not wired to the endpoint.

torkelrogstad (approved):

Christewart:
I get these compiler errors when running locally 

```
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala:5:45: Unused import
[warn] import org.bitcoins.core.protocol.{Address, NetworkElement}
[warn]                                             ^
[info] Done compiling.
[info] Compiling 6 Scala sources to /home/chris/dev/bitcoin-s-core/chain/target/scala-2.12/classes ...
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala:4:18: Unused import
[warn] import java.util.UUID
[warn]                  ^
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala:14:45: Unused import
[warn] import org.bitcoins.core.protocol.{Address, NetworkElement}
[warn]                                             ^
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala:41:15: private val resultKey in class EclairRpcClient is never used
[warn]   private val resultKey = "result"
[warn]               ^
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala:42:15: private val errorKey in class EclairRpcClient is never used
[warn]   private val errorKey = "error"
[warn]               ^
[info] Done compiling.
[warn] /home/chris/dev/bitcoin-s-core/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala:82:34: dead code following this construct
[warn]     val params = Seq().flatMap(_ => nodeId.map(id => "nodeId" -> id.toString))
[warn]                                  ^
[warn] 6 warnings found
[info] Done compiling.
[info] Compiling 1 Scala source to /home/chris/dev/bitcoin-s-core/testkit/target/scala-2.12/classes ...
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:16:49: Unused import
[warn] import org.bitcoins.eclair.rpc.json.{PaymentId, PaymentResult, PaymentStatus}
[warn]                                                 ^
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:17:59: Unused import
[warn] import org.bitcoins.rpc.client.common.{BitcoindRpcClient, BitcoindVersion}
[warn]                                                           ^
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:18:36: Unused import
[warn] import org.bitcoins.rpc.client.v16.BitcoindV16RpcClient
[warn]                                    ^
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:28:41: Unused import
[warn] import org.bitcoins.rpc.util.AsyncUtil.{DEFAULT_INTERNVAL, DEFAULT_MAX_TRIES}
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:28:60: Unused import
[warn] import org.bitcoins.rpc.util.AsyncUtil.{DEFAULT_INTERNVAL, DEFAULT_MAX_TRIES}
[warn]                                                            ^
[warn] /home/chris/dev/bitcoin-s-core/testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala:58:24: Unused import
[warn]     import actorSystem.dispatcher
[warn]                        ^
[warn] 6 warnings found
[info] Done compiling.
```

Christewart:
It also looks like there  are 4 test cases ignored, 
```
should be able to pay to a route !!! IGNORED !!!
must receive gossip messages about channel updates for nodes we do not have a direct channel with !!! IGNORED !!!
should get usable balances !!! IGNORED !!!
should disconnect node !!! IGNORED !!!
```

If the idea here is to cut scope to get something out the door, let's make sure we file an issue so we come back later and implement these. 

EDIT: I'm an idiot :man_facepalming: , don't look at code that early in the morning :-)

PR #537, opened 2019-06-19
Attempt to sync with sendheaders

Christewart:
This PR implements a test case to make sure that our node stays in sync with a bitcoind instance after initial block download.

The sync mechanism is using the [`sendheaders`](https://bitcoin.org/en/developer-reference#sendheaders) message. A peer will automatically send us a header when a block occurs on the network rather than waiting for us to send them a [`getheaders`](https://bitcoin.org/en/developer-reference#getheaders) message.

--------
Fix sync with headers by generating one block in regtest to get out of IBD

Fix comment

Add more comments

Add missing actor system import

more clean up

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
@@ -1,6 +1,8 @@
 package org.bitcoins.node
 
+import akka.actor.ActorSystem
 import org.bitcoins.core.crypto.DoubleSha256DigestBE
+import org.bitcoins.rpc.client.common.BitcoindRpcClient
 import org.bitcoins.rpc.util.RpcUtil
 import org.bitcoins.testkit.node.NodeUnitTest
 import org.bitcoins.testkit.node.fixture.SpvNodeConnectedWithBitcoind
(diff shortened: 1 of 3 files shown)


node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
```diff
@@ -42,6 +46,54 @@ class SpvNodeTest extends NodeUnitTest {
...
+      //start generating a block every 10 seconds with bitcoind
+      //this should result in 5 blocks
+      val startGenF = initSyncF.map { _ =>
+        //generate a block every 5 seconds
+        //until we have generated 5 total blocks
+        genBlock5seconds(bitcoind)
+      }
+
+      startGenF.flatMap { _ =>
+        Thread.sleep(30000)
```
torkelrogstad: Could we do 

```scala
private val desiredBlocks = 5 // somewhere

...
if (count < desiredBlocks) // in genBlock5Seconds
...
val seconds = 5.seconds * desiredBlocks
Thread.sleep(seconds.toMillis) // at the line I'm commenting
```

for easier understanding of what's going on?

torkelrogstad (approved):

rorp (approved):

PR #546, opened 2019-06-21
Use key based encryption in AesCrypt

torkelrogstad:
In this PR we

1) Change the AES encryption mode from CBC to CFB. CFB mode allows for encrypting plaintexts of any size. 
2) Move away from encrypting based on a password (that we salt and extend to the proper key size) to encrypting based on a key of fixed byte size. We also add methods to the old AES password class to extend it to a AES key. This makes our key stretching functionality (PBKDF2) not so tightly coupled to how we do encryption, which in turn makes it much easier to specify a standard that works interchangeable with this. 
3) Introduce tests for decrypting/encrypting across JS/Python/OpenSSL

The parameters we use for AES encryption:
- CFB cipher mode. This cipher mode allows for encrypting plaintexts of arbitrary size, without padding them
- `NoPadding` padding mode. If we didn't use this padding mode the point of using CFB cipher mode would be gone. We want to use this to encrypt REST payloads, so that just makes it more ergonomic.
- Variable key size. We accept AES keys that are 128, 192 and 256 bits long


torkelrogstad:
For context: here are some more code where I attempt using AES in Python and JS, as well as a small samble Python web server I found that integrate `pycrypto` with `crypto-js`. 

Python: https://gist.github.com/9f57c9ec2f14322a9c1ce0a863f4ad50
JS: https://gist.github.com/4611d73567cdcbc40d1da144169c9b03

Sample web server: https://gist.github.com/marcoslin/8026990

Christewart:
Seems to have compiler errors on 2.11 https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/549809629#L241

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -7,54 +7,241 @@ import javax.crypto.{BadPaddingException, Cipher, SecretKey}
...
+/** Represents a salt used to derive a AES key from
+  * a human-readable passphrase.
+  */
+final case class AesSalt private (
+    bytes: ByteVector
+) extends AnyVal
 
-/**
-  * @throws IllegalArgumentException if passed an empty string
+object AesSalt {
```
Christewart: `extends Factory[AesSalt]`

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -7,54 +7,241 @@ import javax.crypto.{BadPaddingException, Cipher, SecretKey}
...
+final case class AesEncryptedData(cipherText: ByteVector, iv: AesIV)
 
-case class AesSalt(
-    value: ByteVector
-)
+/** Represents a salt used to derive a AES key from
+  * a human-readable passphrase.
+  */
+final case class AesSalt private (
+    bytes: ByteVector
```
Christewart, line 76: Not exactly sure what this is suppose to do? Is this what you wanted? 

```scala
final case class AesSalt(
    private val bytes: ByteVector
) extends AnyVal
```

Christewart (commented):
Is there any reason we are concerned about boxing values in this PR? It seems like this made things way more complicated than they need to be. I think looking at boxing is a good idea, especially for the `NumberType` stuff, but I don't think here is a good place to start

torkelrogstad:
> Is there any reason we are concerned about boxing values in this PR? It seems like this made things way more complicated than they need to be. I think looking at boxing is a good idea, especially for the `NumberType` stuff, but I don't think here is a good place to start

So one reason is to avoid boxing (which I partially did to explore what that looks like, for future cases with greater payoff). The other reason (which in this case is more important, is to control what sort of constructors we provide to the end user. For example in the case of an `AesPassword`. There's a general `fromString` constructor that returns an `Option`, and then there's the explicit `fromNonEmptyString`, which returns `AesPassword`. That means the user can call the explicit method and avoiding having to deal with the error case, but they need to acknowledge that the responsibility is on them. Especially with types such as `AesKey`, I think this is very beneficial:

1) When constructing from bytes, we need a very specific number of bytes. 
2) It is very important that those bytes are random

Taking away the possibility of doing `AesKey(hex"12345678")` is a nice thing IMO

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
```diff
@@ -4,129 +4,301 @@ import org.bitcoins.testkit.core.gen.{NumberGenerator, StringGenerators}
...
+    val plainText = "foobar"
+    val Right(plainbytes) = ByteVector.encodeUtf8(plainText)
+    val iv = getIV(hex"455014871CD34F8DCFD7C1E387987BFF")
+    val expectedCipher = ByteVector.fromValidBase64("oE8HErg1lg==")
+
+    val encrypted = AesCrypt.encryptWithIV(plainbytes, iv, key)
+
+    // for some reason we end up with different cipher texts
+    // decrypting works though...
+    // assert(encrypted.cipherText == expectedCipher)
```
Christewart, line 103: hmm, that is concerning. This should be a deterministic process right? There isn't some sort of nonce the openssl is generating that causes this diff is there? It's even weirder that we can properly decrypt it

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
```diff
@@ -4,129 +4,301 @@ import org.bitcoins.testkit.core.gen.{NumberGenerator, StringGenerators}
...
+      val Right(decrypted) = AesCrypt.decrypt(encrypted, key)
+      assert(decrypted == plainbytes)
+
+      val Right(decryptedText) = decrypted.decodeUtf8
+      assert(decryptedText == plaintext)
+    }
+  }
+
+  /** Asserts that the two bytevectors are equal expect for trailing padding */
+  def assertPaddedEqual(first: ByteVector, second: ByteVector): Assertion = {
```
Christewart: So this worries me a bit, can you add a explanation why this is needed? Are we padding bits or bytes? where does this extra padding come from?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -7,54 +7,297 @@ import javax.crypto.{BadPaddingException, Cipher, SecretKey}
...
+    * We serialize IV and ciphertext by prepending the IV
+    * to the ciphertext, and converting it to base64.
+    * Since the IV is of static length, deserializing is a matter
+    * of taking the first bytes as IV, and the rest as
+    * ciphertext.
+    */
+  def fromBase64(base64: String): Option[AesEncryptedData] = {
+    ByteVector.fromBase64(base64) match {
+      case None                                        => None
+      case Some(bytes) if bytes.length <= AesIV.length => None
```
Christewart, line 47: So this is implying that you can have a empty cipher text with an IV? That seems to be the assumption any way

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
Christewart: You are using `-nosalt` with your command above, does that have something to do with it?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -7,54 +7,297 @@ import javax.crypto.{BadPaddingException, Cipher, SecretKey}
...
-case class AesPassword(value: String) {
-  require(value.nonEmpty, "AES passwords cannot be empty!")
+final case class AesIV private (private val underlying: ByteVector)
+    extends NetworkElement {
+  require(
+    underlying.length == 16,
+    s"AES salt must be 16 bytes long! Got: ${underlying.length}"
+  )
+
+  val bytes: ByteVector = underlying
```
Christewart: override

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
```diff
@@ -63,118 +306,77 @@ object AesCrypt {
...
+    val decryptionAttempt = Try {
+      val plainText = cipher.doFinal(encrypted.cipherText.toArray)
+      ByteVector(plainText)
+    }
+
+    decryptionAttempt match {
+      case Success(bytes) => Right(bytes)
+      // here we assume that bad padding is because of a bad password
+      // provided. assuming that our implementation is correct, correct
+      // passwords should never to lead to bad padding
```
Christewart, line 323: is there a way we can test this with property based testing?

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
torkelrogstad: Added a comment explaining the line

core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala
torkelrogstad: I'm not sure how. Generally speaking, I believe its impossible to know if decryption failed because of bad data or a bad key.

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
torkelrogstad: Added explanatory comments

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
torkelrogstad: I'm actually not sure about this. Will do some more investigation tomorrow.

build.sbt
```diff
@@ -35,7 +35,8 @@ lazy val compilerOpts = Seq(
   "-Ywarn-unused",
   "-unchecked",
   "-deprecation",
-  "-feature"
+  "-feature",
+  "-Xsource:2.12"
```
Christewart, line 39: FYI

https://github.com/scala/scala/blob/2f1b5259188698501dbc8430f63972bf7bc68154/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala#L138

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
Christewart: nit: I would add that this is only needed for the pycrypto test vector

Christewart (approved):

build.sbt
torkelrogstad: Actually ended up posting a question Scala Users: https://users.scala-lang.org/t/understanding-the-xsource-option/4749

PR #547, opened 2019-06-21
SipHashKey

### LINKED ISSUE #521: BIP 158 Key Type
SipHash as specified in [BIP 158](https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki) requires a 128-bit key for which we currently use a `ByteVector` when we really should have a unique type for these keys.

cwaldron97:
Currently replacing ByteVector with SipHashKey to ensure only 16 byte length keys are passed. Trying to fix compiler errors.
Closes #521

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
+++ b/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
@@ -51,10 +51,7 @@ object GCS {
       key: SipHashKey): GolombFilter = {
     buildGolombFilter(data, key, BlockFilter.P, BlockFilter.M)
   }
-
-  //item remains bytevector key to be changed create new file in GCS folder that redefines sipkeyhash
-  //will probably have a require statement bytevector length of 16 bytes or 128 bits. Add a couple methods
-  //to turn the key into an array of bytes (look to toArray method on line 65)
+  
   private def sipHash(item: ByteVector, key: SipHashKey): UInt64 = {
     // https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#hashing-data-objects
     val sipHashCParam = 2
@@ -62,7 +59,7 @@ object GCS {
 
     val sh = new SipHash(sipHashCParam, sipHashDParam)
 
-    val keyParam = new KeyParameter(key.toArray)
+    val keyParam = new KeyParameter(key.bytes.toArray)
 
     sh.init(keyParam)
 
--- a/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
+++ b/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
@@ -51,7 +51,7 @@ object GCS {
       key: SipHashKey): GolombFilter = {
     buildGolombFilter(data, key, BlockFilter.P, BlockFilter.M)
   }
-  
+
   private def sipHash(item: ByteVector, key: SipHashKey): UInt64 = {
     // https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#hashing-data-objects
     val sipHashCParam = 2
--- a/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
+++ b/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
@@ -21,7 +21,7 @@ object GCS {
     */
   def buildGCS(
       data: Vector[ByteVector],
-      key: ByteVector,
+      key: SipHashKey,
       p: UInt8,
       m: UInt64): BitVector = {
     val hashedValues = hashedSetConstruct(data, key, m)
@@ -34,7 +34,7 @@ object GCS {
     */
   def buildGolombFilter(
       data: Vector[ByteVector],
-      key: ByteVector,
+      key: SipHashKey,
       p: UInt8,
       m: UInt64): GolombFilter = {
     val encodedData = buildGCS(data, key, p, m)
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -170,7 +172,7 @@ object BlockFilter {
     val (size, filterBytes) = bytes.splitAt(1)
     val n = CompactSizeUInt.fromBytes(size)
 
-    GolombFilter(blockHash.bytes.take(16),
+    GolombFilter(SipHashKey(blockHash.bytes.take(16)),
```
cwaldron97: There is probably a more acceptable way to do this but this is how I got it to work for now

core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -48,11 +48,14 @@ object GCS {
...
-      key: ByteVector): GolombFilter = {
+      key: SipHashKey): GolombFilter = {
     buildGolombFilter(data, key, BlockFilter.P, BlockFilter.M)
   }
 
-  private def sipHash(item: ByteVector, key: ByteVector): UInt64 = {
+  //item remains bytevector key to be changed create new file in GCS folder that redefines sipkeyhash
+  //will probably have a require statement bytevector length of 16 bytes or 128 bits. Add a couple methods
+  //to turn the key into an array of bytes (look to toArray method on line 65)
+  private def sipHash(item: ByteVector, key: SipHashKey): UInt64 = {
```
nkohen, line 55: nit: Generally above `def`s its best to wrap comments with
```
/**
  * Comment
  * Here
  */
```
since that auto-generates scaladocs :)

nkohen: nit: Also in general its best to have the comment above a def describe what the def does and what its non-trivial inputs are so that the comment stays relevant in the future, rather than describing changes to the def

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
nkohen: That's actually perfect, although I'd recommend you use an intermediate val like you did above on line 152 to make this more readable

core/src/main/scala/org/bitcoins/core/gcs/SipHashKey.scala
```diff
@@ -0,0 +1,17 @@
...
+  def toArray: Array[Byte] = {
+    bytes.toArray
+  }
+
+}
+
+
+
+
+
```
nkohen: nit: bunch of blank lines, normally just one is good :P

core/src/main/scala/org/bitcoins/core/gcs/SipHashKey.scala
```diff
@@ -0,0 +1,17 @@
+package org.bitcoins.core.gcs
+import scodec.bits.ByteVector
+case class SipHashKey(bytes: ByteVector) {
+  require(
+    bytes.size == 16,
+    "Can only use a key length of 16 bytes, got: " + bytes.size)
+
+  def toArray: Array[Byte] = {
```
nkohen: nit: since this will be called from a `key: SipHashKey`, I'd recommend naming this `toByteArray` since it isn't obvious what `toArray` might mean for a `key`. (this can be done super easily by right clicking on this def in intellij and going to refactor > rename)

core/src/main/scala/org/bitcoins/core/gcs/SipHashKey.scala
```diff
@@ -0,0 +1,17 @@
+package org.bitcoins.core.gcs
+import scodec.bits.ByteVector
+case class SipHashKey(bytes: ByteVector) {
```
nkohen: from inside `sbt`, run scalafmt and it will autoformat your code to look like the rest of our codebase :)

nkohen:
@cwaldron97 I left you some review, it is very full of nits which I normally wouldn't leave so many of but just trying to help you get more used to conventions and such :) you've done a great job here so don't let all the critique discourage you!

nkohen:
Also if you wouldn't mind, generally good to reference the issue this addresses in the description (with a # in front of the number of the issue) as this will auto-link to that issue and the issue will automatically reference this PR in its history

core/src/main/scala/org/bitcoins/core/gcs/SipHashKey.scala
Christewart: One thing I would recommend doing here is extending [`NetworkElement`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/protocol/NetworkElement.scala)

So the definition would look like this

```scala
case class SipHashKey(bytes: ByteVector) extends NetworkElement

Christewart: `NetworkElement` gives you a `.hex` and `.bytes` method you can call on to always get the byte/hex representation of the element. If you look at the definition of a bitcoin [`Transaction`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala#L19) you see it also extends `NetworkElement` which gives you a nice way to serialize the transaction to be sent to the p2p network.

core/src/main/scala/org/bitcoins/core/gcs/SipHashKey.scala
Christewart: You should be able to get rid of this method completely and just call `sipHashKey.bytes` if you extend `NetworkElement` on the case class.

Christewart:
Looks like you have some compiler errors @cwaldron97 

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/549894644#L625

nkohen (approved):

Issue #453, opened 2019-05-02
Add more complex test cases like reorgs for ChainSync


Christewart:
As it stands in #450 there is no unit tests for reorgs when we query third parties and they have a different blockchain than us. 

We need to create some methods in `Testkit` and the `ChainFixtures` that allows us to simulate these reorg scenarios and then make sure we do the right thing

Christewart:
#548 partially addresses this issue.

We still do not take chain weight into consideration when thinking about reorgs, per Torkel Rogstad on slack: 

```
torkel [7:04 AM]
I've got a question about your PR:

In the tests there's only a 1 block reorg. When I'm looking at the implementation of handling reorgs, we only compare one block at a time, where we select the tip with the most weights.

What happens if we receive a 2 block reorg where the first block of the reorg is not heavy enough to overtake the previous tip, but the 2 blocks together has enough work the be the new heaviest? (edited) 
```

In general, as of d00dff564585957f42b44cd839d260e0b2f66e76 weight is not considered in reorg logic. This needs to change in the future.

PR #548, opened 2019-06-24
Reorg handling in chain project

Christewart:
This begins to address #453 

This starts adding re-org logic for our chain handling. This PR does two things

1. Makes our Blockchain data structure extend `IndexedSeqLike` in the scala std library so we get all of the operations we know and love from scala collections
2. Implements re-orgs in the chain project. 

The re-org logic is still fairly dumb. This will traverse back in time until the genesis block and see if we can connect the block we just received to any block. In theory this is the desired behavior -- if a miner is mining a _very_ old chain we shouldn't outright reject the block in terms of consesus, but it should _not_ force a re-org and be the new best tip.

In practice, as a lite client, we may want to consider rejecting this block since it has no chance in practice to catch up with the best POW chain -- unless there is some sort of crazy selfish mining attack happening. 

TODO: 

- [x] - test case where we receive a new block on a shorter chain that does _not_ force that shorter chain pass the longer chain in terms of height (negative test case). We should assert that the best hash is still the same.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -134,9 +134,12 @@ class ChainHandlerTest extends ChainUnitTest {
     chainHandler: ChainHandler =>
       val reorgFixtureF = buildChainHandlerCompetingHeaders(chainHandler)
       val chainHandlerCF = reorgFixtureF.map(_.chainApi)
+      // header B, best hash ATM
       val newHeaderBF = reorgFixtureF.map(_.headerDb1)
+      // header C, same height as B but was seen later
       val newHeaderCF = reorgFixtureF.map(_.headerDb2)
 
+      // check that header B is the leader
       val assertBBestHashF = for {
         chainHandler <- chainHandlerCF
         headerB <- newHeaderBF
@@ -145,6 +148,8 @@ class ChainHandlerTest extends ChainUnitTest {
         assert(bestHash == headerB.hashBE)
       }
 
+      // build a new header D off of C which was seen later
+      // but has the same height as B
       val newHeaderDF =
         newHeaderCF.map(h => BlockHeaderHelper.buildNextHeader(h))
 
@@ -160,42 +165,9 @@ class ChainHandlerTest extends ChainUnitTest {
         newHeaderD <- newHeaderDF
         hash <- chainHandler.getBestBlockHash
       } yield {
+        // assert that header D overtook header B
         assert(hash == newHeaderD.hashBE)
       }
-  }
-
-  it must "handle a reorg where both chains are at the exact same height when a new block comes in" in {
-    chainHandler: ChainHandler =>
-      val reorgFixtureF = buildChainHandlerCompetingHeaders(chainHandler)
-      val chainHandlerCF = reorgFixtureF.map(_.chainApi)
-      val newHeaderBF = reorgFixtureF.map(_.header1)
-      val newHeaderCF = reorgFixtureF.map(_.header2)
-
-      val bestHashB = for {
-        chainHandler <- chainHandlerCF
-        bestHash <- chainHandler.getBestBlockHash
-        newHeaderB <- newHeaderBF
-      } yield assert(bestHash == newHeaderB.hashBE)
-
-      //let's generate block D that is built ontop of C
-      val newHeaderDF = for {
-        newHeaderC <- newHeaderCF
-        bh <- chainHandlerCF.flatMap(_.getHeader(newHeaderC.hashBE))
-      } yield BlockHeaderHelper.buildNextHeader(bh.get)
-
-      val chainHandlerDF = for {
-        chainHandlerC <- chainHandlerCF
-        newHeaderD <- newHeaderDF
-        ch <- chainHandlerC.processHeader(newHeaderD.blockHeader)
-      } yield ch
-
-      //now our best hash should be D
-      for {
-        chainHandler <- chainHandlerDF
-        bestHashD <- chainHandler.getBestBlockHash
-        _ <- bestHashB
-        newHeaderD <- newHeaderDF
-      } yield assert(bestHashD == newHeaderD.hashBE)
   }
 
   it must "NOT reorg to a shorter chain that just received a new block" in {
@@ -254,7 +226,7 @@ class ChainHandlerTest extends ChainUnitTest {
     }
   }
 
-  /** Builds two competing headers that are built onto of parent */
+  /** Builds two competing headers that are built from the same parent */
   private def buildCompetingHeaders(
       parent: BlockHeaderDb): (BlockHeader, BlockHeader) = {
     val newHeaderB =
@@ -278,12 +250,9 @@ class ChainHandlerTest extends ChainUnitTest {
   /** Builds two competing headers off of the [[ChainHandler.getBestBlockHash best chain tip]] */
   private def buildChainHandlerCompetingHeaders(
       chainHandler: ChainHandler): Future[ReorgFixture] = {
-    val bestTip = chainHandler.getBestBlockHeader
-    val headersBC = bestTip.map(h => buildCompetingHeaders(h))
-
     for {
-      (newHeaderB, newHeaderC) <- headersBC
-      oldBestTip <- bestTip
+      oldBestTip <- chainHandler.getBestBlockHeader
+      (newHeaderB, newHeaderC) = buildCompetingHeaders(oldBestTip)
       newChainApi <- chainHandler.processHeaders(Vector(newHeaderB, newHeaderC))
       newHeaderDbB <- newChainApi.getHeader(newHeaderB.hashBE)
       newHeaderDbC <- newChainApi.getHeader(newHeaderC.hashBE)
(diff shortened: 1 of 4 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -53,28 +72,41 @@ object Blockchain extends BitcoinSLogger {
...
-              tipResult match {
-                case TipUpdateResult.Success(headerDb) =>
-                  val newChain =
-                    Blockchain.fromHeaders(headerDb +: blockchain.headers)
-                  BlockchainUpdate.Successful(newChain, headerDb)
-                case fail: TipUpdateResult.Failure =>
-                  BlockchainUpdate.Failed(blockchain, header, fail)
-              }
+            val prevBlockHeaderOpt =
+              blockchain.find(_.hashBE == header.previousBlockHashBE)
```
torkelrogstad, line 76: When we're looking for block headers to connect to we're more likely to find a matching tip towards the end than the start. Would it make sense to do `blockchain.reverseIterator.find(...)`?

torkelrogstad:
I have a question about `Blockchain.parseSuccessOrFailure)`. We're just finding the first future which is a successful update. What happens if we have multiple successful updates, but the "heaviest" (most PoW) is _not_ the first we find?

torkelrogstad (commented):
I have a question about Blockchain.parseSuccessOrFailure). We're just finding the first future which is a successful update. What happens if we have multiple successful updates, but the "heaviest" (most PoW) is not the first we find?

rorp (approved):

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
Christewart: Actually headers are prepended here: https://github.com/bitcoin-s/bitcoin-s/blob/master/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala#L68

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -141,4 +253,42 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      headerDb2: BlockHeaderDb,
+      oldBestBlockHeader: BlockHeaderDb) {
+    lazy val header1: BlockHeader = headerDb1.blockHeader
+    lazy val header2: BlockHeader = headerDb2.blockHeader
+  }
+
+  /** Builds two competing headers off of the [[ChainHandler.getBestBlockHash best chain tip]] */
+  private def buildChainHandlerCompetingHeaders(
+      chainHandler: ChainHandler): Future[ReorgFixture] = {
+    val bestTip = chainHandler.getBestBlockHeader
```
torkelrogstad: Can do this entire method as a `for`, more readable IMO:

```scala
for {
      oldBestTip <- chainHandler.getBestBlockHeader
      (newHeaderB, newHeaderC) = buildCompetingHeaders(oldBestTip)
      newChainApi <- chainHandler.processHeaders(Vector(newHeaderB, newHeaderC))
      newHeaderDbB <- newChainApi.getHeader(newHeaderB.hashBE)
      newHeaderDbC <- newChainApi.getHeader(newHeaderC.hashBE)
    } yield {
      ReorgFixture(newChainApi, newHeaderDbB.get, newHeaderDbC.get, oldBestTip)
    }
```

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -46,4 +46,10 @@ trait ChainApi {
   /** Gets the hash of the block that is what we consider "best" */
   def getBestBlockHash(
       implicit ec: ExecutionContext): Future[DoubleSha256DigestBE]
+
+  /** Gets the best block header we have */
+  def getBestBlockHeader(
+      implicit ec: ExecutionContext): Future[BlockHeaderDb] = {
+    getBestBlockHash.flatMap(getHeader).map(_.get)
```
torkelrogstad: `.getOrElse(throw something)`? `.get` gives very non-descriptive error messages. Or even: 
```scala
 for {
      hash <- getBestBlockHash
      headerOpt <- getHeader(hash)
    } yield
      headerOpt match {
        case None         => throw ???
        case Some(header) => header
      }
```
to make it super clear what's going on

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -141,4 +253,42 @@ class ChainHandlerTest extends ChainUnitTest {
       case Nil => succeed
     }
   }
+
+  /** Builds two competing headers that are built onto of parent */
```
torkelrogstad: `that are built from the same parent`?

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
       }
   }
 
+  it must "handle a very basic reorg where one chain is one block behind the best chain" in {
```
torkelrogstad, line 133: Unless I'm missing something here, this test and the one below tests the same thing. The flow in both are:

1) We have header B and C at the same height, with B being the leader (because B was processed first in `buildChainHandlerCompetingHeaders`)
2) Build a new header D off of the trailing one 
3) Process D
4) Check that D is the new leader

torkelrogstad: Also, to understand what's going on in the tests I added some comments locally, posting them here (feel free to include them if you think they make sense), hopefully easier for the next person to understand what's going on

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+    chainHandler: ChainHandler =>
+      val reorgFixtureF = buildChainHandlerCompetingHeaders(chainHandler)
+      val chainHandlerCF = reorgFixtureF.map(_.chainApi)
+      val newHeaderBF = reorgFixtureF.map(_.headerDb1)
```
torkelrogstad, line 138:
```scala
// header B, best hash ATM
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      val newHeaderCF = reorgFixtureF.map(_.headerDb2)
```
torkelrogstad, line 140:
```scala
// header C, trailing header B
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+
+      val assertBBestHashF = for {
```
torkelrogstad, line 143:
```scala
// check that header B is the leader
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+        chainHandler <- chainHandlerCF
+        headerB <- newHeaderBF
+        bestHash <- chainHandler.getBestBlockHash
+      } yield {
+        assert(bestHash == headerB.hashBE)
+      }
+
+      val newHeaderDF =
```
torkelrogstad, line 153:
```scala
// build a new header D off of the trailing one (C)
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+        newHeaderCF.map(h => BlockHeaderHelper.buildNextHeader(h))
+
+      val chainHandlerDF = for {
+        _ <- assertBBestHashF
+        newHeaderD <- newHeaderDF
+        chainHandler <- chainHandlerCF
+        chainHandlerD <- chainHandler.processHeader(newHeaderD.blockHeader)
```
torkelrogstad, line 160:
```scala
// header D should now have overtaken header B
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -124,6 +130,112 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      } yield chainHandlerD
+
+      for {
+        chainHandler <- chainHandlerDF
+        newHeaderD <- newHeaderDF
+        hash <- chainHandler.getBestBlockHash
+      } yield {
+        assert(hash == newHeaderD.hashBE)
```
torkelrogstad, line 169:
```scala
// assert that header D overtook header B
```

torkelrogstad (commented):
What's your thoughts on creating a new fixture with `buildChainHandlerCompetingHeaders`, and then avoiding all the `Future` mappings at the start of the tests

PR #553, opened 2019-06-27
 implement Ordered[T] instead of creating custom methods

### LINKED ISSUE #436: Implement Ordering for currency units
In use cases like this: 
https://github.com/bitcoin-s/bitcoin-s-core/blob/f0aa98512c859e8e73db56de7c110c406e12bb61/wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala#L22

It would be handy to have an ordering for currency units. 

Scaladoc on `Ordering`: https://www.scala-lang.org/api/2.12.3/scala/math/Ordering.html

Should be simple to implement with existing comparison operators

piu130:
Closes #436 

Scala provides the trait [Ordered](https://www.scala-lang.org/api/2.12.1/scala/math/Ordered.html).

We can implement `compare(that: A): Int ` from it and get the same functions currently written out (eg. `<=`, `>`).

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
+++ b/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
@@ -9,26 +9,14 @@ import scodec.bits.ByteVector
 
 sealed abstract class CurrencyUnit
     extends NetworkElement
+    with Ordered[CurrencyUnit]
     with BasicArithmetic[CurrencyUnit] {
   type A
 
   def satoshis: Satoshis
 
-  def >=(c: CurrencyUnit): Boolean = {
-    satoshis.underlying >= c.satoshis.underlying
-  }
 
-  def >(c: CurrencyUnit): Boolean = {
-    satoshis.underlying > c.satoshis.underlying
-  }
-
-  def <(c: CurrencyUnit): Boolean = {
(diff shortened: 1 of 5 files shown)


torkelrogstad (approved):
Great! Thanks for the contribution.

Christewart:
@piu130 Another place you might want to implement this is [`ScriptNumber`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/script/constant/Constants.scala#L47) if you have a chance

piu130:
@Christewart done. Found two more in `LnCurrencyUnit` and `MilliSatoshis`.

Christewart (approved):

torkelrogstad:
By the way @piu130 , if you're interested in following development in Bitcoin-S or even continue contributing, we have a open Slack you can join: https://join.slack.com/t/suredbits/shared_invite/enQtNDEyMjY3MTg1MTg3LTYyYjkwOGUzMDQ4NDAwZjE1M2I3MmQyNWNlZjNlYjg4OGRjYTRjNWUwNjRjNjg4Y2NjZjAxYjU1N2JjMTU1YWM

It's the company Slack of Suredbits, the channel `#bitcoin-s` there is all about Bitcoin-S. Hoping to see you there!

PR #555, opened 2019-06-27
Process outgoing transactions

### LINKED ISSUE #538: Manage UTXO state in DB when spending
Umbrella issue: #530

--- 

We need to somehow manage a record of spent vs. unspent TXOs. One possible solution is to change our UTXO table to a TXO table, with a field that distinguishes the two as spent/unspent.

This issus is a logical next step after #534 is merged in.

torkelrogstad:
Closes #538 

In this PR we: 

1) Split wallet functionality into traits that gets mixed in, mimicking what we do in the `bitcoind` RPC client. This has the benefit of making it clearer which methods are intended as public facing API methods, which are for internal use within the wallet, and which are just helper functions that shouldn't be used outside of the trait it is in. 
2) Insert change UTXOs into the DB whenever we send outgoing TXs
3) Update spent/unspent status of UTXOs when `processTransaction` (or its internally facing brother `processOurTransaction` (called when processing TXs we've created and signed) is called 
4) Rework `TransactionOutput` and `TransactionOutPoint` from traits with `*Impl` case classes to just pure case classes. This cuts a bit down on code size, and makes it easier to understand what's going on. 
5) Unify the DB representation of TXOs, leaving everything in one table.


wallet/src/main/scala/org/bitcoins/wallet/internal/AddressAndAccountHandling.scala
```diff
@@ -0,0 +1,162 @@
...
+import org.bitcoins.core.protocol.blockchain.MainNetChainParams
+import org.bitcoins.core.protocol.blockchain.RegTestNetChainParams
+import org.bitcoins.core.protocol.blockchain.TestNetChainParams
+import org.bitcoins.core.hd.HDCoinType
+
+/**
+  * Provides functionality related to accounts and addresses. This includes
+  * enumeratng and creating them, primarily.
+  */
+private[wallet] trait AddressAndAccountHandling { self: LockedWallet =>
```
Christewart: Is there a reason we are coupling these things together? Why not just have a `AddressHandling` and `AccountHandling` trait?

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressAndAccountHandling.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  /** Enumerates the public keys in this wallet */
+  protected[wallet] def listPubkeys(): Future[Vector[ECPublicKey]] =
+    addressDAO.findAllPubkeys().map(_.toVector)
+
+  /**
+    * @param account Account to generate address from
+    * @param chainType What chain do we generate from? Internal change vs. external
+    */
+  private def getNewAddressHelper(
```
Christewart: nit: It would be nice to add a scaladoc to this explaining what this method intends to do, for instance it needs to insert the newly generated address in our address table

core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutPoint.scala
```diff
@@ -4,58 +4,47 @@ import org.bitcoins.core.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
...
-final case object EmptyTransactionOutPoint extends TransactionOutPoint {
-
-  override val txId: DoubleSha256Digest =
-    DoubleSha256Digest(
-      BitcoinSUtil.decodeHex(
-        "0000000000000000000000000000000000000000000000000000000000000000"))
-  override val vout: UInt32 = UInt32("ffffffff")
+final object EmptyTransactionOutPoint
+    extends TransactionOutPoint(txId = DoubleSha256Digest.empty,
+                                vout = UInt32.fromBytes(hex"ffffffff")) {
```
Christewart: nit: I think this can be `UInt32.max`

core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutput.scala
```diff
@@ -7,37 +7,23 @@ import org.bitcoins.core.serializers.transaction.RawTransactionOutputParser
...
   //https://bitcoin.org/en/developer-reference#txout
   override def size = scriptPubKey.size + 8
 
   override def bytes = RawTransactionOutputParser.write(this)
 }
 
-case object EmptyTransactionOutput extends TransactionOutput {
-  override def value = CurrencyUnits.negativeSatoshi
-  override def scriptPubKey = ScriptPubKey.empty
+object EmptyTransactionOutput
```
Christewart: final case object?

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
```diff
@@ -31,4 +32,22 @@ object Implicits {
       loop(0)
     }
   }
+
+  /** Extension methods for sequences of assertions */
+  implicit class AssertionSeqOps(private val assertions: Seq[Assertion]) {
+
+    /** Flattens a sequence of assertions into only one */
+    def toAssertion: Assertion = assertions match {
+      case Seq() =>
```
Christewart, line 42: Seq.empty

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
```diff
@@ -31,4 +32,22 @@ object Implicits {
...
+        // should this instead be a TestFailedExceptin? Not sure what's best
+        throw new IllegalArgumentException(
```
Christewart: if we have zero assertions, shouldn't this be trivially true? Idk. It's a good question..

Christewart: I think the test failed exception would be the best default assumption for now the more i think about it

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
```diff
@@ -31,4 +32,22 @@ object Implicits {
...
+          "Cannot turn an empty list into an assertion!")
+      // this should force all collection kinds to
+      // evaluate all their members, throwing when
+      // evaluating a bad one
+      case nonEmpty =>
+        nonEmpty.foreach(_ => ())
```
Christewart, line 50: maybe something like this?

```scala
        val containsFailure = nonEmpty.exists(_ != Assertions.succeed)
        assert(!containsFailure)
```

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -29,20 +29,20 @@ abstract class LockedWallet
...
+  /** Sums up the value of all unspent, incoming
+    * TXOs in the wallet, filtered by the given predicate */
   private def filterThenSum(
       predicate: IncomingWalletTXO => Boolean): Future[CurrencyUnit] = {
     for (utxos <- incomingTxoDAO.findAllWithSpendingInfo())
       yield
         utxos
           .collect {
-            case (txo, spendInfo) if predicate(txo) => spendInfo.value
+            case (txo, spendInfo) if !txo.spent && predicate(txo) =>
```
Christewart: why the `!txo.spent` here? Shouldn't that be included in the `predicate`?

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressAndAccountHandling.scala
```diff
@@ -34,7 +39,22 @@ private[wallet] trait AddressAndAccountHandling { self: LockedWallet =>
...
   protected[wallet] def listPubkeys(): Future[Vector[ECPublicKey]] =
-    addressDAO.findAllPubkeys().map(_.toVector)
+    addressDAO.findAllPubkeys()
+
+  protected[wallet] def listSPKs(): Future[Vector[ScriptPubKey]] =
+    addressDAO.findAllSPKs()
+
+  /** Given a transaction, returns the outputs (with their corresponding outpoints)
+    * that pay to this wallet */
+  def findOurOuts(transaction: Transaction): Future[
```
Christewart: Maybe follow the bitcoin core convention of `isMine()`?

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTxoDAO.scala
```diff
@@ -36,6 +36,24 @@ final case class IncomingTxoDAO(profile: JdbcProfile)(
...
+    */
+  def findOutputsBeingSpent(
+      tx: Transaction): Future[Seq[(SpendingInfoDb, IncomingWalletTXO)]] = {
+    val joined =
+      spendingInfoTable join table on (_.id === _.spendingInfoID)
+
+    val filtered = joined
+      .filter {
+        case (txoInfo, _) =>
+          txoInfo.outPoint.inSet(tx.inputs.map(_.previousOutput))
```
Christewart: nit: Cache the `tx.inputs.map(_.previousOutput)` outside of the query? Not sure if there is a performance impact there

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTxoDAO.scala
```diff
@@ -54,6 +55,31 @@ final case class IncomingTxoDAO(profile: JdbcProfile)(
...
+    */
+  def markAsSpent(
+      outputs: Seq[TransactionOutput]): Future[Vector[IncomingWalletTXO]] = {
+    val spendingInfos = spendingInfoTable filter (_.output.inSet(outputs))
+    val joined = spendingInfos join table on (_.id === _.spendingInfoID)
+    val utxosQuery = joined.map(_._2)
+
+    for {
+      utxos <- database.run(utxosQuery.result)
+      _ = assert(
```
Christewart: Would prefer to use a `require()` rather than `assert()`. This makes me think of the scalatest assert

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTxoDAO.scala
```diff
@@ -54,6 +55,31 @@ final case class IncomingTxoDAO(profile: JdbcProfile)(
     database.run(filtered.result)
   }
 
+  /** Marks the given outputs as spent. Assumes that all the
+    * given outputs are ours, throwing if numbers aren't
+    * confirming that.
+    */
+  def markAsSpent(
+      outputs: Seq[TransactionOutput]): Future[Vector[IncomingWalletTXO]] = {
+    val spendingInfos = spendingInfoTable filter (_.output.inSet(outputs))
```
Christewart: I would really prefer dot notation with things like this

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTxoDAO.scala
```diff
@@ -54,6 +55,31 @@ final case class IncomingTxoDAO(profile: JdbcProfile)(
...
+    val joined = spendingInfos join table on (_.id === _.spendingInfoID)
```
Christewart: Can these joins be moved into the private vals in this class?

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
```diff
@@ -21,12 +22,30 @@ class WalletIntegrationTest extends BitcoinSWalletTest {
...
+      first: CurrencyUnit,
+      second: CurrencyUnit): Boolean = {
+    val diff =
+      if (first > second) {
+        first - second
+      } else if (first < second) {
+        second - first
+      } else 0.sats
+
+    diff < 1000.sats
```
Christewart: We should make another param for the delta, with this as a the default value perhaps?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,283 @@
...
+  // Internal wallet API
+
+  /**
+    * Processes TXs originating from our wallet.
+    * This is called right after we've signed a TX,
+    * updating our UTXO state.
+    */
+  private[wallet] def processOurTransaction(
+      transaction: Transaction,
+      confirmations: Int) = {
```
Christewart: return type please

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,283 @@
...
+      transaction: Transaction,
+      confirmations: Int): Future[ProcessTxResult] = {
+
+    val incomingTxoFut: Future[Vector[IncomingWalletTXO]] =
+      incomingTxoDAO
+        .findTx(transaction)
+        .flatMap {
+          // no existing elements found
+          case Vector() =>
+            processNewIncomingTx(transaction, confirmations)
```
Christewart: I don't think I really like the distinction between `processNewIncomingTx` and `processExistingIncomingTxo`, I think the behavior seems redundant and it wouldn't be hard to consolidate this down to one method. Also why does `processNewIncomingTx` take a confirmation parameter? Is this because we aren't making strong guarantees about when the wallet is receiving blocks? I.e. the wallet is offline for 2 days before we sync it back up?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,283 @@
...
+        )
+    }.toVector
+  }
+
+  /**
+    * Processes an incoming transaction that already exists in our wallet.
+    * If the incoming transaction has more confirmations than what we
+    * have in the DB, we update the TX
+    */
+  private def processExistingIncomingTxo(
```
Christewart, line 167: I don't like the return type here, let's keep it `Future[IncomingWalletTxo]` and explicitly throw in the case of reorgs. Currently we call `.flatten` inside of `processTransactionImpl` which silently disregards the `Option` type returned here.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,283 @@
...
+  private def createIncomingTxos(
+      transaction: Transaction,
+      confirmations: Int,
+      xs: Seq[(SpendingInfoDb, OutputWithIndex)]): Vector[IncomingWalletTXO] = {
+    xs.map {
+      case (utxo, OutputWithIndex(out, _)) =>
+        IncomingWalletTXO(
+          confirmations = confirmations,
+          // is this always the case?
+          spent = false,
```
Christewart: Yes, i believe it has to be, at least for this transaction. You can't reference a transaction's outpoint within the transaction it was created in. You end up in a recursive dependency situation.

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
torkelrogstad: `stable identifier required, but scala.collection.Seq.empty found.scalac`

testkit/src/main/scala/org/bitcoins/testkit/Implicits.scala
torkelrogstad: Got this error message after trying your suggestion: 

```
exception during macro expansion: 
scala.reflect.macros.TypecheckException: not found: value assertionsHelper
```

I think that the traits we mix in when using ScalaTest introduce a bunch of magic around `assert` that makes it impossible to use in places that aren't test suites

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
torkelrogstad: From the Scaladoc on the method: 
```
Sums up the value of all unspent
TXOs in the wallet, filtered by the given predicate
```

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressAndAccountHandling.scala
torkelrogstad: `isMine` sounds like a predicate. We could do a method like this:

```scala
def isMine(transaction: Transaction): Future[Boolean] = findOurOuts(transaction).map(_.nonEmpty)
```

?

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTxoDAO.scala
torkelrogstad: https://stackoverflow.com/questions/26140757/what-to-choose-between-require-and-assert-in-scala

Good SO question on exactly that. Can change if you still think its a bad idea after reading the answer

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
torkelrogstad: The reason for keeping them as two different method is just to keep the method smaller. Can change it if you feel strongly about it.

The reason for having a confirmations param in `processNewIncomingTx` is exactly what you posted yourself.

wallet-test/src/test/scala/org/bitcoins/wallet/models/SpendingInfoDAOTest.scala
```diff
@@ -17,17 +23,110 @@ class SpendingInfoDAOTest extends BitcoinSWalletTest with WalletDAOFixture {
...
+          val head = randomTX.inputs.head
+          val ourInput =
+            TransactionInput(utxo.outPoint,
+                             ScriptSignature.empty,
+                             head.sequence)
+
+          ourInput +: randomTX.inputs.tail
+        }
+
+        BaseTransaction(randomTX.version,
```
Christewart, line 69: nit: I don't think this matters in this case, but remember when you are going from a `Transaction` -- which can be a `WitnessTransaction`  -- to a `BaseTransaction` you may be discarding witness data.

wallet/src/main/scala/org/bitcoins/wallet/internal/AccountHandling.scala
```diff
@@ -0,0 +1,41 @@
...
+  override def listAccounts(): Future[Vector[AccountDb]] =
+    accountDAO.findAll()
+
+  /** @inheritdoc */
+  override protected[wallet] def getDefaultAccount(): Future[AccountDb] = {
+    for {
+      account <- accountDAO.read((DEFAULT_HD_COIN, 0))
+    } yield
+      account.getOrElse(
+        throw new RuntimeException(
```
Christewart, line 28: Perhaps `Future.failed` instead of a hard throw?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,268 @@
...
+  * spending from our wallet
+  */
+private[wallet] trait TransactionProcessing { self: LockedWallet =>
+  /////////////////////
+  // Public facing API
+
+  /** @inheritdoc */
+  override def processTransaction(
+      transaction: Transaction,
+      confirmations: Int): Future[LockedWallet] = {
```
Christewart, line 25: Should this return a `TransactionProcessing` instead of a `LockedWallet`?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,268 @@
...
+
+            Future
+              .sequence(txoProcessingFutures)
+
+        }
+
+    val outgoingTxFut: Future[Vector[SpendingInfoDb]] = {
+      for {
+        outputsBeingSpent <- spendingInfoDAO.findOutputsBeingSpent(transaction)
+        processed <- FutureUtil.sequentially(outputsBeingSpent)(
```
Christewart, line 98: Why do we we need to process this sequentially?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,268 @@
...
+          markAsSpentIfUnspent)
+      } yield processed.flatten.toVector
```
Christewart, line 100: So what does the `None` case mean that is being returned from `markAsSpentIfUnspent`? Does that mean that utxo was already spent? This seems like a log worth event, as it means that someone is trying to double spend a transaction that we funded, right? 

Perhaps the log need not happen here, but I think we should definitely log in `markAsSpentIfUnspent`

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,268 @@
...
+  /** If the given UTXO is marked as unspent, updates
+    * its spending status. Otherwise returns `None`.
+    */
+  private val markAsSpentIfUnspent: SpendingInfoDb => Future[
+    Option[SpendingInfoDb]] = { out =>
+    if (out.spent) {
+      Future.successful(None)
+    } else {
+      val updatedF =
+        spendingInfoDAO.update(out.copyWithSpent(spent = true))
```
Christewart, line 129: So I actually don't think we want to do the sql update here, we probably want to do it in the caller of this method, so that the transaction will be updated atomically. In the current implementation, we could have some `SpendingInfoDb` be marked as spent, while some others not marked as spent within the same transaction because this database call failed for whatever reason.

TLDR, we probably want to make this update using `.transactionally`.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -0,0 +1,268 @@
...
+    * Inserts the UTXO at the given index into our DB, swallowing the
+    * error if any (this is because we're operating on data we've
+    * already verified).
+    */
+  private def processUtxo(
+      transaction: Transaction,
+      index: Int,
+      spent: Boolean,
+      confirmations: Int): Future[SpendingInfoDb] =
+    addUtxo(transaction,
```
Christewart, line 149: nit: intermediate val here.

Christewart (commented):
In general, I think we need to work on making `TransactionProcessing` more `Transactional` in nature, so that we can't end up in indeterminant states at our database level. I'm not sure if you want to address this in this PR, or another one, but I think that is something we need to implement at some point.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
torkelrogstad: You're right, we probably don't. Changed.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
torkelrogstad: It only means that the TXO was already spent. Added a couple of logs at `debug` and `trace` level to `markAsSpentIfUnspent`

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
torkelrogstad: So the intention here is that `ProcessTxResult` is part of the internal API only, and used to for example log the result of processing a TX. An external user shouldn't need to be aware of any of those details. However, it should return `LockedWalletApi` instead of `LockedWallet`. So I changed that.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
torkelrogstad: Made a change where `markAsSpentIfUnspent` now takes in a sequence of TXOs, and calls `updateAll` in its DAO. `CRUD.updateAll` is also updated to process elements transactionally. 

Is this a solution you think works?

Issue #557, opened 2019-06-28
Calling isStartedF on a bitcoind instance does not work with cookie based auth


torkelrogstad:
**To Reproduce**
1. Use cookie based auth in your `$HOME/.bitcoin.conf` file (this is the default if you don't specify any auth options)
2. Call the `isStartedF` method

This causes the RPC client to crash because it builds a `ping` request, causes the cookie file to be read, which again causes this error: 
```
[E] Exception in thread "main" java.lang.RuntimeException: Could not find /home/torkel/.bitcoin/regtest/.cookie!
```

PR #558, opened 2019-06-28
Somewhat dirty standalone server and CLI binary

### LINKED ISSUE #536: Implement a basic MVP for external communication with wallet - server side
Umbrella issue: https://github.com/bitcoin-s/bitcoin-s/issues/530

--- 

We need some sort of external communication tool for our wallet (and later, node as well).

Some design goals for this communication protocol: 

- Should be accessible through a CLI
- Should be accessible through a standardized communication protocol (HTTP/gRPC/etc)
- Should be accessible through a JVM library interface
- _Nice to have_: small classpath footprint, so it can run on Android/ScalaJS
- _Nice to have_: using the same technology in node and wallet

## HTTP
One alternative is serving requests through a HTTP server.

Here are some tools for setting up a web server in Scala

  - [Scalatra](https://github.com/scalatra/scalatra/) 
  - [Akka-HTTP](https://doc.akka.io/docs/akka-http/current/index.html) - already on the classpath of `node`. Would make it harder to migrate away from the Akka ecosystem later.
  - [http4s](https://github.com/http4s/http4s) - depends on Cats
  - [Finch](https://finagle.github.io/finch/)
  - [Finatra](https://twitter.github.io/finatra/)

## gRPC
A different alternative that has broad ecosystem support is gRPC. Some alternatives here: 
  - [Mu](http://higherkindness.io/mu/) - some deps, supports gRPC
  - [ScalaPB](https://scalapb.github.io/grpc.html) - gRPC
  - [Akka-gRPC](https://github.com/akka/akka-grpc) - Same problem as Akka-HTTP from above

## RPC
A third alternative is just non-standardized RPC. These solutions look easy to implement and would run fine on both JVM and JS, but would make it hard for other people to **not** use our CLI or library.

  - [scala-json-rpc](https://github.com/shogowada/scala-json-rpc) - no deps, JVM/JS compatible. Not much activity on GitHub
  - [Sloth](https://github.com/cornerman/sloth) - RPC, supports JVM/JS

# Scope of this issue

1) Continue research on what technology we should go with, and settle on one
2) Implement a `getBalance` method for the wallet, that returns a simple JSON object looking like ` { "balance": 34503458 } `, or something similar. There's already a method in the wallet that gets the balance, so the hard part here is setting up the external communication protocol.

torkelrogstad:
Closes #536 

In this PR we introduce:

1. An extremely dirty standalone server for the node, chain and wallet projects
   - Wallet
     - `getnewaddress`
     - `getbalance`
     - `sendtoaddress` - note that this does not broadcast the TX anywhere, so doesn't really do much...
   - Node
     - `getpeers` - doesn't return any information about actual peers, just a faux implementation for the time being
   - Chain
     - `getblockcount`
     - `getbestblockhash`
2. En equally dirty CLI that can be used to interact with the wallet. It does validation of given args and commands, but this is by no means a finished thing. Did this quick and dirty to have something we can actually use out there. It can be compiled with GraalVM `native-image` (you need to install this separately, get it on [GitHub](https://github.com/oracle/graal/releases)) to a binary that can be executed.
3. A new project `picklers` that contain common serialization/deserialization logic for CLI and server

## Running the server

> You need to have a running `bitcoind` locally to run the sever, due to #557 we're not able to start it within the server quite yet

If you're running the CLI after compiling with `native-image`, it's easiest to run the server thrugh Bloop:
```bash
$ bloop run walletServer
```

If you're going to run the CLI through Bloop, you have to run it through sbt (Bloop can't run multiple things at the same time): 

```bash
$ sbt walletServer/run
```

### GraalVM `native-image`:

```bash
$ sbt cli/graalvm-native-image:packageBin
$ ./app/cli/target/graalvm-native-image/bitcoin-s-cli 
``` 

### Bloop

```bash
$ bloop run cli -- 
```

Note that you need to add `--` after the Bloop command _and then_ pass the actual arguments to the CLI, otherwise they get picked up as arguments to Bloop

### sbt
Just don't


torkelrogstad:
Update 2.7: I've made some changes to mimic the Bitcoin Core RPC API. As a consequence of this, I broke `sendtoaddress`. Working on a fix for that, but pushed changes now so you guys can have a look.

project/Deps.scala
```diff
@@ -55,6 +63,18 @@ object Deps {
...
+    val uPickle = "com.lihaoyi" %% "upickle" % V.uPickleV
+
+    // make akka-http play nice with upickle
+    val akkaHttpUpickle = "de.heikoseeberger" %% "akka-http-upickle" % V.akkaHttpUpickleV
+
+    // parsing of CLI opts and args
+    val scopt = "com.github.scopt" %% "scopt" % V.scoptV
+
+    // HTTP client lib
+    val sttp = "com.softwaremill.sttp" %% "core" % V.sttpV
```
Christewart, line 78: do we really need _another_ http client lib?

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -0,0 +1,131 @@
...
+import org.bitcoins.node.SpvNodeCallbacks
+
+object Main
+    extends App
+    // TODO we want to log to user data directory
+    // how do we do this?
+    with BitcoinSLogger {
+  implicit val conf = {
+    // val custom = ConfigFactory.parseString("bitcoin-s.network = testnet3")
+    BitcoinSAppConfig()
```
Christewart, line 36: I don't really like this syntax, it would be nicer if we could have something like `fromDefaultConf()` or `fromDefaultDatadir()` rather than just an empty param that looks like magic is happening when you read code

Christewart:
If you don't have the native packager installed, but you have graalvm installed, you can install the `native-image` tool with 

```
./gu install native-image
```

project/Deps.scala
torkelrogstad: Valid point, but in that case I'd say its a better idea to scrap akka-http as a client lib. 

Classpath of `akka-http` + `akka-stream` (need both):
```bash
$ coursier resolve com.typesafe.akka:akka-stream_2.12:2.5.23
com.typesafe:config:1.3.3:default
com.typesafe:ssl-config-core_2.12:0.3.7:default
com.typesafe.akka:akka-actor_2.12:2.5.23:default
com.typesafe.akka:akka-protobuf_2.12:2.5.23:default
com.typesafe.akka:akka-stream_2.12:2.5.23:default
org.reactivestreams:reactive-streams:1.0.2:default
org.scala-lang:scala-library:2.12.8:default
org.scala-lang.modules:scala-java8-compat_2.12:0.8.0:default
org.scala-lang.modules:scala-parser-combinators_2.12:1.1.1:default

$ coursier resolve com.typesafe.akka:akka-http_2.12:10.1.8
com.typesafe.akka:akka-http-core_2.12:10.1.8:default
com.typesafe.akka:akka-http_2.12:10.1.8:default
com.typesafe.akka:akka-parsing_2.12:10.1.8:default
org.scala-lang:scala-library:2.12.8:default
```

classpath of Sttp: 
```bash
$ coursier resolve com.softwaremill.sttp:core_2.12:1.1.13
com.softwaremill.sttp:core_2.12:1.1.13:default
org.scala-lang:scala-library:2.12.5:default
```

app/server/src/main/scala/org/bitcoins/server/Main.scala
torkelrogstad: I agree. File an issue and tackle in a separate PR? Maybe even `default()` would be nice name

Christewart:
As of 920ca62 i get this error
 ```
➜  bitcoin-s-core git:(920ca629a) ✗ bloop run walletServer
09:02:38.637 INFO Main$ - Using pre-existing wallet
09:02:38.786 ERROR WalletStorage$ - Encrypted mnemonic not found at /home/chris/.bitcoin-s/regtest/encrypted_bitcoin-s_seed.json
09:02:38.789 ERROR LockedWallet$LockedWalletImpl - Encrypted mnemonic not found when unlocking the wallet!
[E] Exception in thread "main" scala.MatchError: org.bitcoins.wallet.api.UnlockWalletError$MnemonicNotFound$: Mnemonic nout found (of class org.bitcoins.wallet.api.UnlockWalletError$MnemonicNotFound$)
[E]     at org.bitcoins.server.Main$.delayedEndpoint$org$bitcoins$server$Main$1(Main.scala:72)
[E]     at org.bitcoins.server.Main$delayedInit$body.apply(Main.scala:30)
[E]     at scala.Function0.apply$mcV$sp(Function0.scala:39)
[E]     at scala.Function0.apply$mcV$sp$(Function0.scala:39)
[E]     at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
[E]     at scala.App.$anonfun$main$1$adapted(App.scala:80)
[E]     at scala.collection.immutable.List.foreach(List.scala:392)
[E]     at scala.App.main(App.scala:80)
[E]     at scala.App.main$(App.scala:78)
[E]     at org.bitcoins.server.Main$.main(Main.scala:29)
[E]     at org.bitcoins.server.Main.main(Main.scala)
```

File: app/picklers/src/main/scala/org/bitcoins/picklers/Picklers.scala [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+import upickle.default._
+
+package object picklers {
+  import org.bitcoins.core.crypto.DoubleSha256DigestBE
+  implicit val bitcoinAddressPickler: ReadWriter[BitcoinAddress] =
+    readwriter[String]
+      .bimap(_.value, BitcoinAddress.fromStringExn(_))
+
+  implicit val bitcoinsPickler: ReadWriter[Bitcoins] =
+    readwriter[Double].bimap(_.toBigDecimal.toDouble, Bitcoins(_))
```
rorp, line 15: This line is OK for a POC, but absolutely can't be used in production. See https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

torkelrogstad: So in principle I agree with you 100%. The reason this is a double is that we need to represent bitcoins in the API (both as a return value and a parameter) to stay compatible with Bitcoin Core. Consider the `getbalance` RPC: 

```bash
$ bitcoin-cli getbalance
2963.87631780 # this is on regtest, in case you were wondering
```

Or the `sendtoaddress` RPC:

```bash
$ bitcoin-cli sendtoaddress <ADDRESS> 5.3
...
```

IMO the ideal solution here would be for all crypto APIs to use satoshis, but that's not the case...`

Christewart: ~Can we call `_.toBigDecimal.toDoubleExact()` so we get a exception thrown if we can't round correctly? This means we fail explicitly if we can't cleanly round. I think this is a nice middle ground~

It looks like the `BigDecimal` api does not support `toDoubleExact()`

app/server/src/main/scala/org/bitcoins/server/Main.scala
torkelrogstad: https://github.com/bitcoin-s/bitcoin-s/issues/588

PR #563, opened 2019-07-02
Add windows compatability to bitcoind RPC client

### LINKED ISSUE #562: Starting a bitcoin node cannot find path on Windows
When running `bloop run node` discovered that the `Bitcoindconfig` in `bitcoind-rpc` tries to find `.bitcoin` in the `C:\users\{current_user}\.bitcoin` folder by default. I'll write a default case for Windows users.

```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: C:\Users\rocky\.bitcoin does not exist!  
```

cwaldron97:
Currently trying to fix implementation to create Bitcoind instance on a windows machine. Default Directory is handled for Windows installation of bitcoin currently trying to fix issues with Regex difference between linux and DOS/Powershell. I will need to just commit to a format for or the other and try to get that working. Currently setting up Windows Terminal to see if you can just do it through that.

closes:#562

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindConfig.scala
@@ -331,7 +331,12 @@ object BitcoindConfig extends BitcoinSLogger {
                 "Library",
                 "Application Support",
                 "Bitcoin")
-    } else {
+    } else if (Properties.isWin) {
+      Paths.get("C:",
+        "Program Files",
+                        "Bitcoin")
+    }
+    else {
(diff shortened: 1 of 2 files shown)


cwaldron97:
Successfully initializes a chain and node if you have bitcoind.exe in the default location from out of box install. Otherwise tweaks will need to be made to `DEFAULT_DATADIR` in `BitcoindConfig`. There are 2 folders created when installing bitcoin on Windows, 1 which is contained in the location on the wiki has data generated after syncing and the current directory `DEFAULT_DATADIR` is pointing at which contains the bitcoind.exe file. That file is what is necessary to start up bitcoind on Windows.

Christewart (approved):

PR #565, opened 2019-07-03
Remove redundant function checkResult and refactor Numbers

piu130:
This PR removes the redundant and broken function `checkResult` from `Number`. The check is redundant, because calling apply already checks whether the number is in the required range.

Add new function `isInBound` to check whether the underlying number is in bound of the number.

Split `BaseNumber { zero, one, min, max }` in `BaseNumber { zero, one }` and `Bounded { min, max }`.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -34,7 +34,10 @@ sealed abstract class Number[T <: Number[T]]
     */
   def andMask: BigInt
 
-  /** Factory function to create the underlying T, for instance a UInt32 */
+  /**
+    * Factory function to create the underlying T, for instance a UInt32.
+    * This method must check if the parameter is in the required range.
+    */
   def apply: A => T
 
   override def +(num: T): T = apply(underlying + num.underlying)
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -148,6 +148,7 @@ sealed abstract class UInt64 extends UnsignedNumber[UInt64] {
     * Converts a [[BigInt]] to a 8 byte hex representation.
     * [[BigInt]] will only allocate 1 byte for numbers like 1 which require 1 byte, giving us the hex representation 01
     * this function pads the hex chars to be 0000000000000001
+    *
     * @param bigInt The number to encode
     * @return The hex encoded number
     */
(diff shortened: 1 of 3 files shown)


torkelrogstad:
Thanks for the contribution!

One thing which might make sense to do, is add tests that check explicitly for too small and large numbers. Something like this, in the case of `Int64`: 

```scala
  it must "throw on too small numbers" in {
    val bigint = BigInt("-9223372036854775809")
    val bigintBytes = ByteVector(bigint.toByteArray)
    an[IllegalArgumentException] must be thrownBy {
      Int64(bigint)
    }

    an[IllegalArgumentException] must be thrownBy {
      Int64.fromBytes(bigintBytes)
    }

    an[IllegalArgumentException] must be thrownBy {
      Int64.fromHex(bigintBytes.toHex)
    }
  }

  it must "throw on too large numbers" in {
    val bigint = BigInt("9223372036854775808")
    val bigintBytes = ByteVector(bigint.toByteArray)
    an[IllegalArgumentException] must be thrownBy {
      Int64(bigint)
    }

    an[IllegalArgumentException] must be thrownBy {
      Int64.fromBytes(bigintBytes)
    }

    an[IllegalArgumentException] must be thrownBy {
      Int64.fromHex(bigintBytes.toHex)
    }
  }
```

Another thing which might also be of interest, is a couple of commits I worked on a while ago without making a PR out of: https://github.com/torkelrogstad/bitcoin-s/commit/ca1f6ed8221211d937408d2f5f54d114a834832b and https://github.com/torkelrogstad/bitcoin-s/commit/e15a845ffab16b611bf966712af0b574e6bd719d. The general idea is to remove as many invariants as possible in `NumberType`, and have _one_ place that's able to check all invariants generically. If you think that approach is any good, that might make sense to incorporate here?

piu130:
@torkelrogstad I currently don't have the time to write tests for all those numbers. Sorry.

I think we should generalize the behaviour of 'call require in apply' by abstracting the method `isValid` in `UInt8` for all numbers. Then we can test whether isValid is called in the constructor.

torkelrogstad:
The `isValid` approach sounds good! Do you mind expanding your PR to incorporate that?

piu130:
@torkelrogstad I implemented `isValid` called `isInBound` and moved `min` and `max` to the new trait `Bounded`. `zero` and `one` are more identity elements while `min` and `max` are bounds for some but not all numbers.
This is also to prepare for the [Numeric](https://www.scala-lang.org/api/current/scala/math/Numeric.html) trait in Scala which contains `zero` and `one` but not `min` and `max`.

What do you think about it?

Should we remove the methods marked with `@deprecated`? Or do you prefer it this way?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -188,33 +182,50 @@ sealed abstract class Int64 extends SignedNumber[Int64] {
   override def hex: String = BitcoinSUtil.encodeHex(toLong)
 }
 
-/**
-  * Represents various numbers that should be implemented
-  * inside of any companion object for a number
-  */
+trait Bounded[T] {
```
torkelrogstad, line 190: Do you mind adding a short Scaladoc here? Something like "Represents number types that are bounded by minimum and maximum values"

torkelrogstad (changes requested):
Great work! I think just deleting the deprecated methods is fine, seeing as we haven't reached a 1.x release yet. I'm fine with merging in this PR after that.

piu130:
@torkelrogstad done.

I think you should add scalafmt to travis (if possible) so the build would fail, if the code is not formatted correctly.

torkelrogstad:
Looks like there's some weirdness going on in CI . I'll restart the build and see what happens.

torkelrogstad:
Looks like all of CI is screwed up now for some reason. Think I've seen the error message before, but can't remember how I solved it. I'll restart the build a bit later and see if anything changes, otherwise we'll just merge this.

torkelrogstad (approved):

torkelrogstad:
CI is having issues, so I pulled down and ran tests locally. LGTM!

PR #577, opened 2019-07-08
 Add functionality for broadcasting TXs to node

torkelrogstad:
In this PR we add functionality and tests
for broadcasting a TX from our node. To accomplish
this we introduce a table over broadcastable TXs
that's added to when the externally facing method
`broadcastTransaction(tx)` within `SpvNode` is called.
We send out a `inv` message for the TX we just added,
and upon receiving a `getdata` message we search in
the previously mentioned table for entries where
the hashes match up.

I have created a test that checks that the TX was broadcast correctly. Didn't have much time to manually test this today, but I was having some issues getting this to work when using the server/CLI. Might be a problem with my `bitcoind` configuration.


node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    implicit val nodeConfig: NodeAppConfig = config
+    implicit val chainConfig: ChainAppConfig = config
+
+    def hasSeenTx(transaction: Transaction): Future[Boolean] = {
+      rpc
+        .getRawTransaction(transaction.txIdBE)
+        .map { _ =>
+          true
+        }
+        .recover {
```
Christewart, line 44: nit: It actually might be nice to adjust the `getRawTransaction()` rpc call to incorporate this logic and return a `Future[Option[...]]` cc: @cwaldron97

node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
```diff
@@ -0,0 +1,88 @@
...
+        spv.start()
+      }
+      _ <- spv.sync()
+      _ <- NodeTestUtil.awaitSync(spv, rpc)
+
+      tx <- wallet
+        .sendToAddress(address, 1.bitcoin, SatoshisPerByte(10.sats))
+
+      _ = spv.broadcastTransaction(tx)
+      _ <- Future { Thread.sleep(5.seconds.toMillis) }
```
Christewart: Can we incorporate this Thread.sleep in the awaitConditionF below? Just increase the timeout?

node/src/main/scala/org/bitcoins/node/models/BroadcastAbleTransactionTable.scala
```diff
@@ -0,0 +1,43 @@
...
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.ProvenShape
+import org.bitcoins.db.DbRowAutoInc
+import org.bitcoins.db.TableAutoInc
+import org.bitcoins.core.crypto.DoubleSha256Digest
+import scodec.bits.ByteVector
+import org.bitcoins.core.protocol.transaction.Transaction
+
+/** TXs we can broadcast over the P2P network */
+final case class BroadcastAbleTransaction(
```
Christewart, line 12: I'm not sure if this abstraction is right, perhaps it would be better to called this `SignedTransaction` and then have a field in this that indicates if `isBroadcasted: Boolean`, and then maybe a `isConfirmed` or something like that. Although this starts to get redundant with confirmations else where in our data model. Is the idea here we delete txs from this table after they have been broadcasted?

Christewart: Thinking about this further, there seems to be another state, something like "sent inv", which does not necessarily mean it is broadcasted. This is also peer specific which makes it more complex. Maybe this is making things too complex for this PR, and we should address separately.

node/src/main/scala/org/bitcoins/node/networking/peer/InventoryEsque.scala
```diff
@@ -0,0 +1,31 @@
...
+  def toInv(elem: T): Inventory
+
+}
+
+/** Default implementations of things that can be sent as
+  * a inventory message
+  */
+object InventoryEsque {
+
+  implicit val transactionInventoryEsque = new InventoryEsque[Transaction] {
```
Christewart: I don't really see the added benefit here, I think we should default to not using implicits as it makes things more complex and harder to reason about.

node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
torkelrogstad: That's not how Core does it, though. If we're going to start applying that sort of logic there's _a lot_ of places that can happen. And if we're 100% consistent with Core some places but take liberties other places, we end up with a very confusing library.

node/src/main/scala/org/bitcoins/node/models/BroadcastAbleTransactionTable.scala
torkelrogstad: My reasoning here is that we might not broadcast a TX only once. Our peer (or peers, once that's implemented) might ask for the same TX again. Would be bad if we couldn't send them to them?

node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
cwaldron97: I don't know if it relevant but in V18 `getRawTransaction` fundamentally changes how it works and no longer check unspent `UTXO` for a transaction. 
`"The remaining behaviors are as follows: 1. If a blockhash is provided, check the corresponding block. 2. If no blockhash is provided, check the mempool. 3. If no blockhash is provided but txindex is enabled, also check txindex."`

torkelrogstad: We're looking for transactions in the mempool (the mempool are the set of transactions a node has received, but not yet confirmed in a block), so I don't think this matters to us here.

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -86,6 +94,22 @@ case class SpvNode(
     isStoppedF.map(_ => this)
   }
 
+  /** Broadcasts the given transaction over the P2P network */
+  def broadcastTransaction(transaction: Transaction) = {
```
Christewart: return type

node/src/main/scala/org/bitcoins/node/models/BroadcastAbleTransactionTable.scala
```diff
@@ -0,0 +1,43 @@
...
+      require(tx.txId == txid)
+      BroadcastAbleTransaction(tx, id)
+  }
+
+  private val toTuple: BroadcastAbleTransaction => Option[Tuple] = tx =>
+    Some(tx.transaction.txId, tx.transaction.bytes, tx.id)
+
+  import org.bitcoins.db.DbCommonsColumnMappers._
+
+  def txid: Rep[DoubleSha256Digest] = column("txid", O.Unique)
```
Christewart: Are we storing the `BE` representation or `LE` representation else where? Let's just make sure we are consistent with how we are storing these.

node/src/main/scala/org/bitcoins/node/models/BroadcastAbleTransactionTable.scala
Christewart: I don't think that is the expected behavior, and we should probably consider the peer at best as broken, at worst as malicious if they request the transaction multiple times. Here is the algorithm bitcoin core uses for _requesting_ information about a inv tx they received: https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L295

Here is how bitcoin core processes an _incoming_ TX request on the p2p network. I can't tell what there behavior is from a cursory glance, but we definitely want to at least look at this https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L2470

Christewart (changes requested):

node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
```diff
@@ -0,0 +1,90 @@
...
+        .map { _ =>
+          true
+        }
+        .recover {
+          case err: Throwable
+              if err.getMessage.contains(
+                s"No such mempool or blockchain transaction") =>
+            false
+          case other =>
+            logger.info(s"Received other error: $other")
```
nkohen: `logger.error` and a more descriptive prefix?

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -140,13 +141,43 @@ trait BitcoinSWalletTest
...
     makeDependentFixture(builder, destroy = destroyWalletWithBitcoind)(test)
+
+  }
+
+  def withFundedWalletAndBitcoind(test: OneArgAsyncTest): FutureOutcome = {
+    val builder: () => Future[WalletWithBitcoind] =
+      composeBuildersAndWrapFuture(
+        builder = createDefaultWallet _,
+        dependentBuilder = createWalletWithBitcoind,
+        processResult = (_: UnlockedWalletApi, pair: WalletWithBitcoind) => {
```
nkohen: I think it would be nice to have this as a separate `def`

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -16,32 +16,51 @@ import org.bitcoins.core.p2p.TransactionMessage
...
       payload: DataPayload,
       peerMsgSender: PeerMessageSender): Future[Unit] = {
 
     payload match {
+      case getData: GetDataMessage =>
+        logger.debug(
+          s"Received a getdata message for inventories=${getData.inventories}")
+        getData.inventories.foreach { inv =>
+          logger.debug(s"Looking for inv=$inv")
+          txDAO.findByHash(inv.hash).map {
```
Christewart: I think we want to avoid database accesses for every single `GetDataMessage` we receive. I think something like this is superior 

```scala
        getData.inventories.foreach { inv =>
          logger.debug(s"Looking for inv=$inv")
          inv.typeIdentifier match {
            case TypeIdentifier.MsgTx =>
              txDAO.findByHash(inv.hash).map {
                case Some(tx) =>
                  peerMsgSender.sendTransactionMessage(tx.transaction)
                case None =>
                  logger.warn(
                    s"Got request to send data with hash=${inv.hash}, but found nothing")
              }
            case TypeIdentifier.MsgBlock | TypeIdentifier.MsgFilteredBlock |
                TypeIdentifier.MsgCompactBlock |
                TypeIdentifier.MsgFilteredWitnessBlock |
                TypeIdentifier.MsgWitnessBlock | TypeIdentifier.MsgWitnessTx =>
            //not sure what we want to do here on this PR...

            case unassigned: MsgUnassigned =>
              logger.warn(
                s"Received unassigned message we do not understand, msg=${unassigned}")
          }

        }
```

PR #590, opened 2019-07-10
Add notion of address types in wallet

torkelrogstad:
In this commit we
1) Add the notion of address types in the wallet, and remove
    the emphasis on account types. Users now just request
    a segwit/nested-segwit/legacy address, and we take care
    of querying for the correct account
2) Fix a bug where a wallet could only get addresses for
    its default address type. This was a pretty minor bug,
    where a few values that should have been dynamic was
    hard coded.

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
@@ -11,6 +11,9 @@ import org.bitcoins.wallet.api.UnlockWalletError.MnemonicNotFound
 import com.typesafe.config.ConfigFactory
 import org.bitcoins.core.protocol.P2PKHAddress
 import org.bitcoins.core.hd.HDPurposes
+import org.bitcoins.core.protocol.Bech32Address
+import org.bitcoins.core.protocol.P2SHAddress
+import org.bitcoins.core.hd.AddressType
 
 class LegacyWalletTest extends BitcoinSWalletTest {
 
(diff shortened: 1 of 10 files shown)


wallet/src/main/scala/org/bitcoins/wallet/AddressType.scala
```diff
@@ -0,0 +1,18 @@
+package org.bitcoins.wallet
+
+/** The address types our wallet can generate */
+sealed abstract class AddressType
+
+object AddressType {
```
Christewart: Perhaps move into core?

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -135,12 +135,20 @@ private[wallet] trait AddressHandling { self: LockedWallet =>
...
-  override def getNewAddress(account: AccountDb): Future[BitcoinAddress] = {
-    val addrF = getNewAddressHelper(account, HDChainType.External)
+  def getNewAddress(account: AccountDb): Future[BitcoinAddress] = {
+    val addrF =
+      getNewAddressHelper(account, HDChainType.External)
     addrF
   }
 
+  /** @inheritdoc */
+  def getNewAddress(addressType: AddressType): Future[BitcoinAddress] = {
```
Christewart: nit: `override`

Christewart (changes requested):

wallet-test/src/test/scala/org/bitcoins/wallet/LegacyWalletTest.scala
```diff
@@ -24,13 +26,39 @@ class LegacyWalletTest extends BitcoinSWalletTest {
...
+    for {
+      segwit <- wallet.getNewAddress(AddressType.SegWit)
+      legacy <- wallet.getNewAddress(AddressType.Legacy)
+      // TODO: uncomment this once nested segwit is implemented
+      // nested <- wallet.getNewAddress(AddressType.NestedSegWit)
+    } yield {
+      assert(segwit.isInstanceOf[Bech32Address])
+      assert(legacy.isInstanceOf[P2PKHAddress])
+      // TODO: uncomment this once nested segwit is implemented
+      // assert(nested.isInstanceOf[P2SHAddress])
```
nkohen, line 64: Does this have an issue?

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -83,12 +84,14 @@ trait LockedWalletApi extends WalletApi {
...
-    * Gets a new external address from the specified
-    * account. Calling this method multiple
+    * Gets a new external address with the specified
+    * type. Calling this method multiple
     * times will return the same address, until it has
     * received funds.
+    *
     */
-  def getNewAddress(account: AccountDb): Future[BitcoinAddress]
+  // TODO: Last sentence is not true, implement that
```
nkohen, line 93: issue?

PR #591, opened 2019-07-10
Node cleanup

torkelrogstad:
In this PR we
1) Rename Client to P2PClient. Client is a very generic name.
2) Clean up some Scaladocs
3) Remove some unecessary objects/traits/classes and replace them
    with case classes
4) Add trace logging of bytes received in P2PClient


node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -59,12 +60,12 @@ sealed abstract class ClientActor extends Actor with BitcoinSLogger {
 
   /**
     * The parameters for the network we are connected to
-    * i.e. [[org.bitcoins.core.config.MainNet]] or [[org.bitcoins.core.config.TestNet3]]
-    * @return
     */
-  def network: NetworkParameters = config.network
+  val network: NetworkParameters = config.network
```
cwaldron97, line 64: Why do we make the change from a `def` to a `val`? Would a `lazy val` not be preferred?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
```diff
@@ -57,17 +57,13 @@ class PeerMessageSender(client: Client)(implicit np: NetworkParameters)
...
     logger.debug(s"Sending msg=${msg.commandName} to peer=${socket}")
-    val newtworkMsg = NetworkMessage(np, msg)
+    val newtworkMsg = NetworkMessage(conf.network, msg)
     client.actor ! newtworkMsg
   }
 }
 
 object PeerMessageSender {
 
-  private case class PeerMessageSenderImpl(client: Client)(
```
cwaldron97, line 67: I noticed a few instances of `impl` being removed. Are we just replacing those with `serializers`? Is that the preferred coding practice?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
torkelrogstad: My reasoning is:
- try val
- if it crashes, try lazy val
- if it still crashes, use def

There's sometimes initialization issues with vals and lazy vals, defs don't suffer from this. But defs are the least performant, and lazy vals are slightly less performant than vals.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
torkelrogstad: This is pattern happens a bunch of places in our code base:
- declare a trait
- make a companion object
- create a private case class that implemented the trait, and return this case class from the companion object methods.

Sometimes this is a necessary thing, but a bunch of times this isn't. So I'm trying to move away from this where we don't need it, as it is very verbose and causes more allocation of objects.

PR #594, opened 2019-07-10
Add all wallet outpoints to bloom filter

torkelrogstad:
We add all outpoints to the bloom filter as a way
of working around the fact that bloom filters
was never updated to incorporate SegWit changes.
see this mailing list thread for context: https://www.mail-archive.com/bitcoin-dev@lists.linuxfoundation.org/msg06950.html

especially this email from Jim Posen: https://www.mail-archive.com/bitcoin-dev@lists.linuxfoundation.org/msg06952.html

Once #590 goes in I can add a test to this that verifies we're notified about both legacy and segwit addresses being spent to.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
@@ -14,6 +14,7 @@ import org.bitcoins.core.bloom.BloomFilter
 import org.bitcoins.core.bloom.BloomUpdateAll
 import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.wallet.internal._
+import org.bitcoins.core.protocol.transaction.TransactionOutPoint
 
 abstract class LockedWallet
     extends LockedWalletApi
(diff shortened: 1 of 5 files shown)


wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -82,25 +83,33 @@ abstract class LockedWallet
...
 
   /** Gets the size of the bloom filter for this wallet  */
-  private def getBloomFilterSize(): Future[Int] = {
-    for {
-      pubkeys <- listPubkeys()
-    } yield {
+  private def getBloomFilterSize(
+      pubkeys: Seq[ECPublicKey],
+      outpoints: Seq[TransactionOutPoint]): Int = {
+    {
```
rorp: No need for an extra pair of brackets.

Christewart (commented):
Can we add a unit test verifying that this works?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -96,4 +97,10 @@ case class SpendingInfoDAO()(
 
     database.run(query.result).map(_.toVector)
   }
+
+  /** Enumerates all TX outpoints in the wallet */
+  def findAllOutpoints(): Future[Vector[TransactionOutPoint]] = {
+    val query = table.map(_.outPoint)
+    database.run(query.result).map(_.toVector)
```
nkohen: nit: `database.runVec`

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -110,7 +119,8 @@ abstract class LockedWallet
                     falsePositiveRate = walletConfig.bloomFalsePositiveRate,
                     flags = bloomFlag)
 
-      pubkeys.foldLeft(baseBloom) { _.insert(_) }
+      val withPubs = pubkeys.foldLeft(baseBloom) { _.insert(_) }
+      outpoints.foldLeft(withPubs) { _.insert(_) }
```
nkohen, line 124: Is there any reason we can't have some kind of persistent bloom filter lying around so that we don't have to reconstruct the whole thing each time we add a new outpoint? (I understand that it needs to be reconstructed from scratch if anything ever gets removed, but for insertions this should be constant time not linear/quadratic.)

torkelrogstad: I guess that would be an improvement, but that sounds like introducing a lot of complexity for a very limited gain. Here's a quick benchmark of inserting 10K pubkeys and outpoints: 

```scala
    val pubkeys = (1 to 10000).map(_ => CryptoGenerators.publicKey.sampleSome)
    val outpoints =
      (1 to 10000).map(_ => TransactionGenerators.outPoint.sampleSome)

    val start = Deadline.now
    val baseBloom = BloomFilter(30000, 0.001, BloomUpdateAll)

    val end = Deadline.now
    val withPubs = pubkeys.foldLeft(baseBloom) { _.insert(_) }
    val withPoints = outpoints.foldLeft(withPubs) { _.insert(_) }
    println(s"took time: ${(end - start).toMillis} millis")
```

When executing this with bloop, I get `took time: 20 millis` (the actual execution takes  a lot longer, but that's spent on generating pubkeys and outpoints). So I don't think this is a big issue. Unless you have a nice way of implementing it

PR #600, opened 2019-07-11
Revert payment monitor

rorp:


torkelrogstad (approved):
Would be nice with a description in the PR with the motivation for the changes. Otherwise, LGTM

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -568,4 +667,47 @@ class EclairRpcClient(val instance: EclairInstance)(
...
+
+    val runnable = new Runnable() {
+
+      private var attempt = new AtomicInteger(1)
+
+      override def run(): Unit = {
+        for {
+          results <- getSentInfo(paymentId)
+        } yield {
+          results.find(_.status != PaymentStatus.PENDING) match {
```
Christewart: Why not `results.find(_.status == PaymentStatus.SUCCEEDED)`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala
```diff
@@ -136,6 +138,12 @@ trait EclairApi {
 
   def payInvoice(invoice: LnInvoice, amountMsat: Option[MilliSatoshis], maxAttempts: Option[Int], feeThresholdSat: Option[Satoshis], maxFeePct: Option[Int]): Future[PaymentId]
 
+  def monitorSentPayment(paymentId: PaymentId, notifyPayment: PaymentResult => Future[Unit]): Future[PaymentResult]
```
Christewart: scaladocs on these

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -568,4 +667,47 @@ class EclairRpcClient(val instance: EclairInstance)(
...
+            case Some(result) =>
+              notifyPayment(result).foreach(_ => p.success(result))
+            case None =>
+              if (attempt.incrementAndGet() > 60) {
+                p.failure(new RuntimeException(
```
Christewart: Does this actually fail the payment in eclair itself? I feel like we shouldn't fail the payment attempt until eclair fails it internally. Shouldn't we do something with
 ```scala 
results.find(_.status == PaymentStatus.FAILED)`  match { ... } 
```

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala (hunk 1 above)
Comment (rorp) [reply eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala]: Because the payment can fail.

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala (hunk 3 above)
Comment (rorp) [reply eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala]: We don't want to wait forever if Eclair doesn't respond for some reason.

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala (hunk 1 above)
Comment (Christewart) [reply eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala]: So here is the documentation from eclair

https://acinq.github.io/eclair/

```
Returns a list of attempts to send an outgoing payment, the possible statuses are PENDING, FAILED and SUCCEEDED. The API can be queried by paymentHash OR by uuid, if the latter is used the API returns a list containing at most one element.
```
That means this implementation will find the _first_ attempt at sending the payment, not the last one. I guess this opens a question here of which one we are interested in. It seems we want to search for a `SUCCESS` until the timeout occurrs? I still am not convinced that this is the right solution

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala (hunk 3 above)
Comment (Christewart) [reply eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala]: This seems like something that the user of the library should determine? Maybe this is some sort of a config?

**Review by Christewart — CHANGES_REQUESTED** [review]

File: eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala [hunk 4]
```diff
@@ -663,22 +663,28 @@ class EclairRpcClient(val instance: EclairInstance)(
...
 
       override def run(): Unit = {
-        val resultsF = getSentInfo(paymentId)
-        resultsF.recover { case e: Throwable => logger.error("Cannot check payment status", e) }
-        for {
-          results <- resultsF
-        } yield {
-          results.find(_.status != PaymentStatus.PENDING) match {
-            case Some(result) =>
+        if (attempts.incrementAndGet() > maxAttempts) {
```
Christewart, line 676: This gets rid of a compiler warning, adds the `paymentId`/`interval` to the log message for failing the future. This also updates the scaladoc to talk about the completion scenarios for the future we are returning  

```scala
  /**
    * Pings eclair to see if a invoice has been paid
    * If the invoice has been paid or the payment has failed, we publish a
    * [[org.bitcoins.eclair.rpc.json.PaymentResult PaymentResult]]
    * event to the [[akka.actor.ActorSystem ActorSystem]]'s
    * [[akka.event.EventStream ActorSystem.eventStream]]
    *
    * We also return a Future[PaymentResult] that is completed when one of three things is true
    * 1. The payment has succeeded
    * 2. The payment has failed
    * 3. We have attempted to query the eclair more than maxAttempts, and the payment is still pending
    */
  override def monitorSentPayment(
      paymentId: PaymentId,
      interval: FiniteDuration,
      maxAttempts: Int): Future[PaymentResult] = {
    val p: Promise[PaymentResult] = Promise[PaymentResult]()

    val runnable = new Runnable() {

      private val attempts = new AtomicInteger(0)

      override def run(): Unit = {
        if (attempts.incrementAndGet() > maxAttempts) {
          // too many tries to get info about a payment
          // either Eclair is down or the payment is still in PENDING state for some reason
          // complete the promise with an exception so the runnable will be canceled
          p.failure(
            new RuntimeException(
              s"EclairApi.monitorSentPayment() too many attempts: ${attempts
                .get()} for paymentId=${paymentId} for interval=${interval}"))
        } else {
          val resultsF = getSentInfo(paymentId)
          resultsF.recover {
            case e: Throwable => logger.error(s"Cannot check payment status for paymentId=${paymentId}", e)
          }
          val _ = for {
            results <- resultsF
          } yield {
            results.foreach { result =>
              result.status match {
                case PaymentStatus.PENDING =>
                //do nothing, while we wait for eclair to attempt to process
                case PaymentStatus.SUCCEEDED | PaymentStatus.FAILED =>
                  // invoice has been succeeded or has failed, let's publish to event stream
                  // so subscribers to the event stream can see that a payment
                  // was received or failed
                  // complete the promise so the runnable will be canceled
                  p.success(result)
              }
            }
          }
        }
      }
    }

    val cancellable = system.scheduler.schedule(interval, interval, runnable)

    val f = p.future

    f.onComplete(_ => cancellable.cancel())
    f.foreach(system.eventStream.publish)

    f
  }
```

PR #601, opened 2019-07-11
Publish all artifacts

rorp:


build.sbt
```diff
@@ -287,13 +287,15 @@ lazy val coreTest = project
 lazy val walletServer = project
     .in(file("app/server"))
     .settings(commonSettings: _*)
+    .settings(name := "bitcoin-s-server")
```
torkelrogstad, line 290: This setting is already applied in `server.sbt`, so can be removed from here

build.sbt
```diff
@@ -287,13 +287,15 @@ lazy val coreTest = project
...
     .dependsOn(
       picklers,
       node,
       chain, 
       wallet,
       bitcoindRpc
     )
+  .enablePlugins(GitVersioning)
```
torkelrogstad, line 298: As far as I can tell, we don't actually use this plugin for anything. Versioning of our builds are handled by `sbt-dynver`.

build.sbt
```diff
@@ -307,15 +309,20 @@ lazy val walletServerTest = project
 // and CLI
 lazy val picklers = project
     .in(file("app/picklers"))
+    .settings(name := "bitcoin-s-picklers")
+    .settings(commonSettings: _*)
     .dependsOn(core % testAndCompile)
+  .enablePlugins(GitVersioning)
```
torkelrogstad, line 315: Same as comment on plugin above

build.sbt
```diff
@@ -307,15 +309,20 @@ lazy val walletServerTest = project
 // and CLI
 lazy val picklers = project
     .in(file("app/picklers"))
+    .settings(name := "bitcoin-s-picklers")
```
torkelrogstad, line 312: Duplicate name settings

build.sbt
```diff
@@ -307,15 +309,20 @@ lazy val walletServerTest = project
...
+    .settings(commonSettings: _*)
     .dependsOn(core % testAndCompile)
+  .enablePlugins(GitVersioning)
 
 
 lazy val cli = project
     .in(file("app/cli"))
+    .settings(name := "bitcoin-s-cli")
```
torkelrogstad, line 320: Duplicate name setting

build.sbt
```diff
@@ -307,15 +309,20 @@ lazy val walletServerTest = project
...
     .settings(commonSettings: _*)
     .dependsOn(
       picklers
     )
+  .enablePlugins(GitVersioning)
```
torkelrogstad, line 325: Same comment on plugins

torkelrogstad (commented):
I assume the motivation for this is test testkit needs a class found in `walletServer`? Would be nice to mention that in the PR description. 

The CLI, picklers and wallet server aren't intended for use as libraries, so ideally I'd prefer to not publish them as artifacts. However, we do use things from `walletServer` in testkit... Do you have any thoughts on what the best solution to that problem is?

torkelrogstad:
Needs rebase

PR #617, opened 2019-07-12
Replace akka-http with sttp

### LINKED ISSUE #589: Consider replacing akka-http with sttp
Classpath of `akka-http` + `akka-stream` (need both):
```bash
$ coursier resolve com.typesafe.akka:akka-stream_2.12:2.5.23
com.typesafe:config:1.3.3:default
com.typesafe:ssl-config-core_2.12:0.3.7:default
com.typesafe.akka:akka-actor_2.12:2.5.23:default
com.typesafe.akka:akka-protobuf_2.12:2.5.23:default
com.typesafe.akka:akka-stream_2.12:2.5.23:default
org.reactivestreams:reactive-streams:1.0.2:default
org.scala-lang:scala-library:2.12.8:default
org.scala-lang.modules:scala-java8-compat_2.12:0.8.0:default
org.scala-lang.modules:scala-parser-combinators_2.12:1.1.1:default

$ coursier resolve com.typesafe.akka:akka-http_2.12:10.1.8
com.typesafe.akka:akka-http-core_2.12:10.1.8:default
com.typesafe.akka:akka-http_2.12:10.1.8:default
com.typesafe.akka:akka-parsing_2.12:10.1.8:default
org.scala-lang:scala-library:2.12.8:default
```

classpath of Sttp: 
```bash
$ coursier resolve com.softwaremill.sttp:core_2.12:1.1.13
com.softwaremill.sttp:core_2.12:1.1.13:default
org.scala-lang:scala-library:2.12.5:default
```

_Originally posted by @torkelrogstad in https://github.com/bitcoin-s/bitcoin-s/pull/558_

torkelrogstad:
Closes #589.

In this PR we replace `akka-http` in the Eclair
RPC client and the `bitcoind` RPC client with with
Sttp. Sttp is a Scala HTTP library that acts as a wrapper around other HTTP clients, giving a nice and user friendly interface. For our use case, we have two choices: 

1) Go with the default HTTP client include in Sttp Core, [`java.net.HttpURLConnection`](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html). This gives a _very_ small classpath, but it is a blocking HTTP client that doesn't handle errors very well. 
2) Use the provided `async-http-client-backend-future`. This wraps [`async-http-client`](https://github.com/AsyncHttpClient/async-http-client) which again is built on Netty. 

I went with choice number 2. I can do some investigating on whether or not choice 1 or 2 is best. Regardless of going with 1 or 2, there are some benefits here compared to using Akka: 
    1) Smaller classpath. `akka-http` pulls in `akka-streams`,
        leading to a very big classpath
    2) No need for 3rd parties to provide a `ActorSystem`.
        Now the only thing that's needed is an
        `ExecutionContext`!

For some reason, Eclair tests run very slowly. There's also a warning message printed (see below). I believe the warnings and slow Eclair tests might be related to my reimplementation of AsyncUtil: https://github.com/bitcoin-s/bitcoin-s/compare/master...torkelrogstad:2019-07-12-sttp?expand=1#diff-a9be7d4ae386bc69045a2f8b014f0e12R98. Open to suggestions here. 

Todo: 
 - [ ] Update documentation
 - [ ] Figure out why Eclair tests are taking so long to instantiate
 - [ ] Investigate warnings printed by our retry mechanism:
```
16:03:25.969 ERROR HashedWheelTimer - You are creating too many HashedWheelTimer instances. HashedWheelTimer is a shared resource that must be reused across the JVM,so that only a few instances are created.
```

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/AsyncUtil.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/AsyncUtil.scala
@@ -1,15 +1,15 @@
 package org.bitcoins.rpc.util
 
-import akka.actor.ActorSystem
-import org.bitcoins.core.util.BitcoinSLogger
-
 import scala.concurrent._
-import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.concurrent.duration._
 
-abstract class AsyncUtil extends BitcoinSLogger {
+abstract class AsyncUtil {
   import AsyncUtil.DEFAULT_INTERNVAL
   import AsyncUtil.DEFAULT_MAX_TRIES
 
+  /** Returns a runnable that completes the
+    * given promise with the given condition
+    */
   private def retryRunnable(
       condition: => Boolean,
       p: Promise[Boolean]): Runnable = new Runnable {
@@ -23,7 +23,7 @@ abstract class AsyncUtil extends BitcoinSLogger {
       condition: => Boolean,
       duration: FiniteDuration = DEFAULT_INTERNVAL,
       maxTries: Int = DEFAULT_MAX_TRIES)(
-      implicit system: ActorSystem): Future[Unit] = {
+      implicit ec: ExecutionContext): Future[Unit] = {
     val f = () => Future.successful(condition)
     retryUntilSatisfiedF(f, duration, maxTries)
   }
@@ -40,7 +40,7 @@ abstract class AsyncUtil extends BitcoinSLogger {
       conditionF: () => Future[Boolean],
       duration: FiniteDuration = DEFAULT_INTERNVAL,
       maxTries: Int = DEFAULT_MAX_TRIES)(
-      implicit system: ActorSystem): Future[Unit] = {
+      implicit ec: ExecutionContext): Future[Unit] = {
     val stackTrace: Array[StackTraceElement] =
       Thread.currentThread().getStackTrace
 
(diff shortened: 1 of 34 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/AsyncUtil.scala
```diff
@@ -97,7 +95,10 @@ abstract class AsyncUtil extends BitcoinSLogger {
         val p = Promise[Boolean]()
         val runnable = retryRunnable(condition, p)
 
-        system.scheduler.scheduleOnce(duration, runnable)
+        Future {
+          Thread.sleep(duration.toMillis)
```
nkohen, line 99: Isn't this blocking? I agree with your description that this change is likely what is causing problems in tests.

nkohen: (In the sense that it keeps Threads hostage)

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -466,34 +459,27 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-  private def sendRequest(req: HttpRequest): Future[HttpResponse] = {
-    val respF = Http(m.system).singleRequest(req)
-    respF
-  }
-
   private def buildRequest(
       instance: EclairInstance,
       methodName: String,
-      params: (String, String)*): HttpRequest = {
+      params: (String, String)*): Request[String, Nothing] = {
```
nkohen, line 562: We seem to use `Request[String, Nothing]` a lot now and it isn't really clear what that means from the name, I think we should use a type alias for this with a better name

testkit/src/main/scala/org/bitcoins/testkit/util/BitcoindRpcTest.scala
```diff
@@ -1,22 +1,25 @@
...
   protected val logger: Logger = LoggerFactory.getLogger(getClass)
 
-  implicit val system: ActorSystem =
-    ActorSystem(getClass.getSimpleName, BitcoindRpcTestUtil.AKKA_CONFIG)
-  implicit val ec: ExecutionContext = system.dispatcher
+  /** AsyncFlatSpec provides its own execution context, but for some
+    * reason that doesn't allow doing anything in `afterAll`. We add
+    * our own.
+    */
+  implicit val ec: ExecutionContext = ExecutionContext.Implicits.global
```
nkohen, line 21: It worries me to see so many things using this `ExecutionContext` now that we don't have `ActorSystem`s because if this gets backlogged really bad things happen (generally deadlock is inevitable in these cases)

torkelrogstad:
Did some basic (and unscientific) benchmarking of performance with Sttp versus Akka-HTTP. I compiled everything with bloop (`bloop compile bitcoindRpcTest-test`), and then did `bloop test bitcoindRpcTest` on both this branch and master. 

Tests on this branch: 
- 6.18.82
- 6.02.39
- 6.29.53

Tests on master: 
- 6.08.51
- 5.52.57
- 6.04.57

Issue #539, opened 2019-06-20
Make RPC `sendToAddress` take `CurrencyUnit`


torkelrogstad:
Current method definition: https://bitcoin-s.org/api/org/bitcoins/rpc/client/common/BitcoindRpcClient.html#sendToAddress(address:org.bitcoins.core.protocol.BitcoinAddress,amount:org.bitcoins.core.currency.Bitcoins,localComment:String,toComment:String,subractFeeFromAmount:Boolean):scala.concurrent.Future[org.bitcoins.core.crypto.DoubleSha256DigestBE]

Would be more ergonomic if the signature looked like this instead: 

```scala
 sendToAddress(address: BitcoinAddress, amount: CurrencyUnit, localComment: String = "", toComment: String = "", subractFeeFromAmount: Boolean = false): Future[DoubleSha256DigestBE] 
```

cwaldron97:
Currently looks like we implement it like this 
```
def sendToAddress(
      address: BitcoinAddress,
      amount: CurrencyUnit,
      feeRate: FeeUnit,
      fromAccount: AccountDb): Future[Transaction]
```
does that mean CurrencyUnit was already taken care of and we just need to add the comments?

torkelrogstad:
You're looking at the [`sendToAddress`](https://github.com/bitcoin-s/bitcoin-s/blob/9101aece9b9e48df741e668fce66edbf02f59afc/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L38) method in our wallet (`Wallet.scala`/`WalletApi.scala`), but I'm talking atbout [`sendToAddress`](https://github.com/bitcoin-s/bitcoin-s/blob/4dbb57467ce3b469268b7ddb9faecfc736a87d08/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/TransactionRpc.scala#L149) in our RPC client.

PR #618, opened 2019-07-12
Replace Bitcoins params in RPC client with CurrencyUnit

cwaldron97:
Changed `amount` from `Bitcoins` to `CurrencyUnit` in `sendToAddress` and implemented the implicit writer associated. 

Addresses Issue #539

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/TransactionRpc.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/TransactionRpc.scala
@@ -7,7 +7,6 @@ import org.bitcoins.core.protocol.blockchain.MerkleBlock
 import org.bitcoins.rpc.client.common.RpcOpts.{AddressType, FeeEstimationMode}
 import org.bitcoins.rpc.jsonmodels._
 import org.bitcoins.rpc.serializers.JsonSerializers._
-import org.bitcoins.rpc.serializers.JsonWriters.CurrencyUnitWrites
 import play.api.libs.json._
 
 import scala.concurrent.Future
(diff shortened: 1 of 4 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/TransactionRpc.scala
```diff
@@ -148,7 +149,7 @@ trait TransactionRpc { self: Client =>
 
   def sendToAddress(
       address: BitcoinAddress,
-      amount: Bitcoins,
+      amount: CurrencyUnit,
```
Christewart, line 153: You have to be careful here, even though you are taking in a generic `CurrencyUnit` type, you need to convert it to `Bitcoins` _internally_ in this method as the bitcoind rpc server expects `Bitcoins`. For instance, if you passed in `Satoshis(1)` here and called `Satoshis(1).toBigDecimal` you would receive `1`, which would then bitcoind would interpret as one bitcoin!

```scala
@ val sat = Satoshis.one
sat: Satoshis = SatoshisImpl(Int64Impl(1))

@ sat.toBigDecimal
res5: BigDecimal = 1
```

You would effectively be sending one bitcoin when you meant to send one satoshi! Yikes!

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala
```diff
@@ -89,6 +89,11 @@ object JsonWriters {
     override def writes(addr: AddressType): JsValue = JsString(addr.toString)
   }
 
+  implicit object CurrencyUnitWrites extends Writes[CurrencyUnit] {
+    override def writes(amount: CurrencyUnit): JsValue =
+      JsNumber(amount.toBigDecimal)
```
Christewart: I don't think this is the default behavior we want, and this could be dangerous.

Christewart (changes requested):

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonWriters.scala
torkelrogstad: Chiming in here: I think we shouldn't have a `Writes[CurrencyUnit]` implemented at all, in the same way that we shouldn't have a `Writes[FiniteDuration]`. Both of these types are unit agnostic (could be satoshis/bitcoins or milliseconds/hours), so we should explicitly convert them to the right unit at the call site;

Issue #212, opened 2018-10-03
Add secp256k1jni bindings for the schnorrsig module in libsecp256k1


Christewart:
It looks like schnorr signature support is starting to be added to bitcoin-core/secp256k1. 

We need to modify our java interface to support schnorr signatures: 

https://github.com/bitcoin-s/bitcoin-s-core/tree/master/secp256k1jni

https://github.com/bitcoin-core/secp256k1/pull/558

Christewart:
There is now jni bindings courtesey of @floreslorca and the bitcoin-abc team https://github.com/Bitcoin-ABC/bitcoin-abc/tree/master/src/secp256k1/src

nkohen:
This binding is currently maintained on the `schnorr-dlc` branch on this commit: https://github.com/bitcoin-s/bitcoin-s/commit/4548dc045bbc2e931812b084402cabde6d684dcf

PR #627, opened 2019-07-16
WIP: Get schnorr signature unit tests passing

Christewart:
This begins to address #212

This PR contains binaries for linux 64 bit schnorr as implemented in bitcoin cash.

https://github.com/Bitcoin-ABC/bitcoin-abc/tree/master/src/secp256k1

The commit this is built off of in bitcoin cash is 45c0f7bdb8702b4b6d7f38b6b10db7030b065421

Our submodule for secp256k1 is based on the bitcoin-s fork for secp256k1. You can find the source code from which the binaries can be built here: 

https://github.com/bitcoin-s/secp256k1/tree/schnorr

TODO:

- [ ] Windows binaries
- [ ] mac osx binaries


nkohen (approved):
Looks good, but we still need to have a way of doing signatures where we specify the nonce at some point

Christewart:
So I think this will more than likely be the "official API", so we should probably code against this with the understanding it is subject to change. It will work for experiments https://github.com/bitcoin-core/secp256k1/pull/558

Christewart:
Superseded by #657

PR #633, opened 2019-07-16
WIP fixing compiler warnings during testing

cwaldron97:
I am currently fixing compiler warnings using `sbt` `test:compile` also needing to be noted in the `build.sbt` file I have changed the `lazy val testCompilerOpts` to be equal to `compilerOpts` instead of `commonCompilerOpts` this is just a note to change back before merging.


torkelrogstad (changes requested):
Couple of issues here: 
1) It'd be great to not touch non-Scala files. There's huge diffs here for `.json`, `.js`, `.xml`, `.java` and `.conf` files here. IMO autoformatting of code should happen with the same tool for all developers, and that requires us to have a configuration for it. Right now we only have configuration for a Scala formatter (`.scalafmt.conf`).
2) What was the process behind organizing imports? I assume you didn't do it manually. If it was through Scalafmt I think its :+1:, otherwise I'm not convinced. Using the `Optimize Imports` functionality of IntelliiJ is suboptimal as a formatting tool, as that functionality isn't configured by code, and prone to change from developer to developer based on their setup
3) Judging from the merge conflicts on this PR it looks like it needs to be rebased (that means your changes didn't happen on the most recent history, and needs to be brought up to date)

Issue #622, opened 2019-07-15
Clean up publish logs


Christewart:
When running `sbt publishLocal` you will see a bunch of warnings related to incorrect formatting of scaladocs. 

```
[warn] /**
[warn] ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/p2p/NetworkMessage.scala:39:3: Could not find any member to link for "NetworkParameters".
[warn]   /**
[warn]   ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/p2p/NetworkHeader.scala:42:3: Could not find any member to link for "NetworkParameters".
[warn]   /**
[warn]   ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala:167:3: Could not find any member to link for "org.bitcoins.core.hd.HDPath".
[warn]   /**
[warn]   ^
```

We need to clean up those warnings.

Christewart:
Here is a example of a CI run: https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/558513052#L2527

PR #634, opened 2019-07-16
Cleaning up Publish Logs

cwaldron97:
I tried reformatting some of the scaladoc comment blocks we have set up but it seems there are two issues here. First is that scaladocs does not recognize links that are within the java language and next is that the @throws tag does not work the way its intended. This issue with links as highlighted in issue #622 seems to be rather common and issues just get closed after being told to edit the `build.sbt` file to include `"-no-link-warnings"` in your `scaladoc` settings. I am a bit worried though as some of this is open source and has multiple people contributing the link warnings might be useful. 

I have double checked the `scaladocs` formatting that is required and in instances we are following them but it still throws a warning. There are a few things that could create an issue, 1. I don't think `scaladocs` handles `@throws` the same way that `javadocs` does so it struggles. 2. `scaladocs` doesn't seem to be very good at giving descriptive warnings so there might be an issue deeper in the chain of the doc but it will report them at the top. I have also noticed that it processes the `scaladocs` from the bottom up in a file. Many people seem to have similar issues with scaladocs and they just close the issues and recommend turning off the flags. [[https://github.com/scala/bug/issues/10182]]  [[https://github.com/akka/akka/issues/19998]]  [[https://github.com/scala/bug/issues/9311]] 

I did some more sanity checks and I have no clue why `@inheritdocs` does not work in the code base,
its isolated to `TransactionProcessing` the others instances have been fixed but for whatever reason it is not finding the `scaladocs` written in the `superclass` that the `defs` are overriding. The `superclass` is the `walletApi.scala` file if anyone wants to take a look and I'm missing anything obvious. There has been some removal of warnings so I think this PR is still worthy of being merged and I also fixed some documentation so there is content but it has not solved the problem in its entirety.

So as this pull request stands there has been added scaladocs in locations that lacked them or some scaladocs were expanded on. Some had directed links to their directory. I have removed the no link warning flag so this PR is just added documentation and fixing a few of the publish log errors we ran into but I can't seem to fix a persistent few. Hopefully in a update of scaladocs a lot of these will go away.

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
@@ -2,16 +2,14 @@ package org.bitcoins.rpc.config
 
 import java.io.File
 import java.net.URI
-import java.nio.file.Paths
+import java.nio.file.{Files, Paths}
 
+import org.bitcoins.core.config.NetworkParameters
+import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.rpc.client.common.BitcoindVersion
 
 import scala.sys.process._
-import org.bitcoins.core.util.BitcoinSLogger
-import org.bitcoins.core.config.NetworkParameters
-
-import scala.util.{Properties, Try}
-import java.nio.file.Files
+import scala.util.Properties
 
 /**
   * Created by chris on 4/29/17.
@@ -123,7 +121,7 @@ object BitcoindInstance {
     * Construct a `bitcoind` from the given config file. If no `datadir` setting
     * is found, the parent directory to the given file is used.
     *
-    * @throws IllegalArgumentException if the given config file does not exist
+    * @throws  IllegalArgumentException if the given config file does not exist
     */
   def fromConfigFile(
       file: File = BitcoindConfig.DEFAULT_CONF_FILE): BitcoindInstance = {
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,8 @@ lazy val commonSettings = List(
     "-doc-title",
     "Bitcoin-S",
     "-doc-version",
-    version.value
+    version.value,
+    "-no-link-warnings"
   ),
   // Set apiURL to define the base URL for the Scaladocs for our library.
   // This will enable clients of our library to automatically link against
--- a/build.sbt
+++ b/build.sbt
@@ -61,8 +61,7 @@ lazy val commonSettings = List(
     "-doc-title",
     "Bitcoin-S",
     "-doc-version",
-    version.value,
-    "-no-link-warnings"
+    version.value
   ),
   // Set apiURL to define the base URL for the Scaladocs for our library.
   // This will enable clients of our library to automatically link against
--- a/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
+++ b/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
@@ -1,12 +1,22 @@
 package org.bitcoins.eclair.rpc.json
 
-import org.bitcoins.core.crypto.{DoubleSha256Digest, DoubleSha256DigestBE, ECDigitalSignature, Sha256Digest}
+import org.bitcoins.core.crypto.{
+  DoubleSha256Digest,
+  DoubleSha256DigestBE,
+  ECDigitalSignature,
+  Sha256Digest
+}
 import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.protocol.ln.channel.{ChannelState, FundedChannelId}
 import org.bitcoins.core.protocol.ln.currency.MilliSatoshis
 import org.bitcoins.core.protocol.ln.fee.FeeProportionalMillionths
 import org.bitcoins.core.protocol.ln.node.NodeId
-import org.bitcoins.core.protocol.ln.{LnHumanReadablePart, LnInvoiceSignature, PaymentPreimage, ShortChannelId}
+import org.bitcoins.core.protocol.ln.{
+  LnHumanReadablePart,
+  LnInvoiceSignature,
+  PaymentPreimage,
+  ShortChannelId
+}
 import org.bitcoins.eclair.rpc.network.PeerState
 import play.api.libs.json.JsObject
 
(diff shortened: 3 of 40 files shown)


testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -620,7 +620,7 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
   }
 
   /**
-    * Returns a triple of [[org.bitcoins.rpc.client.v17.BitcoindV17RpcClient BitcoindV17RpcClient]]
+    * @Returns a triple of [[org.bitcoins.rpc.client.v17.BitcoindV17RpcClient BitcoindV17RpcClient]]
```
torkelrogstad:
```suggestion
    * @return a triple of [[org.bitcoins.rpc.client.v17.BitcoindV17RpcClient BitcoindV17RpcClient]]
```

build.sbt
```diff
@@ -59,11 +59,12 @@ lazy val commonSettings = List(
         "Bitcoin-S",
         "-doc-version",
         version.value,
+        "-no-link-warnings",
```
torkelrogstad: This shouldn't go in, IMO

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -102,7 +100,7 @@ object BitcoindInstance {
   /** Constructs a `bitcoind` instance from the given datadir, using the
     * `bitcoin.conf` found within (if any)
     *
-    * @throws IllegalArgumentException if the given datadir does not exist
+    * @throws [[java.lang.IllegalArgumentException]] if the given datadir does not exist
```
torkelrogstad:
```suggestion
    * @throws java.lang.IllegalArgumentException if the given datadir does not exist
```

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -123,7 +121,7 @@ object BitcoindInstance {
     * Construct a `bitcoind` from the given config file. If no `datadir` setting
     * is found, the parent directory to the given file is used.
     *
-    * @throws IllegalArgumentException if the given config file does not exist
+    * @throws [[java.lang.IllegalArgumentException]] if the given config file does not exist
```
torkelrogstad:
```suggestion
    * @throws java.lang.IllegalArgumentException if the given config file does not exist
```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -45,7 +55,7 @@ sealed abstract class ChannelInfo {
 
 /**
   * This represents the case where the
-  * [[org.bitcoins.core.protocol.ln.channel.ChannelState ChannelState]] is
+  * [[org.bitcoins.core.protocol.ln.channel.ChannelState] ChannelState] is
```
torkelrogstad:
```suggestion
  * [[org.bitcoins.core.protocol.ln.channel.ChannelState ChannelState]] is
```

testkit/src/main/scala/org/bitcoins/testkit/node/NodeTestUtil.scala
```diff
@@ -96,8 +96,8 @@ abstract class NodeTestUtil extends BitcoinSLogger {
     new InetSocketAddress(instance.uri.getHost, instance.p2pPort)
   }
 
-  /** Gets the [[Peer]] that
-    * corresponds to [[org.bitcoins.rpc.client.common.BitcoindRpcClient]] */
+  /** Gets the [[org.bitcoins.node.models.Peer]] that
+    * corresponds to org.bitcoins.rpc.client.common.BitcoindRpcClient */
```
torkelrogstad: Shouldn't remove the brackets (`[[]]`)

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -609,7 +609,7 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
   }
 
   /**
-    * Returns a triple of [[org.bitcoins.rpc.client.common.BitcoindRpcClient BitcoindRpcClient]]
+    * Returns a triple of org.bitcoins.rpc.client.common.BitcoindRpcClient BitcoindRpcClient
```
torkelrogstad, line 612:
```suggestion
    * @return a triple of [[org.bitcoins.rpc.client.common.BitcoindRpcClient BitcoindRpcClient]]
```

torkelrogstad (changes requested):
General comment in this file: 
`[[java.lang.Foo Foo]]` appears as `Foo` and makes it clickable to the Javadoc for Foo. `[[java.lang.Foo]]` appears as `java.lang.Foo` and makes it clickable. IMO the first option is better, because the link is shorter.

torkelrogstad:
Needs a rebase. 

To save you from rebase issues, make sure you pull the newest code before starting to make changes:-)

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -51,9 +51,9 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
   def emptyScriptSignature = p2pkhScriptSignature.map(_ => EmptyScriptSignature)
 
   /**
-    * Generates a [[org.bitcoins.core.protocol.script.P2SHScriptSignature P2SHScriptSignature]]
```
torkelrogstad, line 54: I'm only commenting here, but this applies to a bunch of places: The last word shouldn't be removed here, that screws up the link

torkelrogstad (changes requested):
To see how Scaladocs turn out: do `sbt unidoc`, and then look at the end of the printed lines for where the website is put

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala
```diff
@@ -166,8 +166,7 @@ sealed trait SpendingInfoDb
   * it is spent, how many (if any) confirmations it has
   * or which block/transaction it was included in.
   *
-  * That is rather handled by
-  * [[org.bitcoins.wallet.models.WalletTXOTable WalletTXOTable]].
+  * That is rather handled by WalletTXOTable.
```
torkelrogstad: This is my fault, I forgot to update the Scaladoc a while ago. Could you change the doc for this class to

```
This table stores the necessary information to spend
a transaction output (TXO) at a later point in time. It 
also stores how many confirmations it has, whether 
or not it is spent (i.e. if it is a UTXO or not) and the 
TXID of the transaction that created this output. 
```

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -109,6 +111,7 @@ trait LockedWalletApi extends WalletApi {
   /**
     * Mimics the `getaddressinfo` RPC call in Bitcoin Core
     *
+    * @param address
```
torkelrogstad, line 114: No point in having just a `@param` annotation if there's no description of the parameter

torkelrogstad:
Needs rebase

build.sbt
```diff
@@ -61,7 +61,8 @@ lazy val commonSettings = List(
     "-doc-title",
     "Bitcoin-S",
     "-doc-version",
-    version.value
+    version.value,
+    "-no-link-warnings"
```
torkelrogstad: Shouldn't add this IMO

cwaldron97: I have removed it for this PR just so we have some of the added scaladocs where they didn't already exist. I'll leave a note about this in my comment at the top but I agree for now we leave this out of this PR and settle it later.

Christewart:
Can we get this in @torkelrogstad ? It is strictly and improvement over what we have now right?

PR #637, opened 2019-07-16
Remove BitcoinJ from dependencies

### LINKED ISSUE #102: Remove bitcoinj dependency
Early on in this libraries development we would use bitcoinj to test against. I think this library has reached sufficient maturity to remove the bitcoinj test dependency and all corresponding bitcoinj test cases. 

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/project/Build.scala#L22

cwaldron97:
closes #102 removing test cases and dependencies to old testing system `bitcoinj` as it is no longer used.


project/Deps.scala
```diff
@@ -105,7 +105,6 @@ object Deps {
   )
 
   val coreTest = List(
-    Test.bitcoinj,
```
Christewart, line 108: The other thing you should delete in this file is the `val bitcoinj = ...` in the `Test` object

Christewart:
more compile failures here: https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/559638818#L866

PR #640, opened 2019-07-17
Add configurable logging to data directory

### LINKED ISSUE #588: Change syntax for creating app configs
https://github.com/bitcoin-s/bitcoin-s/blob/8fab4e35984b0eaa022a982ec23e5521d7c2c1bf/app/server/src/main/scala/org/bitcoins/server/Main.scala#L36

I don't really like this syntax, it would be nicer if we could have something like `fromDefaultConf()` or `fromDefaultDatadir()` rather than just an empty param that looks like magic is happening when you read code

_Originally posted by @Christewart in https://github.com/bitcoin-s/bitcoin-s/pull/558_

Would need a change in [`BitcoinSAppConfig`](https://github.com/bitcoin-s/bitcoin-s/blob/8fab4e35984b0eaa022a982ec23e5521d7c2c1bf/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala). Could make the constructor private, remove the `apply` method in the companion object and add a method `fromDefaultConf()`.

torkelrogstad:
Closes #588 

In this PR we add the ability for the node, chain
and wallet projects (+ the server) to log to the users
data directory instead of whatever directory the
binaries was launched from. This is inherently a bit
more complicated than our previous setup, because
we need to read the user directory before we can create
loggers. As a result of this, some files/methods were
moved around, so the relevant app config could be
found in scope.

## Discussion item: 
Right now we log in a similar way to how we already do it, by passing in the name of the file (instead of the class, but leads to almost identical behavior), and then constructing a logger with that name. This is why `sourcecode` (from Li Haoyi) is added as a dependency. Should we instead move to a category based system, where we log with a predefined set of categories? This is how both Bitcoin Core and Lnd handles it. Categories could be things such as chainsync, P2P messages, transaction creation, address derivation, peer handling, etc. IMO this is a better solution, primarily because when we have a predefined number of loggers its easy to expose options in `reference.conf` for fine-tuning behavior of each logger. I.e. the user could set P2P messages to be logged at `DEBUG`, but turn off chain sync logs. 

TODO: 
- [x] Make console/file logging something the user can turn on/off
- [x] Make specific loggers tunable wrt. levels
- [x] Add doc on logging for dev purposes

UPDATE: 

Added several logging categories, as well as options for tuning logging levels and file and console logs on/off.


node/src/main/scala/org/bitcoins/node/peer/ConnectedPeerManager.scala
```diff
@@ -1,10 +1,12 @@
...
 
-import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.node.models.Peer
+import org.bitcoins.db.P2PLogger
+import org.bitcoins.chain.config.ChainAppConfig
 
 /** Manages all peers we have active connections with */
-class ConnectedPeerManager extends BitcoinSLogger {
+class ConnectedPeerManager extends P2PLogger {
+  implicit val config: ChainAppConfig = ???
```
nkohen, line 9: ping

torkelrogstad: This file isn't actually used anywhere

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/ChainSync.scala
```diff
@@ -5,11 +5,12 @@ import org.bitcoins.chain.blockchain.ChainHandler
...
 import org.bitcoins.core.crypto.DoubleSha256DigestBE
 import org.bitcoins.core.protocol.blockchain.BlockHeader
-import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.db.AppLoggers
 
 import scala.concurrent.{ExecutionContext, Future}
+import org.bitcoins.chain.config.ChainAppConfig
 
-trait ChainSync extends BitcoinSLogger {
+trait ChainSync {
```
Christewart: `extends ChainVerificationLogger`?

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -62,7 +61,9 @@ object Blockchain extends BitcoinSLogger {
     *         or [[BlockchainUpdate.Failed failed]] to connect to a tip
     */
   def connectTip(header: BlockHeader, blockHeaderDAO: BlockHeaderDAO)(
-      implicit ec: ExecutionContext): Future[BlockchainUpdate] = {
+      implicit ec: ExecutionContext,
+      conf: ChainAppConfig): Future[BlockchainUpdate] = {
+    val logger = AppLoggers.getVerificationLogger
```
Christewart: so why don't we extend the logger?

db-commons/src/main/resources/reference.conf
```diff
@@ -2,6 +2,36 @@ bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
 
+    logging {
+        level = info # trace, debug, info, warn, error, off
+
+        # You can also tune specific module loggers.
+        # They each take the same levels as above.
+        # The available loggers are:
```
Christewart, line 13: can you give an example? For instance if i wanted to tune p2p would i do 

```
logging.p2p = INFO
```
?

db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
```diff
@@ -0,0 +1,214 @@
...
+  def getDatabaseLogger(implicit conf: AppConfig): Logger =
+    getLoggerImpl(LoggerKind.Database)
+
+  private val context = {
+    val context = LoggerFactory.getILoggerFactory() match {
+      case ctx: LoggerContext => ctx
+      case other              => sys.error(s"Expected LoggerContext, got: $other")
+    }
+
+    // following three lines prevents Logback from reading XML conf files
```
Christewart, line 71: So does this mean our dev flow has to go through these loggers to? Or is this only applicable when the app is running?

db-commons/src/main/resources/reference.conf
```diff
@@ -2,6 +2,36 @@ bitcoin-s {
     datadir = ${HOME}/.bitcoin-s
     network = regtest # regtest, testnet3, mainnet
 
+    logging {
```
Christewart, line 5: is there any way we can log everything to one file if we want to?

torkelrogstad: WYM? Right now all loggers log to the same file

db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
torkelrogstad: Both tests and running the app will log to the same place. By default that's both the console and the log file, so this doesn't really change development flow

db-commons/src/main/resources/reference.conf
Christewart: oh -- i thought they were being logged to different files in this PR. In the future they will be right? In that case, can we somehow having a setting to dump everything into one file?

torkelrogstad: It's trivial to add new log files for specific loggers, as well as get one logger to output to several files if we want that in the future.

Christewart:
This fails to compile on 2.11

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/565913446#L621

Christewart:
Tests do not pass for me locally either 
```
[info] ChainAppConfigTest:
[info] - must be overridable *** FAILED ***
[info]   RegTest did not equal TestNet3 (ChainAppConfigTest.scala:21)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:530)
[info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:529)
[info]   at org.scalatest.FlatSpec.newAssertionFailedException(FlatSpec.scala:1685)
[info]   at org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:503)
[info]   at org.bitcoins.chain.ChainAppConfigTest.$anonfun$new$1(ChainAppConfigTest.scala:21)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```

torkelrogstad:
The error you're having is not reproducing locally for me, nor on CI. Are you still seeing them, @Christewart ?

torkelrogstad:
I pushed a new commit that changes our app configs to explictly take in the data directory as a parameter, instead of magically resolving it. This also closes #588

PR #647, opened 2019-07-18
Fix merkle block parsing error

torkelrogstad:
This took an absurd amount of head-bashing to figure out, but here's a fix for a bug that's taken me several hours to struggle with.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/serializers/blockchain/RawMerkleBlockSerializer.scala
+++ b/core/src/main/scala/org/bitcoins/core/serializers/blockchain/RawMerkleBlockSerializer.scala
@@ -18,27 +18,31 @@ sealed abstract class RawMerkleBlockSerializer
     extends RawBitcoinSerializer[MerkleBlock] {
 
   def read(bytes: ByteVector): MerkleBlock = {
-    val blockHeader = RawBlockHeaderSerializer.read(bytes.take(80))
-    val bytesAfterBlockHeaderParsing =
-      bytes.slice(blockHeader.bytes.size, bytes.size)
-    val transactionCount = UInt32(
-      bytesAfterBlockHeaderParsing.slice(0, 4).reverse)
-    val hashCount = CompactSizeUInt.parseCompactSizeUInt(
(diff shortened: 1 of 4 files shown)


node/src/main/scala/org/bitcoins/node/util/BitcoinSpvNodeUtil.scala
```diff
@@ -38,9 +38,9 @@ trait BitcoinSpvNodeUtil extends BitcoinSLogger {
               loop(newRemainingBytes, message :: accum)
             }
           case Failure(exception) =>
-            logger.debug(
-              "Failed to parse network message, could be because tcp frame isn't aligned")
-            logger.debug(exception.getMessage)
+            logger.error(
```
torkelrogstad, line 41: Figured it was best to up the log level, as the fact that this was on debug was the reason I didn't see it while working on the SPV node

core/src/main/scala/org/bitcoins/core/serializers/blockchain/RawMerkleBlockSerializer.scala
```diff
@@ -18,27 +18,32 @@ sealed abstract class RawMerkleBlockSerializer
...
+    val blockHeader = RawBlockHeaderSerializer.read(headerBytes)
 
+    val (txCountBytes, afterTxCount) = afterHeader.splitAt(4)
+
+    val transactionCount = UInt32.fromBytes(txCountBytes.reverse)
+    val hashCount = CompactSizeUInt.parseCompactSizeUInt(afterTxCount)
+
+    val (_, afterHashCountBytes) = afterTxCount.splitAt(hashCount.bytes.length)
+
+    println()
```
Christewart: nit: remove

node/src/main/scala/org/bitcoins/node/util/BitcoinSpvNodeUtil.scala
Christewart: it's fairly common that TCP frames aren't aligned, so this may be noisy

Christewart:
So i'm ok with merginng this as unit tests do pass, but I can't seem to spot the bug. Obviously the code is pretty hairy so i'm not surprised there is something wrong with it.

PR #649, opened 2019-07-19
implement ability to monitor a invoice that we are generated

Christewart:
This implements the ability to monitor an invoice after it is created with the eclair rpc client. When a payment is received it is published to the actor systems event bus, and the Future returned by `monitorInvoice` is completed after 

1. a payment being received
2. the timeout being hit, which is dictated by the parameter `maxAttempts`

### DIFF
--- a/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
+++ b/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
@@ -15,9 +15,14 @@ import org.bitcoins.core.protocol.Address
 import org.bitcoins.core.protocol.ln.channel.{ChannelId, FundedChannelId}
 import org.bitcoins.core.protocol.ln.currency.MilliSatoshis
 import org.bitcoins.core.protocol.ln.node.NodeId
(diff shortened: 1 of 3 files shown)


eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -385,15 +434,21 @@ class EclairRpcClient(val instance: EclairInstance)(
...
-      paymentHash: Sha256Digest): Future[ReceivedPaymentResult] = {
-    eclairCall[ReceivedPaymentResult]("getreceivedinfo",
-                                      "paymentHash" -> paymentHash.hex)
-  }
-
-  override def getReceivedInfo(
-      invoice: LnInvoice): Future[ReceivedPaymentResult] = {
-    eclairCall[ReceivedPaymentResult]("getreceivedinfo",
-                                      "invoice" -> invoice.toString)
+      paymentHash: Sha256Digest): Future[Option[ReceivedPaymentResult]] = {
```
nkohen, line 437: nit: I think it would be cleaner to have `ReceivedPaymentResult` be a `sealed trait` which was extended by the two possible cases

Christewart: i agree, but with it being time pressing i think we should just use `Option` for this for now.

PR #651, opened 2019-07-20
Add the -Xfatal-warnings compiler option to make sure we get compile …

Christewart:
…errors if we have warnings

https://docs.scala-lang.org/overviews/compiler-options/index.html

```
 -Xfatal-warnings

    Fail the compilation if there are any warnings.
```

there are errors currently, i'm not sure what this one really is 

```
[error] sbt-api: Unhandled type class scala.reflect.internal.Types$MethodType : ($this: org.bitcoins.core.crypto.AesPassword)String
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 4 s, completed Jul 20, 2019 1:01:43 PM

```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,6 +28,7 @@ lazy val compilerOpts = Seq(
   "-feature",
   "-deprecation",
   "-Xfuture",
+  "-Xfatal-warnings",
   "-Ywarn-dead-code",
   "-Ywarn-unused-import",
   "-Ywarn-value-discard",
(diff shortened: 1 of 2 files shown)


torkelrogstad:
I think this is a bug in sbt/Zinc: 
- https://github.com/sbt/sbt/issues/3138
- https://stackoverflow.com/questions/29720289/warning-about-unhandled-type-class-during-compilation

Also, I think this is a compiler option we should only enforce on CI, not while working locally. Makes working with WIP code a massive pain in the ass

Christewart:
I said this on a meeting with you, but wanted to comment publicly as well, I think you should comment out that flag when doing heavy WIP stuff. I don't think the best solution is just enforcing on CI.

Christewart:
@torkelrogstad There is no workaround for this currently right?

torkelrogstad:
@Christewart the maintainer of sbt recently made a PR that ignores this warning: https://github.com/sbt/zinc/pull/689. It is not merged yet though, so we'll probably have to wait a little while to get this fixed.

Christewart:
This was done in #1483

PR #652, opened 2019-07-21
Implement missing branch on POW validation for testnet, implement Blo…

Christewart:
…ckHeaderDAO.find

This PR implements a missing branch in our POW validation logic that is found [here](https://github.com/bitcoin/bitcoin/blob/35477e9e4e3f0f207ac6fa5764886b15bf9af8d0/src/pow.cpp#L29-L35). I believe this code is needed to properly sync our spv node with testnet.

This PR also introduces the `find` method to `BlockHeaderDAO`. This is similar to the `find()` method in the scala collections library that allows you to pass in a `f: BlockHeaderDb => Boolean` and returns a `Future[Option[BlockHeaderDb]]`. I think this will be useful later on down the line, and I would be interested in trying to add something like this to `CRUD`

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -195,7 +195,7 @@ case class BlockHeaderDAO()(
     }
   }
 
-  /** Finds a [[BlockHeaderDb block header]] that satisfies the given predicate, else returns None */
+  /** Finds a [[org.bitcoins.chain.models.BlockHeaderDb block header]] that satisfies the given predicate, else returns None */
   def find(f: BlockHeaderDb => Boolean)(
       implicit ec: ExecutionContext): Future[Option[BlockHeaderDb]] = {
     val chainsF = getBlockchains()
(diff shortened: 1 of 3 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -194,4 +194,16 @@ case class BlockHeaderDAO()(
       Future.sequence(nestedFuture)
     }
   }
+
+  /** Finds a [[BlockHeaderDb block header]] that satisfies the given predicate, else returns None */
```
torkelrogstad: I don't think the link here will work in the Scaladoc, you'd need to do `[[org.bitcoins.chain.models.BlockHeaderDb block header]]`

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -194,4 +194,16 @@ case class BlockHeaderDAO()(
...
+  def find(f: BlockHeaderDb => Boolean)(
+      implicit ec: ExecutionContext): Future[Option[BlockHeaderDb]] = {
+    val chainsF = getBlockchains()
+    chainsF.map { chains =>
+      val headersOpt: Vector[Option[BlockHeaderDb]] =
+        chains.map(_.headers.find(f))
+      //if there are multiple, we just choose the first one for now
```
torkelrogstad, line 205: Adding a `logger.warn` in the case of finding multiple would be a good idea IMO

torkelrogstad (approved):
Save for the two nits, LGTM

Issue #653, opened 2019-07-21
Benchmark TipValidation


Christewart:
It looks like, from logs, it takes roughly 100ms to connect a tip in bitcoin-s. 

```
2019-07-21T18:28:33,558Z DEBUG Connected new header to blockchain, height=10077 hash=DoubleSha256BDigestBEImpl(00000000307cf3745615e1926eefe4abe9b95c97c3c2fff5d4aa9be0f38f7762)
2019-07-21T18:28:33,598Z DEBUG Connected new header to blockchain, height=10078 hash=DoubleSha256BDigestBEImpl(00000000130d0a10ac76fe510ad42bf6c88b58b9a190f1d34f8e3ecc3b07abaf)
2019-07-21T18:28:33,666Z DEBUG Connected new header to blockchain, height=10079 hash=DoubleSha256BDigestBEImpl(000000003a7002e1247b0008cba36cd46f57cd7ce56ac9d9dc5644265064df09)
2019-07-21T18:28:33,749Z DEBUG Connected new header to blockchain, height=10080 hash=DoubleSha256BDigestBEImpl(000000007c4fc01a14af8067762fb807144b1b59cd4ec79ffc61efae3439757d)
```

This means for testnet it will take roughly 41.6666667 hours to sync a node, which seems very slow. We should get this down at least an order of magnititude

torkelrogstad:
Would prob. make sense to do some more work on this: https://github.com/bitcoin-s/bitcoin-s/pull/574

Christewart:
Latest:

These results should be reproducible with this branch: https://github.com/Christewart/bitcoin-s-core/tree/2019-07-31-recv-funds-rebase-655

```
sbt bloopInstall && bloop run walletServer
```

It incorporates the optimizations done in #655 

```
14:24:51.479 DEBUG ChainHandler - Connected new header to blockchain, height=111910 hash=DoubleSha256BDigestBEImpl(00000000000867d3817bca0d947aabe17b38062d5f6977a33cd62696e021194c)
14:24:51.496 DEBUG ChainHandler - Connected new header to blockchain, height=111911 hash=DoubleSha256BDigestBEImpl(00000000000aa00995691949b1df7cd76e98154ccf10292a38c971cc2ce221f2)
[...]

PR #654, opened 2019-07-22
POC: UInt32 optimization

Christewart:
In wake of #653 I think some of the lowest hanging fruit that will improve the library overall is optimize our Number implementations.

To do this, I think we will have to get rid of the `Number` trait. I can't figure how to to get the underlying implementation -- defined as `A` in `Number` -- to be different for each concrete type. I.e. a `UInt8` should _not_ have a `BigInt` backing it, it should be a `Short` (maybe a `Byte` if we are ambitious!) The underlying implementation of `UInt64` _should_ be a `BigInteger`, and a `UInt32` should be a `Long` etc.

This PR optimizes our `UInt32` type. It turns `UInt32` into a case class and we extend [`AnyVal`](https://docs.scala-lang.org/overviews/core/value-classes.html) to avoid runtime allocations. This should improve performance and memory usage.

I think this is going to come at the cost of having redundant code for each number implementation we do, but as far as i can tell this is what Scala does with their number types [`Int`](https://github.com/scala/scala/blob/2.13.x/src/library/scala/Int.scala)/[`Long`](https://github.com/scala/scala/blob/2.13.x/src/library/scala/Long.scala) etc

If the overall design is acceptable, I think we should extend this to the rest of Number implementations.

The only unit tests not passing in `coreTest` are the ones that are checking that `UInt32` is not a negative number, unfortunately we cannot preserve the invariant `require()` in the body of the case class while extending `AnyVal`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/BasicArithmetic.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/BasicArithmetic.scala
@@ -9,7 +9,7 @@ import scala.util.Try
   * operator to throw. This method wraps it in a `Try`
   * block.
   */
-trait BasicArithmetic[N] {
+trait BasicArithmetic[N] extends Any {
 
   def +(n: N): N
 
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -129,11 +129,75 @@ sealed abstract class UInt8 extends UnsignedNumber[UInt8] {
 /**
   * Represents a uint32_t in C
   */
-sealed abstract class UInt32 extends UnsignedNumber[UInt32] {
-  override def apply: A => UInt32 = UInt32(_)
-  override def hex: String = BitcoinSUtil.encodeHex(toLong).slice(8, 16)
+final case class UInt32(underlying: Long)
+    extends AnyVal
+    with NetworkElement
+    with BasicArithmetic[UInt32]
+    with Ordered[UInt32] {
+  override def hex: String = BitcoinSUtil.encodeHex(underlying).slice(8, 16)
 
-  override def andMask = 0xffffffffL
+  override def bytes: ByteVector = ByteVector.fromValidHex(hex)
+
+  override def +(y: UInt32): UInt32 = {
+    UInt32(underlying + y.underlying)
+  }
+
+  override def -(y: UInt32): UInt32 = {
+    UInt32(underlying - y.underlying)
+  }
+
+  override def *(y: UInt32): UInt32 = {
+    UInt32(underlying * y.underlying)
+  }
+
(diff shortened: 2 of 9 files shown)


core/src/main/scala/org/bitcoins/core/number/BasicArithmetic.scala
```diff
@@ -9,7 +9,7 @@ import scala.util.Try
   * operator to throw. This method wraps it in a `Try`
   * block.
   */
-trait BasicArithmetic[N] {
+trait BasicArithmetic[N] extends Any {
```
nkohen, line 12: Don't all things automatically extend `Any`?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -129,11 +129,75 @@ sealed abstract class UInt8 extends UnsignedNumber[UInt8] {
 /**
   * Represents a uint32_t in C
   */
-sealed abstract class UInt32 extends UnsignedNumber[UInt32] {
-  override def apply: A => UInt32 = UInt32(_)
-  override def hex: String = BitcoinSUtil.encodeHex(toLong).slice(8, 16)
+final case class UInt32(underlying: Long)
```
nkohen, line 132: I assume we went with `Long` because there's no nice way to use `Int` co-opting the sign bit somehow right?

nkohen (approved):

cwaldron97 (approved):

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
Christewart: Exactly. There might be a clever way to do this, but I"m not going to spend the time on how to figure it out atm. cc: @cwaldron97

core/src/main/scala/org/bitcoins/core/number/BasicArithmetic.scala
Christewart: That's what I thought too, but here is the error i get if I don't do it 

```
[error]  is not a subclass of the superclass Object
[error]  of the mixin trait BasicArithmetic
[error]     with BasicArithmetic[UInt32]
```

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -308,35 +372,35 @@ object UInt8
...
-  }
+  }*/
 
   lazy val zero = UInt32(0)
   lazy val one = UInt32(1)
 
-  private lazy val minUnderlying: A = 0
-  private lazy val maxUnderlying: A = 4294967295L
+  private lazy val minUnderlying = 0
+  private lazy val maxUnderlying = 4294967295L
```
rorp, line 386: I think `0xffffffffL` is more readable in this case.

Christewart:
This is superseded by #659

PR #655, opened 2019-07-22
Start the process of refactoring our ChainHandler to be able to avoid…

Christewart:
… database calls on TipValidation

This part of addressing #653 

A problem we have during TipValidation is that we don't cache our blockchain inside of `ChainHandler`. This means we have to call [`BlockHeaderDAO.getBlockchains()`](https://github.com/bitcoin-s/bitcoin-s/blob/master/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala#L64) every time we want to validate a new tip which is an expensive call.

The idea behind this PR is to cache our set of chains inside of `ChainHandler` to avoid this call. This comes at the risk of us having a stale `ChainHandler` instance lying around that does _not have the real best tip in it's `blockchain` val_. 

I'm particularly concerned about how to handle this inside of `DataMessageHandler`, as currently our [`PeerMessageReceiver` spawns a new one](https://github.com/bitcoin-s/bitcoin-s/blob/master/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala#L141) (does not cache it!) for every message received. This circumvents this optimization.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -44,6 +44,7 @@ case class Blockchain(headers: Vector[BlockHeaderDb])
 object Blockchain extends ChainVerificationLogger {
 
   def fromHeaders(headers: Vector[BlockHeaderDb]): Blockchain = {
+
     Blockchain(headers)
   }
 
@@ -60,59 +61,63 @@ object Blockchain extends ChainVerificationLogger {
     *         we [[org.bitcoins.chain.blockchain.BlockchainUpdate.Successful successful]] connected the tip,
     *         or [[org.bitcoins.chain.blockchain.BlockchainUpdate.Failed Failed]] to connect to a tip
     */
-  def connectTip(header: BlockHeader, blockHeaderDAO: BlockHeaderDAO)(
+  def connectTip(
+      header: BlockHeader,
+      blockHeaderDAO: BlockHeaderDAO,
+      blockchains: Vector[Blockchain])(
       implicit ec: ExecutionContext,
       conf: ChainAppConfig): Future[BlockchainUpdate] = {
     logger.debug(
       s"Attempting to add new tip=${header.hashBE.hex} with prevhash=${header.previousBlockHashBE.hex} to chain")
 
-    //get all competing chains we have
-    val blockchainsF: Future[Vector[Blockchain]] =
-      blockHeaderDAO.getBlockchains()
-
-    val tipResultF: Future[BlockchainUpdate] = blockchainsF.flatMap {
-      blockchains =>
-        val nested: Vector[Future[BlockchainUpdate]] = blockchains.map {
-          blockchain =>
-            val prevBlockHeaderOpt =
-              blockchain.find(_.hashBE == header.previousBlockHashBE)
-            prevBlockHeaderOpt match {
-              case None =>
-                logger.warn(
-                  s"No common ancestor found in the chain to connect header=${header.hashBE.hex}")
-                val err = TipUpdateResult.BadPreviousBlockHash(header)
-                val failed = BlockchainUpdate.Failed(blockchain = blockchain,
-                                                     failedHeader = header,
-                                                     tipUpdateFailure = err)
-                Future.successful(failed)
-
-              case Some(prevBlockHeader) =>
-                //found a header to connect to!
-                logger.trace(
-                  s"Found ancestor=${prevBlockHeader.hashBE.hex} for header=${header.hashBE.hex}")
-                val tipResultF =
-                  TipValidation.checkNewTip(newPotentialTip = header,
-                                            currentTip = prevBlockHeader,
-                                            blockHeaderDAO = blockHeaderDAO)
-
-                tipResultF.map { tipResult =>
-                  tipResult match {
-                    case TipUpdateResult.Success(headerDb) =>
-                      logger.debug(
-                        s"Successfully verified=${headerDb.hashBE.hex}, connecting to chain")
-                      val newChain =
-                        Blockchain.fromHeaders(headerDb +: blockchain.headers)
-                      BlockchainUpdate.Successful(newChain, headerDb)
-                    case fail: TipUpdateResult.Failure =>
-                      logger.warn(
-                        s"Could not verify header=${header.hashBE.hex}, reason=$fail")
-                      BlockchainUpdate.Failed(blockchain, header, fail)
-                  }
-                }
+    val tipResultF: Future[BlockchainUpdate] = {
+      val nested: Vector[Future[BlockchainUpdate]] = blockchains.map {
+        blockchain =>
+          val prevBlockHeaderIdxOpt =
+            blockchain.headers.zipWithIndex.find {
+              case (headerDb, _) =>
+                headerDb.hashBE == header.previousBlockHashBE
             }
-        }
-        parseSuccessOrFailure(nested = nested)
+          prevBlockHeaderIdxOpt match {
+            case None =>
+              logger.warn(
+                s"No common ancestor found in the chain to connect to ${header.hashBE}")
+              val err = TipUpdateResult.BadPreviousBlockHash(header)
+              val failed = BlockchainUpdate.Failed(blockchain = blockchain,
+                                                   failedHeader = header,
+                                                   tipUpdateFailure = err)
+              Future.successful(failed)
+
+            case Some((prevBlockHeader, prevHeaderIdx)) =>
+              //found a header to connect to!
+              logger.debug(
+                s"Attempting to add new tip=${header.hashBE.hex} with prevhash=${header.previousBlockHashBE.hex} to chain")
+              val tipResultF =
+                TipValidation.checkNewTip(newPotentialTip = header,
+                                          currentTip = prevBlockHeader,
+                                          blockHeaderDAO = blockHeaderDAO)
+
+              tipResultF.map { tipResult =>
+                tipResult match {
+                  case TipUpdateResult.Success(headerDb) =>
+                    logger.debug(
+                      s"Successfully verified=${headerDb.hashBE.hex}, connecting to chain")
+                    val oldChain =
+                      blockchain.takeRight(blockchain.length - prevHeaderIdx)
+                    val newChain =
+                      Blockchain.fromHeaders(headerDb +: oldChain)
+                    BlockchainUpdate.Successful(newChain, headerDb)
+                  case fail: TipUpdateResult.Failure =>
+                    logger.warn(
+                      s"Could not verify header=${header.hashBE.hex}, reason=$fail")
+                    BlockchainUpdate.Failed(blockchain, header, fail)
+                }
+              }
+          }
+      }
+      parseSuccessOrFailure(nested = nested)
     }
+
     tipResultF
   }
 
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -19,9 +19,11 @@ import org.bitcoins.chain.validation.TipUpdateResult
   * of [[org.bitcoins.chain.api.ChainApi ChainApi]], this is the entry point in to the
   * chain project.
   */
-case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
-    implicit private[chain] val chainConfig: ChainAppConfig
-) extends ChainApi
+case class ChainHandler(
+    blockHeaderDAO: BlockHeaderDAO,
+    blockchains: Vector[Blockchain])(
+    implicit private[chain] val chainConfig: ChainAppConfig)
+    extends ChainApi
     with ChainVerificationLogger {
 
   override def getBlockCount(implicit ec: ExecutionContext): Future[Long] = {
@@ -49,18 +51,40 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
     logger.debug(
       s"Processing header=${header.hashBE.hex}, previousHash=${header.previousBlockHashBE.hex}")
 
-    val blockchainUpdateF =
-      Blockchain.connectTip(header, blockHeaderDAO)
+    val blockchainUpdateF = Blockchain.connectTip(header = header,
+                                                  blockHeaderDAO =
+                                                    blockHeaderDAO,
+                                                  blockchains = blockchains)
 
     val newHandlerF = blockchainUpdateF.flatMap {
-      case BlockchainUpdate.Successful(_, updatedHeader) =>
+      case BlockchainUpdate.Successful(newChain, updatedHeader) =>
         //now we have successfully connected the header, we need to insert
         //it into the database
         val createdF = blockHeaderDAO.create(updatedHeader)
         createdF.map { header =>
           logger.debug(
-            s"Connected new header to blockchain, height=${header.height} hash=${header.hashBE.hex}")
-          ChainHandler(blockHeaderDAO)
+            s"Connected new header to blockchain, height=${header.height} hash=${header.hashBE}")
+          val chainIdxOpt = blockchains.zipWithIndex.find {
+            case (chain, _) =>
+              val oldTip = newChain(1) //should be safe, even with genesis header as we just connected a tip
+              oldTip == chain.tip
+          }
+
+          val updatedChains = {
+            chainIdxOpt match {
+              case Some((_, idx)) =>
+                logger.trace(
+                  s"Updating chain at idx=${idx} out of competing chains=${blockchains.length} with new tip=${header.hashBE.hex}")
+                blockchains.updated(idx, newChain)
+
+              case None =>
+                logger.info(
+                  s"New competing blockchain with tip=${newChain.tip}")
+                blockchains.:+(newChain)
+            }
+          }
+
+          ChainHandler(blockHeaderDAO, updatedChains)
         }
       case BlockchainUpdate.Failed(_, _, reason) =>
         val errMsg =
@@ -117,11 +141,42 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
     //this does _not_ mean that it is on the chain that has the most work
     //TODO: Enhance this in the future to return the "heaviest" header
     //https://bitcoin.org/en/glossary/block-chain
-    blockHeaderDAO.chainTips.map { tips =>
-      val sorted = tips.sortBy(header => header.blockHeader.difficulty)
-      val hash = sorted.head.hashBE
-      logger.debug(s"getBestBlockHash result: hash=$hash")
-      hash
+    val groupedChains = blockchains.groupBy(_.tip.height)
+    val maxHeight = groupedChains.keys.max
+    val chains = groupedChains(maxHeight)
+
+    val hashBE: DoubleSha256DigestBE = chains match {
+      case Vector() =>
+        val errMsg = s"Did not find blockchain with height $maxHeight"
+        logger.error(errMsg)
+        throw new RuntimeException(errMsg)
+      case chain +: Vector() =>
+        chain.tip.hashBE
+      case chain +: rest =>
+        logger.warn(
+          s"We have multiple competing blockchains: ${(chain +: rest).map(_.tip.hashBE.hex).mkString(", ")}")
+        chain.tip.hashBE
     }
+    Future.successful(hashBE)
+  }
+}
+
+object ChainHandler {
+
+  /** Constructs a [[ChainHandler chain handler]] from the state in the database
+    * This gives us the guaranteed latest state we have in the database
+    * */
+  def fromDatabase(blockHeaderDAO: BlockHeaderDAO)(
+      implicit ec: ExecutionContext,
+      chainConfig: ChainAppConfig): Future[ChainHandler] = {
+    val bestChainsF = blockHeaderDAO.getBlockchains()
+
+    bestChainsF.map(chains =>
+      new ChainHandler(blockHeaderDAO = blockHeaderDAO, blockchains = chains))
+  }
+
+  def apply(blockHeaderDAO: BlockHeaderDAO, blockchains: Blockchain)(
+      implicit chainConfig: ChainAppConfig): ChainHandler = {
+    new ChainHandler(blockHeaderDAO, Vector(blockchains))
   }
 }
--- a/node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
@@ -27,9 +27,10 @@ class SpvNodeTest extends NodeUnitTest {
       val spvNode = spvNodeConnectedWithBitcoind.spvNode
       val bitcoind = spvNodeConnectedWithBitcoind.bitcoind
 
-      assert(spvNode.isConnected)
-
-      assert(spvNode.isInitialized)
+      val assert1F = for {
+        _ <- spvNode.isConnected.map(assert(_))
+        a2 <- spvNode.isInitialized.map(assert(_))
+      } yield a2
 
       val hashF: Future[DoubleSha256DigestBE] = {
         bitcoind.generate(1).map(_.head)
@@ -37,6 +38,7 @@ class SpvNodeTest extends NodeUnitTest {
 
       //sync our spv node expecting to get that generated hash
       val spvSyncF = for {
+        _ <- assert1F
         _ <- hashF
         sync <- spvNode.sync()
       } yield sync
@@ -62,7 +64,9 @@ class SpvNodeTest extends NodeUnitTest {
       //as they happen with the 'sendheaders' message
       //both our spv node and our bitcoind node _should_ both be at the genesis block (regtest)
       //at this point so no actual syncing is happening
-      val initSyncF = gen1F.flatMap(_ => spvNode.sync())
+      val initSyncF = gen1F.flatMap { _ =>
+        spvNode.sync()
+      }
 
       //start generating a block every 10 seconds with bitcoind
       //this should result in 5 blocks
@@ -76,7 +80,8 @@ class SpvNodeTest extends NodeUnitTest {
         //we should expect 5 headers have been announced to us via
         //the send headers message.
         val has6BlocksF = RpcUtil.retryUntilSatisfiedF(
-          conditionF = () => spvNode.chainApi.getBlockCount.map(_ == 6),
+          conditionF =
+            () => spvNode.chainApiFromDb().flatMap(_.getBlockCount.map(_ == 6)),
           duration = 1.seconds)
 
         has6BlocksF.map(_ => succeed)
--- a/node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
@@ -2,6 +2,8 @@ package org.bitcoins.node.networking
 
 import akka.io.Tcp
 import akka.testkit.{TestActorRef, TestProbe}
+import org.bitcoins.chain.db.ChainDbManagement
+import org.bitcoins.node.SpvNodeCallbacks
 import org.bitcoins.node.models.Peer
 import org.bitcoins.node.networking.peer.PeerMessageReceiver
 import org.bitcoins.node.networking.peer.PeerMessageReceiverState.Preconnection
@@ -39,6 +41,7 @@ class P2PClientTest
     BitcoinSTestAppConfig.getTestConfig()
   implicit private val chainConf = config.chainConf
   implicit private val nodeConf = config.nodeConf
+  implicit private val timeout = akka.util.Timeout(10.seconds)
 
   implicit val np = config.chainConf.network
 
@@ -126,8 +129,20 @@ class P2PClientTest
   }
   behavior of "P2PClient"
 
+  override def beforeAll(): Unit = {
+    ChainDbManagement.createHeaderTable()
+  }
+
+  override def afterAll(): Unit = {
+    ChainDbManagement.dropHeaderTable()
+    super.afterAll()
+  }
+
   it must "establish a tcp connection with a bitcoin node" in {
-    bitcoindPeerF.flatMap(remote => connectAndDisconnect(remote))
+    bitcoindPeerF.flatMap { remote =>
+      println(s"Starting test")
+      connectAndDisconnect(remote)
+    }
   }
 
   it must "connect to two nodes" in {
@@ -152,26 +167,32 @@ class P2PClientTest
   def connectAndDisconnect(peer: Peer): Future[Assertion] = {
     val probe = TestProbe()
     val remote = peer.socket
-    val chainHandler = {
-      val dao = BlockHeaderDAO()
-      ChainHandler(dao)
+    val peerMessageReceiverF =
+      PeerMessageReceiver.preConnection(peer, SpvNodeCallbacks.empty)
+
+    val clientActorF: Future[TestActorRef[P2PClientActor]] =
+      peerMessageReceiverF.map { peerMsgRecv =>
+        TestActorRef(P2PClient.props(peer, peerMsgRecv), probe.ref)
+      }
+    val p2pClientF: Future[P2PClient] = clientActorF.map {
+      client: TestActorRef[P2PClientActor] =>
+        P2PClient(client, peer)
     }
-    val peerMessageReceiver =
-      PeerMessageReceiver(state = Preconnection, chainHandler)
-    val client =
-      TestActorRef(P2PClient.props(peer, peerMessageReceiver), probe.ref)
 
-    client ! Tcp.Connect(remote)
-
-    val isConnectedF =
-      TestAsyncUtil.retryUntilSatisfied(peerMessageReceiver.isInitialized)
+    val isConnectedF = for {
+      p2pClient <- p2pClientF
+      _ = p2pClient.actor ! Tcp.Connect(remote)
+      isConnected <- TestAsyncUtil.retryUntilSatisfiedF(p2pClient.isConnected)
+    } yield isConnected
 
     isConnectedF.flatMap { _ =>
-      //disconnect here
-      client ! Tcp.Abort
-      val isDisconnectedF =
-        TestAsyncUtil.retryUntilSatisfied(peerMessageReceiver.isDisconnected,
-                                          duration = 1.seconds)
+      val isDisconnectedF = for {
+        p2pClient <- p2pClientF
+        _ = p2pClient.actor ! Tcp.Abort
+        isDisconnected <- TestAsyncUtil.retryUntilSatisfiedF(
+          p2pClient.isDisconnected,
+          duration = 1.seconds)
+      } yield isDisconnected
 
       isDisconnectedF.map { _ =>
         succeed
--- a/node/src/main/scala/org/bitcoins/node/SpvNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/SpvNode.scala
@@ -2,76 +2,115 @@ package org.bitcoins.node
 
 import akka.actor.ActorSystem
 import org.bitcoins.chain.api.ChainApi
-import org.bitcoins.node.models.Peer
+import org.bitcoins.chain.blockchain.ChainHandler
+import org.bitcoins.chain.config.ChainAppConfig
+import org.bitcoins.chain.models.BlockHeaderDAO
+import org.bitcoins.core.bloom.BloomFilter
+import org.bitcoins.core.p2p.NetworkPayload
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.db.P2PLogger
+import org.bitcoins.node.config.NodeAppConfig
+import org.bitcoins.node.models.{
+  BroadcastAbleTransaction,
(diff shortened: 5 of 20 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -16,7 +16,8 @@ import scala.concurrent.{ExecutionContext, Future}
   */
 case class ChainHandler(
     blockHeaderDAO: BlockHeaderDAO,
```
torkelrogstad, line 23: Chances are you're aware of this, but IMO one goal here should be to remove this parameter from the class, and then give it a method `persistToDisk(dao)` (or something similar) that writes once

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -89,3 +95,24 @@ case class ChainHandler(
     }
   }
 }
+
+object ChainHandler {
+
+  def apply(blockHeaderDAO: BlockHeaderDAO, chainConfig: ChainAppConfig)(
```
torkelrogstad: So the idea is to provide one constructor for DB access (returning a future), and one without DB access returning instantly? I think that's a good idea. Although it might make sense to give them more descriptive names, lately I've been leaning towards that having a bunch of `apply` methods can be confusing and overwhelming

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -43,17 +44,22 @@ case class ChainHandler(
...
         //now we have successfully connected the header, we need to insert
         //it into the database
         val createdF = blockHeaderDAO.create(updatedHeader)
         createdF.map { header =>
           logger.debug(
             s"Connected new header to blockchain, height=${header.height} hash=${header.hashBE}")
-          ChainHandler(blockHeaderDAO, chainConfig)
+
+          //what about old chain? The returned chain is _not_ necessarily the best chain I think
+          ChainHandler(blockHeaderDAO, chainConfig, Vector(newChain))
```
torkelrogstad: Under what circumstances is this _not_ true?

torkelrogstad (commented):
Concept ACK

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
Christewart: Our `ChainHandler` has a `Vector[Blockchain]`, while `BlockchainUpdate.Successful()` only returns _one_ chain. This is effectively discarding the competing chain from `ChainHandler`. There isn't even a guarantee here that the chain that got updated is the best chain.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
Christewart: No, the other constructor is simply a convinence method for passing in one `Blockchain`. It simply does `Vector(blockchain)` and then calls the default case class constructor. I suppose I could use some sort of vargs here but I'm unsure about what collection it defaults to and what the performance implications are.

Christewart:
As of 7a70099d

1. Passing state of our PeerMessageReceiver, DataMessageHandler, P2PClient etc in return types rather than caching internally

2. Keep chain tips in memory so we don't need to do a database read for every new tip we are validating, this also fixes bugs we had in our chain validation logic
The tests do not pass, there is still issues with passing updated states around that I'm trying to think through.

The tough thing here is managing the state that is updated in `PeerMessageReceiver` after a message is processed. For instance, if we a receive a block header we need to update our chain and return a `Future[ChainApi]` that represents the new chain state. 

For instance, our `SpvNode` class has a `sync()` method that queries what our best hash is with this:

```scala
      header <- chainApi
        .getHeader(hash)
        .map(_.get) // .get is safe since this is an internal call
```

However, the `ChainApi` our `SpvNode` has _is not_ the one being updated in `PeerMessageReceiver`. This means we will be sending a get headers message for something that _may not_ be our best tip any more. This is the complexity that is introduced with caching.

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -26,6 +26,9 @@ case class Blockchain(headers: Vector[BlockHeaderDb])
     with BitcoinSLogger {
   val tip: BlockHeaderDb = headers.head
 
+  //this could throw if all we have is the genesis
+  lazy val secondTip: BlockHeaderDb = headers(1)
```
torkelrogstad: This seems dangerous to have in here. It's only used one place. Does that justify introducing a dangerous val? In any case, should probably have a Scaladoc explaining what the purpose here is, and that it is/why it is dangerous

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -81,11 +105,32 @@ case class ChainHandler(
...
     //https://bitcoin.org/en/glossary/block-chain
-    blockHeaderDAO.chainTips.map { tips =>
-      val sorted = tips.sortBy(header => header.blockHeader.difficulty)
-      val hash = sorted.head.hashBE
-      logger.debug(s"getBestBlockHash result: hash=$hash")
-      hash
-    }
+    val groupedChains = blockchains.groupBy(_.tip.height)
+    val maxHeight = groupedChains.keys.max
+    val chains = groupedChains(maxHeight)
```
torkelrogstad, line 146: To make this log more meaningful, could do something like this: 

```scala

    val hashBE: DoubleSha256DigestBE = chains match {
      case Vector() =>
        val errMsg = s"Did not find blockchain with height $maxHeight"
        logger.error(errMsg)
        throw new RuntimeException(errMsg)
      case chain +: Vector() =>
        chain.tip.hashBE
      case chain +: rest =>
        logger.warn(
          s"We have multiple competing blockchains: ${(chain +: rest).map(_.tip.hashBE.hex).mkString(", ")}")
        chain.tip.hashBE
    }
```

node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
```diff
@@ -27,16 +27,18 @@ class SpvNodeTest extends NodeUnitTest {
       val spvNode = spvNodeConnectedWithBitcoind.spvNode
       val bitcoind = spvNodeConnectedWithBitcoind.bitcoind
 
-      assert(spvNode.isConnected)
-
-      assert(spvNode.isInitialized)
+      val assert1F = for {
+        a1 <- spvNode.isConnected.map(assert(_))
```
torkelrogstad:
```suggestion
         _ <- spvNode.isConnected.map(assert(_))
```

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -36,16 +37,25 @@ case class SpvNode(
     extends BitcoinSLogger {
   import system.dispatcher
 
+  implicit private val timeout = akka.util.Timeout(10.seconds)
```
torkelrogstad, line 45: Could we do a comment explaining what this implicit affects?

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -36,16 +37,25 @@ case class SpvNode(
...
   private val txDAO = BroadcastAbleTransactionDAO(SQLiteProfile)
 
-  private val peerMsgRecv =
-    PeerMessageReceiver.newReceiver(callbacks)
+  private val clientF: Future[P2PClient] = {
```
torkelrogstad, line 60: Why is this a future?

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
torkelrogstad: Ok, I see. What's your thoughts on renaming the first `apply`, to something like `fromDatabase`? To signify that its querying the DB for the current state

node/src/main/scala/org/bitcoins/node/SpvNodeMain.scala
```diff
@@ -55,14 +55,21 @@ object SpvNodeMain extends App with BitcoinSLogger {
...
 
-    def run: Unit =
-      spvNodeF
-        .flatMap(_.chainApi.getBlockCount)
-        .foreach(count => logger.debug(s"SPV block height: $count"))
+    def run: Unit = {
+      for {
+        spvNode <- spvNodeF
+        count <- spvNode.chainApi.getBlockCount
+      } yield {
```
torkelrogstad: I get a compiler warning here: It's because we `yield` something in a unit-returning function. This removed the warning:
```scala
   def run: Unit =
      for {
        spvNode <- spvNodeF
        count <- spvNode.chainApi.getBlockCount
      } logger.debug(s"SPV block height: $count")
```

torkelrogstad: That being said this is a leftover from some debugging I did, could probably delete this file altogether (the same functionality is achieved by running the chain+node+wallet server

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -78,8 +84,12 @@ case class P2PClientActor(
       self.forward(networkMsg)
     case message: Tcp.Message =>
       val newUnalignedBytes =
-        handleTcpMessage(message, Some(peer), unalignedBytes)
+        Await.result(handleTcpMessage(message, Some(peer), unalignedBytes),
```
torkelrogstad, line 92: Is the purpose of this `Await` to get rid of the compiler warning? In that case, this is better IMO:


```scala
   case connected: Tcp.Connected =>
      val _ = handleEvent(connected, unalignedBytes = ByteVector.empty)
      ()
```

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -90,16 +100,20 @@ case class P2PClientActor(
       //after receiving Tcp.Connected we switch to the
       //'awaitNetworkRequest' context. This is the main
       //execution loop for the Client actor
-      val _ = handleCommand(cmd, peer = None)
+      handleCommand(cmd, peer = None)
 
     case connected: Tcp.Connected =>
-      val _ = handleEvent(connected, unalignedBytes = ByteVector.empty)
+      Await.result(handleEvent(connected, unalignedBytes = ByteVector.empty),
```
torkelrogstad, line 111: Same as above

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -196,13 +218,27 @@ case class P2PClientActor(
         //send them to 'context.parent' -- this is the
         //PeerMessageHandler that is responsible for
         //creating this Client Actor
-        messages.foreach { m =>
-          val msg = NetworkMessageReceived(m, P2PClient(self, peer))
-          peerMsgHandlerReceiver.handleNetworkMessageReceived(msg)
-
+        val newMsgReceiverF: Future[PeerMessageReceiver] = {
+          val currentF = Future.successful(currentPeerMsgHandlerRecv)
```
torkelrogstad: This is hard to read... What's going on here? 

It looks similar to `FutureUtil.sequentially`, except that it has a custom initial value for the `foldLeft`. Would it be possible to add a method `FutureUtil.sequentiallyWithInit`, or something along those lines, and implement `FutureUtil.sequentially` as a special case of `FutureUtil.sequentiallyWithInit`?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -222,6 +258,19 @@ case class P2PClientActor(
         manager ! bind
     }
 
+  /**
+    * Returns the current state of our peer given the [[P2PClient.MetaMsg]]
+    * @param metaMsg
```
torkelrogstad: Unused `param` and `return` annotations

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -235,10 +284,45 @@ case class P2PClientActor(
...
 
-case class P2PClient(actor: ActorRef, peer: Peer)
+case class P2PClient(actor: ActorRef, peer: Peer) {
+  import akka.pattern.ask
+
+  def isConnected()(
+      implicit timeout: Timeout,
+      ec: ExecutionContext): Future[Boolean] = {
+    val isConnectedF = (actor.ask(P2PClient.IsConnected)).mapTo[Boolean]
+    isConnectedF.recoverWith { case _: Throwable => Future.successful(false) }
```
torkelrogstad: Is it possible to match on a less generic error type?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -235,10 +284,45 @@ case class P2PClientActor(
...
+  }
+
+  def isInitialized()(
+      implicit timeout: Timeout,
+      ec: ExecutionContext): Future[Boolean] = {
+    val isInitF = actor.ask(P2PClient.IsInitialized).mapTo[Boolean]
+    isInitF.recoverWith { case _: Throwable => Future.successful(false) }
```
torkelrogstad: Same as above

torkelrogstad: Same as above

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -235,10 +284,45 @@ case class P2PClientActor(
...
+
+  def isDisconnected()(
+      implicit timeout: Timeout,
+      ec: ExecutionContext): Future[Boolean] = {
+    val isDisconnect: Future[Boolean] =
+      actor.ask(P2PClient.IsDisconnected).mapTo[Boolean]
+
+    //this future can be failed, as we stop the P2PClientActor if we send a disconnect
+    //if that actor has been killed, the peer _has_ to have been disconnected
+    isDisconnect.recoverWith { case _: Throwable => Future.successful(true) }
```
torkelrogstad: Same as above

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -235,10 +284,45 @@ case class P2PClientActor(
...
+  }
+}
 
 object P2PClient {
 
+  sealed trait MetaMsg
```
torkelrogstad, line 324: Scaladoc on this and case objects below

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -233,18 +247,53 @@ object PeerMessageReceiver {
...
-    new PeerMessageReceiver(state, callbacks)
+    import ref.dispatcher
+    val dataHandler = new DataMessageHandler(chainApi, callbacks)
+    new PeerMessageReceiver(dataMessageHandler = dataHandler,
+                            state = state,
+                            peer = peer,
+                            callbacks = callbacks)
+  }
+
+  def preConnection(peer: Peer, callbacks: SpvNodeCallbacks)(
```
torkelrogstad, line 264: Scaladoc

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
Christewart: So this `Await` is here on purpose. Actor mailboxes messages are queued and processing synchronously. If we return a `Future`, it looks like we have completed processing. This `Await` forces the future to be completed before we return, so the Actor doesn't move onto the next message

torkelrogstad: But won't this cause everything else to stop? I'm assuming Akka supports a pattern like this without blocking threads

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
Christewart: The goal here is to process the messages sequentially. 

To process the next message we need the previous message to process, and then we need to use the `PeerMessageReceiver` that is returned from the previously processed message.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
Christewart: I can confirm these are needed, and haven't found a alternative yet

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -71,31 +70,38 @@ class DataMessageHandler(callbacks: SpvNodeCallbacks)(
...
-        chainApiF.map { newApi =>
-          val lastHeader = headers.last
-          val lastHash = lastHeader.hash
-          newApi.getBlockCount.map { count =>
-            logger.trace(
-              s"Processed headers, most recent has height=$count and hash=$lastHash.")
+        val processedHeadersF = chainApi.processHeaders(headers)
+
+        processedHeadersF.flatMap { newApi =>
+          val sendHeadersMsgF = for {
```
torkelrogstad: ping

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -42,18 +44,41 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
...
-      case BlockchainUpdate.Successful(_, updatedHeader) =>
+      case BlockchainUpdate.Successful(newChain, updatedHeader) =>
         //now we have successfully connected the header, we need to insert
         //it into the database
         val createdF = blockHeaderDAO.create(updatedHeader)
         createdF.map { header =>
           logger.debug(
             s"Connected new header to blockchain, height=${header.height} hash=${header.hashBE}")
-          ChainHandler(blockHeaderDAO)
+          val chainIdxOpt = blockchains.zipWithIndex.find {
```
torkelrogstad, line 67: You don't need `zipWithIndex` here

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -42,18 +44,41 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
...
+            case (chain, _) =>
+              val oldTip = newChain(1) //should be safe, even with genesis header as we just connected a tip
+              oldTip == chain.tip
+          }
+
+          val updatedChains = {
+            chainIdxOpt match {
+              case Some((_, idx)) =>
+                logger.trace(
+                  s"Updating chain at idx=${idx} out of ${blockchains.length} with new tip=${header.hashBE.hex}")
```
torkelrogstad: nit: I'd add to the log: `out of competing chains=${blockchains.length}`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -81,11 +106,39 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
...
+          s"We have multiple competing blockchains: ${(chain +: rest).map(_.tip.hashBE.hex).mkString(", ")}")
+        chain.tip.hashBE
     }
+    Future.successful(hashBE)
+  }
+}
+
+object ChainHandler {
+
+  def fromDatabase(blockHeaderDAO: BlockHeaderDAO)(
```
torkelrogstad, line 169: Scaladoc

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -2,46 +2,79 @@ package org.bitcoins.node
...
+  /** This implicit is required for using the [[akka.pattern.ask akka ask]]
+    * to query what the state of our node is, like [[isConnected isConnected]]
+    * */
+  implicit private val timeout = akka.util.Timeout(10.seconds)
   private val txDAO = BroadcastAbleTransactionDAO(SQLiteProfile)
 
-  private val peerMsgRecv =
-    PeerMessageReceiver.newReceiver(chainApi, callbacks)
+  /** This is constructing a chain api from disk every time we call this method
+    * This involves database calls which can be slow and expensive to construct
```
torkelrogstad, line 49: Two nits: I think appending an `F` to a method name because it returns a future isn't the best thing to do, the type signature tells you about the future part. Also, add empty parentheses to the method? Makes it clearer that its a side effecting call. I think `readChainHandlerFromDIsk()` is a nice and descriptive name. It shouldn't be too ergonomic to type out either, considering how expensive it is

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -2,46 +2,79 @@ package org.bitcoins.node
...
+    * our [[org.bitcoins.chain.blockchain.Blockchain Blockchain]]
+    * */
+  def chainApiF: Future[ChainApi] = {
+    ChainHandler.fromDatabase(BlockHeaderDAO())
+  }
 
-  private val client: P2PClient =
-    P2PClient(context = system, peer = peer, peerMessageReceiver = peerMsgRecv)
+  /** Unlike our chain api, this is cached inside our [[SpvNode spv node]]
```
torkelrogstad: The Scaladoc links here are broken. All classes have to be referenced by their full package path (which is stupid, it should follow normal scoping/import rules). Aside from that, I don't think there's much value in making the links clickable. There's already a ton of other clickable links (in method return types, parameter types, etc) in the Scaladoc for this class

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -2,46 +2,79 @@ package org.bitcoins.node
...
+      val p2p = P2PClient(context = system,
+                          peer = peer,
+                          peerMessageReceiver = peerMsgRecv)
+      p2p
+    }
+  }
 
-  private val peerMsgSender: PeerMessageSender = {
-    PeerMessageSender(client)
+  private val peerMsgSenderF: Future[PeerMessageSender] = {
```
torkelrogstad, line 75: Could we add a Scaladoc explaning why this is a future?

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -50,30 +83,35 @@ case class SpvNode(
...
     for {
       _ <- nodeAppConfig.initialize()
       node <- {
-        peerMsgSender.connect()
+        val connectedF = peerMsgSenderF.map(_.connect())
 
-        val isInitializedF =
-          AsyncUtil.retryUntilSatisfied(peerMsgRecv.isInitialized)
+        val isInitializedF = for {
+          _ <- connectedF
```
torkelrogstad:
```suggestion
          _ <- peerMsgSenderF.map(_.connect())
```

No real need for the intermediate val

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -50,30 +83,35 @@ case class SpvNode(
...
-        isInitializedF.failed.foreach(err =>
-          logger.error(s"Failed to connect with peer=$peer with err=${err}"))
+        isInitializedF.failed.foreach(
+          err =>
+            logger(nodeAppConfig).error(
+              s"Failed to connect with peer=$peer with err=${err}"))
 
         isInitializedF.map { _ =>
-          logger.info(s"Our peer=${peer} has been initialized")
+          logger(nodeAppConfig).info(s"Our peer=${peer} has been initialized")
```
torkelrogstad, line 142: As you've mentioned before, this is very unfortunate... I think we should go the route you proposed, and make P2P logger take a node config, chain verification logger take a chain config, etc. Should hopefully clear most of these up

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -82,37 +120,46 @@ case class SpvNode(
 
   /** Stops our spv node */
   def stop(): Future[SpvNode] = {
```
torkelrogstad, line 155: Since we're logging the shutdown on line 131 it'd be nice to add a line at the start of the method, "Stopping SPV node", or something like that

node/src/main/scala/org/bitcoins/node/SpvNodeMain.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import scala.concurrent._
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+import org.bitcoins.rpc.config.BitcoindInstance
+import java.net.InetSocketAddress
+import org.bitcoins.core.bloom.BloomFilter
+import org.bitcoins.core.bloom.BloomUpdateAll
+
+object SpvNodeMain extends App with BitcoinSLogger {
```
torkelrogstad: Looks like this file is new in this PR, believe I deleted it in a different PR that you rebased onto. IMO we can just keep it deleted

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
torkelrogstad: Could we leave a comment in the code, and maybe also open an issue for this? There's got to be a better way...

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -128,43 +141,51 @@ case class P2PClientActor(
...
-        unalignedBytes
+        Future.successful(unalignedBytes)
       case Tcp.CommandFailed(command) =>
         logger.debug(s"Client Command failed: ${command}")
 
-        unalignedBytes
+        Future.successful(unalignedBytes)
       case Tcp.Connected(remote, local) =>
-        logger.debug(s"Tcp connection to: ${remote}")
+        logger.info(s"Tcp connection to: ${remote}")
```
torkelrogstad: ping, forgot to turn this down?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -128,43 +141,51 @@ case class P2PClientActor(
...
   private def handleEvent(
       event: Tcp.Event,
-      unalignedBytes: ByteVector): ByteVector = {
+      unalignedBytes: ByteVector): Future[ByteVector] = {
+    import context.dispatcher
     event match {
       case Tcp.Bound(localAddress) =>
-        logger.debug(
-          s"Actor is now bound to the local address: ${localAddress}")
+        logger.info(s"Actor is now bound to the local address: ${localAddress}")
```
torkelrogstad: ping, forgot to turn this down?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
torkelrogstad: Could we factor this out by adding something like this to `FutureUtil`?

```scala

  def foldLeft[T, U](init: T, items: Seq[U])(fun: (T, U) => Future[T])(
      implicit ec: ExecutionContext): Future[T] = {
    items.foldLeft(Future.successful(init)) {
      case (accumF, elem) =>
        accumF.flatMap { accum =>
          fun(accum, elem)
        }
    }
  }
```

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
torkelrogstad: We probably only want to interpret a network related error as `false`

node/src/main/scala/org/bitcoins/node/networking/peer/PeerHandler.scala
```diff
@@ -58,6 +60,4 @@ object PeerHandler {
 }
  */
 
-case class PeerHandler(
-    peerMsgRecv: PeerMessageReceiver,
-    peerMsgSender: PeerMessageSender)
+case class PeerHandler(peerMsgRecv: P2PClient, peerMsgSender: PeerMessageSender)
```
torkelrogstad: rename first paramter?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -231,20 +247,52 @@ object PeerMessageReceiver {
...
+    import ref.dispatcher
+    val dataHandler = new DataMessageHandler(chainApi, callbacks)
+    new PeerMessageReceiver(dataMessageHandler = dataHandler,
+                            state = state,
+                            peer = peer,
+                            callbacks = callbacks)
+  }
+
+  /**
+    * Creates ta [[PeerMessageReceiver PeerMessageReceiver]] that is ready
```
torkelrogstad: typo in second word

torkelrogstad: by the way, same comment as above wrt to Scaladoc links applies here also

node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
```diff
@@ -152,26 +164,31 @@ class P2PClientTest
...
-    val client =
-      TestActorRef(P2PClient.props(peer, peerMessageReceiver), probe.ref)
-
-    client ! Tcp.Connect(remote)
 
-    val isConnectedF =
-      TestAsyncUtil.retryUntilSatisfied(peerMessageReceiver.isInitialized)
+    val isConnectedF = for {
+      p2pClient <- p2pClientF
+      _ <- p2pClientF.map(p2pClient => p2pClient.actor ! Tcp.Connect(remote))
```
torkelrogstad:
```suggestion
      _ <- p2pClient.actor ! Tcp.Connect(remote)
```

node-test/src/test/scala/org/bitcoins/node/SpvNodeTest.scala
```diff
@@ -62,7 +64,9 @@ class SpvNodeTest extends NodeUnitTest {
       //as they happen with the 'sendheaders' message
       //both our spv node and our bitcoind node _should_ both be at the genesis block (regtest)
       //at this point so no actual syncing is happening
-      val initSyncF = gen1F.flatMap(_ => spvNode.sync())
+      val initSyncF = gen1F.flatMap { hashes =>
```
torkelrogstad: unused param

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -60,53 +61,57 @@ object Blockchain extends ChainVerificationLogger {
...
-                      BlockchainUpdate.Successful(newChain, headerDb)
-                    case fail: TipUpdateResult.Failure =>
-                      BlockchainUpdate.Failed(blockchain, header, fail)
-                  }
-                }
+    val tipResultF: Future[BlockchainUpdate] = {
+      val nested: Vector[Future[BlockchainUpdate]] = blockchains.map {
+        blockchain =>
+          val prevBlockHeaderIdxOpt =
+            blockchain.headers.zipWithIndex.find {
```
torkelrogstad, line 77: Don't need `zipWithindex`?

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
Christewart: It's used below

node/src/main/scala/org/bitcoins/node/SpvNode.scala
Christewart: I filed an issue to resolve this on another PR #670

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -60,53 +61,57 @@ object Blockchain extends ChainVerificationLogger {
...
+              case (headerDb, _) =>
```
nkohen, line 78: Index is not used here right? It is not too cheap an operation to add them so we probably shouldn't.

Is there any way to know if find is going from genesis to tip or tip to genesis? If we can make it do the former we should right?

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
nkohen: Where?

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -81,11 +106,39 @@ case class ChainHandler(blockHeaderDAO: BlockHeaderDAO)(
...
+      implicit ec: ExecutionContext,
+      chainConfig: ChainAppConfig): Future[ChainHandler] = {
```
nkohen, line 171: Since this takes in these implicits, couldn't we just construct a `BlockHeaderDAO` from them?

node/src/main/scala/org/bitcoins/node/SpvNodeMain.scala
```diff
@@ -0,0 +1,89 @@
...
+    logger.info(
+      s"Connecting to bitcoind running on ${bitcoind.network} at ${bitcoind.uri} ")
+    val socket =
+      new InetSocketAddress(bitcoind.uri.getHost(), bitcoind.p2pPort)
+    Peer(socket)
+  }
+
+  logger.info(s"Starting SPV node")
+
+  val emptyBloom =
```
nkohen: nit: Is there a reason not to use `BloomFilter.empty` here?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
nkohen: https://stackoverflow.com/questions/36592074/wait-on-asynchronous-future-call-before-processing-next-message-in-akka

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -194,17 +215,24 @@ case class P2PClientActor(
...
-        messages.foreach { m =>
-          val msg = NetworkMessageReceived(m, P2PClient(self, peer))
-          peerMsgHandlerReceiver.handleNetworkMessageReceived(msg)
-
+        val newMsgReceiverF: Future[PeerMessageReceiver] = {
+          val currentF = Future.successful(currentPeerMsgHandlerRecv)
+          logger.trace(s"About to process ${messages.length} messages")
+          messages.foldLeft(currentF) {
+            case (peerMsgRecvF, m) =>
+              peerMsgRecvF.flatMap { peerMsgRecv: PeerMessageReceiver =>
```
nkohen: Might want to consider calling `recover` on this so that one message handling failure doesn't result in other dropped messages?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -152,68 +163,73 @@ class PeerMessageReceiver(
...
             //we want peers to just send us headers
             //we don't want to have to request them manually
             sender.sendHeadersMessage()
 
-            Success(())
+            val newRecv =
+              new PeerMessageReceiver(dataMessageHandler = dataMessageHandler,
+                                      state = newState,
+                                      peer = peer,
+                                      callbacks = callbacks)
```
nkohen: nit: This seems to be a regular pattern, maybe having a method on the old receiver that takes in a newState would be nice?

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
Christewart: Don't we want to do tip to genesis? That is what we are doing currently. We are more likely to have a reorg closer to the tip.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
Christewart: I would like to get away from passing around tons of implicits.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
Christewart: #671

chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
Christewart: the index is also used below

Christewart:
Some very unscientific benchmarks when running `chainTest/test` off of

pre chain handler optimization: 6258946f22b5d7522c220a5906de1390826d8456

```
[info] ScalaTest
[info] Run completed in 3 minutes, 9 seconds.
[info] Total number of tests run: 27
[info] Suites: completed 8, aborted 0
[info] Tests: succeeded 27, failed 0, canceled 0, ignored 0, pending 0
```


a16716b 
```
[info] Run completed in 1 minute, 44 seconds.
[info] Total number of tests run: 27
[info] Suites: completed 8, aborted 0
[info] Tests: succeeded 27, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 27, Failed 0, Errors 0, Passed 27
[success] Total time: 107 s, completed Aug 6, 2019 12:53:38 PM
```

PR #659, opened 2019-07-26
WIP Number Optimization

cwaldron97:
Looking to optimize the number implementation in the code base. Look to PR: #654  and issue: #653 to hopefully optimize and setup the underlying typing more efficiently so less bits have to be used for each number type. This is currently the implementation of the signed 32 bit int which is backed by Scala.Int type as per https://www.scala-lang.org/api/current/scala/Int.html.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -242,7 +242,11 @@ final case class UInt64 private (private val underlying: java.math.BigInteger)
     i
   }
 
-  def toBigInt: BigInteger = underlying
+  def toInt: Int = underlying.intValueExact()
+
+  def toFloat: Float = underlying.floatValue()
+
+  def toDouble: Double = underlying.doubleValue()
 
   def toLong: Long = underlying.longValueExact()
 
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -209,10 +209,66 @@ final case class UInt64(underlying: BigInt)
...
+
+  override def -(y: UInt64): UInt64 = {
+    UInt64(underlying - y.underlying)
+  }
+
+  override def *(y: UInt64): UInt64 = {
+    UInt64(underlying * y.underlying)
+  }
+
+  override def *(factor: BigInt): UInt64 = ???
```
torkelrogstad, line 220:
```scala
override def *(factor: BigInt): UInt64 = ???
      UInt64(underlying * factor)
  }
```

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -209,10 +209,66 @@ final case class UInt64(underlying: BigInt)
...
+    require(underlying == i,
+            s"Rounded when converting long=${underlying} toInt=${i}")
+    i
+  }
+
+  def toBigInt: BigInt = underlying
+
+  def toLong: Long = toBigInt
+
+  def toDouble(x: UInt64): Double = x.underlying.toDouble
```
torkelrogstad: `def toDouble(x: UInt64)`, `def toFloat(x: UInt64)` and `def toBigInt(x: UInt64)` doesn't really make much sense to have as method on this trait. Why would a `NumberType` be responsible for converting external values?

Issue #673, opened 2019-08-04
UpdateBloomFilterTest do not create chain tables


Christewart:
I get this error in logs when running `*UpdateBloomFilterTest*`

```
2019-08-04T23:22:26 ERROR [DATABASE] Error when executing query result [select "height", "hash", "version", "previous_block_hash", "merkle_root_hash", "time", "n_bits", "nonce", "hex" from "block_headers" where "hash" in ('0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206')]
2019-08-04T23:22:26 ERROR [DATABASE] org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: block_headers)
```

The solution to this IMO is creating a new fixture that constructs
1. Bitcoind
2. Chain
3. Wallet
4. Spv node

and calls the appropriate fixture constructors that we have already defined so we don't have to worry about creating tables for chain/wallet.

[The old fixture only setups up wallet/bitcoind](https://github.com/bitcoin-s/bitcoin-s/blob/master/node-test/src/test/scala/org/bitcoins/node/UpdateBloomFilterTest.scala#L27)

Christewart:
Addressed in #680

Issue #675, opened 2019-08-05
MerkleBuffers property tests are not sastifiable reliably


Christewart:
MerkleBufferTests fail reliably for me locally with this error message

```
[info] MerkleBuffers
[info] - must match a merkle block with its corresponding transactions *** FAILED ***
[info]   Gave up after 2 successful property evaluations. 32 evaluations were discarded.
[info]   org.scalatest.exceptions.GeneratorDrivenPropertyCheckFailedException:
[info]   at org.scalatest.enablers.CheckerAsserting$$anon$2.indicateFailure(CheckerAsserting.scala:234)
[info]   at org.scalatest.enablers.CheckerAsserting$$anon$2.indicateFailure(CheckerAsserting.scala:222)
```

More than likely this is because our generators found [here](https://github.com/bitcoin-s/bitcoin-s/commit/b706e8ea452b1c59f5feea4d582d90e5df1b1dd5#diff-ea82f21fe67f2a2ed4d3f9ede5ef1ec2R29)

PR #676, opened 2019-08-05
Address issue #673, also refactor fixture stuff to be in companion ob…

Christewart:
…jects rather than traits so they can be re-used without extending the trait itself. This allows for more modularity with our fixtures

This addresses #673 and is blocking #655 from getting merged. Previously we were not creating chain project tables even though we were relying on the chain project functionality. This PR refactors fixtures to create the chain project tables and creates a encompassing fixture called `SpvNodeFundedWalletBitcoind` that creates 

1.  a funded bitcoind wallet
2.  a funded bitcoin-s wallet
3. a chain handler with the appropriate tables created
4. a spv node that is connected to the bitcoin instance -- but not started!


Christewart:
70959e9 - addresses #675 and fixes small bug we had in the second test case in `UpdateBloomFilterTest`

nkohen (approved):
ACK

PR #681, opened 2019-08-06
Schnorr Compute PubKey from pre-committed R value

nkohen:
Added computeSchnorrPubKey to the jni and added a property test which passses!

This PR is in parallel with changes introduced here: https://github.com/nkohen/secp256k1/tree/sig_pub_key_gen


nkohen:
Superseded by #682

PR #665, opened 2019-07-30
Schnorr Sign with Specified Nonce

nkohen:
This PR is in parallel with changes introduced here: https://github.com/nkohen/secp256k1/tree/specify_schnorr_nonce

It adds `NativeSecp256k1.schnorrSignWithNonce` which allows the user to specify a `k` value for the schnorr signature.

Tests Pass!

PR #657, opened 2019-07-23
Schnorr secp256k1 rebase

nkohen:
Closes #212 

An attempt at rebasing (and squashing) secp256k1 jni commits to be over jonas' schnorrsig branch (https://github.com/bitcoin-core/secp256k1/pull/558)

EDIT: based off of https://github.com/nkohen/secp256k1/tree/schnorr_expirement

PR #682, opened 2019-08-07
Secp256k1 Support for DLCs

nkohen:
This PR includes all of my work related to Schnorr signatures for DLCs.

It introduces the following calls to `NativeSecp256k1`:

- schnorrSign
- schnorrVerify
- schnorrSignWithNonce
- computeSchnorrPubKey
- schnorrPublicNonce

and there are tests for all of these calls that all pass!

This is based off of my branch of secp256k1: https://github.com/nkohen/secp256k1/tree/all_schnorr
This PR contains all work from #665 and #681


nkohen:
TODO: Build binaries for

- [ ] OSX
- [ ] Windows
- [ ] linux 32-bit

Christewart:
rebase needed now that #657 is in

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.bitcoins.testkit.core.gen.NumberGenerator
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import scodec.bits.ByteVector
+
+class NativeSecp256k1ScalaTest extends BitcoinSUnitTest {
+  behavior of "NativeSecp256k1"
+
+  it must "act like a normal schnorrSign if the normal nonce is specified" in {
+    forAll(
+      NumberGenerator.bytevector(32).filter(_.toArray.exists(_ != 0.toByte)),
```
Christewart: Why can't we have have all zeros? A comment would be nice

Christewart: I'm assuming it has to do with EC math?

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
```diff
@@ -0,0 +1,73 @@
...
+      NumberGenerator.bytevector(32)) {
+      case (privKeyBytes, message) =>
+        val privKey = ECPrivateKey.fromBytes(privKeyBytes)
+
+        // This is the nonce specified in bip-schnorr
+        val nonce =
+          CryptoUtil.sha256(privKey.bytes ++ message).bytes
+
+        assert(nonce.toArray.exists(_ != 0.toByte))
```
Christewart: this is a pretty paranoid assert right? the odds of `CryptoUtil.sha256()` outputing all zeros is extremely low?

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
Christewart: as a matter of code style it would be nice to avoid long generators and make a helper method

```scala
def byteVector32NonEmpty = NumberGenerator.bytevector(32).filter(_.toArray.exists(_ != 0.toByte))
```

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import org.bitcoin.NativeSecp256k1
+import org.bitcoins.core.util.CryptoUtil
+import org.bitcoins.testkit.core.gen.NumberGenerator
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import scodec.bits.ByteVector
+
+class NativeSecp256k1ScalaTest extends BitcoinSUnitTest {
+  behavior of "NativeSecp256k1"
+
```
Christewart: Since this is new code I would 

```scala
implicit override val generatorDrivenConfig = generatorDrivenConfigNewCode
```

That makes it so that we execute the property based tests 100 times instead of 10.

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
```diff
@@ -0,0 +1,73 @@
...
+        val sig: Array[Byte] =
+          NativeSecp256k1.schnorrSign(messageArr, privKey.bytes.toArray)
+
+        val (rKey, s) = sig.splitAt(32)
+
+        assert(ByteVector(rBytes.tail) == ByteVector(rKey))
+
+        val sPub: Array[Byte] =
+          NativeSecp256k1.computeSchnorrPubKey(messageArr,
+                                               rKey.+:(rBytes.head),
```
Christewart: What is this here? Can you make an intermediate val and leave a comment on this?

Christewart: I'm guessing it is something to do with a sign bit?

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
```diff
@@ -0,0 +1,73 @@
...
+                                               publicKey.bytes.toArray)
+
+        val realSPriv = ECPrivateKey.fromBytes(ByteVector(s))
+        val realSPub = realSPriv.publicKey
+
+        assert(ECPublicKey.fromBytes(ByteVector(sPub)) == realSPub)
+    }
+  }
+}
```
Christewart: Would it be appropriate to add a unit test here saying that we should be able to reveal the private key if we sign two things with the pre-commited R value?

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import org.bitcoin.NativeSecp256k1
+import scodec.bits.ByteVector
+
+object Schnorr {
+
+  def sign(
+      msg: ByteVector,
+      privateKey: ECPrivateKey): SchnorrDigitalSignature = {
+    if (msg.length != 32) {
```
Christewart: nit: require() instead of `throw new IllegaArugment`

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+        NativeSecp256k1.schnorrSign(msg.toArray, privateKey.bytes.toArray)
+
+      SchnorrDigitalSignature.fromBytes(ByteVector(sigArr))
+    }
+  }
+
+  def signWithNonce(
+      msg: ByteVector,
+      privateKey: ECPrivateKey,
+      nonce: ECPrivateKey): SchnorrDigitalSignature = {
```
Christewart: Is it appropriate to use the `ECPrivateKey` type here?

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
+import scodec.bits.ByteVector
+
+sealed abstract class SchnorrDigitalSignature extends BitcoinSLogger {
+  require(rx.length == 32, s"R must be 32 bytes, got $rx")
+  require(s.length == 32, s"s must be 32 bytes, got $s")
+
```
Christewart, line 11: Just to be safe I would require `bytes` must be 64 in length to.

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
+import scodec.bits.ByteVector
+
+sealed abstract class SchnorrDigitalSignature extends BitcoinSLogger {
```
Christewart: This should probably `extend NetworkElement`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /** Represents the x coordinate of the R value used in the signature */
+  def rx: ByteVector
+
+  /** The s = k + e*x value of the signature */
+  def s: ByteVector
+
+  def bytes: ByteVector = rx ++ s
+
+  def hex: String = BitcoinSUtil.encodeHex(bytes)
```
Christewart: if we extend `NetworkElement` we won't have to define this

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      SchnorrDigitalSignatureImpl(rx, s)
+    }
+  }
+
+  def apply(r: ECPublicKey, s: ByteVector): SchnorrDigitalSignature = {
+    if (s.length != 32) {
+      throw new IllegalArgumentException(s"s must be 32 bytes, got $s")
+    } else {
+      SchnorrDigitalSignatureImpl(r.bytes.tail, s)
```
Christewart: Shouldn't we call this constructor on line 42? 

Also in general, can we have these applies also call `.fromRxS()` and `fromRS()`. The more I write Scala code the more I prefer the explicit `.from()` calls.

Then I think it makes sense to add a scaladoc on both indicating how they are different.

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
nkohen: Yeah, private keys can't be zero in Schnorr sigs because then there's no commitment to the message being signed (i.e. it would be a valid signature for any message)

And I found that there was a private key `Gen` in `CryptoGenerators` so I put it there :)

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
nkohen: yup :) I just figured if it were to ever happen better error message to have this fail than have a really weird message.

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
nkohen: Oops, that should just be `rBytes` (as you can see from the assert right above this :P)

core-test/src/test/scala/org/bitcoins/core/crypto/NativeSecp256k1ScalaTest.scala
nkohen: Man this took wayy longer than I expected

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,63 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoin.NativeSecp256k1
+import scodec.bits.ByteVector
+
+object Schnorr {
```
torkelrogstad, line 6: Would be nice with some Scaladocs on these methods

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
+import scodec.bits.ByteVector
+
+sealed abstract class SchnorrDigitalSignature extends NetworkElement {
```
torkelrogstad: This could just be a direct case class IMO, no need for the abstract class/companion object/Impl pattern (leads to more allocation as well as more verbose code)

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
...
+  /** The s = k + e*x value of the signature */
+  def s: ByteVector
+
+  override def bytes: ByteVector = rx ++ s
+
+  override def toString: String = s"SchnorrDigitalSignature($rx, $s)"
+}
+
+final case object DummySchnorrDigitalSignature extends SchnorrDigitalSignature {
+  override val rx: ByteVector = ByteVector.fill(32)(0.toByte)
```
torkelrogstad: nit: `ByteVector.low(32)`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
...
+  override val s: ByteVector = rx
+}
+
+object SchnorrDigitalSignature extends Factory[SchnorrDigitalSignature] {
+  private case class SchnorrDigitalSignatureImpl(rx: ByteVector, s: ByteVector)
+      extends SchnorrDigitalSignature
+
+  override def fromBytes(bytes: ByteVector): SchnorrDigitalSignature = {
+    if (bytes.length != 64) {
```
torkelrogstad: `require(bytes.length == 64,  s"SchnorrDigitalSignature must be 64 bytes, got $bytes")`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
...
+      val (rx, s) = bytes.splitAt(32)
+      SchnorrDigitalSignatureImpl(rx, s)
+    }
+  }
+
+  /** Constructs a SchnorrDigitalSignature from the x-coordinate of the R
+    * value and the s value.
+    */
+  def fromRxS(rx: ByteVector, s: ByteVector): SchnorrDigitalSignature = {
+    if (rx.length != 32) {
```
torkelrogstad: `require(rx.length == 32, s"rx must be 32 bytes, got $rx")`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
...
+      throw new IllegalArgumentException(s"rx must be 32 bytes, got $rx")
+    } else if (s.length != 32) {
```
torkelrogstad: `require(s.length == 32, s"s must be 32 bytes, got $s")`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  def apply(rx: ByteVector, s: ByteVector): SchnorrDigitalSignature = {
+    fromRxS(rx, s)
+  }
+
+  /** Consturcts a SchnorrDigitalSignature form the R and s values. */
+  def fromRS(r: ECPublicKey, s: ByteVector): SchnorrDigitalSignature = {
+    if (s.length != 32) {
```
torkelrogstad: `require(s.length == 32, s"s must be 32 bytes, got $s")`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,43 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.{CryptoUtil, Factory}
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+sealed abstract class SchnorrNonce extends NetworkElement {
```
torkelrogstad: Same comment as above, with the abstract class/comp object/impl pattern

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,43 @@
...
+object SchnorrNonce extends Factory[SchnorrNonce] {
+  private case class SchnorrNonceImpl(bytes: ByteVector) extends SchnorrNonce {
+    override def toString: String = {
+      s"SchnorrNonce($bytes)"
+    }
+  }
+
+  @tailrec
+  override def fromBytes(bytes: ByteVector): SchnorrNonce = {
+    require(bytes.length <= 33, s"Schnorr Nonce must be 32 bytes, got $bytes")
```
torkelrogstad: `must be less than 33` as the message instead? Seeing as we pad shorter nonces

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,43 @@
...
+    else if (bytes.size < 32) {
+      //means we need to pad the private key with 0 bytes so we have 32 bytes
+      SchnorrNonce.fromBytes(bytes.padLeft(32))
+    } //this is for the case when java serialies a BigInteger to 33 bytes to hold the signed num representation
+    else {
+      SchnorrNonce.fromBytes(bytes.tail)
+    }
+  }
+
+  def fromBipSchnorr(
```
torkelrogstad, line 42: scaladoc

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -568,6 +568,45 @@ public static boolean isValidPubKey(byte[] pubkey) {
...
+
+        r.lock();
+        try {
+            retByteArray = secp256k1_schnorrsig_compute_pubkey(byteBuff, Secp256k1Context.getContext());
+        } finally {
+            r.unlock();
+        }
+
+        byte[] sigPubKey = retByteArray[0];
+        int retVal = new BigInteger(new byte[] { retByteArray[1][0] }).intValue();
```
torkelrogstad, line 603: Could we do a comment on the intention here? Seems very cryptic

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CryptoGenerators.scala
```diff
@@ -109,6 +109,13 @@ sealed abstract class CryptoGenerators {
 
   def privateKey: Gen[ECPrivateKey] = Gen.delay(ECPrivateKey())
 
+  /** Generates a random non-zero private key */
```
torkelrogstad, line 112: The comment here doesn't match the code, does it? From reading the code it looks like we're generating a privkey where none of the bytes are zero, not the privkey itself

testkit/src/main/scala/org/bitcoins/testkit/core/gen/NumberGenerator.scala
```diff
@@ -98,6 +98,9 @@ trait NumberGenerator {
   /** Generates an arbitrary [[scodec.bits.ByteVector ByteVector]] */
   def bytevector: Gen[ByteVector] = Gen.listOf(byte).map(ByteVector(_))
 
+  def bytevector(num: Int): Gen[ByteVector] =
```
torkelrogstad, line 102: `/** Generates a ByteVector with the given length */`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,71 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
```
torkelrogstad: Could of unused imports up here

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
nkohen: If this were a case class then I couldn't have the `DummySchnorrDigitalSignature` could I?

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
nkohen: I would if I understood it :P

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CryptoGenerators.scala
nkohen: The private key is technically a number (mod the order of the secp256k1 group), so I don't think it is too confusing, the bytes are just the serialization and its most efficient to check that rather than doing something like `NumberUtil.toUnsignedInt(privKey.bytes) != 0`

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+
+        val realSPriv = ECPrivateKey.fromBytes(ByteVector(s))
+        val realSPub = realSPriv.publicKey
+
+        assert(sPubKey == realSPub)
+    }
+  }
+
+  it should "leak keys if two messages are signed" in {
+    // The order of the secp256k1 curve
```
Christewart: There needs to be more context added to this. Why should we be able to leak keys if two messages are signed? At least a link to why this is the case, or a paragraph explanation.

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+class SchnorrTest extends BitcoinSUnitTest {
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
+
+  behavior of "NativeSecp256k1"
+
+  it must "act like a normal schnorrSign if the normal nonce is specified" in {
+    forAll(CryptoGenerators.nonZeroPrivKey, NumberGenerator.bytevector(32)) {
+      case (privKey, message) =>
+        val nonce = SchnorrNonce.fromBipSchnorr(privKey, message)
```
Christewart, line 19: comment saying 
```
//first, let's manually generate the nonce specified in bip-schnorr inside of bitcoin-s and then pass it into secp256k1 and get a sig back
```

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+
+        assert(nonce.bytes.toArray.exists(_ != 0.toByte))
+
+        val sigWithNonce =
+          NativeSecp256k1.schnorrSignWithNonce(message.toArray,
+                                               privKey.bytes.toArray,
+                                               nonce.bytes.toArray)
+
+        assert(
```
Christewart, line 30:
```
//assert we can deterministically sign when specifying a nonce
```

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+          SchnorrDigitalSignature(ByteVector(sigWithNonce)) ==
+            Schnorr.signWithNonce(message, privKey, nonce)
+        )
+
+        val sig =
+          NativeSecp256k1.schnorrSign(message.toArray, privKey.bytes.toArray)
+
+        assert(
```
Christewart, line 40:
```
//make sure secp256k1 deterministically signs when just passing in a message and a privKey
```

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+          SchnorrDigitalSignature(ByteVector(sig)) ==
+            Schnorr.sign(message, privKey)
+        )
+
+        assert(ByteVector(sigWithNonce) == ByteVector(sig))
```
Christewart, line 46:
```
// assert that our nonce generation inside of `SchnorrNonce.fromBipSchnorr()` is the same as secp256k1, if it was not we will receive diffierent digital signatures here, which means we should fail the test
```

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
torkelrogstad: As long as the object is not a `case object` it works out nicely: https://github.com/bitcoin-s/bitcoin-s/blob/9101aece9b9e48df741e668fce66edbf02f59afc/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutPoint.scala#L32

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,166 @@
...
+    forAll(CryptoGenerators.nonZeroPrivKey, NumberGenerator.bytevector(32)) {
+      case (privKey, message) =>
+        val publicKey: ECPublicKey = privKey.publicKey
+        val messageArr: Array[Byte] = message.toArray
+
+        val nonce = SchnorrNonce.fromBipSchnorr(privKey, message)
+
+        assert(nonce.bytes.toArray.exists(_ != 0.toByte))
+
+        val rBytes = nonce.publicKey.bytes
```
Christewart, line 62:
```
//add a self documenting comment of why the nonce's public key is the r value? 
```

I honestly don't understand why this is the case.

Christewart: In general, you are going to have to help 'narrate' me through this test case, I don't fully understand what's going on. Take me step by step through each assertion, and why it should be true, similar to what I kind of did for myself in the first test case in this file.

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,76 @@
...
+      msg.toArray,
+      r.bytes.toArray,
+      publicKey.bytes.toArray)
+
+    ECPublicKey.fromBytes(ByteVector(sigKeyArr))
+  }
+
+  /** Computes the public key associated with a SchnorrNonce as specified in bip-schnorr.
+    * They y-coordinate is chosen to be a quadratic residue.
+    * @see [[https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki]]
```
Christewart: Is there a specific heading that can be linked to here? Or do you need to read the entire document to understand this.

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
Christewart: We probably want to at least make this a trait, so we can `extend AnyVal` in `SchnorrDigitalSignatureImpl()` later on to avoid runtime allocations.

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,46 @@
...
+  def fromBytes(bytes: ByteVector): SchnorrNonce = {
+    require(bytes.length <= 33,
+            s"Schnorr Nonce must be less than 33 bytes, got $bytes")
+
+    if (bytes.length == 32)
+      SchnorrNonce(bytes)
+    else if (bytes.length < 32) {
+      //means we need to pad the private key with 0 bytes so we have 32 bytes
+      SchnorrNonce.fromBytes(bytes.padLeft(32))
+    } //this is for the case when java serialies a BigInteger to 33 bytes to hold the signed num representation
```
Christewart: nit: put this comment inside of the `{ }`

Christewart (commented):
In general, these comments are starting to reflect my lack of understanding of low level cryptography. I hoping the comments where I ask you to "narrate" the assertions we are making in tests will help readers understand what is going on with a beginner level background in cryptography.

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
nkohen: I added these comments, lmk what you think, but RE this specific comment, that's just its definition (`R := nonce.publicKey`)

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
rorp: `secp256k1_schnorrsig_compute_pubkey()` returns a two element array of byte arrays. The first element is a byte array containing the public key, the second one contains just a single byte that represents the error code (`0` means error). Using `BigInteger` here is overkill.
`byte retVal = retByteArray[1][0];` would be more than enough.

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,197 @@
...
+   *
+   * sig1 = nonce + message1*privKey
+   * sig2 = nonce + message2*privKey
+   *
+   * => sig1 - sig2 = (message1 - message2)*privKey
+   * => privKey = (sig1 - sig2) * inverse(message1 - message2)
+   */
+  it should "leak keys if two messages are signed" in {
+    // The order of the secp256k1 curve (and thus the modulus of the field elements)
+    val M = BigInt(
```
Christewart: nit: this can probably be moved to `CryptoParams` if it isn't in there already

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
```diff
@@ -0,0 +1,197 @@
...
+
+    forAll(CryptoGenerators.nonZeroPrivKey,
+           CryptoGenerators.schnorrNonce,
+           NumberGenerator.bytevector(32),
+           NumberGenerator.bytevector(32)) {
+      case (privKey, nonce, message1, message2) =>
+        // This will only work if we sign two different messages
+        assert(message1 != message2)
+
+        // Sign both messages using the same privKey and nonce
```
Christewart, line 162: the top level comment is wrong, it says it just needs to be the same private key. It needs to be the same key _and_ nonce, right?

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,46 @@
...
+  def fromBytes(bytes: ByteVector): SchnorrNonce = {
+    require(bytes.length <= 33,
+            s"Schnorr Nonce must be less than 33 bytes, got $bytes")
+
+    if (bytes.length == 32)
+      SchnorrNonce(bytes)
+    else if (bytes.length < 32) {
+      // means we need to pad the private key with 0 bytes so we have 32 bytes
+      SchnorrNonce.fromBytes(bytes.padLeft(32))
+    } else {
```
Christewart: I think it is probably better to be super explicit here and see 

```scala 
else if (bytes.length == 33) {
  //make recursive call
}
```

this protects ourselves from doing something stupid like removing the require in the future. In the else case we should still throw, even if it is redundant.

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrTest.scala
nkohen: "same keys", so not wrong technically but the nonce terminology seems to complicate things :/ I'll make it more clear

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
torkelrogstad: @nkohen thoughts on @rorp s alternative?

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,76 @@
...
+    val sigArr =
+      NativeSecp256k1.schnorrSign(msg.toArray, privateKey.bytes.toArray)
+
+    SchnorrDigitalSignature.fromBytes(ByteVector(sigArr))
+  }
+
+  /** Generates a Schnorr signature for the 32 byte msg using privateKey
+    * and nonce.
+    *
+    * IMPORTANT: Never sign two messages with the same privateKey and nonce!
```
torkelrogstad: For Scaladoc bonus points: doing `'''IMPORTANT'''` makes it bold I believe

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,50 @@
...
+      SchnorrNonce(bytes)
+    else if (bytes.length < 32) {
+      // means we need to pad the private key with 0 bytes so we have 32 bytes
+      SchnorrNonce.fromBytes(bytes.padLeft(32))
+    } else if (bytes.length == 33) {
+      // this is for the case when java serialies a BigInteger to 33 bytes to hold the signed num representation
+      SchnorrNonce.fromBytes(bytes.tail)
+    } else {
+      throw new IllegalArgumentException(
+        "Private keys cannot be greater than 33 bytes in size, got: " +
```
Christewart: Schnorr Nonce..

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
ysangkok: @nkohen why no thoughts on this in this debate?

PR #685, opened 2019-08-09
Change 'height' types from Long -> Int. Int.MaxValue is 2,147,483,647, w…

Christewart:
…hich is an order of magnitude more than current block heights on mainnet/testnet bitcoin


Christewart:
This also means we don't have to call `.toInt` to index into our [`Blockchain`](https://github.com/bitcoin-s/bitcoin-s/blob/020a9356180efa2e71ef02b0eadd677fee0bc83f/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala#L40) data structure

PR #688, opened 2019-08-10
2019 08 09 Don't use BlockHeaderDAO in TipValidation

Christewart:
This builds off of #655 

This uses our in memory blockchain to do tip validation instead of having to _possibly_ hit the database when validating a new tip. The downside here is is that we need to keep more chain in m

A side effect of this PR is it makes it much easier to benchmark our tip validation logic now that there is any async computation happening in this code path, which allows us to make progress on #653

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.chain.blockchain
 
 import org.bitcoins.chain.config.ChainAppConfig
-import org.bitcoins.chain.models.{BlockHeaderDAO, BlockHeaderDb}
+import org.bitcoins.chain.models.BlockHeaderDb
 import org.bitcoins.chain.validation.{TipUpdateResult, TipValidation}
 import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.db.ChainVerificationLogger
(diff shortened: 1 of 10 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -37,7 +36,22 @@ case class Blockchain(headers: Vector[BlockHeaderDb])
...
-  override def apply(idx: Int): BlockHeaderDb = headers(idx)
+  override def apply(idx: Int): BlockHeaderDb = headers.apply(idx)
+
+  /** Finds a block header at a given height */
+  def findAtHeight(height: Int): Option[BlockHeaderDb] =
+    find(_.height == height)
+
+  /** Splits the blockchain at the header, returning a new blockchain where the best tip is the given header */
+  def fromHeader(header: BlockHeaderDb): Option[Blockchain] = {
+    val headerIdxOpt = headers.zipWithIndex.find(_._1 == header)
```
nkohen, line 47: wouldn't it be more efficient to not `zipWithIndex` and instead to do something like `indexOf` after the fact (to avoid doing processing on the entire header chain)

Christewart: I'm not going to consider performance in this PR as we don't have proper benchmarking, I will investigate this in #689

nkohen (approved):
ACK

rorp (approved):

PR #683, opened 2019-08-08
WIP: Binary Outcome DLC

nkohen:
This PR builds on previous Schnorr work to construct a DLC with a binary outcome between two participants.

PR #694, opened 2019-08-15
WIP: Binary Outcome DLC

nkohen:
This PR builds on previous Schnorr work to construct a DLC with a binary outcome between two participants.

(Was previously #683)


core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import org.bitcoin.NativeSecp256k1
+import scodec.bits.ByteVector
+
+object Schnorr {
+
+  def sign(
+      msg: ByteVector,
+      privateKey: ECPrivateKey): SchnorrDigitalSignature = {
+    if (msg.length != 32) {
```
torkelrogstad: `require` instead? You're basically hand rolling it here

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+        NativeSecp256k1.schnorrSign(msg.toArray, privateKey.bytes.toArray)
+
+      SchnorrDigitalSignature.fromBytes(ByteVector(sigArr))
+    }
+  }
+
+  def signWithNonce(
+      msg: ByteVector,
+      privateKey: ECPrivateKey,
+      nonce: ECPrivateKey): SchnorrDigitalSignature = {
```
torkelrogstad: `nonce` here is a priv key because it needs to be 32 bytes, right? Not sure how I feel about using `ECPrivateKey`, as types are also very important to convey meaning to params etc

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+    if (msg.length != 32) {
```
torkelrogstad: Same thing with `require` from abvoe

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+
+      SchnorrDigitalSignature.fromBytes(ByteVector(sigArr))
+    }
+  }
+
+  def verify(
+      msg: ByteVector,
+      sig: SchnorrDigitalSignature,
+      publicKey: ECPublicKey): Boolean = {
+    if (msg.length != 32) {
```
torkelrogstad: `require`

core/src/main/scala/org/bitcoins/core/crypto/Schnorr.scala
```diff
@@ -0,0 +1,71 @@
...
+                                    sig.bytes.toArray,
+                                    publicKey.bytes.toArray)
+    }
+  }
+
+  def computePubKey(
+      msg: ByteVector,
+      r: ECPublicKey,
+      publicKey: ECPublicKey): ECPublicKey = {
+    if (msg.length != 32) {
```
torkelrogstad: `require`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.core.util.{BitcoinSLogger, BitcoinSUtil, Factory}
+import scodec.bits.ByteVector
+
+sealed abstract class SchnorrDigitalSignature extends BitcoinSLogger {
```
torkelrogstad: IMO there's no real advantage to using the sealed trait/factory object/impl case class pattern here, might as well define `SchnorrDigitialSignature` as a case class (which should lead to fewer allocations as well)

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+  override val rx: ByteVector = ByteVector.fill(32)(0.toByte)
+  override val s: ByteVector = rx
+}
+
+object SchnorrDigitalSignature extends Factory[SchnorrDigitalSignature] {
+  private case class SchnorrDigitalSignatureImpl(rx: ByteVector, s: ByteVector)
+      extends SchnorrDigitalSignature
+
+  override def fromBytes(bytes: ByteVector): SchnorrDigitalSignature = {
+    if (bytes.length != 64) {
```
torkelrogstad: `require`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+      throw new IllegalArgumentException(
+        s"SchnorrDigitalSignature must be 64 bytes, got $bytes")
+    } else {
+      val (rx, s) = bytes.splitAt(32)
+      SchnorrDigitalSignatureImpl(rx, s)
+    }
+  }
+
+  def apply(rx: ByteVector, s: ByteVector): SchnorrDigitalSignature = {
+    if (rx.length != 32) {
```
torkelrogstad: `require`. Also, this is covered by the `require` statements in the abstract class

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+      throw new IllegalArgumentException(s"rx must be 32 bytes, got $rx")
+    } else if (s.length != 32) {
+      throw new IllegalArgumentException(s"s must be 32 bytes, got $s")
+    } else {
+      SchnorrDigitalSignatureImpl(rx, s)
+    }
+  }
+
+  def apply(r: ECPublicKey, s: ByteVector): SchnorrDigitalSignature = {
+    if (s.length != 32) {
```
torkelrogstad: this length check is covered in the abstract class

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -255,6 +255,87 @@ object MultiSignatureScriptPubKey
...
+  }
+
+  def timeout: Int = timeoutSPK.locktime.toInt
+}
+
+object MultiSignatureWithTimeoutScriptPubKey
+    extends ScriptFactory[MultiSignatureWithTimeoutScriptPubKey] {
+  import org.bitcoins.core.script.control._
+
+  private case class MultiSignatureWithTimeoutScriptPubKeyImpl(
```
torkelrogstad, line 291: Same thing as above. Do we need the abstract class/case class indirection?

dlc/README.md
```diff
@@ -0,0 +1 @@
+#TODO
```
torkelrogstad, line 1: The other modules don't really have meaningful READMEs, FYI. Probably makes more sense to move this to the website docs, adding a section on DLCs

nkohen:
Replaced by #871

PR #697, opened 2019-08-19
Bump Scala versions

torkelrogstad:
~This PR depends on https://github.com/hseeberger/akka-http-json/pull/314.~

We add support for Scala 2.12.9 and 2.13.0.

To make this easier, we delete the `scripts` project. Everything
that was in here was covered by content on the website. We also
delete the `doc` folder, as that was a remnant from when `scripts`
was called `doc`. This allows us to remove Ammonite as a dependency.

Notable changes that had to be made for this work, is that `IndexedSeqLike` is removed
in Scala 2.13. See https://gitter.im/scala/scala/archives/2019/08/19 (search for "torkelrogstad" (also, how do I link to specific Gitter messages?)). I solved this by introducing three new folders in `chain/src/main`: `scala-2.1{1,2,3}`. Version specific code goes in here, and sbt/Bloop picks it up automatically. This leads to a bunch of code duplication, which is not ideal. But I'm not sure how to handle it without doing this. I tried splitting the companion object into a different, shared place, but then the compiler complained.


torkelrogstad:
Pushed a commit where I yanked the code we used from `akka-http-upickle` until the original author can publish an updated version of the lib. Package structure, copyright notice and license is kept intact. Did a quick reading of the Apache 2.0 and I think this is acceptable licensing wise. 

Also pushed a commit where I clean up as many compiler warnings as possible.

Christewart:
It does look like some new errors are introduced 

```
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala:270:50: method get in class RightProjection is deprecated (since 2.13.0): Use `Either.getOrElse` instead
[warn]     ByteVector.encodeAscii("Bitcoin seed").right.get
[warn]                                                  ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala:270:44: method right in class Either is deprecated (since 2.13.0): Either is now right-biased, use methods directly on Either
[warn]     ByteVector.encodeAscii("Bitcoin seed").right.get
[warn]                                            ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala:29:41: method get in class RightProjection is deprecated (since 2.13.0): Use `Either.getOrElse` instead
[warn]     ByteVector.encodeAscii(chars).right.get
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/protocol/ln/LnHumanReadablePart.scala:29:35: method right in class Either is deprecated (since 2.13.0): Either is now right-biased, use methods directly on Either
[warn]     ByteVector.encodeAscii(chars).right.get
[warn]                                   ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/protocol/ln/LnTaggedFields.scala:97:34: method right in class Either is deprecated (since 2.13.0): Either is now right-biased, use methods directly on Either
[warn]         (None, descriptionOrHash.right.toOption)
[warn]                                  ^
[warn] 7 warnings found

```

Christewart:
I get a bunch of these errors if I run `+test:compile` too 

```
[error] /home/chris/dev/bitcoin-s-core/eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala:208:20: File name too long
[error] This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
[error]           sentInfo <- client.getSentInfo(invoice.lnTags.paymentHash.hash)
[error]                    ^
[error] /home/chris/dev/bitcoin-s-core/eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala:479:23: File name too long
[error] This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
[error]             succeeded <- client.getSentInfo(invoice.lnTags.paymentHash.hash)
[error]                       ^
[error] /home/chris/dev/bitcoin-s-core/eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala:515:15: File name too long
[error] This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
[error]             _ <- EclairRpcTestUtil.awaitUntilPaymentSucceeded(client, paymentId)
[error]               ^
```

In general, why do we need to distinguish between `2.11`/`2.12` on our `Blockchain` data structure. The 2.12 `Blockchain` works on 2.11 as well (collections are the same), so I don't understand why they need to be duplicated there?

Christewart:
So looks at this closer 

>

Notable changes that had to be made for this work, is that IndexedSeqLike is removed
in Scala 2.13. See https://gitter.im/scala/scala/archives/2019/08/19 (search for "torkelrogstad" (also, how do I link to specific Gitter messages?)). I solved this by introducing three new folders in chain/src/main: scala-2.1{1,2,3}. Version specific code goes in here, and sbt/Bloop picks it up automatically. This leads to a bunch of code duplication, which is not ideal. But I'm not sure how to handle it without doing this. I tried splitting the companion object into a different, shared place, but then the compiler complained.

I don't think there is a hard requirement that our underlying collection type be `Vector`. If we removed this requirement, would it make this all a lot easier? I'm guessing we will have to make everything a `IndexedSeq` in auxiliary code that touches `Vector[BlockHeaderDb]`. Could we try a proof of concept with `IndexedSeq` and see what it looks like?

torkelrogstad:
Pushed some new code that takes a new approach with the `Either` story: Instead of using implicits to convert types we introduce `CompatEither`, `CompatRight` and `CompatLeft`, which is basically a partial reimplementation of `Either` (cribbed from 2.12 stdlib) that works across Scala versions. This gets rid of the `EitherOps` compiler warnings. 

> In general, why do we need to distinguish between 2.11/2.12 on our Blockchain data structure. The 2.12 Blockchain works on 2.11 as well (collections are the same), so I don't understand why they need to be duplicated there?

Because I don't know how to get 2.11 and 2.12 to compile with one set of sources and 2.13 to use a different one. The only way I found was to have distinct folders per version. It's probably possible with some sbt-fu, though :-)

> I don't think there is a hard requirement that our underlying collection type be Vector. If we removed this requirement, would it make this all a lot easier? I'm guessing we will have to make everything a IndexedSeq in auxiliary code that touches Vector[BlockHeaderDb]. Could we try a proof of concept with IndexedSeq and see what it looks like?

I still think we're going to need different code for different versions. Take a look at `IndexedSeq` in [2.11](https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/IndexedSeq.scala) and [2.13](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/IndexedSeq.scala). They extend different traits that need different methods quite a bit. 

Right now we have less than 100 lines of duplicated code for `Blockchain`. I don't think that's really a problem, and I doubt we'll be able to get much below that.

PR #698, opened 2019-08-20
Tweak module logging

### LINKED ISSUE #670: Put project specific loggers in their respective projects
Currently we have all of our loggers in `db-commons`. At a first glance, this seems like a good idea to consolidate our various loggers. Unfortunately this means that our loggers have to take in a generic [`AppConfig`](https://github.com/bitcoin-s/bitcoin-s/blob/master/db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala#L175) rather than a project specific config. 

In the case of the [`KeyHandlingLogger`](https://github.com/bitcoin-s/bitcoin-s/blob/master/db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala#L175) above, it should be taking in a `WalletAppConfig` rather than a `AppConfig`. This reduces implicit conflict resolution which was seen in #655 and talked about in the comment below

https://github.com/bitcoin-s/bitcoin-s/pull/655/files#r310064676

torkelrogstad:
Closes #670 

In this PR we split app loggers into their corresponding modules

This avoids a bunch of places with awkward implicit resolution.

We also make sure that all logging in node, chain and wallet
happens with the app loggers, and not BitcoinSLogger

We also tune the default test logging level for chain, wallet and node to WARN

@Christewart the reason you weren't getting the logs you expected was that `logger` referred to `BitcoinSLogger.logger`. I went through node, chain and wallet and removed all usage of `BitcoinSLogger`, so this should be fixed now.


testkit/src/main/scala/org/bitcoins/testkit/BitcoinSTestAppConfig.scala
```diff
@@ -10,10 +10,20 @@ object BitcoinSTestAppConfig {
...
     *
     * 1) Data directory is set to user temp directory
+    * 2) Logging is turned down to WARN
     */
   def getTestConfig(config: Config*): BitcoinSAppConfig = {
+    val overrideConf = ConfigFactory.parseString {
+      """
+      |bitcoin-s {
+      |  logging {
+      |     level = WARN
```
Christewart, line 20: So does testkit logging need to be adjusted here now?

testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
```diff
@@ -9,7 +9,11 @@ import org.scalatest._
 import scala.concurrent.{Future, Promise}
 import scala.util.{Failure, Success}
 
-trait BitcoinSFixture extends fixture.AsyncFlatSpec with BitcoinSLogger {
+trait BitcoinSFixture extends fixture.AsyncFlatSpec {
+
+  // to avoid this trickling up to things that extend
+  // this trait
+  private val logger = BitcoinSLogger.logger
```
Christewart, line 16: Why are we doing this? I thought the `BitcoinSLogger.logger` is broken?

testkit/src/main/scala/org/bitcoins/testkit/BitcoinSTestAppConfig.scala
torkelrogstad: When you say testkit logging, what do you mean? Test code, or code in the testkit project?

testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
torkelrogstad: It isn't broken, but it is turned off for `chain`, `wallet`, and `node`. So logs produced by `BitcoinSLogger.logger` won't appear when running tests in these modules. I kept this in because there might be a scenario where fixtures are used in a different module. One other solution is to introduce a new logger somewhere in `testkit` (`TestLogger`, or something along those lines).

Christewart: Is there a way to turn it on? Can we specify this in the reference.conf?

torkelrogstad: I guess that would also work. Could have it as an option we don't expose to the end user, but keep around internally?

Christewart: Yeah, I mean I think it is probably a good idea to keep a default logger around.

PR #713, opened 2019-08-22
Bump test coverage

torkelrogstad:
This gets us above 90% in core, at least for me locally


core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
```diff
@@ -109,49 +106,44 @@ object Sha256DigestBE extends Factory[Sha256DigestBE] {
 /**
   * Represents the result of SHA256(SHA256())
   */
-sealed trait DoubleSha256Digest extends Any with HashDigest {
-  def flip: DoubleSha256DigestBE = DoubleSha256DigestBE(bytes.reverse)
+case class DoubleSha256Digest(bytes: ByteVector) extends HashDigest {
```
Christewart, line 109: This causes runtime allocations that were optimized in #690

Christewart: You can't have `require()` in the case class body if you want to avoid allocating https://docs.scala-lang.org/overviews/core/value-classes.html

torkelrogstad: We could make `DoubleSha256Digest` a `AnyVal`, but that gives us a compile error in `BloomFilter`, saying that `DoubleSha256Digtest` and `ByteVector` are the same type after type erasure. What's your thoughts on removing `BloomFilter.insert(DoubleSha256Digest)`?

core-test/src/test/scala/org/bitcoins/core/crypto/AesCryptTest.scala
```diff
@@ -331,6 +331,17 @@ class AesCryptTest extends BitcoinSUnitTest {
     assertDoesNotCompile("""val k = new AesKey(hex"1234")""")
   }
 
+  it must "not be constructable from bad byte lenghts" in {
+    forAll(NumberGenerator.bytevector.suchThat(bytes =>
+      !AesKey.keylengths.contains(bytes.length))) { bytes =>
```
Christewart: This generator looks like it is going to be hard to satisfy, it is generally a bad idea to use `.suchThat()` rather than writing a generator that just directly generators the values you want

core-test/src/test/scala/org/bitcoins/core/crypto/ChainCodeTest.scala
```diff
@@ -0,0 +1,16 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.bitcoins.testkit.core.gen.NumberGenerator
+
+class ChainCodeTest extends BitcoinSUnitTest {
+  behavior of "ChainCode"
+
+  it must "not be constructable from invalid lengths byte vectors" in {
+    forAll(NumberGenerator.bytevector.suchThat(_.length != 32)) { bytes =>
```
Christewart, line 10: See, this usage of `.suchThat()` is acceptable as it is _not_ very likely a 32 byte vector is going to be generated. 

Now if we _inverted_ the predicate (required _only_ byte vectors of length 32!) this would be hard to satisfy and we would get a lot of scalacheck errors

core/src/main/scala/org/bitcoins/core/crypto/HashDigest.scala
Christewart: So in general, this stuff needs to be addressed in #717. I think it is probably out of scope of this PR unless it really helps the test coverage story too?

torkelrogstad: Might help a bit with coverage, but not much. Agree on out of scope.

PR #714, opened 2019-08-22
Add forced stdout prints on CI to avoid timeout

torkelrogstad:
I have a suspicion our tests are failing because Travis kills them on macOS after getting no output for a while. This should fix that.


torkelrogstad:
Closing this, as this doesn't appear to be the cause of the problem.

PR #725, opened 2019-08-28
Simplify RPC clients API

torkelrogstad:
In this PR we make it possible to construct RPC clients without specifying an actor system. If the user chooses not to do so, we create one for them. It is still possible to construct clients with a system.

The documentation on the website is updated to use the simplified API.


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -44,14 +44,36 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
...
 object BitcoindRpcClient {
 
+  /**
+    * Creates an RPC client from the given instance.
+    *
+    * Behind the scenes, we create an actor system for
+    * you. You can use `withActorSystem` if you want to
+    * manually specify an actor system for the RPC client.
+    */
+  def apply(instance: BitcoindInstance): BitcoindRpcClient = {
```
Christewart, line 58: Why not this? 

```scala
  def apply(instance: BitcoindInstance)(
      implicit system: ActorSystem = ActorSystem.create("bitcoind-rpc-client")): BitcoindRpcClient = {
    withActorSystem(instance)
  }
```

This allows you to implicitly bind a system if you have one in scope, else create one. Seems like the best of both worlds? Another nit: I would add a timestamp to the system creation

torkelrogstad: I think that signature leads to a bad API for the end user, as you have no control over what happens to the actor system. Our goal should be a clear and intuitive API, not as few lines as possible. 

There's two scenarios here: 

1) The user doesn't care/has no idea about actor systems. They shouldn't see one in the method signature, as that increases the complexity and gives them no real value.
2) The user cares about what actor system he or she uses. In that case they should get a compiler failure if there's no system in scope. That won't happen with the signature above

Christewart: So this means in all of our test cases we need to retroactively change applies to `withActorSystem` so we can properly shut down actor systems. Also this means when you call `.stop()` on the rpc client there probably should be a call to stop the actor system itself so it isn't lingering after we are done using the rpc client

torkelrogstad: Done

PR #737, opened 2019-09-05
Add receiver param to BitcoindRpcTestUtil.fundBlockChainTransaction()

Christewart:
We have a helper method called `fundBlockChainTransaction()` that is used in our test suite.

In this method, we fund a tx, and then generate a block to confirm that tx. There is an implicit requirement here that both peers have seen the block. What this PR does is add a `receiver` parameter to this method and then makes sure the `sender` and `receiver` have seen the same block.


rorp (approved):
👍

PR #766, opened 2019-09-27
Make tests to not require pre-installed bitcoind

rorp:


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/BlockchainRpcTest.scala
```diff
@@ -14,13 +14,14 @@ import org.bitcoins.core.config.RegTest
 class BlockchainRpcTest extends BitcoindRpcTest {
 
   lazy val clientsF: Future[(BitcoindRpcClient, BitcoindRpcClient)] =
-    BitcoindRpcTestUtil.createNodePair(clientAccum = clientAccum)
+    BitcoindRpcTestUtil.createNodePairV17(clientAccum = clientAccum)
```
cwaldron97, line 17: Do we just make V17 the default now instead of an unknown versioned NodePair?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -51,6 +50,9 @@ sealed trait BitcoindInstance extends BitcoinSLogger {
         .last
 
     foundVersion match {
+      case _: String
```
cwaldron97, line 53: Nit: Convention in case is slightly different between each case.

Christewart (approved):

Christewart:
needs rebase

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/BlockchainRpcTest.scala
rorp: This test doesn't work with other versions.

PR #770, opened 2019-10-01
Remove bitcoind dependency from node

rorp:
The `bitcoins` node cannot start if `bitcoind` is not installed locally.

### DIFF
--- a/app/server/src/main/resources/reference.conf
+++ b/app/server/src/main/resources/reference.conf
@@ -1,3 +1,62 @@
+bitcoin-s {
+    datadir = ${HOME}/.bitcoin-s
+    network = regtest # regtest, testnet3, mainnet
+
+    logging {
+        level = WARN # trace, debug, info, warn, error, off
+
+        # You can also tune specific module loggers.
+        # They each take the same levels as above.
+        # If they are commented out (as they are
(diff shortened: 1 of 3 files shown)


app/server/src/main/resources/reference.conf
```diff
@@ -1,3 +1,62 @@
...
+
+        # whether or not to write to the log file
+        disable-file = false
+
+        # whether or not to log to stdout
+        disable-console = false
+    }
+
+    node {
+        peers = ["localhost"]
```
Christewart: can we you put examples here for things you can specify, for instance i would expect `testnet.neutrino.suredbits.com:18333` to work as well

app/server/src/main/resources/reference.conf
```diff
@@ -1,3 +1,65 @@
...
+        # whether or not to write to the log file
+        disable-file = false
+
+        # whether or not to log to stdout
+        disable-console = false
+    }
+
+    node {
+        peers = ["localhost"] # a list of peer addresses in form "hostname:portnumber"
+                              # (e.g. "neutrino.testnet3.suredbits.com:18333")
```
Christewart, line 42: :tada:

PR #776, opened 2019-10-02
ScriptSignature Impl Refactor

nkohen:
Got rid of Impl pattern and IntelliJ warnings in `ScriptSignature.scala`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
@@ -5,6 +5,7 @@ import org.bitcoins.core.script.constant._
 import org.bitcoins.core.serializers.script.ScriptParser
 import org.bitcoins.core.util._
 
+import scala.annotation.tailrec
 import scala.util.{Failure, Success, Try}
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -24,23 +25,19 @@ sealed abstract class ScriptSignature extends Script {
 
 }
 
-sealed trait NonStandardScriptSignature extends ScriptSignature {
+case class NonStandardScriptSignature(override val asm: Vector[ScriptToken])
+    extends ScriptSignature {
   def signatures: Seq[ECDigitalSignature] = Nil
-  override def toString = "NonStandardScriptSignature(" + hex + ")"
+  override def toString: String = "NonStandardScriptSignature(" + hex + ")"
```
Christewart, line 31: `s"NonStandardScriptSignature($hex)"` while we are at it

Christewart (commented):
So this current PR does not protect the invariants, as you can use the case class constructor directly. You need to add a `require(isP2PKHScriptSigh(asm), "error message...")` in the case class bodies. Otherwise you can do this

```scala
scala> import org.bitcoins.core.protocol.script._
                                                ^
       warning: Unused import
import org.bitcoins.core.protocol.script._

scala> import scodec.bits._

       import org.bitcoins.core.protocol.script._
                                                

scala> val p2pkh = P2PKHScriptSignature(Vector.empty)

       import scodec.bits._
                          ^
<synthetic>:6: warning: Unused import

(To diagnose errors in synthetic code, try adding `// show` to the end of your input.)
p2pkh: org.bitcoins.core.protocol.script.P2PKHScriptSignature = P2PKHScriptSignature(00)

scala> p2pkh

       import org.bitcoins.core.protocol.script._
                                                ^
<synthetic>:3: warning: Unused import

(To diagnose errors in synthetic code, try adding `// show` to the end of your input.)

       import scodec.bits._
                          ^
<synthetic>:6: warning: Unused import

(To diagnose errors in synthetic code, try adding `// show` to the end of your input.)
res1: org.bitcoins.core.protocol.script.P2PKHScriptSignature = P2PKHScriptSignature(00)

```

EDIT: 

You might be able to the invariant into `Script` type itself. Not sure, but that would probably be the best.

Christewart:
Also it is a little frightening i don't have test cases for this, maybe add a few sanity ones ? I don't think you need to do all of the types.

PR #781, opened 2019-10-03
Optimize org.bitcoins.chain.blockchain.BaseBlockChain

rorp:
The current implementation contains a few bottle necks that slow down full sync time to several hours. This PR speeds up full block headers sync to several minutes.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
@@ -58,11 +58,10 @@ private[blockchain] trait BaseBlockChain {
   /** Splits the blockchain at the header, returning a new blockchain where the best tip is the given header */
   def fromHeader(header: BlockHeaderDb): Option[Blockchain] = {
     val headerIdxOpt = findHeaderIdx(header.hashBE)
-    headerIdxOpt.map {
-      case (header, idx) =>
-        val newChain = this.compObjectfromHeaders(headers.splitAt(idx)._2)
-        require(newChain.tip == header)
-        newChain
+    headerIdxOpt.map { idx =>
+      val newChain = this.compObjectfromHeaders(headers.splitAt(idx)._2)
+      require(newChain.tip == header)
+      newChain
     }
   }
 
@@ -71,18 +70,17 @@ private[blockchain] trait BaseBlockChain {
     fromHeader(header).get
   }
 
-  /** Finds a header by the given hash and returns it along with its index in the headers vector */
-  def findHeaderIdx(
-      hashBE: DoubleSha256DigestBE): Option[(BlockHeaderDb, Int)] = {
+  /** Finds a header index by the given hash */
+  def findHeaderIdx(hashBE: DoubleSha256DigestBE): Option[Int] = {
 
     @tailrec
-    def loop(idx: Int): Option[(BlockHeaderDb, Int)] = {
+    def loop(idx: Int): Option[Int] = {
       if (idx >= headers.size)
         None
       else {
         val header = headers(idx)
         if (header.hashBE == hashBE)
-          Some(header, idx)
+          Some(idx)
         else
           loop(idx + 1)
       }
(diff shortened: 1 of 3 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
```diff
@@ -211,4 +198,25 @@ private[blockchain] trait BaseBlockChainCompObject
...
   }
 
+  /**
+    * Finds the parent of the given header
+    */
+  private def findPrevBlockHeaderIdx(
+      header: BlockHeader,
+      blockchain: Blockchain): Option[(BlockHeaderDb, Int)] = {
+    // Let's see if we are lucky and the latest tip is the parent.
+    val prevTipOpt = blockchain.headers.headOption
```
Christewart: I would prefer to use `blockchain.tip` here, in case we ever decide to organize headers a different way in `blockchain`.

chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
```diff
@@ -211,4 +198,25 @@ private[blockchain] trait BaseBlockChainCompObject
...
+    if (prevTipOpt.map(_.hashBE).contains(header.previousBlockHashBE)) {
+      // Yes we are. Returning the latest tip.
+      prevTipOpt.map(b => (b, 0))
+    } else {
+      // No. Scanning the blockchain to find the parent.
+      // We don't want to copy the whole headers vector in zipWithIndex, so we warp it into a Stream
+      blockchain.headers.toStream.zipWithIndex.find {
```
Christewart: `.toStream()` seems to be deprecated 

```scala
  @deprecated("Use .to(LazyList) instead of .toStream", "2.13.0")
  @`inline` final def toStream: immutable.Stream[A] = to(immutable.Stream)
```

chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
```diff
@@ -55,7 +55,8 @@ private[blockchain] trait BaseBlockChain {
 
   /** Splits the blockchain at the header, returning a new blockchain where the best tip is the given header */
   def fromHeader(header: BlockHeaderDb): Option[Blockchain] = {
-    val headerIdxOpt = headers.zipWithIndex.find(_._1 == header)
+    // we don't want to create another copy of headers vector in zipWithIndex, so we wrap it in a Stream
+    val headerIdxOpt = headers.toStream.zipWithIndex.find(_._1 == header)
```
Christewart: `.toStream` seems to be deprecated

```scala
  @deprecated("Use .to(LazyList) instead of .toStream", "2.13.0")
  @`inline` final def toStream: immutable.Stream[A] = to(immutable.Stream)
```

Christewart (commented):
So I don't really understand why this is causing copies. Since we are dealing with immutable data structures, when we create a new `Blockchain` would that re-use the `BlockHeaderDb` instances that are already allocated in memory ? It seems like we would just allocate a another Vector instance, but not actually do a deep copy for all of the `BlockHeaderDb`? 

What am I missing?

chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
rorp: Should we deprecate support for Scala 2.11 and 2.12 then? There's no LazyList in their standard libraries.

chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
rorp: Should we deprecate support for Scala 2.11 and 2.12 then? There's no LazyList in their standard libraries.

rorp:
I don't really understand why this is causing copies either. The fact is that Scala does create new copies. I would remove `IndexedSeq` support for `Blockchain` to eliminate such performance issues in the future.

chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala
Christewart: Take a look at the top of this file. We have separate file structures for `2.11` `2.12` and `2.13` because of this. We should probably move things that use `.toStream()` into the corresponding scala version

Issue #523, opened 2019-06-14
Efficient GolombFilter.matchesHash


nkohen:
Add an alternative implementation to `GolombFilter.matchesHash` that lazily computes only the necessary hashes like the algorithm specified in https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#set-queryingdecompression

PR #786, opened 2019-10-07
Make ChainApi to scan the block filters in order to find matches

rorp:
Rudimentary support for block filters

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -365,8 +365,6 @@ class ChainHandlerTest extends ChainUnitTest {
         val blockHashBE =
           DoubleSha256DigestBE.fromBytes(ECPrivateKey.freshPrivateKey.bytes)
         val golombFilter = BlockFilter.fromHex("017fa880", blockHashBE.flip)
-        val firstFilter = CompactFilterMessage(blockHash = blockHashBE.flip,
-                                               filter = golombFilter)
         val firstFilterHeader = FilterHeader(filterHash = golombFilter.hash,
                                              prevHeaderHash =
                                                DoubleSha256Digest.empty)
@@ -459,8 +457,8 @@ class ChainHandlerTest extends ChainUnitTest {
       matched <- chainHandler.getMatchingBlocks(
         addresses =
           Vector(BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get),
-        start = None,
-        end = None)
+        startOpt = None,
+        endOpt = None)
     } yield {
       assert(Vector(created.blockHashBE) == matched)
     }
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -415,7 +415,8 @@ class ChainHandlerTest extends ChainUnitTest {
   it must "match block filters" in { chainHandler: ChainHandler =>
     import scodec.bits._
 
-    val bytes: ByteVector =
+    // This is a filter for a random block on testnet
+    val filterBytes: ByteVector =
       hex"fd2701f0ed169ad16107a8a74609b9e4de3c6133c564f79923ca228805d3" ++
         hex"8e3efc796c4b35034cb573b10b759cdda5efd19e1cdb4d343afcb06455fa" ++
         hex"820b06eca828ad61d3377fa464f3bd06ff4432310a363f667e13d09ba993" ++
@@ -444,10 +445,11 @@ class ChainHandlerTest extends ChainUnitTest {
         hex"81502f0883d52c6a3bcc956e0ea1787f0717d0205fecfe55b01edb1ac0"
 
     val compactFilterDb = CompactFilterDb(
-      hashBE = CryptoUtil.doubleSHA256(bytes).flip,
+      hashBE = CryptoUtil.doubleSHA256(filterBytes).flip,
       filterType = FilterType.Basic,
-      bytes = bytes,
+      bytes = filterBytes,
       height = 1,
+      // this is the hash of the random testnet block
       blockHashBE = DoubleSha256DigestBE
         .fromHex(
           "00000000496dcc754fabd97f3e2df0a7337eab417d75537fecf97a7ebb0e7c75")
@@ -455,10 +457,11 @@ class ChainHandlerTest extends ChainUnitTest {
     for {
       created <- chainHandler.filterDAO.create(compactFilterDb)
       matched <- chainHandler.getMatchingBlocks(
-        addresses =
+        addresses = // this is a random address which is included into the block
           Vector(BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get),
         startOpt = None,
-        endOpt = None)
+        endOpt = None
+      )
     } yield {
       assert(Vector(created.blockHashBE) == matched)
     }
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -456,13 +456,13 @@ class ChainHandlerTest extends ChainUnitTest {
     )
     for {
       created <- chainHandler.filterDAO.create(compactFilterDb)
-      matched <- chainHandler.getMatchingFilters(
+      matched <- chainHandler.getMatchingBlocks(
         addresses =
           Vector(BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get),
         start = None,
         end = None)
     } yield {
-      assert(Vector(created) == matched)
+      assert(Vector(created.blockHashBE) == matched)
     }
   }
 
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -457,8 +457,9 @@ class ChainHandlerTest extends ChainUnitTest {
     for {
       created <- chainHandler.filterDAO.create(compactFilterDb)
       matched <- chainHandler.getMatchingBlocks(
-        addresses = // this is a random address which is included into the block
-          Vector(BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get),
+        scripts = Vector(
+          // this is a random address which is included into the block
+          BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get.scriptPubKey),
         startOpt = None,
         endOpt = None
       )
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -3,7 +3,11 @@ package org.bitcoins.chain.blockchain
 import akka.actor.ActorSystem
 import org.bitcoins.chain.api.ChainApi
 import org.bitcoins.chain.config.ChainAppConfig
-import org.bitcoins.chain.models.{BlockHeaderDb, BlockHeaderDbHelper}
+import org.bitcoins.chain.models.{
+  BlockHeaderDb,
+  BlockHeaderDbHelper,
+  CompactFilterDb
+}
 import org.bitcoins.core.crypto.{
   DoubleSha256Digest,
   DoubleSha256DigestBE,
@@ -11,7 +15,9 @@ import org.bitcoins.core.crypto.{
 }
 import org.bitcoins.core.gcs.{BlockFilter, FilterHeader, FilterType}
 import org.bitcoins.core.p2p.CompactFilterMessage
+import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.blockchain.BlockHeader
+import org.bitcoins.core.util.CryptoUtil
 import org.bitcoins.testkit.BitcoinSTestAppConfig
 import org.bitcoins.testkit.chain.fixture.ChainFixtureTag
 import org.bitcoins.testkit.chain.{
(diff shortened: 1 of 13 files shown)


chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -164,4 +170,56 @@ trait ChainApi {
...
+
+    def height(blockStamp: BlockStamp): Int = blockStamp match {
+      case blockHeight: BlockHeight => blockHeight.height
+      case header: HeaderBlockStamp => header.height
+      case x: BlockStamp =>
+        throw new RuntimeException(s"Invalid block stamp: $x")
+    }
+  }
+
+  /** This trait represents a point on blockchain, and is used to specify block ranges */
```
Christewart: Why not just use a `BlockHeaderDb` ?

Christewart: If we are going to use this `BlockStamp` idea, let's move it into it's own file.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
   override def getFiltersAtHeight(height: Int)(
       implicit ec: ExecutionContext): Future[Vector[CompactFilterDb]] =
     filterDAO.getAtHeight(height)
+
+  /** @inheritdoc */
+  def getMatchingBlocks(
+      addresses: Vector[BitcoinAddress],
+      start: Option[BlockStamp] = None,
```
Christewart: nit: `startOpt` and `endOpt` as names.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+      s"Starting looking for matching blocks for addresses ${addresses.mkString(",")}")
+
+    val batchSize = chainConfig.filterBatchSize
+
+    val bytes = addresses.map(_.scriptPubKey.asmBytes)
+
+    val startHeightF =
+      Future.successful(start.map(BlockStamp.height).getOrElse(0))
+    val endHeightF = for {
+      filterCount <- getFilterCount
```
Christewart: probably want to check if `end` isDefined before making this database call

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+            s"Cannot find matching blocks for addresses ${addresses.mkString(",")}",
+            e)
+        }
+        acc
+      } else {
+        val newAcc: Future[Vector[DoubleSha256DigestBE]] = for {
+          compactFilterDbs <- filterDAO.getBetweenHeights(i - (batchSize - 1),
+                                                          i)
+          filtered <- Future.sequence(compactFilterDbs.map { filter =>
+            Future {
```
Christewart, line 407: Why do we need to wrap in another `Future` here?

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+            .getOrElse("")}")
+        }
+        acc.failed.foreach { e =>
+          logger.error(
+            s"Cannot find matching blocks for addresses ${addresses.mkString(",")}",
+            e)
+        }
+        acc
+      } else {
+        val newAcc: Future[Vector[DoubleSha256DigestBE]] = for {
```
Christewart:
```scala
        val startHeight = i - (batchSize - 1)
        val endHeight = i
```

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+                  Some(filter.blockHashBE)
+                else None
+              }
+            }
+          })
+          res <- acc
+        } yield {
+          res ++ filtered.flatten
+        }
+        val nextI = Math.max(start, i - batchSize)
```
Christewart:
```scala
val nextI = Math.max(start, endHeight - batchSize)
```

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -84,6 +84,16 @@ case class GolombFilter(
 
     matchesHash(hash)
   }
+
+  def matchesAnyHash(hashes: Vector[UInt64]): Boolean =
```
Christewart, line 86: scaladocs

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -164,4 +170,56 @@ trait ChainApi {
...
+
+  object BlockStamp {
+    implicit def implicitHeight(height: Int): BlockHeight = BlockHeight(height)
+    implicit def implicitHeader(header: BlockHeaderDb): HeaderBlockStamp =
+      HeaderBlockStamp(header)
+
+    def height(blockStamp: BlockStamp): Int = blockStamp match {
+      case blockHeight: BlockHeight => blockHeight.height
+      case header: HeaderBlockStamp => header.height
+      case x: BlockStamp =>
```
nkohen: I don't think this case is needed since `BlockStamp` is `sealed`

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -164,4 +170,56 @@ trait ChainApi {
...
+      end: Option[BlockStamp] = None)(
+      implicit ec: ExecutionContext): Future[Vector[DoubleSha256DigestBE]]
+
+}
+
+object ChainApi {
+
+  object BlockStamp {
+    implicit def implicitHeight(height: Int): BlockHeight = BlockHeight(height)
+    implicit def implicitHeader(header: BlockHeaderDb): HeaderBlockStamp =
```
nkohen: scaladocs on these two

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
nkohen, line 86: Could you leave a TODO and open an issue to refactor this method to do as is specified here https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#golomb-coded-set-multi-match

Probably link to https://github.com/bitcoin-s/bitcoin-s/issues/523 in that new issue since they are related

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+    val bytes = addresses.map(_.scriptPubKey.asmBytes)
+
+    val startHeightF =
+      Future.successful(start.map(BlockStamp.height).getOrElse(0))
+    val endHeightF = for {
+      filterCount <- getFilterCount
+    } yield end.map(BlockStamp.height).getOrElse(filterCount)
+
+    @tailrec
+    def scanFilters(
```
nkohen: scaladoc

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
   override def getFiltersAtHeight(height: Int)(
       implicit ec: ExecutionContext): Future[Vector[CompactFilterDb]] =
     filterDAO.getAtHeight(height)
+
+  /** @inheritdoc */
+  def getMatchingBlocks(
```
nkohen, line 372: scaladoc

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -164,4 +170,56 @@ trait ChainApi {
...
+
+  case class BlockHeight(height: Int) extends BlockStamp {
+    override def compare(that: BlockStamp): Int = that match {
+      case blockHeight: BlockHeight => height.compareTo(blockHeight.height)
+      case header: HeaderBlockStamp => height.compareTo(header.height)
+      case _: BlockStamp            => 1
+    }
+  }
+
+  case class HeaderBlockStamp(header: BlockHeaderDb) extends BlockStamp {
```
nkohen: scaladoc

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
nkohen: Also I like having `BlockHeight` but I don't think we need an ADT here, just add an `apply` or `fromBlockHeaderDb` method that pulls the height member. This should just wrap an `Int`

nkohen: And I think `BlockHeight` could be put somewhere in `core`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +360,77 @@ case class ChainHandler(
...
+
+    logger.info(
+      s"Starting looking for matching blocks for addresses ${addresses.mkString(",")}")
+
+    val batchSize = chainConfig.filterBatchSize
+
+    val bytes = addresses.map(_.scriptPubKey.asmBytes)
+
+    val startHeightF =
+      Future.successful(start.map(BlockStamp.height).getOrElse(0))
```
nkohen: Future not needed

chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
rorp: I wanted decouple `wallet` and `chain`, so used a separate type. I moved `BlockStamp` to `core`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
rorp: To let the execution context to run `filter.golombFilter.matchesAny(bytes)` in separate threads.

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -408,6 +412,58 @@ class ChainHandlerTest extends ChainUnitTest {
       }
   }
 
+  it must "match block filters" in { chainHandler: ChainHandler =>
+    import scodec.bits._
+
+    val bytes: ByteVector =
```
Christewart: Can you add a comment to where these bytes are coming from? My understanding is it's from the first testnet block? Let's document this so if we need to come back later and modify we know where it came from.

core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
```diff
@@ -0,0 +1,12 @@
+package org.bitcoins.core.protocol
+
+import org.bitcoins.core.crypto.DoubleSha256Digest
+import org.bitcoins.core.number.UInt32
+
+object BlockStamp {}
```
Christewart:
```scala
/** This trait represents a point on blockchain, and is used to specify block ranges */
sealed trait BlockStamp

object BlockStamp {
  case class BlockHash(hash: DoubleSha256Digest) extends BlockStamp
  case class BlockHeight(height: Int) extends BlockStamp
  case class BlockTime(time: UInt32) extends BlockStamp
}
```

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -408,6 +412,58 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      height = 1,
+      blockHashBE = DoubleSha256DigestBE
+        .fromHex(
+          "00000000496dcc754fabd97f3e2df0a7337eab417d75537fecf97a7ebb0e7c75")
+    )
+    for {
+      created <- chainHandler.filterDAO.create(compactFilterDb)
+      matched <- chainHandler.getMatchingBlocks(
+        addresses =
+          Vector(BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").get),
```
Christewart: i'm guessing this is the coinbase address in the testnet block? just add a quick specifying that if that is the case.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +359,90 @@ case class ChainHandler(
...
+      Vector[DoubleSha256DigestBE]] = {
+      if (end <= start) {
+        acc
+      } else {
+        val startHeight = end - (batchSize - 1)
+        val endHeight = end
+        val newAcc: Future[Vector[DoubleSha256DigestBE]] = for {
+          compactFilterDbs <- filterDAO.getBetweenHeights(startHeight,
+                                                          endHeight)
+          groupSize = if (compactFilterDbs.size / parallelismLevel * parallelismLevel < compactFilterDbs.size)
```
Christewart: so i implemented two helper functions here so that we can do this:

```scala
          groupSize = calcGroupSize(compactFilterDbs,parallelismLevel)
          grouped = compactFilterDbs.grouped(groupSize)
          filtered <- findMatches(grouped,scriptBytes)
```

and here are the helper functions:

```scala
  //scaladoc...
  private def calcGroupSize(compactFilterDbs: Vector[CompactFilterDb], parallelismLevel: Int): Int = {
    if (compactFilterDbs.size / parallelismLevel * parallelismLevel < compactFilterDbs.size)
      compactFilterDbs.size / parallelismLevel + 1
    else compactFilterDbs.size / parallelismLevel
  }

  //scaladoc...
  private def findMatches(grouped: Iterator[Vector[CompactFilterDb]],scriptBytes: Vector[ByteVector])(implicit ec: ExecutionContext): Future[Vector[DoubleSha256DigestBE]] = {
    val allMatchesF: Future[Iterator[Vector[DoubleSha256DigestBE]]] = {
      val nestedMatches: Iterator[Future[Vector[DoubleSha256DigestBE]]] = grouped.map { filterGroup =>
        //matchings for the subgroup
        val nestedMatch: Future[Vector[DoubleSha256DigestBE]] = {
          //we need to wrap in a future here to make sure we can
          //run these matches in parallel
          Future {
            filterGroup.foldLeft(Vector.empty[DoubleSha256DigestBE]) {
              (blocks, filter) =>
                if (filter.golombFilter.matchesAny(scriptBytes))
                  blocks :+ filter.blockHashBE
                else blocks
            }
          }
        }

        nestedMatch
      }
      Future.sequence(nestedMatches)
    }
    allMatchesF.map(_.flatten.toVector)
  }
```

feel free to remove type annotations inside of the helper methods if you want as it does clutter things up.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -357,6 +359,102 @@ case class ChainHandler(
...
+    /** Calculates group size to split a filter vector into [[parallelismLevel]] groups */
+    def calcGroupSize(vectorSize: Int): Int =
+      if (vectorSize / parallelismLevel * parallelismLevel < vectorSize)
+        vectorSize / parallelismLevel + 1
+      else vectorSize / parallelismLevel
+
+    /** Iterates over the grouped vector of filters to find matches with the given [[bytes]].
+      * Wraps each group into a [[Future]] so that it can be potentially run in a separate thread.
+      */
+    def findMatches(groupedVector: Iterator[Vector[CompactFilterDb]]) = {
```
Christewart: type annotation on the return type please

PR #771, opened 2019-10-01
2019 09 25 script interpreter cleanup

nkohen:
In order of commits, this PR:

1) Removes all IntelliJ warnings and some very minor cleanup in `ScriptInterpreter.scala`
2) Refactors `ScriptInterpreter.run` into nicer pieces
3) Replaces all calls to `isInstanceOf` in `ScriptInterpreter.scala`
4) Does a major refactor to the `ScriptProgram` ADT so that most operations can no longer be performed on `PreExecutionScriptProgram`s or `ExecutedScriptProgram`s and can only be performed on `ExecutionInProgressScriptProgram`s. It also tightens up the input and output types of functions so that not everything is just `ScriptProgram => ScriptProgram` (this removed a lot of `throw`s!). It replaces methods that are `ScriptProgram => ScriptProgram` that match on the three types and return the same type as the input into multiple methods with nicer types. Lastly, it refactors `ScriptInterpreter.loop` to be much nicer and only take in `ExecutionInProgressScriptProgram`s.
5) Fixes the problem I now had with `ScriptInterpreter.loop` where it cannot recognize that it is `tailrec` by moving all calls to `loop` and `loopOrComplete` to the bottom of the method.
6) Fix all tests that are now broken by simply turning all `PreExecutionScriptProgram`s into `ExecutionInProgressScriptProgram`s.

Commits 1-3 are minor refactors all contained in the `ScriptInterpreter` file. I'm open to making this its own smaller PR.

PR #791, opened 2019-10-08
2019 10 01 script program cleanup

nkohen:
In order of commits, this PR:

1) Does a major refactor to the ScriptProgram ADT so that most operations can no longer be performed on PreExecutionScriptPrograms or ExecutedScriptPrograms and can only be performed on ExecutionInProgressScriptPrograms. It also tightens up the input and output types of functions so that not everything is just ScriptProgram => ScriptProgram (this removed a lot of throws!). It replaces methods that are ScriptProgram => ScriptProgram that match on the three types and return the same type as the input into multiple methods with nicer types. Lastly, it refactors ScriptInterpreter.loop to be much nicer and only take in ExecutionInProgressScriptPrograms.
2) Fixes the problem I now had with ScriptInterpreter.loop where it cannot recognize that it is tailrec by moving all calls to loop and loopOrComplete to the bottom of the method.
3) Fix all tests that are now broken by simply turning all PreExecutionScriptPrograms into ExecutionInProgressScriptPrograms.

Formerly the last 3 commits of #771

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
@@ -13,8 +13,6 @@ import org.bitcoins.core.script.{ScriptProgram, _}
 import org.bitcoins.core.util.{BitcoinSLogger, BitcoinScriptUtil, CryptoUtil}
 import scodec.bits.ByteVector
 
-import scala.annotation.tailrec
-
 /**
   * Created by chris on 1/6/16.
   */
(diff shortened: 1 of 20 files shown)


core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
```diff
@@ -160,130 +148,114 @@ sealed abstract class CryptoInterpreter {
     * were placed in the scriptPubKey or redeemScript. If all signatures are valid, 1 is returned, 0 otherwise.
     * Due to a bug, one extra unused value is removed from the stack.
     */
-  @tailrec
-  final def opCheckMultiSig(program: ScriptProgram): ScriptProgram = {
+  final def opCheckMultiSig(
```
Christewart, line 151: removal of `@tailrec`?

core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
```diff
@@ -422,397 +420,448 @@ sealed abstract class ScriptInterpreter extends BitcoinSLogger {
...
+  }
+
   /**
     * The execution loop for a script
     *
     * @param program the program whose script needs to be evaluated
     * @return program the final state of the program after being evaluated by the interpreter
     */
-  @tailrec
+  //@tailrec
```
Christewart: ping

core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
```diff
@@ -422,397 +420,448 @@ sealed abstract class ScriptInterpreter extends BitcoinSLogger {
     }
   }
 
+  private def executeProgram(
```
Christewart, line 424: can you add a scaladoc about what the difference is between `executeProgram()` and `completeProgramExecution()`?

core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
Christewart: if this can't be re-added, i think this PR is a regression in terms of performance. We should definitely take out stuff that forces this not to be tail recursive.

core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
nkohen: No longer recursive ... so yeah

core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
nkohen: This is outdated, it is already back

Issue #760, opened 2019-09-25
Replace `ScriptProgram` companion object with pertinent methods that have names


nkohen:
At the time of this writing, the `ScriptProgram` companion object has 7 `apply` methods and 3 other methods, all 10 of which are state changes on existing `ScriptProgram`s.

https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala#L73

As such, all of these methods should be replaced by giving the actual subtypes of `ScriptProgram` pertinent methods with nice descriptive names! When all subtypes of `ScriptProgram` have a given method, then this method should also be added to the supertype.

For example, the first `apply` method takes in a `ScriptProgram` and a `ScriptError` and returns an `ExecutedScriptProgram` with the given error. This method should be replaced by placing a `failExecution` method in the `ScriptProgram` type and an implementation in each of the subtypes (likely by using `copy` methods given to case classes for free).

https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala#L82-L91

I recommend that this be split up into many smaller PRs (possibly built on top of each other) rather than one monolithic PR so that review can be fast and git history is easier to follow.

nkohen:
- [ ] Error Apply
- [ ] Flags Apply
- [ ] Update Apply
- [ ] Token Update Apply
- [ ] Last Code Separator Apply
- [ ] Token and Last Code Separator Apply
- [ ] Token Update With Alt Stack Apply
- [ ] Execution Completion Apply
- [ ] toExecutionInProgress

PR #794, opened 2019-10-11
Script Program apply method refactor part 2 (#760)

benthecarman:
This method was not used anywhere.


core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -61,7 +61,11 @@ case class PreExecutionScriptProgram(
     originalScript: List[ScriptToken],
     altStack: List[ScriptToken],
     flags: Seq[ScriptFlag])
-    extends ScriptProgram
+    extends ScriptProgram {
+  def addFlags(newFlags: Seq[ScriptFlag]): PreExecutionScriptProgram = {
+    this.copy(flags = newFlags)
```
Christewart: This isn't really adding is it? It's replacing the existing flags. Why not just use `.copy()` directly?

PR #795, opened 2019-10-11
Script Program apply method refactor part 3 (#760)

benthecarman:


core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
```diff
@@ -514,18 +514,14 @@ class ArithmeticInterpreterTest extends FlatSpec with MustMatchers {
   it must "evaluate an OP_WITHIN correctly" in {
     val stack = List(ScriptNumber(2), ScriptNumber.one, ScriptNumber.zero)
     val script = List(OP_WITHIN)
-    val program = ScriptProgram(
-      ScriptProgram(TestUtil.testProgramExecutionInProgress, stack, script),
-      Seq[ScriptFlag]())
+    val program = ScriptProgram(TestUtil.testProgramExecutionInProgress, stack, script).addFlags(Seq[ScriptFlag]())
```
Christewart: We should use `Seq.empty` rather than `Seq[ScriptFlag]()`, this avoids an allocation of a `Seq` object that is empty.

core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -94,7 +94,11 @@ case class ExecutionInProgressScriptProgram(
     altStack: List[ScriptToken],
     flags: Seq[ScriptFlag],
     lastCodeSeparator: Option[Int])
-    extends StartedScriptProgram
+    extends StartedScriptProgram {
+  def addFlags(newFlags: Seq[ScriptFlag]): ExecutionInProgressScriptProgram = {
```
nkohen: I agree with @Christewart s comment that this should probably be called `replaceFlags`, but additionally I think that we should also have a `removeFlags` method that does the equivalent of passing in an empty `newFlags` here since this does seem to be the most common use case by far

benthecarman: That makes a lot of sense, I'll do that instead

core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -108,6 +108,19 @@ case class ExecutionInProgressScriptProgram(
   override def failExecution(error: ScriptError): ExecutedScriptProgram = {
     ScriptProgram.toExecutedProgram(this).failExecution(error)
   }
+
+  def addFlags(newFlags: Seq[ScriptFlag]): ExecutionInProgressScriptProgram = {
+    this.copy(flags = newFlags)
```
Christewart: If you want to keep this, you need to do 

```
this.copy(flags = (flags ++ newFlags).distinct)
```

benthecarman: I meant to remove it, I think my rebase might have added it back on accident

Christewart: This adds the new flags to our existing flags, and then makes sure we do not have duplicates.

Christewart: ok, go ahead and remove and ignore comments. Rebasing can be tricky

Christewart:
Since we can't seem to get travis to work for you, i took the liberty of pulling down and running locally and got a failure:

```
$ sbt
sbt:bitcoin-s> coreTest/testOnly *ConstantInterpreterTest* -- -z  "return ScriptErrorMinimalData if program contains ScriptVerifyMinimalData flag and 2nd item in script is zero
```

is the test that is failing

benthecarman:
Fixed

PR #797, opened 2019-10-12
P2WSH Signer!

nkohen:
This PR is the first of what will be a fun series of PRs that clean up TxBuilder's signing logic. The end goal of this series will be to clean up technical debt and introduce script signing for scripts containing `OP_IF`. This will require better modularity than we currently have.

This PR specifically creates a new `Signer` for `P2WSHWitnessSPKV0`s. It was previously the responsibility of ever `Signer` to cover this case and this lead to a lot of redundancy as well as just being suboptimal design. Now, `P2WSHSigner` delegates work to the other `Signer`s who no longer need to understand `P2WSH`.

Future PRs will do similar things with `LockTimeScriptPubKey`, `P2SHScriptPubKey` and of eventually add `ConditionalScriptPubKey`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -76,17 +76,28 @@ sealed abstract class Signer {
   }
 }
 
+/** For use when signing nested ScriptPubKeys.
+  *
+  * This will require the hash to be signed to come from some external
+  * (to the Signer being used) TxSigComponent (think P2SH, P2WSH).
+  *
+  * This will also require that the ScriptPubKey that the nested Signer
+  * looks at needs to be overriden (not output.scriptPubKey)
+  */
 sealed abstract class OverridesForNestedSigning {
   def txSigComponentOpt: Option[TxSigComponent]
   def scriptPubKeyOpt: Option[ScriptPubKey]
 }
 
 object NestedSigning {
+
+  /** The NoOverrides case where a Signer is used raw, without nesting */
   case object NoOverrides extends OverridesForNestedSigning {
     override val txSigComponentOpt: Option[TxSigComponent] = None
     override val scriptPubKeyOpt: Option[ScriptPubKey] = None
   }
 
+  /** For P2WSH signing, this will be passed to the nested Signer */
   case class P2WSHOverrides(
       externalSigComponent: TxSigComponent,
       nestedSPK: ScriptPubKey)
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -85,26 +85,44 @@ sealed abstract class Signer {
   * looks at needs to be overriden (not output.scriptPubKey)
   */
 sealed abstract class OverridesForNestedSigning {
-  def txSigComponentOpt: Option[TxSigComponent]
-  def scriptPubKeyOpt: Option[ScriptPubKey]
+
+  /** Depending on the external [[ScriptPubKey scriptpubkey]] type, there
+    * are different rules for which script is meant to be signed.
+    *
+    * For instance, if you are spending a [[P2PKHScriptPubKey p2pkh]]
+    * you sign the p2pkh script itself.
+    *
+    * If you are signing a [[P2WPKHWitnessSPKV0 p2wpkh spk]], you sign the
+    * re-constructed [[P2PKHScriptPubKey p2pkh]] that is associated with p2wpkh script.
+    *
+    * If you are signing a [[LockTimeScriptPubKey locktime spk]], you must sign
+    * the entire [[LockTimeScriptPubKey]], not the nested script.
+    */
+  def sigComponentToSignOpt: Option[TxSigComponent]
+
+  /** This represents the [[ScriptPubKey scriptpubkey]] for which a valid ScriptSignature is
+    * being generated. This is usually some nested SPK such as a redeem script in P2SH and P2WSH.
+    */
+  def scriptPubKeyToSatisfyOpt: Option[ScriptPubKey]
 }
 
 object NestedSigning {
 
   /** The NoOverrides case where a Signer is used raw, without nesting */
   case object NoOverrides extends OverridesForNestedSigning {
-    override val txSigComponentOpt: Option[TxSigComponent] = None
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -61,7 +61,28 @@ sealed abstract class P2PKSigner extends BitcoinSigner {
...
+  }
+
+  def sign(
+      signers: Seq[Sign],
+      output: TransactionOutput,
+      unsignedTx: Transaction,
+      inputIndex: UInt32,
+      hashType: HashType,
+      isDummySignature: Boolean,
+      p2wshOverride: Option[(TxSigComponent, ScriptPubKey)])(
```
nkohen: Hoping to find a nice general type for whatever this will be in the future. `timeLockOverride` will have just a `ScriptPubKey` while `OP_IF` will require an override with an SPK as well, actually now that I am writing this down I can think of the perfect type, an ADT :) will go do this now

nkohen: Done :)

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -47,6 +76,27 @@ sealed abstract class Signer {
   }
 }
 
+sealed abstract class OverridesForNestedSigning {
```
Christewart, line 87: Scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -47,6 +76,27 @@ sealed abstract class Signer {
...
+  def txSigComponentOpt: Option[TxSigComponent]
+  def scriptPubKeyOpt: Option[ScriptPubKey]
```
Christewart: It seems weird to me that there are two options defined in the parent type? Why not just do 

```scala
sealed abstract class OverridesForNestedSigning

object NestedSigning {
  case object NoOverrides extends OverridesForNestedSigning

  case class P2WSHOverrides(
      externalSigComponent: TxSigComponent,
      nestedSPK: ScriptPubKey)
      extends OverridesForNestedSigning
}
```

Christewart: We may be able to tighten the type on `externalSigComponent` -- if I understand what is going on here -- to `WitnessTxSigComponent`

nkohen: When I (in the next couple PRs) introduce signers for Locktime scripts and P2SH then the Options are the interface that general signers (like P2PKH and MultiSig) will use. (and in the P2SH case I don't believe that `WitnessTxSigComponenet` will do. Will add a scaladoc of course :)

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -47,6 +76,38 @@ sealed abstract class Signer {
...
+  /** The NoOverrides case where a Signer is used raw, without nesting */
+  case object NoOverrides extends OverridesForNestedSigning {
+    override val txSigComponentOpt: Option[TxSigComponent] = None
+    override val scriptPubKeyOpt: Option[ScriptPubKey] = None
+  }
+
+  /** For P2WSH signing, this will be passed to the nested Signer */
+  case class P2WSHOverrides(
+      externalSigComponent: TxSigComponent,
+      nestedSPK: ScriptPubKey)
```
Christewart: So I think this name should be changed to `redeemScript` as that is less confusing than `nestedSPK` which is used in the locktime PR. This is the `redeemScript` right?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
Christewart: So here is what I typed out as a scaladoc for this method: 

```
  /** Depending on the [[ScriptPubKey scriptpubkey]] type, there
    * are different rules for which script is meant to be signed.
    *
    * For instance, if you are spending a [[P2PKHScriptPubKey p2pkh]] you sign
    * the p2pkh script itself.
    *
    * If you are signing a [[P2WSHWitnessSPKV0 p2wsh spk]], you sign the [[P2WSHWitnessV0.redeemScript redeemscript]]
    * associated with that p2wsh script. In that case, the [[scriptPubKeyOpt]] field MUST be popualted with the redeemScript
    *
    * If you are signing a [[P2WPKHWitnessSPKV0 p2wpkh spk]], you sign the re-constructed [[P2PKHScriptPubKey p2pkh]]
    * that is associated with p2wpkh script. In that case, the [[scriptPubKeyOpt]] must be populated with the p2pkh spk
    *
    * If you are signing a [[P2SHScriptPubKey p2sh spk]], you sign the p2sh script itself as BIP16 did not specify
    * any new rules for signature serialization like segwit did. This means you must populate [[scriptPubKeyOpt]]
    * with the [[P2SHScriptPubKey p2sh spk]] you are signing for.
    *
    * If you are signing a [[LockTimeScriptPubKey locktime spk]], you must sign the entire [[LockTimeScriptPubKey]]
    * itself, but you must _satisfy_ the script requirements of the [[LockTimeScriptPubKey.nestedScriptPubKey nested spk]]
    * inside of the Locktime spk. This means you modify [[scriptPubKeyOpt]] to
    * be the [[LockTimeScriptPubKey.nestedScriptPubKey nested spk]] ?
    * */
```

and here is what I realized: There is a distinction between satisfying and signing the `scriptPubKeyOpt` and we are using them interchangeably here and #798 . 

[With `p2wsh` we need to satisfy/sign the `redeemScript`/`nestedSPK`](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification)

[With `p2wpkh` we need to satisfy/sign the "redeemScript" (rebuilt p2pkh)](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification)

With `p2sh`/`locktime spk` we need to satisfy the `redeemScript`/`nestedSPK`, but _sign_ the `p2sh`/`locktime` spk

It seems like our ADT should represent each of these possibilities

Christewart: I would also bet in favor of renaming this field `scriptForSigning` or something like that

PR #798, opened 2019-10-12
LockTimeSigner

nkohen:
Part 2 following #797 (and built on top of it so that only the last two commits are new).

This PR creates a new `Signer` for `LockTimeScriptPubKey`s so that it is no longer the case that every `Signer` is responsible for this case.

Since this is relatively small, I also included a tiny bit of clean-up

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -124,6 +124,31 @@ object NestedSigning {
     override val scriptPubKeyToSatisfyOpt: Option[ScriptPubKey] = Some(
       redeemScript)
   }
+
+  /** For Locktime signing, this will be passed to the nested Signer.
+    *
+    * @param nestedSPK This should be [[LockTimeScriptPubKey.nestedScriptPubKey]]
+    */
+  case class LockTimeOverrides(nestedSPK: ScriptPubKey)
+      extends OverridesForNestedSigning {
+    override val sigComponentToSignOpt: Option[TxSigComponent] = None
+    override val scriptPubKeyToSatisfyOpt: Option[ScriptPubKey] = Some(
+      nestedSPK)
+  }
+
+  /** For signing when a LockTime script is nested, such as within P2WSH.
+    *
+    * @param nestedSPK This should be [[LockTimeScriptPubKey.nestedScriptPubKey]]
+    */
+  case class NestedLockTimeOverrides(
+      externalSigComponent: TxSigComponent,
+      nestedSPK: ScriptPubKey)
+      extends OverridesForNestedSigning {
+    override val sigComponentToSignOpt: Option[TxSigComponent] = Some(
+      externalSigComponent)
+    override val scriptPubKeyToSatisfyOpt: Option[ScriptPubKey] = Some(
+      nestedSPK)
+  }
 }
 
 /** Represents all signers for the bitcoin protocol, we could add another network later like litecoin */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -106,6 +106,12 @@ object NestedSigning {
       externalSigComponent)
     override val scriptPubKeyOpt: Option[ScriptPubKey] = Some(nestedSPK)
   }
+
+  case class LockTimeOverrides(nestedSPK: ScriptPubKey)
```
Christewart, line 132: What does nested spk mean in this case? I'm guessing it means `LockTimeScriptPubKey.nestedScriptPubKey`, lets link to that in the scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -641,12 +503,73 @@ sealed abstract class P2WSHSigner extends BitcoinSigner {
...
+      signers: Seq[Sign],
+      output: TransactionOutput,
+      unsignedTx: Transaction,
+      inputIndex: UInt32,
+      hashType: HashType,
+      isDummySignature: Boolean,
+      overrides: OverridesForNestedSigning)(
+      implicit ec: ExecutionContext): Future[TxSigComponent] = {
+    val spk = overrides.scriptPubKeyOpt match {
+      case None               => output.scriptPubKey
```
Christewart, line 565: Shouldn't this always be defined? 

https://github.com/bitcoin-s/bitcoin-s/pull/798/commits/2998fd6cbe40cc35fa4431966afe58477bb2c7e1#diff-5491ddd7f432d5aea914d2d4bd4d65c8R113

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -641,12 +503,73 @@ sealed abstract class P2WSHSigner extends BitcoinSigner {
...
+              _: CLTVScriptPubKey | _: CSVScriptPubKey | _: WitnessCommitment |
+              EmptyScriptPubKey | _: UnassignedWitnessScriptPubKey =>
+            Future.fromTry(TxBuilderError.WrongSigner)
+        }
+        signerF.flatMap { signer =>
+          val overridesForNestedSigning = overrides match {
+            case NestedSigning.NoOverrides =>
+              NestedSigning.LockTimeOverrides(lockSPK.nestedScriptPubKey)
+            case lockTimeOverride: NestedSigning.LockTimeOverrides =>
+              lockTimeOverride.copy(nestedSPK = lockSPK.nestedScriptPubKey)
```
Christewart: This seems a little hacky, couldn't we just add this as a field on `NestedSigning.LockTimeOverrides` and then extract the approprate script inside of the various Signer?  I.e. 

```scala
    val spk = overrides match {
      case NestedSigning.NoOverrides => output.scriptPubKey
      case p : NestedSigning.P2WSHOverrides => p.nestedSPK
      case l : NestedSigning.LockTimeOverrides => l.nestedSPK
    }
```

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
nkohen: As discussed on our call, in the case of a raw CSV or CLTV script this will be None

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
nkohen: I figured out that weird stuff was coming from there not being a `NestedSigning.NestedLockTimeOverrides`, fixed this.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -124,6 +124,31 @@ object NestedSigning {
...
       redeemScript)
   }
+
+  /** For Locktime signing, this will be passed to the nested Signer.
+    *
+    * @param nestedSPK This should be [[LockTimeScriptPubKey.nestedScriptPubKey]]
+    */
+  case class LockTimeOverrides(nestedSPK: ScriptPubKey)
+      extends OverridesForNestedSigning {
+    override val sigComponentToSignOpt: Option[TxSigComponent] = None
```
Christewart, line 134: Why is this `None`?

nkohen: So I've realized that things work in a kinda weird way right now where even on nested calls, we pass in the external parameters `output, unsignedTx, inputIndex`. Thus, unless the nesting is withing a Witness SPK which would then require a `WitnessSigComponent` (as opposed to `Base`) then the `TxSigComponent` is actually constructed correctly.

I did not change this weirdness in the spirit of not changing any control flow I don't have to in my refactors.

I opened #828 as an attempt to make this better and clearer and generally simplify things. Let me know what you think of the approach.

nkohen: #828 has been replaced by #830

PR #799, opened 2019-10-12
Update script_tests.json, fix bugs that were unveiled with that updat…

Christewart:
…e. Specifically around handling negative zero in Script, and cleanstack behavior for segwit scripts. We were not checking for equality at the byte level for negative zero. With cleanstack, we were failing the script, but not failing with the correct error. If you look at interpreter.cpp in bitcoin core, cleanstack check is done before checking if the stacktop is true or false

Here is the latest script_tests.json file that I copied over from bitcoin core: https://github.com/bitcoin/bitcoin/blob/master/src/test/data/script_tests.json


core/src/main/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreter.scala
```diff
@@ -111,7 +116,11 @@ sealed abstract class ArithmeticInterpreter {
     require(program.script.headOption.contains(OP_BOOLOR),
             "Script top must be OP_BOOLOR")
     performBinaryBooleanOperation(program, (x, y) => {
-      if (x == y && (x == ScriptNumber.zero || x == OP_0)) false else true
+      if (x == y && (x == ScriptNumber.zero
+          || x == OP_0
+          || x == ScriptNumber.negativeZero
+          || x.toLong == 0)) false
```
nkohen: nit: I think this is long enough now to justify a util method. This will simplify the operation on line 104-108 to
```
SomethingUtil.isZeroOrFalse(x) || SomethingUtil.isZeroOrFalse(y)
```

nkohen (approved):

PR #811, opened 2019-10-14
Removed the ScriptProgram companion object

### LINKED ISSUE #760: Replace `ScriptProgram` companion object with pertinent methods that have names

nkohen:
Closes #760

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/script/stack/StackInterpreter.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/stack/StackInterpreter.scala
@@ -2,7 +2,6 @@ package org.bitcoins.core.script.stack
 
 import org.bitcoins.core.script.{
   ExecutionInProgressScriptProgram,
-  ScriptProgram,
   StartedScriptProgram
 }
 import org.bitcoins.core.script.constant._
@@ -29,7 +28,7 @@ sealed abstract class StackInterpreter {
             "Top of the script stack must be OP_DUP")
     program.stack match {
       case h :: _ =>
-        ScriptProgram(program, h :: program.stack, program.script.tail)
+        program.updateStackAndScript(h :: program.stack, program.script.tail)
       case Nil =>
         logger.error("Cannot duplicate the top element on an empty stack")
         program.failExecution(ScriptErrorInvalidStackOperation)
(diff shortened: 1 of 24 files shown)


nkohen:
There's a lot of fuzz in this review but make sure to take a close look at `ScriptProgram.scala`

core/src/main/scala/org/bitcoins/core/script/stack/StackInterpreter.scala
```diff
@@ -69,10 +68,10 @@ sealed abstract class StackInterpreter {
     require(program.script.headOption.contains(OP_TOALTSTACK),
             "Top of script stack must be OP_TOALTSTACK")
     if (program.stack.nonEmpty) {
-      ScriptProgram(program,
-                    program.stack.tail,
-                    program.script.tail,
-                    program.stack.head :: program.altStack)
+      program
+        .updateStack(program.stack.tail)
```
benthecarman, line 72: nit: This could be
```
program.
updateStackAndScript(program.stack.tail, program.script.tail).
updateAltStack(program.stack.head :: program.altStack)
```

Not sure what you guys would consider cleaner

core/src/main/scala/org/bitcoins/core/script/stack/StackInterpreter.scala
```diff
@@ -85,10 +84,10 @@ sealed abstract class StackInterpreter {
     require(program.script.headOption.contains(OP_FROMALTSTACK),
             "Top of script stack must be OP_FROMALTSTACK")
     if (program.altStack.nonEmpty) {
-      ScriptProgram(program,
-                    program.altStack.head :: program.stack,
-                    program.script.tail,
-                    program.altStack.tail)
+      program
+        .updateStack(program.altStack.head :: program.stack)
```
benthecarman, line 88: Same as previous comment

benthecarman (approved):
LGTM

Christewart (approved):

Issue #813, opened 2019-10-16
link to  txbuilder examples on website broken


philbertw4:
**Describe the bug**
Great work you guys are doing by providing a clean interface into bitcoin from scala! Unfortunately when clicking the link on the home page to view the TxBuilder example, I was greeted with a 404.

**To Reproduce**
1. go to bitcoin-s website
2. click on "examples"

**Expected behavior**
I'm not sure where you intended for it to point, so that's why i'm opening the issue. But it would be great if in fact it did point to some nice example uses of the library :-)

philbertw4:
closing this in favor of the pull request #815 that hopefully fixes it

PR #815, opened 2019-10-16
fix of link to TxBuilder example on home page

philbertw4:
Found where the link should point (currently getting a 404) mentioned in #813. Hopefully this fixes it.


Christewart:
Thank you!

philbertw4:
No problem. However I noticed there is another link that is not working:
the slack invite link seems to be expired. The link appears both on the
help page and also in the gitter chat.

For this I cannot open a proper pull request for you since I do not have a
slack invite link :-) If you can supply the proper invite link here I can
update my branch and push a PR if that would be helpful?

PR #818, opened 2019-10-20
Upgrade to ecalir v0.3.2

rorp:
Note that /allupdates API is broken in Eclair 0.3.2 (https://github.com/ACINQ/eclair/issues/1179), so all test cases that use it are marked as pending.

### DIFF
--- a/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
+++ b/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
@@ -13,7 +13,6 @@ import org.bitcoins.core.protocol.ln.fee.FeeProportionalMillionths
 import org.bitcoins.core.protocol.ln.node.NodeId
 import org.bitcoins.core.protocol.ln.{
   LnHumanReadablePart,
-  LnInvoiceSignature,
   PaymentPreimage,
   ShortChannelId
 }
@@ -100,31 +99,39 @@ case class NetworkFeesResult(
     remoteNodeId: NodeId,
     channelId: FundedChannelId,
     txId: DoubleSha256DigestBE,
-    feeSat: Satoshis,
+    fee: Satoshis,
     txType: String,
     timestamp: FiniteDuration
 )
 
 case class ChannelStats(
     channelId: FundedChannelId,
-    avgPaymentAmountSatoshi: Satoshis,
+    avgPaymentAmount: Satoshis,
     paymentCount: Long,
-    relayFeeSatoshi: Satoshis,
-    networkFeeSatoshi: Satoshis
+    relayFee: Satoshis,
+    networkFee: Satoshis
 )
 
 case class UsableBalancesResult(
-    canSendMsat: MilliSatoshis,
-    canReceiveMsat: MilliSatoshis,
+    remoteNodeId: NodeId,
+    shortChannelId: ShortChannelId,
+    canSend: MilliSatoshis,
+    canReceive: MilliSatoshis,
     isPublic: Boolean
 )
 
 case class ReceivedPayment(
-    amount: MilliSatoshis,
     paymentHash: Sha256Digest,
-    fromChannelId: FundedChannelId,
-    timestamp: FiniteDuration
+    parts: Vector[ReceivedPayment.Part]
 )
+
+object ReceivedPayment {
+  case class Part(
+      amount: MilliSatoshis,
+      fromChannelId: FundedChannelId,
+      timestamp: FiniteDuration
+  )
+}
 
 case class RelayedPayment(
     amountIn: MilliSatoshis,
@@ -136,13 +143,19 @@ case class RelayedPayment(
 )
 
 case class SentPayment(
-    amount: MilliSatoshis,
-    feesPaid: MilliSatoshis,
     paymentHash: Sha256Digest,
-    paymentPreimage: String,
-    toChannelId: FundedChannelId,
-    timestamp: FiniteDuration
+    paymentPreimage: String
 )
+
+object SentPayment {
+  case class Part(
+      id: String,
+      amount: MilliSatoshis,
+      feesPaid: MilliSatoshis,
+      toChannelId: FundedChannelId,
+      timestamp: FiniteDuration
+  )
+}
 
 case class ChannelUpdate(
     signature: ECDigitalSignature,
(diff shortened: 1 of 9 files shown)


eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -136,14 +143,20 @@ case class RelayedPayment(
 )
 
 case class SentPayment(
-    amount: MilliSatoshis,
-    feesPaid: MilliSatoshis,
     paymentHash: Sha256Digest,
-    paymentPreimage: String,
-    toChannelId: FundedChannelId,
-    timestamp: FiniteDuration
+    paymentPreimage: String
```
Christewart: `paymentPreimage: PaymentPreimage`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -136,14 +143,20 @@ case class RelayedPayment(
...
 )
 
+object SentPayment {
+  case class Part(
+      id: String,
```
Christewart: should this be `id: UUID` ?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -290,48 +303,63 @@ case class InvoiceResult(
     paymentHash: Sha256Digest,
     expiry: FiniteDuration)
 
+case class PaymentId(value: String) {
```
Christewart: Can we tighten the type here? Is this a `UUID`?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -290,48 +303,63 @@ case class InvoiceResult(
...
-    completedAt: Option[FiniteDuration],
+    paymentRequest: Option[PaymentRequest],
     status: PaymentStatus)
 
 case class ReceivedPaymentResult(
-    paymentHash: Sha256Digest,
-    amountMsat: MilliSatoshis,
-    receivedAt: FiniteDuration)
+    paymentRequest: Option[PaymentRequest],
+    paymentPreimage: String,
```
Christewart: `paymentPreimage: PaymentPreimage`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -290,48 +303,63 @@ case class InvoiceResult(
...
-    signature: LnInvoiceSignature)
+    serialized: String,
+    description: String,
+    paymentHash: Sha256Digest,
+    expiry: FiniteDuration,
+    amount: Option[MilliSatoshis])
 
 case class PaymentResult(
     id: String,
+    parentId: String,
```
Christewart: both of these can be `UUID` i believe

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -290,48 +303,63 @@ case class InvoiceResult(
...
+    extends PaymentStatus
+
+case class Hop(
+    nodeId: NodeId,
+    nextNodeId: NodeId,
+    shortChannelId: ShortChannelId)
+
+case class PaymentFailed(failures: Seq[PaymentFailure]) extends PaymentStatus
+
+case class PaymentFailureType(name: String)
```
Christewart: How many different `PaymentFailureType` is there? We should probably enumerate them with an ADT so we handle these in a more robust way.

testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -255,7 +263,7 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
       failFast: Boolean = true)(implicit system: ActorSystem): Future[Unit] = {
     awaitUntilPaymentStatus(client,
                             paymentId,
-                            PaymentStatus.SUCCEEDED,
+                            "sent",
```
Christewart: I don't like removing types here and replacing them with `String`. Now we will never get compiler errors if these things are changed again. Why did you remove the `PaymentStatus` here?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
```diff
@@ -290,48 +303,63 @@ case class InvoiceResult(
...
-  def apply(s: String): PaymentStatus = s match {
-    case "PENDING"   => PENDING
-    case "SUCCEEDED" => SUCCEEDED
-    case "FAILED"    => FAILED
-    case err =>
-      throw new IllegalArgumentException(
-        s"Unknown payment status code `${err}`")
-  }
-}
+case object PaymentPending extends PaymentStatus
```
Christewart: Why did we get rid of `PaymentStatus`?

testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
rorp: Because there's no such entity in the RPC responses anymore

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
rorp: Because this RPC was changed significantly.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
rorp: Why?

testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
Christewart: So do we need to change the `awaitUntilPaymentStatus` interface to account for this then? 

Could we keep the `PaymentStatus` stuff, but make it internal to `EclairRpcTestUtil` and make it `private`?

rorp: Those string  literals are only used locally inside `EclairRpcTestUtil.awaitUntilPaymentStatus` I cant see any problem with doing it. I can replace them with `Int`s.

Christewart: Fair enough, as long as we don't expose `awaitUntilPaymentStatus()`. I don't like using raw Strings as parameters as it is very hard to update  as we don't get compiler errors if we add a new String that represents a state. For instance if we had to segregate "failed" into "failed-onchain"/"failed-routing". We need to remember to update it, rather than having the compiler work for us and remind us to update it.

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/json/EclairModels.scala
Christewart: It seems that every other `id` in eclair is a UUID.

However I deleted this `object SentPayment` in src and it is only used in one other place -- json serializers. Why do we have this?

PR #823, opened 2019-10-22
POC: Create an ADT to represent the various types of utxos we can have. No…

Christewart:
…w we have RawScriptUTXOSpendingInfo, SegwitV0NativeUTXOSpendingInfo, P2SHSpendingInfo, P2SHNestedSegwitV0UTXOSpendingInfo. This hopefully can be used in combination with the P2WSHSigner and LockTimeSigner work to make TxBuilder and the Signers more comprehensible

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
@@ -1,11 +1,9 @@
 package org.bitcoins.core.wallet.utxo
 
 import org.bitcoins.core.crypto.Sign
-import org.bitcoins.core.protocol.script.{ScriptPubKey, ScriptWitness}
-import org.bitcoins.core.protocol.transaction.{
-  TransactionOutPoint,
-  TransactionOutput
-}
+import org.bitcoins.core.currency.CurrencyUnit
+import org.bitcoins.core.protocol.script.{P2SHScriptPubKey, ScriptPubKey, ScriptWitnessV0, WitnessScriptPubKeyV0}
+import org.bitcoins.core.protocol.transaction.{TransactionOutPoint, TransactionOutput}
 import org.bitcoins.core.script.crypto.HashType
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -17,26 +15,71 @@ sealed abstract class UTXOSpendingInfo {
...
+                                     amount: CurrencyUnit,
+                                     scriptPubKey: ScriptPubKey,
+                                     signers: Seq[Sign],
+                                     hashType: HashType) extends BitcoinUTXOSpendingInfo
+
+
+/** This is the case where we are spending a [[org.bitcoins.core.protocol.script.WitnessScriptPubKeyV0 witness v0 script]]  */
+case class SegwitV0NativeUTXOSpendingInfo(outPoint: TransactionOutPoint,
+                                          amount: CurrencyUnit,
+                                          scriptPubKey: WitnessScriptPubKeyV0,
```
Christewart, line 49: I think the most important things here are the tightenting of types on scripts, for instance `WitnessScriptPubKeyV0` here rather than just `ScriptPubKey`. This should reduce the amount of pattern matching required significantly.

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -17,26 +15,71 @@ sealed abstract class UTXOSpendingInfo {
...
+                                          signers: Seq[Sign],
+                                          hashType: HashType,
+                                          scriptWitness: ScriptWitnessV0) extends BitcoinUTXOSpendingInfo
+
+/** This is the case were we are attempting to spend a [[org.bitcoins.core.protocol.script.P2SHScriptPubKey p2sh spk]] */
+case class P2SHSpendingInfo(outPoint: TransactionOutPoint,
+                            amount: CurrencyUnit,
+                            scriptPubKey: P2SHScriptPubKey,
```
Christewart, line 57: Same could be said here with `P2SHScriptPubKey` rather than `ScriptPubKey`

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -17,26 +15,71 @@ sealed abstract class UTXOSpendingInfo {
...
+                                               hashType: HashType,
+                                               redeemScript: WitnessScriptPubKeyV0,
+                                               scriptWitness: ScriptWitnessV0
+) extends BitcoinUTXOSpendingInfo {
+  require(P2SHScriptPubKey(redeemScript) == output.scriptPubKey,
+    s"Given redeem script did not match hash in output script, " +
+      s"got=${P2SHScriptPubKey(redeemScript).scriptHash.hex}, " +
+      s"expected=${scriptPubKey.scriptHash.hex}")
+
+  //can we put an invariant here with scriptWitness and the witness spk redeemScript?
```
nkohen, line 84: Yes, I believe if you match on the `redeemScript` you should be able to add an invariant depending on whether it is pubkey or script

nkohen (commented):
I see no reason from what I'm looking at in this diff that this isn't a great improvement, I would still like to see that the interface is good enough with the `redeemScriptOpt` and `scriptWitnessOpt` gone and such

Christewart:
Superseded by #834

PR #826, opened 2019-10-23
Improve EclairRpcClientTest execution time

rorp:


testkit/src/main/scala/org/bitcoins/testkit/eclair/rpc/EclairRpcTestUtil.scala
```diff
@@ -119,7 +119,7 @@ trait EclairRpcTestUtil extends BitcoinSLogger {
         "eclair.api.port" -> apiPort,
         "eclair.mindepth-blocks" -> 2,
         "eclair.max-htlc-value-in-flight-msat" -> 100000000000L,
-        "eclair.router-broadcast-interval" -> "2 second",
+        "eclair.router.broadcast-interval" -> "2 second",
```
Christewart, line 122: wow... :man_facepalming:

PR #827, opened 2019-10-23
OP_IF Signing

nkohen:
Adds the `ConditionalScriptPubKey` and `ConditionalScriptSignature` types which allow for non-nested `OP_IF ... OP_ELSE ... OP_ENDIF` scripts and their signing.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -4,7 +4,7 @@ import org.bitcoins.core.consensus.Consensus
 import org.bitcoins.core.crypto._
 import org.bitcoins.core.script.bitwise.{OP_EQUAL, OP_EQUALVERIFY}
 import org.bitcoins.core.script.constant.{BytesToPushOntoStack, _}
-import org.bitcoins.core.script.control.OP_RETURN
+import org.bitcoins.core.script.control.{OP_ELSE, OP_ENDIF, OP_IF, OP_RETURN}
 import org.bitcoins.core.script.crypto.{
   OP_CHECKMULTISIG,
   OP_CHECKMULTISIGVERIFY,
@@ -555,6 +555,85 @@ object CSVScriptPubKey extends ScriptFactory[CSVScriptPubKey] {
 
 }
 
+/** Currently only supports a single OP_IF ... OP_ELSE ... OP_ENDIF ScriptPubKey */
+sealed trait ConditionalScriptPubKey extends RawScriptPubKey {
+  require(asm.nonEmpty, "ConditionalScriptPubKey cannot be empty")
+  require(asm.head.equals(OP_IF),
+          "ConditionalScriptPubKey must begin with OP_IF")
+  require(opElseIndex != -1,
+          "ConditionalScriptPubKey has to contain OP_ELSE asm token")
+  require(asm.last.equals(OP_ENDIF),
+          "ConditionalScriptPubKey must end in OP_ENDIF")
+  require(asm.count(_.equals(OP_IF)) == 1,
+          "ConditionalScriptPubKey does not currently support nesting OP_IFs")
+  require(asm.count(_.equals(OP_ELSE)) == 1,
+          "ConditionalScriptPubKey does not currently support nesting OP_ELSEs")
+
+  require(!P2SHScriptPubKey.isP2SHScriptPubKey(trueSPK.asm) && !P2SHScriptPubKey
+            .isP2SHScriptPubKey(falseSPK.asm),
+          "ConditionalScriptPubKey cannot wrap P2SH")
+  require(
+    !WitnessScriptPubKey
+      .isWitnessScriptPubKey(trueSPK.asm) && !WitnessScriptPubKey
+      .isWitnessScriptPubKey(falseSPK.asm),
+    "ConditionalScriptPubKey cannot wrap SegWit ScriptPubKey"
+  )
+
+  def opElseIndex: Int = {
+    asm.indexOf(OP_ELSE)
+  }
+
+  def trueSPK: RawScriptPubKey = {
+    RawScriptPubKey
+      .fromAsm(asm.slice(1, opElseIndex))
+  }
+
+  def falseSPK: RawScriptPubKey = {
+    RawScriptPubKey
+      .fromAsm(asm.slice(opElseIndex + 1, asm.length - 1))
+  }
+}
+
+object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
+  private case class ConditionalScriptPubKeyImpl(
+      override val asm: Vector[ScriptToken])
+      extends ConditionalScriptPubKey {
+    override def toString: String =
+      s"ConditionalScriptPubKey($trueSPK, $falseSPK)"
+  }
+
+  override def fromAsm(asm: Seq[ScriptToken]): ConditionalScriptPubKey = {
+    buildScript(
+      asm = asm.toVector,
+      constructor = ConditionalScriptPubKeyImpl.apply,
+      invariant = isConditionalScriptPubKey,
+      errorMsg = "Given asm was not a ConditionalScriptPubKey, got: " + asm
+    )
+  }
+
+  def apply(
+      trueSPK: RawScriptPubKey,
+      falseSPK: RawScriptPubKey): ConditionalScriptPubKey = {
+    val asm = Vector(OP_IF) ++ trueSPK.asm ++ Vector(OP_ELSE) ++ falseSPK.asm ++ Vector(
+      OP_ENDIF)
+
+    fromAsm(asm)
+  }
+
+  def isConditionalScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
+    val headIsOpIf = asm.headOption.contains(OP_IF)
+    lazy val containsOpElse = {
+      val opElseIndex = asm.indexOf(OP_ELSE)
+      opElseIndex != -1
+    }
+    lazy val singleOpIf = asm.count(_.equals(OP_IF)) == 1
+    lazy val singleOpElse = asm.count(_.equals(OP_ELSE)) == 1
+    lazy val endsWithEndIF = asm.last == OP_ENDIF
+
+    headIsOpIf && containsOpElse && singleOpIf && singleOpElse && endsWithEndIF
+  }
+}
+
 sealed trait NonStandardScriptPubKey extends RawScriptPubKey
 
 object NonStandardScriptPubKey extends ScriptFactory[NonStandardScriptPubKey] {
@@ -585,6 +664,8 @@ object RawScriptPubKey extends ScriptFactory[RawScriptPubKey] {
 
   def fromAsm(asm: Seq[ScriptToken]): RawScriptPubKey = asm match {
     case Nil => EmptyScriptPubKey
+    case _ if ConditionalScriptPubKey.isConditionalScriptPubKey(asm) =>
+      ConditionalScriptPubKey.fromAsm(asm)
     case _ if P2PKHScriptPubKey.isP2PKHScriptPubKey(asm) =>
       P2PKHScriptPubKey(asm)
     case _ if P2PKScriptPubKey.isP2PKScriptPubKey(asm) => P2PKScriptPubKey(asm)
@@ -696,9 +777,9 @@ object WitnessScriptPubKey {
 
     //we can also have a LockTimeScriptPubKey with a nested 0 public key multisig script, need to check that as well
     val bytes = BitcoinSUtil.toByteVector(asm)
-    val isMultiSig =
+    lazy val isMultiSig =
       MultiSignatureScriptPubKey.isMultiSignatureScriptPubKey(asm)
-    val isLockTimeSPK = {
+    lazy val isLockTimeSPK = {
       LockTimeScriptPubKey.isValidLockTimeScriptPubKey(asm)
     }
 
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
@@ -235,6 +235,7 @@ object P2SHScriptSignature extends ScriptFactory[P2SHScriptSignature] {
               _: CSVScriptPubKey | _: WitnessScriptPubKeyV0 |
               _: UnassignedWitnessScriptPubKey =>
             true
+          case _: ConditionalScriptPubKey                        => false
           case _: NonStandardScriptPubKey | _: WitnessCommitment => false
           case EmptyScriptPubKey                                 => false
         }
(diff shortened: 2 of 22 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -545,6 +545,82 @@ object CSVScriptPubKey extends ScriptFactory[CSVScriptPubKey] {
 
 }
 
+/** Currently only supports a single OP_IF ... OP_ELSE ... OP_ENDIF ScriptPubKey */
+sealed trait ConditionalScriptPubKey extends ScriptPubKey {
+  require(asm.nonEmpty, "ConditionalScriptPubKey cannot be empty")
+  require(asm.head.equals(OP_IF),
```
Christewart, line 561: probably should allow `OP_NOTIF`

Christewart: normally i would say skip it to simplify for now, but it is used in Lightning scripts:

https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#offered-htlc-outputs

Christewart: Another consideration: 

What if the condition is _not_ the first thing evaluated in the Script? Is that still a conditional Script? I.e. 

https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#offered-htlc-outputs

Christewart: Finally, I know you have the stated goal of not fully supporting conditionals here, which I think is wise. I just want to provide an example of a valid conditional Script that shows the insanity of Script, so we can at least mull over what it would look like to support nesting: https://github.com/bitcoin-s/bitcoin-s/blob/master/core-test/src/test/resources/script_tests.json#L62

nkohen: So I'm leaning towards getting this in and then adding an `OP_NOTIF` immediately after which should be much easier once `OP_IF` is in, at the very least I plan on getting all of the `OP_IF` code done on this PR before adding any `OP_NOTIF` related commits.

Re `OP_IF` not being first if this is the case I would say that good design would dictate a new SPK that has a `ConditionalScriptPubKey` nested within it, much like how my `MultiSignatureWithTimeoutScriptPubKey` will be nested within a `ConditionalScriptPubKey`. As a currently supported example, if a both conditions have the same locktime then one should construct a `CLTVScriptPubKey(ConditionalScriptPubKey(...`

That being said, I am aware that this will not work for nesting `OP_IF` scripts since my current asm processing code assumes a single `OP_IF ... OP_ELSE ... OP_ENDIF` but I will improve on this in the future.

nkohen: Also Re BOLT 3 offered HTLCs, those would require nesting anyway

Christewart: Something that might be helpful with understanding nesting is looking at the implementation of OP_IF in the interpreter, we use a binary tree data structure that represents the program. 

https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/script/control/ControlOperationsInterpreter.scala#L22

Tangentially, we will probably need to integrate something like this for Taproot, so we might as well start thinking about the best way to handle this generically.

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -282,22 +323,64 @@ case class MultiSignatureSpendingInfo(
...
+    case ConditionalPath.ConditionTrue =>
+      true
+    case ConditionalPath.ConditionFalse =>
+      false
+    case ConditionalPath.NoConditionsLeft =>
+      throw new IllegalStateException(
+        "This should be covered by invariant above")
+  }
+
+  def nestedSpendingInfo: RawScriptUTXOSpendingInfo = {
```
nkohen: should be `val`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -696,9 +777,9 @@ object WitnessScriptPubKey {
 
     //we can also have a LockTimeScriptPubKey with a nested 0 public key multisig script, need to check that as well
     val bytes = BitcoinSUtil.toByteVector(asm)
-    val isMultiSig =
+    lazy val isMultiSig =
       MultiSignatureScriptPubKey.isMultiSignatureScriptPubKey(asm)
-    val isLockTimeSPK = {
+    lazy val isLockTimeSPK = {
```
nkohen, line 782: For the record, this was a bug where we were checking asm under the assumption that it was non-empty in cases when it was empty

core/src/main/scala/org/bitcoins/core/script/constant/Constants.scala
```diff
@@ -191,6 +191,15 @@ sealed abstract class ScriptNumberOperation
...
   override def hex = opCode.toHexString
+
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case number: ScriptNumber =>
+        number.toLong == toLong
+      case _ =>
+        super.equals(obj)
+    }
+  }
```
nkohen, line 205: For the record, this was a bug where we had `OP_TRUE == OP_1` and `OP_FALSE == OP_0` as `false`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -681,10 +848,20 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
...
+        val spk = lock match {
+          case csv: CSVScriptPubKey =>
+            CSVScriptPubKey(csv.locktime, conditional.trueSPK)
+          case cltv: CLTVScriptPubKey =>
+            CLTVScriptPubKey(cltv.locktime, conditional.trueSPK)
+        }
+        val scriptSig = lockTimeHelperScriptSig(spk, sigs, keys).scriptSig
+        ConditionalScriptSignature(scriptSig, true)
+      case EmptyScriptPubKey =>
+        CSVScriptSignature(NonStandardScriptSignature("0151"))
```
nkohen, line 865: For the record, this was a bug that was never hit (since previously CLTV/CSV never wrapped `EmptyScriptPubKey`) where we must use `NonStandardScriptSignature(OP_TRUE)` rather than `EmptyScriptSignature` to satisfy an `EmptyScriptPubKey` in the `ScriptInterpreter`'s eyes

nkohen (commented):
Bugs found and fixed:

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -432,6 +433,59 @@ object CSVScriptSignature extends ScriptFactory[CSVScriptSignature] {
   }
 }
 
+sealed trait ConditionalScriptSignature extends ScriptSignature {
+  require(asm.lastOption.contains(OP_TRUE) || asm.lastOption.contains(OP_FALSE),
+          "ConditionalScriptSignature must end in true or false")
+
+  def isTrue: Boolean = {
+    asm.last == OP_TRUE
```
Christewart: It would be nice to have a `BitcoinScriptUtil.isTrue()` and `BitcoinScriptUtil.isFalse()`, note we do do have a [`BitcoinScriptUtil.castToBool()`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala#L516).

Sadly something being considered a bool isn't as strict as `OP_TRUE/OP_1/OP_0/OP_FALSE` which is why we have a policy flag for `MINIMALIF`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -432,6 +433,59 @@ object CSVScriptSignature extends ScriptFactory[CSVScriptSignature] {
...
+  }
+
+  def isFalse: Boolean = {
+    asm.last == OP_FALSE
+  }
+
+  def nestedScriptSig: ScriptSignature = {
```
Christewart, line 456: scaladoc

core/src/main/scala/org/bitcoins/core/script/constant/Constants.scala
```diff
@@ -191,6 +191,15 @@ sealed abstract class ScriptNumberOperation
     extends ScriptNumber
     with ScriptOperation {
   override def hex = opCode.toHexString
+
+  override def equals(obj: Any): Boolean = {
```
Christewart, line 198: scaladoc on why this is necessary

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -402,68 +411,43 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
-                                                  o,
-                                                  signers,
-                                                  redeemScriptOpt,
-                                                  None,
-                                                  hashType)
-                  loop(i +: t, currentLockTime)
-              }
-            } else {
-              loop(t, currentLockTime)
+      case spendingInfo :: newRemaining =>
```
Christewart: just a note: we should probably use `+:` for pattern matching on collections as I believe `::` can throw a exception for certain types of collections. 

I know `::` is used elsewhere in this file -- it is unsafe and depends on the scala runtime "deciding" on what collections to use. We should be moving to use `+:`

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -481,62 +465,40 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
-                                                  None,
-                                                  hashType)
-                  loop(i +: t, accum)
-              }
-            } else loop(t, accum)
-          case _: P2PKScriptPubKey | _: P2PKHScriptPubKey |
-              _: MultiSignatureScriptPubKey | _: P2WPKHWitnessSPKV0 |
-              _: NonStandardScriptPubKey | _: WitnessCommitment |
-              EmptyScriptPubKey | _: UnassignedWitnessScriptPubKey =>
+      case spendingInfo :: newRemaining =>
```
Christewart: same comment as above

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -458,3 +461,33 @@ sealed abstract class LockTimeSigner
   }
 }
 object LockTimeSigner extends LockTimeSigner
+
+sealed abstract class ConditionalSigner
```
Christewart, line 468: Scaladoc explaining what this does at a high level

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -57,6 +58,16 @@ sealed abstract class UTXOSpendingInfo {
   def redeemScriptOpt: Option[ScriptPubKey]
 
   def scriptWitnessOpt: Option[ScriptWitness]
+
+  def conditionalPath: ConditionalPath
+}
+
+sealed trait ConditionalPath
```
Christewart, line 77: Scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -57,6 +58,16 @@ sealed abstract class UTXOSpendingInfo {
...
+
+object ConditionalPath {
+  case object NoConditionsLeft extends ConditionalPath
```
Christewart, line 80: nit: maybe just call this `None`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -282,22 +323,64 @@ case class MultiSignatureSpendingInfo(
     hashType: HashType
 ) extends RawScriptUTXOSpendingInfo {
   require(signers.length >= scriptPubKey.requiredSigs, "Not enough signers!")
+
+  override def conditionalPath: ConditionalPath =
+    ConditionalPath.NoConditionsLeft
+}
+
+case class ConditionalSpendingInfo(
```
Christewart, line 344: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -57,6 +58,16 @@ sealed abstract class UTXOSpendingInfo {
   def redeemScriptOpt: Option[ScriptPubKey]
 
   def scriptWitnessOpt: Option[ScriptWitness]
+
+  def conditionalPath: ConditionalPath
```
Christewart, line 62: Why does this have to be defined on every `UTXOSpendingInfo`? It seems to be it should just need to be defined `ConditionalSpendingInfo`?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -49,6 +49,39 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
...
+      .map(_._1)
+
+    scriptSigGen.flatMap { scriptSig =>
+      NumberGenerator.bool.map { condition =>
+        ConditionalScriptSignature(scriptSig, condition)
+      }
+    }
+  }
+
+  def nonLockTimeConditionalScriptSignature: Gen[ConditionalScriptSignature] = {
```
Christewart, line 72: scaladoc why it is useful not to include locktime

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
Christewart: Let's definitely add a comment in the source code as to what this is... `"0151"` is a magic value (I'm guessing it is 1 byte to push onto the stack and then OP_TRUE)

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
nkohen: I'm not sure why we would ever use a `BitcoinScriptUtil.isTrue` as opposed to `== OP_TRUE`? Could you elaborate? (note that in this specific case it will certainly be either `OP_TRUE` or `OP_FALSE` by the invariant)

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
nkohen: I'm not sure what you mean?

This is only used by `randomNonLockTimeScriptSignature` which in turn is only used for the `CLTV` and `CSV` script signature generators in this file. (this is likely because TxBuilder does not currently handle nested locktimes and such correctly? idk why tbh)

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: Any spend with nesting (P2SH, P2WSH, Locktime, etc.) can contain a `ConditionalScriptPubKey` within it

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: So long-term (i.e. in the next PR) this will represent the leaf of a condition binary tree which is why I went with the name, I also think it makes more clear what kind of error message should be written if you are expecting more conditions (say deep in some tree).

Your call though, still think I should change it? (I don't have a strong opinion on this one)

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -57,6 +58,19 @@ sealed abstract class UTXOSpendingInfo {
...
+}
+
+/** Represents the spending branch being taken in a ScriptPubKey's execution
+  *
```
Christewart, line 66: Probably should mention that this is meant to be used in conjunction with `ConditionalScriptPubKey`, and specifically used for control flow in a Script and give an example with with a simple `OP_IF/OP_ELSE` control flow

PR #828, opened 2019-10-23
Signer TxSigComponent refactor

nkohen:
An experiment where I replace a bunch of signing parameters with a single `TxSigComponent`

Thoughts? Is this better?

(The first commit is #798)


nkohen:
Replaced by #830

PR #830, opened 2019-10-24
Signer UTXOSpendingInfo refactor

nkohen:
Replaces all the parameters to `Signer.sign` with just a `UTXOSpendingInfo`, `TxSigComponent`, and a `ScriptPubKey`. This gets rid of the need for `OverridesForNestedSigning` and the `Option`s involved there. It also makes things more extensible since in the future `UTXOSpendingInfo` will be a much nicer ADT and will allow us to pass which condition is being signed with on ConditionalSigner and such.

Replaces #828

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -6,7 +6,9 @@ import org.bitcoins.core.policy.Policy
 import org.bitcoins.core.protocol.script._
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.script.crypto.HashType
+import org.bitcoins.core.script.flag.ScriptFlag
 import org.bitcoins.core.wallet.builder.TxBuilderError
+import org.bitcoins.core.wallet.utxo.UTXOSpendingInfo
 import scodec.bits.ByteVector
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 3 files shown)


nkohen:
The first commit can be ignored as it is #798

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -16,50 +18,37 @@ sealed abstract class Signer {
...
     * @param isDummySignature - do not sign the tx for real, just use a dummy signature this is useful for fee estimation
     * @return
     */
   def sign(
-      signers: Seq[Sign],
-      output: TransactionOutput,
-      unsignedTx: Transaction,
-      inputIndex: UInt32,
-      hashType: HashType,
+      spendingInfo: UTXOSpendingInfo,
```
Christewart, line 27: So one concern I do have here is that it is possible that the `output` in `UTXOSpendingInfo` and the `output` in `TxSigComponent` could possible be different.

In general, it seems like `TxSigComponent` = `UTXOSpendingInfo` + partially/fully signed transaction that spends that `UTXOSpendingInfo`. Does that sound right? 

I'm not really sure how to handle them interacting with each other, and I don't really have concrete suggestions for making this better, but it does seem like that is a possible invalid state that could be represented currently -- a different output in `TxSigComponent` and `UTXOSpendingInfo`

nkohen: Good point, the only new thing `TxSigComponent` brought to the table was the `unsignedTx`, so I just replaced it with this and derived everything from the two.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -74,55 +63,57 @@ sealed abstract class Signer {
...
+            WitnessTransaction(btx.version,
+                               btx.inputs,
+                               btx.outputs,
+                               btx.lockTime,
+                               transactionWitness)
+          case wtx: WitnessTransaction => wtx
+        }
+
+        WitnessTxSigComponent(wtx, index, spendingInfo.output, flags)
+      case _: ScriptPubKey =>
```
Christewart, line 118: I don't think this accounts for the case of `P2SH(P2WSH())` does it? Shoouldn't that be a `WitnessTxSigComponentP2SH`?

PR #835, opened 2019-10-24
Rescan and fetch blocks

rorp:


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -429,12 +431,16 @@ case class ChainHandler(
...
           val startHeight = end - (batchSize - 1)
           val endHeight = end
-          val newAcc: Future[Vector[DoubleSha256DigestBE]] = for {
+          val newAcc = for {
             compactFilterDbs <- filterDAO.getBetweenHeights(startHeight,
                                                             endHeight)
-            groupSize = calcGroupSize(compactFilterDbs.size)
-            grouped = compactFilterDbs.grouped(groupSize)
-            filtered <- findMatches(grouped)
+            filtered <- if (compactFilterDbs.isEmpty)
```
Christewart: Do we really need this check here? Can't we just delegate this to the else clause and have `findMatches()` not work on a `Vector.empty`?

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
```diff
@@ -90,33 +121,68 @@ class NeutrinoNodeTest extends NodeUnitTest {
       startGenF.flatMap { _ =>
         //we should expect 5 headers have been announced to us via
         //the send headers message.
-        def has6BlocksF =
+        def hasBlocksF =
           RpcUtil.retryUntilSatisfiedF(conditionF = () => {
             node
               .chainApiFromDb()
-              .flatMap(_.getBlockCount.map(_ == 6))
+              .flatMap(_.getBlockCount.map(_ == 113))
```
Christewart: Can we mave `113` to something like 

```scala
val EXPECTED_BLOCK_COUNT = 113
```

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -160,14 +162,16 @@ trait Node extends P2PLogger {
     }
   }
 
+  def onStop(): Future[Unit]
```
Christewart: What are these? Scaladoc please

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -176,9 +180,11 @@ trait Node extends P2PLogger {
     } yield {
       peerMsgSenderF.map(_.sendGetHeadersMessage(hash.flip))
       logger.info(s"Starting sync node, height=${header.height} hash=$hash")
-    }
+    }).flatMap(_ => onSync())
   }
 
+  def onSync(): Future[Unit]
```
Christewart: What is this? Scaladoc please.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -176,9 +180,11 @@ trait Node extends P2PLogger {
     } yield {
       peerMsgSenderF.map(_.sendGetHeadersMessage(hash.flip))
       logger.info(s"Starting sync node, height=${header.height} hash=$hash")
-    }
+    }).flatMap(_ => onSync())
```
Christewart: I don't think this actually guarantees this callback is called when we are synced. So this is a little misleading IMO. Basically this should be called `onStartSync()`.

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -76,4 +76,7 @@ case class SpvNode(
     }
   }
 
+  override def onStop(): Future[Unit] = Future.successful(())
```
Christewart: nit: `FutureUtil.unit`

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
```diff
@@ -90,33 +121,68 @@ class NeutrinoNodeTest extends NodeUnitTest {
...
-          _ <- has6FiltersF
+          _ <- hasBlocksF
+          _ <- hasFilterHeadersF
+          _ <- hasFiltersF
         } yield succeed
       }
   }
 
+  it must "download a block that matches a compact block filter" taggedAs (UsesExperimentalBitcoind) in {
+    nodeConnectedWithBitcoind: NeutrinoNodeFundedWalletBitcoind =>
```
Christewart, line 152: Let's start using the bitcoin-s wallet when we can!

```scala
     val node = nodeConnectedWithBitcoind.node
      val bitcoind = nodeConnectedWithBitcoind.bitcoindRpc
      val bitcoinsWallet = nodeConnectedWithBitcoind.wallet
      var cancelable: Option[Cancellable] = None

      for {
        addressFromBitcoins <- bitcoinsWallet.getNewAddress()
        _ = addressFromBitcoindP.success(addressFromBitcoins)
        ...
```

Christewart (changes requested):

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
rorp: We do. `Iterable.grouped()` throws if group size is less than 1. Also there's no reason to allocate so many entities for an empty DB response.

PR #838, opened 2019-10-26
Refactor compact filter matching

rorp:
Refactored filter matching code and added simple sequential matching algorithm, which works almost 2 time faster in rescan.


core/src/main/scala/org/bitcoins/core/gcs/GCS.scala
```diff
@@ -192,22 +192,34 @@ object GCS {
...
+
+    res
+  }
+
+  /**
+    * Decodes all hashes while the given predicate returns true
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#set-queryingdecompression]]
+    */
+  def golombDecodeSetsWithPredicate(encodedData: BitVector, p: UInt8)(
+      predicate: UInt64 => Boolean): Unit = {
```
nkohen: I think it would be cleaner to have this return a `Vector[UInt64]` so that users aren't forced to side-effect in the `predicate` like you do above.

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -46,7 +46,98 @@ case class GolombFilter(
     n.bytes ++ encodedData.bytes
   }
 
-  // TODO: Offer alternative that stops decoding when it finds out if data is there
+  def hashToRange(item: ByteVector): UInt64 = GCS.hashToRange(item, f, key)
+
+}
+
+trait BlockFilterMatcher {
```
nkohen: New file please

nkohen: Also I believe this can be `sealed`

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -46,7 +46,98 @@ case class GolombFilter(
...
+    */
+  def matches(data: ByteVector): Boolean
+
+  /**
+    * Checks if the underlying filter matches any item from the given collection
+    */
+  def matchesAny(data: Vector[ByteVector]): Boolean
+}
+
+class SimpleFilterMatcher(filter: GolombFilter) extends BlockFilterMatcher {
```
nkohen: `case class`

nkohen: (this gives us a nice extractor method for free)

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
```diff
@@ -46,7 +46,98 @@ case class GolombFilter(
...
+    }
+
+    GCS.golombDecodeSetsWithPredicate(filter.encodedData, filter.p)(predicate)
+
+    matches
+  }
+
+}
+
+class BinarySearchFilterMatcher(filter: GolombFilter)
```
nkohen: Do we need this for any reason?

nkohen: Although I think it would be interesting to have a hybrid Matcher that cached when it was done in the above case and used Binary search when the cache already had everything you needed, if this slows things down too much though I'd say ditch it

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+    * Returns all ScriptPubKeys from a Block's inputs that are relevant
+    * to BIP 158 Basic Block Filters
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#contents]]
+    */
+  /*
+  TODO uncomment and add unit tests for this method
+  def getInputScriptPubKeysFromBlock(
```
Christewart: Is this not needed anymore? If so, let's remove?

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.bitcoins.core.crypto.DoubleSha256Digest
+import org.bitcoins.core.protocol.CompactSizeUInt
+import org.bitcoins.core.protocol.blockchain.Block
+import org.bitcoins.core.protocol.script.{EmptyScriptPubKey, ScriptPubKey}
+import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutput}
+import org.bitcoins.core.script.control.OP_RETURN
+import org.bitcoins.core.util.BitcoinSUtil
+import scodec.bits.ByteVector
+
+object BlockFilter {
```
Christewart, line 12: Shouldn't we be `extend Factory[BlockFilter]`? 

Isn't there a way to read a `BlockFilter` from strictly binary?

core/src/main/scala/org/bitcoins/core/gcs/BlockFilterMatcher.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+    * Checks if the underlying filter matches the given data
+    */
+  def matches(data: ByteVector): Boolean
+
+  /**
+    * Checks if the underlying filter matches any item from the given collection
+    */
+  def matchesAny(data: Vector[ByteVector]): Boolean
```
Christewart, line 18: Couldn't we implement this here? I.e. 

```scala
def matchesAny(data: Vector[ByteVector]): Boolean = {
  data.exists(d => matches(d))
}
```

Christewart (commented):
```
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:5:46: Unused import
[warn] import org.bitcoins.core.protocol.blockchain.Block
[warn]                                              ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:6:43: Unused import
[warn] import org.bitcoins.core.protocol.script.{EmptyScriptPubKey, ScriptPubKey}
[warn]                                           ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:6:62: Unused import
[warn] import org.bitcoins.core.protocol.script.{EmptyScriptPubKey, ScriptPubKey}
[warn]                                                              ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:7:48: Unused import
[warn] import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutput}
[warn]                                                ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:7:61: Unused import
[warn] import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutput}
[warn]                                                             ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:9:41: Unused import
[warn] import org.bitcoins.core.script.control.OP_RETURN
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:10:32: Unused import
[warn] import org.bitcoins.core.util.{BitcoinSUtil, CryptoUtil}
[warn]                                ^
[warn] /home/chris/dev/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala:13:25: Unused import
[warn] import scala.annotation.tailrec
[warn]                         ^

```

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
rorp: `object BlockFilter` is the only `BlockFilter`, I'm not sure if we need a factory for it.

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
rorp: I don't know I commented them because they don't have test coverage to make `core` buildable.

core/src/main/scala/org/bitcoins/core/gcs/BlockFilterMatcher.scala
rorp: What would be a benefit? The current recommended implementation doesn't work this way. And we don't know how other filter types will be implemented.

core/src/main/scala/org/bitcoins/core/gcs/GolombFilter.scala
rorp: We'll probably need some more performance testing before deprecating it.

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
Christewart: cc @nkohen

core/src/main/scala/org/bitcoins/core/gcs/BlockFilterMatcher.scala
Christewart: Normal "good" software engineering practices. It seems to me the only difference between `matchesAny` and `matches` is that `matchesAny()` is a `Vector`. Therefore we just need to iterate through and call `matches()` ? 

Also, with `.exists()` i believe we exit early if we match something, so we don't have to evaluate the entire sequence of elements

>The current recommended implementation doesn't work this way. And we don't know how other filter types will be implemented.

That's interesting, but I guess my understanding is that we just need to iterate through each of the elements in the Vector and see if _any_ of them match? 

If that is wrong, maybe we should add a scaladoc indicating what the goal of this method is as I might be misunderstanding

core/src/main/scala/org/bitcoins/core/gcs/BlockFilter.scala
nkohen: If this is never used I think it is fine to delete along with `TempUtxoProvider.scala` (it was all intended to be temporary anyway)

core/src/main/scala/org/bitcoins/core/gcs/BlockFilterMatcher.scala
rorp: I'd rather leave `BlockFilterMatcher` trait fully abstract, so that we don't impose the algorithm on the future implementations.

PR #840, opened 2019-10-28
Spending Info ADT use

nkohen:
Replaced TxBuilder signAndAddInput match with UTXOSpendingInfo match

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
@@ -19,7 +19,19 @@ import org.bitcoins.core.script.locktime.LockTimeInterpreter
 import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.core.wallet.signer._
-import org.bitcoins.core.wallet.utxo.{BitcoinUTXOSpendingInfo, UTXOSpendingInfo}
+import org.bitcoins.core.wallet.utxo.{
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -244,167 +256,147 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
+    val input =
+      TransactionInput(outpoint, EmptyScriptSignature, sequence)
+
+    val updatedTx =
+      unsignedTx.updateInput(inputIndex, input)
+
+    val updatedOutput =
+      TransactionOutput(spendingInfo.amount, redeemScript)
+
+    val updatedUTXOInfo = BitcoinUTXOSpendingInfo(outpoint,
```
Christewart, line 365: Do we need to call this here? I would like to get rid of this if possible and directly call constructors on the subtype of `BitcoinUTXOSpendingInfo()`. I guess that would force a pattern match here then though?

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
nkohen: I think that this is the correct use of `BitcoinUTXOSpendingInfo.apply`, what would go here if I didn't use this would be exactly the match and construction defined in that method for reuse. I think there are times when you have a `ScriptPubKey` on your hands that you have the spending info for but don't have the elaborated type, and that is what the `apply` is for.

PR #842, opened 2019-10-28
Replace scriptPubKeyToSatisfy with spendingInfoToSatisfy in Signer

nkohen:
This PR pairs `Signer`s with `UTXOSpendingInfo`s with a type parameter and now has `Signer.sign` take in a `spendingInfoToSatisfy` rather than just the `ScriptPubKey` from that `UTXOSpendingInfo`. This change will make `OP_IF` signing possible since this requires more than just the `ScriptPubKey` be passed in delegation.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
@@ -21,14 +21,9 @@ import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.core.wallet.signer._
 import org.bitcoins.core.wallet.utxo.{
   BitcoinUTXOSpendingInfo,
-  LockTimeSpendingInfo,
-  MultiSignatureSpendingInfo,
-  P2PKHSpendingInfo,
-  P2PKSpendingInfo,
   P2SHNestedSegwitV0UTXOSpendingInfo,
   P2SHNoNestSpendingInfo,
   P2SHSpendingInfo,
-  SegwitV0NativeUTXOSpendingInfo,
   UTXOSpendingInfo,
   UnassignedSegwitNativeUTXOSpendingInfo
 }
@@ -265,38 +260,6 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
       val oldInput = unsignedTx.inputs(inputIndex.toInt)
 
       utxo match {
-        case p2pkInfo: P2PKSpendingInfo =>
-          P2PKSigner
-            .sign(p2pkInfo, unsignedTx, dummySignatures)
-            .map(_.transaction)
-        case p2pkhInfo: P2PKHSpendingInfo =>
-          P2PKHSigner
-            .sign(p2pkhInfo, unsignedTx, dummySignatures)
-            .map(_.transaction)
-        case multiSigInfo: MultiSignatureSpendingInfo =>
-          MultiSigSigner
-            .sign(multiSigInfo, unsignedTx, dummySignatures)
-            .map(_.transaction)
-        case lockTimeInfo: LockTimeSpendingInfo =>
-          LockTimeSigner
-            .sign(lockTimeInfo, unsignedTx, dummySignatures)
-            .map(_.transaction)
-        case segWitInfo @ SegwitV0NativeUTXOSpendingInfo(_,
-                                                         _,
-                                                         witnessSPK,
-                                                         _,
-                                                         _,
-                                                         _) =>
-          witnessSPK match {
-            case _: P2WPKHWitnessSPKV0 =>
-              P2WPKHSigner
-                .sign(segWitInfo, unsignedTx, dummySignatures)
-                .map(_.transaction)
-            case _: P2WSHWitnessSPKV0 =>
-              P2WSHSigner
-                .sign(segWitInfo, unsignedTx, dummySignatures)
-                .map(_.transaction)
-          }
         case p2shInfo: P2SHNoNestSpendingInfo =>
           signAndAddInputForP2SH(p2shInfo,
                                  unsignedTx,
@@ -336,10 +299,17 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
           }
         case _: UnassignedSegwitNativeUTXOSpendingInfo =>
           Future.fromTry(TxBuilderError.NoSigner)
+        case _: BitcoinUTXOSpendingInfo =>
+          BitcoinSigner
+            .sign(utxo, unsignedTx, dummySignatures)
+            .map(_.transaction)
       }
     }
   }
 
+  /** Signs P2SH(P2WSH) and non-segwit P2SH inputs by calling signAndAddInput on the altered
+    * transaction without P2SH nesting and then re-wrapping with the tweaked ScriptSignature
+    */
   private def signAndAddInputForP2SH(
       spendingInfo: P2SHSpendingInfo,
       unsignedTx: Transaction,
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -322,7 +322,7 @@ sealed abstract class P2WPKHSigner extends BitcoinSigner[P2WPKHV0SpendingInfo] {
       spendingInfoToSatisfy: P2WPKHV0SpendingInfo)(
       implicit ec: ExecutionContext): Future[TxSigComponent] = {
     if (spendingInfoToSatisfy != spendingInfo) {
-      Future.fromTry(TxBuilderError.NestedWitnessSPK)
+      Future.fromTry(TxBuilderError.WrongSigner)
     } else {
       val (_, output, inputIndex, hashType) =
         relevantInfo(spendingInfo, unsignedTx)
@@ -402,7 +402,7 @@ sealed abstract class P2WSHSigner extends BitcoinSigner[P2WSHV0SpendingInfo] {
       spendingInfoToSatisfy: P2WSHV0SpendingInfo)(
       implicit ec: ExecutionContext): Future[TxSigComponent] = {
     if (spendingInfoToSatisfy != spendingInfo) {
-      Future.fromTry(TxBuilderError.NestedWitnessSPK)
+      Future.fromTry(TxBuilderError.WrongSigner)
     } else {
       val (_, output, inputIndex, _) = relevantInfo(spendingInfo, unsignedTx)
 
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -244,163 +251,115 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
+              _,
+              amount,
+              scriptPubKey,
+              _,
+              hashType,
+              redeemWitnessScript,
+              _) =>
+          redeemWitnessScript match {
+            case _: P2WSHWitnessSPKV0 =>
+              signAndAddInputForP2SH(p2shNestedSegwitInfo,
```
Christewart: Don't we need to call 

```scala
val uwtx = WitnessTransaction.toWitnessTx(unsignedTx)
```

here?

Christewart: Actually it doesn't look like it, with the p2sh hack where we replace the output script?

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -244,163 +251,115 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
-
-                signedTxF
+      }
+    }
+  }
 
-              }
-            case None => Future.fromTry(TxBuilderError.NoRedeemScript)
-          }
+  private def signAndAddInputForP2SH(
```
Christewart, line 313: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -119,271 +130,248 @@ sealed abstract class Signer {
         BaseTxSigComponent(unsignedTx, index, spendingInfo.output, flags)
     }
   }
+
+  protected def updateScriptSigInSigComponent(
```
Christewart, line 137: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -119,271 +130,248 @@ sealed abstract class Signer {
...
+      unsignedTx: Transaction,
+      inputIndex: Int,
+      output: TransactionOutput,
+      signatureF: Future[ScriptSignature])(
```
Christewart: please name `scriptSignatureF` as I would expect `signatureF` to be a `Future[ECDigitalSignature]`

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -119,271 +130,248 @@ sealed abstract class Signer {
...
+      implicit ec: ExecutionContext): Future[BaseTxSigComponent] = {
```
Christewart, line 142: Why is this always a `BaseTxSigComponent`?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -8,13 +8,24 @@ import org.bitcoins.core.protocol.transaction._
...
+  UTXOSpendingInfo,
+  UnassignedSegwitNativeUTXOSpendingInfo
+}
 import scodec.bits.ByteVector
 
 import scala.concurrent.{ExecutionContext, Future}
 
 /** The class used to represent a signing process for a specific [[org.bitcoins.core.protocol.script.ScriptPubKey]] type */
-sealed abstract class Signer {
+sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
```
Christewart, line 28: Why are these contravariant? If I remove the `-` from `Signer` and `BitcoinSigner` things still seem to compile

https://docs.scala-lang.org/tour/variances.html

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -119,271 +130,248 @@ sealed abstract class Signer {
...
   override def sign(
       spendingInfo: UTXOSpendingInfo,
       unsignedTx: Transaction,
       isDummySignature: Boolean,
-      scriptPubKeyToSatisfy: ScriptPubKey)(
+      spendingInfoToSatisfy: P2WPKHV0SpendingInfo)(
       implicit ec: ExecutionContext): Future[TxSigComponent] = {
-    if (scriptPubKeyToSatisfy != spendingInfo.output.scriptPubKey) {
+    if (spendingInfoToSatisfy != spendingInfo) {
       Future.fromTry(TxBuilderError.NestedWitnessSPK)
```
Christewart: is this this right error here still? Shouldn't it be like `TxBuilderError.WrongSigner`?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -119,271 +130,248 @@ sealed abstract class Signer {
...
-                } else Future.successful(p2wpkh)
-              case _: P2PKScriptPubKey | _: P2PKHScriptPubKey |
-                  _: MultiSignatureScriptPubKey | _: P2SHScriptPubKey |
-                  _: P2WSHWitnessSPKV0 | _: NonStandardScriptPubKey |
-                  _: CLTVScriptPubKey | _: CSVScriptPubKey |
-                  _: WitnessCommitment | EmptyScriptPubKey |
-                  _: UnassignedWitnessScriptPubKey =>
-                Future.fromTry(TxBuilderError.NonWitnessSPK)
-            }
+          val sign = signers.head.signFunction
```
Christewart:
```scala
          val sign = spendingInfoToSatisfy.signer
          val pubKey = sign.publicKey
```
i think?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -394,133 +382,75 @@ sealed abstract class P2WPKHSigner extends BitcoinSigner {
...
   override def sign(
       spendingInfo: UTXOSpendingInfo,
       unsignedTx: Transaction,
       isDummySignature: Boolean,
-      scriptPubKeyToSatisfy: ScriptPubKey)(
+      spendingInfoToSatisfy: P2WSHV0SpendingInfo)(
       implicit ec: ExecutionContext): Future[TxSigComponent] = {
-    if (scriptPubKeyToSatisfy != spendingInfo.output.scriptPubKey) {
+    if (spendingInfoToSatisfy != spendingInfo) {
       Future.fromTry(TxBuilderError.NestedWitnessSPK)
```
Christewart: Same comment as above wrt to error message

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -275,4 +483,12 @@ case class P2SHNestedSegwitV0UTXOSpendingInfo(
   override val redeemScriptOpt: Option[ScriptPubKey] = Some(redeemScript)
 
   override val scriptWitnessOpt: Option[ScriptWitnessV0] = Some(scriptWitness)
+
+  override def nestedSpendingInfo: SegwitV0NativeUTXOSpendingInfo =
```
Christewart: nit: can be a `lazy val` or `val`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -239,6 +243,19 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
     )
   }
 
+  def nonWitnessScriptPubKey: Gen[(NonWitnessScriptPubKey, Seq[ECPrivateKey])] = {
+    Gen.oneOf(
+      p2pkScriptPubKey.map(privKeyToSeq),
+      p2pkhScriptPubKey.map(privKeyToSeq),
+      multiSigScriptPubKey,
+      emptyScriptPubKey,
+      cltvScriptPubKey,
```
Christewart: nit: use `lockTimeScriptPubKey`

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
nkohen: If something can sign given a more generic type then it is more specific. Marking it as contra-variant lets us use this fact, and the fact that it compiles with the `-` is just proof that `Signer` is contra-variant in its parameter.

To make things explicit with an example:
`Signer[P2SHNestedSegwitSpendingInfo]` is the type for all things that can sign p2sh(segwit).
`Signer[P2SHNoNestSpendingInfo]` is the type for all things that can sign p2sh(raw).
`Signer[P2SHSpendingInfo]` is the type for all things that can sign both and hence it is a subtype of those two.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
nkohen: yeah, `spendingInfoToSatisfy != spendingInfo` => Nesting

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: I've actually been wrestling with this a lot, I'm starting to think that these should be `val`s so that you get an error right away if anything is wrong, rather than `lazy` or `def` which is perhaps more performant, thoughts?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
nkohen: because that is what this constructs, the two witness signers do their own funky stuff while everything else uses this so I figured I'd factor it out

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
nkohen: Yeah, shouldn't need to do anything special since p2sh recurses on this external function

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
Christewart: How do you know that `spendingInfo` is a witness (specifically `SegwitV0NativeUTXOSpendingInfo`)? The error says `nestedWitnessSPK`

nkohen: If they weren't equal then that means this is a nested call that was made via a match on the nested info and this is a witness signer

PR #843, opened 2019-10-28
RawScriptPubKey

nkohen:
Further elaborated the `NonWitnessScriptPubKey` type with a subtype `RawScriptPubKey` which excludes `P2SH`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -26,8 +26,10 @@ import scala.util.{Failure, Success, Try}
   */
 sealed abstract class ScriptPubKey extends Script
 
+/** Trait for all Non-SegWit ScriptPubKeys  */
 sealed trait NonWitnessScriptPubKey extends ScriptPubKey
 
+/** Trait for all raw, non-nested ScriptPubKeys (no P2SH) */
 sealed trait RawScriptPubKey extends NonWitnessScriptPubKey
 
 /**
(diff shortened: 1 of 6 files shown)


core-test/src/test/resources/logback-test.xml
```diff
@@ -1,7 +1,7 @@
 <configuration>
     <include resource="common-logback.xml" />
 
-    <root level="OFF">
+    <root level="INFO">
```
Christewart: ping

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -26,19 +26,24 @@ import scala.util.{Failure, Success, Try}
   */
 sealed abstract class ScriptPubKey extends Script
 
+sealed trait NonWitnessScriptPubKey extends ScriptPubKey
```
Christewart, line 30: quick scaladocs please

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -565,33 +574,61 @@ object NonStandardScriptPubKey extends ScriptFactory[NonStandardScriptPubKey] {
...
+case object EmptyScriptPubKey extends RawScriptPubKey {
   override def asm: Seq[ScriptToken] = Vector.empty
 }
 
-/** Factory companion object used to create
-  * [[org.bitcoins.core.protocol.script.ScriptPubKey ScriptPubKey]] objects */
-object ScriptPubKey extends ScriptFactory[ScriptPubKey] {
-  def empty: ScriptPubKey = fromAsm(Nil)
+object RawScriptPubKey extends ScriptFactory[RawScriptPubKey] {
+  def empty: RawScriptPubKey = fromAsm(Nil)
```
Christewart: nit: Let's change this to a `val`

PR #851, opened 2019-11-04
Fix getsentinfo Eclair Rpc call

rorp:


nkohen (approved):
ACK assuming tests pass

PR #854, opened 2019-11-06
Rescan RPC

rorp:
Implements the wallet rescan RPC call and the corresponding CLI command


app/cli/src/main/scala/org/bitcoins/cli/Cli.scala
```diff
@@ -70,6 +75,41 @@ object Cli extends App {
         .hidden()
         .action((_, conf) => conf.copy(command = GetBestBlockHash))
         .text(s"Get the best block hash"),
+      cmd("rescan")
```
Christewart, line 78: Can we define a `help` message on this? https://github.com/scopt/scopt#help-version-and-notes

app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
```diff
@@ -4,16 +4,37 @@ import akka.actor.ActorSystem
...
     extends ServerRoute {
+  import system.dispatcher
   implicit val materializer = ActorMaterializer()
 
   def handleCommand: PartialFunction[ServerCommand, StandardRoute] = {
     case ServerCommand("getpeers", _) =>
       complete {
         Server.httpSuccess("TODO implement getpeers")
       }
+    case ServerCommand("rescan", arr) =>
```
Christewart, line 21: I know there isn't any unit tests for this stuff yet, but can you please begin to implement them?

core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
```diff
@@ -1,13 +1,40 @@
...
+  case class BlockHash(hash: DoubleSha256DigestBE) extends BlockStamp {
+    override def mkString: String = hash.hex
+  }
+  case class BlockHeight(height: Int) extends BlockStamp {
+    require(height >= 0, "block height must be a positive number")
+    override def mkString: String = height.toString
+  }
+  case class BlockTime(time: UInt32) extends BlockStamp {
+    // TODO implement me
+    override def mkString: String = ???
```
Christewart: ping

Christewart (commented):
We you able to build the cli as a graal native image and try to test it out?

app/cli/src/main/scala/org/bitcoins/cli/CliReaders.scala
```diff
@@ -39,4 +41,25 @@ object CliReaders {
...
+            val time = ZonedDateTime.of(year.toInt,
+                                        month.toInt,
+                                        day.toInt,
+                                        0,
+                                        0,
+                                        0,
+                                        0,
+                                        ZoneId.of("UTC"))
+            BlockTime(time)
+          case _ => BlockStamp.fromString(str).get
```
Christewart, line 62: Will this blow up if the `str` is not well formatted?

app/server-test/src/test/scala/org/bitcoins/server/MockObjects.scala
```diff
@@ -0,0 +1,209 @@
...
+import org.bitcoins.node.SpvNodeCallbacks
+import org.bitcoins.node.config.NodeAppConfig
+import org.bitcoins.node.models.Peer
+import org.bitcoins.wallet.api._
+import org.bitcoins.wallet.config.WalletAppConfig
+import org.bitcoins.wallet.models.{AccountDb, AddressDb, SpendingInfoDb}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object MockObjects {
```
Christewart: Is there any way we can just integrate this wit hthe `Fixture` system rather than creating these mock objects with a bunch of unimplemented methods?

rorp: @Christewart I don't think we need an integration test just to verify that the URI and form body parser works as expected. I can replace `???` with something else if it's an issue.

Christewart (commented):
One compiler warning: 

```
[warn] /home/chris/dev/bitcoin-s-core/app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala:37:28: match may not be exhaustive.
[warn] It would fail on the following inputs: Arr(_), False, Null, Obj(_), True
[warn]       nullToOpt(value).map {
[warn]                            ^
[warn] one warning found
```

PR #855, opened 2019-11-07
ScriptInterpreter Conditional Refactor

nkohen:
Rewrote all conditional handling so that we no longer use binary trees and instead, do something even better than what core currently does! Because as @benthecarman pointed out to me, this exists: https://github.com/bitcoin/bitcoin/pull/16902 and makes our `ScriptInterpreterTest` run faster because O(1) handling is good :)

In my personal opinion this is much simpler than what was being done and I'd also found bugs in the previous OP_IF handling on a local branch that is being blocked by this bug. (Tests for this will be added for free on that branch because it introduces nested conditionals which are then run through the ScriptInterpreter)

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
@@ -66,15 +66,16 @@ class ScriptProgramFactoryTest extends BitcoinSUnitTest {
       ScriptFlagFactory.empty)
     val program = PreExecutionScriptProgram(t)
     val inProgress =
-      ExecutionInProgressScriptProgram(t,
-                                       stack,
-                                       script,
-                                       Nil,
-                                       Nil,
-                                       Nil,
-                                       None,
-                                       0,
-                                       0)
+      ExecutionInProgressScriptProgram(
+        txSignatureComponent = t,
+        stack = stack,
+        script = script,
+        originalScript = Nil,
+        altStack = Nil,
+        flags = Nil,
+        lastCodeSeparator = None,
+        conditionalCounter = ConditionalCounter.empty
+      )
     inProgress.stack must be(stack)
     inProgress.script must be(script)
   }
--- a/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
@@ -73,7 +73,8 @@ class ScriptProgramFactoryTest extends BitcoinSUnitTest {
                                        Nil,
                                        Nil,
                                        None,
-                                       Vector.empty)
+                                       0,
+                                       0)
     inProgress.stack must be(stack)
     inProgress.script must be(script)
   }
--- a/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
@@ -66,7 +66,14 @@ class ScriptProgramFactoryTest extends BitcoinSUnitTest {
       ScriptFlagFactory.empty)
     val program = PreExecutionScriptProgram(t)
     val inProgress =
-      ExecutionInProgressScriptProgram(t, stack, script, Nil, Nil, Nil, None)
+      ExecutionInProgressScriptProgram(t,
+                                       stack,
+                                       script,
+                                       Nil,
+                                       Nil,
+                                       Nil,
+                                       None,
+                                       Vector.empty)
     inProgress.stack must be(stack)
     inProgress.script must be(script)
   }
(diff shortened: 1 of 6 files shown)


core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
```diff
@@ -66,7 +66,15 @@ class ScriptProgramFactoryTest extends BitcoinSUnitTest {
       ScriptFlagFactory.empty)
     val program = PreExecutionScriptProgram(t)
     val inProgress =
-      ExecutionInProgressScriptProgram(t, stack, script, Nil, Nil, Nil, None)
+      ExecutionInProgressScriptProgram(t,
```
Christewart: Let's start using named parameters when we touch this now

```scala
                                       Nil,
                                       Nil,
                                       Nil,
                                       None,
                                       0,
                                       0
```

is not very descriptive

core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -143,18 +154,26 @@ case class ExecutionInProgressScriptProgram(
     originalScript: List[ScriptToken],
     altStack: List[ScriptToken],
     flags: Seq[ScriptFlag],
-    lastCodeSeparator: Option[Int])
+    lastCodeSeparator: Option[Int],
+    trueCount: Int,
```
Christewart: Can we just make this a type please? `ConditionalCounter` or something like that?

core/src/main/scala/org/bitcoins/core/script/control/ControlOperationsInterpreter.scala
```diff
@@ -22,111 +20,104 @@ sealed abstract class ControlOperationsInterpreter {
...
             "Script top was not OP_NOTIF")
-    val minimalIfEnabled = ScriptFlagUtil.minimalIfEnabled(program.flags)
-    val sigVersion = program.txSignatureComponent.sigVersion
-    val oldStackTop = program.stack.headOption
 
-    if (isNotMinimalStackTop(oldStackTop, sigVersion, minimalIfEnabled)) {
-      //need to duplicate minimal check, we cannot accurately invert the stack
-      //top for OP_IF otherwise
-      program.failExecution(ScriptErrorMinimalIf)
+    if (program.isInExecutionBranch) {
```
Christewart: nit: You might be clever enough to figure this out:

Can we invert the stack top, and the call `OP_IF` code? `OP_NOTIF` is just the inverted version of `OP_IF` right?

core-test/src/test/scala/org/bitcoins/core/script/ScriptProgramFactoryTest.scala
Christewart: I guess this is like the only place you didn't do it. So basically this is just a nit

core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
```diff
@@ -135,6 +138,14 @@ sealed trait StartedScriptProgram extends ScriptProgram
   * evaluated by the [[org.bitcoins.core.script.interpreter.ScriptInterpreter ScriptInterpreter]].
   *
   * @param lastCodeSeparator The index of the last [[org.bitcoins.core.script.crypto.OP_CODESEPARATOR OP_CODESEPARATOR]]
+  * @param trueCount The depth of OP_IFs/OP_NOTIFs we've entered on the true condition before the first false.
+  * @param falseAndIgnoreCount The depth of OP_IFs/OP_NOTIFs we've entered after and including the first false condition.
```
Christewart: Is there any invariants we can add around this? I think I get what it is suppose to do conceptually, but it seems like every time we encounter a `OP_IF/NOT_IF` one of these numbers should get incremented?

core/src/main/scala/org/bitcoins/core/script/ScriptProgram.scala
Christewart: Also then we can link to the PR that sipa implements this on, which is extremely useful reading

Christewart (commented):
So in general things LGTM, just a few software structuring suggestions. It looks correct AFAICT.

core/src/main/scala/org/bitcoins/core/script/control/ControlOperationsInterpreter.scala
nkohen: While this is certainly doable and not even hard. It turns out not to clean anything up since before passing over to `opIf`, `opNotIf` must first check `program.isInExecutionBranch` (since we don't want to invert stack top if not) as well as checking for `MinimalIf` because if it is not minimally encoded we can't invert it into something that is without avoiding a necessary ScriptError.

All of this means that the only code that gets replaced is:
```
else {
            val stackTopFalse: Boolean = program.stackTopIsFalse
            logger.debug(s"OP_IF stack top was $stackTopFalse")
            logger.debug(s"Stack top: ${program.stack}")

            program
              .updateStackAndScript(program.stack.tail, program.script.tail)
              .addCondition(stackTopFalse)
          }
```
and I don't think that replacing that with a stackTop inversion followed by a call to `opIf` is cleaner

nkohen: That being said, I was able to factor out the common code

PR #857, opened 2019-11-07
Nested OP_IF ScriptPubKey and signing

nkohen:
Implemented parsing for generally nested conditionals as well as re-writing `ConditionalPath` to support general branch spending.

(This is built on top #855 which fixes bugs in ScriptInterpreter that this PR requires. Only the last two commits are new to this PR)

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -35,7 +35,11 @@ sealed abstract class ScriptPubKey extends Script
 /** Trait for all Non-SegWit ScriptPubKeys  */
 sealed trait NonWitnessScriptPubKey extends ScriptPubKey
 
-/** Trait for all raw, non-nested ScriptPubKeys (no P2SH) */
+/** Trait for all raw, non-nested ScriptPubKeys (no P2SH)
+  *
+  * Note that all WitnessScriptPubKeys including P2WPKH are not
+  * considered to be non-nested and hence not RawScriptPubKeys.
+  */
 sealed trait RawScriptPubKey extends NonWitnessScriptPubKey
 
 /**
@@ -568,11 +572,11 @@ sealed trait ConditionalScriptPubKey extends RawScriptPubKey {
   require(asm.last.equals(OP_ENDIF),
           "ConditionalScriptPubKey must end in OP_ENDIF")
 
-  val (validConditional: Boolean, opElseIndex: Int) = {
+  val (isValidConditional: Boolean, opElseIndex: Int) = {
     ConditionalScriptPubKey.isConditionalScriptPubKeyWithElseIndex(asm)
   }
 
-  require(validConditional, "Must be valid ConditionalScriptPubKey syntax")
+  require(isValidConditional, "Must be valid ConditionalScriptPubKey syntax")
   require(opElseIndex != -1,
           "ConditionalScriptPubKey has to contain OP_ELSE asm token")
 
@@ -623,6 +627,9 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
     fromAsm(asm)
   }
 
+  /** Validates the correctness of the conditional syntax.
+    * If valid, also returns the index of the first outer-most OP_ELSE
+    */
   def isConditionalScriptPubKeyWithElseIndex(
       asm: Seq[ScriptToken]): (Boolean, Int) = {
     val headIsOpIf = asm.headOption.contains(OP_IF)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -557,17 +563,18 @@ object CSVScriptPubKey extends ScriptFactory[CSVScriptPubKey] {
...
-  require(opElseIndex != -1,
-          "ConditionalScriptPubKey has to contain OP_ELSE asm token")
   require(asm.last.equals(OP_ENDIF),
           "ConditionalScriptPubKey must end in OP_ENDIF")
-  require(asm.count(_.equals(OP_IF)) == 1,
-          "ConditionalScriptPubKey does not currently support nesting OP_IFs")
-  require(asm.count(_.equals(OP_ELSE)) == 1,
-          "ConditionalScriptPubKey does not currently support nesting OP_ELSEs")
+
+  val (validConditional: Boolean, opElseIndex: Int) = {
```
Christewart: nit: `isValidConditional`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -620,17 +623,67 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
     fromAsm(asm)
   }
 
-  def isConditionalScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
+  def isConditionalScriptPubKeyWithElseIndex(
```
Christewart, line 633: scaladoc

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -620,17 +623,67 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
...
+      asm: Seq[ScriptToken]): (Boolean, Int) = {
     val headIsOpIf = asm.headOption.contains(OP_IF)
-    lazy val containsOpElse = {
-      val opElseIndex = asm.indexOf(OP_ELSE)
-      opElseIndex != -1
+    lazy val endsWithEndIf = asm.last == OP_ENDIF
+
+    var opElseIndexOpt: Option[Int] = None
+
+    lazy val validConditionalTree = {
```
Christewart, line 640: so first thought before reviewing this implementation: Can't we we use the `ConditionalCounter` some how validate this? From my first look, it seems we are just making sure that every `OP_IF` has a matching `OP_ENDIF`?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -620,17 +623,67 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
...
+      // Already validate that the first token is OP_IF, go to tail
+      // and start with depth 1 and no OP_ELSE found for that depth
+      val opElsePendingOpt = asm.zipWithIndex.tail
```
Christewart, line 643: This seems to be more readable IMO

```scala
      val opElsePendingOpt = asm.zipWithIndex.tail
        .foldLeft[Option[Vector[Boolean]]](Some(Vector(false))) {
          case (None, _) => // Invalid tree case, do no computation
            None
          case (Some(Vector()),_) =>
            // Case of additional asm after final OP_ENDIF
            None

          case (Some(opElseFoundAtDepth), (OP_IF | OP_NOTIF, _)) =>
            // Increase depth by one with OP_ELSE yet to be found for this depth
            Some(opElseFoundAtDepth :+ false)
          case (Some(opElseFoundAtDepth), (OP_ELSE, asmIndex)) =>
            if (opElseFoundAtDepth == Vector(false)) {
              // If first OP_ELSE at depth 1, set opElseIndex
              opElseIndexOpt = Some(asmIndex)
            }

            if (opElseFoundAtDepth.last) {
              // If OP_ELSE already found at this depth, invalid
              None
            } else {
              // Otherwise, set to found at this depth
              Some(
                opElseFoundAtDepth.updated(opElseFoundAtDepth.length - 1,
                  true))
            }

          case (Some(opElseFoundAtDepth), (OP_ENDIF, _)) =>
            if (opElseFoundAtDepth.last) {
              // If OP_ELSE found at this depth then valid, decrease depth by 1
              Some(opElseFoundAtDepth.dropRight(1))
            } else {
              // Otherwise, invalid
              None
            }
          case (Some(opElseFoundAtDepth), (_, _)) =>
            // Token not related to conditional structure, ignore
            Some(opElseFoundAtDepth)
        }
```

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -67,19 +67,22 @@ sealed abstract class UTXOSpendingInfo {
...
   * then the remaining over-specified path will be ignored.
   *
-  * Note that we do not yet support nested conditionals.
-  *
   * For example, if you wanted to spend a ConditionalScriptPubKey(P2PK1, P2PK2)
   * (which looks like OP_IF <P2PK1> OP_ELSE <P2PK2> OP_ENDIF) with the P2PK1 case,
-  * then you would construct a ConditionalSpendingInfo using ConditionTrue as your
-  * ConditionalPath. Otherwise if you wanted to use P2PK2 you would use ConditionFalse.
+  * then you would construct a ConditionalSpendingInfo using conditionTrue as your
+  * ConditionalPath. Otherwise if you wanted to use P2PK2 you would use conditionFalse.
```
Christewart: nit: `P2PK2`

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -67,19 +67,22 @@ sealed abstract class UTXOSpendingInfo {
...
 object ConditionalPath {
   case object NoConditionsLeft extends ConditionalPath
-  case object ConditionTrue extends ConditionalPath
-  case object ConditionFalse extends ConditionalPath
+  case class ConditionTrue(nextCondition: ConditionalPath)
+      extends ConditionalPath
+  case class ConditionFalse(nextCondition: ConditionalPath)
+      extends ConditionalPath
+
+  val conditionTrue: ConditionalPath = ConditionTrue(NoConditionsLeft)
```
Christewart: I'm not sure if I understand these, are they just helpers for when we have no conditions left? Why do we need to wrap this in `ConditionTrue`/`ConditionFalse`?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -179,29 +226,43 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
...
-
   /** Creates a ConditionalScriptPubKey with keys for the true case */
-  def conditionalScriptPubKey: Gen[
-    (ConditionalScriptPubKey, Seq[ECPrivateKey])] = {
-    nonConditionalRawScriptPubKey.flatMap {
-      case (spk1, keys1) =>
-        nonConditionalRawScriptPubKey.map(_._1).map { spk2 =>
-          (ConditionalScriptPubKey(spk1, spk2), keys1)
-        }
+  def conditionalScriptPubKey(
```
Christewart, line 233: Add a `@param maxDepth` explaining what it is for

Christewart: This seems more readable to me 

```scala
  def conditionalScriptPubKey(
      maxDepth: Int): Gen[(ConditionalScriptPubKey, Seq[ECPrivateKey])] = {
    if (maxDepth > 0) {
      for {
        (trueSPK, keys1) <- rawScriptPubKey(maxDepth - 1)
        (falseSPK, _) <- rawScriptPubKey(maxDepth - 1)
      } yield (ConditionalScriptPubKey(trueSPK, falseSPK), keys1)
    } else {
        for {
          (trueSPK,keys1) <- nonConditionalRawScriptPubKey
          (falseSPK,_) <- nonConditionalRawScriptPubKey
        } yield (ConditionalScriptPubKey(trueSPK,falseSPK),keys1)
      }
  }
```

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -179,29 +226,43 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
...
-    (ConditionalScriptPubKey, Seq[ECPrivateKey])] = {
-    nonConditionalNonLocktimeRawScriptPubKey.flatMap {
-      case (spk1, keys1) =>
-        nonConditionalNonLocktimeRawScriptPubKey.map(_._1).map { spk2 =>
-          (ConditionalScriptPubKey(spk1, spk2), keys1)
-        }
+  def nonLocktimeConditionalScriptPubKey(
+      maxDepth: Int): Gen[(ConditionalScriptPubKey, Seq[ECPrivateKey])] = {
+    if (maxDepth > 0) {
+      nonLocktimeRawScriptPubKey(maxDepth - 1).flatMap {
```
Christewart: same as above with for expression rather than `.flatMap()`/`.map()`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -332,25 +383,40 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
     )
   }
 
+  def nonLocktimeRawScriptPubKey(
```
Christewart, line 381: technically isn't `P2WPKH` a raw script?

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: not sure which part you are referring to?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: We are also validating that there is only a single `OP_ELSE` at every level and that no asm appears after the top-level `OP_ENDIF`. For this reason, primarily the first one dealing with `OP_ELSE`, we cannot use `ConditionalCounter` as it would not be enough.

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: Better now with the new names?

PR #858, opened 2019-11-07
OP_NOTIF ScriptPubKey and Signing

nkohen:
Implements `OP_NOTIF` SPKs with full nesting and inter-op with `OP_IF`.

Built on top of #857 (only the last commit is new)

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -5,6 +5,7 @@ import org.bitcoins.core.crypto._
 import org.bitcoins.core.script.bitwise.{OP_EQUAL, OP_EQUALVERIFY}
 import org.bitcoins.core.script.constant.{BytesToPushOntoStack, _}
 import org.bitcoins.core.script.control.{
+  ConditionalOperation,
   OP_ELSE,
   OP_ENDIF,
   OP_IF,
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -682,11 +677,93 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
 
     lazy val opElseIndex = opElseIndexOpt.getOrElse(-1)
```
Christewart: probably would be better if this was an `Option[Int]` rather than a `-1`?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -682,11 +677,93 @@ object ConditionalScriptPubKey extends ScriptFactory[ConditionalScriptPubKey] {
...
+      trueSPK: RawScriptPubKey): NotIfConditionalScriptPubKey = {
+    val asm = Vector(OP_NOTIF) ++ falseSPK.asm ++ Vector(OP_ELSE) ++ trueSPK.asm ++ Vector(
+      OP_ENDIF)
+
+    fromAsm(asm)
   }
 
-  def isConditionalScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
-    isConditionalScriptPubKeyWithElseIndex(asm)._1
+  def isNotIfConditionalScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
```
Christewart, line 763: not that this should be done this PR, but we should probably just start calling these `isValid()` and add them as a def in `ScriptFactory`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -579,10 +590,14 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
...
       case (scriptSig, spk, keys) =>
-        rawScriptPubKey(defaultMaxDepth).map(_._1).map { spk2 =>
-          (ConditionalScriptSignature(scriptSig, true),
-           ConditionalScriptPubKey(spk.asInstanceOf[RawScriptPubKey], spk2),
-           keys)
+        conditionalOperation.flatMap { op =>
+          rawScriptPubKey(defaultMaxDepth).map(_._1).map { spk2 =>
+            (ConditionalScriptSignature(scriptSig, true),
+             ConditionalScriptPubKey(op,
+                                     spk.asInstanceOf[RawScriptPubKey],
```
Christewart: is this cast avoidable by using a diff generator?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: I think that could be nice :)

nkohen: Definitely won't clutter it here though

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
nkohen: Added a type parameter to `packageToSequenceOfPrivateKeys` which was the cause for this cast

PR #863, opened 2019-11-13
bitcoind v19 new RPC calls and tests

benthecarman:
First PR, this just adds the new RPC methods and tests for them. I am not sure how exhaustive the tests should be so I just put pretty basic ones in.

New RPCs

- `getblockfilter`

- `getbalances`

- `setwalletflag`

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
@@ -1,0 +1,62 @@
+package org.bitcoins.rpc.v19
+import org.bitcoins.rpc.client.common.BitcoindVersion
+import org.bitcoins.rpc.client.v19.BitcoindV19RpcClient
+import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
+import org.bitcoins.testkit.util.BitcoindRpcTest
+
+import scala.concurrent.Future
+
+class BitcoindV19RpcClientTest extends BitcoindRpcTest {
+  lazy val clientF: Future[BitcoindV19RpcClient] = {
+    val client = new BitcoindV19RpcClient(BitcoindRpcTestUtil.v19Instance())
+    val clientIsStartedF = BitcoindRpcTestUtil.startServers(Vector(client))
+    clientIsStartedF.map(_ => client)
+  }
+  lazy val clientPairF: Future[(BitcoindV19RpcClient, BitcoindV19RpcClient)] =
+    BitcoindRpcTestUtil.createNodePairV19(clientAccum)
+
+  clientF.foreach(c => clientAccum.+=(c))
+
+  behavior of "BitcoindV19RpcClient"
+
+  it should "be able to start a V19 bitcoind instance" in {
+
+    clientF.map { client =>
+      assert(client.version == BitcoindVersion.V19)
+    }
+
+  }
+
+  it should "get a block filter given a block hash" in {
+    for {
+      (client, _) <- clientPairF
+      blocks <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      blockFilter <- client.getBlockFilter(blocks.head.hex, "basic")
+    } yield {
+      assert(blockFilter.filter.length > 0)
+      assert(blockFilter.header.length > 0)
+    }
+  }
+
+  it should "be able to get the balances" in {
+    for {
+      (client, _) <- clientPairF
+      immatureBalance <- client.getBalances
+      _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      newImmatureBalance <- client.getBalances
+    } yield {
+      assert(immatureBalance.mine.immature.toBigDecimal > 0)
+      assert(immatureBalance.mine.immature < newImmatureBalance.mine.immature)
+    }
+  }
+
+  it should "be able to set the wallet flag 'avoid_reuse'" in {
+    for {
+      (client, _) <- clientPairF
+      result <- client.setWalletFlag("avoid_reuse", value = true)
+    } yield {
+      assert(result.flag == "avoid_reuse")
+      assert(result.state)
+    }
+  }
+}
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
@@ -38,14 +38,15 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
       block <- client.getBlockRaw(blocks.head)
       txs <- Future.sequence(
         block.transactions
-          .filter(!_.isCoinbase)
-          .map(x => client.getTransaction(x.txIdBE)))
+          .filterNot(_.isCoinbase)
+          .map(tx => client.getTransaction(tx.txIdBE)))
 
       prevFilter <- client.getBlockFilter(block.blockHeader.previousBlockHashBE, FilterType.Basic)
     } yield {
       val pubKeys = txs.flatMap(_.hex.outputs.map(_.scriptPubKey)).toVector
-      assert(BlockFilter(block, pubKeys).hash == blockFilter.filter.hash)
-      assert(blockFilter.header == BlockFilter(block, pubKeys).getHeader(prevFilter.header.flip).hash.flip)
+      val filter = BlockFilter(block, pubKeys)
+      assert(filter.hash == blockFilter.filter.hash)
+      assert(blockFilter.header == filter.getHeader(prevFilter.header.flip).hash.flip)
     }
   }
 
@@ -56,9 +57,10 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
       _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
       newImmatureBalance <- client.getBalances
     } yield {
+      val blockReward = 12.5
       assert(immatureBalance.mine.immature.toBigDecimal >= 0)
       assert(
-        immatureBalance.mine.immature.toBigDecimal + 12.5 == newImmatureBalance.mine.immature.toBigDecimal)
+        immatureBalance.mine.immature.toBigDecimal + blockReward == newImmatureBalance.mine.immature.toBigDecimal)
     }
   }
 
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
@@ -1,5 +1,7 @@
 package org.bitcoins.rpc.v19
+import org.bitcoins.core.gcs.{BlockFilter, FilterType}
 import org.bitcoins.rpc.client.common.BitcoindVersion
+import org.bitcoins.rpc.client.common.RpcOpts.WalletFlag
 import org.bitcoins.rpc.client.v19.BitcoindV19RpcClient
 import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
 import org.bitcoins.testkit.util.BitcoindRpcTest
@@ -31,10 +33,19 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
     for {
       (client, _) <- clientPairF
       blocks <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
-      blockFilter <- client.getBlockFilter(blocks.head.hex, "basic")
+      blockFilter <- client.getBlockFilter(blocks.head, FilterType.Basic)
+
+      block <- client.getBlockRaw(blocks.head)
+      txs <- Future.sequence(
+        block.transactions
+          .filter(!_.isCoinbase)
+          .map(x => client.getTransaction(x.txIdBE)))
+
+      prevFilter <- client.getBlockFilter(block.blockHeader.previousBlockHashBE, FilterType.Basic)
     } yield {
-      assert(blockFilter.filter.length > 0)
-      assert(blockFilter.header.length > 0)
+      val pubKeys = txs.flatMap(_.hex.outputs.map(_.scriptPubKey)).toVector
+      assert(BlockFilter(block, pubKeys).hash == blockFilter.filter.hash)
+      assert(blockFilter.header == BlockFilter(block, pubKeys).getHeader(prevFilter.header.flip).hash.flip)
     }
   }
 
@@ -45,18 +56,20 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
       _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
       newImmatureBalance <- client.getBalances
     } yield {
-      assert(immatureBalance.mine.immature.toBigDecimal > 0)
-      assert(immatureBalance.mine.immature < newImmatureBalance.mine.immature)
+      assert(immatureBalance.mine.immature.toBigDecimal >= 0)
+      assert(
+        immatureBalance.mine.immature.toBigDecimal + 12.5 == newImmatureBalance.mine.immature.toBigDecimal)
     }
   }
 
   it should "be able to set the wallet flag 'avoid_reuse'" in {
     for {
       (client, _) <- clientPairF
-      result <- client.setWalletFlag("avoid_reuse", value = true)
+      result <- client.setWalletFlag(WalletFlag.AvoidReuse, value = true)
     } yield {
-      assert(result.flag == "avoid_reuse")
-      assert(result.state)
+      assert(result.flag_name == "avoid_reuse")
+      assert(result.flag_state)
+      // TODO: Add test cases for when updated RPCs are added
     }
   }
 }
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.rpc.client.v19
 
 import akka.actor.ActorSystem
-import org.bitcoins.core.currency.Bitcoins
+import org.bitcoins.rpc.client.common.RpcOpts.WalletFlag
 import org.bitcoins.rpc.client.common.{BitcoindRpcClient, BitcoindVersion}
 import org.bitcoins.rpc.config.BitcoindInstance
 import org.bitcoins.rpc.jsonmodels.{GetBalancesResult, SetWalletFlagResult}
@@ -14,8 +14,6 @@ import scala.util.Try
 
 /**
   * Class for creating a BitcoindV19 instance that can access RPCs
-  * @param instance
-  * @param actorSystem
   */
 class BitcoindV19RpcClient(override val instance: BitcoindInstance)(
     implicit
@@ -26,16 +24,15 @@ class BitcoindV19RpcClient(override val instance: BitcoindInstance)(
   override lazy val version: BitcoindVersion = BitcoindVersion.V19
 
   /**
-    * setWalletFlag
-    *
     * Change the state of the given wallet flag for a wallet.
     */
   def setWalletFlag(
-      flag: String,
+      flag: WalletFlag,
       value: Boolean
   ): Future[SetWalletFlagResult] =
-    bitcoindCall[SetWalletFlagResult]("setwalletflag",
-                                      List(JsString(flag), Json.toJson(value)))
+    bitcoindCall[SetWalletFlagResult](
+      "setwalletflag",
+      List(JsString(flag.toString), Json.toJson(value)))
 
   def getBalances: Future[GetBalancesResult] = {
     bitcoindCall[GetBalancesResult]("getbalances")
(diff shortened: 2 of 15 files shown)


nkohen:
Would you mind listing the RPCs implemented in the description?

benthecarman:
> Would you mind listing the RPCs implemented in the description?

Added

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,26 @@
...
+/**
+  * Gets the BIP158 filter for the specified block.
+  * This RPC is only enabled if block filters have been created using the -blockfilterindex configuration option
+  * @see [[https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblockfilter]]
+  */
+trait V19BlockFilterRpc {
+  self: Client =>
+
+  def getBlockFilter(
+      blockhash: String,
```
nkohen: This can be `DoubleSha256DigestBE`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,26 @@
...
+      filtertype: String): Future[GetBlockFilterResult] = {
```
nkohen: I believe we have a `FilterType`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
```diff
@@ -202,3 +202,6 @@ case class GetTxOutSetInfoResult(
     disk_size: Int,
     total_amount: Bitcoins)
     extends BlockchainResult
+
+case class GetBlockFilterResult(filter: String, header: String)
```
nkohen: These can be `GolombFilter` and `DoubleSha256DigestBE`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala
```diff
@@ -229,6 +232,18 @@ object JsonSerializers {
...
 
+  implicit val setWalletFlagResultReads: Reads[SetWalletFlagResult] =
+    ((__ \ "flag_name").read[String] and
+      (__ \ "flag_state").read[Boolean] and
+      (__ \ "warnings").readNullable[String])(SetWalletFlagResult)
+
+  implicit val balanceInfoReads: Reads[BalanceInfo] =
+    ((__ \ "trusted").read[Bitcoins] and
+      (__ \ "untrusted_pending").read[Bitcoins] and
+      (__ \ "immature").read[Bitcoins])(BalanceInfo)
```
nkohen: I think that we should stay consistent and even use underscores in our result case classes rather than changing names and adding complexity. Really the only time where this should be used is if names are illegal like if they use the name `type` or any other reserved words

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -678,6 +696,15 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
     (BitcoindV18RpcClient, BitcoindV18RpcClient)] =
     createNodePairInternal(BitcoindVersion.V18, clientAccum)
 
+  /**
+    * Returns a pair of [[org.bitcoins.rpc.client.v19.BitcoindV19RpcClient BitcoindV19RpcClient]]
+    * that are connected with some blocks in the chain
+    */
+  def createNodePairV19(clientAccum: RpcClientAccum = Vector.newBuilder)(
```
nkohen, line 703: I believe you need to add a case to `createNodeSequence`.

You should get a warning for this on compile time. If it doesn't show up you can `clean` and then `compile`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
```diff
@@ -0,0 +1,77 @@
...
+import play.api.libs.json.JsString
+import org.bitcoins.rpc.serializers.JsonSerializers._
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+  * Class for creating a BitcoindV19 instance that can access RPCs
+  * @param instance
+  * @param actorSystem
```
nkohen: In general we just leave off the params unless we have something to say about them

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
```diff
@@ -0,0 +1,77 @@
...
+    with V19BlockFilterRpc {
+
+  override lazy val version: BitcoindVersion = BitcoindVersion.V19
+
+  /**
+    * setWalletFlag
+    *
+    * Change the state of the given wallet flag for a wallet.
+    */
+  def setWalletFlag(
```
nkohen, line 68: nit: I don't think you need to restate the name in the scaladoc

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
```diff
@@ -0,0 +1,77 @@
...
+      flag: String,
```
nkohen: Is there a fixed set of `flag`s? If so we should make an ADT for this.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/WalletResult.scala
```diff
@@ -50,6 +50,20 @@ case class GetTransactionResult(
     hex: Transaction)
     extends WalletResult
 
+case class SetWalletFlagResult(
+    flag: String,
```
nkohen: ditto comment above on type for this

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  it should "get a block filter given a block hash" in {
+    for {
+      (client, _) <- clientPairF
+      blocks <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      blockFilter <- client.getBlockFilter(blocks.head.hex, "basic")
+    } yield {
+      assert(blockFilter.filter.length > 0)
+      assert(blockFilter.header.length > 0)
```
nkohen: should probably assert that the block filter matches the block and the block filter header matches the block filter. Let me know if you want to know how these are computed.

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  it should "be able to get the balances" in {
+    for {
+      (client, _) <- clientPairF
+      immatureBalance <- client.getBalances
+      _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      newImmatureBalance <- client.getBalances
+    } yield {
+      assert(immatureBalance.mine.immature.toBigDecimal > 0)
+      assert(immatureBalance.mine.immature < newImmatureBalance.mine.immature)
```
nkohen: we probably know the exact difference here right?

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  it should "be able to get the balances" in {
+    for {
+      (client, _) <- clientPairF
+      immatureBalance <- client.getBalances
+      _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      newImmatureBalance <- client.getBalances
+    } yield {
+      assert(immatureBalance.mine.immature.toBigDecimal > 0)
```
nkohen: Let's make this `>=` just in case this ever moves to the top of the file

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  it should "be able to set the wallet flag 'avoid_reuse'" in {
+    for {
+      (client, _) <- clientPairF
+      result <- client.setWalletFlag("avoid_reuse", value = true)
+    } yield {
+      assert(result.flag == "avoid_reuse")
+      assert(result.state)
+    }
```
nkohen, line 75: is there any way we can functionally check that the flag is active?

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,62 @@
+package org.bitcoins.rpc.v19
+import org.bitcoins.rpc.client.common.BitcoindVersion
+import org.bitcoins.rpc.client.v19.BitcoindV19RpcClient
+import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
+import org.bitcoins.testkit.util.BitcoindRpcTest
+
+import scala.concurrent.Future
+
+class BitcoindV19RpcClientTest extends BitcoindRpcTest {
```
nkohen, line 11: I think we are missing an `afterAll` method here shutting things down

nkohen (commented):
Looks pretty good, just a couple minor things

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
benthecarman: There is only one currently `avoid_reuse`

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
benthecarman: I believe I did here https://github.com/bitcoin-s/bitcoin-s/blob/a9d896013cb6a61945f06182e95a3bffd22a3093/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala#L619

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
benthecarman: The flag will show up in some of the updated RPCs, I could add that test case when the I update those RPCs

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
benthecarman: None of the other `RpcClientTest`s have one, should I add one to them in a future PR or do I not need one?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
nkohen: Gotchya, well in any case I'd say it would be nice to have an ADT (Abstract Data Type) for this.

This will entail creating a
```sealed abstract class WalletFlag```
with just the one member for now, and usually for readability with things like this we put the instances inside a companion object:
```
object WalletFlag {
  case object AvoidReuse extends WalletFlag {
    override def toString: String = "avoid_reuse"
  }
}
```
Using the keyword `sealed` enforces that this `abstract class` can only be `extend`ed from within the same file. This gives us nice features like if you ever add a `WalletFlag`, you will get compiler warnings about all of the now non-exhaustive `match` statements where the new case needs to be handled.

File: bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala (hunk 13 above)
Comment (nkohen) [reply bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala:75]: Sounds good, maybe just leave a TODO comment for now and either make yourself a note or file yourself an issue to get back to this :+1:

File: bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala (hunk 14 above)
Comment (nkohen) [reply bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala:11]: @Christewart how is it that we don't have `afterAll`s in these tests? Am I missing something or is this definitely a problem?

Comment (Christewart) [reply bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala:11]: Take a look at the `BitcoindRpcTest` trait

File: bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala (hunk 3 above)
Comment (Christewart) [reply bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala]: is `header` a hash or a `BlockHeader`?

File: testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala [hunk 15]
```diff
@@ -229,12 +230,14 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
...
-      rpcUri,
-      zmqPort,
-      pruneMode,
-      blockFilterIndex = versionOpt.contains(BitcoindVersion.Experimental))
+    val configFile =
+      writtenConfig(uri,
+                    rpcUri,
+                    zmqPort,
+                    pruneMode,
+                    blockFilterIndex = versionOpt.contains(
```
Christewart: maybe move this up a few lines into a intermediate val and say 

```scala
val hasNeutrinoSupport = Vector(BitcoindVersion.Experimental, BitcoindVersion19).contains(version)
.
.
.
blockFilterIndex = hasNeutrinoSupport)
```

benthecarman: > is header a hash or a BlockHeader?

It's the header for the block filter 

https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#filter-headers

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
nkohen: It is a hash since we don’t have access to additional information, specifically the previous hash

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  it should "get a block filter given a block hash" in {
+    for {
+      (client, _) <- clientPairF
+      blocks <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      blockFilter <- client.getBlockFilter(blocks.head, FilterType.Basic)
+
+      block <- client.getBlockRaw(blocks.head)
+      txs <- Future.sequence(
+        block.transactions
+          .filter(!_.isCoinbase)
```
nkohen: nit: `.filterNot(_.isCoinbase)`

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,75 @@
...
+          .map(x => client.getTransaction(x.txIdBE)))
```
nkohen: nit: rename `x` to `tx` or something that is a little more descriptive

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,75 @@
...
+
+      prevFilter <- client.getBlockFilter(block.blockHeader.previousBlockHashBE, FilterType.Basic)
+    } yield {
+      val pubKeys = txs.flatMap(_.hex.outputs.map(_.scriptPubKey)).toVector
+      assert(BlockFilter(block, pubKeys).hash == blockFilter.filter.hash)
```
nkohen: Let's make `BlockFilter(block, pubKeys)` a `val` so it isn't computed twice

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  it should "be able to get the balances" in {
+    for {
+      (client, _) <- clientPairF
+      immatureBalance <- client.getBalances
+      _ <- client.getNewAddress.flatMap(client.generateToAddress(1, _))
+      newImmatureBalance <- client.getBalances
+    } yield {
+      assert(immatureBalance.mine.immature.toBigDecimal >= 0)
+      assert(
+        immatureBalance.mine.immature.toBigDecimal + 12.5 == newImmatureBalance.mine.immature.toBigDecimal)
```
nkohen: If we could give `12.5` a name (as a `val`) that would be nice for readability

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,44 @@
...
+trait V19BlockFilterRpc {
+  self: Client =>
+
+  private case class TempBlockFilterResult(
+      filter: String,
+      header: DoubleSha256DigestBE)
+  implicit private val tempBlockFilterResultReads: Reads[
+    TempBlockFilterResult] =
+    ((__ \ "filter").read[String] and
+      (__ \ "header").read[DoubleSha256DigestBE])(TempBlockFilterResult)
```
nkohen: You should be able to do a simple `Json.reads[TempBlockFilterResult]` here assuming `DoubleSha256DigestBEReads` is in scope

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def getBlockFilter(
+      blockhash: DoubleSha256DigestBE,
+      filtertype: FilterType): Future[GetBlockFilterResult] = {
+    val temp = Await.result(bitcoindCall[TempBlockFilterResult](
```
nkohen: Rather than doing `Await.result`, this should value should be `map`ped

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala
```diff
@@ -229,6 +229,13 @@ object JsonSerializers {
 
   implicit val bumpFeeReads: Reads[BumpFeeResult] = Json.reads[BumpFeeResult]
 
+  implicit val setWalletFlagResultReads: Reads[SetWalletFlagResult] =
+    Json.reads[SetWalletFlagResult];
+
+  implicit val balanceInfoReads: Reads[BalanceInfo] = Json.reads[BalanceInfo];
```
nkohen: nit: semi-colons not necessary :)

nkohen (commented):
Round 2 :) Getting close!

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,37 @@
...
+  implicit private val tempBlockFilterResultReads: Reads[
+    TempBlockFilterResult] = Json.reads[TempBlockFilterResult]
+
+  def getBlockFilter(
+      blockhash: DoubleSha256DigestBE,
+      filtertype: FilterType): Future[GetBlockFilterResult] = {
+    bitcoindCall[TempBlockFilterResult](
+      "getblockfilter",
+      List(JsString(blockhash.hex), JsString(filtertype.toString.toLowerCase)))
+      .map(t =>
```
nkohen: nit: more descriptive name would be nice. It is also nicer IMO to use `.map` with `{ }` but this is just an aesthetic choice

nkohen (approved):
LGTM, I'd like @Christewart to take another look before merging though :)

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/V19BlockFilterRpc.scala
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+  * Gets the BIP158 filter for the specified block.
+  * This RPC is only enabled if block filters have been created using the -blockfilterindex configuration option
+  * @see [[https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblockfilter]]
+  */
+trait V19BlockFilterRpc {
+  self: Client =>
+
+  private case class TempBlockFilterResult(
```
Christewart, line 25: Why is this needed? Can we add a comment here indicating why we need this temp result?

PR #865, opened 2019-11-14
Conditional Signing Tests

nkohen:
Added `ConditionalScriptPubKey`s to `CreditingTxGen` so that we are actually testing Conditional signing now (apparently we weren't before).

Found and fixed some bugs that had not been activate-able previously:

1. `BitcoinScriptUtil.minimalIfOp` which is used to turn `ScriptSignature`s into `P2WSH` `ScriptWintess`es only changed a single `OP_0` or `OP_1` into a `ScriptNumber` rather than all of them.
2. `P2SH` and `P2WSH` nested scripts cannot be serialized to > 520 bytes. This is now possible because conditionals allow multiple large multisigs to appear at once. I fixed this by adding `suchThat` calls on the affected generators

I will also note that this PR should increase coverage by a full `0.21%` according to my local measurements!

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
@@ -537,13 +537,11 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
   /** Since witnesses are not run through the interpreter, replace
     * `OP_0`/`OP_1` with `ScriptNumber.zero`/`ScriptNumber.one` */
   def minimalIfOp(asm: Seq[ScriptToken]): Seq[ScriptToken] = {
-    if (asm == Nil) asm
-    else if (asm.last == OP_0) {
-      asm.dropRight(1) ++ Seq(ScriptNumber.zero)
-    } else if (asm.last == OP_1) {
-      asm.dropRight(1) ++ Seq(ScriptNumber.one)
-    } else asm
-
+    asm.map {
+      case OP_0               => ScriptNumber.zero
+      case OP_1               => ScriptNumber.one
+      case token: ScriptToken => token
+    }
   }
 
   /** Replaces the [[org.bitcoins.core.script.constant.OP_0 OP_0]] dummy for
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
@@ -5,6 +5,7 @@ import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.protocol.script._
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.script.crypto.HashType
+import org.bitcoins.core.script.interpreter.ScriptInterpreter
 import org.bitcoins.core.wallet.utxo.{
   BitcoinUTXOSpendingInfo,
   ConditionalPath,
@@ -62,7 +63,7 @@ sealed abstract class CreditingTxGen {
              p2wpkhOutput,
              p2wshOutput)
       .suchThat(output =>
-        !ScriptGenerators.scriptPubKeyTooBig(output.scriptPubKey))
+        !ScriptGenerators.redeemScriptTooBig(output.scriptPubKey))
       .suchThat {
         case P2SHNestedSegwitV0UTXOSpendingInfo(_,
                                                 _,
@@ -72,7 +73,7 @@ sealed abstract class CreditingTxGen {
                                                 _,
                                                 witness: P2WSHWitnessV0,
                                                 _) =>
-          witness.stack.exists(_.length > 520)
+          witness.stack.exists(_.length > ScriptInterpreter.MAX_PUSH_SIZE)
         case _ => true
       }
   }
@@ -232,7 +233,7 @@ sealed abstract class CreditingTxGen {
   def p2wshOutput: Gen[BitcoinUTXOSpendingInfo] =
     nonP2WSHOutput
       .suchThat(output =>
-        !ScriptGenerators.scriptPubKeyTooBig(output.scriptPubKey))
+        !ScriptGenerators.redeemScriptTooBig(output.scriptPubKey))
       .flatMap {
         case BitcoinUTXOSpendingInfo(_, txOutput, signer, _, _, _, _) =>
           val spk = txOutput.scriptPubKey
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
@@ -1,12 +1,19 @@
 package org.bitcoins.testkit.core.gen
 
 import org.bitcoins.core.crypto.Sign
-import org.bitcoins.core.number.UInt32
+import org.bitcoins.core.number.{UInt32, UInt64}
+import org.bitcoins.core.protocol.CompactSizeUInt
 import org.bitcoins.core.protocol.script._
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.script.crypto.HashType
-import org.bitcoins.core.wallet.utxo.{BitcoinUTXOSpendingInfo, ConditionalPath}
+import org.bitcoins.core.wallet.utxo.{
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -537,13 +537,11 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
...
     * `OP_0`/`OP_1` with `ScriptNumber.zero`/`ScriptNumber.one` */
   def minimalIfOp(asm: Seq[ScriptToken]): Seq[ScriptToken] = {
-    if (asm == Nil) asm
-    else if (asm.last == OP_0) {
-      asm.dropRight(1) ++ Seq(ScriptNumber.zero)
-    } else if (asm.last == OP_1) {
-      asm.dropRight(1) ++ Seq(ScriptNumber.one)
-    } else asm
-
+    asm.map {
```
Christewart, line 540: sometimes things like this makes me :man_facepalming: at my old code....

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -22,15 +29,16 @@ sealed abstract class CreditingTxGen {
       Gen.listOfN(n, TransactionGenerators.realisticOutput)
     }
 
-  def rawOutput: Gen[BitcoinUTXOSpendingInfo] = {
+  def nonSHOutput: Gen[BitcoinUTXOSpendingInfo] = {
```
Christewart, line 33: scaladoc please, I assume `SH` is script-hash?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -40,15 +48,34 @@ sealed abstract class CreditingTxGen {
...
+  /** Only for use in constructing P2SH outputs */
+  private def nonP2SHOutput: Gen[BitcoinUTXOSpendingInfo] = {
+    Gen
+      .oneOf(p2pkOutput,
+             p2pkhOutput,
+             multiSigOutput, /*cltvOutput,*/ csvOutput,
+             conditionalOutput,
+             p2wpkhOutput,
+             p2wshOutput)
+      .suchThat(_.scriptPubKey.compactSizeUInt.toLong + CompactSizeUInt(
```
Christewart: 1. I think this is actually incorrect, this is checking if the _entire script_ is less than 520 bytes, where as the consesus rule is that largest _single element_ that can be pushed onto the [stack is 520 bytes](https://github.com/bitcoin-s/bitcoin-s/blob/d86acfffedda99ad73f6997fbf750f873f7ab2d3/core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala#L41). The largest _total script size_ is [10,000 bytes](https://github.com/bitcoin-s/bitcoin-s/blob/d86acfffedda99ad73f6997fbf750f873f7ab2d3/core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala#L436). It probably makes sense to make a helper method in `BitcoinScriptUtil.isValidStackElementSize()` or something like that. 

2. Another thing I generally don't like to use `.suchThat()` as it tends to generate a lot of generator failures where we cannot create a satisfiable `T` (in this case `BitcoinUTXOSpendingInfo`). 

I guess in this case it might be ok though, as it is pretty rare to generate a single stack element in the SPK with > 520 bytes in size.

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -40,15 +48,34 @@ sealed abstract class CreditingTxGen {
...
+      .suchThat {
+        case P2SHNestedSegwitV0UTXOSpendingInfo(_,
+                                                _,
+                                                _,
+                                                _,
+                                                _,
+                                                _,
+                                                witness: P2WSHWitnessV0,
+                                                _) =>
+          witness.redeemScript.compactSizeUInt.toLong + CompactSizeUInt(
```
Christewart: I think the same thing here applies https://github.com/bitcoin-s/bitcoin-s/blob/d86acfffedda99ad73f6997fbf750f873f7ab2d3/core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala#L352

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -230,6 +287,42 @@ sealed abstract class CreditingTxGen {
   def randoms: Gen[Seq[BitcoinUTXOSpendingInfo]] =
     Gen.choose(min, max).flatMap(n => Gen.listOfN(n, random))
 
+  private def computeAllTrueConditionalPath(
```
Christewart, line 290: Any way to make this better by just taking in a `UTXOSpendingInfo`?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -219,7 +220,8 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
 
   def p2shScriptPubKey: Gen[(P2SHScriptPubKey, Seq[ECPrivateKey])] =
     for {
-      (randomScriptPubKey, privKeys) <- randomNonP2SHScriptPubKey
+      (randomScriptPubKey, privKeys) <- randomNonP2SHScriptPubKey.suchThat(
+        _._1.compactSizeUInt.toLong + CompactSizeUInt(UInt64(520)).bytes.length < 520L)
```
Christewart: this is really used in a lot of different places at this point, do we need it everywhere? At minimum we should isolate this generator check into a single helper method.

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
Christewart: Here is a stackoverflow post talking about what happens when you use `.suchThat()` too much: https://stackoverflow.com/questions/45304489/why-is-my-scalacheck-test-with-a-custom-generator-failing-after-discarding-many

nkohen: In response to 1, When it comes to `P2SH` we push the redeem script onto the stack, hence the limit being needed.

In response to 2, I agree that in most cases this would not be ideal but the probability is low enough that this should be okay, as you say.

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
nkohen: Sadly no, because we make the nested SpendingInfo's into `val`s so there would be errors if we did not construct the conditional path ahead of time

testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -26,7 +27,11 @@ import scala.concurrent.duration.DurationInt
 //TODO: Need to provide generators for [[NonStandardScriptSignature]] and [[NonStandardScriptPubKey]]
 sealed abstract class ScriptGenerators extends BitcoinSLogger {
   val timeout = 5.seconds
-  private val defaultMaxDepth = 2
+  val defaultMaxDepth: Int = 2
+
+  private[gen] def scriptPubKeyTooBig(spk: ScriptPubKey): Boolean = {
+    spk.compactSizeUInt.toLong + CompactSizeUInt(UInt64(520)).bytes.length >= 520L
```
Christewart: This is what this implementation needs to be: 

```scala
  private[gen] def scriptPubKeyTooBig(spk: ScriptPubKey): Boolean = {
    spk.asm.exists(_.bytes.length > ScriptInterpreter.MAX_PUSH_SIZE) && 
      spk.bytes.length == ScriptInterpreter.MAX_SCRIPT_SIZE
  }
```

I think we should just make `MAX_PUSH_SIZE` public, and add a new constant for `MAX_SCRIPT_SIZE = 10000` and refactor the usages in `ScriptInterpreter`

PR #866, opened 2019-11-15
Increase some Test Coverage

nkohen:
This PR primarily adds negative test cases and increases test coverage by about `0.8%`


core-test/src/test/scala/org/bitcoins/core/protocol/script/ConditionalScriptSignatureSpec.scala
```diff
@@ -0,0 +1,28 @@
+package org.bitcoins.core.protocol.script
+
+import org.bitcoins.testkit.core.gen.{NumberGenerator, ScriptGenerators}
+import org.scalacheck.{Prop, Properties}
+
+class ConditionalScriptSignatureSpec
```
Christewart: So let's not use the old scalacheck stuff, and start using scalatest for the property based testing. I would like to migrate all old test cases at some point, but let's not add more burden. Basically just use the `BitcoinSUnitTest` trait here and the normal scalatest framework. 

One thing to note though, since these test cases are new make sure you 

```scala
  implicit override val generatorDrivenConfig: PropertyCheckConfiguration = {
    generatorDriveConfigNewCode
  }
```

which will cause 100 iterations on the property based tests, rather than the scalatest default of 10.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -348,10 +348,7 @@ sealed abstract class P2WPKHSigner extends BitcoinSigner[P2WPKHV0SpendingInfo] {
                                                unsignedTxWitness)
 
           val witSPK = output.scriptPubKey match {
-            case p2wpkh: P2WPKHWitnessSPKV0 =>
-              if (p2wpkh != P2WPKHWitnessSPKV0(pubKey)) {
-                Future.fromTry(TxBuilderError.WrongPublicKey)
-              } else Future.successful(p2wpkh)
+            case p2wpkh: P2WPKHWitnessSPKV0 => Future.successful(p2wpkh)
```
Christewart, line 351: Is this condition being checked somewhere else?

nkohen: It is, that is now `P2WPKHSigner`'s responsibility

PR #867, opened 2019-11-15
MultiSignatureWithTimeoutScriptPubKey

nkohen:
Introduces the explicit subtype of `IfConditional` which has a `MultiSignatureScriptPubKey` in the true case and a `CLTVScriptPubKey` in the false case. This will be used by DLCs.

Only the last couple commits are new as this is built on top of #859

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -688,8 +688,16 @@ object ConditionalScriptPubKey {
       trueSPK: RawScriptPubKey,
       falseSPK: RawScriptPubKey): ConditionalScriptPubKey = {
     conditional match {
-      case OP_IF    => IfConditionalScriptPubKey(trueSPK, falseSPK)
-      case OP_NOTIF => NotIfConditionalScriptPubKey(falseSPK, trueSPK)
+      case OP_IF =>
+        (trueSPK, falseSPK) match {
+          case (multisig: MultiSignatureScriptPubKey,
+                timeout: CLTVScriptPubKey) =>
+            MultiSignatureWithTimeoutScriptPubKey(multisig, timeout)
+          case _ =>
+            NonStandardIfConditionalScriptPubKey(trueSPK, falseSPK)
+        }
+      case OP_NOTIF =>
+        NonStandardNotIfConditionalScriptPubKey(falseSPK, trueSPK)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -723,20 +733,97 @@ object IfConditionalScriptPubKey
...
+    MultiSignatureScriptPubKey.isMultiSignatureScriptPubKey(super.firstSPK.asm),
+    "True case must be MultiSignatureSPK")
+  require(CLTVScriptPubKey.isCLTVScriptPubKey(super.secondSPK.asm),
+          "False case must be CLTVSPK")
+
+  override def firstSPK: MultiSignatureScriptPubKey = {
+    MultiSignatureScriptPubKey.fromAsm(asm.slice(1, opElseIndex))
+  }
+
+  override def secondSPK: CLTVScriptPubKey = {
```
Christewart, line 760: Any reason for using `CLTVScriptPubKey` rather than a more generic `LockTimeScriptPubKey`?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -723,20 +733,97 @@ object IfConditionalScriptPubKey
...
+  def timeout: Int = timeoutSPK.locktime.toInt
+}
+
+object MultiSignatureWithTimeoutScriptPubKey
+    extends ScriptFactory[MultiSignatureWithTimeoutScriptPubKey] {
+  private case class MultiSignatureWithTimeoutScriptPubKeyImpl(
+      override val asm: Vector[ScriptToken])
+      extends MultiSignatureWithTimeoutScriptPubKey {
+    override def toString: String =
+      s"IfConditionalScriptPubKey($trueSPK, $falseSPK)"
```
Christewart: `s"MultiSignatureWithTimeoutScriptPubKey($trueSPK, $falseSPK)"`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -723,20 +733,97 @@ object IfConditionalScriptPubKey
...
+  }
+
+  override def fromAsm(
+      asm: Seq[ScriptToken]): MultiSignatureWithTimeoutScriptPubKey = {
+    buildScript(
+      asm = asm.toVector,
+      constructor = MultiSignatureWithTimeoutScriptPubKeyImpl.apply,
+      invariant = isMultiSignatureWithTimeoutScriptPubKey,
+      errorMsg = "Given asm was not a IfConditionalScriptPubKey, got: " + asm
```
Christewart: `"Given asm was not a MultiSignatureWithTimeoutScriptPubKey, got: $asm"`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: I'm open to doing that in the future but I wanted to narrow the scope to what I'll actually use on the first iteration of DLCs for now so as not to bloat, I believe making it `LockTime` does add some complexity IIRC

PR #868, opened 2019-11-18
bitcoind v19 Update mempool RPCs and tests

benthecarman:
This should fix all the test issues so it is compatible with master

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -107,7 +107,7 @@ object BitcoindVersion {
   }
 
   case object Experimental extends BitcoindVersion {
-    override def toString: String = "v0.19.99"
+    override def toString: String = "v0.18.99"
   }
 
   case object Unknown extends BitcoindVersion {
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RawTransactionRpc.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RawTransactionRpc.scala
@@ -90,10 +90,10 @@ trait RawTransactionRpc { self: Client =>
 
   def sendRawTransaction(
       transaction: Transaction,
-      allowHighFees: Boolean = false): Future[DoubleSha256DigestBE] = {
+      maxfeerate: Double = 0.10): Future[DoubleSha256DigestBE] = {
     bitcoindCall[DoubleSha256DigestBE](
       "sendrawtransaction",
-      List(JsString(transaction.hex), JsBoolean(allowHighFees)))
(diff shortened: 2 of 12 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -107,7 +107,7 @@ object BitcoindVersion {
   }
 
   case object Experimental extends BitcoindVersion {
-    override def toString: String = "v0.19.99"
+    override def toString: String = "v0.18.99"
```
benthecarman, line 110: I shouldn't have changed this in my first PR

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RawTransactionRpc.scala
```diff
@@ -90,10 +91,18 @@ trait RawTransactionRpc { self: Client =>
 
   def sendRawTransaction(
```
nkohen, line 95: add a scaladoc explaining the semantics of setting `maxfeerate = 0`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/BlockchainResult.scala
```diff
@@ -159,22 +175,72 @@ case class GetMemPoolResult(
...
+  def fee: Bitcoins
+  def modifiedfee: Bitcoins
+  def time: UInt32
+  def height: Int
+  def descendantcount: Int
+  def descendantsize: Int
+  def descendantfees: Bitcoins // Should be BitcoinFeeUnit
+  def ancestorcount: Int
+  def ancestorsize: Int
+  def ancestorfees: Bitcoins // Should be BitcoinFeeUnit
```
nkohen: ping

nkohen (commented):
Looking good :) just a couple nits

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RawTransactionRpc.scala
```diff
@@ -88,12 +89,23 @@ trait RawTransactionRpc { self: Client =>
     bitcoindCall[Transaction]("getrawtransaction", params)
   }
 
+  /**
+    * @param maxfeerate On versions before 0.19 set to 0 if you want to enable allowhighfees
```
nkohen: Even for v0.19 `0` means enable `allowhighfees` right? I think just get rid of the first part here

nkohen (approved):
Just one thing

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/RawTransactionRpc.scala
benthecarman: You're right, that makes more sense

PR #870, opened 2019-11-19
Add test to check avoid_reuse flag is on

benthecarman:


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -68,10 +68,11 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
     for {
       (client, _) <- clientPairF
       result <- client.setWalletFlag(WalletFlag.AvoidReuse, value = true)
+      unspent <- client.listUnspent
     } yield {
       assert(result.flag_name == "avoid_reuse")
       assert(result.flag_state)
-      // TODO: Add test cases for when updated RPCs are added
+      assert(unspent.forall(utxo => utxo.reused.isDefined))
```
nkohen: any chance we could assert that this is not the case before calling the RPC just to make this test a little more robust?

nkohen (commented):
Just one thing

nkohen (approved):

PR #871, opened 2019-11-19
Binary Outcome DLC

nkohen:
Adds support for constructing and signing binary outcome DLCs.

(was previously #694)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -0,0 +1,306 @@
...
+
+  /** Subtracts the estimated fee from the last output's value */
+  private def subtractFeeAndSign(
+      txBuilder: BitcoinTxBuilder): Future[Transaction] = {
+    txBuilder.unsignedTx.flatMap { tx =>
+      val fee = feeRate.calc(tx)
+      val output = txBuilder.destinations.last
+      val newOutput = TransactionOutput(output.value - fee, output.scriptPubKey)
+      val newBuilder =
+        BitcoinTxBuilder(txBuilder.destinations.dropRight(1).:+(newOutput),
```
Christewart, line 136: This seems pretty brittle, as it assumes the change output is the last output. Is this just a hack for now for a longer term solution (proper fee estimation for the script type?)

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -0,0 +1,306 @@
...
+  val fundingSPK: MultiSignatureScriptPubKey = {
+    MultiSignatureScriptPubKey(2,
+                               Vector(fundingLocalPubKey, fundingRemotePubKey))
+  }
+
+  /** Subtracts the estimated fee from the last output's value */
+  private def subtractFeeAndSign(
+      txBuilder: BitcoinTxBuilder): Future[Transaction] = {
+    txBuilder.unsignedTx.flatMap { tx =>
+      val fee = feeRate.calc(tx)
```
Christewart, line 132: So i think i might be missing something here

Why doesn't this "dummy signing" process work for this script? 

https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala#L174

You can see that the flag there is `true`, so that we should place "dummy" signatures where real ones should go. Everything else should be the same as this signing dry run expect when we call the real `.sign()` method, those dummy signatures will be replaced with real ones.

There is a problem of digital signatures varying in size -- IIRC they can vary between 68 bytes and 73 bytes -- but I don't think that should matter for a simple 2 of 2 multisig or a timeout clause.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: Yup, this is my temporary solution and better fee handling will be on future PRs as you reminded me not to make this PR too large :P

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -0,0 +1,118 @@
...
+import org.bitcoins.testkit.util.BitcoinSAsyncTest
+import scodec.bits.ByteVector
+
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+
+class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
+  behavior of "BinaryOutcomeDLCWithSelf"
+
+  it should "work" in {
```
Christewart: so shouldn't there be test cases for 

1. Local wins, cooperative close
2. Remote wins, cooperative close
3. Local wins, non-cooperative close
4. Remote wins, non cooperative close
5. Anything I am missing?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -0,0 +1,306 @@
...
+      fundingSpendingInfo: BitcoinUTXOSpendingInfo): Future[Transaction] = {
+    createCETRemote(
+      sigPubKey = sigPubKeyLose,
+      fundingSpendingInfo = fundingSpendingInfo,
+      localPayout = localLosePayout,
+      remotePayout = remoteLosePayout
+    )
+  }
+
+  def executeDLC(oracleSigF: Future[SchnorrDigitalSignature]): Future[
```
Christewart: How can we tell what close case is happening based on this signature? Let's actually break things up into the different close cases to make this more understandable. 

Comments are great! :-)

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
nkohen: I am not adding support for cooperative close on this PR.

But I did add test cases so that we now cover all local and remote wins and loses cases.

Christewart:
Since you added a new project you need to add `dlc/test` to the CI matrix. You can add it to a existing matrix in `.travis.yml` or add a new matrix.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.scalatest.Assertion
+import scodec.bits.ByteVector
+
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+
+class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
+  behavior of "BinaryOutcomeDLCWithSelf"
+
+  it should "work" in {
```
Christewart: maybe a more descriptive name of what this is actually testing :-)

PR #872, opened 2019-11-19
Remove minconf argument from sendmany for new versions

benthecarman:


nkohen (approved):
LGTM

nkohen (changes requested):
Actually, could you make sure that both cases are tested? This will likely involve putting two similar tests in some two places where one is a v19 node and one isn't.

benthecarman:
When adding the extra test found out this was done wrong, turns out that the `minconf` argument still needs to be entered for future versions it just is now a dummy value

PR #874, opened 2019-11-20
Add new services names parameter to P2P rpcs

benthecarman:


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/jsonmodels/NetworkResult.scala
```diff
@@ -39,6 +39,7 @@ case class GetNetworkInfoResult(
     subversion: String,
     protocolversion: Int,
     localservices: String,
+    localservicesnames: Option[Vector[String]],
```
nkohen: I believe we have a type for this, also could you make sure that there is an assert on this field in the test for this RPC?

core/src/main/scala/org/bitcoins/core/p2p/ServiceIdentifier.scala
```diff
@@ -173,6 +173,16 @@ object ServiceIdentifier extends Factory[ServiceIdentifier] {
   def fromBytes(bytes: ByteVector): ServiceIdentifier =
     RawServiceIdentifierSerializer.read(bytes)
 
+  def fromString(string: String): ServiceIdentifier = string match {
```
nkohen, line 176: let's add a catch-all case `_: String` and throw an `IllegalArgumentException` with a more descriptive error message than `unmatched` which is what would happen as is

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -543,4 +544,15 @@ object JsonReaders {
         case "test"    => TestNet3
       }
   }
+
+  implicit object ServiceIdentifierReads extends Reads[ServiceIdentifier] {
+    override def reads(json: JsValue): JsResult[ServiceIdentifier] =
+      json match {
+        case JsString(s) =>
+          JsSuccess(ServiceIdentifier.fromString(s))
```
nkohen: I believe you actually have to call `fromString` inside of `Try(_)` and then match on the result. Currently it will throw instead of returning a `JsError` if it gets an unexpected `String`

core/src/main/scala/org/bitcoins/core/p2p/ServiceIdentifier.scala
```diff
@@ -173,6 +173,19 @@ object ServiceIdentifier extends Factory[ServiceIdentifier] {
...
+    case "NETWORK"         => NODE_NETWORK
+    case "NETWORK_LIMITED" => NODE_NETWORK_LIMITED
+    case "WITNESS"         => NODE_WITNESS
+    case "BLOOM"           => NODE_BLOOM
+    case "GETUTXO"         => NODE_GET_UTXO
+    case "COMPACT_FILTERS" => NODE_COMPACT_FILTERS
+    case "XTHIN"           => NODE_XTHIN
+    case _: String =>
+      throw new IllegalArgumentException(
+        "\"" + string + "\" does not represent a ServiceIdentifier")
```
nkohen: nit: Just fyi, in scala this can be `s"\"$string\" does not represent a ServiceIdentifier"`

nkohen (approved):
LGTM

PR #877, opened 2019-11-20
Move PsbtRpc to be able to be used by future versions of bitcoind

benthecarman:


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/PsbtRpc.scala
```diff
@@ -28,4 +28,12 @@ trait V18PsbtRpc {
     bitcoindCall[String]("utxoupdatepsbt", List(JsString(psbt)))
   }
 
+  def utxoUpdatePsbt(
```
nkohen: could you add a test for this new method

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/DescriptorRpc.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  def deriveAddresses(
+      descriptor: String,
+      range: Option[Vector[Double]]): Future[DeriveAddressesResult] = {
+    bitcoindCall[DeriveAddressesResult](
+      "deriveaddresses",
+      List(JsString(descriptor), Json.toJson(range)))
+  }
+
+  def getDescriptorInfo(descriptor: String): Future[GetDescriptorInfoResult] = {
```
nkohen, line 29: tests for these two

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/DescriptorRpc.scala
```diff
@@ -0,0 +1,32 @@
...
+  */
+trait DescriptorRpc {
+  self: Client =>
+
+  def deriveAddresses(
+      descriptor: String,
+      range: Option[Vector[Double]]): Future[DeriveAddressesResult] = {
+    bitcoindCall[DeriveAddressesResult](
+      "deriveaddresses",
+      List(JsString(descriptor), Json.toJson(range)))
```
nkohen: In the case where `range == None` shouldn't this second param be omitted?

benthecarman: Does `Json.toJson` do that automatically?

btw this is just copy paste from the original `v18DescriptorRpc.scala`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/DescriptorRpc.scala
benthecarman: They already exist in `BitcoindV18ClientTest`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/DescriptorRpc.scala
nkohen: `Json.toJson(None)` returns `JsNull`. Not sure if this is the intended functionality but I have a feeling that this param list should be a val defined on matching on `range`

nkohen (approved):
LGTM

PR #883, opened 2019-11-21
Add passphrase argument to createwallet for later versions

benthecarman:
Last one :smiley:

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
@@ -186,6 +186,12 @@ trait WalletRpc { self: Client =>
       List(JsString(currentPassphrase), JsString(newPassphrase)))
   }
 
+  /**
+    *
+    * @param blank Not available to versions before v19
+    * @param passphrase Not available to versions before v19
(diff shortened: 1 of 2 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
```diff
@@ -187,11 +188,21 @@ trait WalletRpc { self: Client =>
...
   def createWallet(
       walletName: String,
-      disablePrivateKeys: Boolean = false): Future[CreateWalletResult] = {
-    bitcoindCall[CreateWalletResult](
-      "createwallet",
-      List(JsString(walletName), Json.toJson(disablePrivateKeys)))
-  }
+      disablePrivateKeys: Boolean = false,
+      blank: Boolean = false,
+      passphrase: String = ""): Future[CreateWalletResult] =
```
nkohen, line 199: Since this isn't used pre V19, we definitely need to say in the scaladoc that this param is only used for versions after and including 19 so that people don't think they're getting encrypted wallets when they aren't. Also it might be a good idea to be extra safe and `require(passphrase.isEmpty)` in the `v16 | v17 | v18` case for the same reason.

nkohen (approved):
LGTM

PR #885, opened 2019-11-22
Improve DLC Unit Test

nkohen:
Have `executeDLC` return all transactions and their respective `UTXOSpendingInfo`s so that all of them can be tested.


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -268,7 +269,13 @@ case class BinaryOutcomeDLCWithSelf(
     */
   def executeDLC(
       oracleSigF: Future[SchnorrDigitalSignature],
-      local: Boolean): Future[(Transaction, BitcoinUTXOSpendingInfo)] = {
+      local: Boolean): Future[
+    (
```
Christewart: Can we a create a new type for this with documentation rather than a massive nested tuple?

PR #886, opened 2019-11-22
Regtest Binary DLC

nkohen:
Add bitcoind test which publishes all DLC transactions to Regtest for validation

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -9,6 +9,7 @@ import org.bitcoins.core.crypto.{
 }
 import org.bitcoins.core.currency.CurrencyUnit
 import org.bitcoins.core.number.UInt32
+import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.script.{
   CLTVScriptPubKey,
   ConditionalScriptPubKey,
@@ -25,7 +26,7 @@ import org.bitcoins.core.protocol.transaction.{
 }
 import org.bitcoins.core.script.constant.ScriptNumber
 import org.bitcoins.core.script.crypto.HashType
-import org.bitcoins.core.util.{BitcoinSLogger, CryptoUtil}
+import org.bitcoins.core.util.{BitcoinSLogger, CryptoUtil, FutureUtil}
 import org.bitcoins.core.wallet.builder.BitcoinTxBuilder
 import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.core.wallet.utxo.{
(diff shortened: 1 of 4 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -268,7 +270,10 @@ case class BinaryOutcomeDLCWithSelf(
     */
   def executeDLC(
       oracleSigF: Future[SchnorrDigitalSignature],
-      local: Boolean): Future[(Transaction, BitcoinUTXOSpendingInfo)] = {
+      local: Boolean,
+      rpcClientAndAddressOpt: Option[
```
Christewart: there should not be a dependency on bitcoind rpc client inside of the DLC project source code. This breaks bitcoin-s modularity for DLCs. Signing/construction/storing of DLC contracts should not depend on a bitcoind rpc client.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfIntegrationTest.scala
```diff
@@ -0,0 +1,189 @@
...
+      val oracleSig =
+        Schnorr.signWithNonce(outcomeHash.bytes, oraclePrivKey, preCommittedK)
+
+      val fundedInputsTxidF = for {
+        client <- clientF
+        transactionWithoutFunds <- client
+          .createRawTransaction(Vector.empty,
+                                Map(localAddress.get -> Bitcoins(4),
+                                    remoteAddress.get -> Bitcoins(4)))
+        transactionResult <- client.fundRawTransaction(transactionWithoutFunds)
```
Christewart, line 85: So is the thing blocking you from using the bitcoin-s wallet is the fact we don't have a `.fundRawTransaction()`/`signRawTransactionWithWallet()`?

nkohen: It also doesn't have any functionality to broadcast to RegTest, in short I don't see how it has any utility here.

Christewart: Network functionality should not be touching signing functionality. They are two different systems that should be kept far apart for security reasons. I'm interested in what is blocking from signing be executed with the bitcoin-s wallet, is it the two methods I specified above?

nkohen: I believe it is just `fundRawTransaction` since that creates inputs (with signatures present). Also, do note that in this iteration Bitcoind RPC client is only used in test and the place where you are proposing I use bitcoin-s wallet is in that test. Nowhere am I using bitcoind for signing functionality in actual source code, only as a setup to a test, which I think should be acceptable?

Christewart: Well technically the way you have it setup the `testkit` is in `src` in the build.sbt file which brings in the entire testkit to the classpath in source. You can see this with `show dependencyClasspath` in sbt

I think this is fine for now with #896 filed. We have an understanding of moving away from the DLC project in the future to having things in the wallet and core project.

PR #887, opened 2019-11-22
Enable bloom filters for v19

benthecarman:
0.19.0 disables bloom filters by default, we need to enable it for some tests


nkohen (approved):
LGTM

PR #888, opened 2019-11-22
[Neutrino] Update balances

rorp:
Implements `OnBlockReceived` handler that extracts relevant transaction and updates wallet balances

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -1,0 +1,111 @@
+package org.bitcoins.node
+
+import akka.actor.Cancellable
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+import org.bitcoins.core.currency._
+import org.bitcoins.node.networking.peer.DataMessageHandler
+import org.bitcoins.rpc.client.common.BitcoindVersion
+import org.bitcoins.server.BitcoinSAppConfig
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.fixtures.UsesExperimentalBitcoind
+import org.bitcoins.testkit.node.NodeUnitTest.NeutrinoNodeFundedWalletBitcoind
+import org.bitcoins.testkit.node.{NodeTestUtil, NodeUnitTest}
+import org.bitcoins.wallet.api.UnlockedWalletApi
+import org.scalatest.exceptions.TestFailedException
+import org.scalatest.{DoNotDiscover, FutureOutcome}
+
+import scala.concurrent.duration._
+import scala.concurrent.{Future, Promise}
+
+@DoNotDiscover
+class NeutrinoNodeWithWalletTest extends NodeUnitTest {
+
+  /** Wallet config with data directory set to user temp directory */
+  implicit override protected def config: BitcoinSAppConfig =
+    BitcoinSTestAppConfig.getNeutrinoTestConfig()
+
+  override type FixtureParam = NeutrinoNodeFundedWalletBitcoind
+
+  def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withNeutrinoNodeFundedWalletBitcoind(test,
+                                         callbacks,
+                                         Some(BitcoindVersion.Experimental))
+  }
+
+  private val assertionP: Promise[Boolean] = Promise()
+
+  private val walletP: Promise[UnlockedWalletApi] = Promise()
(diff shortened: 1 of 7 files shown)


node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.bitcoins.testkit.node.NodeUnitTest.NeutrinoNodeFundedWalletBitcoind
+import org.bitcoins.testkit.node.{NodeTestUtil, NodeUnitTest}
+import org.bitcoins.wallet.api.UnlockedWalletApi
+import org.scalatest.exceptions.TestFailedException
+import org.scalatest.{DoNotDiscover, FutureOutcome}
+
+import scala.concurrent.duration._
+import scala.concurrent.{Future, Promise}
+
+@DoNotDiscover
```
nkohen, line 24: ping? Should this be here?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -35,6 +35,31 @@ private[wallet] trait TransactionProcessing extends KeyHandlingLogger {
...
+          newWallet
+        }
+    }
+    res.foreach(
+      _ =>
+        logger.info(
+          s"Finished processing of block=${block.blockHeader.hash.flip}."))
+    res.failed.foreach(e =>
+      logger.error(s"Error processing of block=${block.blockHeader.hash.flip}.",
+                   e))
```
nkohen, line 59: nit: These two things can be done in a single `res.onComplete`

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
rorp: It should. We still don't have a statically linked Neutrino-enabled bitcoind binary for CI, so this test will break in CI.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -0,0 +1,111 @@
...
+            }
+          }
+        }
+
+        logger.debug(s"Setting timeout for receiving TX through node")
+        cancellable = Some(system.scheduler.scheduleOnce(delay, failTest))
+        tx
+      }
+
+      for {
```
Christewart, line 84: So it seems to me the test case should 

1. Send money to the bitcoin-s wallet
2. Confirm that money on the bitcoin network
3. Erase the `SpendingInfoTable` 
4. Read the addresses we have generated from the `AddressTable` 
5. Rescan the chain against those SPKs we read from the `AddressTable`
6. For the scripts that matched, insert them into `SpendingInfoTable`.
7. Assert we have the same amount of money as we had in step 2. 
8. Assert we have the same contents in `SpendingInfoTable` as we had in step 3.


This seems to be what we want to do, and it does simplify the problem of generating addresses from an HDKey but avoiding it all together but just assuming `AddressTable` is populated. In the future, we will have to populate `AddressTable` from the HDKey, but we can do that on a separate PR.

Christewart: I think the name on this test case shouldn't include anything about rescanning as it is just about receiving money. I would rename the current `rescan` method to `fetchBlock()` or something like that.

This starts to tie into what we talked about on Friday with moving filters into the wallet/chain project and also having rescan logic live there. This should simplify operations in the node project. Let me know if you want me to clarify this comment.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -0,0 +1,121 @@
...
+    )
+  }
+
+  it must "receive information about received payments" taggedAs (UsesExperimentalBitcoind) in {
+    param =>
+      val NeutrinoNodeFundedWalletBitcoind(node, wallet, bitcoind) = param
+
+      walletP.success(wallet)
+
+      def clearSpendingInfoTable(): Future[Int] = {
```
Christewart, line 64: I don't think we need to clear spending info table if we are just receiving information about a normal payment. This only needs to be done in the case of a rescan. So we can simplify this test case by just receiving the payment, not deleting the table too.

rorp: But this is the case of a rescan.

PR #892, opened 2019-11-25
DLC Fee distribution

nkohen:
Fees paid on each DLC transaction are now evenly distributed among the two nodes.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -130,10 +130,21 @@ case class BinaryOutcomeDLCWithSelf(
       txBuilder: BitcoinTxBuilder): Future[Transaction] = {
     txBuilder.unsignedTx.flatMap { tx =>
       val fee = feeRate.calc(tx)
-      val output = txBuilder.destinations.last
-      val newOutput = TransactionOutput(output.value - fee, output.scriptPubKey)
+
+      val outputs = txBuilder.destinations
+
+      val feePerOutput = Satoshis(Int64(fee.satoshis.toLong / outputs.length))
+      val feeRemainder = Satoshis(Int64(fee.satoshis.toLong % outputs.length))
+
+      val newOutputsWithoutRemainder = outputs.map(output =>
+        TransactionOutput(output.value - feePerOutput, output.scriptPubKey))
+      val lastOutput = newOutputsWithoutRemainder.last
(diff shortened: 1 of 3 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -125,15 +125,26 @@ case class BinaryOutcomeDLCWithSelf(
...
+      val newOutputsWithoutRemainder = outputs.map(output =>
+        TransactionOutput(output.value - feePerOutput, output.scriptPubKey))
+      val lastOutput = newOutputsWithoutRemainder.last
+      val newLastOutput = TransactionOutput(lastOutput.value - feeRemainder,
+                                            lastOutput.scriptPubKey)
+      val newOutputs = newOutputsWithoutRemainder.dropRight(1).:+(newLastOutput)
+
       val newBuilder =
-        BitcoinTxBuilder(txBuilder.destinations.dropRight(1).:+(newOutput),
+        BitcoinTxBuilder(newOutputs,
```
Christewart: Can we get a test case on this that this is actually right? I think that `TxBuilder` will try to re-subtract fees again when we call `.sign()`?

nkohen: Added a test that checks that building fails before subtraction and works (as expected, with evenly distributed fees among outputs) after.

To clarify, `TxBuilder` never subtracts fees from anything, it expects `sum(inputValues) > sum(outputValues)` and throws the difference minus fees into a change SPK. That is why this is necessary and why it works as expected.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -9,23 +9,92 @@ import org.bitcoins.core.crypto.{
...
+          tx <- BinaryOutcomeDLCWithSelf.subtractFeeAndSign(txBuilder)
+        } yield {
+          val diffs = outputs.zip(tx.outputs).map {
+            case (before, after) =>
+              before.value - after.value
+          }
+
+          val firstDiff = diffs.head
+          // Fee has been evenly distributed (up to some remainder)
+          assert(diffs.forall(diff => diff.compare(firstDiff) < diffs.length))
```
Christewart: I don't understand what this is actually checking? `diff.compare(firstDiff)` just checks returns `-1`, `0`, or `1` doesn't it?

PR #895, opened 2019-11-26
DLC Refund Tx Construction

nkohen:
Adds the construction and signing of the refund transaction in DLC execution.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
@@ -114,6 +114,8 @@ sealed abstract class TxBuilder {
   /** The script witnesses that are needed in this transaction */
   def scriptWitOpt: Seq[Option[ScriptWitness]] = utxos.map(_.scriptWitnessOpt)
 
+  def lockTimeOverrideOpt: Option[UInt32]
+
   /**
     * The unsigned version of the tx with dummy signatures instead of real signatures in
     * the [[org.bitcoins.core.protocol.script.ScriptSignature ScriptSignature]]s.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -151,21 +153,32 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
     val emptyChangeOutput = TransactionOutput(CurrencyUnits.zero, changeSPK)
-    val unsignedTxNoFee = lockTime.map { l =>
+    val unsignedTxNoFee = calculatedLockTime.map { cLockTime =>
+      val lockTime = lockTimeOverrideOpt match {
+        case None => cLockTime
+        case Some(overrideLockTime) =>
+          if (overrideLockTime >= cLockTime) {
+            overrideLockTime
+          } else {
+            throw new IllegalStateException(
```
Christewart: Can we just do `Future.failed()` rather than a hard throw?

Christewart: Also put the locktimes in the error message

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -713,11 +727,46 @@ object BitcoinTxBuilder {
...
+                     lockTimeOverrideOpt = None)
+  }
+
+  def apply(
+      destinations: Seq[TransactionOutput],
+      utxos: BitcoinTxBuilder.UTXOMap,
+      feeRate: FeeUnit,
+      changeSPK: ScriptPubKey,
+      network: BitcoinNetwork,
+      lockTimeOverride: UInt32): Future[BitcoinTxBuilder] = {
```
Christewart, line 753: Maybe rather than adding a bunch of applys we just provide it as a default

```scala
lockTimeOverrideOpt: Option[UInt32] = None
```

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -221,6 +222,35 @@ case class BinaryOutcomeDLCWithSelf(
...
 
+  /** Constructs the (time-locked) refund transaction for when the oracle disappears
+    * or signs an unknown message.
+    * Note that both parties have the same refund transaction.
+    */
+  def createRefundTx(
+      fundingSpendingInfo: MultiSignatureSpendingInfo): Future[Transaction] = {
+    val toLocalValueNotSat =
+      (fundingSpendingInfo.amount * localInput).satoshis.toLong / totalInput.satoshis.toLong
+    val toLocalValue = Satoshis(Int64(toLocalValueNotSat))
```
Christewart, line 233: Can you make a small PR with a constructor on `Satoshis` that just takes a scala `Long` so we don't always have to wrap in a `Int64()`? This is annoying and i've been meaning to do it forever..

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
nkohen: It isn't legal in scala to have overloaded methods define defaults as it can cause the compiler to be unable to distinguish which method is being called:
```
in object BitcoinTxBuilder, multiple overloaded alternatives of method apply define default arguments.
```

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: #899

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -151,52 +153,72 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
       implicit ec: ExecutionContext): Future[Transaction] = {
     val utxos = utxoMap.values.toList
     val unsignedTxWit = TransactionWitness.fromWitOpt(scriptWitOpt.toVector)
-    val lockTime = calcLockTime(utxos)
+    val calculatedLockTime = calcLockTime(utxos)
```
Christewart, line 156: nit: Append `T` to this name for `Try`

Christewart (commented):
Do we need a test case for this still?

PR #899, opened 2019-11-26
Nicer Satoshis.apply

nkohen:
Replaced `Satoshis(Int64(_))` with `Satoshis(_)`


core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -84,6 +84,8 @@ object Satoshis
   override def fromBytes(bytes: ByteVector): Satoshis =
     RawSatoshisSerializer.read(bytes)
   def apply(int64: Int64): Satoshis = SatoshisImpl(int64)
+  def apply(satoshis: Long): Satoshis = SatoshisImpl(Int64(satoshis))
+  def apply(satoshis: BigInt): Satoshis = SatoshisImpl(Int64(satoshis))
```
Christewart, line 88: I'd like to leave the old apply method as well, i don't see a reason in removing it, but it is _really_ annoying to have to use it.

PR #900, opened 2019-11-26
Fix OP_PUSHDATA1

nkohen:
Moved BitcoinTxBuilder property tests into ScalaTest context, fixed bug where `opPushData` was marking valid short P2SH scripts as invalid.

This bug was causing spurious failures in what was formerly `BitcoinTxBuilderSpec` and those are gone now :)


core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
```diff
@@ -364,4 +386,102 @@ class BitcoinTxBuilderTest extends AsyncFlatSpec with MustMatchers {
       )
     }
   }
+
+  def verifyScript(tx: Transaction, utxos: Seq[UTXOSpendingInfo]): Boolean = {
```
Christewart, line 390: Don't we have this in `BitcoinScriptUtil` now?

Christewart (commented):
LGTM, although i would like to add a hard coded test case for this to make sure we don't regress in future changes. 

EDIT: 

Actually it seems like CI is timing out now on this PR? https://travis-ci.org/bitcoin-s/bitcoin-s/builds/617476488?utm_source=github_status&utm_medium=notification

I also saw this behavior on #892 for one of the `coreTest` CI runs there

core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
nkohen: Sadly that change only got merged into the `dlc` branch and isn't on `master` yet

PR #901, opened 2019-11-27
BitcoinTxBuilderTest Hang

nkohen:
Maybe fixes spurious hanging bug in `BitcoinTxBuilderTest`.

Also adds `forAllAsync` method to `BitcoinSAsyncTest` to avoid this problem in the future.


core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
```diff
@@ -464,12 +466,23 @@ class BitcoinTxBuilderTest extends BitcoinSAsyncTest {
...
+        testRunFs += {
+          txF.map { tx =>
+            assert(verifyScript(tx, creditingTxsInfo))
+          }
+        }
+
+        succeed
     }
+
+    Future.sequence(testRunFs.result()).map(_.reduce((_, next) => next))
```
Christewart: So i think we should definitely move this stuff into `testkit` and actually write a test to make sure that this fails as expected? I think we can use this else where.

PR #904, opened 2019-12-01
2019 11 30 key manager

Christewart:
This PR segregates key handling/storage logic from the wallet project. For the most part, this is a move only PR taking things inside the `wallet` project and moving them to the new `key-manager` project.

For now, this just delegates storage of the encrypted mnemonic seed to the key manager project. Over the log run, we want to make it so that the wallet project needs to communicate with the key-manager to access private keys. 

This means that ALL SIGNING should be done inside of the key-manager, and private keys should not be leaked outside of the key manager. This makes it easier to reason about the security characteristics of our private keys, and a way to provide a uniform interface for alternative key storage systems (hsm, cloud based key storage, etc) to be plugged into the bitcoin-s library. 

##### TODO 
- [x] key-manager.md
- [x] duplicate encrypted seed unit test


docs/applications/key-manager.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+
+#### Key Manager
+
+The key manager module's goal is to encapusulate all private key interactions with the [wallet](wallet.md) project.
+
+As of this writing, the wallet just delegates storage of the encrypted mnemonic seed to the key manager project. Over the log run, we want to make it so that the wallet project needs to communicate with the key-manager to access private keys.
+
+This means that ALL SIGNING should be done inside of the key-manager, and private keys should not be leaked outside of the key manager.
```
nkohen: nit "should not be leaked outside of" -> "should not leave" (they shouldn't be leaked anywhere ;)

testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
```diff
@@ -36,6 +36,9 @@ trait BitcoinSFixture extends BitcoinSAsyncFixtureTest {
       test(fixture.asInstanceOf[FixtureParam]).toFuture
     }
 
+    outcomeF.failed.foreach(err =>
+      println(s"Failed fixture test execution=${err}"))
```
nkohen, line 40: nit: `println -> logger.error`

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -73,6 +72,16 @@ case class WalletAppConfig(
     initF
   }
 
+  /** The path to our encrypted mnemonic seed */
+  protected[wallet] def seedPath: Path = {
```
nkohen: can't this be `private[wallet]`? case classes can't be extended and I think this grants more access than we want (protected is weird)

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    assert(walletConf.seedExists())
+
+    val exn: RuntimeException = intercept[RuntimeException] {
+      //attempt to write another mnemonic
+      getAndWriteMnemonic(walletConf)
+    }
+
+    //need this since we are using async testing suite
+    //and exn is not a Future
+    succeed
```
rorp: `assertThrows()` returns an `Assertion`

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      WalletStorage.decryptMnemonicFromDisk(seedPath,passphrase)
+    read match {
+      case ReadMnemonicSuccess(readMnemonic) =>
+        assert(writtenMnemonic == readMnemonic)
+      case err: ReadMnemonicError => fail(err.toString)
+    }
+  }
+
+  it must "fail to read a mnemonic with bad password" in { walletConf =>
+    val writtenMnemonic = getAndWriteMnemonic(walletConf)
```
benthecarman: This val is never used

benthecarman:
CI is failing for almost all the cases

testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
Christewart: I'm going to keep this now to help with debugging on CI as we have logging turned off there

PR #905, opened 2019-12-02
Neutrino Wallet: OnCompactFilter handler

rorp:


core/src/main/scala/org/bitcoins/core/node/NodeApi.scala
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+  * Represent a bitcoin-s node API
+  */
+trait NodeApi {
+
+  /**
+    * Fetches the given blocks from the peers.
+    */
+  def fetchBlocks(blockHashes: Vector[DoubleSha256Digest]): Future[Unit]
```
Christewart: Do we want this to return `Future[Vector[Block]]`?

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -31,7 +31,10 @@ case class NeutrinoNode(
 
   override val peer: Peer = nodePeer
 
-  override val callbacks: SpvNodeCallbacks = nodeCallbacks
+  override def setCallbacks(callbacks: SpvNodeCallbacks): Node =
+    copy(nodeCallbacks = callbacks)
+
+  override def setBloomFilter(bloomFilter: BloomFilter): Node = this
```
Christewart: It seems this method should only be available on `SpvNode`?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,7 +48,11 @@ trait Node extends P2PLogger {
 
   val peer: Peer
 
-  val callbacks: SpvNodeCallbacks
+  val nodeCallbacks: SpvNodeCallbacks
+
+  def setCallbacks(callbacks: SpvNodeCallbacks): Node
+
+  def setBloomFilter(bloomFilter: BloomFilter): Node
```
Christewart: I don't think this should be in the `Node` trait

node/src/main/scala/org/bitcoins/node/SpvNodeCallbacks.scala
```diff
@@ -28,11 +29,16 @@ object SpvNodeCallbacks {
   def onMerkleBlockReceived(f: OnMerkleBlockReceived): SpvNodeCallbacks =
     SpvNodeCallbacks(onMerkleBlockReceived = Seq(f))
 
+  /** Constructs a set of callbacks that only acts on block received */
```
Christewart: this scaladoc is wrong I think. Should be on compact  filter received?

node/src/main/scala/org/bitcoins/node/SpvNodeCallbacks.scala
```diff
@@ -9,6 +9,7 @@ import org.bitcoins.node.networking.peer.DataMessageHandler._
   *
   */
 case class SpvNodeCallbacks(
```
Christewart: We should probably rename this class from `SpvNodeCallbacks` -> `NodeCallbacks`

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -94,6 +95,12 @@ case class DataMessageHandler(
             }
           }
           newChainApi <- chainApi.processFilter(filter)
+          _ <- Future {
```
Christewart: So is this `Future` needed because the signature is

```scala
type OnCompactFilterReceived = (DoubleSha256Digest, GolombFilter) => Unit
```

rather than  returning a `Future[Unit]`? Are you worried that the callback is going to take a long time to process? It seems like a valid concern to me.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -20,6 +21,8 @@ import scala.util.{Failure, Success, Try}
 
 sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
 
+  implicit val node: NodeApi
```
Christewart: Hmm... not sure about this. I'm guessing this is needed to fetch the blocks?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -152,7 +156,8 @@ object Wallet extends CreateWalletApi with KeyHandlingLogger {
 
   def apply(mnemonicCode: MnemonicCode)(
       implicit config: WalletAppConfig,
-      ec: ExecutionContext): Wallet =
+      ec: ExecutionContext,
+      node: NodeApi): Wallet =
```
Christewart: I don't think this should be passed as an implicit every where as it is hard to follow implicits in my opinion.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -36,26 +34,109 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
...
+        } yield {
+          (expectedConfirmedAmount == confirmedBalance) &&
+          (expectedUnconfirmedAmount == unconfirmedBalance) &&
+          (expectedAddresses == addresses.size) &&
+          (expectedUtxos == utxos.size)
+        }
+      }
+
+      val condition1 = () => {
+        condition(0.sats,
```
Christewart: nit: can you use named arguments (`alt+enter` on your keyboard in intellij). This makes this kind of stuff so much more readable for code reviewers :-)

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -200,4 +205,16 @@ trait Node extends P2PLogger {
       scriptPubKeysToWatch: Vector[ScriptPubKey],
       startOpt: Option[BlockStamp] = None,
       endOpt: Option[BlockStamp] = None): Future[Unit]
+
+  /**
+    * Fetches the given blocks from the peers and calls the appropriate [[callbacks]] when done.
+    */
+  def fetchBlocks(blockHashes: Vector[DoubleSha256Digest]): Future[Unit] = {
```
Christewart: override

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -30,7 +30,11 @@ case class SpvNode(
 
   override val peer: Peer = nodePeer
 
-  override val callbacks: SpvNodeCallbacks = nodeCallbacks
+  override def setCallbacks(callbacks: NodeCallbacks): Node =
```
Christewart: is this kind of dangerous? It replaces all existing call backs right? maybe we just want a `addCallback()`? I would at least add a scaladoc saying it removes existing callbacks and replaces them with the given ones.

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -244,16 +245,19 @@ object NodeUnitTest extends P2PLogger {
...
     import system.dispatcher
     require(appConfig.isSPVEnabled && !appConfig.isNeutrinoEnabled)
-    val fundedWalletF = BitcoinSWalletTest.fundedWalletAndBitcoind(versionOpt)
     for {
-      fundedWallet <- fundedWalletF
-      node <- createSpvNode(fundedWallet.bitcoind, callbacks)
+      bitcoind <- BitcoinSFixture.createBitcoindWithFunds(versionOpt)
+      node <- createSpvNode(bitcoind, callbacks)
+      fundedWallet <- {
+        implicit val nodeApi: NodeApi = node
```
Christewart: Can we get rid of this implicit too?

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -263,16 +267,19 @@ object NodeUnitTest extends P2PLogger {
...
     import system.dispatcher
     require(appConfig.isNeutrinoEnabled && !appConfig.isSPVEnabled)
-    val fundedWalletF = BitcoinSWalletTest.fundedWalletAndBitcoind(versionOpt)
     for {
-      fundedWallet <- fundedWalletF
-      node <- createNeutrinoNode(fundedWallet.bitcoind, callbacks)
+      bitcoind <- BitcoinSFixture.createBitcoindWithFunds(versionOpt)
+      node <- createNeutrinoNode(bitcoind, callbacks)
+      fundedWallet <- {
+        implicit val nodeApi: NodeApi = node
```
Christewart: Same as I said above, let's make these explicit calls

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -113,7 +121,8 @@ object BitcoinSWalletTest extends WalletLogger {
     */
   private def createNewWallet(extraConfig: Option[Config])(
       implicit config: BitcoinSAppConfig,
-      ec: ExecutionContext): () => Future[UnlockedWalletApi] =
+      ec: ExecutionContext,
+      nodeApi: NodeApi): () => Future[UnlockedWalletApi] =
```
Christewart: and make this `nodeApi` a normal parameter rather than implicit one

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -141,10 +150,9 @@ object BitcoinSWalletTest extends WalletLogger {
   /** Creates a wallet with the default configuration  */
   private def createDefaultWallet()(
       implicit config: BitcoinSAppConfig,
-      ec: ExecutionContext): Future[UnlockedWalletApi] = {
-    val defaultF = createNewWallet(None)(config, ec)() // get the standard config
-    defaultF
-  }
+      ec: ExecutionContext,
+      nodeApi: NodeApi): Future[UnlockedWalletApi] =
```
Christewart: same as above

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -165,10 +173,18 @@ object BitcoinSWalletTest extends WalletLogger {
...
+  )(implicit system: ActorSystem): Future[WalletWithBitcoind] = {
+    Future.successful(WalletWithBitcoind(wallet, bitcoindRpcClient))
+  }
+
   /** Creates a default wallet, and then pairs it with a bitcoind instance that has money in the bitcoind wallet */
   def createWalletWithBitcoind()(
       implicit system: ActorSystem,
-      config: BitcoinSAppConfig): Future[WalletWithBitcoind] = {
+      config: BitcoinSAppConfig,
+      nodeApi: NodeApi): Future[WalletWithBitcoind] = {
```
Christewart: remove implicit `nodeApi`

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -177,7 +193,8 @@ object BitcoinSWalletTest extends WalletLogger {
   /** Gives us a funded bitcoin-s wallet and the bitcoind instance that funded that wallet */
   def fundedWalletAndBitcoind(versionOpt: Option[BitcoindVersion])(
       implicit config: BitcoinSAppConfig,
-      system: ActorSystem): Future[WalletWithBitcoind] = {
+      system: ActorSystem,
+      nodeApi: NodeApi): Future[WalletWithBitcoind] = {
```
Christewart: remove implicit `nodeApi`

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -186,6 +203,18 @@ object BitcoinSWalletTest extends WalletLogger {
     } yield funded
   }
 
+  def fundedWalletAndBitcoind(bitcoindRpcClient: BitcoindRpcClient)(
+      implicit config: BitcoinSAppConfig,
+      system: ActorSystem,
+      nodeApi: NodeApi): Future[WalletWithBitcoind] = {
```
Christewart: remove implicit `nodeApi`

core/src/main/scala/org/bitcoins/core/node/NodeApi.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+  * Represent a bitcoin-s node API
+  */
+trait NodeApi {
+
+  /**
+    * Request the underlying node to download the given blocks from its peers and feed the blocks to [[org.bitcoins.node.NodeCallbacks]].
+    */
+  def requestBlocks(blockHashes: Vector[DoubleSha256Digest]): Future[Unit]
```
nkohen: nit: This name makes it sound like the user is requesting blocks that the `NodeApi` has, maybe rename to `downloadBlocks`?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -95,6 +97,18 @@ case class DataMessageHandler(
...
+
+          Try {
+            val blockFilter =
+              BlockFilter.fromBytes(filter.filterBytes, filter.blockHash)
+            callbacks.onCompactFilterReceived.foreach(
+              _.apply(filter.blockHash, blockFilter))
+          } match {
+            case Failure(ex) =>
+              logger.error("Error processing compact filter", ex)
+            case Success(_) => // ignore
```
nkohen: return explicit `()`

node/src/main/scala/org/bitcoins/node/util/BitcoinSNodeUtil.scala
```diff
@@ -0,0 +1,51 @@
...
+    * @return
+    */
+  def createActorName(className: Class[_]): String =
+    createActorName(className.getSimpleName)
+}
+
+object BitcoinSNodeUtil extends BitcoinSNodeUtil {
+
+  class Mutable[A](initialValue: A) {
+    private val lock = new java.util.concurrent.locks.ReentrantReadWriteLock()
```
nkohen, line 26: you probably know better than me but isn't best practice to use `AtomicReference`?

rorp: Definitely, `AtomicReference.accumulateAndGet()` can be used for updating callbacks, because their `+` is a binary operator. But it's not that easy to use it for updating Bloom filters, because their `insert()` and `update()` methods are not binary operators.

nkohen (approved):
LGTM

PR #907, opened 2019-12-02
Derive DLC keys using BIP 32

nkohen:
This PR implements key derivation as follows:
```
Funding key = EventExtKey / 0
CET key = EventExtKey / 1 / outcome_index
Closing key = EventExtKey / 2
```
where `outcome_index = 0` for refunds and the index (counting from 1 since 0 is taken) of the outcome in `outcomeSeq.sorted`.

We also derive the remote's payouts from the `totalInput` and local payouts.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -116,6 +116,9 @@ case class BinaryOutcomeDLCWithSelf(
   val finalRemotePrivKey: ECPrivateKey =
     remoteExtPrivKey.deriveChildPrivKey(UInt32(2)).key
 
+  /** The derivation index for the win and lose cases respectively.
+    * We assign this index based on lexicographical order to keep things deterministic.
+    */
   private val (winIndex, loseIndex) =
     if (outcomeWin.compareTo(outcomeLose) > 0) {
       (1, 2)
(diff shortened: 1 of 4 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -77,20 +73,14 @@ case class BinaryOutcomeDLCWithSelf(
...
     oraclePubKey: ECPublicKey,
     preCommittedR: ECPublicKey,
-    fundingLocalPrivKey: ECPrivateKey,
-    fundingRemotePrivKey: ECPrivateKey,
-    cetLocalPrivKey: ECPrivateKey,
-    cetRemotePrivKey: ECPrivateKey,
-    finalLocalPrivKey: ECPrivateKey,
-    finalRemotePrivKey: ECPrivateKey,
+    localExtPrivKey: ExtPrivateKey,
+    remoteExtPrivKey: ExtPrivateKey,
```
Christewart, line 77: Shouldn't this be a `ExtPubKey` ?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -114,8 +104,68 @@ case class BinaryOutcomeDLCWithSelf(
...
+    remoteExtPrivKey.deriveChildPrivKey(UInt32(0)).key
+
+  val finalLocalPrivKey: ECPrivateKey =
+    localExtPrivKey.deriveChildPrivKey(UInt32(2)).key
+
+  val finalRemotePrivKey: ECPrivateKey =
+    remoteExtPrivKey.deriveChildPrivKey(UInt32(2)).key
+
+  private val (winIndex, loseIndex) =
+    if (outcomeWin.compareTo(outcomeLose) > 0) {
```
Christewart, line 123: What's going on here? We are comparing two Strings to determine who the winner is?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -114,8 +104,68 @@ case class BinaryOutcomeDLCWithSelf(
...
+    }
+
+  val cetLocalRefundPrivKey: ECPrivateKey = localExtPrivKey
+    .deriveChildPrivKey(
+      BIP32Path(BIP32Node(1, hardened = false), BIP32Node(0, hardened = false)))
+    .key
+
+  val cetLocalWinPrivKey: ECPrivateKey = localExtPrivKey
+    .deriveChildPrivKey(
+      BIP32Path(BIP32Node(1, hardened = false),
```
Christewart: maybe move this path construction into a helper method? The only thing that is different is the `extKey` and the `index` right?

Christewart (commented):
We probably want to follow BIP44 for the key derivation path, specifically i think we want to advocate for a `purpose` field to be set for DLC?

I guess we don't have to follow this on this PR, but I think that is probably the right direction to go.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: No, I haven't separated things out yet, as it currently stands, this class is responsible for building and signing DLC transactions, not for actually representing communication between two parties. This will change of course.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: My current spec that is detailed in the PR description states that you must use the indices of the ordered set of possible outcomes (to keep things deterministic). Hence the different outcomes get a different index depending on how they compare lexicographically.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: This is just replacing many private keys with one

nkohen:
> We probably want to follow BIP44 for the key derivation path, specifically i think we want to advocate for a `purpose` field to be set for DLC?
> 
> I guess we don't have to follow this on this PR, but I think that is probably the right direction to go.

I agree though this should not actually ever be dealt with at this level. Rather the extended key passed to this level must be beneath this (hardened) purpose just like an Account in BIP 44. It would be overly restrictive to enforce this derivation at the DLC level (should happen at wallet)

PR #910, opened 2019-12-02
Bitcoind v19 RPC

benthecarman:
This is all the changes being added to master


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
```diff
@@ -31,7 +29,7 @@ class MempoolRpcTest extends BitcoindRpcTest {
 
         val instanceWithoutBroadcast =
           BitcoindInstance.fromConfig(configNoBroadcast,
-                                      BitcoindRpcTestUtil.newestBitcoindBinary)
+                                      BitcoindRpcTestUtil.getBinary(V18))
```
Christewart, line 32: Why v18 here?

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -115,6 +116,8 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
                   |port=${uri.getPort}
                   |debug=1
                   |walletbroadcast=1
+                  |enablebip61=1
```
Christewart: Why bip61?

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -223,12 +226,14 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
       versionOpt: Option[BitcoindVersion] = None): BitcoindInstance = {
     val uri = new URI("http://localhost:" + port)
     val rpcUri = new URI("http://localhost:" + rpcPort)
-    val configFile = writtenConfig(
-      uri,
-      rpcUri,
-      zmqPort,
-      pruneMode,
-      blockFilterIndex = versionOpt.contains(BitcoindVersion.Experimental))
+    val hasNeutrinoSupport = versionOpt.contains(BitcoindVersion.V19) || versionOpt
```
Christewart, line 228: is this right? `v19` doesn't have p2p neutrino support does it? What type of neutrino support are we talking about? We should probably be more specific if we are talking about cli or p2p

benthecarman: This is correct, this means that the `blockfilterindex` config option works on the version of bitcoind

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
benthecarman: Because on line 19 we are creating a pair of v18 nodes, if it is left at `newestBitcoindBinary` it will be expecting a v18 bitcoind binary and get a v19 bitcoind binary

PR #915, opened 2019-12-03
DLC Refund Transaction Test

nkohen:
This PR makes the `timeout` type a `BlockStamp` instead of an `Int`.

It also adds an execution case with new tests for spending the refund transaction.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
@@ -6,6 +6,7 @@ import java.time.temporal.{ChronoField, TemporalAccessor}
 
 import org.bitcoins.core.crypto.DoubleSha256DigestBE
 import org.bitcoins.core.number.UInt32
+import org.bitcoins.core.script.constant.ScriptNumber
 
 import scala.util.{Failure, Try}
 
@@ -17,6 +18,7 @@ sealed trait BlockStamp {
 /** This trait represents a point on the blockchain, including future points */
 sealed trait BlockStampWithFuture extends BlockStamp {
   def toUInt32: UInt32
+  def toScriptNumber: ScriptNumber
 }
 
 object BlockStamp {
@@ -30,6 +32,7 @@ object BlockStamp {
     require(height >= 0, "block height must be a positive number")
     override def mkString: String = height.toString
     override def toUInt32: UInt32 = UInt32(height)
+    override def toScriptNumber: ScriptNumber = ScriptNumber(height)
   }
   case class BlockTime(time: UInt32) extends BlockStampWithFuture {
     override def mkString: String = {
@@ -37,6 +40,7 @@ object BlockStamp {
       DateTimeFormatter.ISO_INSTANT.format(instant)
     }
     override def toUInt32: UInt32 = time
+    override def toScriptNumber: ScriptNumber = ScriptNumber(time.toLong)
   }
 
   /** This is Tue Nov  5 00:53:20 1985 UTC
--- a/core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
@@ -14,6 +14,11 @@ sealed trait BlockStamp {
   def mkString: String
 }
 
+/** This trait represents a point on the blockchain, including future points */
+sealed trait BlockStampWithFuture extends BlockStamp {
+  def toUInt32: UInt32
+}
+
 object BlockStamp {
   case class InvalidBlockStamp(blockStamp: String)
       extends RuntimeException(s"Invalid blockstamp: ${blockStamp}")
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
```diff
@@ -14,22 +14,43 @@ sealed trait BlockStamp {
...
+    override def toUInt32: UInt32 = time
+  }
+
+  /** This is Tue Nov  5 00:53:20 1985 UTC
+    * @see [[https://github.com/bitcoin/bitcoin/blob/master/src/script/script.h#L39]]
+    */
+  final val LOCKTIME_THRESHOLD: Int = 500000000
+
+  /** @see [[https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki#detailed-specification]] */
+  def apply(timelockNumber: Int): BlockStampWithFuture = {
```
benthecarman, line 52: Should `timelockNumber` be a`UInt32`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -370,97 +373,253 @@ case class BinaryOutcomeDLCWithSelf(
...
-              finalLocalPrivKey
+  /** Constructs and executes on the unilateral spending branch of a DLC
+    *
+    * @return Each transaction published and its spending info
+    */
+  def executeUnilateralDLC(
+      oracleSigF: Future[SchnorrDigitalSignature],
+      local: Boolean,
+      messengerOpt: Option[BitcoinP2PMessenger] = None): Future[DLCOutcome] = {
+    setupDLC(messengerOpt).flatMap {
```
Christewart: setup shouldn't be happening in the `executeUnilateralDLC()` method. You are co-mingling test code and source code.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -370,97 +373,253 @@ case class BinaryOutcomeDLCWithSelf(
...
+          // The prefix other refers to remote if local == true and local otherwise
+          val output = cet.outputs.head
+          val otherOutput = cet.outputs.last
+
+          // Spend the true case on the correct CET
+          val cetSpendingInfo = ConditionalSpendingInfo(
+            TransactionOutPoint(cet.txIdBE, UInt32.zero),
+            output.value,
+            output.scriptPubKey.asInstanceOf[ConditionalScriptPubKey],
+            Vector(cetPrivKey, ECPrivateKey(oracleSig.s)),
```
Christewart: this seems kind of hacky

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -231,7 +232,7 @@ case class BinaryOutcomeDLCWithSelf(
       requiredSigs = 2,
       pubKeys = Vector(cetRemotePrivKey.publicKey, sigPubKey))
     val timeoutSPK = CLTVScriptPubKey(
-      locktime = ScriptNumber(timeout),
+      locktime = ScriptNumber(timeout.toUInt32.toInt),
```
Christewart: I would add a `toScriptNumber()` method on `BlockStamp`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -323,15 +324,17 @@ case class BinaryOutcomeDLCWithSelf(
...
+      fundingTx: Transaction,
+      fundingSpendingInfo: MultiSignatureSpendingInfo,
+      cetWinLocal: Transaction,
+      cetLoseLocal: Transaction,
+      cetWinRemote: Transaction,
+      cetLoseRemote: Transaction,
+      refundTx: Transaction
+  )
+
+  def setupDLC(messengerOpt: Option[BitcoinP2PMessenger]): Future[SetupDLC] = {
```
Christewart: Should this be test code?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -370,97 +373,253 @@ case class BinaryOutcomeDLCWithSelf(
...
+    }
+  }
+
+  /** Constructs and executes on the refund spending branch of a DLC
+    *
+    * @return Each transaction published and its spending info
+    */
+  def executeRefundDLC(
+      messengerOpt: Option[BitcoinP2PMessenger] = None): Future[DLCOutcome] = {
+    setupDLC(messengerOpt).flatMap {
```
Christewart: setup should be in test code

Christewart (commented):
Two things

1. It's pretty clear we are co-mingling test code and source code at this point, all setup stuff should be moved into test cases. This code is going to have to be reworked pretty significantly in the future as you don't have access to the other persons private keys. If this stuff was setup properly in tests, you can have the test case setup the remote side of the DLC and the local side of the DLC, using the same code. Right now you just assume everything is available in src.

2. Can you please use named parameters (`alt+enter` in intellij) so that it is easier for reviewers to review your code?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: Yup, DLCs are essentially just this hack (assuming you are referring to `ECPrivateKey(oracleSig.s)`

nkohen:
I plan on removing `BitcoinP2PMessenger` in the subsequent PR (this should de-futurify things).

And of course, separating this out into two nodes with different information is work for a future PR once construction is finalized.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
Christewart: I would like to see if we can do something with the `Sign` interface, but i guess this isn't the time for that.

PR #917, opened 2019-12-04
PSBT Serialization/Deserialization

benthecarman:
This is my preliminary work I have done on PSBTs before I start testing. Just want some feedback on basic structure or some glaring issues I am missing


core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -144,6 +149,21 @@ object BIP32Path {
...
   }
+
+  def fromBytes(bytes: ByteVector): BIP32Path = {
+    require(bytes.size % 4 == 0, "ByteVector is not suited for KeyPath")
+
+    val parts = bytes.toIterable.grouped(4).toVector
+
+    val path = parts.map { part =>
+      val uInt32: UInt32 = UInt32.fromBytes(ByteVector(part))
+      val hardened = uInt32 > ExtKey.hardenedIdx
```
benthecarman: Should this be `>=`?

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+    val key = bytes.drop(keySize.size).take(keySize.toLong)
+    val valueSize =
+      CompactSizeUInt.parse(bytes.drop(keySize.size + keySize.toLong))
+    val value = bytes
+      .drop(keySize.size + keySize.toLong + valueSize.size)
+      .take(valueSize.toLong)
+    val totalSize = keySize.size + keySize.toLong + valueSize.size + valueSize
+      .toLong
+    (totalSize, key, value)
+  }
```
benthecarman: I think something is wrong in this function, can't figure out what but I think `value` has an extra byte

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializationTest.scala
```diff
@@ -0,0 +1,19 @@
+package org.bitcoins.core.psbt
+
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import scodec.bits._
+
+class PSBTSerializationTest extends BitcoinSUnitTest {
+
+  val validPsbts: List[ByteVector] = List(
```
Christewart: In general it is best to use `Vector` in scala for collections. It has the "least" amount of surprises in terms of pathological performance. See http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html for more information on the various collections performance characteristics

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializationTest.scala
```diff
@@ -0,0 +1,19 @@
...
+    // PSBT with one P2PKH input and one P2SH-P2WPKH input. First input is signed and finalized. Outputs are empty
+    hex"70736274ff0100a00200000002ab0949a08c5af7c49b8212f417e2f15ab3f5c33dcf153821a8139f877a5b7be40000000000feffffffab0949a08c5af7c49b8212f417e2f15ab3f5c33dcf153821a8139f877a5b7be40100000000feffffff02603bea0b000000001976a914768a40bbd740cbe81d988e71de2a4d5c71396b1d88ac8e240000000000001976a9146f4620b553fa095e721b9ee0efe9fa039cca459788ac000000000001076a47304402204759661797c01b036b25928948686218347d89864b719e1f7fcf57d1e511658702205309eabf56aa4d8891ffd111fdf1336f3a29da866d7f8486d75546ceedaf93190121035cdc61fc7ba971c0b501a646a2a83b102cb43881217ca682dc86e2d73fa882920001012000e1f5050000000017a9143545e6e33b832c47050f24d3eeb93c9c03948bc787010416001485d13537f2e265405a34dbafa9e3dda01fb82308000000"
+  )
+
+  it must "successfully serialize and deserialize valid psbts" in {
+//    assert(validPsbts.forall(psbt => PSBT.fromBytes(psbt).toBytes == psbt))
+    assert(PSBT.fromBytes(validPsbts.head) != null)
```
Christewart: `null` is evil, never use it :-). Use `None` instead

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -144,6 +147,21 @@ object BIP32Path {
 
     BIP32PathImpl(path)
   }
+
+  def fromBytes(bytes: ByteVector): BIP32Path = {
+    require(bytes.size % 4 == 0, "ByteVector is not suited for KeyPath")
```
Christewart: it's always good to include the thing we are requiring an invariant on in the error message, like this:

```scala
require(bytes.size % 4 == 0, s"ByteVector is not suited for KeyPath, got=${bytes.length}")
```

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -144,6 +147,21 @@ object BIP32Path {
...
+
+    val parts = bytes.toIterable.grouped(4).toVector
+
+    val path = parts.map { part =>
+      val uInt32: UInt32 = UInt32.fromBytes(ByteVector(part))
```
Christewart: if you remove the `.toIterable` on line 154 and remove change `ByteVector(part)` -> `part` this should be a bit cleaner. Here is what i have for this method

```scala
  def fromBytes(bytes: ByteVector): BIP32Path = {
    require(bytes.size % 4 == 0, s"ByteVector is not suited for KeyPath, got=${bytes.length}")

    val parts: Vector[ByteVector] = bytes.grouped(4).toVector

    val path = parts.map { part =>
      val uInt32: UInt32 = UInt32.fromBytes(part)
      val hardened = uInt32 > ExtKey.hardenedIdx
      val index = if (hardened) uInt32 - ExtKey.hardenedIdx else uInt32
      BIP32Node(index.toInt, hardened)
    }

    BIP32Path(path)
  }
```

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -144,6 +147,21 @@ object BIP32Path {
 
     BIP32PathImpl(path)
   }
+
+  def fromBytes(bytes: ByteVector): BIP32Path = {
```
Christewart: This seems like an important addition to BIP32Path, let's make sure we get some tests around this. This should be ripe for property based tests. We might be able to `extend Factory[BIP32Path]` on this companion object.

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
Christewart: I believe it should be: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys 

```
The normal child keys use indices 0 through 231-1. The hardened child keys use indices 231 through 232-1. To ease notation for hardened key indices, a number iH represents i+231. 
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+  BaseTransaction,
+  EmptyWitness,
+  Transaction,
+  WitnessTransaction
+}
+import org.bitcoins.core.script.crypto.HashType
+import scodec.bits._
+
+object PSBT {
+  final def magicBytes = hex"70736274ff"
```
Christewart: `val` rather than `def` so we dont' re-evaluate the expression everytime `magicBytes` is called

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+
+  def fromBytes(byteVector: ByteVector): PSBT = {
+    require(byteVector.take(5) == hex"70736274ff",
```
Christewart: can't this be `== magicBytes`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+            "A PSBT must start with the magic bytes 0x70736274ff")
```
Christewart: again, with these invariants you will thank yourself later if you include the input in the error message so 

```scala
    require(byteVector.take(5) == hex"70736274ff",
            s"A PSBT must start with the magic bytes 0x70736274ff, got=${byteVector}")
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+
+    val bytes = byteVector.drop(5)
+
+    val (bytes1: ByteVector, global: Seq[GlobalPSBTRecord], tx: Transaction) =
+      PSBTHelper.parseGlobalMap(bytes, accum = Nil, tx = null)
```
Christewart: `null` is evil :-). Always use an `Option[]` and pattern match on it. This gives `null` circuvments the type system as `null` can type check with anything.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+         outputMaps.toVector,
+         tx)
+  }
+}
+
+case class PSBT(
+    globalMap: GlobalPSBTMap,
+    inputMaps: Vector[InputPSBTMap],
+    outputMaps: Vector[OutputPSBTMap],
+    tx: Transaction) {
```
Christewart: we probably want to `extend NetworkElement` here.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+  require(inputMaps.size == tx.inputs.size)
+  require(outputMaps.size == tx.outputs.size)
+
+  val bytes: ByteVector = PSBT.magicBytes ++ globalMap.bytes ++ inputMaps
+    .foldLeft(ByteVector.empty)(_ ++ _.bytes ++ hex"ff") ++ outputMaps.foldLeft(
+    ByteVector.empty)(_ ++ _.bytes ++ hex"ff")
+
+  def toBytes: ByteVector = bytes
```
Christewart: we should be able to get rid of this when we `extend NetworkElement`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+
+case class PSBT(
+    globalMap: GlobalPSBTMap,
+    inputMaps: Vector[InputPSBTMap],
+    outputMaps: Vector[OutputPSBTMap],
+    tx: Transaction) {
+  require(inputMaps.size == tx.inputs.size)
+  require(outputMaps.size == tx.outputs.size)
+
+  val bytes: ByteVector = PSBT.magicBytes ++ globalMap.bytes ++ inputMaps
```
Christewart: I think this is a bit more readable 

```scala
  private val inputBytes: ByteVector = {
    inputMaps
      .foldLeft(ByteVector.empty)(_ ++ _.bytes ++ hex"ff")
  }

  private val outputBytes = {
    outputMaps.foldLeft(
      ByteVector.empty)(_ ++ _.bytes ++ hex"ff")
  }
  val bytes: ByteVector = PSBT.magicBytes ++
    globalMap.bytes ++
    inputBytes ++
    outputBytes
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+    .foldLeft(ByteVector.empty)(_ ++ _.bytes ++ hex"ff") ++ outputMaps.foldLeft(
+    ByteVector.empty)(_ ++ _.bytes ++ hex"ff")
+
+  def toBytes: ByteVector = bytes
+}
+
+sealed trait PSBTRecord {
```
Christewart: `extends NetworkElement`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+    override val value: ByteVector = transaction.bytes
+  }
+
+  case class XPubKey(
+      xpub: ExtPublicKey,
+      masterFingerprint: ByteVector,
+      derivationPath: BIP32Path)
+      extends GlobalPSBTRecord {
+    require(derivationPath.path.length == xpub.depth.toInt,
+            "Derivation path length does not match xpubkey depth")
```
Christewart: same comment applies for adding input that was given to the error message for easier to debugging :-)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,336 @@
...
+  val bytes: ByteVector = PSBT.magicBytes ++ globalMap.bytes ++ inputMaps
+    .foldLeft(ByteVector.empty)(_ ++ _.bytes ++ hex"ff") ++ outputMaps.foldLeft(
+    ByteVector.empty)(_ ++ _.bytes ++ hex"ff")
+
+  def toBytes: ByteVector = bytes
+}
+
+sealed trait PSBTRecord {
+
+  def key: ByteVector
```
Christewart: Can we change the type of this to `PSBTKey`?

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+import org.bitcoins.core.serializers.script.RawScriptWitnessParser
+import org.bitcoins.core.util.BitcoinScriptUtil
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+object PSBTHelper {
+
+  private def parseKeyAndValue(
+      bytes: ByteVector): (Long, ByteVector, ByteVector) = {
```
Christewart: it would be nice to have a case class that wraps the `(Long, ByteVector, ByteVector)`, in general i don't really like returning tuples.

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
Christewart: i can't see it either, but this seems a bit cleaner 

```scala
  private def parseKeyAndValue(
      bytes: ByteVector): (Long, ByteVector, ByteVector) = {
    val keyCmpctUInt = CompactSizeUInt.parse(bytes)
    val key = bytes.drop(keyCmpctUInt.size).take(keyCmpctUInt.toLong)
    val valueBytes = bytes.drop(keyCmpctUInt.size + keyCmpctUInt.toLong)
    val valueCmpctUInt = CompactSizeUInt.parse(valueBytes)
    val value = valueBytes
      .drop(valueCmpctUInt.size)
      .take(valueCmpctUInt.toLong)
    val totalSize = keyCmpctUInt.size + keyCmpctUInt.toLong + valueCmpctUInt.toLong + valueCmpctUInt.size
    (totalSize, key, value)
  }
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+      .toLong
+    (totalSize, key, value)
+  }
+
+  @tailrec
+  final def parseGlobalMap(
+      remainingBytes: ByteVector,
+      accum: Seq[GlobalPSBTRecord],
+      tx: Transaction): (ByteVector, Seq[GlobalPSBTRecord], Transaction) = {
+    if (remainingBytes.head == 0x00.byteValue) // terminator
```
Christewart: this is unsafe, as we don't know if `remaingBytes.nonEmpty`

Christewart: while scala does allow you to not use `{}` to wrap if expressions, I think it is best practice to use this syntax 

```scala
if (...whatever...) { 
  //if stuff in here 
} else { 
  //else stuff in here
}
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+      rest.head == 0x00.byteValue,
+      s"Key-value maps must be separated by the separator char 0x00 was: ${ByteVector(rest.head)}")
+    val next = rest.tail
+    PSBTGlobalKey.fromByte(key.head) match {
+      case UnsignedTransactionKey =>
+        println("DEBUGGING HERE")
+        println(key)
+        println(value)
+        println(next)
+        if (tx != null)
```
Christewart: if we use an `Option[]` like i suggested else where, we can just do a simple pattern match here: 

```scala
txOpt match { 
  case Some(tx) => //do whatever
  case None => throw IllegalArugementException(...)
}
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+      .take(valueSize.toLong)
+    val totalSize = keySize.size + keySize.toLong + valueSize.size + valueSize
+      .toLong
+    (totalSize, key, value)
+  }
+
+  @tailrec
+  final def parseGlobalMap(
+      remainingBytes: ByteVector,
+      accum: Seq[GlobalPSBTRecord],
```
Christewart: Use a `Vector[GlobalPSBTRecord]`

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }
+
+  @tailrec
+  def parseInputMaps(
+      remainingBytes: ByteVector,
+      remainingInputs: Int,
+      accum: Seq[InputPSBTMap]): (ByteVector, Seq[InputPSBTMap]) = {
+    if (remainingInputs <= 0)
+      return (remainingBytes, accum)
```
Christewart: is general `return` isn't needed in scala. You need it here because you aren't using `{}` around your `if () { } else { }` expression. 

This should fix your problem 

```scala
    if (remainingInputs <= 0) {
      (remainingBytes, accum)
    } else {
      val (next, map) = PSBTHelper.parseInputMap(remainingBytes, Nil)

      require(next.head == 0xff.byteValue,
        "Input maps should be separated by 0xff")

      parseInputMaps(next.tail,
        remainingInputs - 1,
        accum :+ InputPSBTMap(map.toVector))      
    }
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
Christewart: Also we should definitely have some `val` some where to represent the terminator rather than just having `0x00` hard coded everywhere.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: No we can't because some keys are longer than just the single byte identifier

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Does that make sense, I don't think there is ever a situation where we would send a `PSBTRecord` on its own when not inside of a `PSBT`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,387 @@
...
+    override val key: ByteVector = hex"01"
+    override val value: ByteVector = witnessScript.asmBytes
+  }
+
+  case class BIP32DerivationPath(
+      pubKey: ECPublicKey,
+      masterFingerprint: ByteVector,
+      path: BIP32Path)
+      extends OutputPSBTRecord {
+    require(pubKey.size == 33, s"pubKey must be 33 bytes, got: ${pubKey.size}")
```
benthecarman, line 399: Should this be moved to `ECPublickKey`?

core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
benthecarman: We discussed this in a call, will be doing in a separate PR

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
```diff
@@ -0,0 +1,94 @@
...
+    hex"70736274ff0100730200000001301ae986e516a1ec8ac5b4bc6573d32f83b465e23ad76167d68b38e730b4dbdb0000000000ffffffff02747b01000000000017a91403aa17ae882b5d0d54b25d63104e4ffece7b9ea2876043993b0000000017a914b921b1ba6f722e4bfa83b6557a3139986a42ec8387000000000001011f00ca9a3b00000000160014d2d94b64ae08587eefc8eeb187c601e939f9037c00010016001462e9e982fff34dd8239610316b090cd2a3b747cb000100220020876bad832f1d168015ed41232a9ea65a1815d9ef13c0ef8759f64b5b2b278a6521010025512103b7ce23a01c5b4bf00a642537cdfabb315b668332867478ef51309d2bd57f8a8751ae00"
+  )
+
+  it must "fail to serialize invalid PSBTs" in {
+    assert(invalidPsbts.forall(psbt =>
+      Try(PSBT.fromBytes(psbt)) match {
+        case Success(_) =>
+          fail(s"The following psbt was able to serialize: $psbt")
+        case Failure(_) => true
+      }))
```
nkohen: I would replace this with a call to `assertThrows`

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+/**
+  * Test vectors are taken directly from the BIP 174 reference sheet
+  * https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#test-vectors
+  */
+class PSBTSerializerTest extends BitcoinSUnitTest {
+
+  val validPsbts: Vector[ByteVector] = Vector(
```
nkohen, line 29: I think it could be nice to assert the things we know should be in these PSBTs to ensure that deserialization happened correctly

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,387 @@
...
+  EmptyWitness,
+  Transaction,
+  TransactionOutput,
+  WitnessTransaction
+}
+import org.bitcoins.core.script.crypto.HashType
+import scodec.bits._
+
+object PSBT {
+  final val magicBytes = hex"70736274ff"
```
nkohen: probably link to where this comes from in the bip

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,387 @@
...
+  BaseTransaction,
+  EmptyWitness,
+  Transaction,
+  TransactionOutput,
+  WitnessTransaction
+}
+import org.bitcoins.core.script.crypto.HashType
+import scodec.bits._
+
+object PSBT {
```
nkohen: This can probably extend `Factory[PSBT]` as `fromBytes` is defined

nkohen: nit: Generally I like to have the companion object under the class

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: No, so it turns out there are a couple different serializations of `ECPublicKey` which are valid in different contexts. We want the compressed version here (hence we require 33 bytes)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,387 @@
...
+      s"A PSBT must start with the magic bytes $magicBytes, got: $byteVector")
+
+    val bytes = byteVector.drop(5)
+
+    val (bytes1: ByteVector,
+         global: Vector[GlobalPSBTRecord],
+         txOpt: Option[Transaction]) =
+      PSBTHelper.parseGlobalMap(bytes, accum = Nil, txOpt = None)
+    if (txOpt.isEmpty)
+      throw new Exception("Invalid PSBT. No global transaction")
```
nkohen: nit: `IllegalArgumentException`

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,256 @@
...
+    } else {
+
+      val (totalSize, key, value) = parseKeyAndValue(remainingBytes)
+      val next = remainingBytes.drop(totalSize) // Drop the key-value map
+
+      PSBTGlobalKey.fromByte(key.head) match {
+        case UnsignedTransactionKey =>
+          txOpt match {
+            case Some(_) =>
+              throw new Exception("Duplicate Key, unsigned tx already provided")
```
nkohen: nit: `IllegalArgumentException`

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,256 @@
...
+            case None =>
+              require(
+                key.size == 1,
+                s"The key must only contain the 1 byte type, got: ${key.size}")
+
+              val transaction: Transaction = Transaction.fromBytes(value)
+              parseGlobalMap(next,
+                             accum :+ UnsignedTransaction(transaction),
+                             Option(transaction))
```
nkohen: nit: `Some(transaction)`

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,256 @@
...
+        .take(valueCmpctUInt.toLong)
+      val totalSize = keyCmpctUInt.size + keyCmpctUInt.toLong + valueCmpctUInt.size + valueCmpctUInt
+        .toLong
+
+      (totalSize, key, value)
+    }
+  }
+
+  @tailrec
+  def parseGlobalMap(
```
nkohen: For every single `@tailrec` function in this file, I think it would be nice to have a `def` to make the interface nicer for users that didn't have all of the params that aren't necessary to the internal workings of the recursion. For example, this function should have a
```
def parseGlobalMap(bytes: ByteVector) = {
  parseGlobalMap(remainingBytes = bytes, accum = Vector.empty, txOpt = None)
} 
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,256 @@
...
+                         txOpt)
+      }
+    }
+  }
+
+  @tailrec
+  def parseInputMaps(
+      remainingBytes: ByteVector,
+      remainingInputs: Int,
+      accum: Seq[InputPSBTMap]): (ByteVector, Vector[InputPSBTMap]) = {
```
nkohen: I think we should introduce a 
```private[psbt] case class PSBTParseResult[T](remainingBytes: ByteVector, parsed: T)```
and use that as the return type on all of the functions in this file rather than using tuples.
So this return type becomes `PSBTParseResult[Vector[InputPSBTMap]]` for example.
And the global one becomes `PSBTParseResult[(Vector[GlobalPSBTRecord], Option[Transaction])]` and so on

File: core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala (hunk 24 above)
Comment (benthecarman) [reply core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala]: Wouldn't that only test the first case then?

File: core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala [hunk 33]
```diff
@@ -0,0 +1,215 @@
...
+    // PSBT With invalid output witnessScript typed key
+    hex"70736274ff0100730200000001301ae986e516a1ec8ac5b4bc6573d32f83b465e23ad76167d68b38e730b4dbdb0000000000ffffffff02747b01000000000017a91403aa17ae882b5d0d54b25d63104e4ffece7b9ea2876043993b0000000017a914b921b1ba6f722e4bfa83b6557a3139986a42ec8387000000000001011f00ca9a3b00000000160014d2d94b64ae08587eefc8eeb187c601e939f9037c00010016001462e9e982fff34dd8239610316b090cd2a3b747cb000100220020876bad832f1d168015ed41232a9ea65a1815d9ef13c0ef8759f64b5b2b278a6521010025512103b7ce23a01c5b4bf00a642537cdfabb315b668332867478ef51309d2bd57f8a8751ae00"
+  )
+
+  it must "fail to serialize invalid PSBTs" in {
+    invalidPsbts.foreach(invalidPsbt =>
+      assertThrows[Exception](PSBT(invalidPsbt)))
```
nkohen: nit: Is there a more specific type than `Exception` that could go here? At the very least `RuntimeException` but if there's something more specific that would be great

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,423 @@
...
+import org.bitcoins.core.psbt.PSBTGlobalKey.UnsignedTransactionKey
+import org.bitcoins.core.script.crypto.HashType
+import org.bitcoins.core.util.Factory
+import scodec.bits._
+
+case class PSBT(
+    globalMap: GlobalPSBTMap,
+    inputMaps: Vector[InputPSBTMap],
+    outputMaps: Vector[OutputPSBTMap],
+    tx: Transaction)
```
nkohen: I think that rather than having `tx` as a param, you should make it a `val` in side the definition of `PSBT` by pulling it out of `globalMap` and you also should have a `require` statement that the `globalMap` contains a tx as it says in the BIP that `A PSBT must have a transaction, otherwise it is invalid.`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,423 @@
...
+
+  def elements: Vector[PSBTRecord]
+
+  def bytes: ByteVector =
+    elements.foldLeft(ByteVector.empty)(_ ++ _.bytes) ++ hex"00"
+}
+
+// Key may be confusing because this is just the first byte of the key
+// Maybe change to Identifier, Byte,  or something else
+sealed trait PSBTKey {
```
nkohen: I like `PSBTKeyId` and then in the scaladoc explain that it is the first byte of a record's key

core/src/main/scala/org/bitcoins/core/psbt/PSBTHelper.scala
```diff
@@ -0,0 +1,269 @@
...
+import org.bitcoins.core.serializers.script.RawScriptWitnessParser
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+private[psbt] case class PSBTParseResult[T <: PSBTMap](
+    remainingBytes: ByteVector,
+    maps: Vector[T])
+
+object PSBTHelper {
```
nkohen: Also, I thought I had left this comment on previous review but I couldn't find the comment so assuming it got deleted before I submitted it somehow. I would prefer that the `PSBTHelper` object be broken up into companion objects for things in `PSBT.scala`, if you want to keep it in a separate file I'm okay with that, maybe name it `PSBTCompanionObjects.scala` or `PSBTUtils.scala`. But what I want is to have a `PSBTRecord` object, as well as a `GlobalMap`, `InputMap`, and `OutputMap` object so that rather than calling `PSBTHelper.parseKeyAndValue`, I would instead call `PSBTRecord.fromBytes` and instead of `PSBTHelper.parseGlobalMap` I'd have `GlobalMap.fromBytes` and so on with the other two objects

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
benthecarman: It'll be an `IllegalArgumentException` or a failed `require()`, is there a type that encompasses both?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -84,18 +86,48 @@ object PSBT extends Factory[PSBT] {
...
+        inputLoop(bytes.drop(map.bytes.size), numInputs - 1, accum :+ map)
+      }
+    }
 
-    val inputsResult: PSBTParseResult[InputPSBTMap] =
-      PSBTHelper.parseInputMaps(globalResult.remainingBytes, tx.inputs.size)
+    val inputMaps = inputLoop(inputBytes, tx.inputs.size, Nil)
+
+    val outputBytes =
+      inputBytes.drop(inputMaps.foldLeft(ByteVector.empty)(_ ++ _.bytes).size)
```
nkohen: nit: `inputBytes.drop(inputMaps.foldLeft(0)(_ + _.bytes.size))`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -84,18 +86,48 @@ object PSBT extends Factory[PSBT] {
...
 
-    val outputsResult: PSBTParseResult[OutputPSBTMap] =
-      PSBTHelper.parseOutputMaps(inputsResult.remainingBytes, tx.outputs.size)
+    val outputMaps = outputLoop(outputBytes, tx.outputs.size, Nil)
 
-    require(
-      outputsResult.remainingBytes.isEmpty,
-      s"The PSBT should be empty now, got: ${outputsResult.remainingBytes}")
+    val remainingBytes =
+      outputBytes.drop(outputMaps.foldLeft(ByteVector.empty)(_ ++ _.bytes).size)
```
nkohen: same as above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -108,6 +140,48 @@ object PSBT extends Factory[PSBT] {
...
+      val valueCmpctUInt = CompactSizeUInt.parse(valueBytes)
+      val value1 = valueBytes
+        .drop(valueCmpctUInt.size)
+        .take(valueCmpctUInt.toLong)
+
+      new PSBTRecord {
+        override def key: ByteVector = key1
+
+        override def value: ByteVector = value1
+      }
```
nkohen: This breaks our ADT :(
`fromBytes` should probably return `(ByteVector, ByteVector)` so that it can be used in context

nkohen: And will need a scaladoc explaining what is returned

nkohen:
Also this probably needs fixing: https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/623965401#L476

PR #919, opened 2019-12-05
Decouple DLC tx construction and publishing to regtest

nkohen:
This is done by getting rid of `BitcoinP2PMessenger`


dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfIntegrationTest.scala
```diff
@@ -34,6 +35,16 @@ import scala.concurrent.Future
 class BinaryOutcomeDLCWithSelfIntegrationTest extends BitcoindRpcTest {
   private val clientsF = BitcoindRpcTestUtil.createNodePairV18(clientAccum)
   private val clientF = clientsF.map(_._1)
+  private val addressForMiningF = clientF.flatMap(_.getNewAddress)
+
+  def publishTransaction(tx: Transaction): Future[DoubleSha256DigestBE] = {
```
Christewart, line 40: :tada:

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -333,8 +332,7 @@ case class BinaryOutcomeDLCWithSelf(
       refundTx: Transaction
   )
 
-  def setupDLC(
-      messengerOpt: Option[BitcoinP2PMessenger] = None): Future[SetupDLC] = {
+  def setupDLC(): Future[SetupDLC] = {
     // Construct Funding Transaction
     createFundingTransaction.flatMap { fundingTx =>
```
Christewart, line 337: This seems like this should be moved into tests too?

PR #920, opened 2019-12-05
Timelock bug fix

nkohen:
Before this PR, the test case that is added here fails.

The problem was that the current default of `UInt32.zero` was being interpreted as a block height and since BIP 65 does not allow block height timelocks alongside timestamp ones this caused an error any time anyone tried using a second-timestamp timelock.

I also ran `test:scalafmt` and it seems a couple other files were hit by that in the diff

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
@@ -3,6 +3,7 @@ package org.bitcoins.core.wallet.builder
 import org.bitcoins.core.config.TestNet3
 import org.bitcoins.core.crypto.{
   BaseTxSigComponent,
+  DoubleSha256DigestBE,
   ECPrivateKey,
   WitnessTxSigComponentP2SH,
   WitnessTxSigComponentRaw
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -410,47 +410,54 @@ sealed abstract class BitcoinTxBuilder extends TxBuilder {
...
-              _: EmptySpendingInfo =>
-            // none of these scripts affect the locktime of a tx
-            loop(newRemaining, currentLockTime)
-        }
-    }
+                    } else Success(UInt32(cltv.locktime.toLong))
+                  val result = lockTime.flatMap { l: UInt32 =>
+                    currentLockTimeOpt match {
+                      case Some(currentLockTime) =>
+                        if (currentLockTime < l) {
```
benthecarman, line 432: Doesn't this assume that `l` is a block time and not a block height, if so wouldn't this break when a block height is given for `l` and `currentLockTime` is a block time

benthecarman (approved):

core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
```diff
@@ -387,6 +387,36 @@ class BitcoinTxBuilderTest extends BitcoinSAsyncTest {
...
+        Vector(
+          TransactionOutput(Bitcoins.one - CurrencyUnits.oneMBTC,
+                            EmptyScriptPubKey)),
+        Vector(cltvSpendingInfo),
+        SatoshisPerByte(Satoshis.one),
+        EmptyScriptPubKey,
+        RegTest
+      )
+
+    txBuilderF.flatMap(_.sign).map(_ => succeed)
```
Christewart: Shouldn't we verify this in the script interpreter?

core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
```diff
@@ -387,6 +387,36 @@ class BitcoinTxBuilderTest extends BitcoinSAsyncTest {
     }
   }
 
+  it must "succeed to sign a cltv spk that uses a second-based locktime" in {
```
Christewart, line 390: should we add test cases for block height as well? Do you think negative test cases are necessary to?

core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
nkohen: That shouldn't be necessary as we are only testing `TxBuilder.calcLockTime` here. That said, I did change this to verify that the transaction's `lockTime` is correct rather than just `suceed`ing

PR #923, opened 2019-12-06
DLC Justice Execution Case

nkohen:
Implemented justice transaction with passing positive tests on both RegTest and the ScriptInterpreter.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -372,14 +372,13 @@ case class BinaryOutcomeDLCWithSelf(
   }
 
   def constructClosingTx(
-      cetOutput: TransactionOutput,
       privKey: ECPrivateKey,
       spendingInfo: BitcoinUTXOSpendingInfo,
       isLocal: Boolean): Future[Transaction] = {
     // Construct Closing Transaction
     val txBuilder = BitcoinTxBuilder(
       Vector(
-        TransactionOutput(cetOutput.value,
+        TransactionOutput(spendingInfo.output.value,
                           P2PKHScriptPubKey(privKey.publicKey))),
       Vector(spendingInfo),
       feeRate,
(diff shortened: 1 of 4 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -500,6 +490,75 @@ case class BinaryOutcomeDLCWithSelf(
     }
   }
 
+  def executeJusticeDLC(
+      dlcSetup: SetupDLC,
+      timedOutCET: Transaction,
+      local: Boolean): Future[DLCOutcome] = {
+    val justiceOutput = timedOutCET.outputs.head
```
Christewart, line 496: can't we just use a `.find()` here instead of hard coding it? I guess this could probably be applied every where else in the file too where we use `.head` and `.last`?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -500,6 +490,75 @@ case class BinaryOutcomeDLCWithSelf(
...
+    )
+
+    val finalPrivKey = if (local) {
+      finalLocalPrivKey
+    } else {
+      finalRemotePrivKey
+    }
+
+    val justiceSpendingTxF =
+      constructClosingTx(justiceOutput,
```
Christewart: isn't it a bit redundant to pass in `justiceOutput` here when it is already inside of `justiceSpendingInfo`? It looks like `cetOuptut` isn't even really used in `constructClosingTx`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -500,6 +490,75 @@ case class BinaryOutcomeDLCWithSelf(
     }
   }
 
+  def executeJusticeDLC(
```
Christewart, line 492: scaladoc saying we are spending the timeout clause of the dlc contract

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfIntegrationTest.scala
```diff
@@ -46,6 +46,16 @@ class BinaryOutcomeDLCWithSelfIntegrationTest extends BitcoindRpcTest {
     } yield txid
   }
 
+  def waitUntilBlock(blockHeight: Int): Future[Unit] = {
```
Christewart, line 49: This seems like a pretty useful util that should probably be moved to `BitcoindRpcTestUtil`. I'd add a scaladoc saying that this will generate blocks until we reach the given `blockHeight`. This is useful for relative locktime schemes.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfIntegrationTest.scala
```diff
@@ -220,28 +230,60 @@ class BinaryOutcomeDLCWithSelfIntegrationTest extends BitcoindRpcTest {
...
+            setup.cetLoseLocal
+          }
+        }
+      }
+
+      for {
+        dlc <- constructDLC()
+        setup <- dlc.setupDLC()
+        _ <- publishTransaction(setup.fundingTx)
+        outcome <- dlc.executeJusticeDLC(setup, chooseCET(setup), local)
```
Christewart: so i think i am misunderstanding something here, the justice transaction should be spending from a CET right? From my understanding of this, we publish the funding transaction, and then attempt to spend from the funding transaction with a justice transaction? 

If I look below this, it does look like a CET is published, maybe what might make this clearer to me is if we publish a CET before determining the dlc outcome? Isn't that what would actually happen in the real world? If we are publishing a justice transaction, that presumably means that the counter party has cheated no?

nkohen: I'll reorder, but just to clarify things for you, a justice tx is not a CET, it spends a CET. Just like in lightning a justice tx isn't a commitment tx it spends one

PR #926, opened 2019-12-09
ChainQueryApi

rorp:
A layer of indirection to decouple `wallet` from `chain` and `node`.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
@@ -5,6 +5,7 @@ import org.bitcoins.chain.models.{
   CompactFilterDb,
   CompactFilterHeaderDb
 }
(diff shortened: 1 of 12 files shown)


chain/src/main/scala/org/bitcoins/chain/api/ChainApi.scala
```diff
@@ -186,4 +187,10 @@ trait ChainApi {
   /** Returns the block height of the given block stamp */
   def getHeightByBlockStamp(blockStamp: BlockStamp)(
       implicit ec: ExecutionContext): Future[Int]
+
+  /** @inheritdoc */
+  override def getBlockHeight(blockHash: DoubleSha256DigestBE)(
+      implicit ec: ExecutionContext): Future[Option[Int]] =
+    getHeader(blockHash).map(_.map(_.height))
```
benthecarman, line 194: Couldn't this be a `flatMap`

PR #928, opened 2019-12-09
BIP32Path Factory and subsequent tests

benthecarman:
Addresses https://github.com/bitcoin-s/bitcoin-s/pull/917#discussion_r354638548


core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -86,9 +88,16 @@ abstract class BIP32Path {
       }
       .fold("m")((accum, curr) => accum + "/" + curr)
 
+  def bytes: ByteVector =
+    path
```
Christewart: i think this can be simplified with 

```scala
path.foldLeft(ByteVector.empty)(_ ++_.toUInt32.bytes)
```

Issue #897, opened 2019-11-26
Fee modes in TxBuilder


Christewart:
With our work on DLCs (#895, #871) , it appears there are different types of fee schemes we want to use when signing transactions. 

The obvious one is for the situation we want to pay some one 1 BTC, and we have a 10 BTC utxo. This means that we need to create 2 outputs in `TxBuilder` where we pay 

1. 1 BTC output to Bob
2. (9 - `networkFee`) back to Alice via a change output.

There is another situation where we have a 10 BTC utxo and we want to pay 2 utxos and have NO CHANGE OUTPUT.  We want the fee to be evently distributed between Alice and Bob's utxos that are constructed in `TxBuilder`.

So our outputs would look like this:

1. (5BTC - `networkFee/2`) to Bob
2. (5BTC - `networkFee/2`) to Alice

It would be nice to natively support this in `TxBuilder`, so we don't have to do it ad-hoc like we did here: https://github.com/bitcoin-s/bitcoin-s/blob/b05d9b0021a95da9d0f38ed765862438349fd7fb/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala#L129

Christewart:
It looks like the bitcoin core `fundrawtransaction` does have a `subtractFeeFromOutputs` option too https://bitcoin.org/en/developer-reference#fundrawtransaction

PR #929, opened 2019-12-10
Refund fee tweak

nkohen:
Rather than having equal fees on refunds, parties now pay proportionally.


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -640,6 +640,45 @@ object BinaryOutcomeDLCWithSelf {
       newBuilder.flatMap(_.sign)
     }
   }
+
+  /** Subtracts the estimated fee by removing from each output proportionally */
+  def subtractFeeProportionalAndSign(txBuilder: BitcoinTxBuilder)(
```
Christewart, line 645: Is this something that should be addressed in #897?

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
     }
   }
 
+  def closeTogether(num1: Double, num2: Double): Boolean = {
+    Math.abs(num1 - num2) < 1.0e-6
+  }
+
+  it should "correctly subtract fees proportionally amongst outputs" in {
+    // Can't use TransactionGenerators.realisiticOutputs as that can return List.empty
+    val nonEmptyRealisticOutputsGen = Gen
```
Christewart:
```scala
    val nonEmptyRealisticOutputsGen = TransactionGenerators.realisticOutputs
      .suchThat(_.nonEmpty)
```

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
...
+      .choose(1, 5)
+      .flatMap(n => Gen.listOfN(n, TransactionGenerators.realisticOutput))
+      .suchThat(_.nonEmpty)
+
+    // CurrencyUnitGenerator.feeRate gives too high of fees
+    val feeRateGen = Gen.choose(0, CurrencyUnits.oneBTC.satoshis.toLong).map {
```
Christewart: I think this would probably be a useful generator to add in `CurrencyUnitGenerator`? Not sure what a good name is tho.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
...
+      n =>
+        SatoshisPerByte(Satoshis(Int64(n)))
```
Christewart: nit: `SatoshisPerByte(Satoshis(n))`

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
...
+          }
+
+          val (firstSize, firstFee) = sizeAndDiffs.head
+          val firstFeeProportion = firstFee.satoshis.toLong.toDouble / firstSize.satoshis.toLong
+          assert {
+            // Fee has been proportionally distributed (up to some remainder)
+            sizeAndDiffs.forall {
+              case (size, fee) =>
+                closeTogether(
+                  (fee.satoshis.toLong.toDouble / size.satoshis.toLong) / firstFeeProportion,
```
Christewart: whaa? How about some comments and names?

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
...
+            txBuilder)
+        } yield {
+          val sizeAndDiffs = outputs.zip(tx.outputs).map {
+            case (before, after) =>
+              (before.value, before.value - after.value)
+          }
+
+          val (firstSize, firstFee) = sizeAndDiffs.head
+          val firstFeeProportion = firstFee.satoshis.toLong.toDouble / firstSize.satoshis.toLong
+          assert {
```
Christewart: any reason we aren't using intermediate `val`s here?

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
```diff
@@ -93,6 +93,76 @@ class BinaryOutcomeDLCWithSelfTest extends BitcoinSAsyncTest {
     }
   }
 
+  def closeTogether(num1: Double, num2: Double): Boolean = {
+    Math.abs(num1 - num2) < 1.0e-6
```
Christewart, line 97: What is `1.0e-6` in terms of sats? Does this mean our remainder here would be theoretically less than 1 satoshi?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: yeah, essentially everything that ends up in this companion object will be moved in #897

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
nkohen: Better not to have a 20% failure with a `suchThat`

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfTest.scala
nkohen: These are unit-less numbers, not satoshis. We're just dealing with comparing proportions and there will be rounding error and also remainders (off by a bit because of discrete stuff) so I figured I'd just assert that things are close enough.

PR #931, opened 2019-12-10
Add fromBytesLE and bytesLE to NumberTypes

benthecarman:


core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -365,6 +371,10 @@ object UInt64
     UInt64(NumberUtil.toUnsignedInt(bytes))
   }
 
+  def fromBytesLE(bytes: ByteVector): UInt64 = {
```
Christewart: can you add this to `Factory`?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -83,6 +83,8 @@ sealed abstract class Number[T <: Number[T]]
   }
 
   override def bytes: ByteVector = BitcoinSUtil.decodeHex(hex)
+
+  def bytesLE: ByteVector = bytes.reverse
```
Christewart: can you add this to `NetworkElement`?

PR #932, opened 2019-12-10
Implement CurrencyUnit as Numeric

### LINKED ISSUE #930: Implement `.sum` on `CurrencyUnit`
Currently to sum a `Vector[CurrencyUnit]` we need to do 

```scala
satsVec.foldLeft(CurrencyUnits.zero) {
          case (accum, sats) =>
            accum + sats        
}
```

In normal Scala, we can sum a `Vector[Int]` with

```scala
@ val vec = Vector(1,2,3)
vec: Vector[Int] = Vector(1, 2, 3)
@ vec.sum
res1: Int = 6
```

It would be nice to implement `scala.math.Numeric` on `CurrencyUnit` so we can take advantage of this with our number types as well

benthecarman:
Closes #930

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
+++ b/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
@@ -7,7 +7,7 @@ import org.bitcoins.core.serializers.RawSatoshisSerializer
 import org.bitcoins.core.util.Factory
 import scodec.bits.ByteVector
 
-import scala.math.{Numeric, Ordering}
+import scala.math.Numeric
 import scala.util.{Failure, Success, Try}
 
 sealed abstract class CurrencyUnit
--- a/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
+++ b/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
@@ -17,27 +17,15 @@ sealed abstract class CurrencyUnit
 
   def satoshis: Satoshis
 
-  override def plus(x: CurrencyUnit, y: CurrencyUnit): CurrencyUnit =
-    Satoshis(x.satoshis.underlying + y.satoshis.underlying)
+  override def plus(x: CurrencyUnit, y: CurrencyUnit): CurrencyUnit = x + y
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -6,15 +6,52 @@ import org.bitcoins.core.protocol.NetworkElement
...
     extends NetworkElement
     with Ordered[CurrencyUnit]
-    with BasicArithmetic[CurrencyUnit] {
+    with BasicArithmetic[CurrencyUnit]
+    with Numeric[CurrencyUnit] {
   type A
 
   def satoshis: Satoshis
 
+  override def plus(x: CurrencyUnit, y: CurrencyUnit): CurrencyUnit =
```
Christewart: Your effictively implementing all of these twice. Can you delegate to the other ones? i.e. 

```scala
  override def +(c: CurrencyUnit): CurrencyUnit = {
    plus(this,c)
  }
```

core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -6,15 +6,52 @@ import org.bitcoins.core.protocol.NetworkElement
...
+  override def negate(x: CurrencyUnit): CurrencyUnit =
+    Satoshis(-x.satoshis.underlying)
+
+  override def fromInt(x: Int): CurrencyUnit = Satoshis(x.toLong)
+
+  override def parseString(str: String): Option[CurrencyUnit] = {
+    if (str.isEmpty) {
+      None
+    } else {
+      Some(Satoshis(str.toLong))
```
rorp: `toLong` will throw if `str` is not correctly formatted long integer. Is it the behavior we want?

benthecarman: This function was actually removed because it is no compatible with version `2.12.x`

Christewart:
Can we move this into here and have it work? https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/currency/package.scala

PR #934, opened 2019-12-10
Revert "BIP32Path Factory and subsequent tests"

benthecarman:
Reverts bitcoin-s/bitcoin-s#928

This wasn't supposed to be merged yet, #931 should be done first so this can implement it


benthecarman:
Oh wait actually it is fine

PR #938, opened 2019-12-11
Compute confirmations dynamically

rorp:
Currently confirmation numbers are stored in the wallet database and get updated by revealing UTXOs to the node. This PR makes the wallet compute confirmation numbers internally when they are needed.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -15,8 +15,8 @@ import org.bitcoins.core.crypto.{
 }
 import org.bitcoins.core.gcs.{BlockFilter, FilterHeader, FilterType}
 import org.bitcoins.core.p2p.CompactFilterMessage
-import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.blockchain.BlockHeader
+import org.bitcoins.core.protocol.{BitcoinAddress, BlockStamp}
 import org.bitcoins.core.util.CryptoUtil
 import org.bitcoins.testkit.BitcoinSTestAppConfig
 import org.bitcoins.testkit.chain.fixture.ChainFixtureTag
@@ -400,7 +400,7 @@ class ChainHandlerTest extends ChainUnitTest {
   it must "generate a range for a block filter query" in {
     chainHandler: ChainHandler =>
       for {
-        bestBlock <- chainHandler.getBestBlockHeader
+        bestBlock <- chainHandler.getBestBlockHeader()
         bestBlockHashBE = bestBlock.hashBE
         rangeOpt <- chainHandler.nextHeaderBatchRange(
           DoubleSha256DigestBE.empty,
@@ -410,6 +410,32 @@ class ChainHandlerTest extends ChainUnitTest {
         assert(rangeOpt.get._1 == 0)
         assert(rangeOpt.get._2 == bestBlockHashBE.flip)
       }
+  }
+
+  it must "return the number of confirmations" in {
+    chainHandler: ChainHandler =>
+      for {
+        bestBlockHashBE <- chainHandler.getBestBlockHash()
+        confirmations <- chainHandler.getNumberOfConfirmations(bestBlockHashBE)
+      } yield {
+        assert(confirmations == Some(1))
+      }
+  }
+
+  it must "return the height by block stamp" in { chainHandler: ChainHandler =>
+    for {
+      bestBlock <- chainHandler.getBestBlockHeader()
+      stamp1 = BlockStamp.BlockHash(bestBlock.hashBE)
+      stamp2 = BlockStamp.BlockHeight(bestBlock.height)
+      stamp3 = BlockStamp.BlockTime(bestBlock.time)
+      height1 <- chainHandler.getHeightByBlockStamp(stamp1)
+      height2 <- chainHandler.getHeightByBlockStamp(stamp2)
+      // TODO implement BlockTime
+//      height3 <- chainHandler.getHeightByBlockStamp(stamp3)
+    } yield {
+      assert(height1 == height2)
+//      assert(height1 == height3)
+    }
   }
 
   it must "match block filters" in { chainHandler: ChainHandler =>
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -14,6 +14,7 @@ import org.bitcoins.core.crypto.{
   ECPrivateKey
 }
 import org.bitcoins.core.gcs.{BlockFilter, FilterHeader, FilterType}
+import org.bitcoins.core.number.{Int32, UInt32}
 import org.bitcoins.core.p2p.CompactFilterMessage
 import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.protocol.{BitcoinAddress, BlockStamp}
@@ -51,7 +52,9 @@ class ChainHandlerTest extends ChainUnitTest {
   source.close()
 
   import org.bitcoins.rpc.serializers.JsonReaders.BlockHeaderReads
-  val headersResult = Json.parse(arrStr).validate[Vector[BlockHeader]].get
+
+  val headersResult: Vector[BlockHeader] =
+    Json.parse(arrStr).validate[Vector[BlockHeader]].get
 
   override val defaultTag: ChainFixtureTag = ChainFixtureTag.GenisisChainHandler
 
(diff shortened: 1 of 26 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -33,22 +33,35 @@ case class ChainHandler(
...
     }
   }
 
+  override def getBestBlockHeader(): Future[BlockHeaderDb] = {
+    for {
+      hash <- getBestBlockHash()
+      headerOpt <- getHeader(hash)
+    } yield headerOpt match {
+      case None =>
+        throw new RuntimeException(
```
Christewart, line 56: `Future.failed()` rather than a hard throw?

wallet-test/src/test/scala/org/bitcoins/wallet/api/CoinSelectorTest.scala
```diff
@@ -37,7 +38,8 @@ class CoinSelectorTest extends BitcoinSWalletTest {
       outPoint = TransactionGenerators.outPoint.sampleSome,
       output = TransactionOutput(10.sats, ScriptPubKey.empty),
       privKeyPath = WalletTestUtil.sampleSegwitPath,
-      scriptWitness = WitnessGenerators.scriptWitness.sampleSome
+      scriptWitness = WitnessGenerators.scriptWitness.sampleSome,
+      blockHash = Some(DoubleSha256DigestBE.empty)
```
Christewart: Can all of these be 

```scala
blockHash = None
```

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -20,32 +21,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
   /////////////////////
   // Public facing API
 
+  def getNumberOfConfirmations(
+      blockHashOpt: Option[DoubleSha256DigestBE]): Future[Int] =
+    blockHashOpt match {
+      case None => Future.successful(0)
```
Christewart: I think the public facing API should change here to be `Future[Option[Int]]` to indicate the block might not be found.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -20,32 +21,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+      case Some(blockHash) =>
+        for {
+          blockHeight <- chainQueryApi.getBlockHeight(blockHash)
+          tipHash <- chainQueryApi.getBestBlockHash()
```
Christewart: maybe add a helper method in `ChainQueryApi` called `getBestBlockHashHeight()` ?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -20,32 +21,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+          tipHeight <- chainQueryApi.getBlockHeight(tipHash)
+        } yield {
+          println(s"$blockHash $blockHeight $tipHash $tipHeight")
+          tipHeight
+            .flatMap(x => blockHeight.map(y => x - y + 1))
+            .getOrElse(0)
```
Christewart: if we use `Future[Option[Int]]` we can get rid of this too.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -193,47 +226,62 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
-        val errMsg =
-          Seq(
-            s"Found TXO has txid=${foundTxo.txid}, tx we were given has txid=${transaction.txIdBE}.",
-            "This is either a reorg or a double spent, which is not implemented yet"
-          ).mkString(" ")
-        logger.error(errMsg)
-        Future.failed(new RuntimeException(errMsg))
+      (foundTxo.blockHash, blockHashOpt) match {
+        case (None, Some(blockHash)) =>
+          logger.debug(
```
Christewart, line 215: it would be nice if there was some sort of an invariant here that we could put to make sure the `foundTxo` occurs in the `blockHashOpt`.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -245,7 +293,8 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
     */
   private def processNewIncomingTx(
       transaction: Transaction,
-      confirmations: Int): Future[Seq[SpendingInfoDb]] = {
+      confirmations: Int,
+      blockHashOpt: Option[DoubleSha256DigestBE]): Future[Seq[SpendingInfoDb]] = {
```
Christewart, line 268: So if `confirmations = 0`, `blockHashOpt = None` right? Maybe we should be making a helper case class like this:

```scala
/** This represents a block hash combined with the number of confirmations
  * a transaction should have relative to this block hash
  * 
  * I.e. 
  * If a transaction was included 10 blocks ago, this case class would be 
  * {{{
  *   BlockHashWithConfirmations(hash, 10)
  * }}}
  * 
  * */
case class BlockHashWithConfirmations(blockHash: DoubleSha256DigestBE, confirmations: Int)
```

At least I think that is what is intended here?

rorp: I'm going to remove `confirmations` soon. I still have them to let some of the tests pass.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -384,7 +390,7 @@ case class ChainHandler(
       endOpt: Option[BlockStamp] = None,
       batchSize: Int = chainConfig.filterBatchSize,
       parallelismLevel: Int = Runtime.getRuntime.availableProcessors())(
-      implicit ec: ExecutionContext): Future[Vector[DoubleSha256DigestBE]] = {
+      ec: ExecutionContext): Future[Vector[DoubleSha256DigestBE]] = {
```
Christewart, line 393: why no implicit here if the `ec` still needs to be a parameter?

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -274,10 +275,9 @@ object NodeUnitTest extends P2PLogger {
     for {
       bitcoind <- BitcoinSFixture.createBitcoindWithFunds(versionOpt)
       node <- createNeutrinoNode(bitcoind, callbacks)
-      chain <- node.chainApiFromDb()
       fundedWallet <- BitcoinSWalletTest.fundedWalletAndBitcoind(bitcoind,
                                                                  node,
-                                                                 chain)
+                                                                 node)
```
Christewart, line 280: same params here?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
rorp: How? We don't have access from `wallet` to the full blockchain. For neutrino mode it's done automatically though, because it receives a full block and `wallet` knows for sure that all transactions occur in that block.

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
rorp: Yes

rorp: Yes

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
rorp: The answer is in the scaladoc: this method performs better with a separate thread pool. But implicit `ec` clashes with the implicit `ExecutionContext` in the scope. This method is going to be moved to `wallet` soon anyway.

rorp: The answer is in the scaladoc: this method performs better with a separate thread pool. But implicit `ec` clashes with the implicit `ExecutionContext` in the scope. This method is going to be moved to `wallet` soon anyway.

rorp: The answer is in the scaladoc: this method performs better with a separate thread pool. But implicit `ec` clashes with the implicit `ExecutionContext` in the scope. This method is going to be moved to `wallet` soon anyway.

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
Christewart: Can you please start using [named arguments](https://docs.scala-lang.org/tour/named-arguments.html) in Scala so this is easier to code review, that would make this look like  

```scala
      fundedWallet <- BitcoinSWalletTest.fundedWalletAndBitcoind(bitcoindRpcClient = bitcoind,
                                                                 nodeApi = node,
                                                                 chainQueryApi = node)
```

which makes more sense to me! You can use if you click inside of the parameter list and press `alt+enter` in intellij they will automatically pop up for you.

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -412,6 +393,47 @@ class ChainHandlerTest extends ChainUnitTest {
...
+      }
+  }
+
+  it must "return the number of confirmations" in {
+    chainHandler: ChainHandler =>
+      for {
+        bestBlockHashBE <- chainHandler.getBestBlockHash()
+        confirmations <- chainHandler.getNumberOfConfirmations(bestBlockHashBE)
+      } yield {
+        assert(confirmations == Some(1))
```
benthecarman, line 417: this can be `confirmations.contains(1)`

rorp: I don't use `contains[A1 >: A](elem: A1)` because  the compiler can infer `A1` as `AnyVal` without warning. So `confirmations.contains("abc")` will compile, but it will always return `false` because `1 != "acb"`

PR #940, opened 2019-12-12
scalafmt On Compile

benthecarman:


.travis.yml
```diff
@@ -8,62 +8,71 @@ jobs:
       os: linux
       name: "Compile & Formatting Check"
       env:
+        - CI="1"
```
Christewart: @nkohen thought this was setup for the whole environment? Can't we configure these to be matrix wide?

benthecarman: Looks like `CI` is set to `true` automatically for us, fixed to use this.

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

benthecarman: Looks like it didn't work :(

https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/701135720#L765

PR #950, opened 2019-12-16
DLC: Two Change SPKs

nkohen:
Added change spk so both local and remote have one

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -15,6 +15,7 @@ import org.bitcoins.core.protocol.BlockStampWithFuture
 import org.bitcoins.core.protocol.script.{
   CLTVScriptPubKey,
   ConditionalScriptPubKey,
+  EmptyScriptPubKey,
   MultiSignatureScriptPubKey,
   MultiSignatureWithTimeoutScriptPubKey,
   P2PKHScriptPubKey,
@@ -64,7 +65,8 @@ import scala.concurrent.{ExecutionContext, Future}
   * @param localLosePayout Local's payout in the Lose case
   * @param timeout The CLTV timeout in milliseconds used in all CETs
   * @param feeRate The predicted fee rate used for all transactions
-  * @param changeSPK The place-holder change ScriptPubKey used for all transactions
+  * @param localChangeSPK Local's change ScriptPubKey used in the funding tx
+  * @param remoteChangeSPK Remote's change ScriptPubKey used in the funding tx
   */
 case class BinaryOutcomeDLCWithSelf(
     outcomeWin: String,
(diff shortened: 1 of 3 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -150,15 +153,22 @@ case class BinaryOutcomeDLCWithSelf(
...
+  private val remoteFunding =
+    remoteFundingUtxos.foldLeft(0L)(_ + _.amount.satoshis.toLong)
+
   /** Remote's payout in the Win case (in which Remote loses) */
   val remoteWinPayout: CurrencyUnit = totalInput - localWinPayout
 
   /** Remote's payout in the Lose case (in which Remote wins) */
   val remoteLosePayout: CurrencyUnit = totalInput - localLosePayout
 
+  private val emptyChangeSPK: ScriptPubKey = EmptyScriptPubKey
```
Christewart, line 175: Why is this needed now?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -172,12 +182,20 @@ case class BinaryOutcomeDLCWithSelf(
...
+      TransactionOutput(Satoshis(localFunding) - localInput, localChangeSPK)
+    val remoteChange =
+      TransactionOutput(Satoshis(remoteFunding) - remoteInput, remoteChangeSPK)
 
-    val outputs: Vector[TransactionOutput] = Vector(output)
+    val outputs: Vector[TransactionOutput] =
+      Vector(output, localChange, remoteChange)
     val txBuilderF: Future[BitcoinTxBuilder] =
-      BitcoinTxBuilder(outputs, fundingUtxos, feeRate, changeSPK, network)
+      BitcoinTxBuilder(outputs, fundingUtxos, feeRate, emptyChangeSPK, network)
```
Christewart, line 198: So I think i understand why this is needed, but it seems like the better way to handle this is with an `Option` type where the `TxBuilder` fails if there is change. If we have a bug where we send money to the empty changeSPK that means that money is lost forever. This is probably something else that should be addressed in #897 

One way I would maybe feel a bit more comfortable with this is if we use the [`sign()`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala#L216) method that allows you specify an invariant about the resulting transaction. The invariant in this case would be that we do NOT have an output that is the empty SPK.

dlc/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelfIntegrationTest.scala
```diff
@@ -189,7 +191,8 @@ class BinaryOutcomeDLCWithSelfIntegrationTest extends BitcoindRpcTest {
         localLosePayout = localInput - CurrencyUnits.oneMBTC,
         timeout = tomorrowInBlocks,
         feeRate = feeRate,
-        changeSPK = changeSPK,
+        localChangeSPK = localChangeSPK,
+        remoteChangeSPK = remoteChangeSPK,
         network = RegTest
```
Christewart, line 196: We probably should have assertions in this test case that the resulting signed tx does not contain the empty spk. Same with `BinaryOutcomeDLCWithSelfTest`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: Added the invariant for now and opened #961 for later

PR #952, opened 2019-12-16
DLC Up-Front Fees

nkohen:
Built on top of #951 

CET fees are now computed in advance and added to the funding tx so that users get the nice UX of having CET outputs show the exact values agreed to in the setup phase.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -2,6 +2,7 @@ package org.bitcoins.dlc
 
 import org.bitcoins.core.config.BitcoinNetwork
 import org.bitcoins.core.crypto.{
+  DoubleSha256DigestBE,
   ECPrivateKey,
   ECPublicKey,
   ExtPrivateKey,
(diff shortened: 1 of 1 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -183,21 +184,48 @@ case class BinaryOutcomeDLCWithSelf(
...
-
-    txBuilderF.flatMap { txBuilder =>
-      subtractFeeFromOutputsAndSign(txBuilder,
-                                    Vector(localChangeSPK, remoteChangeSPK))
+
+    val emptyOutPoint =
+      TransactionOutPoint(DoubleSha256DigestBE.empty, UInt32.zero)
+    val dummySpendingInfo = P2WSHV0SpendingInfo(
+      outPoint = emptyOutPoint,
+      amount = totalInput * 2,
```
Christewart: Why is this multiplied by 2?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -183,21 +184,48 @@ case class BinaryOutcomeDLCWithSelf(
...
+      scriptPubKey = P2WSHWitnessSPKV0(fundingSPK),
+      signers = Vector(fundingLocalPrivKey, fundingRemotePrivKey),
+      hashType = HashType.sigHashAll,
+      scriptWitness = P2WSHWitnessV0(fundingSPK),
+      conditionalPath = ConditionalPath.NoConditionsLeft
+    )
+    val cetWinLocalF = createCETWinLocal(dummySpendingInfo)
```
Christewart: There is symmetry between `cetWinLocal` and `cetLoseRemote` right? I guess to me it seems that this scheme is not going to work when parties are trying to "come to consensus" on what fees are. There are 2 different outcomes in the binary DLC case, is there a reason you picked `CETWinLocal` as the one to estimate fees on and not `CETLoseLocal`? Comments are helpeful here.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -183,21 +184,48 @@ case class BinaryOutcomeDLCWithSelf(
...
+    val cetFeeF = cetWinLocalF.map { tx =>
+      feeRate.calc(tx._1)
+    }
+
+    cetFeeF.flatMap { cetFee =>
+      val halfCetFee = Satoshis(cetFee.satoshis.toLong / 2)
+
+      val output: TransactionOutput =
+        TransactionOutput(totalInput + cetFee, P2WSHWitnessSPKV0(fundingSPK))
```
Christewart, line 210: It would be good to have some comment to narrate what is going on in here. Here is how I understand it:

>We need to commit to the fee to execute the `cet` during the construction of the funding transaction. Our approach at the moment is to estimate the currenty blockchain's fee rate, and then approximate how many bytes the `cetWinLocal` transaction will be (why did we choose this one?). We then add that amount to the `fundingOutput` so it can be used for fees later.

Is this right?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: It must be greater than `totalInput` since we now add fees to funding rather than subtracting them from CET, I could add 10000 or something like that but I figured this would be nicer

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: In my testing, all of the CETs had the same fees since they are all the same size (since they are all the same contract). Refund is a couple sats fewer as it is shorter. I'll add a comment explaining this though. (also keep in mind this is a fee estimation, we're almost never gonna get it right on the dot so if there's like a couple bytes difference this should be fine)

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: yup, with the only additional comment that this is all done so that the CETs output the amount specified in the contract and not less.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: Also to your point about fee consensus, the initiator will specify fees over messages in the future. Or something of this nature.

Christewart (approved):

PR #954, opened 2019-12-17
Add invariant to WitnessTransaction that says inputs.length == witnes…

Christewart:
…ses.length

From [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#specification) 

> The witness is a serialization of all witness data of the transaction. **Each txin is associated with a witness field**. A witness field starts with a var_int to indicate the number of stack items for the txin. It is followed by stack items, with each item starts with a var_int to indicate the length. Witness data is NOT script. 

This creates an invariant that enforces that inside of `WitnessTransaction`.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionWitness.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionWitness.scala
@@ -75,10 +75,6 @@ object EmptyWitness {
   def fromInputs(inputs: Seq[TransactionInput]): EmptyWitness = {
     fromN(inputs.length)
   }
-
-  /** Used in some test cases, this is only safe to use with transactions
-    * if your transaction has 0 inputs */
-  val empty: EmptyWitness = fromN(0)
 }
 
 object TransactionWitness {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionWitness.scala
```diff
@@ -55,10 +55,30 @@ sealed abstract class TransactionWitness extends NetworkElement {
...
+    new EmptyWitness(wits)
+  }
+
+  def fromInputs(inputs: Seq[TransactionInput]): EmptyWitness = {
+    fromN(inputs.length)
+  }
+
+  /** Used in some test cases, this is only safe to use with transactions
+    * if your transaction has 0 inputs */
+  val empty: EmptyWitness = fromN(0)
```
benthecarman: It might be better to remove this and just do `.fromN(0)` in the test cases so people don't accidentally use this outside of tests

core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionWitness.scala
```diff
@@ -69,7 +89,8 @@ object TransactionWitness {
     if (witnesses.exists(_ != EmptyScriptWitness)) {
       TransactionWitnessImpl(witnesses)
     } else {
-      EmptyWitness
+      //means that everything must be a empty ScriptWitness
+      EmptyWitness(witnesses.map(_.asInstanceOf[EmptyScriptWitness.type]))
```
nkohen: nit: no strong feelings about this but you could also just do `EmptyWitness.fromN(witnesses.length)` to avoid casting here

benthecarman (approved):

PR #955, opened 2019-12-17
PSBT Updater

benthecarman:
The test case follows the example in BIP 174 starting at `Given the above PSBT, an updater with only the following: `


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -33,6 +36,11 @@ case class PSBT(
   require(inputMaps.size == transaction.inputs.size)
   require(outputMaps.size == transaction.outputs.size)
 
+  // Need to define these so when we compare the PSBTs
+  // the map lexicographical ordering is enforced
+  def ==(p: PSBT): Boolean = this.bytes == p.bytes
```
nkohen, line 46: Does it do the trick to just override `equals`? I could be wrong on this

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+    * @param tx Transaction to add to PSBT
+    * @param i index of inputMap to add tx to
+    * @return PSBT with added tx
+    */
+  def addTransaction(tx: Transaction, i: Int): PSBT = {
+    require(
+      i < inputMaps.size,
+      s"i must be less than the number of input maps present in the psbt, $i >= ${inputMaps.size}")
+    if (inputMaps(i).isFinalized) {
+      this
```
nkohen: Shouldn't you ensure that the `tx` doesn't conflict with `this.inputMaps` and throw if it does?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
       vec.head.version
     }
   }
+
+  /**
+    * @param tx Transaction to add to PSBT
+    * @param i index of inputMap to add tx to
+    * @return PSBT with added tx
+    */
+  def addTransaction(tx: Transaction, i: Int): PSBT = {
```
nkohen: I think it would be nice to rename this to be more descriptive of what is happening here since I could mis-interpret this to mean messing with the underlying unsigned tx of the PSBT, also a scaladoc describing what you mean by `add` would be nice

nkohen: maybe `addTransactionToInputs` or something along these lines if I understand what is happening here correctly

nkohen: also renaming `i` to `index` would make this easier to understand at first glance

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+      PSBT(globalMap, newInputMaps, outputMaps)
+    }
+  }
+
+  /**
+    * @param script ScriptPubKey to add to PSBT
+    * @param i index of inputMap to add script to
+    * @return PSBT with added script
+    */
+  def addScriptToInput(script: ScriptPubKey, i: Int): PSBT = {
```
nkohen: similar to above, `i -> index` and would be nice to have a more descriptive scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+      PSBT(globalMap, newInputMaps, outputMaps)
+    }
+  }
+
+  /**
+    * @param script ScriptPubKey to add to PSBT
+    * @param i index of outputMap to add script to
+    * @return PSBT with added script
+    */
+  def addScriptToOutput(script: ScriptPubKey, i: Int): PSBT = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+    }
+  }
+
+  /**
+    * @param extKey ExtKey to derive key from
+    * @param path path of key to add to PSBT
+    * @param i index of inputMap to add the BIP32Path to
+    * @return PSBT with added BIP32Path
+    */
+  def addKeyPathToInput(extKey: ExtKey, path: BIP32Path, i: Int): PSBT = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+    }
+  }
+
+  /**
+    * @param extKey ExtKey to derive key from
+    * @param path path of key to add to PSBT
+    * @param i index of outputMap to add the BIP32Path to
+    * @return PSBT with added BIP32Path
+    */
+  def addKeyPathToOutput(extKey: ExtKey, path: BIP32Path, i: Int): PSBT = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+      PSBT(globalMap, inputMaps, newOutputMaps)
+    }
+  }
+
+  /**
+    * @param hashType HashType to add to the input
+    * @param i index of the input to add the HashType to
+    * @return PSBT with added HashType
+    */
+  def addSigHashTypeToInput(hashType: HashType, i: Int): PSBT = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -605,6 +817,35 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
       .filter(element => PSBTInputKeyId.fromByte(element.key.head) == key)
       .asInstanceOf[Vector[T]]
   }
+
+  def isFinalized: Boolean =
+    getRecords(FinalizedScriptSigKeyId).nonEmpty || getRecords(
+      FinalizedScriptWitnessKeyId).nonEmpty
+
+  def slimMap(txIn: TransactionInput): InputPSBTMap = {
```
nkohen: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +68,211 @@ case class PSBT(
...
+        outputMaps(i).getRecords[psbt.OutputPSBTRecord.RedeemScript](
+          PSBTOutputKeyId.RedeemScriptKeyId)
+      val hasWitScript = redeemScript.size == 1 && WitnessScriptPubKey
+        .isWitnessScriptPubKey(redeemScript.head.redeemScript.asm)
+
+      if (!isWitScript && hasWitScript) {
+        elements = elements.filterNot(_.key.startsWith(hex"01")) :+ OutputPSBTRecord
+          .WitnessScript(script)
+      } else {
+        elements = elements.filterNot(_.key.startsWith(hex"00")) :+ OutputPSBTRecord
```
nkohen: I think for all of these `key.startsWith` wit would be nice to have some simple `ADT` someplace wrapping these bytes so that we could give them nice readable names rather than using numbers

nkohen (commented):
I'll likely have more review on the next iteration but I had some trouble understanding what these functions intend to do since you now know more about PSBTs than me and I need scaladocs :P

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Tried it, doesn't work :/

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +67,218 @@ case class PSBT(
...
+          .size == 1
+
+      if (outIsWitnessScript || hasWitScript) {
+        elements = elements.filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
+          out)
+      } else {
+        elements = elements.filterNot(_.key.head == NonWitnessUTXOKeyId.byte) :+ NonWitnessOrUnknownUTXO(
+          tx)
+      }
+    }
```
nkohen: Rather than using a `var` for elements, I think it would be more idiomatic to do `val elements = if ...` and then just drop the `elements =` on the ends, and rather than elements call `inputMaps(index).elements`, and also add an else case which just returns `inputMaps(index).elements`. Or something of this nature

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +67,218 @@ case class PSBT(
...
+    * @return PSBT with added script
+    */
+  def addScriptToInput(script: ScriptPubKey, index: Int): PSBT = {
+    require(
+      index < inputMaps.size,
+      s"i must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    var elements = inputMaps(index).elements
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: you can also give `inputMaps(index).elements` a `val` if it comes up enough

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +67,218 @@ case class PSBT(
...
+    * @param index index of the OutputPSBTMap to add script to
+    * @return PSBT with added script
+    */
+  def addScriptToOutput(script: ScriptPubKey, index: Int): PSBT = {
+    require(
+      index < outputMaps.size,
+      s"i must be less than the number of output maps present in the psbt, $index >= ${outputMaps.size}")
+    require(!isFinalized, "Cannot update a PSBT that is finalized")
+
+    var elements = outputMaps(index).elements
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +67,218 @@ case class PSBT(
...
+    */
+  def addKeyPathToInput(extKey: ExtKey, path: BIP32Path, index: Int): PSBT = {
+    require(
+      index < inputMaps.size,
+      s"i must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val map = inputMaps(index)
+    var elements = map.elements
```
nkohen: here too

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,11 +67,218 @@ case class PSBT(
...
+    * @return PSBT with added BIP32Path
+    */
+  def addKeyPathToOutput(extKey: ExtKey, path: BIP32Path, index: Int): PSBT = {
+    require(
+      index < outputMaps.size,
+      s"i must be less than the number of output maps present in the psbt, $index >= ${outputMaps.size}")
+    require(!isFinalized, "Cannot update a PSBT that is finalized")
+
+    val map = outputMaps(index)
+    var elements = map.elements
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -605,6 +823,43 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+    * 1. Compresses the size of the data
+    * 2. Allows for segwit only compatibility
+    * @param txIn The TransactionInput that this InputPSBTMap represents
+    * @return The compressed InputPSBTMap
+    */
+  def compressMap(txIn: TransactionInput): InputPSBTMap = {
+    if (isFinalized) {
+      this
+    } else {
+      var newElements = elements
```
nkohen: same kinda thing here

nkohen (approved):

benthecarman:
Rebased

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -59,9 +68,8 @@ case class PSBT(
     }
   }
 
-  def isFinalized: Boolean = inputMaps.forall(_.isFinalized)
-
   /**
+<<<<<<< HEAD
```
nkohen: nit: I think you forgot to delete this

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -82,11 +90,231 @@ case class PSBT(
...
+    * Adds tx to the indexed InputPSBTMap to either the NonWitnessOrUnknownUTXO
+    * or WitnessUTXO field depending on the tx and available information in the PSBT
+    * @param tx Transaction to add to PSBT
+    * @param index index of the InputPSBTMap to add tx to
+    * @return PSBT with added tx
+    */
+  def addTransactionToInput(tx: Transaction, index: Int): PSBT = {
+    require(
+      index < inputMaps.size,
+      s"i must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
```
nkohen: nit: still says `i` instead of `index` in these error messages (below too)

nkohen (approved):

PR #956, opened 2019-12-17
PSBT Combiner

benthecarman:


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,6 +58,20 @@ case class PSBT(
       vec.head.version
     }
   }
+
+  def isFinalized: Boolean = inputMaps.forall(_.isFinalized)
+
+  def combinePSBT(other: PSBT): PSBT = {
```
nkohen, line 71: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -576,6 +590,13 @@ case class GlobalPSBTMap(elements: Vector[GlobalPSBTRecord]) extends PSBTMap {
       .filter(element => PSBTGlobalKeyId.fromByte(element.key.head) == key)
       .asInstanceOf[Vector[T]]
   }
+
+  def combine(other: GlobalPSBTMap): GlobalPSBTMap = {
```
nkohen, line 609: scaladocs for all of these and I would prefer more descriptive names to overloading

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -576,6 +590,13 @@ case class GlobalPSBTMap(elements: Vector[GlobalPSBTRecord]) extends PSBTMap {
...
+    val newElements: Vector[GlobalPSBTRecord] =
+      other.elements.filterNot(element =>
+        this.elements.exists(_.key == element.key))
```
nkohen: here and with all of these functions, I believe it is simpler to just call `(this.elements ++ other.elements).distinct`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,6 +58,20 @@ case class PSBT(
...
   }
+
+  def isFinalized: Boolean = inputMaps.forall(_.isFinalized)
+
+  def combinePSBT(other: PSBT): PSBT = {
+    require(other.transaction.txId == this.transaction.txId,
+            "Can only combine PSBTs with the same global transaction.")
+
+    val global = globalMap.combine(other.globalMap)
+    val inputs = inputMaps.zipWithIndex.map(input =>
```
nkohen: rather than using `zipWithIndex` when you have two collections of the same size, you can actually just call `inputMaps.zip(other.inputMaps)` and get a Vector where they are paired by place/index

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -0,0 +1,20 @@
+package org.bitcoins.core.psbt
+
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import scodec.bits._
+
+class PSBTTest extends BitcoinSUnitTest {
+
+  it must "successfully combine two PSBTs" in {
```
nkohen, line 8: Not saying it has to happen on this PR but it might be nice to get in some property based tests for things like this once we have nice generators for PSBTs and PSBT components

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Wasn't really sure what a more descriptive name could be, but I think the added scaladocs should better explain

nkohen (approved):

PR #959, opened 2019-12-17
2019 12 17 sign ext key

Christewart:
This PR creates `ExtSign` which is a version of `Sign` that is meant to work with `ExtKey`. Now you shouldn't have to access the private/public key yourself when using `ExtKey`. You can now just pass a `ExtKey` into `TxBuilder`. 

One note: It seems that sign is painfully slow. 

```
sbt:bitcoin-s-core-test> testOnly *ExtSignTest*
[info] ExtSignTest:
[info] - must be able to sign something that extends ExtSignKey
[info] ScalaTest
[info] Run completed in 24 seconds, 909 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 1, Failed 0, Errors 0, Passed 1
[success] Total time: 27 s, completed Dec 17, 2019 9:18:26 AM
```

I'm not sure why it takes 24 seconds to sign 20 times in the property based test. My first guess is slow generators, but i don't have time to look into it for now.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/crypto/ExtSignTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/crypto/ExtSignTest.scala
@@ -1,15 +1,27 @@
 package org.bitcoins.core.crypto
 
-import org.bitcoins.testkit.core.gen.CryptoGenerators
-import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, HDGenerators}
+import org.bitcoins.testkit.util.{BitcoinSAsyncTest, BitcoinSUnitTest}
 
-class ExtSignTest extends BitcoinSUnitTest {
+class ExtSignTest extends BitcoinSAsyncTest {
 
   it must "be able to sign something that extends ExtSignKey" in {
     forAll(CryptoGenerators.extPrivateKey, CryptoGenerators.sha256Digest) {
       case (extPrivKey, hash) =>
         val sig = extPrivKey.sign(hash.bytes)
         assert(extPrivKey.publicKey.verify(hash,sig))
+    }
+  }
+
+  it must "be able to sign a specific path of a ext key" in {
+    forAllAsync(CryptoGenerators.extPrivateKey, CryptoGenerators.sha256Digest, HDGenerators.bip32Path) {
+      case (extPrivKey, hash, path) =>
+        val sigF = extPrivKey.sign(hash.bytes,path)
+        val childPubKey = extPrivKey.deriveChildPubKey(path).get
+        sigF.map { sig =>
+          assert(childPubKey.key.verify(hash,sig))
+        }
+
     }
   }
 }
--- a/core-test/src/test/scala/org/bitcoins/core/crypto/ExtSignTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/crypto/ExtSignTest.scala
@@ -16,7 +16,7 @@ class ExtSignTest extends BitcoinSAsyncTest {
   it must "be able to sign a specific path of a ext key" in {
     forAllAsync(CryptoGenerators.extPrivateKey, CryptoGenerators.sha256Digest, HDGenerators.bip32Path) {
       case (extPrivKey, hash, path) =>
-        val sigF = extPrivKey.sign(hash.bytes,path)
+        val sigF = extPrivKey.deriveAndSignFuture(hash.bytes,path)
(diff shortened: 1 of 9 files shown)


core-test/src/test/scala/org/bitcoins/core/crypto/ExtSignTest.scala
```diff
@@ -0,0 +1,27 @@
...
+        val sig = extPrivKey.sign(hash.bytes)
+        assert(extPrivKey.publicKey.verify(hash,sig))
+    }
+  }
+
+  it must "be able to sign a specific path of a ext key" in {
+    forAllAsync(CryptoGenerators.extPrivateKey, CryptoGenerators.sha256Digest, HDGenerators.bip32Path) {
+      case (extPrivKey, hash, path) =>
+        val sigF = extPrivKey.sign(hash.bytes,path)
+        val childPubKey = extPrivKey.deriveChildPubKey(path).get
```
Christewart, line 20: I think this needs to be revisted when this is fixed: https://github.com/bitcoin-s/bitcoin-s/issues/969

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -58,3 +59,17 @@ object Sign {
     }, publicKey)
   }
 }
+
+/** A signing interface for [[ExtKey]] */
+trait ExtSign extends Sign { this: ExtPrivateKey =>
```
nkohen: I don't think it is good to require that all `ExtSign`s be `ExtPrivateKey`s and that instead you should make the `def sign` into a `def deriveAndSignFunction: (ByteVector, BIP32Path) => Future[ECDigitalSignature]` and implement this function on `ExtPrivateKey`. This way things will be consistent with the `Sign` trait.

nkohen: I also think that we need consistency between the `ExtPrivateKey - ECPrivateKey` and the `ExtSign - Sign` relationships. Either both `ExtPrivateKey` and `ExtSign` has-a `ECPrivateKey` and `Sign`, or both is-a. Currently we have `ExtSign is-a Sign` and `ExtPrivateKey has-a ECPrivateKey`

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -58,3 +59,14 @@ object Sign {
...
 }
+
+/** A signing interface for [[ExtKey]] */
+trait ExtSign extends Sign {
+
+  def deriveAndSignFuture: (ByteVector, BIP32Path) => Future[ECDigitalSignature]
+
+  /** First derives the child key that corresponds to [[BIP32Path path]] and then signs */
+  def sign(bytes: ByteVector, path: BIP32Path): ECDigitalSignature = {
+    Await.result(deriveAndSignFuture(bytes, path), 30.seconds)
```
nkohen, line 70: I don't think I like having this non-future version here, shouldn't the decision to block be left to the caller?

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -58,3 +59,14 @@ object Sign {
     }, publicKey)
   }
 }
+
+/** A signing interface for [[ExtKey]] */
+trait ExtSign extends Sign {
```
nkohen, line 64: I still think that either `ExtSign` should have a `Sign` rather than extending or `ExtPrivateKey` should extend `ECPrivateKey` rather than having one

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
Christewart: This is following the pattern in `Sign`: https://github.com/bitcoin-s/bitcoin-s/blob/74f7f73797c04924994da0c6c1ee12e47aacbbfc/core/src/main/scala/org/bitcoins/core/crypto/Sign.scala#L28

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
Christewart: We can revisit this later, as [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format) does say that `ExtKey` does have a `BaseECKey`

>33 bytes: the public key or private key data (serP(K) for public keys, 0x00 || ser256(k) for private keys)

PR #964, opened 2019-12-18
No-Multisig CET

nkohen:
Replace multisig with oracle pub key to addition of pub key with oracle pub key.

In the next PR I'll be introducing a new ScriptPubKey for CET outputs.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -1,5 +1,6 @@
 package org.bitcoins.dlc
 
+import org.bitcoin.NativeSecp256k1
 import org.bitcoins.core.config.BitcoinNetwork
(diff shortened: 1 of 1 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -251,14 +252,18 @@ case class BinaryOutcomeDLCWithSelf(
       cetRemotePrivKey.deriveChildPrivKey(UInt32.one).key
     val toRemotePrivKey = cetRemotePrivKey.deriveChildPrivKey(UInt32(2)).key
 
-    val multiSig = MultiSignatureScriptPubKey(
-      requiredSigs = 2,
-      pubKeys = Vector(localToLocalPrivKey.publicKey, sigPubKey))
+    val pubKeyBytes = NativeSecp256k1.pubKeyTweakAdd(
+      sigPubKey.bytes.toArray,
```
Christewart, line 256: named args here please, `true` isn't very descriptive

PR #967, opened 2019-12-18
P2PKWithTimeoutSPK

nkohen:
Introduces a new SPK type used by [DLCs](https://github.com/bitcoin-s/dlcspecs/blob/master/Transactions.md#CetOutputs)

The new SPK is of the following form:
```
OP_IF
    BytesToPushOntoStackImpl(33)
    ScriptConstantImpl(ByteVector(33 bytes, pubKeyBytes))
OP_ELSE
    BytesToPushOntoStackImpl(4)
    ScriptConstantImpl(ByteVector(4 bytes, timeLockBytes))
    OP_CHECKLOCKTIMEVERIFY
    OP_DROP

    BytesToPushOntoStackImpl(33)
    ScriptConstantImpl(ByteVector(33 bytes, timeoutPubKeyBytes))
OP_ENDIF

OP_CHECKSIG
```

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -912,14 +912,14 @@ object P2PKWithTimeoutScriptPubKey
       .calculatePushOp(timeoutPubKey.bytes)
       .toVector ++ Vector(ScriptConstant(timeoutPubKey.bytes))
 
-    P2PKWithTimeoutScriptPubKeyImpl {
+    P2PKWithTimeoutScriptPubKeyImpl(
       Vector(Vector(OP_IF),
              pubKeyAsm,
              Vector(OP_ELSE),
              timeoutAsm,
              timeoutPubKeyAsm,
              Vector(OP_ENDIF, OP_CHECKSIG)).flatten
-    }
+    )
   }
 
   def isP2PKWithTimeoutScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -866,6 +866,15 @@ object NonStandardNotIfConditionalScriptPubKey
   }
 }
 
+/** The type for ScriptPubKeys of the form:
+  * OP_IF
+  *   <Public Key>
+  * OP_ELSE
+  *   <Timeout> OP_CHECKLOCKTIMEVERIFY OP_DROP
+  *   <Timeout Public Key>
+  * OP_ENDIF
+  * OP_CHECKSIG
+  */
 sealed trait P2PKWithTimeoutScriptPubKey extends RawScriptPubKey {
 
   lazy val pubKey: ECPublicKey =
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -866,6 +866,70 @@ object NonStandardNotIfConditionalScriptPubKey
   }
 }
 
+sealed trait P2PKWithTimeoutScriptPubKey extends RawScriptPubKey {
+
+  val pubKey: ECPublicKey =
+    ECPublicKey.fromBytes(asm(2).bytes)
+
+  val lockTime: ScriptNumber = ScriptNumber.fromBytes(asm(5).bytes)
+
+  val timeoutPubKey: ECPublicKey =
+    ECPublicKey.fromBytes(asm(9).bytes)
+}
+
+object P2PKWithTimeoutScriptPubKey
+    extends ScriptFactory[P2PKWithTimeoutScriptPubKey] {
+  private case class P2PKWithTimeoutScriptPubKeyImpl(asm: Vector[ScriptToken])
+      extends P2PKWithTimeoutScriptPubKey
+
+  override def fromAsm(asm: Seq[ScriptToken]): P2PKWithTimeoutScriptPubKey = {
+    buildScript(
+      asm = asm.toVector,
+      constructor = P2PKWithTimeoutScriptPubKeyImpl.apply,
+      invariant = isP2PKWithTimeoutScriptPubKey,
+      errorMsg = s"Given asm was not a P2PKWithTimeoutScriptPubKey, got $asm"
+    )
+  }
+
+  def apply(
+      pubKey: ECPublicKey,
+      lockTime: ScriptNumber,
+      timeoutPubKey: ECPublicKey): P2PKWithTimeoutScriptPubKey = {
+    val timeoutAsm = CLTVScriptPubKey(lockTime, EmptyScriptPubKey).asm.toVector
+    val pubKeyAsm = BitcoinScriptUtil
+      .calculatePushOp(pubKey.bytes)
+      .toVector ++ Vector(ScriptConstant(pubKey.bytes))
+    val timeoutPubKeyAsm = BitcoinScriptUtil
+      .calculatePushOp(timeoutPubKey.bytes)
+      .toVector ++ Vector(ScriptConstant(timeoutPubKey.bytes))
+
+    P2PKWithTimeoutScriptPubKeyImpl {
+      Vector(Vector(OP_IF),
+             pubKeyAsm,
+             Vector(OP_ELSE),
+             timeoutAsm,
+             timeoutPubKeyAsm,
+             Vector(OP_ENDIF, OP_CHECKSIG)).flatten
+    }
+  }
+
+  def isP2PKWithTimeoutScriptPubKey(asm: Seq[ScriptToken]): Boolean = {
+    if (asm.length == 12) {
+      val pubKey = ECPublicKey.fromBytes(asm(2).bytes)
+      val lockTimeTry = Try(ScriptNumber.fromBytes(asm(5).bytes))
+      val timeoutPubKey = ECPublicKey.fromBytes(asm(9).bytes)
+
+      lockTimeTry match {
+        case Success(lockTime) =>
+          asm == P2PKWithTimeoutScriptPubKey(pubKey, lockTime, timeoutPubKey).asm
+        case Failure(_) => false
+      }
+    } else {
+      false
+    }
+  }
+}
+
 sealed trait NonStandardScriptPubKey extends RawScriptPubKey
 
 object NonStandardScriptPubKey extends ScriptFactory[NonStandardScriptPubKey] {
@@ -896,6 +960,8 @@ object RawScriptPubKey extends ScriptFactory[RawScriptPubKey] {
 
   def fromAsm(asm: Seq[ScriptToken]): RawScriptPubKey = asm match {
     case Nil => EmptyScriptPubKey
+    case _ if P2PKWithTimeoutScriptPubKey.isP2PKWithTimeoutScriptPubKey(asm) =>
+      P2PKWithTimeoutScriptPubKey.fromAsm(asm)
     case _
         if MultiSignatureWithTimeoutScriptPubKey
           .isMultiSignatureWithTimeoutScriptPubKey(asm) =>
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -866,6 +866,70 @@ object NonStandardNotIfConditionalScriptPubKey
   }
 }
 
+sealed trait P2PKWithTimeoutScriptPubKey extends RawScriptPubKey {
+
+  val pubKey: ECPublicKey =
```
Christewart: can you make these all `lazy`, i worry about weird initialization errors if they aren't `lazy`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -866,6 +866,70 @@ object NonStandardNotIfConditionalScriptPubKey
...
+    buildScript(
+      asm = asm.toVector,
+      constructor = P2PKWithTimeoutScriptPubKeyImpl.apply,
+      invariant = isP2PKWithTimeoutScriptPubKey,
+      errorMsg = s"Given asm was not a P2PKWithTimeoutScriptPubKey, got $asm"
+    )
+  }
+
+  def apply(
+      pubKey: ECPublicKey,
```
Christewart, line 904: nit: Could you name this like `ifPubKey` or something like that, or something a bit more descriptive.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -866,6 +866,70 @@ object NonStandardNotIfConditionalScriptPubKey
...
+      lockTime: ScriptNumber,
+      timeoutPubKey: ECPublicKey): P2PKWithTimeoutScriptPubKey = {
+    val timeoutAsm = CLTVScriptPubKey(lockTime, EmptyScriptPubKey).asm.toVector
```
Christewart, line 907: why `EmptyScriptPubKey` here? Can't we construct the nested `P2PK`/`P2PKH` ? The more I think about this, shouldn't `timeoutPubkey` really just be a generic `ScriptPubKey`?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -866,6 +866,70 @@ object NonStandardNotIfConditionalScriptPubKey
...
+    val pubKeyAsm = BitcoinScriptUtil
+      .calculatePushOp(pubKey.bytes)
+      .toVector ++ Vector(ScriptConstant(pubKey.bytes))
+    val timeoutPubKeyAsm = BitcoinScriptUtil
+      .calculatePushOp(timeoutPubKey.bytes)
+      .toVector ++ Vector(ScriptConstant(timeoutPubKey.bytes))
+
+    P2PKWithTimeoutScriptPubKeyImpl {
```
Christewart: This is different, `{}` rather than `()`? I didn't know this is valid scala syntax for constructors.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -365,6 +365,29 @@ object P2PKScriptSignature extends ScriptFactory[P2PKScriptSignature] {
...
+    buildScript(
+      asm.toVector,
+      ConditionalScriptSignature.fromAsm,
+      isP2PKWithTimeoutScriptSignature,
+      s"The given asm tokens were not a P2PKWithTimeoutScriptSignature, got $asm"
+    )
+  }
+
+  def apply(
+      beforeTimeout: Boolean,
```
Christewart, line 380: nit `isBeforeTimeout`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
Christewart: I think i understand what you are doing now, basically forcing both branches to be single public keys. Is there a reason both branches are forced to be public keys?

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -72,14 +73,22 @@ sealed abstract class UTXOSpendingInfo {
   * then you would construct a ConditionalSpendingInfo using nonNestedTrue as your
   * ConditionalPath. Otherwise if you wanted to use P2PK2 you would use nonNestedFalse.
   */
-sealed trait ConditionalPath
+sealed trait ConditionalPath {
+  def head: Option[Boolean]
```
Christewart: Let's follow the collections api and call this `.headOption`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -72,14 +73,22 @@ sealed abstract class UTXOSpendingInfo {
...
+}
 
 object ConditionalPath {
-  case object NoConditionsLeft extends ConditionalPath
+  case object NoConditionsLeft extends ConditionalPath {
+    override def head: Option[Boolean] = None
```
Christewart: `val`

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -72,14 +73,22 @@ sealed abstract class UTXOSpendingInfo {
...
+  }
   case class ConditionTrue(nextCondition: ConditionalPath)
-      extends ConditionalPath
+      extends ConditionalPath {
+    override def head: Option[Boolean] = Some(true)
```
Christewart: `val`

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -72,14 +73,22 @@ sealed abstract class UTXOSpendingInfo {
...
+  }
   case class ConditionFalse(nextCondition: ConditionalPath)
-      extends ConditionalPath
+      extends ConditionalPath {
+    override def head: Option[Boolean] = Some(false)
```
Christewart: `val`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -29,12 +29,15 @@ sealed abstract class CreditingTxGen {
...
-    Gen.oneOf(p2pkOutput,
-              p2pkhOutput,
-              multiSigOutput, /*cltvOutput,*/ csvOutput,
-              multiSignatureWithTimeoutOutput,
-              conditionalOutput,
-              p2wpkhOutput)
+    Gen.oneOf(
+      p2pkOutput,
+      p2pkhOutput,
+      /*p2pkWithTimeoutOutput,*/
```
Christewart, line 35: this has to be commented out as well since it uses the `nLockTime` field? Ugh, that is frustrating.

Christewart (commented):
Can you add a few hard coded unit tests as well? Also adding the `asm` you listed out in this PR description as a scaladoc on the script would be a good idea to IMO.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: I'm not trying to do something general of some kind of
```
OP_IF SPK_YIELDING_PUBKEY
OP_ELSE SPK_YIELDING_PUBKEY
OP_ENDIF
OP_CHECKSIG
```
I want a Pay to PubKey with a timeout pubkey option.
Should this pattern every show up again I'd likely be partial to adding something more general but I think it would be bad to try to do so with only one data point.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: Blocks `{}` are allowed anywhere, I don't believe there is any real difference between `{}` and `()` other than formatting

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
nkohen: yeah :/ I have some ideas for fixing this, will write up an issue

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
nkohen: I think this could make sense if I was doing something more general (see my response to the comment below), but the semantics here are that this is a P2PK script with a timelocked timeout public key.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
Christewart: I would prefer to just stick the normal syntax unless there is good reason to chagne

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -86,22 +86,47 @@ sealed abstract class CreditingTxGen {
...
+      .map { output =>
+        Vector(output)
+      }
+    val nonCltvGen = Gen.choose(min, 5).flatMap(n => Gen.listOfN(n, output))
+
+    val cltvSize = cltvOutputGens.length
+    val nonCltvSize = nonCltvOutputGens.length
+    val totalSize = cltvSize + nonCltvSize
+
+    cltvGen.flatMap { cltv =>
```
Christewart: Can we just simplify this using [`Gen.frequency`](https://www.scalacheck.org/files/scalacheck_2.11-1.9-api/index.html#org.scalacheck.Gen$@frequency[T](gs:(Int,org.scalacheck.Gen[T])*):org.scalacheck.Gen[T]) in scalacheck ? This gives us a weighted distribution built into the library itself.

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -86,22 +86,47 @@ sealed abstract class CreditingTxGen {
...
-      /*p2pkWithTimeoutOutput,*/
-      multiSigOutput,
-      p2shOutput,
-      csvOutput, /*cltvOutput,*/
-      multiSignatureWithTimeoutOutput,
-      conditionalOutput,
-      p2wpkhOutput,
-      p2wshOutput
-    )
+    Gen.oneOf(nonCltvOutputGens.head,
```
Christewart: This is confusing to me, why are we calling `.head`, `.tail.head`, and then adding the rest of the elements? 

Comments are helpful here :-) 

If it is a weird scala collections issue this code seems to work:

```scala
  def output: Gen[BitcoinUTXOSpendingInfo] = {
    import scala.collection.JavaConverters._
    Gen.sequence(nonCltvOutputGens).flatMap(arrayList => Gen.oneOf(arrayList.asScala))
  }
```

testkit/src/main/scala/org/bitcoins/testkit/core/gen/NumberGenerator.scala
```diff
@@ -87,6 +87,9 @@ trait NumberGenerator {
   def scriptNumbers: Gen[ScriptNumber] =
     Gen.choose(Int64.min.toLong, Int64.max.toLong).map(ScriptNumber(_))
 
+  def timeLockScriptNumbers: Gen[ScriptNumber] =
+    Gen.choose(1L, UInt32.max.toLong - 1L).map(ScriptNumber(_))
```
Christewart, line 92: This seems artificially limited, IIRC `ScriptNumber` can be 5 bytes? Is this the consensus valid range that `ScriptNumbers` can be in for time lock operations? If so let's just add a quick scaladoc to the top of the method to indicate that.

PR #970, opened 2019-12-18
Move scalafmtCheck in CI to beginning

benthecarman:
I think it makes sense to have this run first so we can know right away if we need to make formatting changes, rather than have to wait for all the tests to pass then have to go back and re run then because we had to make a formatting change


benthecarman:
Don't think it is possible / it is not worth more effort than I have put into this

PR #971, opened 2019-12-18
PSBTInputMap to UTXOSigningInfo

benthecarman:
Pre-req for signing and `TxBuilder`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -118,6 +118,7 @@ case class PSBT(
       index: Int,
       signers: Seq[Sign],
       conditionalPath: ConditionalPath): UTXOSpendingInfo = {
+    require(index >= 0 && index < inputMaps.size, s"Index must be within 0 and the number of inputs, got: $index")
     inputMaps(index).toUTXOSpendingInfo(transaction.inputs(index),
                                         signers,
                                         conditionalPath)
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -1,22 +1,34 @@
 package org.bitcoins.core.psbt
 
-import org.bitcoins.core.crypto.{ECDigitalSignature, ECPublicKey, ExtPublicKey}
+import org.bitcoins.core.crypto.{
+  ECDigitalSignature,
+  ECPublicKey,
+  ExtPublicKey,
+  Sign
+}
 import org.bitcoins.core.hd.BIP32Path
 import org.bitcoins.core.number.UInt32
-import org.bitcoins.core.protocol.script.{
-  EmptyScriptSignature,
-  ScriptPubKey,
-  ScriptSignature,
-  ScriptWitness
-}
+import org.bitcoins.core.protocol.script._
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.protocol.{CompactSizeUInt, NetworkElement}
 import org.bitcoins.core.psbt.GlobalPSBTRecord._
+import org.bitcoins.core.psbt.InputPSBTRecord.{
+  NonWitnessOrUnknownUTXO,
+  RedeemScript,
+  SigHashType,
+  WitnessScript,
+  WitnessUTXO
+}
 import org.bitcoins.core.psbt.PSBTGlobalKeyId._
 import org.bitcoins.core.psbt.PSBTInputKeyId._
 import org.bitcoins.core.script.crypto.HashType
 import org.bitcoins.core.serializers.script.RawScriptWitnessParser
 import org.bitcoins.core.util.Factory
+import org.bitcoins.core.wallet.utxo.{
+  BitcoinUTXOSpendingInfo,
+  ConditionalPath,
+  UTXOSpendingInfo
+}
 import scodec.bits._
 
 import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -605,6 +645,77 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+
+    val redeemScriptVec = getRecords[RedeemScript](RedeemScriptKeyId)
+    val redeemScriptOpt =
+      if (redeemScriptVec.size == 1) Some(redeemScriptVec.head.redeemScript)
+      else None
+
+    val scriptWitnessVec = getRecords[WitnessScript](WitnessScriptKeyId)
+    val scriptWitnessOpt =
+      if (scriptWitnessVec.size == 1)
+        Some(P2WSHWitnessV0(scriptWitnessVec.head.witnessScript))
```
benthecarman, line 724: Will this cause issues if the input is a P2WPKH?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -58,6 +70,34 @@ case class PSBT(
...
+
+  /**
+    * Takes the InputPSBTMap at the given index and returns a UTXOSpendingInfo
+    * that can be used to sign the input
+    * @param index index of the InputPSBTMap
+    * @param signers Signers that will be used to sign the input
+    * @return A corresponding UTXOSpendingInfo
+    */
+  def getUTXOSpendingInfo(index: Int, signers: Seq[Sign]): UTXOSpendingInfo = {
+    getUTXOSpendingInfo(index, signers)
```
nkohen: infinite loop?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -605,6 +645,77 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+      val tx = txVec.head.transactionSpent
+      tx.outputs(txIn.previousOutput.vout.toInt)
+    } else {
+      throw new UnsupportedOperationException(
+        "Not enough information in the InputPSBTMap to get a valid UTXOSpendingInfo")
+    }
+
+    val redeemScriptVec = getRecords[RedeemScript](RedeemScriptKeyId)
+    val redeemScriptOpt =
+      if (redeemScriptVec.size == 1) Some(redeemScriptVec.head.redeemScript)
```
nkohen, line 718: nit: I think you could use `redeemScriptVec.headOption.map(_.redeemScript)` instead

nkohen (commented):
Also, tests would be nice

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Doesn't work :( the return type is `Any`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -82,6 +94,35 @@ case class PSBT(
...
   }
+
+  /**
+    * Takes the InputPSBTMap at the given index and returns a UTXOSpendingInfo
+    * that can be used to sign the input
+    * @param index index of the InputPSBTMap
+    * @param signers Signers that will be used to sign the input
+    * @return A corresponding UTXOSpendingInfo
+    */
+  def getUTXOSpendingInfo(index: Int, signers: Seq[Sign]): UTXOSpendingInfo = {
```
nkohen: nit: I think in this case I'd prefer having just one method with a default value specified for `conditionalPath` rather than overloading

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -652,6 +693,73 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+  /**
+    * Takes the InputPSBTMap returns a UTXOSpendingInfo
+    * that can be used to sign the input
+    * @param txIn The transaction input that this InputPSBTMap represents
+    * @param signers Signers that will be used to sign the input
+    * @return A corresponding UTXOSpendingInfo
+    */
+  def toUTXOSpendingInfo(
+      txIn: TransactionInput,
+      signers: Seq[Sign]): UTXOSpendingInfo = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Yeah, I think this will be an issue. Not really sure what goes in `WitnessScript` in the `P2WPKH` case though, since the witness is not a `ScriptPubKey` but instead is just a PubKey

nkohen: Also, looking at `WitnessScript` now, that `ScriptPubKey` parameter should probably be a `RawScriptPubKey`, doesn't have to happen on this PR though

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -304,7 +335,7 @@ object InputPSBTRecord {
     override val value: ByteVector = redeemScript.asmBytes
   }
 
-  case class WitnessScript(witnessScript: ScriptPubKey)
+  case class WitnessScript(witnessScript: RawScriptPubKey)
```
nkohen, line 338: So I looked into this more and I think this type should actually be `ScriptWitnessV0` since `RawScriptPubKey` is not sufficient for `P2WPKH` inputs

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Hopefully my comment above will help resolve this

PR #985, opened 2019-12-26
Automated wallet recovery

rorp:


app/cli/src/main/scala/org/bitcoins/cli/Cli.scala
```diff
@@ -46,7 +37,7 @@ object CliCommand {
   case object GetBestBlockHash extends CliCommand
   case object GetBlockCount extends CliCommand
   case class Rescan(
-      addresses: Vector[BitcoinAddress],
+      addressBatchSize: Option[Int],
```
Christewart, line 40: So we are assuming that they have a populated `AddressTable` and just rescans the contents of this right?

core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
```diff
@@ -30,40 +29,11 @@ trait ChainQueryApi {
 
   def getFiltersBetweenHeights(
       startHeight: Int,
-      endHeight: Int): Future[Vector[(GolombFilter, DoubleSha256DigestBE)]]
+      endHeight: Int): Future[Vector[(GolombFilter, DoubleSha256DigestBE, Int)]]
```
Christewart: can we just make this return type a `case class` rather than a `Tuple` now? I'm guessing the `Int` is height?

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
```diff
@@ -148,7 +148,7 @@ class NeutrinoNodeTest extends NodeUnitTest {
       }
   }
 
-  it must "download a block that matches a compact block filter" taggedAs (UsesExperimentalBitcoind) in {
+  it must "download a block that matches a compact block filter" taggedAs (UsesExperimentalBitcoind) ignore {
```
Christewart: why the change to `ignore` here?

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -320,15 +245,16 @@ trait UnlockedWalletApi extends LockedWalletApi {
...
-                    } else {
-                      blocks
-                    }
-                }
+                filterGroup
+                  .foldLeft(Vector.empty[(DoubleSha256DigestBE, Int)]) {
+                    (blocks, filter) =>
+                      val matcher = new SimpleFilterMatcher(filter._1)
+                      if (matcher.matchesAny(bytes)) {
+                        blocks :+ (filter._2 -> filter._3)
```
Christewart: Adding a specific type for `(GolombFilter, DoubleSha256DigestBE, Int)` would be very helpful in reading this code :+1:

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+import org.bitcoins.wallet.models.AddressDb
+import org.bitcoins.wallet.{LockedWallet, WalletLogger}
+
+import scala.annotation.tailrec
+import scala.concurrent.{ExecutionContext, Future}
+
+private[wallet] trait RescanHandling extends WalletLogger {
+  self: LockedWallet =>
+
+  val AddressGapLimit = 20
```
Christewart: So if I remember our conversation, the difference between `AddressGapLimit` and `addressBatchSize` is: 

`addressBatchSize` - the number of addresses we should generate from a keychain to attempt to match in in a rescan
`AddressGapLimit` - the number of addresses required to go _without a match_ before we determine that our wallet is "synced". For instance, if `addressBatchSize=100`, and `AddressGapLimit=20` we do a rescan and the last address we find containing funds is at index 75, we would _not_ generate more addresses to try and rescan. However if the last index containing funds was `81`, we would generate _another_ `100` addresses from the keychain and attempt to rescan those.

If that is correct, I think that would be a very helpful scaladoc to add. Also maybe we should make `AddressGapLimit` configurable in the `bitcoin-s.conf`.

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+  /////////////////////
+  // Public facing API
+
+  /** @inheritdoc */
+  override def rescanNeutrinoWallet(
+      startOpt: Option[BlockStamp],
+      endOpt: Option[BlockStamp],
+      addressBatchSize: Int): Future[Unit] = {
+
+    logger.info(s"Starting rescanning the wallet.")
```
Christewart, line 26: Maybe add `startOpt` and `endOpt` in the log message?

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    val res = for {
+      _ <- spendingInfoDAO.deleteAll()
+      _ <- addressDAO.deleteAll()
```
Christewart, line 30: I'm not sure if I agree with deleting everything? It seems like it is a bad idea to destroy data that already exists in the wallet, is there a reason we can't just update/create?

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+    } yield ()
+
+    res.onComplete(_ => logger.info("Finished rescanning the wallet"))
+
+    res
+  }
+
+  /** @inheritdoc */
+  override def rescanSPVWallet(): Future[Unit] =
+    Future.failed(new RuntimeException("Rescan not implemented"))
```
Christewart: "Rescan not implemented for spv wallet"

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+            addressDb.scriptPubKey)
+          _ <- if (spendingInfoDbs.isEmpty) addressDAO.delete(addressDb)
+          else FutureUtil.unit
+        } yield ()
+      }
+    } yield ()
+  }
+
+  private def calcAddressGap(chainType: HDChainType): Future[Int] = {
+    for {
```
Christewart, line 78: Maybe refactor to this? 

```scala
  private def calcAddressGap(chainType: HDChainType): Future[Int] = {
    val addressDbsF = addressDAO.findAll()
    //make sure all addressDb are of the correct chainType
    //and they are sorted according to their index so we can
    //calculate the gap accurately
    val sortedF = addressDbsF.map {
      _.filter(_.path.chain.chainType == chainType)
        .sortBy(_.path.address.index)
    }
    
    for {
      addressDbs <- sortedF
      addressGap <- addressDbs
        .foldLeft(Future.successful(0)) { (prevNF, addressDb) =>
          for {
            prevN <- prevNF
            spendingInfoDbs <- spendingInfoDAO.findByScriptPubKey(
              addressDb.scriptPubKey)
          } yield {
            if (spendingInfoDbs.isEmpty) prevN + 1 else 0
          }
        }
    } yield {
      logger.debug(s"Address gap: $addressGap")
      addressGap
    }
  }
```

Christewart: Also there might be a potential optmization here if we pass in the `AddressGapLimit` where we can return early... but maybe that is pre-mature. For large wallet's it might make a difference.

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+    } yield {
+      logger.debug(s"Address gap: $addressGap")
+      addressGap
+    }
+  }
+
+  private def downloadAndProcessBlocks(
+      blocks: Vector[DoubleSha256Digest]): Future[Unit] = {
+    logger.debug(s"Requesting ${blocks.size} block(s)")
+    blocks.foldLeft(Future.successful(())) { (prevF, blockHash) =>
```
Christewart: nit: `Future.successful(())` -> `FutureUtil.unit`

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -0,0 +1,157 @@
...
+      blocks.sortBy(_._2).map(_._1.flip)
+    }
+
+    blocksF.onComplete(_ => threadPool.shutdown())
+    blocksF
+  }
+
+  private def generateScriptPubKeys(
+      count: Int): Future[Vector[ScriptPubKey]] = {
+    for {
```
Christewart, line 134: maybe this? This should allow us to run things in parallel. Is this going to cause synchronization issues with `getNewAddress()` you were talking about the other day?

```scala

    val receiveSPKsF: Future[Vector[BitcoinAddress]] = {
      1.to(count)
        .foldLeft(Future.successful(Vector.empty[BitcoinAddress])) {
          (prevFuture, _) =>
            for {
              prev <- prevFuture
              address <- getNewAddress()
            } yield prev :+ address
        }
    }

    val changeAddressesF: Future[Vector[BitcoinAddress]] = {
      1.to(count)
        .foldLeft(Future.successful(Vector.empty[BitcoinAddress])) {
          (prevFuture, _) =>
            for {
              prev <- prevFuture
              address <- getNewChangeAddress()
            } yield prev :+ address
        }
    }

    for {
      receive <- receiveSPKsF
      change <- changeAddressesF
    } yield receive.map(_.scriptPubKey) ++ change.map(_.scriptPubKey)
```

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -89,6 +89,27 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
   /////////////////////
   // Private methods
 
+  private var blockProcessingSignals =
+    Map.empty[DoubleSha256Digest, Promise[DoubleSha256Digest]]
+
+  private[wallet] def subscribeForBlockProcessingCompletionSignal(
+      blockHash: DoubleSha256Digest): Future[DoubleSha256Digest] =
+    synchronized {
+      val signal = Promise[DoubleSha256Digest]()
```
Christewart: so this overwrites existing signals right? I think this is what we want? 

```scala
  private[wallet] def subscribeForBlockProcessingCompletionSignal(
      blockHash: DoubleSha256Digest): Future[DoubleSha256Digest] =
    synchronized {
      blockProcessingSignals.get(blockHash) match {
        case Some(existingP) => existingP.future
        case None =>
          val signal = Promise[DoubleSha256Digest]()
          blockProcessingSignals = blockProcessingSignals.updated(blockHash, signal)
          signal.future
      }
    }
```

wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
```diff
@@ -25,6 +24,10 @@ case class AddressDAO()(
   override def createAll(ts: Vector[AddressDb]): Future[Vector[AddressDb]] =
     SlickUtil.createAllNoAutoInc(ts, database, table)
 
+  def deleteAll(): Future[Int] = {
```
Christewart: maybe add this in `CRUD.scala`?

app/cli/src/main/scala/org/bitcoins/cli/Cli.scala
rorp: Right.

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
rorp: There's no acute need in running this in parallel

Christewart:
There seems to be some imports that need to be cleaned up: 


>[W]  [E2] wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala:18:37
[W]       Unused import
[W]       L18: import org.bitcoins.wallet.{Wallet, WalletLogger}
[W]            ^^^^^^^
[W]  [E1] wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala:15:33
[W]       Unused import
[W]       L15: import org.bitcoins.keymanager.{KeyManager, KeyManagerParams}
[W]            ^^^^^^^
[W] wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala: L18 [E2]
[W] wallet/src/main/scala/org/bitcoins/wallet/models/AddressTable.scala: L114 [E3], L115 [E4], L116 [E5]
[W] wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala: L15 [E1]
[W] wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala: L130 [E6], L132 [E7]
Compiled wallet (2599ms)
Compiling appServer (3 Scala sources)
[W]  [E2] app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala:28:9
[W]       local method parseAddresses in method fromJsArr is never used
[W]       L28:     def parseAddresses(value: Value): Vector[BitcoinAddress] = value match {
[W]                    ^
[W]  [E1] app/server/src/main/scala/org/bitcoins/server/Main.scala:136:36
[W]       discarded non-Unit value
[W]       L136:         wallet.processCompactFilter(blockHash, blockFilter)
[W]                                                ^
[W] app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala: L28 [E2]
[W] app/server/src/main/scala/org/bitcoins/server/Main.scala: L136 [E1]
Compiled appServer (513ms)
Compiling testkit (6 Scala sources)
[W]  [E1] testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala:13:32
[W]       Unused import
[W]       L13: import org.bitcoins.keymanager.KeyManager
[W]            ^^^^^^^
[W] testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala: L13 [E1]
Compiled testkit (1205ms)
Compiling testkit (1 Scala source)
Compiled testkit (230ms)
Compiling keyManagerTest-test (4 Scala sources)
Compiling nodeTest-test (5 Scala sources)
Compiling appServerTest-test (2 Scala sources)
Compiling chainTest-test (1 Scala source)
Compiling walletTest-test (14 Scala sources)
Compiling coreTest-test (12 Scala sources)
Compiled chainTest-test (1198ms)
[W]  [E-1] there were two deprecation warnings (since 2.13.0); re-run with -deprecation for details
Compiled keyManagerTest-test (1363ms)
[W]  [E-1] there were two deprecation warnings (since 3.1.0); re-run with -deprecation for details
Compiled appServerTest-test (1623ms)
Compiled nodeTest-test (1998ms)
[W]  [E-1] there were 8 deprecation warnings (since 3.1.0); re-run with -deprecation for details
Compiled coreTest-test (2824ms)
Compiled walletTest-test (3403ms)

key-manager/src/main/scala/org/bitcoins/keymanager/bip39/BIP39KeyManager.scala
```diff
@@ -56,7 +56,7 @@ case class BIP39KeyManager(
 object BIP39KeyManager
     extends KeyManagerCreateApi[BIP39KeyManager]
     with BitcoinSLogger {
-  val badPassphrase = AesPassword.fromString("bad-password").get
+  val badPassphrase = AesPassword.fromString("changeMe").get
```
Christewart, line 59: why did you change this?

rorp: I did.

Issue #972, opened 2019-12-19
Implement proper password handling in the key-manager


Christewart:
Currently we just have hard coded passwords which is obviously not ideal. We need to break interfaces and add optional password parameters: 

https://github.com/bitcoin-s/bitcoin-s/blob/1ba8205b7cda8c94a8a0bdd1140da7ba9a382413/key-manager/src/main/scala/org/bitcoins/keymanager/KeyManager.scala#L66

Christewart:
This also encompasses the [BIP39 password problem](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed). Basically there are two types of passwords that need to be considered: The BIP39 password and the password that is used to encrypt the entire mnemonic on disk. Perhaps these should be the same password?

Christewart:
So from talking with @nkohen and @rorp it seems that there are a couple things that need to be considered here: 

### Raw BIP32 HD Key

What I mean by this is just storing the raw [BIP32 private key](https://github.com/bitcoin-s/bitcoin-s/blob/80b5c4c25bf73808dad758d9fadd2c64f345c020/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala#L109) on the file system some where. 

There should definitely be an option to encrypt this on disk as this is the _only_ information needed to re-generate the entire wallet's private keys. The current scheme we use is encrypting the information with [AES](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/crypto/AesCrypt.scala) and then storing it to our file system in a file called [`encrypted-bitcoin-s-seed.json`](https://github.com/bitcoin-s/bitcoin-s/blob/master/key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala#L15)

This can then be decrypted with a user password (not BIP39 related!) so they can use their private keys.

### BIP39
[With BIP39, a password is optional](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed): 

>To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the derived key is 512 bits (= 64 bytes).

[...]

PR #966, opened 2019-12-18
Move initialization of wallet entropy into the key manager

Christewart:
This moves initialization of the mneomnic seed in bitcoin-s into the key-manager project rather than having it done inside of the wallet project. 

Mostly this renames things to use the name `KeyManager` rather than `{Unlock,Lock}Wallet`

This does not encapsulate all key handling logic into the key manager yet, this will be done on a separate PR. 

##### TODO
- [x] documentation
- [x] ci coverage minimums for key-manager
- [x] testing

I filed #972 to handle passwords in a future PR.

PR #988, opened 2019-12-27
This creates a subtype BIP39KeyManager and moves all existing KeyMana…

Christewart:
…ger functionality to that subtype. We leave a empty 'KeyManager' trait for now

Basically this takes everything named `KeyManager` from #966 and renames it to `BIP39KeyManager` and creates a new `bip39` package inside of `keymanager` to contain all BIP39 key related functionality. 

I realized we should really start to segregate the different key generation/storage types from #972 as with different types of key storage there will be no passwords. Putting a password in the parent `KeyManager` type really doesn't make sense if there is no password needed for the specific type of key manager.


key-manager/src/main/scala/org/bitcoins/keymanager/KeyManager.scala
```diff
@@ -1,132 +1,3 @@
...
-    val mnemonicCodeE =
-      WalletStorage.decryptMnemonicFromDisk(kmParams.seedPath, password)
-
-    mnemonicCodeE match {
-      case Right(mnemonic) => Right(new KeyManager(mnemonic, kmParams))
-      case Left(v)         => Left(v)
-    }
-  }
-}
+trait KeyManager
```
nkohen, line 3: nit: scaladoc

key-manager/src/main/scala/org/bitcoins/keymanager/KeyManagerCreateApi.scala
```diff
@@ -16,22 +16,21 @@ import scodec.bits.BitVector
   *                           can write it down. They should also be prompted
   *                           to confirm at least parts of the code.
   */
-trait KeyManagerCreateApi {
+trait KeyManagerCreateApi[T <: KeyManager] {
```
nkohen, line 19: nit: `KM` at the very least instead of `T` if not something even more descriptive

nkohen (approved):
Just some nits

rorp (approved):

PR #989, opened 2019-12-27
DLC fee simplification

nkohen:
Replaced fee computation with constant values from empirical fee computation.

Also added closing tx fee to pre-added fees since this output must be swept immediately (as per discussions with CG)


nkohen:
This should also make it easier to have deterministic fees for inter-op with other clients

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -187,46 +186,38 @@ case class BinaryOutcomeDLCWithSelf(
                                Vector(fundingLocalPubKey, fundingRemotePubKey))
   }
 
+  private val estimatedCETFee: CurrencyUnit = Satoshis(190 * feeRate.toLong)
```
Christewart: So i'm guessing `190` and `122` are estimated transaction sizes? Can you use constants for these with an appropriate scaladoc?

Christewart (approved):

PR #990, opened 2019-12-28
Implement abililty to use BIP39 password. This means this password ne…

Christewart:
…eds to be password through our various projects to be able to correctly generate the key that controls the wallet. This also renames 'CreateKeyManagerApi' -> 'BIP39CreateKeymanagerApi' as the bip39 password is needed when creating the KeyManager

This partially addresses #972 . However there are still problems where we are requiring two passwords in certain places -- one to encrypt the mnemonic on disk and one to properly generate the root `xpriv` key. This needs to be thought out more, and it won't be done on this PR.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -101,7 +101,7 @@ object Main extends App {
       val locked = LockedWallet(nodeApi, chainQueryApi)
 
       // TODO change me when we implement proper password handling
-      locked.unlock(BIP39KeyManager.badPassphrase,bip39PasswordOpt) match {
+      locked.unlock(BIP39KeyManager.badPassphrase, bip39PasswordOpt) match {
         case Right(wallet) =>
           Future.successful(wallet)
         case Left(kmError) =>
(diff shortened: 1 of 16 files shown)


benthecarman (approved):

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -38,13 +38,13 @@ object Main extends App {
   val peerSocket =
     parseInetSocketAddress(nodeConf.peers.head, nodeConf.network.port)
   val peer = Peer.fromSocket(peerSocket)
-
+  val bip39PasswordOpt = None //todo need to prompt user for this
```
nkohen, line 41: ping, or maybe this needs an issue (or comment on existing #972) if you plan on doing this on a future PR

testkit/src/main/scala/org/bitcoins/keymanager/KeyManagerTestUtil.scala
```diff
@@ -15,5 +19,24 @@ object KeyManagerTestUtil {
...
+  def bip39PasswordOpt: Option[String] = {
+    if (scala.util.Random.nextBoolean()) {
+      Some(bip39Password)
+    } else {
+      None
+    }
+  }
+
+  def bip39Password: String = {
+    CryptoGenerators.bip39Password.sample.get
```
nkohen, line 38: you should use the `GenUtil` sample here in case `suchThat` is ever introduced or anything like that

nkohen (approved):

rorp (approved):

PR #993, opened 2019-12-30
P2SH Signing

benthecarman:
This is needed for PSBT signing

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
@@ -53,6 +53,20 @@ sealed abstract class BaseTxSigComponent extends TxSigComponent {
   override def sigVersion = SigVersionBase
 }
 
+sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
+  override def scriptPubKey: P2SHScriptPubKey =
+    output.scriptPubKey.asInstanceOf[P2SHScriptPubKey]
+
+  override def scriptSignature: P2SHScriptSignature = {
+    val s = transaction.inputs(inputIndex.toInt).scriptSignature
+    require(s.isInstanceOf[P2SHScriptSignature],
+            "Must have P2SHScriptSignature for P2SH, got: " + s)
+    s.asInstanceOf[P2SHScriptSignature]
+  }
+
+  override def amount: CurrencyUnit = output.value
+}
+
 /**
   * The [[org.bitcoins.core.crypto.TxSigComponent TxSigComponent]]
   * used to represent all the components necessarily for
@@ -164,6 +178,17 @@ object BaseTxSigComponent {
     BaseTxSigComponentImpl(transaction, inputIndex, output, flags)
   }
 
+}
+
+object P2SHTxSigComponent {
+
+  def apply(
+      transaction: Transaction,
+      inputIndex: UInt32,
+      output: TransactionOutput,
+      flags: Seq[ScriptFlag]): BaseTxSigComponent = {
+    BaseTxSigComponent(transaction, inputIndex, output, flags)
+  }
 }
 
 object WitnessTxSigComponent {
--- a/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
@@ -54,17 +54,21 @@ sealed abstract class BaseTxSigComponent extends TxSigComponent {
 }
 
 sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
-  override def scriptPubKey: P2SHScriptPubKey =
-    output.scriptPubKey.asInstanceOf[P2SHScriptPubKey]
 
-  override def scriptSignature: P2SHScriptSignature = {
-    val s = transaction.inputs(inputIndex.toInt).scriptSignature
-    require(s.isInstanceOf[P2SHScriptSignature],
-            "Must have P2SHScriptSignature for P2SH, got: " + s)
-    s.asInstanceOf[P2SHScriptSignature]
-  }
+  require(
+    input.scriptSignature
+      .isInstanceOf[P2SHScriptSignature],
+    "Must have P2SHScriptSignature for P2SH, got: " + input.scriptSignature
+  )
+  require(
+    output.scriptPubKey
+      .isInstanceOf[P2SHScriptSignature],
+    "Must have P2SHScriptPubKey for P2SH, got: " + output.scriptPubKey
+  )
 
-  override def amount: CurrencyUnit = output.value
+  override def scriptPubKey: P2SHScriptPubKey = scriptPubKey
+
+  override def scriptSignature: P2SHScriptSignature = scriptSignature
 }
 
 /**
@@ -182,12 +186,19 @@ object BaseTxSigComponent {
 
 object P2SHTxSigComponent {
 
+  private case class P2SHTxSigComponentImpl(
+      transaction: Transaction,
+      inputIndex: UInt32,
+      output: TransactionOutput,
+      flags: Seq[ScriptFlag])
+      extends P2SHTxSigComponent
+
   def apply(
       transaction: Transaction,
       inputIndex: UInt32,
       output: TransactionOutput,
-      flags: Seq[ScriptFlag]): BaseTxSigComponent = {
-    BaseTxSigComponent(transaction, inputIndex, output, flags)
+      flags: Seq[ScriptFlag]): P2SHTxSigComponent = {
+    P2SHTxSigComponentImpl(transaction, inputIndex, output, flags)
   }
 }
 
--- a/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
@@ -62,7 +62,7 @@ sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
   )
   require(
     output.scriptPubKey
-      .isInstanceOf[P2SHScriptSignature],
+      .isInstanceOf[P2SHScriptPubKey],
     "Must have P2SHScriptPubKey for P2SH, got: " + output.scriptPubKey
   )
 
--- a/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
@@ -66,9 +66,11 @@ sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
     "Must have P2SHScriptPubKey for P2SH, got: " + output.scriptPubKey
   )
 
-  override def scriptPubKey: P2SHScriptPubKey = scriptPubKey
+  override def scriptPubKey: P2SHScriptPubKey =
+    output.scriptPubKey.asInstanceOf[P2SHScriptPubKey]
 
-  override def scriptSignature: P2SHScriptSignature = scriptSignature
+  override def scriptSignature: P2SHScriptSignature =
+    input.scriptSignature.asInstanceOf[P2SHScriptSignature]
 }
 
 /**
--- a/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
@@ -50,7 +50,7 @@ sealed abstract class TxSigComponent {
   * P2SH(witness script) [[org.bitcoins.core.protocol.script.ScriptPubKey ScriptPubKey]]
   */
 sealed abstract class BaseTxSigComponent extends TxSigComponent {
-  override def sigVersion = SigVersionBase
+  override def sigVersion: SignatureVersion = SigVersionBase
 }
 
 sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
@@ -84,13 +84,13 @@ sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
   */
 sealed trait WitnessTxSigComponent extends TxSigComponent {
 
-override def transaction: WitnessTransaction
+  override def transaction: WitnessTransaction
 
   def witness: ScriptWitness = transaction.witness.witnesses(inputIndex.toInt)
 
   def witnessVersion: WitnessVersion
 
-  override def sigVersion = SigVersionWitnessV0
+  override def sigVersion: SignatureVersion = SigVersionWitnessV0
 }
 
 /** This represents checking the [[org.bitcoins.core.protocol.transaction.WitnessTransaction WitnessTransaction]]
@@ -107,7 +107,9 @@ sealed abstract class WitnessTxSigComponentRaw extends WitnessTxSigComponent {
 
 /** This represents checking the [[org.bitcoins.core.protocol.transaction.WitnessTransaction WitnessTransaction]]
   * against a P2SH(P2WSH) or P2SH(P2WPKH) scriptPubKey */
-sealed abstract class WitnessTxSigComponentP2SH extends P2SHTxSigComponent with WitnessTxSigComponent {
+sealed abstract class WitnessTxSigComponentP2SH
+    extends P2SHTxSigComponent
+    with WitnessTxSigComponent {
   override def scriptPubKey: P2SHScriptPubKey =
     output.scriptPubKey.asInstanceOf[P2SHScriptPubKey]
 
@@ -138,7 +140,7 @@ sealed abstract class WitnessTxSigComponentP2SH extends P2SHTxSigComponent with 
     case Failure(err) => throw err
   }
 
-  override def amount = output.value
+  override def amount: CurrencyUnit = output.value
 
 }
 
@@ -161,9 +163,9 @@ sealed abstract class WitnessTxSigComponentRebuilt extends TxSigComponent {
     * rebuild the scriptPubKey above */
   def witnessScriptPubKey: WitnessScriptPubKey
 
-  override def sigVersion = SigVersionWitnessV0
+  override def sigVersion: SignatureVersion = SigVersionWitnessV0
 
-  def witnessVersion = witnessScriptPubKey.witnessVersion
+  def witnessVersion: WitnessVersion = witnessScriptPubKey.witnessVersion
 
 }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -53,6 +53,20 @@ sealed abstract class BaseTxSigComponent extends TxSigComponent {
   override def sigVersion = SigVersionBase
 }
 
+sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
```
Christewart, line 56: how does is this handled relative to `WitnessTxSigComponentP2SH` ? It seems that they are essentially the same wrapper around a `redeemScript`. Any way we can share code here?

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -166,6 +180,17 @@ object BaseTxSigComponent {
 
 }
 
+object P2SHTxSigComponent {
+
+  def apply(
+      transaction: Transaction,
+      inputIndex: UInt32,
+      output: TransactionOutput,
+      flags: Seq[ScriptFlag]): BaseTxSigComponent = {
```
Christewart: This should return a `: P2SHTxSigComponent`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -166,6 +180,17 @@ object BaseTxSigComponent {
 
 }
 
+object P2SHTxSigComponent {
+
+  def apply(
+      transaction: Transaction,
```
Christewart, line 201: This should probably be typed as a `: BaseTransaction`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -166,6 +180,17 @@ object BaseTxSigComponent {
...
+      inputIndex: UInt32,
+      output: TransactionOutput,
+      flags: Seq[ScriptFlag]): BaseTxSigComponent = {
+    BaseTxSigComponent(transaction, inputIndex, output, flags)
```
Christewart: You should probably define a `private case class P2SHTxSigComponentImpl(...)` and use it as a constructor here.

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -53,6 +53,20 @@ sealed abstract class BaseTxSigComponent extends TxSigComponent {
...
 
+sealed abstract class P2SHTxSigComponent extends BaseTxSigComponent {
+  override def scriptPubKey: P2SHScriptPubKey =
+    output.scriptPubKey.asInstanceOf[P2SHScriptPubKey]
+
+  override def scriptSignature: P2SHScriptSignature = {
+    val s = transaction.inputs(inputIndex.toInt).scriptSignature
+    require(s.isInstanceOf[P2SHScriptSignature],
+            "Must have P2SHScriptSignature for P2SH, got: " + s)
+    s.asInstanceOf[P2SHScriptSignature]
```
nkohen: This require should be top-level rather than inside this `def`, and also there should be a `require` for the `output.scriptPubKey`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
nkohen: I'm actually hoping to slowly get rid of that pattern as it obfuscates the invariants imo as its only purpose is to restrict construction where I think we should just make things correct by construction

nkohen: That said something has to be concrete and i see no reason not to make `P2SHTxSigComponent` concrete

nkohen (commented):
Some non-structural review. More to come on actual structure

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
benthecarman: There is some code re-use but not sure if it is possible to share code here.  `WitnessTxSigComponentP2SH` cannot extend `P2SHTxSigComponent` because it needs to extend WitnessTxSigComponent`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
benthecarman: This will require us to have to do something like `P2SHTxSigComponent(unsignedTx.asInstanceOf[BaseTransaction],...` for most/every call in `Signer.scala`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
benthecarman: I think it makes more sense to do Chris's suggestion here because that is what is done throughout `TxSigCompenent.scala` and in a later refactor we can change it

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -53,6 +53,24 @@ sealed abstract class BaseTxSigComponent extends TxSigComponent {
...
+      .isInstanceOf[P2SHScriptSignature],
+    "Must have P2SHScriptSignature for P2SH, got: " + input.scriptSignature
+  )
+  require(
+    output.scriptPubKey
+      .isInstanceOf[P2SHScriptSignature],
+    "Must have P2SHScriptPubKey for P2SH, got: " + output.scriptPubKey
+  )
+
+  override def scriptPubKey: P2SHScriptPubKey = scriptPubKey
```
Christewart: These will result in infinite loops if you call them

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -361,14 +346,13 @@ sealed abstract class MultiSigSigner
     val signers = signersWithPubKeys.map(_.signFunction)
 
     val requiredSigs = spendingInfoToSatisfy.scriptPubKey.requiredSigs
+    val component =
+      if (spendingInfo.isInstanceOf[P2SHSpendingInfo])
+        sigComponent(spendingInfoToSatisfy, unsignedTx)
```
nkohen: I'm worried here about the case of `P2SH(Conditional(MultiSig, ...))` and things like it. Won't it be wrong to use `spendingInfoToSatisfy` as you're supposed to sign `Conditional(MultiSig, ...)` not `MultiSig`?

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
nkohen: Is it not possible for this to be a `WitnessTransaction` though? (If there are other outputs or inputs that use segwit)

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -166,6 +186,24 @@ object BaseTxSigComponent {
...
 }
 
+object P2SHTxSigComponent {
+
+  private case class P2SHTxSigComponentImpl(
+      transaction: Transaction,
+      inputIndex: UInt32,
+      output: TransactionOutput,
+      flags: Seq[ScriptFlag])
+      extends P2SHTxSigComponent
```
nkohen, line 198: I think I'd prefer to have the top level class be a `case class` rather than using the `impl` pattern as it makes things nicer to pattern match on

nkohen: In the past this pattern has been used in place of `require` statements to restrict what constructions are legal but you have invariants stated in the super class so using the `impl` pattern is unnecessary

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +369,40 @@ sealed abstract class MultiSigSigner
...
+      isDummySignature: Boolean,
+      spendingInfoToSatisfy: P2SHSpendingInfo)(
+      implicit ec: ExecutionContext): Future[TxSigComponent] = {
+    if (spendingInfoToSatisfy != spendingInfo) {
+      Future.fromTry(TxBuilderError.WrongSigner)
+    } else {
+      val (_, output, inputIndex, _) = relevantInfo(spendingInfo, unsignedTx)
+
+      val signedSigComponentF = BitcoinSigner.sign(
+        spendingInfo,
```
nkohen: I think that as a remedy to the concern I had above, you should not add special cases elsewhere for `P2SH` and instead pass in `spendingInfoToSatisfy.nestedSpendingInfo` for this param here

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +369,40 @@ sealed abstract class MultiSigSigner
 
 object MultiSigSigner extends MultiSigSigner
 
+/** Used to sign a [[org.bitcoins.core.protocol.script.P2SHScriptPubKey]] */
+sealed abstract class P2SHSigner extends BitcoinSigner[P2SHSpendingInfo] {
+  override def sign(
```
nkohen, line 373: I don't think that this method currently handles `P2SH(P2WPKH)` and `P2SH(P2WSH)` correctly (see `TxBuilder.signAndAddInput` for more details)

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,70 @@ sealed abstract class MultiSigSigner
...
+    } else {
+
+      val (_, output, inputIndex, _) =
+        relevantInfo(spendingInfo, unsignedTx)
+
+      if (spendingInfoToSatisfy
+            .isInstanceOf[P2SHNestedSegwitV0UTXOSpendingInfo] && spendingInfoToSatisfy.redeemScript
+            .isInstanceOf[P2WPKHWitnessSPKV0]) {
+        val p2wpkh =
+          spendingInfoToSatisfy.redeemScript.asInstanceOf[P2WPKHWitnessSPKV0]
```
nkohen: Use `match` rather than `isInstanceOf` followed by `asInstanceOf`, you can `match` a tuple like `(spendingInfo, spendingInfoToSatisfy.redeemScript)` to get what you want here

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,70 @@ sealed abstract class MultiSigSigner
...
+        val pubKey = spendingInfo.signers.head.publicKey
+        if (P2WPKHWitnessSPKV0(pubKey) != p2wpkh) {
+          Future.fromTry(TxBuilderError.WrongPublicKey)
```
nkohen: I beleive you don't need to do this check as it is covered in `P2WPKHV0SpendingInfo` as an invariant

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,70 @@ sealed abstract class MultiSigSigner
...
+          val uwtx = {
+            val u = unsignedTx.updateInput(inputIndex.toInt, updatedInput)
+            WitnessTransaction.toWitnessTx(u)
+          }
+
+          Future.successful(
+            WitnessTxSigComponentP2SH(transaction = uwtx,
+                                      inputIndex = inputIndex,
+                                      output = output,
+                                      flags = Policy.standardFlags))
```
nkohen: This is incorrect/incomplete: you never generate a signature and update the `uwtx` TransactionWitness with that signature. See `TxBuilder.signP2SHP2WPKH` for what needs to happen here.

Also, I really recommend you replace `TxBuilder.signAndAddInput's P2SH logic with what you are developing here. This way you will have tests failing if the nested P2SH segwit signing is not working.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,70 @@ sealed abstract class MultiSigSigner
...
+        val scriptSigF = signedSigComponentF.map { signedSigComponent =>
+          P2SHScriptSignature(signedSigComponent.scriptSignature,
+                              spendingInfoToSatisfy.redeemScript)
+        }
+
+        updateScriptSigInSigComponent(unsignedTx,
+                                      inputIndex.toInt,
+                                      output,
+                                      scriptSigF)
+      }
```
nkohen, line 430: I'm still not sure this covers this case either, take a look at `TxBuilder.signAndAddInputForP2SH` and ditto the comment above about replacing stuff in `TxBuilder` for test coverage

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+            TransactionOutput(spendingInfo.amount, redeemScript)
+
+          val updatedUTXOInfo = BitcoinUTXOSpendingInfo(
+            outpoint,
+            updatedOutput,
+            spendingInfo.signers,
+            None,
+            spendingInfo.scriptWitnessOpt,
+            spendingInfo.hashType,
+            spendingInfo.conditionalPath)
```
nkohen: nit: I think that p2sh spending info has a field called `nestedSpendingInfo` that should do this for you IIRC

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+          val scriptSigF = signedSigComponentF.map { signedSigComponent =>
+            P2SHScriptSignature(signedSigComponent.scriptSignature,
+                                spendingInfoToSatisfy.redeemScript)
+          }
+
+          updateScriptSigInSigComponent(unsignedTx,
+                                        inputIndex.toInt,
+                                        output,
+                                        scriptSigF)
+      }
```
nkohen: Is there a reason this case differs from the one above it? Looking at the old implementation in `TxBuilder` above it looks like they used to do the same thing

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+                                signedInputs,
+                                btx.outputs,
+                                btx.lockTime)
+              case wtx: WitnessTransaction =>
+                WitnessTransaction(wtx.version,
+                                   signedInputs,
+                                   wtx.outputs,
+                                   wtx.lockTime,
+                                   wtx.witness)
+            }
```
nkohen: If you are spending a `P2SH(P2WSH)` output isn't it true that this tx must be a `WitnessTransaction`? Maybe I'm misunderstanding something here

nkohen (commented):
My only other remaining concern (that isn't mentioned in my review) is that we now have multiple representations of `TxSigComponent`s for `P2SH(Segwit)`: `P2SHTxSigComponent` and `WitnessTxSigComponentP2SH`. What should happen there IMO is that `P2SHTxSigComponent` should become an ADT with two cases (nested segwit and not) and just have the existing `WitnessTxSigComponentP2SH` extend `P2SHTxSigComponent` (as well as `WitnessTxSigComponent`) and restrict the non-nested subtype to only allow non-nested stuff.

LMK if you have any questions about this.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
benthecarman: That makes sense, this was copied from the old `signAndAddInputForP2SH` but it looks like it can be removed

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
Christewart: You are right :+1:

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+          val hashForSig = TransactionSignatureSerializer.hashForSignature(
+            txSigComponent = wtxSigComp,
+            hashType = spendingInfo.hashType)
+
+          //sign the hash
+          val signature: ECDigitalSignature = {
+            if (isDummySignature) {
+              EmptyDigitalSignature
+            } else {
+              val sig = spendingInfo.signers.head.sign(hashForSig.bytes)
```
Christewart: Hmm... this seems wrong. the `doSign()` method takes care of this everywhere else. Can't we just call that here?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+            TransactionInput(outpoint, EmptyScriptSignature, oldInput.sequence)
+
+          val updatedTx =
+            unsignedTx.updateInput(inputIndex.toInt, input)
+
+          val updatedOutput =
+            TransactionOutput(spendingInfo.amount, redeemScript)
+
+          val updatedUTXOInfo = BitcoinUTXOSpendingInfo(
+            outpoint,
```
Christewart: please use `alt+enter` on intellij for named arguments. Also don't we know that this _has_ to be a `P2SHNoNestSpendingInfo` ? Why not use that type here?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -385,6 +368,141 @@ sealed abstract class MultiSigSigner
...
+              case wtx: WitnessTransaction =>
+                WitnessTransaction(wtx.version,
+                                   signedInputs,
+                                   wtx.outputs,
+                                   wtx.lockTime,
+                                   wtx.witness)
+            }
+            P2SHTxSigComponent(finalTx, inputIndex, output, flags)
+          }
+        case (_, _) =>
```
Christewart: In general it is a bad idea to use `_` in a pattern match. 

For one, the first `(_,` can never be anything, and the second `_` can be typed with a `_: RawScriptPubKey` if I understand this. You probably should just pattern match on `spendingInfoToSatisfy` and then inside of that match on `spendingInfoToSatisfy.redeemScript`

nkohen (approved):
Assuming CI is good, LGTM

Christewart (approved):

PR #994, opened 2019-12-30
ToLocal Closing Tx Amount

nkohen:
As agreed upon with CG, the ToLocal closing tx should output the "payout" amount. It turns out that our approximation code is so consistent with how fees are calculated internally that we were already doing this ... to the satoshi! (I can tell that this is true because I didn't have to touch the test vectors and the test passed, and I inspected the test vectors and we have nice round output values). But nonetheless I hard-coded the payouts in ToLocal closing txs to ensure that we are never off in the future.

### DIFF
--- a/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
@@ -462,19 +462,43 @@ case class BinaryOutcomeDLCWithSelf(
   def constructClosingTx(
       privKey: ECPrivateKey,
       spendingInfo: BitcoinUTXOSpendingInfo,
-      isLocal: Boolean): Future[Transaction] = {
-    // Construct Closing Transaction
-    val txBuilder = BitcoinTxBuilder(
-      destinations = Vector(
-        TransactionOutput(spendingInfo.output.value,
-                          P2WPKHWitnessSPKV0(privKey.publicKey))),
-      utxos = Vector(spendingInfo),
-      feeRate = feeRate,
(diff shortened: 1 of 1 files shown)


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
```diff
@@ -462,19 +462,43 @@ case class BinaryOutcomeDLCWithSelf(
...
-        TransactionOutput(spendingInfo.output.value,
-                          P2WPKHWitnessSPKV0(privKey.publicKey))),
-      utxos = Vector(spendingInfo),
-      feeRate = feeRate,
-      changeSPK = emptyChangeSPK,
-      network = network
-    )
+      isLocal: Boolean,
+      isWin: Boolean,
+      isToLocal: Boolean): Future[Transaction] = {
```
Christewart, line 467: we now have a `isLocal` and `isToLocal`? :thinking:

Christewart (commented):
Frankly I don't understand the difference between the locals, as when this stuff gets made into real wallet software you don't have have any of the remote transactions. Everything should just be symmetric. I'm just going to assume this is good for now

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCWithSelf.scala
nkohen: For posterity, `isLocal` is true when the method is called from the "local" view of this DLCWithSelf, while `isToLocal` refers to whether or not this spends the `ToLocalOutput` or not.

Of course, this will be much easier to follow once I separate into two views so that only the local view exists.

PR #997, opened 2020-01-02
PSBT Constructors and Extraction

nkohen:
Added PSBT constructors that take in an unsigned transaction and `UTXOSpendingInfo`s.

Also introduced [transaction extraction](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#transaction-extractor).

TODO:
- [x] Test that `TxBuilder.sign` agrees with PSBT.fromUnsignedTxAndInputs(...).extractTransaction


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -413,6 +476,32 @@ object PSBT extends Factory[PSBT] {
...
+
+    PSBT(globalMap, inputMaps, outputMaps)
+  }
+
+  def fromUnsignedTxAndInputs(
+      unsignedTx: Transaction,
+      spendingInfos: Vector[UTXOSpendingInfo])(
+      implicit ec: ExecutionContext): Future[PSBT] = {
+    require(spendingInfos.length == unsignedTx.inputs.length,
+            "Must have a UTXOSpendingInfo for every input")
```
benthecarman, line 559: Could validate that the `spendingInfos` correspond to inputs in the `unsignedTx` as well

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -907,6 +996,15 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
     InputPSBTMap((this.elements ++ other.elements).distinct)
   }
 
+  def toUTXOSpendingInfo(
```
benthecarman, line 1070: I don't understand the point of this function, why would you want a spending info after it is signed?

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -44,6 +44,10 @@ object Sign {
     SignImpl(signFunction, pubKey)
   }
 
+  def withSig(sig: ECDigitalSignature, pubKey: ECPublicKey): Sign = {
```
Christewart: maybe call this `constant`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1000,6 +1098,28 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+        case _: BaseTransaction => InputPSBTMap(Vector(scriptSig))
+        case wtx: WitnessTransaction =>
+          val witness = wtx.witness.witnesses(sigComponent.inputIndex.toInt)
+          val scriptWitness = FinalizedScriptWitness(witness)
+          InputPSBTMap(Vector(scriptSig, scriptWitness))
+      }
+    }
+  }
+
+  def empty: InputPSBTMap = InputPSBTMap(Vector.empty)
```
Christewart: val

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1039,6 +1159,8 @@ case class OutputPSBTMap(elements: Vector[OutputPSBTRecord]) extends PSBTMap {
 
 object OutputPSBTMap {
 
+  def empty: OutputPSBTMap = OutputPSBTMap(Vector.empty)
```
Christewart: val

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: This is for the case where someone is doing something interactive and given some finalized `InputMap`, wants to do something else (non-interactive) whose interface expects `UTXOSpendingInfo`s instead of `PSBT` things (since `UTXOSpendingInfo` is the canonical signing information thing).

I guess one could argue that they could just use `TransactionInput` somehow and also that this case doesn't really exist yet (I originally thought I would use this in DLC stuff but now it seems I will be using PSBTs internally), but since the work is done and it isn't too complicated I figured I'd leave it here and we can get rid of it if it becomes an annoyance.

@Christewart thoughts?

benthecarman (approved):

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -350,6 +351,20 @@ sealed abstract class CreditingTxGen {
...
+      .flatMap { creditingTxsInfo =>
+        val creditingOutputs = creditingTxsInfo.map(c => c.output)
+        val creditingOutputsAmt = creditingOutputs.map(_.value)
+        val totalAmount = creditingOutputsAmt.fold(CurrencyUnits.zero)(_ + _)
+
+        TransactionGenerators.smallOutputs(totalAmount).map { destinations =>
+          (creditingTxsInfo, destinations)
+        }
+      }
+      .suchThat(_._1.nonEmpty)
```
Christewart, line 408: nit it's nice to avoid the use of `.suchThat()` in simple cases like this where you can make another generator pretty easily.

Christewart (approved):

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
nkohen: So actually in this case it should always be the case that the list is non-empty since it uses a number between 1 and 5 ... but then if things fail and it tries to shrink for some reason it occasionally uses an empty list and gives that error message instead, and this seems to fix that.

PR #1001, opened 2020-01-03
Add more expressive txo states for the wallet

Christewart:
This adds more states for a txo in our wallet. Previously we just had a flag that was called `spent`. As you can guess, `spent` just indicated if the utxo has been spent or not. There are many more states that a wallet txo can be in: 

1. DoesNotExist
2. UnconfirmedReceived
3. ConfirmedReceived
4. UnconfirmedSpent
5. ConfirmedSpent

There are still parts of the wallet that need clarification on when a txo can transition from `Unconfirmed` -> `Confirmed`, but that will have to be done in a separate PR.

NOTE: 

This breaks the existing wallet schema, specifically `SpendingInfoTable`.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
@@ -322,6 +322,11 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
       (nWord > UInt32(0xffff) && nSize > 32)
     )
   }
+
+  /** Generates a random positive integer */
+  def posInt: Int = {
+    Math.abs(scala.util.Random.nextInt())
+  }
 }
 
 object NumberUtil extends NumberUtil
--- a/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
@@ -325,7 +325,7 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
 
   /** Generates a random positive integer */
   def posInt: Int = {
-    Math.abs(scala.util.Random.nextInt())
+    scala.util.Random.between(0, Int.MaxValue)
   }
 }
 
--- a/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
@@ -325,7 +325,7 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
 
   /** Generates a random positive integer */
   def posInt: Int = {
-    scala.util.Random.between(0, Int.MaxValue)
+    Math.abs(scala.util.Random.nextInt())
   }
 }
 
--- a/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
@@ -13,21 +13,21 @@ object TxoState {
   final case object DoesNotExist extends TxoState
 
   /** Means we have received funds to this utxo, but they are not confirmed */
-  final case object UnconfirmedReceived extends ReceivedState
+  final case object PendingConfirmationsReceived extends ReceivedState
 
   /** Means we have received funds and they are fully confirmed for this utxo */
   final case object ConfirmedReceived extends ReceivedState
 
   /** Means we have spent this utxo, but it is not fully confirmed */
-  final case object UnconfirmedSpent extends SpentState
+  final case object PendingConfirmationsSpent extends SpentState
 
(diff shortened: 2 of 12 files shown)


wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -99,9 +101,13 @@ case class SpendingInfoDAO()(
...
 
-  /** Enumerates all unspent TX outputs in the wallet */
+  private val receivedStates: Vector[TxoState] =
+    Vector(TxoState.UnconfirmedReceived, TxoState.ConfirmedReceived)
+
+  /** Enumerates all unspent TX outputs in the wallet with the state
+    * [[TxoState.UnconfirmedReceived]] or [[TxoState.ConfirmedReceived]] */
   def findAllUnspent(): Future[Vector[SpendingInfoDb]] = {
-    val query = table.filter(!_.spent)
+    val query = table.filter(_.state.inSet(receivedStates))
```
benthecarman, line 110: It might be useful to have one that gets only the confirmed txs if someone doesn't want to try and spend unconfirmed txs

core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /** Means we have received funds and they are fully confirmed for this utxo */
+  final case object ConfirmedReceived extends ReceivedState
+
+  /** Means we have spent this utxo, but it is not fully confirmed */
+  final case object UnconfirmedSpent extends SpentState
+
+  /** Means we have spent this utxo, and it is fully confirmed */
+  final case object ConfirmedSpent extends SpentState
+
```
benthecarman, line 26: An immature state might be useful as well for txs that have been mined but not able to be spent yet

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -322,6 +322,11 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
       (nWord > UInt32(0xffff) && nSize > 32)
     )
   }
+
+  /** Generates a random positive integer */
+  def posInt: Int = {
+    Math.abs(scala.util.Random.nextInt())
```
rorp, line 328: Probably `scala.util.Random.between(0, Int.MaxValue)` can get more uniformly distibuted values.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -37,9 +38,16 @@ abstract class LockedWallet
...
-              txo.output.value
+          .filter(predicate)
+          .map {
+            case txo: SpendingInfoDb =>
+              txo.state match {
+                case TxoState.UnconfirmedReceived | TxoState.ConfirmedReceived =>
+                  txo.output.value
+                case TxoState.UnconfirmedSpent | TxoState.ConfirmedSpent |
+                     TxoState.DoesNotExist =>
+                  CurrencyUnits.zero
```
rorp, line 50: Why `sum`  zeros?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -99,9 +101,13 @@ case class SpendingInfoDAO()(
     database.runVec(filtered.result)
   }
 
-  /** Enumerates all unspent TX outputs in the wallet */
+  private val receivedStates: Vector[TxoState] =
```
rorp: Should it be `Set[TxoState]`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
Christewart: Hmm... I guess that is what I meant by "Unconfirmed" but i see how this can be confusing if you take it literally. There are two "Unconfirmeds" -- unconfirmed in the network and unconfirmed in your wallet. The latter setting is dependent on how many confirmations you think your transaction needs before it is fully "settled".

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
Christewart: I think this is useful, but I'll let this go into a separate PR. I want to try and keep this one as small as possible.

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
Christewart: It's a `.map` so we need to return some value. We want to ignore the values of outputs that are `Spent` or `DoesNotExist`.

PR #1002, opened 2020-01-03
PSBT Finalizer

nkohen:
Added [finalizing](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer) functionality to PSBTs


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -96,6 +97,19 @@ case class PSBT(
     PSBT(global, inputs, outputs)
   }
 
+  def finalizePSBT: Option[PSBT] = {
```
nkohen, line 111: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
     getRecords(FinalizedScriptSigKeyId).nonEmpty || getRecords(
       FinalizedScriptWitnessKeyId).nonEmpty
 
+  def finalize(input: TransactionInput): Option[InputPSBTMap] = {
```
nkohen, line 1082: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+            .Try(outputs(input.previousOutput.vout.toInt).scriptPubKey)
+            .toOption
+      }
+    }
+
+    scriptPubKeyToSatisfyOpt.flatMap(finalize)
+  }
+
+  private def finalize(spkToSatisfy: ScriptPubKey): Option[InputPSBTMap] = {
+    def wipeAndAdd(
```
nkohen, line 1112: scaladocs

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+      case _: P2WPKHWitnessSPKV0 =>
+        val sigOpt =
+          getRecords[PartialSignature](PartialSignatureKeyId).headOption
+        sigOpt.map { sig =>
+          val witness = P2WPKHWitnessV0(sig.pubKey, sig.signature)
+          val scriptSig = EmptyScriptSignature
+          wipeAndAdd(scriptSig, Some(witness))
+        }
+      case conditional: ConditionalScriptPubKey =>
+        // Parse into BinaryTree[Vector[KeyHash]], traverse leaves
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+        val builder =
+          Vector.newBuilder[(
+              ConditionalPath,
+              Vector[Sha256Hash160Digest],
+              RawScriptPubKey)]
+
+        def buildSPK(
```
nkohen: not needed, get rid of all the cltv stuff

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+          } else {
+            CLTVScriptPubKey(cltvNesting.head,
+                             buildSPK(rawSPK, cltvNesting.tail))
+          }
+        }
+
+        def addLeaves(
+            rawSPK: RawScriptPubKey,
+            path: Vector[Boolean],
+            cltvNesting: Vector[ScriptNumber]): Unit = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+                           hashes,
+                           buildSPK(multiSig, cltvNesting)))
+            case EmptyScriptPubKey | _: NonStandardScriptPubKey |
+                _: WitnessCommitment =>
+              throw new UnsupportedOperationException(
+                s"$rawSPK is not yet supported")
+          }
+        }
+
+        val sigs = getRecords[PartialSignature](PartialSignatureKeyId)
```
nkohen, line 1256: comment

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -897,6 +911,219 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+      case locktime: LockTimeScriptPubKey =>
+        val nestedFinalizedOpt = finalize(locktime.nestedScriptPubKey)
+
+        nestedFinalizedOpt.map { nestedFinalized =>
+          val nestedScriptSig = nestedFinalized
+            .getRecords[FinalizedScriptSig](FinalizedScriptSigKeyId)
+            .head
+          val otherElements = nestedFinalized.elements.filterNot(
+            _.isInstanceOf[FinalizedScriptSig])
+          val scriptSig = locktime match {
```
nkohen: match not needed

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -896,6 +913,193 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
     getRecords(FinalizedScriptSigKeyId).nonEmpty || getRecords(
       FinalizedScriptWitnessKeyId).nonEmpty
 
+  /** Finalizes this input if possible, returning None if not */
+  def finalize(input: TransactionInput): Option[InputPSBTMap] = {
+    val scriptPubKeyToSatisfyOpt: Option[ScriptPubKey] = {
```
benthecarman: Should check if the input is already finalized at the beginning

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -96,6 +96,23 @@ case class PSBT(
     PSBT(global, inputs, outputs)
   }
 
+  /** Finalizes this PSBT if possible
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer]]
+    * @return None if there exists any input which is not ready for signing
+    */
+  def finalizePSBT: Option[PSBT] = {
+    val finalizedInputOpts = inputMaps.zip(transaction.inputs).map {
```
benthecarman: Should check if the psbt is already finalized at the beginning

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -482,6 +483,28 @@ object ConditionalScriptSignature
...
   }
 
+  @scala.annotation.tailrec
+  def apply(
+      nestedScriptSig: ScriptSignature,
+      conditionalPath: ConditionalPath): ConditionalScriptSignature = {
+    conditionalPath match {
+      case ConditionalPath.NoConditionsLeft =>
+        throw new IllegalArgumentException("ConditionalPath cannot be empty")
+      case ConditionalPath.nonNestedTrue =>
```
Christewart, line 516: This is really unfortunate that `.nonNestedFalse` and `nonNestedTrue` aren't real ADTs :-(, especially since they are the terminating cases's for how `ConditionalPath` is defined.

benthecarman (changes requested):
Meant to not approve

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -896,6 +913,193 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+    val scriptPubKeyToSatisfyOpt: Option[ScriptPubKey] = {
+      val witnessUTXOOpt = getRecords[WitnessUTXO](WitnessUTXOKeyId).headOption
+      val nonWitnessUTXOOpt =
+        getRecords[NonWitnessOrUnknownUTXO](NonWitnessUTXOKeyId).headOption
+
+      (witnessUTXOOpt, nonWitnessUTXOOpt) match {
+        case (None, None) => None
+        case (Some(witnessUtxo), _) =>
+          Some(witnessUtxo.spentWitnessTransaction.scriptPubKey)
+        case (_, Some(utxo)) =>
```
Christewart: This is missing the `case (Some(_), Some(_))`. It is concerning that there isn't a compiler warning here!

Christewart: Oh, it gets caught by the `_` in the above statements, is this the exact ordering you want? Seems a little dangerous.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -896,6 +913,193 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+        }
+      case conditional: ConditionalScriptPubKey =>
+        val builder =
+          Vector.newBuilder[(
+              ConditionalPath,
+              Vector[Sha256Hash160Digest],
+              RawScriptPubKey)]
+
+        /** Traverses the ConditionalScriptPubKey tree for leaves and adds them to builder */
+        def addLeaves(rawSPK: RawScriptPubKey, path: Vector[Boolean]): Unit = {
```
Christewart, line 1219: can this be marked as `tailrec`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -83,6 +83,18 @@ object ConditionalPath {
 
   val nonNestedTrue: ConditionalPath = ConditionTrue(NoConditionsLeft)
   val nonNestedFalse: ConditionalPath = ConditionFalse(NoConditionsLeft)
+
+  def fromBranch(branch: Vector[Boolean]): ConditionalPath = {
```
Christewart, line 96: tailrec with accumulator?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: It should be fine, if you have a `WitnessUTXO` you can remove the `NonWitnessOrUnknownUTXO`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
nkohen: I would argue that the only terminating case for `ConditionalPath` is `NoConditionsLeft` (as it should be) and that these two vals are just short-hand for exactly what I'm going for in this match case (which is to extract further than just terminal cases)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: what ben said ^ the spec says to remove `NonWitnessOrUnknownUTXO` if a `WitnessUTXO` record is present

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: sadly no, since there are two calls made to `addLeaves` in the `ConditionalScriptPubKey` case. The good news is that people don't do giant nesting of `ConditionalScriptPubKey`s and I doubt it is even possible to do so to such an extent that this recursion gets out of hand

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: sadly no again because the last call in this method is to `ConditionalPath` constructors. Essentially the below is equivalent to:
```
val condition = fromBranch(branch.tail)
ConditionTrue(condition)
```

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
Christewart: This layout breaks pattern matching exhaustiveness though. Why not just make them `final case object` if they are important?

nkohen: It doesn't break exhaustiveness: I removed the case beneath this and got
```
[warn] /home/nkohen/Desktop/SuredBits/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala:486:5: match may not be exhaustive.
[warn] It would fail on the following input: ConditionTrue(_)
[warn]     conditionalPath match {
[warn]     ^
[warn] one warning found
```

Essentially when representing branches of a binary tree a single step/leg is going to be more frequent than other things (especially in this context) but it is not trivial or fundamental, only leaves are trivial.

benthecarman (approved):

Christewart (approved):

PR #1006, opened 2020-01-03
Eclair Web Socket client

rorp:


Christewart (approved):
Can we get a basic unit test for the websocket?

PR #1012, opened 2020-01-07
Payment secret and features LN invoice tags

rorp:


benthecarman (approved):

Christewart (approved):

Christewart:
Actually looking a this closer: It looks like you didn't add generators for the new tags in `LnInvoiceGen`, can you add the generators and then incorporate them into the `optionalTags`/`allTags` generator as well?

PR #1013, opened 2020-01-07
Eclair performance tests

rorp:


Christewart:
Built on top of #1012

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -821,7 +823,7 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
   // We spawn fresh clients in this test because the test
   // needs nodes with activity both related and not related
   // to them
-  it should "get all channel updates for a given node ID" in {
+  it should "get all channel updates for a given node ID" ignore {
```
Christewart: Why did these get `ignore`d?

PR #1017, opened 2020-01-08
Increase core test coverage

benthecarman:
On my tests it increased from 90.54 to 91.01

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/config/NetworkParametersTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/config/NetworkParametersTest.scala
@@ -21,4 +21,12 @@ class NetworkParametersTest extends BitcoinSUnitTest {
   it must "create the correct magic network bytes for regtest" in {
     BitcoinSUtil.encodeHex(RegTest.magicBytes) must be("fabfb5da")
   }
+
+  it must "get the correct Network from string" in {
+    assert(Networks.fromString("mainnet").contains(MainNet))
+    assert(Networks.fromString("testnet").contains(TestNet3))
+    assert(Networks.fromString("regtest").contains(RegTest))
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -190,7 +190,7 @@ object Bech32Address extends AddressFactory[Bech32Address] {
       np: NetworkParameters): Try[Bech32Address] =
     spk match {
       case witSPK: WitnessScriptPubKey =>
-        Bech32Address.fromScriptPubKey(witSPK, np)
+        Success(Bech32Address(witSPK, np))
```
benthecarman, line 193: This function was an infinite loop so I fixed it

Christewart: yikes!

core-test/src/test/scala/org/bitcoins/core/config/NetworkParametersTest.scala
```diff
@@ -21,4 +21,12 @@ class NetworkParametersTest extends BitcoinSUnitTest {
   it must "create the correct magic network bytes for regtest" in {
     BitcoinSUtil.encodeHex(RegTest.magicBytes) must be("fabfb5da")
   }
+
+  it must "get the correct Network from string" in {
+    assert(Networks.fromString("mainnet").contains(MainNet))
+    assert(Networks.fromString("testnet").contains(TestNet3))
+    assert(Networks.fromString("regtest").contains(RegTest))
+    assert(Networks.fromString("").isEmpty)
+    assert(Networks.fromString("craig wright is a fraud").isEmpty)
```
Christewart, line 30: :joy:

Christewart (approved):

PR #1020, opened 2020-01-10
Single Signing

nkohen:
Implements `UTXOSpendingInfoSingle` and `SingleSigner` which together allows for users to using single `Sign`ers to get `ECDigitalSignature`s on transactions, without requiring all `Sign`ers be present.

Adds a property-based test to ensure that `SingleSigner` and `Signer` don't diverge

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
@@ -1,19 +1,25 @@
 package org.bitcoins.core.wallet.signer
 
+import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
 import org.bitcoins.core.protocol.script.WitnessScriptPubKey
+import org.bitcoins.core.wallet.builder.BitcoinTxBuilder
+import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
 import org.bitcoins.core.wallet.utxo.{
   P2WPKHV0SpendingInfo,
   P2WSHV0SpendingInfoFull,
+  UTXOSpendingInfoSingle,
   UnassignedSegwitNativeUTXOSpendingInfo
 }
 import org.bitcoins.testkit.core.gen.{
+  ChainParamsGenerator,
   CreditingTxGen,
   GenUtil,
+  ScriptGenerators,
   TransactionGenerators
 }
 import org.bitcoins.testkit.util.BitcoinSAsyncTest
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 
 class SignerTest extends BitcoinSAsyncTest {
 
@@ -63,6 +69,71 @@ class SignerTest extends BitcoinSAsyncTest {
     val tx = GenUtil.sample(TransactionGenerators.baseTransaction)
     recoverToSucceededIf[IllegalArgumentException] {
       P2WSHSigner.sign(dumbSpendingInfo, tx, isDummySignature = false, p2wsh)
+    }
+  }
+
+  private val outputGen = CreditingTxGen.outputs
+    .flatMap { creditingTxsInfo =>
+      val creditingOutputs = creditingTxsInfo.map(c => c.output)
+      val creditingOutputsAmt = creditingOutputs.map(_.value)
+      val totalAmount = creditingOutputsAmt.fold(CurrencyUnits.zero)(_ + _)
+
+      TransactionGenerators.smallOutputs(totalAmount).map { destinations =>
+        (creditingTxsInfo, destinations)
+      }
+    }
+    .suchThat(_._1.nonEmpty)
+
+  it must "sign a mix of spks in a tx and then verify that single signing agrees" in {
+    forAllAsync(outputGen,
+                ScriptGenerators.scriptPubKey,
+                ChainParamsGenerator.bitcoinNetworkParams) {
+      case ((creditingTxsInfos, destinations), changeSPK, network) =>
+        val fee = SatoshisPerVirtualByte(Satoshis(1000))
+
+        for {
+          builder <- BitcoinTxBuilder(destinations,
+                                      creditingTxsInfos,
+                                      fee,
+                                      changeSPK._1,
+                                      network)
+          unsignedTx <- builder.unsignedTx
+          signedTx <- builder.sign
+
+          singleSigs <- {
+            val singleInfosVec: Vector[Vector[UTXOSpendingInfoSingle]] =
+              creditingTxsInfos.toVector.map(_.toSingles)
+            val sigVecFs = singleInfosVec.map { singleInfos =>
+              val sigFs = singleInfos.map { singleInfo =>
+                val keyAndSigF =
+                  BitcoinSignerSingle.signSingle(singleInfo,
+                                                 unsignedTx,
+                                                 isDummySignature = false)
+
+                keyAndSigF.map(_._2)
+              }
+
+              Future.sequence(sigFs)
+            }
+
+            Future.sequence(sigVecFs)
+          }
+        } yield {
+          signedTx.inputs.foreach { input =>
+            val indexOpt = creditingTxsInfos.zipWithIndex
+              .find(_._1.outPoint == input.previousOutput)
+              .map(_._2)
+            assert(indexOpt.isDefined)
+            val index = indexOpt.get
+            val sigs = singleSigs(index)
+            val expectedSigs = input.scriptSignature.signatures
+
+            assert(sigs.length == expectedSigs.length)
+            assert(sigs.forall(expectedSigs.contains))
+          }
+
+          succeed
+        }
     }
   }
 }
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
@@ -1,7 +1,16 @@
 package org.bitcoins.core.wallet.signer
 
+import org.bitcoins.core.crypto.ECDigitalSignature
 import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
-import org.bitcoins.core.protocol.script.WitnessScriptPubKey
+import org.bitcoins.core.protocol.script.{
+  EmptyScriptWitness,
+  P2WPKHWitnessV0,
+  P2WSHWitnessV0,
+  ScriptSignature,
+  WitnessScriptPubKey
+}
+import org.bitcoins.core.protocol.transaction.WitnessTransaction
+import org.bitcoins.core.serializers.script.ScriptParser
 import org.bitcoins.core.wallet.builder.BitcoinTxBuilder
 import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
 import org.bitcoins.core.wallet.utxo.{
@@ -24,6 +33,9 @@ import scala.concurrent.{ExecutionContext, Future}
 class SignerTest extends BitcoinSAsyncTest {
 
   implicit val ec: ExecutionContext = ExecutionContext.global
+
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
 
   behavior of "Signer"
 
@@ -100,7 +112,7 @@ class SignerTest extends BitcoinSAsyncTest {
           unsignedTx <- builder.unsignedTx
           signedTx <- builder.sign
 
-          singleSigs <- {
+          singleSigs: Vector[Vector[ECDigitalSignature]] <- {
             val singleInfosVec: Vector[Vector[UTXOSpendingInfoSingle]] =
               creditingTxsInfos.toVector.map(_.toSingles)
             val sigVecFs = singleInfosVec.map { singleInfos =>
@@ -119,17 +131,29 @@ class SignerTest extends BitcoinSAsyncTest {
             Future.sequence(sigVecFs)
           }
         } yield {
-          signedTx.inputs.foreach { input =>
-            val indexOpt = creditingTxsInfos.zipWithIndex
-              .find(_._1.outPoint == input.previousOutput)
-              .map(_._2)
-            assert(indexOpt.isDefined)
-            val index = indexOpt.get
-            val sigs = singleSigs(index)
-            val expectedSigs = input.scriptSignature.signatures
+          signedTx.inputs.zipWithIndex.foreach {
+            case (input, inputIndex) =>
+              val infoAndIndexOpt = creditingTxsInfos.zipWithIndex
+                .find(_._1.outPoint == input.previousOutput)
+              assert(infoAndIndexOpt.isDefined)
+              val (info, index) = infoAndIndexOpt.get
+              val sigs = singleSigs(index)
 
-            assert(sigs.length == expectedSigs.length)
-            assert(sigs.forall(expectedSigs.contains))
+              val expectedSigs = if (info.scriptWitnessOpt.isEmpty) {
+                input.scriptSignature.signatures
+              } else {
+                signedTx
+                  .asInstanceOf[WitnessTransaction]
+                  .witness
+                  .witnesses(inputIndex) match {
+                  case p2wpkh: P2WPKHWitnessV0 => Vector(p2wpkh.signature)
+                  case p2wsh: P2WSHWitnessV0   => p2wsh.signatures
+                  case EmptyScriptWitness      => Vector.empty
+                }
+              }
+
+              assert(sigs.length == expectedSigs.length)
+              assert(sigs.forall(expectedSigs.contains))
           }
 
           succeed
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
@@ -3,7 +3,7 @@ package org.bitcoins.core.wallet.signer
 import org.bitcoins.core.protocol.script.WitnessScriptPubKey
 import org.bitcoins.core.wallet.utxo.{
   P2WPKHV0SpendingInfo,
-  P2WSHV0SpendingInfo,
+  P2WSHV0SpendingInfoFull,
   UnassignedSegwitNativeUTXOSpendingInfo
 }
 import org.bitcoins.testkit.core.gen.{
@@ -59,7 +59,7 @@ class SignerTest extends BitcoinSAsyncTest {
     val dumbSpendingInfo = GenUtil.sample(CreditingTxGen.output)
     val p2wsh = GenUtil
       .sample(CreditingTxGen.p2wshOutput)
-      .asInstanceOf[P2WSHV0SpendingInfo]
+      .asInstanceOf[P2WSHV0SpendingInfoFull]
     val tx = GenUtil.sample(TransactionGenerators.baseTransaction)
     recoverToSucceededIf[IllegalArgumentException] {
       P2WSHSigner.sign(dumbSpendingInfo, tx, isDummySignature = false, p2wsh)
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
@@ -6,11 +6,9 @@ import org.bitcoins.core.protocol.script.{
   EmptyScriptWitness,
   P2WPKHWitnessV0,
   P2WSHWitnessV0,
-  ScriptSignature,
   WitnessScriptPubKey
 }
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -63,6 +62,31 @@ sealed abstract class UTXOSpendingInfo {
...
+    * where full signing and single key singing happen.
+    */
+  override def signer: Sign = signers.head
+
+  def requiredSigs: Int
+
+  def toSingle(signerIndex: Int): UTXOSpendingInfoSingle
+
+  def toSingles: Vector[UTXOSpendingInfoSingle] = {
+    signers.indices.take(requiredSigs).toVector.map(toSingle)
```
benthecarman, line 89: If we had a 2-of-3 multisig, this would only allow us to sign with the first 2 keys

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -404,32 +551,62 @@ case class P2PKWithTimeoutSpendingInfo(
...
+    MultiSignatureSpendingInfoSingle(outPoint,
+                                     amount,
+                                     scriptPubKey,
+                                     signers(signerIndex),
+                                     hashType)
+  }
+
+  override def toSingles: Vector[MultiSignatureSpendingInfoSingle] = {
+    signers.take(requiredSigs).map { signer =>
+      MultiSignatureSpendingInfoSingle(outPoint,
```
benthecarman, line 618: If we had a 2-of-3 multisig, this would only allow us to sign with the first 2 keys

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -596,7 +597,11 @@ sealed abstract class P2WPKHSigner
                              unsignedTx,
                              isDummySignature,
                              spendingInfoToSatisfy)
-        sig = sigComponent.scriptSignature.signatures.head
+        sig = sigComponent
+          .asInstanceOf[WitnessTxSigComponent]
+          .witness
+          .asInstanceOf[script.P2WPKHWitnessV0]
+          .signature
```
benthecarman: It'd be safer to use a `match` here and throw errors when you don't get the expected types

core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
```diff
@@ -1,24 +1,42 @@
 package org.bitcoins.core.wallet.signer
 
-import org.bitcoins.core.protocol.script.WitnessScriptPubKey
+import org.bitcoins.core.crypto.ECDigitalSignature
+import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
+import org.bitcoins.core.protocol.script.{
+  EmptyScriptWitness,
+  P2WPKHWitnessV0,
+  P2WSHWitnessV0,
+  ScriptSignature,
```
benthecarman: Unused

core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
```diff
@@ -1,24 +1,42 @@
...
+  WitnessScriptPubKey
+}
+import org.bitcoins.core.protocol.transaction.WitnessTransaction
+import org.bitcoins.core.serializers.script.ScriptParser
```
benthecarman: Unused

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -404,32 +551,62 @@ case class P2PKWithTimeoutSpendingInfo(
...
-    conditionalPath: ConditionalPath)
-    extends RawScriptUTXOSpendingInfo {
+    hashType: HashType
+) extends RawScriptUTXOSpendingInfo
+    with MultiSignatureSpendingInfo {
+  require(signers.length >= scriptPubKey.requiredSigs,
+          s"Not enough signers!: $this")
+
+  override val requiredSigs: Int = scriptPubKey.requiredSigs
+
```
benthecarman, line 606: `signer` should be overrun to throw an error so people cannot mistakenly use it

benthecarman: Also, then all the `UXTOSpendingInfo`s with nested spending infos should have `signer` be `nestedSpendingInfo.signer`

benthecarman (changes requested):
This is a huge improvement on mine, thank you for doing this. Just a couple small things :)

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -94,14 +118,121 @@ object ConditionalPath {
...
+                                  redeemScriptOpt,
+                                  scriptWitnessOpt,
+                                  hashType,
+                                  conditionalPath)
+  }
+}
+
 object BitcoinUTXOSpendingInfo {
 
   def apply(
```
benthecarman, line 267: Could this function be changed so if `signers.size == 1` then it uses `BitcoinUTXOSpendingInfoSingle.apply`

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -454,35 +666,106 @@ case class ConditionalSpendingInfo(
...
   }
+
+  override val requiredSigs: Int = nestedSpendingInfo.requiredSigs
 }
 
-/** This is the case where we are spending a [[org.bitcoins.core.protocol.script.WitnessScriptPubKeyV0 witness v0 script]]  */
-sealed trait SegwitV0NativeUTXOSpendingInfo extends BitcoinUTXOSpendingInfo {
+sealed trait SegwitV0NativeUTXOSpendingInfoSingle
+    extends BitcoinUTXOSpendingInfoSingle {
   def scriptWitness: ScriptWitnessV0
```
benthecarman, line 749: We could add `def scriptPubKey: WitnessScriptPubKeyV0` here

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: So actually it will allow you to sign with any two keys but if you happen to have all three then it will only use the first two in the `Vector` (which could still be any two). Will add a doc tho

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
nkohen: I messed around with this idea for quite a while but in the end it turned out to be quite complicated and I think

1. It isn't really necessary since BitcoinUTXOSpendingInfo extends BitcoinUTXOSpendingInfoSingle and
2. I think that it is good to keep the two applies somewhat separate so as to make sure people get expected behavior and failing if impossible rather than trying correct for people when they make mistakes.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -118,6 +87,68 @@ sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
         BaseTxSigComponent(unsignedTx, index, spendingInfo.output, flags)
     }
   }
+}
+
+sealed trait SingleSigner[-SpendingInfo <: UTXOSpendingInfoSingle]
```
nkohen, line 108: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -118,6 +87,68 @@ sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
...
+    extends SignerUtils {
+
+  def signSingle(
```
nkohen, line 117: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -151,6 +182,99 @@ sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
   }
 }
 
+sealed trait BitcoinSignerSingle[-SpendingInfo <: BitcoinUTXOSpendingInfoSingle]
```
nkohen, line 220: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -185,74 +309,62 @@ object BitcoinSigner {
...
-    val satisfyEmptyScriptSig =
-      Future.successful(NonStandardScriptSignature("0151"))
-
-    updateScriptSigInSigComponent(unsignedTx,
-                                  inputIndex.toInt,
-                                  output,
-                                  satisfyEmptyScriptSig)
-  }
-}
+sealed abstract class SingleKeyBitcoinSigner[
```
nkohen, line 369: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -185,74 +309,62 @@ object BitcoinSigner {
...
+    -SpendingInfo <: BitcoinUTXOSpendingInfo]
+    extends BitcoinSigner[SpendingInfo]
+    with BitcoinSignerSingle[SpendingInfo]
 
-object EmptySigner extends EmptySigner
+sealed abstract class RawSingleKeyBitcoinSigner[
```
nkohen, line 375: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -63,6 +62,35 @@ sealed abstract class UTXOSpendingInfo {
   def conditionalPath: ConditionalPath
 }
 
+/**
+  * Contains the information required to spend an unspent transaction output (UTXO)
```
nkohen: nit: to fully spend

nkohen: reference UTXOSpendingInfoSingle

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -265,21 +366,65 @@ object BitcoinUTXOSpendingInfo {
   }
 }
 
-/** This represents the information needed to be spend scripts like
-  * [[org.bitcoins.core.protocol.script.P2PKHScriptPubKey p2pkh]] or [[org.bitcoins.core.protocol.script.P2PKScriptPubKey p2pk]]
-  * scripts. Basically there is no nesting that requires a redeem script here*/
-sealed trait RawScriptUTXOSpendingInfo extends BitcoinUTXOSpendingInfo {
-  override def outPoint: TransactionOutPoint
-  override def amount: CurrencyUnit
+sealed trait RawScriptUTXOSpendingInfoSingle
```
nkohen: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -454,35 +680,111 @@ case class ConditionalSpendingInfo(
...
   }
+
+  override def signer: Sign = nestedSpendingInfo.signer
+
+  override val requiredSigs: Int = nestedSpendingInfo.requiredSigs
 }
 
-/** This is the case where we are spending a [[org.bitcoins.core.protocol.script.WitnessScriptPubKeyV0 witness v0 script]]  */
-sealed trait SegwitV0NativeUTXOSpendingInfo extends BitcoinUTXOSpendingInfo {
+sealed trait SegwitV0NativeUTXOSpendingInfoSingle
```
nkohen: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -578,52 +914,117 @@ case class UnassignedSegwitNativeUTXOSpendingInfo(
...
-case class P2SHNoNestSpendingInfo(
-    outPoint: TransactionOutPoint,
-    amount: CurrencyUnit,
-    scriptPubKey: P2SHScriptPubKey,
-    signers: Seq[Sign],
-    hashType: HashType,
-    redeemScript: RawScriptPubKey,
-    conditionalPath: ConditionalPath)
-    extends P2SHSpendingInfo {
+sealed trait P2SHSpendingInfoSingle extends BitcoinUTXOSpendingInfoSingle {
```
nkohen: scaladoc

nkohen (commented):
Scaladoc markers for myself

benthecarman (approved):
LGTM

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -13,43 +13,7 @@ import scodec.bits.ByteVector
...
-    * @param spendingInfoToSatisfy - specifies the UTXOSpendingInfo whose ScriptPubKey needs a ScriptSignature to be generated
-    * @return
-    */
-  def sign(
-      spendingInfo: UTXOSpendingInfo,
-      unsignedTx: Transaction,
-      isDummySignature: Boolean,
-      spendingInfoToSatisfy: SpendingInfo)(
-      implicit ec: ExecutionContext): Future[TxSigComponent]
+abstract private[signer] class SignerUtils {
```
Christewart: nit: Can't we just make this `sealed` instead of using this `private[signer]` stuff?

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -151,6 +216,102 @@ sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
...
+  */
+sealed trait BitcoinSignerSingle[-SpendingInfo <: BitcoinUTXOSpendingInfoSingle]
+    extends SingleSigner[SpendingInfo] {
+  override def signSingle(
+      spendingInfo: UTXOSpendingInfoSingle,
+      unsignedTx: Transaction,
+      isDummySignature: Boolean,
+      spendingInfoToSatisfy: SpendingInfo)(
+      implicit ec: ExecutionContext): Future[
+    (ECPublicKey, ECDigitalSignature)] = {
```
Christewart, line 228: I would prefer to make a case class for `(ECPublicKey, ECDigitalSignature)`, i think this will be useful else where.

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -67,16 +31,21 @@ sealed abstract class Signer[-SpendingInfo <: UTXOSpendingInfo] {
   protected val flags: Seq[ScriptFlag] = Policy.standardFlags
 
   protected def relevantInfo(
-      spendingInfo: UTXOSpendingInfo,
+      spendingInfo: UTXOSpendingInfoSingle,
       unsignedTx: Transaction): (Seq[Sign], TransactionOutput, UInt32, HashType) = {
-    (spendingInfo.signers,
+    val signers = spendingInfo match {
+      case info: UTXOSpendingInfo    => info.signers
```
Christewart: This is an unfortunate consequence of the design, ordering is is _very_ important here isn't it? Since `UTXOSpendingInfo` and `UTXOSpendingInfoSingle` aren't a mutually exclusive ADT (`UTXOSpendingInfo extends UTXOSpendingInfoSingle`). 

If this ordering isn't correct we will just silently sign with Single by default. This is really more of a standard OO inheritance based programming than using ADTs in my opinion

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -34,10 +34,9 @@ import org.bitcoins.core.script.crypto.HashType
 import org.bitcoins.core.util.CryptoUtil
 
 /**
-  * Contains the information required to spend a unspent transaction output (UTXO)
-  * on a blockchain.
+  * Contains the information required to sign an unspent transaction output (UTXO) for a single key
   */
-sealed abstract class UTXOSpendingInfo {
+sealed abstract class UTXOSpendingInfoSingle {
```
Christewart: My proposal to make this a true ADT: 

```scala
sealed abstract class UTXOSpendingInfo{

  /** The funding transaction's txid and the index of the output in the transaction we are spending */
  def outPoint: TransactionOutPoint

  def amount: CurrencyUnit

  def scriptPubKey: ScriptPubKey

  /** the actual output itself we are spending */
  def output: TransactionOutput = {
    TransactionOutput(value = amount, scriptPubKey = scriptPubKey)
  }

  /** The signer signing in the output above */
  def signer: Sign

  def hashType: HashType

  def redeemScriptOpt: Option[ScriptPubKey]

  def scriptWitnessOpt: Option[ScriptWitness]

  def conditionalPath: ConditionalPath
}

sealed trait UTXOSpendingInfoSingle extends UTXOSpendingInfo

/**
  * Contains the information required to fully sign an unspent transaction output (UTXO)
  * on a blockchain.
  *
  * If you want to partially sign a UTXO you will likely need to use UTXOSpendingInfoSingle.
  */
sealed abstract class UTXOSpendingInfoFull extends UTXOSpendingInfo { ... 
```

This gives us a disjoint type between `UTXOSpendingInfoFull` and `UTXOSpendingInfoSingle` so we can safely pattern match and use the compiler's exhaustiveness checks.

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -536,33 +848,67 @@ case class P2WPKHV0SpendingInfo(
...
   override def conditionalPath: ConditionalPath =
     ConditionalPath.NoConditionsLeft
 }
 
-case class P2WSHV0SpendingInfo(
+sealed trait P2WSHV0SpendingInfo extends SegwitV0NativeUTXOSpendingInfoSingle {
+  require(
+    CryptoUtil
+      .sha256(scriptWitness.redeemScript.asmBytes) == scriptPubKey.scriptHash,
+    "Witness has incorrect script")
```
Christewart: nit: In general with require statements it is very nice to put the inputs into the error message so you can see what it was that failed

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
benthecarman: We will have one once the PSBT side of this is implemented

benthecarman:
After working with this, I think it might be useful to have a `UTXOSpendingInfoSingle.apply(signers: Seq(Sign)): UTXOSpendingInfo` to turn a single singer into a full signer

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -527,132 +847,229 @@ case class P2WPKHV0SpendingInfo(
...
+    outPoint: TransactionOutPoint,
+    amount: CurrencyUnit,
+    scriptPubKey: P2WSHWitnessSPKV0,
+    signer: Sign,
+    hashType: HashType,
+    scriptWitness: P2WSHWitnessV0,
+    conditionalPath: ConditionalPath)
+    extends P2WSHV0SpendingInfo
+    with SegwitV0NativeUTXOSpendingInfoSingle {
+  override val nestedSpendingInfo: RawScriptUTXOSpendingInfoSingle = {
```
Christewart, line 888: Maybe in a later PR we can just pass in the `nestedSpendingInfo` as a parameter, along with the `scriptPubKey`, and `scriptWitness`? This seems to be a pattern here where we re-construct the `nestedSpendingInfo` inside of the case class.

Christewart (approved):

PR #1019, opened 2020-01-09
PSBT Generators

nkohen:
1. Introduces PSBT constructors and turns them into generators.
2. Adds a test using these generators for serialization symmetry
3. Adds a test using these generators for PSBT finalizing.
4. Fixes bugs found

PR #1025, opened 2020-01-14
Psbt Signer

benthecarman:
Built on top of #1020, to review just look at the diff in `84d04d7`

Property based tests can be added when #1019 is finalized


core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -231,4 +232,66 @@ class PSBTTest extends BitcoinSUnitTest {
...
+        .sign(psbt = firstSig,
+              inputIndex = 1,
+              signer = privKey1,
+              isDummySignature = false)
+
+    } yield {
+      assert(signedPsbt.bytes == expectedPsbt0.bytes)
+    }
+
+    Await.result(firstResultF, 5.seconds)
```
nkohen: Rather than using `Await.result` on either of these you should instead make `PSBTTest extend BitcoinSAsyncTest` like I do in #1019

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -231,4 +232,66 @@ class PSBTTest extends BitcoinSUnitTest {
...
+        .sign(psbt = firstSig,
+              inputIndex = 1,
+              signer = privKey3,
+              isDummySignature = false)
+
+    } yield {
+      assert(signedPsbt.bytes == expectedPsbt1.bytes)
+    }
+
+    Await.result(secondResultF, 5.seconds)
```
nkohen: ditto

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -243,6 +246,46 @@ sealed trait BitcoinSignerSingle[-SpendingInfo <: BitcoinUTXOSpendingInfoSingle]
...
+
+    sigAndPubKeyF.map { sigAndPubKey =>
+      val pubKey = sigAndPubKey._1
+      val signature = sigAndPubKey._2
+
+      val map = psbt.inputMaps(inputIndex)
+      val newElements = map.elements :+ PartialSignature(pubKey, signature)
+
+      val newInputMaps =
+        psbt.inputMaps.updated(inputIndex, InputPSBTMap(newElements))
```
nkohen: It would be nice if `PSBT` had more methods available for doing stuff like this so that users of `PSBT` don't have to go updating the guts of `Vector`s. For example in this case it would be nice to have a `PSBT.addSignature(index: Int, pubKey: ECPublicKey, sig: ECDigitalSignature)`

nkohen (commented):
I assume more will change as 1020 gets worked on but looks pretty much good to me for now

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -247,7 +244,7 @@ class PSBTTest extends BitcoinSUnitTest {
...
 
-    val firstResultF = for {
+    for {
```
nkohen: the `Future` returned here needs to be put in a val that is mapped on in the for below, otherwise this could fail but isn't part of the `FutureOutcome` registered for this test and so the test could still pass

nkohen (commented):
Just one nit on the new stuff

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -11,19 +11,14 @@ import org.bitcoins.core.protocol.script.{
...
-  GenUtil,
-  PSBTGenerators,
-  ScriptGenerators
-}
+import org.bitcoins.testkit.core.gen._
 import org.bitcoins.testkit.util.BitcoinSAsyncTest
 import scodec.bits._
 
 import scala.annotation.tailrec
+import scala.concurrent.{Await, ExecutionContext}
```
nkohen: Unused

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -237,4 +232,57 @@ class PSBTTest extends BitcoinSAsyncTest {
...
+
+    val privKey2 = ECPrivateKey.fromWIFToPrivateKey(
+      "cT7J9YpCwY3AVRFSjN6ukeEeWY6mhpbJPxRaDaP5QTdygQRxP9Au")
+
+    val privKey3 = ECPrivateKey.fromWIFToPrivateKey(
+      "cNBc3SWUip9PPm1GjRoLEJT6T41iNzCYtD7qro84FMnM5zEqeJsE")
+
+    for {
+      firstSig0 <- BitcoinSignerSingle
+        .sign(psbt = unsignedPsbt,
```
nkohen: Rather than having users calling `BitcoinSignerSingle.sign` I think it would be nice to put a sign method on PSBT itself so that you could do
```
unsignedPSBT.sign(inputIndex = 0, signer = privKey0)
```
(where `isDummySignature` defaults to `false`)

Not a strong opinion here but I think it looks nicer and requires less bitcoin-s knowledge from psbt users

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1436,7 +1449,25 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord]) extends PSBTMap {
...
+
+    BitcoinUTXOSpendingInfoFull(
+      outPoint = spendingInfoSingle.outPoint,
+      output = spendingInfoSingle.output,
+      signers = signers,
+      redeemScriptOpt = spendingInfoSingle.redeemScriptOpt,
+      scriptWitnessOpt = spendingInfoSingle.scriptWitnessOpt,
+      hashType = spendingInfoSingle.hashType,
+      conditionalPath = conditionalPath
+    )
```
nkohen, line 1476: If you want, feel free to add a util method to some companion object that does this. For example you could add another constructor for `BitcoinUTXOSpendingInfoFull` that takes in a `Vector[Sign]` and a `BitcoinUTXOSpendingInfoSingle` and does this.

nkohen (commented):
Just a couple small things

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -237,4 +230,38 @@ class PSBTTest extends BitcoinSAsyncTest {
...
+
+    for {
+      firstSig0 <- unsignedPsbt.sign(inputIndex = 0, signer = privKey0)
+      signedPsbt0 <- firstSig0.sign(inputIndex = 1, signer = privKey1)
+
+      firstSig1 <- unsignedPsbt.sign(inputIndex = 0, signer = privKey2)
+      signedPsbt1 <- firstSig1.sign(inputIndex = 1, signer = privKey3)
+    } yield {
+      assert(signedPsbt0.bytes == expectedPsbt0.bytes)
+      assert(signedPsbt1.bytes == expectedPsbt1.bytes)
```
nkohen: Is there a reason that calls to `_.bytes` is required here? Are they not equal in the usual sense?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -132,6 +115,18 @@ case class PSBT(
     }
   }
 
+  def sign(
```
nkohen, line 127: nit: add a scaladoc with `@param` tags so that this is a nice interface for new users

nkohen (commented):
Just two more nits

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
benthecarman: Was doing this because I was getting weird error messages before, however, don't need them now

nkohen (approved):

Christewart (approved):

PR #1026, opened 2020-01-14
Increase code coverage on PSBT tests

benthecarman:
Adds some basic tests for PSBTs to increase code coverage, after generators are added then coverage should be increased even more


core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
```diff
@@ -157,6 +159,24 @@ class PSBTSerializerTest extends BitcoinSUnitTest {
...
+
+    val psbtWithPoRC =
+      PSBT(psbt.globalMap,
+           psbt.inputMaps.updated(0, InputPSBTMap(inputElements)),
+           psbt.outputMaps)
+
+    assert(
+      psbtWithPoRC.inputMaps.head
+        .getRecords(ProofOfReservesCommitmentKeyId)
+        .size == 1)
```
nkohen: nit: rather than asserting `size == 1` you should assert that `.headOption.contains(proofOfReservesCommitment)`

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -134,6 +177,13 @@ class PSBTTest extends BitcoinSUnitTest {
     assert(psbt.extractTransaction == expected)
   }
 
+  it must "fail to extract a transaction from a non-finalized PSBT" in {
+    val psbt = PSBT(
+      "70736274ff0100a00200000002ab0949a08c5af7c49b8212f417e2f15ab3f5c33dcf153821a8139f877a5b7be40000000000feffffffab0949a08c5af7c49b8212f417e2f15ab3f5c33dcf153821a8139f877a5b7be40100000000feffffff02603bea0b000000001976a914768a40bbd740cbe81d988e71de2a4d5c71396b1d88ac8e240000000000001976a9146f4620b553fa095e721b9ee0efe9fa039cca459788ac000000000001076a47304402204759661797c01b036b25928948686218347d89864b719e1f7fcf57d1e511658702205309eabf56aa4d8891ffd111fdf1336f3a29da866d7f8486d75546ceedaf93190121035cdc61fc7ba971c0b501a646a2a83b102cb43881217ca682dc86e2d73fa882920001012000e1f5050000000017a9143545e6e33b832c47050f24d3eeb93c9c03948bc787010416001485d13537f2e265405a34dbafa9e3dda01fb82308000000")
+
+    assertThrows[RuntimeException](psbt.extractTransaction)
```
nkohen: nit: I don't know how easy this would be to do but if it is easy, we should try to make this an `IllegalStateException`

nkohen (approved):
ACK

PR #1028, opened 2020-01-14
Formatting fix

benthecarman:


nkohen (approved):
Very odd that this was causing problems

PR #1030, opened 2020-01-15
Validate the PSBT unknown is not a known field

benthecarman:


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -757,7 +746,9 @@ object GlobalPSBTRecord {
   }
 
   case class Unknown(key: ByteVector, value: ByteVector)
-      extends GlobalPSBTRecord
+      extends GlobalPSBTRecord {
+    require(PSBTGlobalKeyId.fromKey(key) == PSBTGlobalKeyId.UnknownKeyId)
```
nkohen: nit: error message please (you can even make the `fromKey` into a `private val` and put it in the error message)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -857,7 +848,11 @@ object InputPSBTRecord {
     override val value: ByteVector = porCommitment
   }
 
-  case class Unknown(key: ByteVector, value: ByteVector) extends InputPSBTRecord
+  case class Unknown(key: ByteVector, value: ByteVector)
+      extends InputPSBTRecord {
+    require(key.size > 1)
+    require(PSBTInputKeyId.fromKey(key) == PSBTInputKeyId.UnknownKeyId)
```
nkohen: same as above, also, do you need both requires? Does the second not cover the first?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -944,7 +939,9 @@ object OutputPSBTRecord {
   }
 
   case class Unknown(key: ByteVector, value: ByteVector)
-      extends OutputPSBTRecord
+      extends OutputPSBTRecord {
+    require(PSBTOutputKeyId.fromKey(key) == PSBTOutputKeyId.UnknownKeyId)
```
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -990,6 +987,14 @@ object PSBTGlobalKeyId {
     case _: Byte                       => UnknownKeyId
   }
 
+  def fromKey(key: ByteVector): PSBTGlobalKeyId = {
```
nkohen: Since we are in the `PSBTGlobalKeyId` object can't this be called `fromBytes`? Same with the other two

nkohen: At the very least it should be `fromKeyBytes`

nkohen (approved):

PR #1031, opened 2020-01-15
PSBT Support

nkohen:
Introduces support for [Partially Signed Bitcoin Transactions (PSBTs)](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki).

Todo:

- [x] More property based tests

- [x] Increase coverage

- [x] PSBT.md documentation


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -50,6 +51,21 @@ object NonStandardScriptSignature
                 },
                 errorMsg = "")
   }
+
+  /** A script signature to be used in tests for signing EmptyScriptPubKey.
+    * This script pushes an OP_TRUE onto the stack, causing a successful spend.
+    */
+  val trivalTrue: NonStandardScriptSignature = fromAsmHex("0151")
```
nkohen: oops, meant to delete this val now

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -702,4 +702,26 @@ object BitcoinTxBuilder {
     val map = loop(utxos, Map.empty)
     BitcoinTxBuilder(destinations, map, feeRate, changeSPK, network)
   }
+
+  def emptyAllSigs(tx: Transaction): Transaction = {
```
nkohen: scaladoc

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOSpendingInfo.scala
```diff
@@ -353,6 +365,19 @@ object BitcoinUTXOSpendingInfoFull {
     }
   }
 
+  def apply(
+      spendingInfoSingle: UTXOSpendingInfoSingle,
+      signers: Vector[Sign]): BitcoinUTXOSpendingInfoFull =
```
nkohen: Should probably require that `signers.contains(spendingInfoSingle.signer)` so that people don't accidentally omit it

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -0,0 +1,327 @@
...
+    @tailrec
+    def loop(i: Int, accum: Vector[Sign]): Vector[Sign] = {
+      if (i <= 0) {
+        accum
+      } else {
+        loop(i - 1, accum :+ Sign.dummySign(ECPublicKey.freshPublicKey))
+      }
+    }
+    loop(size, Vector.empty)
+  }
```
nkohen: nit: I think this is nicer:
```
private def getDummySigners(size: Int, pubKey: ECPublicKey = ECPublicKey.freshPublicKey): Vector[Sign] = {
    Vector.fill(size)(Sign.dummySign(pubKey))
}
```

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -0,0 +1,327 @@
...
+    assert(spendingInfo.hashType == HashType.sigHashAll)
+    assert(spendingInfo.redeemScriptOpt.isEmpty)
+    assert(
+      spendingInfo.scriptWitnessOpt.contains(
+        P2WSHWitnessV0(RawScriptPubKey.fromAsmHex(
+          "5221029da12cdb5b235692b91536afefe5c91c3ab9473d8e43b533836ab456299c88712103372b34234ed7cf9c1fea5d05d441557927be9542b162eb02e1ab2ce80224c00b52ae"))))
+    assert(spendingInfo.conditionalPath == ConditionalPath.NoConditionsLeft)
+  }
+
+  it must "fail to create a valid UTXOSpendingInfo" in {
```
nkohen: nit: under what conditions?

nkohen (commented):
Initial review of all smaller files (i.e. all files but PSBT.scala which I will review next on its own)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+    globalMap: GlobalPSBTMap,
+    inputMaps: Vector[InputPSBTMap],
+    outputMaps: Vector[OutputPSBTMap])
+    extends NetworkElement {
+  require(
+    globalMap
+      .getRecords(UnsignedTransactionKeyId)
+      .size == 1)
+  require(inputMaps.size == transaction.inputs.size)
+  require(outputMaps.size == transaction.outputs.size)
```
nkohen: error messages

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+    outputMaps.foldLeft(ByteVector.empty)(_ ++ _.bytes)
+
+  val bytes: ByteVector = PSBT.magicBytes ++
+    globalMap.bytes ++
+    inputBytes ++
+    outputBytes
+
+  def transaction: Transaction =
+    globalMap
+      .getRecords[UnsignedTransaction](UnsignedTransactionKeyId)
```
nkohen: As a higher level structural comment:

I think that it is fine to have `getRecords` for internal `PSBT` calls but I think we should provide `def`s in each of `GlobalPSBTMap`, `InputPSBTMap`, and `OutputPSBTMap` for known things that return `Vector[...]` so that users don't have to call `getRecords`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+
+  /**
+    * Combiner defined by https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#combiner
+    * Takes another PSBT and adds all records that are not contained in this PSBT
+    * A record's distinctness is determined by its key
+    * @param other PSBT to be combined with
+    * @return A PSBT with the combined data of the two PSBTs
+    */
+  def combinePSBT(other: PSBT): PSBT = {
+    require(this.transaction.txId == other.transaction.txId,
```
nkohen, line 71: nit: is there a reason we are comparing `txid`s here and not just the whole `transaction`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+            "Can only combine PSBTs with the same global transaction.")
+
+    val global = this.globalMap.combine(other.globalMap)
+    val inputs = this.inputMaps
+      .zip(other.inputMaps)
+      .map(input => input._1.combine(input._2))
+    val outputs = this.outputMaps
+      .zip(other.outputMaps)
+      .map(output => output._1.combine(output._2))
```
nkohen: nit: on these `map`s I think that
```
.map { case (output, otherOutput) => output.combine(otherOutput) }
```
is nicer

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+    PSBT(global, inputs, outputs)
+  }
+
+  /** Finalizes this PSBT if possible
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer]]
+    * @return None if there exists any input which is not ready for signing
+    */
+  def finalizePSBT: Option[PSBT] = {
+    if (isFinalized) {
+      Some(this)
```
nkohen: probably want to throw an `IllegalStateException` here

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      .map(output => output._1.combine(output._2))
+
+    PSBT(global, inputs, outputs)
+  }
+
+  /** Finalizes this PSBT if possible
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer]]
+    * @return None if there exists any input which is not ready for signing
+    */
+  def finalizePSBT: Option[PSBT] = {
```
nkohen: I think we should try to move this to a `Try[PSBT]` with nicer error messages in failure cases rather than just `None`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+   * @param conditionalPath Represents the spending branch being taken in a ScriptPubKey's execution
+   * @param isDummySignature Do not sign the tx for real, just use a dummy signature, this is useful for fee estimation
+   * @return
+   */
+  def sign(
+      inputIndex: Int,
+      signer: Sign,
+      conditionalPath: ConditionalPath = ConditionalPath.NoConditionsLeft,
+      isDummySignature: Boolean = false)(
+      implicit ec: ExecutionContext): Future[PSBT] =
```
nkohen: nit: braces

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  /**
+    * Adds tx to the indexed InputPSBTMap to either the NonWitnessOrUnknownUTXO
+    * or WitnessUTXO field depending on the tx and available information in the PSBT
+    * @param tx Transaction to add to PSBT
+    * @param index index of the InputPSBTMap to add tx to
+    * @return PSBT with added tx
+    */
+  def addTransactionToInput(tx: Transaction, index: Int): PSBT = {
```
nkohen: nit: I think `addUTXO` or `addUTXOToInput` is a little more descriptive

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      index < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val elements = {
+      val previousElements = inputMaps(index).elements
+      val txIn = transaction.inputs(index)
+
+      if (txIn.previousOutput.vout.toInt < tx.outputs.size) {
```
nkohen: nit:
```
val previousElements = inputMaps(index).elements
val txIn = transaction.inputs(index)

val elements = if (txIn.previousOutput.vout.toInt < tx.outputs.size) { ...
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+          previousElements.filterNot(_.key.head == NonWitnessUTXOKeyId.byte) :+ NonWitnessOrUnknownUTXO(
+            tx)
+        }
+      } else {
+        previousElements
+      }
+    }
+
+    val newInputMaps =
+      inputMaps.patch(index, Seq(InputPSBTMap(elements)), 1)
```
nkohen: nit: I think that `inputMaps.updated(...)` is more readable

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  /**
+    * Adds script to the indexed InputPSBTMap to either the RedeemScript
+    * or WitnessScript field depending on the script and available information in the PSBT
+    * @param script ScriptPubKey to add to PSBT
+    * @param index index of the InputPSBTMap to add script to
+    * @return PSBT with added script
+    */
+  def addScriptToInput(script: ScriptPubKey, index: Int): PSBT = {
```
nkohen: nit: I think I prefer `addRedeemOrWitnessScript(script: ScriptPubKey, inputIndex: Int)`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      if (!isWitScript && hasWitScript) {
+        previousElements.filterNot(
+          _.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte) :+ InputPSBTRecord
+          .WitnessScript(script.asInstanceOf[RawScriptPubKey])
+      } else {
+        previousElements.filterNot(
+          _.key.head == PSBTInputKeyId.RedeemScriptKeyId.byte) :+ InputPSBTRecord
+          .RedeemScript(script)
+      }
+    }
```
nkohen: nit:
```
      val previousElements = inputMaps(index).elements

      val isWitScript = WitnessScriptPubKey.isWitnessScriptPubKey(script.asm)
      val redeemScriptOpt =
        inputMaps(index).getRecords[InputPSBTRecord.RedeemScript](
          PSBTInputKeyId.RedeemScriptKeyId).headOption
      val hasWitScript = redeemScript.isDefined && WitnessScriptPubKey
        .isWitnessScriptPubKey(redeemScript.get.redeemScript.asm)

      val elements = if (!isWitScript && hasWitScript) {
        previousElements.filterNot(
          _.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte) :+ InputPSBTRecord
          .WitnessScript(script.asInstanceOf[RawScriptPubKey])
      } else {
        previousElements.filterNot(
          _.key.head == PSBTInputKeyId.RedeemScriptKeyId.byte) :+ InputPSBTRecord
          .RedeemScript(script)
      }
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+
+    val newMap = InputPSBTMap(elements).compressMap(transaction.inputs(index))
+    val newInputMaps = inputMaps.patch(index, Seq(newMap), 1)
```
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      if (!isWitScript && hasWitScript) {
+        previousElements.filterNot(
+          _.key.head == PSBTOutputKeyId.WitnessScriptKeyId.byte) :+ OutputPSBTRecord
+          .WitnessScript(script)
+      } else {
+        previousElements.filterNot(
+          _.key.head == PSBTOutputKeyId.RedeemScriptKeyId.byte) :+ OutputPSBTRecord
+          .RedeemScript(script)
+      }
+    }
```
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+
+    val newMap = OutputPSBTMap(elements)
+    val newOutputMaps = outputMaps.patch(index, Seq(newMap), 1)
```
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      index < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val elements = {
+      val previousElements = inputMaps(index).elements
+
+      val keyOpt = extKey.deriveChildPubKey(path)
+      if (keyOpt.isSuccess && !previousElements.exists(_.key == keyOpt.get.bytes
```
nkohen: nit:
```
val previousElements = inputMaps(index).elements

val keyOpt = extKey.deriveChildPubKey(path)
val elements = if ...
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+          else extKey.fingerprint
+        previousElements :+ InputPSBTRecord.BIP32DerivationPath(keyOpt.get.key,
+                                                                fp,
+                                                                path)
+      } else {
+        previousElements
+      }
+    }
+
+    val newInputMaps = inputMaps.patch(index, Seq(InputPSBTMap(elements)), 1)
```
nkohen: ditto above

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  /**
+    * Adds the BIP32Path to the indexed InputPSBTMap to the BIP32DerivationPath field
+    * @param extKey ExtKey to derive key from
+    * @param path path of key to add to PSBT
+    * @param index index of the InputPSBTMap to add the BIP32Path to
+    * @return PSBT with added BIP32Path
+    */
+  def addKeyPathToInput(extKey: ExtKey, path: BIP32Path, index: Int): PSBT = {
```
nkohen, line 413: do we need to require that there is some relationship between `extKey` and a global record? Just asking to be sure cause currently we don't require anything like that to my knowledge

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  /**
+    * Adds the BIP32Path to the indexed OutputPSBTMap to the BIP32DerivationPath field
+    * @param extKey ExtKey to derive key from
+    * @param path path of key to add to PSBT
+    * @param index index of the OutputPSBTMap to add the BIP32Path to
+    * @return PSBT with added BIP32Path
+    */
+  def addKeyPathToOutput(extKey: ExtKey, path: BIP32Path, index: Int): PSBT = {
```
nkohen, line 434: ditto both of the above here, also if there is any easy way to factor stuff out with these two functions that would be nice

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+        bytes: ByteVector,
+        numOutputs: Int,
+        accum: Seq[OutputPSBTMap]): Vector[OutputPSBTMap] = {
+      if (numOutputs <= 0 || bytes.isEmpty) {
+        accum.toVector
+      } else {
+        val map = OutputPSBTMap.fromBytes(bytes)
+        outputLoop(bytes.drop(map.bytes.size), numOutputs - 1, accum :+ map)
+      }
+    }
```
nkohen: I believe you can consolidate these two `loop` functions into one:
```
@tailrec
def mapLoop[MapType <: PSBTMap](bytes: ByteVector, numMaps: Int, accum: Seq[MapType], factory: Factory[MapType]): Vector[MapType] = {
    if (numMaps <=0 || bytes.isEmpty) {
        accum.toVector
    } else {
        val newMap = factory.fromBytes(bytes)
        mapLoop(bytes.drop(newMap.bytes.size, numMaps - 1, accum :+ map, factory)
    }
}
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  def fromBase64(base64: String): PSBT = {
+    ByteVector.fromBase64(base64) match {
+      case None =>
+        throw new IllegalArgumentException(
+          s"String given was not in base64 format, got: $base64")
+      case Some(bytes) => fromBytes(bytes)
+    }
+  }
+
+  def fromUnsignedTx(unsignedTx: Transaction): PSBT = {
```
nkohen, line 682: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+        OutputPSBTRecord.Unknown(key, value)
+    }
+  }
+}
+
+/**
+  * A PSBTKeyId refers to the first byte of a key that signifies which kind of key-value map
+  * is in a given PSBTRecord
+  */
+sealed trait PSBTKeyId {
```
nkohen: I think we might be able to avoid the `getRecords[RecordType](KeyObj)` by introducing a type member in `PSBTKeyId`
```
type RecordType <: PSBTRecord
```
and override it in subtypes to have things like
```
final case object UnsignedTransactionKeyId extends PSBTGlobalKeyId {
  type RecordType = GlobalPSBTRecord.UnsignedTransaction
  override val byte: Byte = 0x00.byteValue
}
```
And if I'm right this should let us define
```
def getRecords(key: PSBTGlobalKeyId): Vector[key.RecordType] = { ...
```
which will make all calls to `getRecords` much nicer. Again, not sure this works but I think we should do it if it does work.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+sealed trait PSBTMap extends NetworkElement {
+  require(elements.map(_.key).groupBy(identity).values.forall(_.length == 1),
+          "All keys must be unique.")
+
+  def elements: Vector[PSBTRecord]
+
+  def bytes: ByteVector =
+    elements
+      .sortBy(_.key)
+      .foldLeft(ByteVector.empty)(_ ++ _.bytes) ++ hex"00"
```
nkohen: nit: give `hex"00"` a name like `endByte` or something like that for readability

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+      .filter(element => PSBTGlobalKeyId.fromByte(element.key.head) == key)
+      .asInstanceOf[Vector[T]]
+  }
+
+  /**
+    * Takes another GlobalPSBTMap and adds all records that are not contained in this GlobalPSBTMap
+    * @param other GlobalPSBTMap to be combined with
+    * @return A GlobalPSBTMap with the combined data of the two GlobalPSBTMaps
+    */
+  def combine(other: GlobalPSBTMap): GlobalPSBTMap = {
```
nkohen: should this function start with a require that the `UnsignedTransaction`s within are the same?

nkohen: wouldn't want to end up with two of them

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  def isFinalized: Boolean =
+    getRecords(FinalizedScriptSigKeyId).nonEmpty || getRecords(
+      FinalizedScriptWitnessKeyId).nonEmpty
+
+  /** Finalizes this input if possible, returning None if not */
+  def finalize(input: TransactionInput): Option[InputPSBTMap] = {
+    if (isFinalized) {
+      Some(this)
```
nkohen: ditto above about throwing here, and turning this function into returning a `Try` rather than an `Option`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+  }
+
+  /**
+    * Takes another InputPSBTMap and adds all records that are not contained in this InputPSBTMap
+    * A record's distinctness is determined by its key
+    * @param other InputPSBTMap to be combined with
+    * @return A InputPSBTMap with the combined data of the two InputPSBTMaps
+    */
+  def combine(other: InputPSBTMap): InputPSBTMap = {
+    InputPSBTMap((this.elements ++ other.elements).distinct)
```
nkohen: we should probably ensure they don't conflict right?

nkohen: same with other `combine`s

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+          InputPSBTMap(utxos ++ finalizedSigs)
+      }
+    }
+  }
+
+  /** Constructs a full (ready to be finalized) but unfinalized InputPSBTMap
+    * from a UTXOSpendingInfoFull, the corresponding PSBT's unsigned transaction,
+    * and if this is a non-witness spend, the transaction being spent
+    */
+  def fromUTXOSpendingInfo(
```
nkohen: this entire function should be re-written to use `spendingInfo.toSingles` rather than injecting side-effects into `Sign.signFunction`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1799 @@
...
+        case p2sh: P2SHNestedSegwitV0UTXOSpendingInfoFull =>
+          builder.+=(RedeemScript(p2sh.redeemScript))
+          p2sh.scriptWitness match {
+            case p2wsh: P2WSHWitnessV0 =>
+              builder.+=(WitnessScript(p2wsh.redeemScript))
+            case _: P2WPKHWitnessV0 => ()
+          }
+        case p2wsh: P2WSHV0SpendingInfo =>
+          builder.+=(WitnessScript(p2wsh.scriptWitness.redeemScript))
+        case _: RawScriptUTXOSpendingInfoFull | _: P2WPKHV0
```
nkohen: might be nice to give this a name and put it somewhere so it can be used here and elsewhere

nkohen: This `fromBytes` pattern can likely be defined in the super class

nkohen (commented):
Reviewed PSBT.scala

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: `txid` is just a hash of the transaction so it's a faster comparison

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: We don't current do any `require`s, that seems like it may be a good idea however, it is not specified in the BIP that an `xpub` must be defined to add a `BIP32Path`.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: How would they conflict?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen this is a beast of a function and I think you have a better understanding of it

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: two different signatures for the same public key, different UTXOs for a given input, etc.
calling `_.distinct` here will not fail on cases like these

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Not really sure how to do this, maybe you can take a stab at it
@nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: If that happens they would get an error for making a PSBTMap with the same key.

When I made this I tried doing `e => e.distinctBy(e.key)` however, this function isn't available in older versions

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen not sure what you mean here

benthecarman: @nkohen not sure what you mean here

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -0,0 +1,326 @@
...
+
+    val txT = psbt.extractTransactionAndValidate
+
+    assert(txT.isSuccess)
+
+    val tx = psbt.extractTransaction
+
+    assert(tx == txT.get)
```
benthecarman: Why do we need to run `extractTransaction` twice

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -52,6 +53,19 @@ object NonStandardScriptSignature
   }
 }
 
+/** A script signature to be used in tests for signing EmptyScriptPubKey.
+  * This script pushes an OP_TRUE onto the stack, causing a successful spend.
+  */
+case object TrivialTrueScriptSignature extends ScriptSignature {
+  override def signatures: Seq[ECDigitalSignature] = Nil
```
Christewart: should be `val` rather than `def`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -505,6 +519,28 @@ object ConditionalScriptSignature
     fromAsm(nestedScriptSig.asm.:+(conditionAsm))
   }
 
+  @scala.annotation.tailrec
+  def apply(
```
Christewart: can we use a real name here rather than just `apply()`? `fromNestedScriptSig()` or `fromNested()`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+import scala.util.{Failure, Success, Try}
+
+case class PSBT(
+    globalMap: GlobalPSBTMap,
+    inputMaps: Vector[InputPSBTMap],
+    outputMaps: Vector[OutputPSBTMap])
+    extends NetworkElement {
+  require(
+    inputMaps.size == transaction.inputs.size,
+    "There must be an input map for every input in the global transaction")
```
Christewart: adding the inputs to the error messages are always a good idea so it makes it easier to debug.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  def ==(p: PSBT): Boolean = this.bytes == p.bytes
+  def !=(p: PSBT): Boolean = !(this == p)
+
+  private val inputBytes: ByteVector =
+    inputMaps.foldLeft(ByteVector.empty)(_ ++ _.bytes)
+
+  private val outputBytes: ByteVector =
+    outputMaps.foldLeft(ByteVector.empty)(_ ++ _.bytes)
+
+  val bytes: ByteVector = PSBT.magicBytes ++
```
Christewart: override

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    val outputs = this.outputMaps
+      .zip(other.outputMaps)
+      .map { case (output, otherOutput) => output.combine(otherOutput) }
+
+    PSBT(global, inputs, outputs)
+  }
+
+  /** Finalizes this PSBT if possible
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer]]
+    * @return None if there exists any input which is not ready for signing
```
Christewart: Outdated scaladoc i think? We return a `Try`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+        if (outIsWitnessScript || hasWitScript) {
+          previousElements.filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
+            out)
+        } else {
+          previousElements.filterNot(_.key.head == NonWitnessUTXOKeyId.byte) :+ NonWitnessOrUnknownUTXO(
+            tx)
+        }
+      } else {
+        previousElements
```
Christewart: isn't this a failure case here? Shouldn't we do something else besides return the previous elements? Essentially this is the case where we have a index out of bounds no?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val previousElements = inputMaps(index).elements
+
+    val isWitScript = WitnessScriptPubKey.isWitnessScriptPubKey(script.asm)
+    val redeemScriptOpt = inputMaps(index).redeemScriptOpt
+    val hasWitScript = redeemScriptOpt.isDefined && WitnessScriptPubKey
+      .isWitnessScriptPubKey(redeemScriptOpt.get.redeemScript.asm)
```
Christewart, line 220: this .get isn't safe is it?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    val elements =
+      if (txIn.previousOutput.vout.toInt < tx.outputs.size) {
+        val out = tx.outputs(txIn.previousOutput.vout.toInt)
+
+        val outIsWitnessScript =
+          WitnessScriptPubKey.isWitnessScriptPubKey(out.scriptPubKey.asm)
+        val hasWitScript = inputMaps(index).witnessScriptOpt.isDefined
+
+        if (outIsWitnessScript || hasWitScript) {
+          previousElements.filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
```
Christewart: this pattern is everywhere: `previousElements.filterNot()`, let's move this into a helper method?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  }
+
+  /**
+    * Adds script to the indexed OutputPSBTMap to either the RedeemScript
+    * or WitnessScript field depending on the script and available information in the PSBT
+    * @param script ScriptPubKey to add to PSBT
+    * @param index index of the OutputPSBTMap to add script to
+    * @return PSBT with added script
+    */
+  def addRedeemOrWitnessScriptToOutput(
```
Christewart, line 304: Are `addRedeemOrWitnessScriptToOutput` and `addRedeemOrWitnessScriptToInput` the exact same except for what map it gets added to? If so, we should just make a helper method and pass in the Map it needs to be added to as a parameter so we don't have this redundant code

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+      maps: Vector[MapType],
+      makeRecord: (ECPublicKey, ByteVector, BIP32Path) => RecordType,
+      makeMap: Vector[RecordType] => MapType): Vector[MapType] = {
+    require(
+      index < maps.size,
+      s"index must be less than the number of output maps present in the psbt, $index >= ${outputMaps.size}")
+    require(!isFinalized, "Cannot update a PSBT that is finalized")
+
+    val previousElements = maps(index).elements
+    val keyOpt = extKey.deriveChildPubKey(path)
```
Christewart: this is a `Try` not a `Option`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+    lazy val expectedBytes =
+      keyOpt.get.bytes.+:(PSBTOutputKeyId.BIP32DerivationPathKeyId.byte)
+
+    val elements =
+      if (keyOpt.isSuccess && !previousElements.exists(_.key == expectedBytes)) {
+        val fp =
+          if (extKey.fingerprint == hex"00000000") {
```
Christewart: What's going on here?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  def extractTransactionAndValidate: Try[Transaction] = {
+    inputMaps.zipWithIndex.foldLeft(Try(extractTransaction)) {
+      case (txT, (inputMap, index)) =>
+        txT.flatMap { tx =>
+          val wUtxoOpt = inputMap.witnessUTXOOpt
+          val utxoOpt = inputMap.nonWitnessOrUnknownUTXOOpt
+
+          (wUtxoOpt, tx) match {
+            case (Some(wUtxo), wtx: WitnessTransaction) =>
+              val output = wUtxo.spentWitnessTransaction
```
Christewart: `.spentWitnessTransaction` probably should be named to `witnessUtxo` or something like that

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+    val newElements = inputMaps(inputIndex).elements :+ partialSignature
+
+    val newInputMaps =
+      inputMaps.updated(inputIndex, InputPSBTMap(newElements))
+
+    PSBT(globalMap, newInputMaps, outputMaps)
+  }
+
+  def extractTransactionAndValidate: Try[Transaction] = {
```
Christewart, line 504: scaladoc would be helpful

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+                  } else {
+                    Failure(
+                      new RuntimeException(
+                        s"Input $index was invalid: $inputResult"))
+                  }
+                case None =>
+                  logger.info(
+                    s"No UTXO record was provided for input $index, hence no validation was done for this input")
+
+                  Success(tx)
```
Christewart, line 553: This should be a success if there is no utxo record?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+object PSBT extends Factory[PSBT] {
+
+  // The known version of PSBTs this library can process defined by https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#version-numbers
+  final val knownVersions: Vector[UInt32] = Vector(UInt32.zero)
+
+  // The magic bytes and separator defined by https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#specification
+  final val magicBytes = hex"70736274ff"
+
+  def fromBytes(bytes: ByteVector): PSBT = {
```
Christewart: override

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    }
+  }
+
+  /**
+    * Creates an empty PSBT with only the global transaction record filled
+    * @param unsignedTx global transaction for the PSBT
+    * @return Created PSBT
+    */
+  def fromUnsignedTx(unsignedTx: Transaction): PSBT = {
+    val emptySigTx = BitcoinTxBuilder.emptyAllSigs(unsignedTx)
```
Christewart: aren't we assuming that the tx is already unsigned when passed in here? I don't think i would expect this method to empty signatures for me. Probably should have a `require()` verifying that the transaction is unsigned

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+    def map[T](
+        func: (UTXOSpendingInfoFull, Option[Transaction]) => T): Vector[T] = {
+      infoAndTxOpts.map { case (info, txOpt) => func(info, txOpt) }
+    }
+  }
+
+  object SpendingInfoAndNonWitnessTxs {
+
+    /** For use in tests, makes dummy transactions with the correct output for the PSBT */
```
Christewart: Why is this in source code if it is just used for tests?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    val outputMaps = unsignedTx.outputs.map(_ => OutputPSBTMap.empty).toVector
+
+    PSBT(globalMap, inputMaps, outputMaps)
+  }
+
+  /**
+    * Wraps a Vector of pairs of UTXOSpendingInfos and the Transactions whose outputs are spent.
+    * Note that this Transaction is only necessary when the output is non-segwit.
+    */
+  case class SpendingInfoAndNonWitnessTxs(
```
Christewart, line 703: If this is a `NonWitnessTx` can't we have the type be `Vector[(..., Option[BaseTransaction])]`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+
+    Future.sequence(inputMapFs).map { inputMaps =>
+      PSBT(globalMap, inputMaps, outputMaps)
+    }
+  }
+}
+
+sealed trait PSBTRecord extends NetworkElement {
+
+  def key: ByteVector
```
Christewart: we should probably have some sort of helper method in here to avoid the `filterNot()` we have everywhere when trying to determine key types.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    val keySize = CompactSizeUInt.calc(key)
+    val valueSize = CompactSizeUInt.calc(value)
+
+    keySize.bytes ++ key ++ valueSize.bytes ++ value
+  }
+}
+
+object PSBTRecord {
+
+  def fromBytes(bytes: ByteVector): (ByteVector, ByteVector) = {
```
Christewart: why doesn't this just return a `PSBTRecord`? If we do that, we can `extend Factory[PSBTRecord]` on the object as well.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+        val sig = ScriptSignature.fromAsmBytes(value)
+        FinalizedScriptSig(sig)
+      case FinalizedScriptWitnessKeyId =>
+        require(key.size == 1,
+                s"The key must only contain the 1 byte type, got: ${key.size}")
+
+        FinalizedScriptWitness(RawScriptWitnessParser.read(value))
+      case ProofOfReservesCommitmentKeyId =>
+        ProofOfReservesCommitment(value)
+      case _ =>
```
Christewart: `case _ =>` is a bad idea. isn't the only case remaining `UnknownKeyId`? We've designed this so nicely with types, let's get the full benefit of doing so!

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+                s"The key must only contain the 1 byte type, got: ${key.size}")
+
+        OutputPSBTRecord.WitnessScript(ScriptPubKey.fromAsmBytes(value))
+      case PSBTOutputKeyId.BIP32DerivationPathKeyId =>
+        val pubKey = ECPublicKey(key.tail)
+        val fingerprint = value.take(4)
+        val path = BIP32Path.fromBytesLE(value.drop(4))
+
+        OutputPSBTRecord.BIP32DerivationPath(pubKey, fingerprint, path)
+      case _ =>
```
Christewart: Same comment as above with respect to `case _ =>`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+        OutputPSBTRecord.Unknown(key, value)
+    }
+  }
+}
+
+/**
+  * A PSBTKeyId refers to the first byte of a key that signifies which kind of key-value map
+  * is in a given PSBTRecord
+  */
+sealed trait PSBTKeyId {
```
Christewart: Shouldn't this `PSBTKeyId` be a field in `PSBTRecord`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  def bytes: ByteVector =
+    elements
+      .sortBy(_.key)
+      .foldLeft(ByteVector.empty)(_ ++ _.bytes) :+ PSBTMap.separatorByte
+
+  protected def getRecords[KeyIdType <: PSBTKeyId](
+      key: KeyIdType,
+      keyIdFactory: PSBTKeyIdFactory[KeyIdType]): Vector[key.RecordType] = {
+    elements
+      .filter(element => keyIdFactory.fromByte(element.key.head) == key)
```
Christewart: Could we just use `.collect { } ` here and avoid the cast?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  def unsignedTransaction: UnsignedTransaction = {
+    getRecords(UnsignedTransactionKeyId).head
+  }
+
+  def extendedPublicKeys: Vector[XPubKey] = {
+    getRecords(XPubKeyKeyId)
+  }
+
+  def version: Version = {
+    getRecords(VersionKeyId).headOption.getOrElse(Version(UInt32.zero))
```
Christewart: any reason why we are defaulting to `UInt32.zero` here?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+    extends PSBTMap[InputPSBTRecord] {
+  import org.bitcoins.core.psbt.InputPSBTRecord._
+  import org.bitcoins.core.psbt.PSBTInputKeyId._
+
+  def nonWitnessOrUnknownUTXOOpt: Option[NonWitnessOrUnknownUTXO] = {
+    getRecords(NonWitnessUTXOKeyId).headOption
+  }
+
+  def witnessUTXOOpt: Option[WitnessUTXO] = {
+    getRecords(WitnessUTXOKeyId).headOption
```
Christewart: Why do we know that `.headOption` is right one? Couldn't it be at index 1 in the `Vector` returned here? The `getRecords()` stuff seems wonky

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+            val outputs = utxo.transactionSpent.outputs
+            Try(outputs(input.previousOutput.vout.toInt).scriptPubKey)
+        }
+      }
+
+      scriptPubKeyToSatisfyT.flatMap(finalize)
+    }
+  }
+
+  /** Finalizes this input if possible, returning None if not */
```
Christewart: scaladoc needs update, it's a `Try` not `None`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+      val records = utxos ++ recordsToAdd ++ unknowns
+      InputPSBTMap(records)
+    }
+
+    /** Turns the required PartialSignatures into a ScriptSignature and calls wipeAndAdd
+      * @return None if the requirement is not met
+      */
+    def collectSigs(
+        required: Int,
+        constructScriptSig: Seq[(ECDigitalSignature, ECPublicKey)] => ScriptSignature): Try[
```
Christewart: can't we use the `PartialSignature` type here? `Seq[PartialSignature] => ScriptSignature`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -0,0 +1,1870 @@
...
+  /** Constructs a finalized InputPSBTMap from a UTXOSpendingInfoFull,
+    * the corresponding PSBT's unsigned transaction, and if this is
+    * a non-witness spend, the transaction being spent
+    */
+  def finalizedFromUTXOSpendingInfo(
+      spendingInfo: UTXOSpendingInfoFull,
+      unsignedTx: Transaction,
+      nonWitnessTxOpt: Option[Transaction])(
+      implicit ec: ExecutionContext): Future[InputPSBTMap] = {
+    val sigComponentF = BitcoinSig
```
Christewart: There is way to much stuff in this file. This should be 3 or 4 files

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -702,4 +702,32 @@ object BitcoinTxBuilder {
...
     BitcoinTxBuilder(destinations, map, feeRate, changeSPK, network)
   }
+
+  /**
+    * Sets the ScriptSignature for every input in the given transaction to an EmptyScriptSignature
+    * as well as sets the witness to an EmptyWitness
+    * @param tx Transaction to empty signatures
+    * @return Transaction with no signatures
+    */
+  def emptyAllSigs(tx: Transaction): Transaction = {
```
Christewart: nit: `emptyAllScriptSigs`

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -234,18 +235,94 @@ sealed trait BitcoinSignerSingle[-SpendingInfo <: BitcoinUTXOSpendingInfoSingle]
...
+      signer: Sign,
+      conditionalPath: ConditionalPath = ConditionalPath.NoConditionsLeft,
+      isDummySignature: Boolean = false)(
+      implicit ec: ExecutionContext): Future[PSBT] = {
+    // if already signed by this signer
+    if (psbt
+          .inputMaps(inputIndex)
+          .partialSignatures
+          .exists(_.pubKey == signer.publicKey)) {
+      throw new IllegalArgumentException(
```
Christewart: `Future.failed(new IllegalArgumentException)` and add the public key to the error message

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -234,18 +235,94 @@ sealed trait BitcoinSignerSingle[-SpendingInfo <: BitcoinUTXOSpendingInfoSingle]
...
+    val tx = psbt.transaction
+    val spendingInfo =
+      psbt
+        .inputMaps(inputIndex)
+        .toUTXOSpendingInfoSingle(tx.inputs(inputIndex),
+                                  signer,
+                                  conditionalPath)
+
+    val txToSign = spendingInfo.output.scriptPubKey match {
+      case _: WitnessScriptPubKey =>
```
Christewart, line 281: why does this only apply to `WitnessScriptPubKey`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: It is safe because we are checking if it is defined first

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#version-numbers

If no version is given then it is 0

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: For all the `Option[x]` ones there can only be one of them in the map

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
benthecarman: We don't need the witness data for non `WitnessScriptPubKey`s

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: @nkohen I took a stab at this but couldn't figure it out

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: `.fingerprint` gives the fingerprint of the parent key, if it is the parent key it returns `0x00000000`, this is a check for if it is the parent key

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: yup, validation only done where possible, added this to the scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Because more context is needed as keys are reused per record type. (i.e. a key of 2 means something different for inputs and global)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Sadly no, using collect with `key.RecordType` causes a warning message because in fact either way this is done, there is type erasure and nothing is actually happening at run time.

PR #1035, opened 2020-01-16
PSBT stuff assigned to nkohen

nkohen:
Responded to psbt review assigned to me

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -1212,14 +1212,16 @@ sealed trait PSBTMap[+RecordType <: PSBTRecord] extends NetworkElement {
   require(elements.map(_.key).groupBy(identity).values.forall(_.length == 1),
           "All keys must be unique.")
 
-  final val separatorByte: Byte = 0x00.byteValue
-
   def elements: Vector[RecordType]
 
   def bytes: ByteVector =
     elements
       .sortBy(_.key)
-      .foldLeft(ByteVector.empty)(_ ++ _.bytes) :+ separatorByte
+      .foldLeft(ByteVector.empty)(_ ++ _.bytes) :+ PSBTMap.separatorByte
+}
+
+object PSBTMap {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -43,6 +41,9 @@ case class PSBT(
     outputMaps.size == transaction.outputs.size,
     "There must be an output map for every output in the global transaction")
 
+  import org.bitcoins.core.psbt.GlobalPSBTRecord._
+  import org.bitcoins.core.psbt.InputPSBTRecord._
```
benthecarman, line 44: Why move these?

nkohen: It is required because of new inter-dependencies which cause the following if these are left top-level:
```
[error] /home/nkohen/Desktop/SuredBits/bitcoin-s-core/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala:11:31: encountered unrecoverable cycle resolving import.
[error] Note: this is often due in part to a class depending on a definition nested within its companion.
```
essentially these must be moved to avoid circular dependencies

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1184,21 +1187,50 @@ object PSBTOutputKeyId {
...
+  def constructMap(elements: Vector[RecordType]): MapType
+
+  lazy val empty: MapType = constructMap(Vector.empty)
+
+  override def fromBytes(bytes: ByteVector): MapType = {
+    @tailrec
+    def loop(
+        remainingBytes: ByteVector,
+        accum: Vector[RecordType]): Vector[RecordType] = {
+      if (remainingBytes.head == 0x00.byteValue) {
```
benthecarman: > `0x00.byteValue`

use `PSBTMap.separatorByte`

benthecarman (changes requested):

benthecarman (approved):

PR #1037, opened 2020-01-16
PSBT independent record calls

benthecarman:


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1775,8 +1791,22 @@ object InputPSBTMap extends PSBTMapFactory[InputPSBTRecord, InputPSBTMap] {
...
+  def witnessScriptOpt: Option[WitnessScript] = {
+    getRecords(WitnessScriptKeyId).headOption
+  }
+
+  def BIP32DerivationPaths: Vector[BIP32DerivationPath] = {
+    getRecords(BIP32DerivationPathKeyId)
+  }
 
-  def getRecords(key: PSBTOutputKeyId): Vector[key.RecordType] = {
+  private def getRecords(key: PSBTOutputKeyId): Vector[key.RecordType] = {
```
nkohen, line 1813: I think that this def can likely move up to `PSBTMap` superclass

nkohen (commented):
Just one thing

nkohen (approved):

PR #1038, opened 2020-01-17
PSBT signer property based tests

benthecarman:


testkit/src/main/scala/org/bitcoins/testkit/core/gen/PSBTGenerators.scala
```diff
@@ -1,24 +1,199 @@
...
+
+  private def unknownGlobals: Gen[Vector[GlobalPSBTRecord.Unknown]] = {
+    Gen.choose(0, 4).flatMap(num => unknownGlobals(num))
+  }
+
+  private def unknownGlobals(
+      num: Int): Gen[Vector[GlobalPSBTRecord.Unknown]] = {
+    Gen
+      .sequence(0.until(num).map(_ => unknownGlobal))
+      .map(_.asScala.toVector)
```
nkohen: You should be able to use `Gen.listOfN`

testkit/src/main/scala/org/bitcoins/testkit/core/gen/PSBTGenerators.scala
```diff
@@ -1,24 +1,199 @@
...
+  }
+
+  private def unknownInput: Gen[InputPSBTRecord.Unknown] = {
```
nkohen, line 71: You should have one private def that generates a random record like this whose key matches nothing in all three realms (or in just the one if you pass in the factory), and then interpret those bytes differently in three one-liner defs rather than repeating

testkit/src/main/scala/org/bitcoins/testkit/core/gen/PSBTGenerators.scala
```diff
@@ -1,24 +1,199 @@
...
+
+  private def unknownInputs: Gen[Vector[InputPSBTRecord.Unknown]] = {
+    Gen.choose(0, 4).flatMap { num =>
+      unknownInputs(num)
+    }
+  }
+
+  private def unknownInputs(num: Int): Gen[Vector[InputPSBTRecord.Unknown]] = {
+    Gen
+      .sequence(0.until(num).map(_ => unknownInput))
```
nkohen: ditto above

testkit/src/main/scala/org/bitcoins/testkit/core/gen/PSBTGenerators.scala
```diff
@@ -1,24 +1,199 @@
...
+  private def unknownOutputs: Gen[Vector[OutputPSBTRecord.Unknown]] = {
+    Gen.choose(0, 4).flatMap { num =>
+      unknownOutputs(num)
+    }
+  }
+
+  private def unknownOutputs(
+      num: Int): Gen[Vector[OutputPSBTRecord.Unknown]] = {
+    Gen
+      .sequence(0.until(num).map(_ => unknownOutput))
```
nkohen: ditto above

testkit/src/main/scala/org/bitcoins/testkit/core/gen/PSBTGenerators.scala
```diff
@@ -1,24 +1,199 @@
...
+
+        PSBT(newGlobal, newInputMaps, newOutputMaps)
+      }
+    }
+  }
+
+  def psbtWithVersion(implicit ec: ExecutionContext): Gen[Future[PSBT]] = {
+    for {
+      psbtF <- psbtWithUnknowns
+      versionNumber <- Gen.choose(min = 0, max = 4294967295L)
```
nkohen: nit: give this max a descriptive name

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
```diff
@@ -220,6 +220,17 @@ class PSBTSerializerTest extends BitcoinSAsyncTest {
     succeed
   }
 
+  it must "fail to serialize PSBTs with unknown version numbers" in {
+    forAllAsync(PSBTGenerators.psbtWithVersion) { psbtF =>
```
nkohen: I think you should make a generator that specifically creates unknown versions for this test so that you don't need the if/else inside

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -233,6 +239,39 @@ class PSBTTest extends BitcoinSAsyncTest {
...
+                             test = Some(info)))
+                }
+              }
+          }
+          signedPSBTF.map { signedPSBT =>
+            val finalizedPsbtT = signedPSBT.finalizePSBT
+            finalizedPsbtT match {
+              case Success(finalizedPsbt) =>
+                assert(finalizedPsbt.extractTransactionAndValidate.isSuccess)
+              case Failure(exception) => throw exception
```
nkohen: use `fail` method instead which will wrap the exception in a nice and descriptive `TestFailedException` (which uses reflection and stuff)

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -276,8 +276,47 @@ object BitcoinSignerSingle {
                                   signer,
                                   conditionalPath)
 
+    val txToSign = spendingInfo.output.scriptPubKey match {
```
nkohen, line 279: Is there a reason that `P2WPKH` doesn't need to be accounted for here as well?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1652,9 +1652,15 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
     val scriptWitnessVec = getRecords(WitnessScriptKeyId)
     val scriptWitnessOpt =
-      if (scriptWitnessVec.size == 1)
+      if (scriptWitnessVec.size == 1) {
         Some(P2WSHWitnessV0(scriptWitnessVec.head.witnessScript))
-      else None
+      } else if (output.scriptPubKey
+                   .isInstanceOf[P2WPKHWitnessSPKV0] || redeemScriptOpt.exists(
+                   _.isInstanceOf[P2WPKHWitnessSPKV0])) {
+        Some(P2WPKHWitnessV0(signer.publicKey))
```
nkohen, line 1660: as this could be a common pattern, it might be nice to add a `scriptWitnessOpt: Option[ScriptWitnessV0]` to `InputPSBTMap` so that we don't have to go digging for `P2WSH` and `P2WPKH` witnesses

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
benthecarman: I am not sure, but it works if I make it only `P2WPKH`

nkohen: do you mean only `P2WSH`?

nkohen: maybe this has to do with the fact that what you sign in the `p2wpkh` case is the reconstructed `p2pkh` which doesn't involve witness stuff

PR #1039, opened 2020-01-17
PSBT Combiner property based tests

benthecarman:


core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -136,6 +136,48 @@ class PSBTTest extends BitcoinSAsyncTest {
     assert(psbt2.combinePSBT(psbt2) != expected)
   }
 
+  it must "correctly combine PSBTs" in {
+    forAllAsync(PSBTGenerators.psbtWithUnknowns) { psbtF =>
```
nkohen: why with Unknowns instead of using the arbitrary generator?

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -136,6 +136,48 @@ class PSBTTest extends BitcoinSAsyncTest {
...
+      psbtF.map { psbt =>
+        val global = psbt.globalMap.elements
+        val inputs = psbt.inputMaps
+        val outputs = psbt.outputMaps
+
+        val newGlobalElements = PSBTGenerators.pruneVec(global) :+ GlobalPSBTRecord
+          .UnsignedTransaction(psbt.transaction)
+        val newGlobal = GlobalPSBTMap(newGlobalElements.distinct)
```
nkohen: This seems like a common thing to do, we should add a util next to `pruneVec` called `def pruneGlobal(globalMap: GlobalPSBTMap): GlobalPSBTMap`

nkohen (commented):
Just some small things

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
nkohen: I think using the arbitrary generator will result in a more general test here

benthecarman: Since the arbitrary just prunes parts of `psbtWithUnknowns` figured it'd be pointless to prune twice

nkohen (approved):

PR #1040, opened 2020-01-17
Expand DLC Test Vectors

nkohen:
In response to review from CG/Tibo, I have de-serialized outputs and outpoints, re-serialized change spks as addresses, and added the oracle signature and R value to the test vectors


nkohen:
Here is the exact review I got:
```
Is there a strong reason for having output (within the funding utxos) be serialized? We could maybe add some function to deserialize it, but as it doesn’t seem like a very useful function in general I would rather have it deserialized in the test vector.


A couple of things that would make it easier on our side:

having the UTXO deserialized (at the moment I only use txid and vout plus the total input amount for each party as well as the total collateral for each party)

having the outpoint decomposed to txid and vout. this is easy to parse though so not a big deal.

Have addresses for change instead of ScriptPubKey, unless we actually need more information?

Have the ext key be in Base58

For the oracle, have the R point instead of the K value, and have the signature instead of the private key. At the moment I don’t handle the oracle signatures in the library and will need to wait until schnorr is in secp256k1 . I guess maybe you wanted to try the oracle part as well, but if it’s ok with you I’d rather delay that a bit.
```

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVector.scala
```diff
@@ -357,11 +378,27 @@ object SerializedDLCTestVectorSerializers {
       json.validate[String].map(factory.fromHex)
   }
 
-  implicit val transactionOutPointReads: Reads[TransactionOutPoint] = hexReads(
-    TransactionOutPoint)
-  implicit val transactionOutputReads: Reads[TransactionOutput] = hexReads(
-    TransactionOutput)
-  implicit val extPrivKeyReads: Reads[ExtPrivateKey] = hexReads(ExtPrivateKey)
+  implicit val ecPublicKeyReads: Reads[ECPublicKey] = hexReads(ECPublicKey)
```
Christewart, line 381: unrelated, but i think we should be moving serializers like these into `db-commons` (which should be renamed). Aren't these already implemented for the bitcoind-rpc project? https://github.com/bitcoin-s/bitcoin-s/blob/3a5af7e3ebcbd7dbe535c08b79ddda9a4a5612c0/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala#L35

Christewart: #1043

PR #1041, opened 2020-01-18
PSBT Input Updater property based tests

benthecarman:
Needed to add `addScriptWitnessToInput` because there was no way to do native segwit

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -230,16 +230,19 @@ case class PSBT(
             s"Cannot update an InputPSBTMap that is finalized, index: $index")
 
     val inputMap = inputMaps(index)
-    val redeemScriptOpt = inputMap.redeemScriptOpt.map(_.redeemScript)
 
-    val elements = if (isP2SHNestedSegwit(script, redeemScriptOpt)) {
+    val isWitScript = WitnessScriptPubKey.isWitnessScriptPubKey(script.asm)
+    val redeemScriptOpt = inputMap.redeemScriptOpt
+    val hasRedeemScript = redeemScriptOpt.isDefined && WitnessScriptPubKey
+      .isWitnessScriptPubKey(redeemScriptOpt.get.redeemScript.asm)
+
+    val elements = if (!isWitScript && hasRedeemScript) {
       inputMap.filterRecords(WitnessScriptKeyId) :+ InputPSBTRecord
         .WitnessScript(script.asInstanceOf[RawScriptPubKey])
     } else {
       inputMap.filterRecords(RedeemScriptKeyId) :+ InputPSBTRecord.RedeemScript(
         script)
     }
-
     val newMap = InputPSBTMap(elements).compressMap(transaction.inputs(index))
     val newInputMaps = inputMaps.updated(index, newMap)
 
@@ -274,6 +277,42 @@ case class PSBT(
         throw new IllegalArgumentException(
           s"Output script does not need a redeem script, got: $outputScript")
     }
+  }
+
+  def addScriptWitnessToInput(
+      scriptWitness: ScriptWitness,
+      index: Int): PSBT = {
+    require(
+      index < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val previousElements = inputMaps(index).elements
+      .filterNot(_.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte)
+
+    val newElement = scriptWitness match {
+      case p2wpkh: P2WPKHWitnessV0 =>
+        InputPSBTRecord.WitnessScript(P2PKHScriptPubKey(p2wpkh.pubKey))
+      case p2wsh: P2WSHWitnessV0 =>
+        InputPSBTRecord.WitnessScript(p2wsh.redeemScript)
+      case _ =>
+        throw new IllegalArgumentException(
+          s"Invalid scriptWitness given, got: $scriptWitness")
+    }
+
+    val compressedMap = InputPSBTMap(previousElements :+ newElement)
+      .compressMap(transaction.inputs(index))
+    val newMap = if (scriptWitness.isInstanceOf[P2WPKHWitnessV0]) {
+      // Compress to remove NonWitnessOrUnknownUTXO then remove P2WPKHWitnessV0
+      // because it is not necessary
+      InputPSBTMap(
+        compressedMap.elements.filterNot(
+          _.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte))
+    } else compressedMap
+
+    val newInputMaps = inputMaps.updated(index, newMap)
+    PSBT(globalMap, newInputMaps, outputMaps)
   }
 
   /**
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -282,14 +282,18 @@ case class PSBT(
   def addScriptWitnessToInput(
       scriptWitness: ScriptWitness,
       index: Int): PSBT = {
+    require(index >= 0,
+            s"index must be greater than or equal to 0, got: $index")
     require(
       index < inputMaps.size,
       s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
     require(!inputMaps(index).isFinalized,
             s"Cannot update an InputPSBTMap that is finalized, index: $index")
+    require(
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -242,6 +242,42 @@ case class PSBT(
...
+
+    val previousElements = inputMaps(index).elements
+      .filterNot(_.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte)
+
+    val newElement = scriptWitness match {
+      case p2wpkh: P2WPKHWitnessV0 =>
+        InputPSBTRecord.WitnessScript(P2PKHScriptPubKey(p2wpkh.pubKey))
+      case p2wsh: P2WSHWitnessV0 =>
+        InputPSBTRecord.WitnessScript(p2wsh.redeemScript)
+      case _ =>
```
Christewart: don't use `case _ =>`, it breaks pattern match exhaustiveness. You need to think of all the possible cases here and enumerate them. The compiler will even tell you what case you are missing: 

```
[W]  [E1] core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala:257:22
[W]       match may not be exhaustive.
[W]       It would fail on the following input: EmptyScriptWitness
[W]       L257:     val newElement = scriptWitness match {
[W]                                  ^
[W] core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala: L257 [E1]
Compiled core (3024ms)
Compiling core (30 Scala sources)
[W]  [E3] core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala:257:22
[W]       match may not be exhaustive.
[W]       It would fail on the following input: EmptyScriptWitness
[W]       L257:     val newElement = scriptWitness match {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -242,6 +242,42 @@ case class PSBT(
...
+        throw new IllegalArgumentException(
+          s"Invalid scriptWitness given, got: $scriptWitness")
+    }
+
+    val compressedMap = InputPSBTMap(previousElements :+ newElement)
+      .compressMap(transaction.inputs(index))
+    val newMap = if (scriptWitness.isInstanceOf[P2WPKHWitnessV0]) {
+      // Compress to remove NonWitnessOrUnknownUTXO then remove P2WPKHWitnessV0
```
Christewart: any reason this isn't done the pattern match above? It seems like this `if(scriptWitness.isInstanceOf[P2WPKHWitnessV0])` isn't necessary.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: I tried to do that originally, but I get this
```
[error] /home/ben/projects/bitcoin-s/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala:272:15: not found: type EmptyScriptWitness
[error]       case _: EmptyScriptWitness =>
[error]               ^
```

importing it does not fix it either

Christewart: It's because `EmptyScriptWitness` is a case object, rather than a `case class`. You need to do `case EmptyScriptWitness => ...`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -242,6 +242,42 @@ case class PSBT(
...
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val previousElements = inputMaps(index).elements
+      .filterNot(_.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte)
+
+    val newMap = scriptWitness match {
+      case p2wpkh: P2WPKHWitnessV0 =>
+        val newElement =
+          InputPSBTRecord.WitnessScript(P2PKHScriptPubKey(p2wpkh.pubKey))
```
nkohen: I was under the impression that no `WitnessScript` record was needed for `p2wpkh`, did you find something suggesting otherwise? If so other things will have to change as well as we derive it in different ways elsewhere.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -242,6 +242,42 @@ case class PSBT(
...
+      scriptWitness: ScriptWitness,
+      index: Int): PSBT = {
+    require(
+      index < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+    require(!inputMaps(index).isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val previousElements = inputMaps(index).elements
+      .filterNot(_.key.head == PSBTInputKeyId.WitnessScriptKeyId.byte)
```
nkohen: Should probably require that one is not already there

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Okay, I see below that it gets removed but don't understand why you add it just to remove it

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1695,23 +1732,34 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
             val out = nonWitUtxo.outputs(txIn.previousOutput.vout.toInt)
-            elements
-              .filterNot(_.key.head == NonWitnessUTXOKeyId.byte)
-              .filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
-              out)
+            out.scriptPubKey match {
+              case _: WitnessScriptPubKey =>
+                sortedElements
+                  .filterNot(_.key.head == NonWitnessUTXOKeyId.byte)
+                  .filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
```
nkohen: We should probably require that you can't have both `WitnessUTXO` and `NonWitnessUTXO` populated, then you can get rid of the `WitnessUTXO` filter here

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -1695,23 +1732,34 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+                  out)
+              case _: P2SHScriptPubKey =>
+                if (redeemScriptOpt.isDefined && redeemScriptOpt.get.redeemScript
+                      .isInstanceOf[WitnessScriptPubKey]) {
+                  sortedElements
+                    .filterNot(_.key.head == NonWitnessUTXOKeyId.byte)
+                    .filterNot(_.key.head == WitnessUTXOKeyId.byte) :+ WitnessUTXO(
+                    out)
```
nkohen: It seems that turning a `NonWitnessUTXO` into a `WitnessUTXO` should get its own method

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: So it knows that it is a witness utxo so it changes from a `NonWitnessOrUnknownUTXO` to a `WitnessUTXO`

nkohen: Nowhere in the code below for compressing the map does it look at the `WitnessScript` field, it exclusively looks at the `ScriptPubKey`. Still don't see why we add and remove this `newElement`

nkohen: Nowhere in the code below for compressing the map does it look at the `WitnessScript` field, it exclusively looks at the `ScriptPubKey`. Still don't see why we add and remove this `newElement`

benthecarman: My bad was thinking we used `WitnessScript` but we actually use `RedeemScript`

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -563,18 +587,17 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
     if (isFinalized) {
       this
     } else {
+      require(
+        this.witnessUTXOOpt.isEmpty || this.nonWitnessOrUnknownUTXOOpt.isEmpty)
```
nkohen: move this up to the top (still holds in the finalized case and should really be an invariant of InputPSBTMap), also error message

nkohen: What I meant was that what this actually ensures is that they aren't both populated and that this should go all the way up to the top of `InputPSBTMap`

nkohen (approved):

PR #1046, opened 2020-01-20
PSBTs response to Chris's review

benthecarman:


Christewart (approved):

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -173,28 +173,28 @@ case class PSBT(
...
-            out)
+          map.filterRecords(WitnessUTXOKeyId) :+ WitnessUTXO(out)
         } else {
-          previousElements.filterNot(_.key.head == NonWitnessUTXOKeyId.byte) :+ NonWitnessOrUnknownUTXO(
-            tx)
+          map.filterRecords(NonWitnessUTXOKeyId) :+ NonWitnessOrUnknownUTXO(tx)
         }
       } else {
-        previousElements
+        throw new IllegalArgumentException(
```
nkohen, line 197: nit: `IllegalStateException` right?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -257,21 +255,18 @@ case class PSBT(
       s"index must be less than the number of output maps present in the psbt, $index >= ${outputMaps.size}")
     require(!isFinalized, "Cannot update a PSBT that is finalized")
 
-    val previousElements = outputMaps(index).elements
+    val map = outputMaps(index)
```
nkohen: lets call this `outputMap` and the above `inputMap`, scala lets your overshadow names but it can lead to confusion when reading and changing code

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -296,26 +292,30 @@ case class PSBT(
...
 
-    val elements =
-      if (keyOpt.isSuccess && !previousElements.exists(_.key == expectedBytes)) {
-        val fp =
-          if (extKey.fingerprint == hex"00000000") {
-            extKey.deriveChildPubKey(path.path.head).get.fingerprint
+        val elements =
+          if (!previousElements.exists(_.key == expectedBytes)) {
+            val fp =
+              if (extKey.fingerprint == hex"00000000") {
```
nkohen: That `ByteVector` probably deserves a `val` in `ExtKey.scala` someplace

nkohen (commented):
just a couple nits

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: I think `IllegalArgumentException` is correct here because if we reach here it means that the transaction given as an argument had less outputs than it would need to be a valid previous transaction for this input

PR #1048, opened 2020-01-20
PSBT Docs

### LINKED ISSUE #1016: Add website psbt.md documenting/showcasing how psbts work

benthecarman:
I added the small function `PSBT.base64` because it was used in the docs

Closes #1016

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -54,6 +54,8 @@ case class PSBT(
     inputBytes ++
     outputBytes
 
+  val base64: String = bytes.toBase64
+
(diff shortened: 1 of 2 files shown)


nkohen (commented):
I think it would be nice to cut down on the amount of nesting and use more intermediate vals, especially with the updates where you construct the inputs to the updates inside as params.

Also, and this is up for debate, once all the inputs are vals, I think it could be cool to show off how the interface is "fluent" meaning you can call
```
psbt
    .updateSomething(input)
    .updateSomethingElse(otherInput)
```

benthecarman:
> Also, and this is up for debate, once all the inputs are vals, I think it could be cool to show off how the interface is "fluent" meaning you can call

I think this could be a good idea but might just end up confusing a user.

Christewart (approved):

PR #1049, opened 2020-01-21
Psbt changes nkohen

nkohen:
Built on top of #1046 to avoid merge conflicts

Responds to review on #1031 assigned to nkohen.

Biggest change here is that PSBT.scala gets split up into multiple files.

First commit responds to most review and second commit is move-only.


benthecarman (approved):
lgtm

Christewart (approved):

PR #1052, opened 2020-01-22
Txo state flyway

### LINKED ISSUE #1045: TXO State migration
In #1001 we added a more expressive field in the spending info table to indicate what the state of a txo is. With #1044 we now have the ability to properly manage this migration. We need to properly migrate the existing fields of `true` and `false` to be one of the new `TxoState`. Probably we want to map `false` -> `PendingConfirmationsReceived` and `true` -> `PendingConfirmationsSpent`. 

This needs to be addressed before all test cases will pass on #1044.

benthecarman:
Not entirely sure if I did this correctly, but tests are passing AFAIK.
Diff is just the `f0f231c` commit

Logic is that the previous booleans (which are stored as integers in SQLite), will be read in as VARCHAR and then we can convert as needed.

Closes #1045


wallet/src/main/resources/walletdb/migration/V1__wallet_db_baseline.sql
```diff
@@ -1,5 +1,8 @@
-CREATE TABLE IF NOT EXISTS "txo_spending_info" ("id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,"tx_outpoint" VARCHAR(254) NOT NULL,"script_pub_key" VARCHAR(254) NOT NULL,"value" INTEGER NOT NULL,"hd_privkey_path" VARCHAR(254) NOT NULL,"redeem_script" VARCHAR(254),"script_witness" VARCHAR(254),"confirmations" INTEGER NOT NULL,"spent" INTEGER NOT NULL,"txid" VARCHAR(254) NOT NULL,constraint "fk_scriptPubKey" foreign key("script_pub_key") references "addresses"("script_pub_key") on update NO ACTION on delete NO ACTION);
+CREATE TABLE IF NOT EXISTS "txo_spending_info" ("id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,"tx_outpoint" VARCHAR(254) NOT NULL, "txo_state" VARCHAR(254) NOT NULL, "script_pub_key" VARCHAR(254) NOT NULL,"value" INTEGER NOT NULL,"hd_privkey_path" VARCHAR(254) NOT NULL,"redeem_script" VARCHAR(254),"script_witness" VARCHAR(254),"confirmations" INTEGER NOT NULL,"spent" INTEGER NOT NULL,"txid" VARCHAR(254) NOT NULL,constraint "fk_scriptPubKey" foreign key("script_pub_key") references "addresses"("script_pub_key") on delete NO ACTION);
+
+UPDATE "txo_spending_info" SET "txo_state" = "PendingConfirmationsSpent" WHERE "txo_state" = 1;
```
Christewart: so the sql itself looks right, but what you generally don't want to modify existing migrations. You want to create a second migration called `V2__nice_name.sql` and then put the sql that needs to be run in there. 

Basically if modify existing migrations, it breaks the checksum system flyway uses to determine if a migration has already been applied. https://flywaydb.org/documentation/migrations#naming-1

Christewart: Also another thing I think you need to do is alter the table to drop the old "spent" column and add the new "txo_state" column. Make sure the schema is the same as what is defined in `SpendingInfoTable`

PR #1054, opened 2020-01-22
PSBT unit tests by Ben

benthecarman:
Removed `toUTXOSpendingInfo`, `getUTXOSpendingInfo` and the test that only checked if we didn't give a negative index to those functions.

These functions weren't used and didn't have much use as they returned spend infos for already signed inputs

All tests pass locally

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
@@ -3,10 +3,15 @@ package org.bitcoins.core.psbt
 import org.bitcoins.core.crypto.{ECPublicKey, ExtPublicKey}
 import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.hd.BIP32Path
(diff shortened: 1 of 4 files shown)


core-test/src/test/scala/org/bitcoins/core/psbt/PSBTSerializerTest.scala
```diff
@@ -156,6 +169,23 @@ class PSBTSerializerTest extends BitcoinSAsyncTest {
     assert(
       psbtWithPoRC.inputMaps.head.proofOfReservesCommitmentOpt
         .contains(proofOfReservesCommitment))
+    assert(psbtWithPoRC == PSBT(psbtWithPoRC.bytes))
+  }
+
+  it must "successfully serialize a PSBT with a Output Witness Script" in {
+    val psbt = PSBT(validPsbts.head)
+
+    val witScript = OutputPSBTRecord.WitnessScript(ScriptPubKey.empty)
```
nkohen, line 178: nit: Use `EmptyScriptPubKey`

nkohen (commented):
Just one nit

PR #1055, opened 2020-01-22
Psbt unit tests by Nadav

nkohen:
Added a bunch of tests, caught a log bug, removed an old method whose use has been replaced.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
@@ -1,16 +1,11 @@
 package org.bitcoins.core.psbt
 
 import org.bitcoins.core.crypto._
-import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.currency.{CurrencyUnits, Satoshis}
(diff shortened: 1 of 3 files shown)


core-test/src/test/scala/org/bitcoins/core/psbt/PSBTTest.scala
```diff
@@ -476,4 +491,160 @@ class PSBTTest extends BitcoinSAsyncTest {
...
+    assert(compressedInputMap == expectedInputMap)
+  }
+
+  it must "do nothing when compressing a finalized InputPSBTMap" in {
+    val finalizedInputMap = InputPSBTMap(
+      Vector(InputPSBTRecord.FinalizedScriptSig(EmptyScriptSignature)))
+
+    val dummyInput = dummyTx().inputs.head
+
+    val compressedInputMap = finalizedInputMap.compressMap(dummyInput)
```
benthecarman: Should just be able to use `EmptyTransactionInput` instead of all this `dummyTx` stuff

benthecarman (approved):
LGTM, could make that one change but doesn't really matter

PR #1058, opened 2020-01-23
If migrations fail, attempt to baseline the database and apply migrat…

Christewart:
…ions again

This is for databases that were created pre-flyway. We need to [baseline](https://flywaydb.org/documentation/command/baseline) the existing database and then apply migrations so flyway works.


benthecarman:
Is there a way to add a test for this?

benthecarman (approved):
If there no way to add a test, LGTM

PR #1073, opened 2020-01-27
Split DLC into two clients that must communicate

nkohen:
Replaced `BinaryOutcomeDLCWithSelf` with two `BinaryOutcomeDLCClient`s and wrote tests for parity.


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+    oraclePubKey: ECPublicKey,
+    preCommittedR: ECPublicKey,
+    isInitiator: Boolean,
+    extPrivKey: ExtPrivateKey,
+    remoteExtPubKey: ExtPublicKey,
+    input: CurrencyUnit,
+    remoteInput: CurrencyUnit,
+    fundingUtxos: Vector[BitcoinUTXOSpendingInfoSingle],
+    remoteFundingInputs: Vector[
+      (DoubleSha256DigestBE, TransactionOutput, UInt32)],
```
Christewart: Why not `(TransactionOutPoint, TransactionOutput)`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+
+  private val cetFee: CurrencyUnit = Satoshis(approxCETVBytes * feeRate.toLong)
+  private val toLocalClosingFee: CurrencyUnit = Satoshis(
+    approxToLocalClosingVBytes * feeRate.toLong)
+
+  private val isRBFEnabled = Policy.isRBFEnabled
+  private val sequence =
+    if (isRBFEnabled) UInt32.zero else TransactionConstants.sequence
+
+  lazy val createUnsignedFundingTransaction: Transaction = {
```
nkohen, line 226: important change: split funding tx construction into steps, unsigned, signed by one party, signed by both

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+     *
+     * Once computed, we add the estimated amount to the fundingOutput so it can be used for fees later.
+     */
+    val halfCetFee = Satoshis((cetFee + toLocalClosingFee).satoshis.toLong / 2)
+
+    val output: TransactionOutput =
+      TransactionOutput(totalInput + halfCetFee + halfCetFee,
+                        P2WSHWitnessSPKV0(fundingSPK))
+
+    val (initiatorChange, initiatorChangeSPK, otherChange, otherChangeSPK) =
```
nkohen, line 238: important change: order of outputs now matters as both parties must generate the same txs

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+      sequence)
+
+    val psbt = PSBT.fromUnsignedTx(
+      BaseTransaction(TransactionConstants.validLockVersion,
+                      Vector(fundingInput),
+                      outputs,
+                      timeouts.contractMaturity.toUInt32)
+    )
+
+    val readyToSignPSBT = psbt
```
nkohen, line 385: important change: one signature is given as input to local CETs so PSBTs are used

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+                                     Vector(fundingInput),
+                                     outputs,
+                                     timeouts.contractMaturity.toUInt32)
+
+    val sigF = PSBT
+      .fromUnsignedTx(unsignedTx)
+      .addUTXOToInput(fundingTx, index = 0)
+      .addScriptWitnessToInput(P2WSHWitnessV0(fundingSPK), index = 0)
+      .sign(inputIndex = 0, fundingPrivKey)
+      .map(_.inputMaps.head.partialSignatures.head)
```
nkohen, line 449: Important change: signature is constructed rather than signed remote CET

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+
+    sigF.map((unsignedTx, _, P2WSHWitnessV0(toLocalSPK)))
+  }
+
+  lazy val createUnsignedRefundTx: Transaction = {
```
nkohen, line 454: important change: like funding tx, refund tx is split into stages, and order of outputs matters

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+      sendSigs: (
+          PartialSignature,
+          PartialSignature,
+          PartialSignature) => Future[Unit],
+      getSigs: Future[
+        (
+            PartialSignature,
+            PartialSignature,
+            PartialSignature,
+            Vector[PartialSignature])]): Future[SetupDLC] = {
```
nkohen: Important change: there are now two different setup methods for the two different parties, and they take interaction lambdas and futures as inputs

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+    spendingTxF.foreach(tx => logger.info(s"Closing Tx: ${tx.hex}"))
+
+    spendingTxF
+  }
+
+  /** Constructs and executes on the unilateral spending branch of a DLC
+    *
+    * @return Each transaction published and its spending info
+    */
+  def executeUnilateralDLC(
```
nkohen, line 720: important change: execution of unilateral DLC is not split into local and remote methods, one for each output. (note that the remote case is used in justice case as well)

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+        cetSpendingInfo = justiceSpendingInfo
+      )
+    }
+  }
+
+  /** Constructs and executes on the refund spending branch of a DLC
+    *
+    * @return Each transaction published and its spending info
+    */
+  def executeRefundDLC(dlcSetup: SetupDLC): Future[DLCOutcome] = {
```
nkohen, line 880: important change, both parties must now call this method separately to get each of their closing txs

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+    }
+  }
+
+  /** Subtracts the estimated fee by removing from each output with a specified spk evenly */
+  def subtractFeeFromOutputs(
+      tx: BaseTransaction,
+      feeRate: FeeUnit,
+      spks: Vector[ScriptPubKey]): BaseTransaction = {
+    // tx has empty script sigs and we need to account for witness data in fees
+    val dummyTxWit = TransactionWitness(
```
nkohen, line 993: important change: fee estimation is done by filling a `TransactionWitness` with `DummyECDigitalSignature`s rather than using `BitcoinTxBuilder.loop(dummy = true)`. It seems that this creates exactly the same fees for funding tx but for some reason I don't understand, the fees on the refund tx (spends the funding tx only) have been reduced by a couple satoshis

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVector.scala
```diff
@@ -141,7 +124,18 @@ case class DLCTestVector(
 
   /** Tests that regenerating from inputs yields same outputs */
   def test(): Future[Boolean] = {
-    regenerate.map(_ == this)
+    regenerate.map { regenerated =>
+      if (regenerated != this) {
+        println(s"BEFORE: $this")
```
nkohen: ping: oops, get rid of this

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVector.scala
```diff
@@ -165,46 +159,115 @@ object DLCTestVector {
...
+      changeSPK = remoteChangeSPK,
+      remoteChangeSPK = localChangeSPK,
+      network = RegTest
+    )
+
     val outcomeHash = CryptoUtil.sha256(ByteVector(realOutcome.getBytes)).flip
     val oracleSig =
       Schnorr.signWithNonce(outcomeHash.bytes, oracleKey, oracleKValue)
 
+    val offerSigReceiveP =
```
nkohen, line 221: important change: this is how I am mocking out interaction in tests across the board

nkohen (commented):
Marked important to help focus review

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+    val fundingInput = TransactionInput(
+      TransactionOutPoint(fundingTxid, UInt32.zero),
+      EmptyScriptSignature,
+      sequence)
+    val fundingOutput = fundingTx.outputs.head
+
+    val (initiatorValue, initiatorKey, otherValue, otherKey) =
+      if (isInitiator) {
+        val toLocalValue = Satoshis(
+          (fundingOutput.value * input).satoshis.toLong / totalInput.satoshis.toLong)
```
Christewart, line 466: division always makes me nervous when handling money, I don't see anything wrong here but just something to be wary of. Probably want to make sure that this is tested well

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+  def createCETLoseRemote(): Future[
+    (Transaction, PartialSignature, P2WSHWitnessV0)] = {
+    createCETRemote(
+      sigPubKey = sigPubKeyLose,
+      payout = losePayout,
+      remotePayout = remoteLosePayout
+    )
+  }
+
+  def setupDLCAccept(
```
Christewart, line 585: scaladoc with parameter descriptions

Also link to the relevant dlcspec documentation

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+        cetWinRemote.txIdBE,
+        cetWinRemoteWitness,
+        cetLoseRemote.txIdBE,
+        cetLoseRemoteWitness,
+        refundTx
+      )
+    }
+  }
+
+  def setupDLCOffer(
```
Christewart, line 630: Scaladoc, and link to the relevant dlspec documentation

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+      val sigForLose = Schnorr.verify(messageLose, oracleSig, oraclePubKey)
+
+      // Pick the CET to use and payout by checking which message was signed
+      val (cet, extCetPrivKey, cetScriptWitness) =
+        if (sigForWin) {
+          (cetWinLocal, cetWinPrivKey, cetWinLocalWitness)
+        } else if (sigForLose) {
+          (cetLoseLocal, cetLosePrivKey, cetLoseLocalWitness)
+        } else {
+          throw new IllegalStateException(
```
Christewart, line 745: Do we want to do something like broadcast the refund transaction in this case?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
Christewart: Link to the dlcspec defining what the unlilateral dlc execution is

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
Christewart: Link to the spot in dlc specs where this is defined

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -649,6 +483,177 @@ object TxBuilder {
...
+    */
+  def calcLockTime(utxos: Seq[BitcoinUTXOSpendingInfoFull]): Try[UInt32] = {
+    def computeNextLockTime(
+        currentLockTimeOpt: Option[UInt32],
+        locktime: Long): Try[UInt32] = {
+      val lockTime =
+        if (locktime > UInt32.max.toLong || locktime < 0) {
+          TxBuilderError.IncompatibleLockTimes
+        } else Success(UInt32(locktime))
+      lockTime.flatMap { l: UInt32 =>
```
benthecarman: Use a more descriptive name than `l`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+    }
+
+    val signedFundingPSBTF =
+      fundingUtxos.zipWithIndex.foldLeft(Future.successful(fundingPSBT)) {
+        case (psbtF, (utxo, index)) =>
+          psbtF.flatMap { psbt =>
+            psbt
+              .addWitnessUTXOToInput(output = utxo.output, index + localTweak)
+              .addScriptWitnessToInput(
+                scriptWitness = utxo.scriptWitnessOpt.get,
```
benthecarman, line 320: Is this `.get` safe?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+                index + localTweak)
+              .sign(index + localTweak, utxo.signers.head)
```
benthecarman: Can use `utxo.signer` instead

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+
+    (signedCETF, P2WSHWitnessV0(toLocalSPK))
+  }
+
+  /** Constructs Remote's CET given sig*G, the funding tx's UTXOSpendingInfo and payouts */
+  def createCETRemote(
+      sigPubKey: ECPublicKey,
+      payout: CurrencyUnit,
+      remotePayout: CurrencyUnit): Future[
+    (Transaction, PartialSignature, P2WSHWitnessV0)] = {
```
benthecarman: nit: I feel like it makes more sense to order this as `(Transaction, P2WSHWitnessV0, PartialSignature)`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -0,0 +1,1056 @@
...
+      .addScriptWitnessToInput(P2WSHWitnessV0(fundingSPK), index = 0)
+      .sign(inputIndex = 0, fundingPrivKey)
+      .map(_.inputMaps.head.partialSignatures.head)
+
+    sigF.map((unsignedTx, _, P2WSHWitnessV0(toLocalSPK)))
+  }
+
+  lazy val createUnsignedRefundTx: Transaction = {
+    val fundingTx = createUnsignedFundingTransaction
+    val fundingTxid = fundingTx.txId
```
benthecarman, line 456: How come above making the `fundingTxid` uses `.txIdBE` and here you use `.txId`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
benthecarman: Can't these be `Vector`s and just put a `require` for the length

benthecarman (changes requested):

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: Agreed, and this (as well as fee calculations) will make it to the spec eventually. At this point I'd think of it as nearly right but probably still a placeholder

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: Not necessarily, under the current client design (which will change) I would prefer a failure followed later, if wanted, by a call to `executeRefundDLC`

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: So `TransactionOutPoint` has constructors for both the `DoubleSha256Digest` and `DoubleSha256DigestBE` types and handles the flipping in the `apply`, since this val is only used in an outpoint and not returned or passed anywhere, I figured it didn't matter and did the shorter one cause I'm lazy

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: That would be one thing I could do but I imagine in the future (once this part of the code becomes more stable when I've written interaction code) this will all end up in a case class which is why I have it elaborated rn.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: yes since all outputs in a dlc (including funding) must be segwit (because of malleability)

Christewart (approved):

PR #1074, opened 2020-01-28
CLI Commands for PSBTs

benthecarman:


app/cli/src/main/scala/org/bitcoins/cli/Cli.scala
```diff
@@ -157,6 +166,76 @@ object Cli extends App {
...
+            .action((seq, conf) =>
+              conf.copy(command = conf.command match {
+                case joinPSBTs: JoinPSBTs =>
+                  joinPSBTs.copy(psbts = seq)
+                case other => other
+              }))
+        ),
+      cmd("finalizepsbt")
+        .hidden()
+        .action((_, conf) => conf.copy(command = FinalizePSBT(null)))
```
Christewart: Do we need to define an `.action` on this line and on line 204? I would like to get rid of this one if possible as we are using `null` here and `null` is evil. 

I realize we are doing this in other places too -- can we remove all of the outer `.action()` definitions?

https://github.com/scopt/scopt

from a quick glance at the README it looks like they can avoid using `null`. Is it because they provide default values in the `Config` object? Can we create a `PSBT.empty` or `EmptyTransaction` to use as default values rather than `null`?

PR #1075, opened 2020-01-28
Attempt to keep GCSTest for timing out on CI

Christewart:
#845 does not fix the timeout problem, another attempt at fixes the timeout, specifically on scala 2.12.x


nkohen (approved):
Still skeptical that this will fix everything but certainly can only be an improvement if anything? LGTM

PR #1078, opened 2020-01-29
Created methods for constructing DLCClients from DLCMessages

nkohen:


dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -914,6 +913,125 @@ case class BinaryOutcomeDLCClient(
...
+      preCommittedR: ECPublicKey,
+      isInitiator: Boolean,
+      extPrivKey: ExtPrivateKey,
+      remoteExtPubKey: ExtPublicKey,
+      input: CurrencyUnit,
+      remoteInput: CurrencyUnit,
+      fundingUtxos: Vector[BitcoinUTXOSpendingInfoSingle],
+      remoteFundingInputs: Vector[(TransactionOutPoint, TransactionOutput)],
+      winPayout: CurrencyUnit,
+      losePayout: CurrencyUnit,
```
benthecarman, line 929: `winPayout` and `losePayout` can be removed because they are in the `contractInfo`

Should be removed from the `BinaryOutcomeDLCClient` constructor as well

PR #1080, opened 2020-01-29
Added Mutual Closing DLC case

nkohen:
Added functions for mutual closing of DLCs, with passing unit and integration tests

### DIFF
--- a/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
+++ b/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
@@ -349,7 +349,7 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
     } yield (acceptDLC, acceptSetup, offerDLC, offerSetup)
   }
 
-  def executeMutualForCase(
+  def executeForMutualCase(
       outcomeHash: Sha256DigestBE,
       local: Boolean): Future[Assertion] = {
     val oracleSig =
@@ -568,15 +568,15 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
 
   it should "be able to publish all DLC txs to Regtest for the mutual Win case" in {
     for {
-      _ <- executeMutualForCase(outcomeWinHash, local = true)
-      _ <- executeMutualForCase(outcomeWinHash, local = false)
+      _ <- executeForMutualCase(outcomeWinHash, local = true)
+      _ <- executeForMutualCase(outcomeWinHash, local = false)
     } yield succeed
   }
 
   it should "be able to publish all DLC txs to Regtest for the mutual Lose case" in {
     for {
-      _ <- executeMutualForCase(outcomeLoseHash, local = true)
-      _ <- executeMutualForCase(outcomeLoseHash, local = false)
+      _ <- executeForMutualCase(outcomeLoseHash, local = true)
+      _ <- executeForMutualCase(outcomeLoseHash, local = false)
     } yield succeed
   }
 
--- a/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
+++ b/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
@@ -379,19 +379,21 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
         val watchForMutualCloseTx = new Runnable {
           override def run(): Unit = {
             if (!mutualCloseTxP.isCompleted) {
+              val fundingTxId = initDLC
+                .createUnsignedMutualClosePSBT(oracleSig, initSetup.fundingTx)
+                .transaction
+                .txIdBE
+
               clientF.foreach { client =>
-                val fundingTxResultF = client.getRawTransaction(
-                  initDLC
-                    .createUnsignedMutualClosePSBT(oracleSig,
-                                                   initSetup.fundingTx)
-                    .transaction
-                    .txIdBE)
+                val fundingTxResultF = client.getRawTransaction(fundingTxId)
 
                 fundingTxResultF.onComplete {
                   case Success(fundingTxResult) =>
                     if (fundingTxResult.confirmations.isEmpty) {
                       ()
                     } else {
+                      logger.info(
+                        s"Found funding tx on chain! $fundingTxResult")
                       mutualCloseTxP.trySuccess(fundingTxResult.hex)
                     }
                   case Failure(_) => ()
--- a/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
+++ b/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
@@ -8,6 +8,7 @@ import org.bitcoins.core.crypto.{
   ECPublicKey,
   ExtPrivateKey,
   Schnorr,
+  SchnorrDigitalSignature,
   SchnorrNonce,
   Sha256DigestBE
 }
@@ -26,6 +27,7 @@ import org.bitcoins.core.protocol.script.{
   P2WPKHWitnessV0
 }
 import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutPoint}
+import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import org.bitcoins.core.script.crypto.HashType
 import org.bitcoins.core.util.{CryptoUtil, FutureUtil}
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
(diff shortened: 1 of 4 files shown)


dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
     } yield (acceptDLC, acceptSetup, offerDLC, offerSetup)
   }
 
+  def executeMutualForCase(
```
benthecarman: > executeMutualForCase

executeForMutualCase

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientTest.scala
```diff
@@ -295,6 +297,52 @@ class BinaryOutcomeDLCClientTest extends BitcoinSAsyncTest {
     }
   }
 
+  def executeMutualForCase(
```
benthecarman, line 300: same as above

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -330,6 +330,70 @@ case class BinaryOutcomeDLCClient(
...
+  def createUnsignedMutualClosePSBT(
+      sig: SchnorrDigitalSignature,
+      fundingTx: Transaction): PSBT = {
+    val (toLocalPayout, toRemotePayout) =
+      if (Schnorr.verify(messageWin, sig, oraclePubKey)) {
+        (winPayout, remoteWinPayout)
+      } else if (Schnorr.verify(messageLose, sig, oraclePubKey)) {
+        (losePayout, remoteLosePayout)
+      } else {
+        throw new IllegalStateException(
```
benthecarman: I feel this should be `IllegalArugmentExeception` because the `sig` is an argument

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -330,6 +330,70 @@ case class BinaryOutcomeDLCClient(
...
+          "Signature does not correspond to either possible outcome!")
```
benthecarman: nit: include `sig` in the message

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -330,6 +330,70 @@ case class BinaryOutcomeDLCClient(
...
+  }
+
+  def createMutualCloseTxSig(
+      sig: SchnorrDigitalSignature,
+      fundingTx: Transaction): Future[PartialSignature] = {
+    val unsignedPSBT = createUnsignedMutualClosePSBT(sig, fundingTx)
+
+    unsignedPSBT
+      .sign(inputIndex = 0, fundingPrivKey)
+      .map(_.inputMaps.head.partialSignatures.head)
```
benthecarman: This won't work if there is multiple sigs for the input, you should do,

`.map(_.inputMaps.head.partialSignatures.find(_.pubKey == fundingPrivKey.pubKey))`

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+                    .txIdBE)
+
+                fundingTxResultF.onComplete {
+                  case Success(fundingTxResult) =>
+                    if (fundingTxResult.confirmations.isEmpty) {
+                      ()
+                    } else {
+                      mutualCloseTxP.trySuccess(fundingTxResult.hex)
+                    }
+                  case Failure(_) => ()
```
benthecarman, line 399: Error message here?

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+                  initDLC
+                    .createUnsignedMutualClosePSBT(oracleSig,
+                                                   initSetup.fundingTx)
+                    .transaction
+                    .txIdBE)
+
+                fundingTxResultF.onComplete {
+                  case Success(fundingTxResult) =>
+                    if (fundingTxResult.confirmations.isEmpty) {
+                      ()
```
benthecarman, line 393: Error message here?

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
nkohen: No need to fail, if nothing is found then nothing is done and another call is made after `1.second`

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
nkohen: No need to fail, if no confirmations then nothing is done and another call is made after `1.second`

benthecarman (approved):
LGTM

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
     } yield (acceptDLC, acceptSetup, offerDLC, offerSetup)
   }
 
+  def executeForMutualCase(
```
Christewart, line 352: Is there a document for this case in the DLC spec repo? If so link in the scaladoc

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+        }
+
+        val mutualCloseTxP = Promise[Transaction]()
+
+        val watchForMutualCloseTx = new Runnable {
+          override def run(): Unit = {
+            if (!mutualCloseTxP.isCompleted) {
+              clientF.foreach { client =>
+                val fundingTxResultF = client.getRawTransaction(
+                  initDLC
```
Christewart: can we use an intermediate val for calculating the closing transaction's txid?

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+                    .createUnsignedMutualClosePSBT(oracleSig,
+                                                   initSetup.fundingTx)
+                    .transaction
+                    .txIdBE)
+
+                fundingTxResultF.onComplete {
+                  case Success(fundingTxResult) =>
+                    if (fundingTxResult.confirmations.isEmpty) {
```
Christewart, line 392: At some point we will have to come back and adjust this I think to this when we implement it: https://github.com/bitcoin-s/bitcoin-s/issues/1083

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -347,6 +349,98 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+                      ()
+                    } else {
```
Christewart, line 394: logging in the case where we actually find the transaction would be nice.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -712,6 +785,36 @@ case class BinaryOutcomeDLCClient(
     spendingTxF
   }
 
+  def initiateMutualClose(
+      dlcSetup: SetupDLC,
+      sig: SchnorrDigitalSignature,
+      sendSigs: (SchnorrDigitalSignature, PartialSignature) => Future[Unit],
+      getMutualCloseTx: Future[Transaction]): Future[CooperativeDLCOutcome] = {
+    val fundingTx = dlcSetup.fundingTx
```
Christewart, line 807: log that we are beginning to attempt a mutual close of dlc, with the txids at minimum so we can follow this.

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -330,6 +330,73 @@ case class BinaryOutcomeDLCClient(
     }
   }
 
+  def createUnsignedMutualClosePSBT(
```
Christewart, line 338: link to dlc spec and scaladoc

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -330,6 +330,73 @@ case class BinaryOutcomeDLCClient(
...
+      sig: SchnorrDigitalSignature,
+      fundingTx: Transaction): Future[PartialSignature] = {
+    val unsignedPSBT = createUnsignedMutualClosePSBT(sig, fundingTx)
+
+    unsignedPSBT
+      .sign(inputIndex = 0, fundingPrivKey)
+      .map(
+        _.inputMaps.head.partialSignatures
+          .find(_.pubKey == fundingPrivKey.publicKey)
+          .get)
```
Christewart: rather than using `.get here it would be nice to just pattern match and then give a better error message in the `case None => ` case. 

I.e. 
```scala
case None => sys.error(s"You have constructed the psbt incorrectly! We could not find a public key=${fundingPrivKey.publicKey}")
```

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -446,7 +513,10 @@ case class BinaryOutcomeDLCClient(
       .addUTXOToInput(fundingTx, index = 0)
       .addScriptWitnessToInput(P2WSHWitnessV0(fundingSPK), index = 0)
       .sign(inputIndex = 0, fundingPrivKey)
-      .map(_.inputMaps.head.partialSignatures.head)
+      .map(
+        _.inputMaps.head.partialSignatures
+          .find(_.pubKey == fundingPrivKey.publicKey)
+          .get)
```
Christewart: same comment as above, if this pattern is re-used in more places maybe just factor it out into a helper method?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -712,6 +785,36 @@ case class BinaryOutcomeDLCClient(
     spendingTxF
   }
 
+  def initiateMutualClose(
```
Christewart, line 802: scaladoc with link to dlcspec

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -712,6 +785,36 @@ case class BinaryOutcomeDLCClient(
...
+    for {
+      fundingSig <- createMutualCloseTxSig(sig, fundingTx)
+      _ <- sendSigs(sig, fundingSig)
+      mutualCloseTx <- getMutualCloseTx
+    } yield {
+      CooperativeDLCOutcome(fundingTx, mutualCloseTx)
+    }
+  }
+
+  def executeMutualClose(
```
Christewart, line 825: scaladoc with link to dlcspec

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
nkohen: sure, but this is just for integration tests, wallet code probably won't even use this code

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
nkohen: This isn't source code so not sure it is necessary but will do

Christewart (approved):

Issue #1061, opened 2020-01-23
Logging breaks if you make cli requests to two separate sub projects


Christewart:
**Describe the bug**
I was syncing my neutrino node this morning having logs come in just fine to my `$HOME/testnet3/bitcoin-s.log` file, however when i made another rpc call during the sync, (bitcoin-s-cli getnewaddress), it appears that all of the logging that was happening in the chain project related to my IBD broke. I could not get them to restart. 

**To Reproduce**
Watch your `.bitcoin-s/testnet3/bitcoin-s.log` file
Start syncing your node
Make a rpc call to another project (such as `getnewaddress`)
Observe how other projects logs seem to break.

**Expected behavior**
I expected logging to work for all projects even if rpc calls are made that are related to other projects.

This is possibly related to #807

nkohen:
Called `getfilterheadercount`, then `getnewaddress` then `getfilterheadercount` again and got these logs:
```
2020-01-29T21:29:03 DEBUG [chain-verification] Querying for filter header count
2020-01-29T21:29:03 DEBUG [chain-verification] getFilterHeaderCount result: count=1664264
2020-01-29T21:30:45 DEBUG [WALLET] Getting new External adddress for m/84'/1'/0'
2020-01-29T21:30:45 DEBUG [WALLET] Did not find previous address, next=m/84'/1'/0'/0/0
2020-01-29T21:30:45 DEBUG [WALLET] Writing SegWitAddressDb(m/84'/1'/0'/0/0,ECPublicKey(030dcf5200b5a18a36c8f3d5b03050b580175ad0dd37fc9cd5715c6959d8182dab),Sha256Hash160DigestImpl(3acd7e5bf0d994670d06c5b9e83ec2cef0aff3de),Bech32Address(tb1q8txhuklsmx2xwrgxcku7s0kzemc2lu779n45jr),P2WPKHWitnessV0(List(030dcf5200b5a18a36c8f3d5b03050b580175ad0dd37fc9cd5715c6959d8182dab, )),wpkh(Sha256Hash160DigestImpl(3acd7e5bf0d994670d06c5b9e83ec2cef0aff3de))) to DB
[...]

PR #1086, opened 2020-01-31
Multiple Logger Fix

nkohen:
An attempt to fix logging to file using MDC.

See #1061

See http://logback.qos.ch/manual/mdc.html

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.db
 
-import org.slf4j.Logger
-import org.slf4j.LoggerFactory
+import ch.qos.logback.classic
+import org.slf4j.{LoggerFactory, MDC}
 import ch.qos.logback.classic.spi.ILoggingEvent
 import ch.qos.logback.classic.LoggerContext
 import ch.qos.logback.classic.encoder.PatternLayoutEncoder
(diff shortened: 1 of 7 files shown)


db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
```diff
@@ -83,14 +87,33 @@ private[bitcoins] trait AppLoggers {
 
     logFileAppender.setRollingPolicy(logFilePolicy)
 
+    val baseDir = AppConfig.DEFAULT_BITCOIN_S_DATADIR
```
Christewart: This is going to cause problems if the user isn't using the `DEFAULT_BITCOIN_S_DATADIR`, i'm not sure of a better way to handle this though.

Christewart: Maybe we need to make this into a class that has the `AppConfig` passed into it implicitly, this means we can also get rid of the implicit on `getLoggerImpl()`

db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala
```diff
@@ -27,8 +28,10 @@ private[bitcoins] trait AppLoggers {
     case object Database extends LoggerKind
   }
 
+  def network: NetworkParameters
```
Christewart: This should really come from the `AppConfig.network`

PR #1088, opened 2020-02-02
DLC Dust Outputs

nkohen:
Added the logic to remove outputs with value below `Policy.dustThreshold`, finally enabling us to remove the weird over-collateralization in tests (where each party was putting in 1 BTC to gain or lose 1 mBTC).

Tests now have single output CETs as is expected from binary outcome DLCs.

### DIFF
--- a/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
+++ b/dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
@@ -82,6 +82,7 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
   val preCommittedR: ECPublicKey = preCommittedK.publicKey
   val localInput: CurrencyUnit = CurrencyUnits.oneBTC
   val remoteInput: CurrencyUnit = CurrencyUnits.oneBTC
+  val totalInput: CurrencyUnit = localInput + remoteInput
 
   val inputPrivKeyLocal: ECPrivateKey = ECPrivateKey.freshPrivateKey
   val inputPubKeyLocal: ECPublicKey = inputPrivKeyLocal.publicKey
@@ -212,8 +213,8 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
         remoteFundingInputs = Vector(
           (TransactionOutPoint(fundingTx.txIdBE, remoteVout),
            fundingTx.outputs(remoteVout.toInt))),
-        winPayout = localInput + CurrencyUnits.oneMBTC,
-        losePayout = localInput - CurrencyUnits.oneMBTC,
+        winPayout = totalInput,
+        losePayout = CurrencyUnits.zero,
         timeouts = DLCTimeouts(penaltyTimeout = csvTimeout,
                                tomorrowInBlocks,
                                twoDaysInBlocks),
@@ -237,8 +238,8 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
         remoteFundingInputs = Vector(
           (TransactionOutPoint(fundingTx.txIdBE, localVout),
            fundingTx.outputs(localVout.toInt))),
-        winPayout = remoteInput - CurrencyUnits.oneMBTC,
-        losePayout = remoteInput + CurrencyUnits.oneMBTC,
+        winPayout = CurrencyUnits.zero,
+        losePayout = totalInput,
         timeouts = DLCTimeouts(penaltyTimeout = csvTimeout,
                                tomorrowInBlocks,
                                twoDaysInBlocks),
@@ -257,18 +258,26 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
   }
 
   def validateOutcome(outcome: DLCOutcome): Future[Assertion] = {
-    for {
-      client <- clientF
-      regtestLocalClosingTx <- client.getRawTransaction(
-        outcome.closingTx.txIdBE)
-    } yield {
-      assert(regtestLocalClosingTx.hex == outcome.closingTx)
-      assert(regtestLocalClosingTx.confirmations.isDefined)
-      assert(regtestLocalClosingTx.confirmations.get >= 6)
+    clientF.flatMap { client =>
+      outcome.closingTxOpt match {
+        case None =>
+          Future {
+            assert(noEmptySPKOutputs(outcome.fundingTx))
+            assert(noEmptySPKOutputs(outcome.cet))
+          }
+        case Some(closingTx) =>
+          val txResultF = client.getRawTransaction(closingTx.txIdBE)
 
-      assert(noEmptySPKOutputs(outcome.fundingTx))
-      assert(noEmptySPKOutputs(outcome.cet))
-      assert(noEmptySPKOutputs(outcome.closingTx))
+          txResultF.map { regtestLocalClosingTx =>
+            assert(regtestLocalClosingTx.hex == outcome.closingTxOpt.get)
+            assert(regtestLocalClosingTx.confirmations.isDefined)
+            assert(regtestLocalClosingTx.confirmations.get >= 6)
+
+            assert(noEmptySPKOutputs(outcome.fundingTx))
+            assert(noEmptySPKOutputs(outcome.cet))
+            assert(noEmptySPKOutputs(outcome.closingTxOpt.get))
+          }
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -257,18 +258,26 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+            assert(noEmptySPKOutputs(outcome.cet))
+          }
+        case Some(closingTx) =>
+          val txResultF = client.getRawTransaction(closingTx.txIdBE)
 
-      assert(noEmptySPKOutputs(outcome.fundingTx))
-      assert(noEmptySPKOutputs(outcome.cet))
-      assert(noEmptySPKOutputs(outcome.closingTx))
+          txResultF.map { regtestLocalClosingTx =>
+            assert(regtestLocalClosingTx.hex == outcome.closingTxOpt.get)
```
benthecarman: Should add `assert(outcome.closingTxOpt.isDefined)` before this

benthecarman: Should add `assert(outcome.closingTxOpt.isDefined)` before this

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -364,7 +364,7 @@ case class BinaryOutcomeDLCClient(
       sequence)
     val utx = BaseTransaction(TransactionConstants.validLockVersion,
                               Vector(input),
-                              outputs,
+                              outputs.filter(_.value >= Policy.dustThreshold),
```
benthecarman, line 367: The `dustThreshold` should be a multiple of the `feeRate`

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
benthecarman: Should `assert(outcome.closingTxOpt.isDefined)` before

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
benthecarman, line 367: The `dustThreshold` should be a multiple of the feeRate

benthecarman (changes requested):
Seems like the current bitcoin-s way to determine if an output is dust is that it just under 1000 sats.  This however should be changed to be based off of the tx fee rate. For example, if an output is 31 vbytes and the fee rate is 33 sats/vbyte, you will be paying 1023 sats to create 1000 sats.

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -470,10 +479,14 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
       _ <- waitUntilBlock(
         unilateralDLC.timeouts.contractMaturity.toUInt32.toInt)
       _ <- publishTransaction(unilateralOutcome.cet)
-      _ <- publishTransaction(unilateralOutcome.closingTx)
+      _ <- unilateralOutcome.closingTxOpt
+        .map(publishTransaction)
+        .getOrElse(FutureUtil.unit)
```
Christewart: do we want to fail silently here? is that what `FutureUtil.unit` will do here?

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -470,10 +479,14 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
       otherOutcome <- otherDLC.executeRemoteUnilateralDLC(otherSetup,
                                                           unilateralOutcome.cet)
-      _ <- publishTransaction(otherOutcome.closingTx)
+      _ <- otherOutcome.closingTxOpt
+        .map(publishTransaction)
+        .getOrElse(FutureUtil.unit)
```
Christewart: ditto

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -552,15 +565,29 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
         cetWronglyPublished)
       _ = assert(toRemoteOutcome.cet == cetWronglyPublished)
       _ = assert(justiceOutcome.cet == cetWronglyPublished)
-      _ <- publishTransaction(toRemoteOutcome.closingTx)
-      _ <- recoverToSucceededIf[BitcoindException](
-        publishTransaction(justiceOutcome.closingTx))
+      _ <- toRemoteOutcome.closingTxOpt
+        .map(publishTransaction)
+        .getOrElse(FutureUtil.unit)
```
Christewart: ditto

dlc-test/src/test/scala/org/bitcoins/dlc/BinaryOutcomeDLCClientIntegrationTest.scala
```diff
@@ -552,15 +565,29 @@ class BinaryOutcomeDLCClientIntegrationTest extends BitcoindRpcTest {
...
+      _ <- justiceOutcome.closingTxOpt
+        .map { tx =>
+          recoverToSucceededIf[BitcoindException](
+            publishTransaction(tx)
+          )
+        }
+        .getOrElse(FutureUtil.unit)
```
Christewart, line 635: ditto

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
Christewart: I don't think it is `feeRate`, rather `dustRelayFee`, at least that is what I'm reading from these two links. 

https://bitcoin.stackexchange.com/a/41082

https://github.com/bitcoin/bitcoin/commit/9022aa3

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -518,10 +518,11 @@ case class BinaryOutcomeDLCClient(
...
       TransactionInput(fundingOutPoint, EmptyScriptSignature, sequence)
 
-    val unsignedTx = BaseTransaction(TransactionConstants.validLockVersion,
-                                     Vector(fundingInput),
-                                     outputs,
-                                     timeouts.contractMaturity.toUInt32)
+    val unsignedTx = BaseTransaction(
+      TransactionConstants.validLockVersion,
+      Vector(fundingInput),
+      outputs.filter(_.value >= Policy.dustThreshold),
```
Christewart, line 524: So i'm torn on this. I don't really like removing outputs silently from a transaction, rather I would rather have the caller of this API have to explicitly remove the output themselves. When we start removing outputs internally it makes it easy to fudge something up and pay a large fee.

dlc/src/main/scala/org/bitcoins/dlc/DLCOutcome.scala
```diff
@@ -7,7 +7,7 @@ import org.bitcoins.core.wallet.utxo.BitcoinUTXOSpendingInfoFull
 case class DLCOutcome(
     fundingTx: Transaction,
     cet: Transaction,
-    closingTx: Transaction,
+    closingTxOpt: Option[Transaction],
```
Christewart: So what exactly does this state mean? Does this mean we cannot ultimately settle the DLC because the payout is too small?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
nkohen: At some point me and Tibo plan on explicitly putting the algorithms for doing this as well as fee estimation up on the spec, as all parties must be doing things deterministically (and so allowing users to specify is off the table). But as per the discussion above I would view this in its current form as more of a place-holder (whose main function is to wipe 0 sat outputs so that tests pass as that is a common case)

benthecarman (approved):

Christewart (approved):

PR #1092, opened 2020-02-03
Reworked DLC sequence numbers to disable RBF

nkohen:
Using `0xffffffffe` as the sequence numbers on non-CSV inputs in DLC txs to opt-out of RBF (see BIP 125).


core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionConstants.scala
```diff
@@ -11,6 +11,7 @@ trait TransactionConstants {
   lazy val validLockVersion = Int32(2)
   lazy val lockTime = UInt32.zero
   lazy val sequence = UInt32(4294967295L)
+  lazy val disableRBFSequence = UInt32(4294967294L)
```
benthecarman: nit: maybe make this `sequence - 1` so it is clearer to see that it is `MAX - 1`

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -619,12 +616,10 @@ object TxBuilder {
               //none of these script types affect the sequence number of a tx
               //the sequence only needs to be adjustd if we have replace by fee (RBF) enabled
               //see BIP125 for more information
-              val sequence =
-                if (isRBFEnabled) UInt32.zero else TransactionConstants.sequence
```
benthecarman: This comment is no longer relevant

core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionConstants.scala
Christewart: This enables RBF: https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#summary


>Explicit signaling: A transaction is considered to have opted in to allowing replacement of itself if any of its inputs have an nSequence number less than (0xffffffff - 1).

core/src/main/scala/org/bitcoins/core/wallet/builder/TxBuilder.scala
```diff
@@ -570,7 +570,7 @@ object TxBuilder {
     */
   def calcSequenceForInputs(
       utxos: Seq[UTXOSpendingInfo],
-      isRBFEnabled: Boolean): Seq[TransactionInput] = {
+      defaultSequence: UInt32 = UInt32.zero): Seq[TransactionInput] = {
```
Christewart: why not make this `Policy.sequence` or something like that? I think users should have to explicitly enable things like RBF/CSV

PR #1094, opened 2020-02-03
Added server handling for empty account tables while a seed exists

nkohen:
Now the wallet is not considered initialized by the Server if the accounts table is empty.

And when a BIP39KeyManager is being initialized while a seed already exists, that seed is used rather than failing.


rorp:
Should it simply create a missing record in `wallet_accounts` table?

PR #1097, opened 2020-02-06
Support for Eclair 0.3.3

rorp:
See https://github.com/ACINQ/eclair/releases/tag/v0.3.3

### DIFF
--- a/eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
+++ b/eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
@@ -947,8 +947,6 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
     val client1F = connectedClientsF.map(_.c1)
     val client2F = connectedClientsF.map(_.c2)
 
-    // allupdates for a single node is broken in Eclair 0.3.2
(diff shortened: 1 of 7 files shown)


eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -1100,12 +1097,11 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
     //which is why i am choosing to use them for this test
     // allupdates for a single node is broken in Eclair 0.3.2
```
Christewart: Looks like this got fixed, can we remove it?

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -937,11 +949,11 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
 
     // allupdates for a single node is broken in Eclair 0.3.2
     // TODO remove recoverToPendingIf when https://github.com/ACINQ/eclair/issues/1179 is fixed
```
Christewart: this is fixed, can we remove it?

Issue #1100, opened 2020-02-06
Graalvm native image cli  java.lang.NoClassDefFoundError: Lorg/codehaus/janino/ScriptEvaluator;


Christewart:
Jdk version: 
```
java -version
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment GraalVM CE 19.3.1 (build 11.0.6+9-jvmci-19.3-b07)
OpenJDK 64-Bit Server VM GraalVM CE 19.3.1 (build 11.0.6+9-jvmci-19.3-b07, mixed mode, sharing)
```

Native image version: 
```
native-image --version
GraalVM Version 19.3.1 CE
```

This was caused when running Scala `2.13.1` on  7fbc6423a5c81aec1fa7c31e3307fa2452585b95 

with the command ` cli/graalvm-native-image:packageBin` inside of sbt

The stack trace:

>[error] Fatal error: com.oracle.svm.core.util.VMError$HostedError: java.lang.NoClassDefFoundError: Lorg/codehaus/janino/ScriptEvaluator;
[error]         at com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:70)
[error]         at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:636)
[error]         at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:445)
[error]         at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
[error]         at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
[error]         at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
[error]         at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
[error]         at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
[error]         at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
[error] Caused by: java.lang.NoClassDefFoundError: Lorg/codehaus/janino/ScriptEvaluator;
[error]         at java.base/java.lang.Class.getDeclaredFields0(Native Method)
[...]

PR #1101, opened 2020-02-06
Add codehause dependecy as a work around for issue 1110, there is a i…

Christewart:
…ssue in graalvm native image generator where images aren't being generated correctly. THis is related to the logback dependecy

See #1100 

As a workaround for the graalvm issue, we explictly include the `codehaus`. This should be able to be removed when the underlying oracle graavm issue is fixed.

### DIFF
--- a/project/Deps.scala
+++ b/project/Deps.scala
@@ -193,6 +193,9 @@ object Deps {
     else Compile.newMicroPickle,
     Compile.logback,
     Compile.scopt,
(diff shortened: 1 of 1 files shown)


benthecarman:
If this fixes the problem you should update the cli build docs, so that it no longer lists you need to use `++ 2.12.10`

Christewart:
I need to use `++2.12.10` for this to work, see #1104 (different error)

PR #1111, opened 2020-02-11
Reserved TxoState

### LINKED ISSUE #1008: Mark 'SpendingInfoDb' as in use
We should have the ability to mark a `SpendingInfoDb` as 'in use'. 

This is needed when you are operating a wallet that is processing a lot of transactions, or when you are doing offline signing. You don't want _another_ transaction using the utxos you are attempting to sign for. 

If they do end up using the utxos you sign for, you will be effectively attempting to double spend yourself which means only one of these transactions will be confirmed on the blockchain.

There are two ways to implement this: 

Add another row/field to `SpendingInfoDb`, or have some sort of table that controls usage of the utxos and foreign keys into `txo_spending_info` table which controls the `SpendingInfoDb`.

benthecarman:
Closes #1008

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/utxo/TxoStateTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/utxo/TxoStateTest.scala
@@ -15,6 +15,8 @@ class TxoStateTest extends BitcoinSUnitTest {
     TxoState.fromString("ConfirmedReceived").get must be(
       TxoState.ConfirmedReceived)
 
+    TxoState.fromString("Reserved").get must be(TxoState.Reserved)
+
     TxoState.fromString("PendingConfirmationsSpent").get must be(
       TxoState.PendingConfirmationsSpent)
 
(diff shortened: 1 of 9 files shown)


Christewart (commented):
Things look good so far, but do we want to integrate this further into the code base? Specifically I was thinking of adding a flag to [`fundRawTransaction`](https://github.com/bitcoin-s/bitcoin-s/blob/a6e21fe43b0e34f22cd3a839e225e814fa6f6550/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala#L17) and [`fundRawTransactionInternal`](https://github.com/bitcoin-s/bitcoin-s/blob/a6e21fe43b0e34f22cd3a839e225e814fa6f6550/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala#L51) that allows for you reserve the utxo. If that flag is set, we change the state to reserved in the database.

Christewart (commented):
Add unit test to make sure that this actually work when the `markAsReserved` is enabled.

PR #1112, opened 2020-02-11
Initial Signet support

pavel-main:
## Description

Initial PR to support default Signet ("Signet Global Test Net III"). Custom Signet is not implemented in this PR.

More info about [Signet](https://en.bitcoin.it/wiki/Signet):

- [BIP-0325](https://github.com/bitcoin/bips/blob/master/bip-0325.mediawiki)
- https://github.com/bitcoin/bitcoin/pull/16411
- https://github.com/rust-bitcoin/rust-bitcoin/pull/291

# TODO

- [x] New network parameters
- [x] Bech32 human-readable part changes
- [x] New configuration option
- [x] Genesis block creation
- [ ] Tests to verify genesis block hash

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
+++ b/core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
@@ -160,13 +160,10 @@ sealed abstract class SigNet extends BitcoinNetwork {
   /**
    * @inheritdoc
    */
-  override def dnsSeeds = {
-    Seq(
-      "178.128.221.177",
-      "2a01:7c8:d005:390::5",
-      "ntv3mtqw5wt63red.onion:38333",
-    )
-  }
+  override def dnsSeeds: Seq[String] =
+    Seq("178.128.221.177",
+        "2a01:7c8:d005:390::5",
+        "ntv3mtqw5wt63red.onion:38333")
 
   /**
    * @inheritdoc
--- a/core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
+++ b/core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
@@ -143,10 +143,43 @@ sealed abstract class RegTest extends BitcoinNetwork {
 }
 
 final case object RegTest extends RegTest
+
+sealed abstract class SigNet extends BitcoinNetwork {
+  override def chainParams: SigNetChainParams.type = SigNetChainParams
+
+  /**
+   * @inheritdoc
+   */
+  override def port = 38333
+
+  /**
+   * @inheritdoc
+   */
+  override def rpcPort = 38332
+
+  /**
+   * @inheritdoc
+   */
+  override def dnsSeeds = {
+    Seq(
+      "178.128.221.177",
+      "2a01:7c8:d005:390::5",
+      "ntv3mtqw5wt63red.onion:38333",
+    )
+  }
+
+  /**
+   * @inheritdoc
+   */
+  override def magicBytes = ByteVector(0xf0, 0xc7, 0x70, 0x6a)
+
+}
+
+final case object SigNet extends SigNet
 // $COVERAGE-ON$
 
 object Networks {
-  val knownNetworks: Seq[NetworkParameters] = Seq(MainNet, TestNet3, RegTest)
+  val knownNetworks: Seq[NetworkParameters] = Seq(MainNet, TestNet3, RegTest, SigNet)
   val secretKeyBytes: Seq[ByteVector] = knownNetworks.map(_.privateKey)
   val p2pkhNetworkBytes: Seq[ByteVector] = knownNetworks.map(_.p2pkhNetworkByte)
   val p2shNetworkBytes: Seq[ByteVector] = knownNetworks.map(_.p2shNetworkByte)
@@ -156,6 +189,7 @@ object Networks {
     case "mainnet" => Some(MainNet)
     case "testnet" => Some(TestNet3)
     case "regtest" => Some(RegTest)
+    case "signet"  => Some(SigNet)
     case _: String => None
   }
 
@@ -163,7 +197,8 @@ object Networks {
   def magicToNetwork: Map[ByteVector, NetworkParameters] = Map(
     MainNet.magicBytes -> MainNet,
     TestNet3.magicBytes -> TestNet3,
-    RegTest.magicBytes -> RegTest
+    RegTest.magicBytes -> RegTest,
+    SigNet.magicBytes -> SigNet
   )
 
   def bytesToNetwork: Map[ByteVector, NetworkParameters] = Map(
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -104,6 +100,52 @@ sealed abstract class ChainParams {
...
+      Seq(input),
+      Seq(output),
+      TransactionConstants.lockTime)
+    val prevBlockHash = DoubleSha256Digest(
+      "0000000000000000000000000000000000000000000000000000000000000000")
+    val merkleRootHash = Merkle.computeMerkleRoot(Seq(tx))
+    val genesisBlockHeader =
+      BlockHeader(version, prevBlockHash, merkleRootHash, time, nBits, nonce)
+    val genesisBlock = Block(genesisBlockHeader, Seq(tx))
+    genesisBlock
```
benthecarman, line 146: nit: you could not define this as a `val` and just have `Block(genesisBlockHeader, Seq(tx))`

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
```diff
@@ -143,10 +143,43 @@ sealed abstract class RegTest extends BitcoinNetwork {
...
+   */
+  override def rpcPort = 38332
+
+  /**
+   * @inheritdoc
+   */
+  override def dnsSeeds = {
+    Seq(
+      "178.128.221.177",
+      "2a01:7c8:d005:390::5",
```
Christewart: i'm not sure how well supported ipv6 is in our code base and also onion addresses  -- I haven't tried them before. Just an FYI.

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -104,6 +100,52 @@ sealed abstract class ChainParams {
...
+      amount: CurrencyUnit): Block = {
+    val defaultSignetKey = DoubleSha256Digest(
+      "03ad5e0edad18cb1f0fc0d28a3d4f1f3e445640337489abb10404f2d1e086be430")
+    val const = ScriptConstant(defaultSignetKey.bytes)
+
+    val asm = {
+      List(OP_1, const, OP_1, OP_CHECKMULTISIG)
+    }
+
+    val blockScript = ScriptSignature.fromAsm(asm)
```
Christewart, line 131: i'm not sure how signet's signing works, but you might want to use our built in script types.

You can create a multisig script with

```scala
scala> import org.bitcoins.core.protocol.script._
import org.bitcoins.core.protocol.script._

scala> import org.bitcoins.core.crypto._

scala> val pubKey = ECPublicKey.fromHex("03ad5e0edad18cb1f0fc0d28a3d4f1f3e445640337489abb10404f2d1e086be430")
pubKey: org.bitcoins.core.crypto.ECPublicKey = ECPublicKey(03ad5e0edad18cb1f0fc0d28a3d4f1f3e445640337489abb10404f2d1e086be430)

scala> val m = MultiSignatureScriptPubKey(1,Vector(pubKey))
m: org.bitcoins.core.protocol.script.MultiSignatureScriptPubKey = MultiSignatureScriptPubKey(1, 1, ECPublicKey(03ad5e0edad18cb1f0fc0d28a3d4f1f3e445640337489abb10404f2d1e086be430))

scala> m.asm
res0: Seq[org.bitcoins.core.script.constant.ScriptToken] = Vector(OP_1, BytesToPushOntoStackImpl(33), ScriptConstantImpl(ByteVector(33 bytes, 0x03ad5e0edad18cb1f0fc0d28a3d4f1f3e445640337489abb10404f2d1e086be430)), OP_1, OP_CHECKMULTISIG)
```

This also makes sure you get the asm format correct :-).

Depending on how signet works (raw multisig, vs wrapped in p2wsh/p2sh) you can wrap that script like this:

```scala
scala> val p2sh = P2SHScriptPubKey(m)
p2sh: org.bitcoins.core.protocol.script.P2SHScriptPubKey = P2SHScriptPubKey(Sha256Hash160DigestImpl(04699beafffcd5b24d48e5b5a9f02979aa92d2c0))

scala> val p2wsh = P2WSHWitnessSPKV0(m)
p2wsh: org.bitcoins.core.protocol.script.P2WSHWitnessSPKV0 = wsh(Sha256DigestImpl(d64e78dc1940a14c251ab244ee7d54999018e72894ae85d8d7baf710be6215b6))

```

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -399,6 +438,57 @@ object RegTestNetChainParams extends BitcoinChainParams {
...
+   * Regtest allows trivial difficulty blocks
+   * [[https://github.com/bitcoin/bitcoin/blob/006d9ab9e59a00fa02b8d5f32919f9d57567fa65/src/chainparams.cpp#L312 regtest min difficulty]]
+   */
+  override lazy val allowMinDifficultyBlocks: Boolean = false
+
+  /**
+   * Regtest allows pow retargetting
+   * [[https://github.com/bitcoin/bitcoin/blob/006d9ab9e59a00fa02b8d5f32919f9d57567fa65/src/chainparams.cpp#L313 signet pow retargetting]]
+   */
+  override lazy val noRetargeting: Boolean = false
```
Christewart, line 484: Interesting, i'm kind of surprised this is not `true`. So signet allows for POW retargetting? I thought signet got rid of POW all together. cc @kallewoof

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
kallewoof: Signet is raw, I believe; no wrappery. (In fact, I don't think it supports segwit right now. Probably worth looking at, but not hugely important.)

core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
kallewoof: Signet in fact includes proof of work! This is, generally, kept at a very low difficulty, however (even lower than main net's lowest difficulty).

Christewart:
Re: Ci failures:

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/649678802?utm_medium=notification&utm_source=github_status

That should be able to be fixed by `sbt test:scalafmt`

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/649678795#L2362

^ this CI failure is because we maintain a high test coverage requirement for our `core` module to make sure we don't regress our test suite over time. You can take a look at the coverage reports here: 

https://coveralls.io/builds/28696651

PR #1123, opened 2020-02-14
Four Key DLC

nkohen:
Re-implemented key derivation to be compatible with traditional key discovery as described [here](https://github.com/discreetlogcontracts/dlcspecs/issues/8#issuecomment-586391136)


dlc/src/main/scala/org/bitcoins/dlc/DLCPublicKeys.scala
```diff
@@ -0,0 +1,57 @@
...
+      extPubKey
+        .deriveChildPubKey(BIP32Path.fromString(s"m/0/${nextAddressIndex + 3}"))
+        .get
+        .key
+
+    DLCPublicKeys(
+      fundingKey = fundingPubKey,
+      toLocalCETKey = cetToLocalPubKey,
+      toRemoteCETKey = cetToRemotePubKey,
+      finalAddress = Bech32Address(P2WPKHWitnessSPKV0(finalPubKey), RegTest)
```
benthecarman: Should make the network a parameter and maybe give it a default val

benthecarman (approved):

dlc/src/main/scala/org/bitcoins/dlc/DLCPublicKeys.scala
```diff
@@ -0,0 +1,59 @@
...
+      toLocalCETKey = cetToLocalPubKey,
+      toRemoteCETKey = cetToRemotePubKey,
+      finalAddress = Bech32Address(P2WPKHWitnessSPKV0(finalPubKey), network)
+    )
+  }
+
+  def fromExtPrivKeyAndIndex(
+      extPrivKey: ExtPrivateKey,
+      nextAddressIndex: Int,
+      network: NetworkParameters = RegTest): DLCPublicKeys = {
```
Christewart: why hard coded network?

dlc/src/main/scala/org/bitcoins/dlc/DLCPublicKeys.scala
```diff
@@ -0,0 +1,59 @@
...
+    toLocalCETKey: ECPublicKey,
+    toRemoteCETKey: ECPublicKey,
+    finalAddress: BitcoinAddress)
+
+object DLCPublicKeys {
+
+  def fromExtPubKeyAndIndex(
+      extPubKey: ExtPublicKey,
+      nextAddressIndex: Int,
+      network: NetworkParameters = RegTest): DLCPublicKeys = {
```
Christewart: why hard coded network?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -883,7 +883,8 @@ case class BinaryOutcomeDLCClient(
     */
   def executeRemoteUnilateralDLC(
       dlcSetup: SetupDLC,
-      publishedCET: Transaction): Future[UnilateralDLCOutcome] = {
+      publishedCET: Transaction,
+      sweepPrivKey: ECPrivateKey): Future[UnilateralDLCOutcome] = {
```
benthecarman, line 887: Would it be better to pass in a `SpendingInfo` so we can sweep to any addr we want?

nkohen: at some point I think that would be nice, would prefer to do it on another PR though as this one is slowly growing in scope :P

PR #1124, opened 2020-02-14
Funding Change Dust Removal

nkohen:
Remove dust change outputs from funding transaction.


benthecarman (approved):

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -272,9 +273,15 @@ case class BinaryOutcomeDLCClient(
...
-                           feeRate,
-                           Vector(changeSPK, remoteChangeSPK))
+    val txWithFee = subtractFeeFromOutputs(txWithoutFee,
+                                           feeRate,
+                                           Vector(changeSPK, remoteChangeSPK))
+
+    BaseTransaction(version = txWithFee.version,
+                    inputs = txWithFee.inputs,
+                    outputs =
+                      txWithFee.outputs.filter(_.value >= Policy.dustThreshold),
```
Christewart, line 283: doesn't this remove all outputs under dust, not just the change?

nkohen: yup :sunglasses:

Issue #1121, opened 2020-02-13
Make it so `BitcoindRpcTestUtil.baseBinaryDirectory` is configurable


Christewart:
This is related to #1109 

Currently we just hard code the `baseBinaryDirectory` to be the `user.dir`. This means we can't share binaries across projects. We should allow this to be configured in some way by a user. 

https://github.com/bitcoin-s/bitcoin-s/blob/5c77d549ac8c8d23f0c9b248b72ecfda56f69cf0/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala#L161

Issue #1109, opened 2020-02-11
Make it so that bitcoind/eclair binaries are stored under $HOME so they can be used cross project


Christewart:
Currently when you run  the `downloadBitcoind`, `downloadEclair` sbt tasks it stores the bitcoind/eclair binaries are stored in this projects root directory:

```zshrc  
➜ pwd
/home/suredbits/dev/bitcoin-s
➜  bitcoin-s git:(master) ls binaries
bitcoind  eclair
```

However if we have other projects, that use the same scripts, we have to duplicate the storage of these  binaries. 

I think we should rename the `binaries` folder to something else, and store it as a hidden directory in the user's `$HOME` so that we don't have duplicate downloads. We can always just check to see if it exists in that directory, rather than each project's directory.

PR #1128, opened 2020-02-16
Move binaries to be stored inside of the HOME/.bitcoin-s/binaries ins…

Christewart:
…tead of usrdir/binaries. This makes it so different projects that use bitcoin-s can re-use binaries instead of having to download them for every project

This addresses #1121 and #1109. 

Basically if you run `downloadBitcoind` and `downloadEclair` the binaries that are downloaded will be stored under `$HOME/.bitcoin-s/binaries` rather than the current working directory `${Properties.userDir}/binaries/`. 

This means after this PR first gets merged, you will have to re-download these binaries, or manually move the existing `binaries/` to `~/.bitcoin-s/`

### DIFF
--- a/project/CommonSettings.scala
+++ b/project/CommonSettings.scala
@@ -1,4 +1,6 @@
 // these two imports are needed for sbt syntax to work
+import java.nio.file.Paths
+
 import sbt._
 import sbt.Keys._
 
@@ -116,4 +118,6 @@ object CommonSettings {
   ) ++ testSettings
 
   lazy val prodSettings: Seq[Setting[_]] = settings
+
+  lazy val binariesPath = Paths.get(Properties.userHome, ".bitcoin-s", "binaries")
 }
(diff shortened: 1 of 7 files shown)


rorp (approved):

project/CommonSettings.scala
```diff
@@ -116,4 +118,6 @@ object CommonSettings {
   ) ++ testSettings
 
   lazy val prodSettings: Seq[Setting[_]] = settings
+
+  lazy val binariesPath = Paths.get(Properties.userHome, ".bitcoin-s", "binaries")
```
nkohen, line 122: Any reason not to read the conf file and go to the correct `baseDataDir`? What if someone is running from a different datadir?

Christewart: You are talking about running the application, where as this is for running tests. You don't need an `AppConfig` to run `bitcoindRpcTest` or `eclairTest`.

nkohen (approved):

Issue #916, opened 2019-12-04
chain tests async bug with chain tables and fixtures


Christewart:
https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/620409556#L445

it looks like we have an async issue with table creation in our chain project. It showed up on #904, here are the logs 

```
2019-12-04T02:16:39 ERROR [DATABASE] Error when executing query flatMap org.bitcoins.chain.models.BlockHeaderDAO$$Lambda$1074/0x0000000100885840@48cc67f3

2019-12-04T02:16:39 ERROR [DATABASE] org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: block_headers)

Failed to build fixture with err=org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: block_headers)

Failed fixture test execution=org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: block_headers)
```

Either 

1. We are not calling `.map()`/`.flatMap()` some where we should be when we are constructing sql tables
2. We are not properly initializing one of the chain tests

Christewart:
Another instance of this: 

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/651137497#L787

PR #1129, opened 2020-02-16
Fix async bug in chain fixtures setup phase

Christewart:
Address #916 

In our chain project fixtures we did not make sure that tables were fully created before trying to insert information into tables in futures. This causes race conditions on slow CI machines that are fixed now because we call 'makeChainHandler()' inside of the setupTableF flatMap

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
@@ -436,15 +436,13 @@ object ChainUnitTest extends ChainVerificationLogger {
       appConfig: ChainAppConfig): Future[(ChainHandler, BlockHeaderDb)] = {
     val tableSetupF = setupAllTables()
 
-    val chainHandlerF = makeChainHandler()
-
     val genesisHeaderF = tableSetupF.flatMap { _ =>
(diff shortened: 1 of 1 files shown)


testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
```diff
@@ -436,15 +436,13 @@ object ChainUnitTest extends ChainVerificationLogger {
       appConfig: ChainAppConfig): Future[(ChainHandler, BlockHeaderDb)] = {
     val tableSetupF = setupAllTables()
 
-    val chainHandlerF = makeChainHandler()
-
     val genesisHeaderF = tableSetupF.flatMap { _ =>
+      val chainHandlerF = makeChainHandler()
       for {
         chainHandler <- chainHandlerF
```
nkohen, line 442: super nit: Any reason not to simply
```
for {
    chainHandler <- makeChainHandler()
...
```

nkohen (approved):

testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
Christewart: no, but i'm not going to do it because I don't want to wait for the build :-)

PR #1131, opened 2020-02-17
Seq and Map Wrappers

nkohen:
Introduced traits for case classes that wrap a single `Seq` or `Map` to extend in order to inherit all of the methods from the thing they wrap.

### DIFF
--- a/chain/src/main/scala-2.11/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala-2.11/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -6,7 +6,7 @@ import scala.collection.mutable
 
 /** @inheritdoc */
 case class Blockchain(headers: Vector[BlockHeaderDb])
-    with BaseBlockChain {
+    extends BaseBlockChain {
 
   protected[blockchain] def compObjectfromHeaders(
       headers: scala.collection.immutable.Seq[BlockHeaderDb]): Blockchain =
--- a/chain/src/main/scala-2.11/org/bitcoins/chain/blockchain/Blockchain.scala
+++ b/chain/src/main/scala-2.11/org/bitcoins/chain/blockchain/Blockchain.scala
@@ -2,11 +2,10 @@ package org.bitcoins.chain.blockchain
 
 import org.bitcoins.chain.models.BlockHeaderDb
 
-import scala.collection.{mutable, IndexedSeqLike}
+import scala.collection.mutable
 
 /** @inheritdoc */
 case class Blockchain(headers: Vector[BlockHeaderDb])
-    extends IndexedSeqLike[BlockHeaderDb, Vector[BlockHeaderDb]]
     with BaseBlockChain {
 
   protected[blockchain] def compObjectfromHeaders(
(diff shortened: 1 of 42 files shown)


chain/src/main/scala-2.11/org/bitcoins/chain/blockchain/Blockchain.scala
```diff
@@ -2,11 +2,10 @@ package org.bitcoins.chain.blockchain
 
 import org.bitcoins.chain.models.BlockHeaderDb
 
-import scala.collection.{mutable, IndexedSeqLike}
+import scala.collection.mutable
 
 /** @inheritdoc */
 case class Blockchain(headers: Vector[BlockHeaderDb])
-    extends IndexedSeqLike[BlockHeaderDb, Vector[BlockHeaderDb]]
```
Christewart, line 9: Why get rid of this?

Christewart: your wrapper classes just `extend Seq` which doesn't make any guarantees about Indexing, so this can devolve to worst case `O(n)` when attempting to do a indexed lookup. As far as I know, Scala doesn't make any guarantees about the implementation type behind this, which is why people never used `Seq` in older versions of Scala. I think things changed in `2.13` with the collections rework, but we don't just support 2.13

Christewart: For instance, this is how `Vector` is defined, you probably want to default to `IndexedSeq` IMO. 

```scala
final class Vector[+A] private[immutable] (private[collection] val startIndex: Int, private[collection] val endIndex: Int, focus: Int)
extends AbstractSeq[A]
   with IndexedSeq[A]
   with GenericTraversableTemplate[A, Vector]
   with IndexedSeqLike[A, Vector[A]]
   with VectorPointer[A @uncheckedVariance]
   with Serializable
   with CustomParallelizable[A, ParVector[A]
```

core/src/main/scala/org/bitcoins/core/util/Wrappers.scala
```diff
@@ -0,0 +1,19 @@
+package org.bitcoins.core.util
+
+trait SeqWrapper[+T] extends Seq[T] {
+  def wrapped: Seq[T]
```
Christewart: I think this is confusing, as you are technically _unwrapping_ the type the extends this and exposing the underlying implementation type.

Christewart:
It looks like things do not compile on `2.12`/`2.11`. Hopefully it's a simple fix, but i would try compiling locally as i wouldn't be surprised if there are other issues

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/651749168#L520

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/651749170?utm_medium=notification&utm_source=github_status

Christewart (commented):
Is there a `WrappersTest.scala` so we don't lose test coverage? 

https://travis-ci.org/bitcoin-s/bitcoin-s/jobs/652024169#L2376

PR #1134, opened 2020-02-18
Update native secp256k1 binaries with schnorr for osx_64

Tibo-lg:
Update the native binaries of secp256k1 including schnorr signatures for osx 64bits.


secp256k1jni/natives/osx_64/libsecp256k1.la
```diff
@@ -17,7 +17,7 @@ old_library='libsecp256k1.a'
 inherited_linker_flags=' '
 
 # Libraries that this one depends upon.
-dependency_libs=''
+dependency_libs=' -L/usr/local/opt/gmp/lib -lgmp'
```
Christewart: Hmm.. this seems strange. My understanding is that libsecp256k1 has no external deps

Christewart:
If you want to get CI to run follow the instructions here: 

https://github.com/bitcoin-s/bitcoin-s/blob/c884a149cfac7e57df506fe8dc11493060f4d425/docs/contributing.md#my-ci-isnt-running

secp256k1jni/natives/osx_64/libsecp256k1.la
Tibo-lg: Fixed. For reference I ran configure with `--with-bignum=no` to get rid of this.

PR #1156, opened 2020-02-21
2020 02 21 datadir configurable

Christewart:
This makes it so we can pass in a datadir that is configurable.

This is done with the [sbt-native-packager](https://www.scala-sbt.org/sbt-native-packager/archetypes/java_app/index.html#start-script-options) tool. You can read the docs to see how it works. 

You can now pass in a datadir to our server with 

```
 ./app/server/target/universal/stage/bin/bitcoin-s-server --datadir /path/to/datadir
```

you can also omit it, and the datadir defaults to `$HOME/.bitcoin-s/`

- [x] Test with old scala versions (`2.11.x` and `2.12.x`)
- [x] figure out if we can use named parameters (see below)

~I would like to make it so we don't just pass it into the `args: Array[String]` in `App`, as this means the ordering is important. It would be much nicer to have something like `--datadir /path/to/datadir/` than what I have now. @benthecarman do you have any idea of how to accomplish this?~

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.server
 
 import java.net.InetSocketAddress
-import java.nio.file.{Files, Path, Paths}
+import java.nio.file.{Files, Paths}
 
 import akka.actor.ActorSystem
 import org.bitcoins.chain.config.ChainAppConfig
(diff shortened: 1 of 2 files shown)


app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -22,7 +23,14 @@ import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
 
 object Main extends App {
-  implicit val conf = BitcoinSAppConfig.fromDefaultDatadir()
+  implicit val conf = {
+    val path: Path = if (args.isEmpty) {
```
Christewart: this is less than ideal as we add args this doesn't really scale (i.e. we can't omit the `datadir` arg)

benthecarman:
Instead of requiring an ordering you could just do 
```
val dataDirIndex = args.find("--datadir")
val dataDir = if (daraDirIndex >= 0)
	args(dataDirIndex + 1)
else DEFAULT_DATA_DIR
```

benthecarman (approved):

nkohen (approved):

PR #1164, opened 2020-02-24
DLC Database Improvements

### LINKED ISSUE #1151: Change DLC Database Schema
Give each `DLCPubKey` their own column

Give `OracleInfo`'s values each their own column (`pubkey`, `rValue`)

Give each member from `DLCTimeouts` their own column

Give each `CETSignatures` their own column

Create separate table for `fundingInputs` and their corresponding signatures

Properly set foreign keys for each table

benthecarman:
Closes #1151

Done in this PR:

Give `OracleInfo`'s values each their own column (`pubkey`, `rValue`)

Give each member from `DLCTimeouts` their own column

Give each `CETSignature` their own column

Give each `DLCPubkey` their own column, as well as use an address for `toRemote`

Create separate table for `fundingInputs` and their corresponding signatures, as well as, allows for multiple signatures per input

Properly set foreign keys for each table

---
if you had a DLC database before this you will need to do the following:
```
execute the following in walletdb.sqlite
-----------------------------------------
delete from flyway_schema_history where version=4;
DROP TABLE IF EXISTS wallet_dlcs;
DROP TABLE IF EXISTS wallet_dlc_offers;
DROP TABLE IF EXISTS wallet_dlc_accepts;
```


benthecarman:
Rename Executed DLC Table

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -1234,7 +1241,8 @@ object BinaryOutcomeDLCClient {
   }
 }
 
-case class FundingSignatures(sigs: Vector[PartialSignature])
+case class FundingSignatures(
+    sigs: Map[TransactionOutPoint, Vector[PartialSignature]])
```
benthecarman, line 1247: Make use `MapWrapper`

Christewart (approved):

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -489,22 +489,20 @@ case class PSBT(
...
       inputIndex < inputMaps.size,
       s"index must be less than the number of input maps present in the psbt, $inputIndex >= ${inputMaps.size}")
     require(
       !inputMaps(inputIndex).isFinalized,
       s"Cannot update an InputPSBTMap that is finalized, index: $inputIndex")
-    require(
-      !inputMaps(inputIndex).partialSignatures
-        .exists(_.pubKey == partialSignature.pubKey),
-      s"Input has already been signed by ${partialSignature.pubKey}"
-    )
```
nkohen: I'm pretty sure we still need a requirement of this nature don't we?

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -62,8 +62,8 @@ case class BinaryOutcomeDLCClient(
     losePayout: CurrencyUnit,
     timeouts: DLCTimeouts,
     feeRate: FeeUnit,
-    changeSPK: WitnessScriptPubKey,
-    remoteChangeSPK: WitnessScriptPubKey,
+    changeSPK: ScriptPubKey,
```
nkohen, line 65: This must be a segwit script to avoid malleability, not sure why you've changed this and the one below it but they should both change back

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -233,15 +234,20 @@ case class BinaryOutcomeDLCClient(
...
+          .map(sig => (utxo.outPoint, sig))
         vec :+ sigF
     }
+    val sigsF = Future.sequence(sigFs)
 
-    Future.sequence(sigFs).map(FundingSignatures.apply)
+    val sigMapF = sigsF.map { sigs =>
+      sigs.map(sig => (sig._1, sigs.filter(_._1 == sig._1).map(_._2))).toMap
+    }
+    sigMapF.map(FundingSignatures.apply)
```
nkohen: Rename to `outPointAndSigF` and all other names accordingly, then use `_.map { case (outPoint, sig) => ...` at the bottom, this is pretty hard to read as is

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -1234,7 +1241,17 @@ object BinaryOutcomeDLCClient {
...
+case class FundingSignatures(
+    sigs: Map[TransactionOutPoint, Vector[PartialSignature]])
+    extends MapWrapper[TransactionOutPoint, Vector[PartialSignature]] {
+  override protected def wrapped: Map[
+    TransactionOutPoint,
+    Vector[PartialSignature]] = sigs
+
+  def removed(key: TransactionOutPoint): Map[
+    TransactionOutPoint,
+    Vector[PartialSignature]] = sigs.filter(_._1 != key)
```
nkohen: this can be gotten rid of if you rebase

dlc/src/main/scala/org/bitcoins/dlc/DLCMessage.scala
```diff
@@ -117,7 +113,7 @@ object DLCMessage {
     def pubKeys: DLCPublicKeys
     def totalCollateral: Satoshis
     def fundingInputs: Vector[OutputReference]
-    def changeAddress: Bech32Address
+    def changeAddress: BitcoinAddress
```
nkohen, line 112: Same as above, this must be a segwit address.

dlc/src/main/scala/org/bitcoins/dlc/DLCMessage.scala
```diff
@@ -139,7 +135,7 @@ object DLCMessage {
       pubKeys: DLCPublicKeys,
       totalCollateral: Satoshis,
       fundingInputs: Vector[OutputReference],
-      changeAddress: Bech32Address,
+      changeAddress: BitcoinAddress,
```
nkohen, line 134: ditto

dlc/src/main/scala/org/bitcoins/dlc/DLCMessage.scala
```diff
@@ -305,7 +301,7 @@ object DLCMessage {
       totalCollateral: Satoshis,
       pubKeys: DLCPublicKeys,
       fundingInputs: Vector[OutputReference],
-      changeAddress: Bech32Address,
+      changeAddress: BitcoinAddress,
```
nkohen, line 297: ditto

wallet/src/main/scala/org/bitcoins/wallet/models/DLCAcceptTable.scala
```diff
@@ -2,22 +2,30 @@ package org.bitcoins.wallet.models
...
     totalCollateral: CurrencyUnit,
-    fundingInputs: Vector[OutputReference],
-    cetSigs: CETSignatures,
-    changeAddress: Bech32Address) {
-
-  def toDLCAccept: DLCAccept = {
+    winSig: PartialSignature,
+    loseSig: PartialSignature,
+    refundSig: PartialSignature,
+    changeAddress: BitcoinAddress) {
```
nkohen, line 22: ditto

wallet/src/main/scala/org/bitcoins/wallet/models/DLCAcceptTable.scala
```diff
@@ -2,22 +2,30 @@ package org.bitcoins.wallet.models
...
+import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.transaction.OutputReference
+import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import org.bitcoins.dlc.DLCMessage.DLCAccept
 import org.bitcoins.dlc.{CETSignatures, DLCPublicKeys}
 import slick.jdbc.SQLiteProfile.api._
-import slick.lifted.{PrimaryKey, ProvenShape}
+import slick.lifted.{ForeignKeyQuery, PrimaryKey, ProvenShape}
 
 case class DLCAcceptDb(
```
nkohen, line 13: Why is the auto-incrementing primary id not represented here?

wallet/src/main/scala/org/bitcoins/wallet/models/DLCFundingInputTable.scala
```diff
@@ -0,0 +1,64 @@
...
+import org.bitcoins.core.protocol.transaction.{
+  OutputReference,
+  TransactionOutPoint,
+  TransactionOutput
+}
+import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.{ForeignKeyQuery, PrimaryKey, ProvenShape}
+
+case class DLCFundingInputDb(
```
nkohen, line 13: Why is the auto-inc primary key not represented here

nkohen: (This is one reason to use slick to generate the tables)

wallet/src/main/scala/org/bitcoins/wallet/models/DLCOfferTable.scala
```diff
@@ -3,27 +3,38 @@ package org.bitcoins.wallet.models
...
 import org.bitcoins.core.protocol.transaction.OutputReference
 import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
 import org.bitcoins.dlc.DLCMessage.{ContractInfo, DLCOffer, OracleInfo}
-import org.bitcoins.dlc.{DLCPublicKeys, DLCTimeouts}
+import org.bitcoins.dlc.{DLCMessage, DLCPublicKeys, DLCTimeouts}
 import slick.jdbc.SQLiteProfile.api._
-import slick.lifted.{PrimaryKey, ProvenShape}
+import slick.lifted.{ForeignKeyQuery, PrimaryKey, ProvenShape}
 
 case class DLCOfferDb(
```
nkohen, line 15: auto-inc?

wallet/src/main/scala/org/bitcoins/wallet/models/DLCOfferTable.scala
```diff
@@ -3,27 +3,38 @@ package org.bitcoins.wallet.models
...
+    toLocalCETKey: ECPublicKey,
+    toRemoteCETAddr: BitcoinAddress,
+    finalAddress: BitcoinAddress,
     totalCollateral: CurrencyUnit,
-    fundingInputs: Vector[OutputReference],
     feeRate: SatoshisPerVirtualByte,
-    changeAddress: Bech32Address) {
-
-  def toDLCOffer: DLCOffer = {
+    changeAddress: BitcoinAddress) {
```
nkohen, line 29: ditto

wallet/src/main/scala/org/bitcoins/wallet/models/DLCOfferTable.scala
```diff
@@ -46,55 +83,79 @@ class DLCOfferTable(tag: Tag)
...
+  def toRemoteCETAddr: Rep[BitcoinAddress] = column("toRemoteCETAddr")
+
+  def finalAddress: Rep[BitcoinAddress] = column("finalAddress")
+
+  def totalCollateral: Rep[CurrencyUnit] = column("totalCollateral")
 
   def feeRate: Rep[SatoshisPerVirtualByte] = column("feeRate")
 
-  def changeAddress: Rep[Bech32Address] = column("changeAddress")
+  def changeAddress: Rep[BitcoinAddress] = column("changeAddress")
```
nkohen, line 106: ditto

wallet/src/main/scala/org/bitcoins/wallet/models/DLCTable.scala
```diff
@@ -2,21 +2,21 @@ package org.bitcoins.wallet.models
...
 import org.bitcoins.core.crypto._
 import org.bitcoins.core.hd.HDAccount
-import org.bitcoins.dlc.{CETSignatures, FundingSignatures}
+import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import slick.jdbc.SQLiteProfile.api._
-import slick.lifted.{PrimaryKey, ProvenShape}
+import slick.lifted.{ForeignKeyQuery, PrimaryKey, ProvenShape}
 
-case class ExecutedDLCDb(
+case class DLCDb(
```
nkohen, line 9: auto-inc?

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -90,23 +93,15 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
-            dlcOfferDb.pubKeys,
-            collateral.satoshis,
-            dlcOfferDb.fundingInputs,
-            dlcOfferDb.changeAddress,
-            satoshisPerVirtualByte,
-            dlcOfferDb.timeouts
-          )
-          Future.successful(offer)
+          Future.successful(
+            dlcOfferDb.toDLCOffer(fundingInputs.map(_.toOutputReference)))
```
nkohen: nit: intermediate vals would be nice

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -200,20 +200,14 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
-            dlcAcceptDb.totalCollateral.satoshis,
-            dlcAcceptDb.pubKeys,
-            dlcAcceptDb.fundingInputs,
-            dlcAcceptDb.changeAddress,
-            dlcAcceptDb.cetSigs,
-            dlcAcceptDb.eventId
-          )
-          Future.successful(accept)
+          Future.successful(
+            dlcAcceptDb.toDLCAccept(fundingInputs.map(_.toOutputReference)))
```
nkohen: nit: same

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -332,71 +331,104 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
+      updatedDLCDb = dlc.copy(winSigOpt = Some(cetSigs.winSig),
+                              loseSigOpt = Some(cetSigs.loseSig),
+                              refundSigOpt = Some(cetSigs.refundSig))
       _ <- dlcDAO.update(updatedDLCDb)
     } yield {
       DLCSign(cetSigs, fundingSigs, accept.eventId)
     }
   }
 
+  def addFundingSigs(sign: DLCSign): Future[Vector[DLCFundingInputDb]] = {
```
nkohen, line 360: I'm mildly concerned about async-safety here

nkohen: It is probably fine but a comment saying that this is the only way one should insert sigs to the db would be nice

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -332,71 +331,104 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
+      fundingInputs.filter(_.isInitiator).map(_.toOutputReference)
+    val acceptInputs =
+      fundingInputs.filterNot(_.isInitiator).map(_.toOutputReference)
+
     val (setupMsg, remoteSetupMsg) = if (dlcDb.isInitiator) {
-      (dlcOffer.toDLCOffer, dlcAccept.toDLCAccept)
+      (dlcOffer.toDLCOffer(offerInputs), dlcAccept.toDLCAccept(acceptInputs))
     } else {
-      (dlcAccept.toDLCAccept, dlcOffer.toDLCOffer)
+      (dlcAccept.toDLCAccept(acceptInputs), dlcOffer.toDLCOffer(offerInputs))
```
nkohen: should name the two values before the if to make this a little less redundant

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -516,11 +558,12 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
   override def acceptDLCMutualClose(
       mutualCloseSig: DLCMutualCloseSig): Future[Transaction] = {
     for {
-      _ <- updateDLCOracleSig(mutualCloseSig.eventId, mutualCloseSig.oracleSig)
+      dlcDb <- updateDLCOracleSig(mutualCloseSig.eventId,
+                                  mutualCloseSig.oracleSig)
+      (_, dlcOffer, dlcAccept, fundingInputs) <- getAllDLCData(
```
nkohen: was there a reason to use the older value rather than the newer one for `dlcDb`?

nkohen (changes requested):
Also, could you maybe document somewhere (here by comment or on an issue) what changes we still plan on making surrounding the db (e.g. having ADT states rather than Options and such)

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
benthecarman: No, but I think it makes more sense to get it from `getAllDLCData` so I'll switch it

wallet/src/main/scala/org/bitcoins/wallet/models/DLCAcceptTable.scala
benthecarman: None of the other tables that have an auto-incrementing primary id have it in their `XDb` or anywhere in the `XTable.scala` at all.

Christewart: I think i'm on the record saying I would like autoinc primary key, so we can decouple the protocol itself from our own internal representation. Plus we have a a bunch of infra to get this stuff for free `DbRowAutoInc`, `CRUDAutoInc`, `TableAutoInc` etc

nkohen: I think you might have misunderstood my intention here, I'm saying that since the auto inc primary key is in the db, it should be a parameter in these `case classes`

nkohen: Also just generally it seems that our slick tables only represent a subset of the real ones

nkohen: (If I use slick to generate this table it will not have an autoinc primary key)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -489,22 +489,26 @@ case class PSBT(
...
     require(
       !inputMaps(inputIndex).isFinalized,
       s"Cannot update an InputPSBTMap that is finalized, index: $inputIndex")
+    val intersect = inputMaps(inputIndex).partialSignatures
+      .intersect(partialSignatures)
     require(
-      !inputMaps(inputIndex).partialSignatures
-        .exists(_.pubKey == partialSignature.pubKey),
-      s"Input has already been signed by ${partialSignature.pubKey}"
+      intersect.isEmpty,
```
nkohen: You have to `map(_.pubKey)` before taking the intersection otherwise someone could add a `PartialSignature` for a pub key that has already been signed for

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
```diff
@@ -226,15 +226,23 @@ case class BinaryOutcomeDLCClient(
...
     }
+    val sigsF = Future.sequence(outPointAndSigFs)
 
-    Future.sequence(sigFs).map(FundingSignatures.apply)
+    val outPointAndSigsF = sigsF.map { sigs =>
+      sigs.map {
+        case (outPoint, _) =>
+          (outPoint, sigs.filter(_._1 == outPoint).map(_._2))
+      }.toMap
+    }
```
nkohen: I'm very confused by what is happening here and why this extra code is needed?

nkohen (commented):
Just two things here, will probably add another round of review if I run into any issues in running this locally

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: ECDSA sig malleability 👎

dlc/src/main/scala/org/bitcoins/dlc/BinaryOutcomeDLCClient.scala
benthecarman: It is turning a `Vector[(TransactionOutpoint, PartialSignature)]` into a `Map[TransactionOutpoint, Vector[PartialSignature]`

nkohen (approved):

PR #1169, opened 2020-02-26
 DLC Wallet Database tests 

benthecarman:
Built on top of #1164 

Basic tests for the wallet's dlc database

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
@@ -1,15 +1,14 @@
 package org.bitcoins.wallet.models
 
 import org.bitcoins.core.config.RegTest
+import org.bitcoins.core.crypto.Sha256DigestBE
 import org.bitcoins.core.currency.Satoshis
-import org.bitcoins.core.hd.{BIP32Path, HDAccount}
 import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.protocol.script.EmptyScriptPubKey
 import org.bitcoins.core.protocol.transaction.{
   TransactionOutPoint,
   TransactionOutput
 }
-import org.bitcoins.dlc.DLCMessage
 import org.bitcoins.testkit.fixtures.WalletDAOFixture
 import org.bitcoins.testkit.wallet.{BitcoinSWalletTest, DLCWalletTestUtil}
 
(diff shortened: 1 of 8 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
```diff
@@ -0,0 +1,199 @@
...
+          eventId = eventId,
+          isInitiator = true,
+          account = HDAccount.fromPath(BIP32Path.fromString("m/84'/0'/0'")).get,
+          keyIndex = 0,
+          winSigOpt = None,
+          loseSigOpt = None,
+          refundSigOpt = None,
+          oracleSigOpt = None
+        )
+        dlcDAO.create(dlcDb)
```
nkohen: nit: move the val out of the `for` to avoid the need for `{}` here

nkohen: same a couple lines below

wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      }
+      _ <- {
+        val offerDb = DLCOfferDb.fromDLCOffer(offer, RegTest)
+        offerDAO.create(offerDb)
+      }
+
+      readOffer <- offerDAO.read(eventId)
+    } yield {
+      assert(
+        readOffer.get.toDLCOffer(offer.fundingInputs).toJson == offer.toJson)
```
nkohen: is there a reason for the `toJson` calls?

wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  it should "correctly insert a DLCAccept into the database" in { daos =>
```
nkohen, line 57: the `it should "correctly insert a *** into the database` tests seem to follow a similar pattern and I think that we can factor it out to being a function that takes in an element, a DAO capabale of writing and reading that kind of element, and returning a `Future[Assertion]`

wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
```diff
@@ -0,0 +1,199 @@
...
+          isInitiator = true,
+          outPoint = TransactionOutPoint(testBlockHash, UInt32(3)),
+          output = TransactionOutput(Satoshis.one, EmptyScriptPubKey),
+          sigs = Vector(DLCWalletTestUtil.dummyPartialSig)
+        )
+      )
+
+      for {
+        _ <- {
+          val dlcDb = DLCDb(
```
nkohen: Since the `DLCDb`s in this file are the same across tests, it is probably worth throwing this val up in `DLCWalletTestUtil` with the rest of the vals. (Same might apply for offer/accept/funding inputs)

wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
benthecarman: Yeah, I'd have the exact same offers but the `==` would fail, so I changed it to `toJson`

benthecarman: fixed now

wallet-test/src/test/scala/org/bitcoins/wallet/models/DLCDAOTest.scala
```diff
@@ -0,0 +1,121 @@
...
+  it should "correctly insert a DLCAccept into the database" in { daos =>
+    val dlcDAO = daos.dlcDAO
+    val acceptDAO = daos.dlcAcceptDAO
+
+    val acceptDb = DLCAcceptDb.fromDLCAccept(DLCWalletTestUtil.sampleDLCAccept)
+
+    verifyDatabaseInsertion(acceptDb, acceptDAO, dlcDAO)
+  }
+
+  it should "correctly insert funding inputs into the database" in { daos =>
```
nkohen, line 66: any reason we couldn't use `verifyDatabaseInsertion` here?

PR #1170, opened 2020-02-26
Update scodec-bits dep to 1.14

Christewart:


Christewart:
It looks like scodec-bits has not been published for 2.11, they may stop supporting 2.11 soon.

PR #1174, opened 2020-02-27
Allow getbalance to return in sats

benthecarman:
This allows the user to add `--sats` to get their balance in sats, it now also will print the user's balance much prettier

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -91,8 +91,18 @@ object ConsoleCli {
         ),
       cmd("getbalance")
         .hidden()
-        .action((_, conf) => conf.copy(command = GetBalance))
-        .text("Get the wallet balance"),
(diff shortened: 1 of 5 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -91,8 +91,18 @@ object ConsoleCli {
         ),
       cmd("getbalance")
         .hidden()
-        .action((_, conf) => conf.copy(command = GetBalance))
-        .text("Get the wallet balance"),
+        .action((_, conf) => conf.copy(command = GetBalance(false)))
+        .text("Get the wallet balance")
```
Christewart, line 95: nit: We should probably start adding config flags in the help message, which is what the text message is I'm guessing?

Christewart (approved):

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: I think that is out of the scope of this PR but I will make an issue for it.

PR #1175, opened 2020-02-27
Wallet DLC Negotiation Tests

benthecarman:
Built on top of #1169

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
@@ -3,9 +3,12 @@ package org.bitcoins.testkit.wallet
 import org.bitcoins.core.crypto._
 import org.bitcoins.core.currency._
 import org.bitcoins.core.hd.{BIP32Path, HDAccount}
+import org.bitcoins.core.number.UInt32
+import org.bitcoins.core.protocol.script.EmptyScriptPubKey
 import org.bitcoins.core.protocol.transaction.{
-  EmptyOutputReference,
-  EmptyTransactionOutPoint
+  OutputReference,
+  TransactionOutPoint,
+  TransactionOutput
 }
 import org.bitcoins.core.protocol.{BitcoinAddress, BlockStampWithFuture}
 import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
@@ -16,8 +19,7 @@ import org.bitcoins.dlc._
 import org.bitcoins.wallet.models.DLCDb
 import scodec.bits.ByteVector
 
-object DLCWalletTestUtil {
-
+object DLCWalletUtil {
   lazy val oraclePrivKey: ECPrivateKey = ECPrivateKey.freshPrivateKey
   lazy val nonce: SchnorrNonce = SchnorrNonce.freshNonce
   lazy val rValue: ECPublicKey = nonce.publicKey
@@ -49,6 +51,8 @@ object DLCWalletTestUtil {
 
   lazy val dummyKey: ECPublicKey = ECPublicKey.freshPublicKey
 
+  lazy val dummyKey2: ECPublicKey = ECPublicKey.freshPublicKey
+
   lazy val dummyPartialSig: PartialSignature =
     PartialSignature(dummyKey, DummyECDigitalSignature)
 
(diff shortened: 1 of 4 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.scalatest.FutureOutcome
+
+class DLCNegotiationTest extends BitcoinSWalletTest {
+  type FixtureParam = FundedWallet
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withFundedSegwitWallet(test)
+  }
+
+  behavior of "DLCNegotiation"
```
nkohen: Should this not be `DLCWallet`? I'm not able to parse `"DLCNegotiation must create a dlc offer"`

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
```diff
@@ -0,0 +1,101 @@
+package org.bitcoins.wallet.dlc
+
+import org.bitcoins.dlc.DLCMessage
+import org.bitcoins.testkit.wallet.FundWalletUtil.FundedWallet
+import org.bitcoins.testkit.wallet.{BitcoinSWalletTest, DLCWalletUtil}
+import org.scalatest.FutureOutcome
+
+class DLCNegotiationTest extends BitcoinSWalletTest {
```
nkohen: nit: `WalletDLCSetupTest` is probably a better descriptor

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  behavior of "DLCNegotiation"
+
+  it must "create a dlc offer" in { funded: FundedWallet =>
+    val fundedWallet = funded.wallet
+
+    val offerData = DLCWalletUtil.sampleDLCOffer
+
+    for {
+      offer <- fundedWallet.createDLCOffer(
```
nkohen: might be worthwhile to create an overloaded version of `createDLCOffer` that takes in a `DLCOffer`

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
```diff
@@ -0,0 +1,101 @@
...
+        offerData.totalCollateral,
+        Some(offerData.feeRate),
+        offerData.timeouts.contractMaturity.toUInt32,
+        offerData.timeouts.contractTimeout.toUInt32
+      ) // must initialize the dlc in the database
+
+      sign <- fundedWallet.signDLC(accept)
+    } yield {
+      assert(sign.eventId == accept.eventId)
+      assert(sign.fundingSigs.nonEmpty)
```
nkohen: can we validate these sigs or is there a catch?

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  it must "add dlc sigs" in { funded: FundedWallet =>
+    val fundedWallet = funded.wallet
+
+    val offer = DLCWalletUtil.sampleDLCOffer
+    val sigs = DLCWalletUtil.sampleDLCSign
+
+    for {
+      _ <- fundedWallet.acceptDLCOffer(offer) // must initialize the dlc in the database
```
nkohen: We might want to consider making `initDLC` package private rather than fully private to avoid this hack? I guess I don't have strong feelings about it but it seems like the cleaner option

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
benthecarman: We don't have the tx they are signing

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCNegotiationTest.scala
benthecarman: Doesn't really solve the problem because we need the funding inputs in the database or some of the offer data

nkohen (approved):

PR #1176, opened 2020-02-28
DLC Wallet Execution Tests

benthecarman:
Built on top of #1175

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
@@ -1,0 +1,65 @@
+package org.bitcoins.testkit.wallet
+
+import org.bitcoins.db.AppConfig
+import org.bitcoins.server.BitcoinSAppConfig
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.wallet.DLCWalletUtil.InitializedDLCWallet
+import org.bitcoins.testkit.wallet.FundWalletUtil.FundedWallet
+import org.bitcoins.wallet.config.WalletAppConfig
+import org.scalatest.FutureOutcome
+
+trait BitcoinSDualWalletTest extends BitcoinSWalletTest {
+  import BitcoinSWalletTest._
+
+  implicit protected def config2: BitcoinSAppConfig =
+    BitcoinSTestAppConfig.getSpvTestConfig()
+
+  implicit protected def wallet2AppConfig: WalletAppConfig = {
+    config2.walletConf
+  }
+
+  override def beforeAll(): Unit = {
+    AppConfig.throwIfDefaultDatadir(config.walletConf)
+    AppConfig.throwIfDefaultDatadir(config2.walletConf)
+  }
+
+  def withDualFundedSegwitWallets(test: OneArgAsyncTest): FutureOutcome = {
+    makeDependentFixture(
+      build = () =>
+        for {
+          walletA <- FundWalletUtil.createFundedWallet(nodeApi,
+                                                       chainQueryApi,
+                                                       Some(segwitWalletConf))
+          walletB <- FundWalletUtil.createFundedWallet(
+            nodeApi,
+            chainQueryApi,
+            Some(segwitWalletConf))(config2, system)
+        } yield (walletA, walletB),
+      destroy = { fundedWallets: (FundedWallet, FundedWallet) =>
+        destroyWallet(fundedWallets._1.wallet)
+        destroyWallet(fundedWallets._2.wallet)
+      }
+    )(test)
+  }
+
+  def withDualDLCWallets(test: OneArgAsyncTest): FutureOutcome = {
+    makeDependentFixture(
+      build = () =>
+        for {
+          walletA <- FundWalletUtil.createFundedWallet(nodeApi,
+                                                       chainQueryApi,
+                                                       Some(segwitWalletConf))
+          walletB <- FundWalletUtil.createFundedWallet(
+            nodeApi,
+            chainQueryApi,
+            Some(segwitWalletConf))(config2, system)
+          (dlcWalletA, dlcWalletB) <- DLCWalletUtil.createDLCWallets(walletA,
+                                                                     walletB)
+        } yield (dlcWalletA, dlcWalletB),
+      destroy = { dlcWallets: (InitializedDLCWallet, InitializedDLCWallet) =>
+        destroyWallet(dlcWallets._1.wallet)
+        destroyWallet(dlcWallets._2.wallet)
+      }
+    )(test)
+  }
+}
--- a/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
@@ -21,6 +21,7 @@ trait BitcoinSDualWalletTest extends BitcoinSWalletTest {
   override def beforeAll(): Unit = {
     AppConfig.throwIfDefaultDatadir(config.walletConf)
     AppConfig.throwIfDefaultDatadir(config2.walletConf)
+    super.beforeAll()
   }
 
   /** Creates two segwit wallets that are funded with some bitcoin, these wallets are NOT
--- a/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
@@ -23,6 +23,9 @@ trait BitcoinSDualWalletTest extends BitcoinSWalletTest {
     AppConfig.throwIfDefaultDatadir(config2.walletConf)
   }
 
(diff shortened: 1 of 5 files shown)


testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.bitcoins.db.AppConfig
+import org.bitcoins.server.BitcoinSAppConfig
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.wallet.DLCWalletUtil.InitializedDLCWallet
+import org.bitcoins.testkit.wallet.FundWalletUtil.FundedWallet
+import org.bitcoins.wallet.config.WalletAppConfig
+import org.scalatest.FutureOutcome
+
+trait BitcoinSDualWalletTest extends BitcoinSWalletTest {
+  import BitcoinSWalletTest._
```
nkohen, line 12: is the `import` necessary given the extension?

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
```diff
@@ -0,0 +1,65 @@
...
+  implicit protected def wallet2AppConfig: WalletAppConfig = {
+    config2.walletConf
+  }
+
+  override def beforeAll(): Unit = {
+    AppConfig.throwIfDefaultDatadir(config.walletConf)
+    AppConfig.throwIfDefaultDatadir(config2.walletConf)
+  }
+
+  def withDualFundedSegwitWallets(test: OneArgAsyncTest): FutureOutcome = {
```
nkohen, line 30: scaladoc

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
```diff
@@ -0,0 +1,65 @@
...
+            Some(segwitWalletConf))(config2, system)
+        } yield (walletA, walletB),
+      destroy = { fundedWallets: (FundedWallet, FundedWallet) =>
+        destroyWallet(fundedWallets._1.wallet)
+        destroyWallet(fundedWallets._2.wallet)
+      }
+    )(test)
+  }
+
+  def withDualDLCWallets(test: OneArgAsyncTest): FutureOutcome = {
```
nkohen, line 50: scaladoc

testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
```diff
@@ -101,4 +107,45 @@ object DLCWalletUtil {
 
   lazy val sampleDLCSign: DLCSign =
     DLCSign(dummyCETSigs, dummyFundingSignatures, sampleDLCEventId)
+
+  def initDLC(fundedWalletA: FundedWallet, fundedWalletB: FundedWallet)(
```
nkohen, line 125: `setupDLC` is probably a better name as we use `initDLC` to mean something else internally. Also a scaladoc would be nice here

testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
```diff
@@ -101,4 +107,45 @@ object DLCWalletUtil {
...
+
+object DLCWalletUtil extends DLCWalletUtil {
+
+  case class InitializedDLCWallet(funded: FundedWallet) {
+    val wallet: Wallet = funded.wallet
+  }
+
+  def createDLCWallets(
+      fundedWalletA: FundedWallet,
+      fundedWalletB: FundedWallet)(implicit system: ActorSystem): Future[
```
nkohen: From what I see here you probably want an implicit `ExecutionContext` rather than a full-fledged `ActorSystem`

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCExecutionTest.scala
```diff
@@ -0,0 +1,214 @@
...
+import org.scalatest.FutureOutcome
+
+class DLCExecutionTest extends BitcoinSDualWalletTest {
+  type FixtureParam = (InitializedDLCWallet, InitializedDLCWallet)
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withDualDLCWallets(test)
+  }
+
+  behavior of "DLCExecution"
```
nkohen: should this be `DLCWallet`?

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCExecutionTest.scala
```diff
@@ -0,0 +1,214 @@
...
+        (cet, closeTxOpt) <- dlcA.executeDLCForceClose(eventId, winSig)
+
+      } yield {
+        assert(closeTxOpt.isDefined)
+        val closeTx = closeTxOpt.get
+
+        assert(cet.inputs.size == 1)
+        assert(cet.outputs.size == 1)
+        assert(ScriptInterpreter.checkTransaction(cet))
+        assert(ScriptInterpreter.checkTransaction(closeTx))
```
nkohen: Is there some reason that here (and with every call to `checkTransaction` we can't use `ScriptInterpreter.runVerify` which would actually verify sigs and such? I know that this requires more than just the tx but I imagine that we would be able to get the extra info in at least some of these cases?

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
benthecarman: Yes, I am not sure why though, the `destroyWallet` function will not be defined

nkohen: Oh I get it, it is in the companion object which is what is being imported here

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
```diff
@@ -0,0 +1,69 @@
...
+    BitcoinSTestAppConfig.getSpvTestConfig()
+
+  implicit protected def wallet2AppConfig: WalletAppConfig = {
+    config2.walletConf
+  }
+
+  override def beforeAll(): Unit = {
+    AppConfig.throwIfDefaultDatadir(config.walletConf)
+    AppConfig.throwIfDefaultDatadir(config2.walletConf)
+  }
```
nkohen: don't forget to end with a call to `super.beforeAll`

testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
```diff
@@ -116,4 +121,44 @@ object DLCWalletUtil {
...
+    val sampleOfferData = sampleDLCOffer
+    for {
+      offer <- walletA.createDLCOffer(
+        sampleOfferData.oracleInfo,
+        sampleOfferData.contractInfo,
+        Satoshis(6000),
+        Some(sampleOfferData.feeRate),
+        sampleOfferData.timeouts.contractMaturity.toUInt32,
+        sampleOfferData.timeouts.contractTimeout.toUInt32
+      )
```
nkohen: Is there a reason not to just pass in `sampleDLCOffer`?

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCExecutionTest.scala
```diff
@@ -0,0 +1,326 @@
...
+        ScriptInterpreter.runVerify(
+          PreExecutionScriptProgram(closeSigComponent)))
+    }
+  }
+
+  it must "do a unilateral close as the recipient" in { wallets =>
+    val dlcB = wallets._2.wallet
+
+    for {
+      (cet, closeTxOpt) <- dlcB.executeDLCForceClose(eventId, loseSig)
```
nkohen: These tests are all almost the same, it would be nice to have the pattern factored out and turn these tests into two or three liners

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCExecutionTest.scala
```diff
@@ -0,0 +1,272 @@
...
+          val output = fundingTxPrevOutputRefs
+            .find(_.outPoint == input.previousOutput)
+            .get
+            .output
+          val sigComponent =
+            WitnessTxSigComponent(fundingTx.asInstanceOf[WitnessTransaction],
+                                  UInt32(index),
+                                  output,
+                                  flags)
+          ScriptInterpreter.runVerify(PreExecutionScriptProgram(sigComponent))
```
nkohen: nit: This pattern is widely used, I think it might be nice to make a def that takes a tx, index, and output and does the rest. Might even be worth adding to `ScriptInterpreter` or at the very least the top of this file

nkohen (commented):
just one last thing

nkohen (approved):

PR #1178, opened 2020-02-28
Update TxoState for transactions after they've been confirmed

### LINKED ISSUE #1083: Make number of confirmations the wallet requires configurable
This should be in the [`bitcoin-s.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/resources/reference.conf#L57) file. 

``` 
bitcoin-s { 
  wallet { 
    confirmations = 6 //or whatever number 
  } 
}
```

This setting should be used to determine when utxos in the `SpendingInfoTable` should be transititioned from [TxoStates](https://github.com/bitcoin-s/bitcoin-s/blob/1ad402ea56718052dac4ea260c4bf8f7fbff8c55/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala#L10). For instance, from `PendingConfirmationsReceived` -> `ConfirmedReceived`. 

The `TxoState` in `SpendingInfoDb` should be updated here in [processExistingIncomingTxo](https://github.com/bitcoin-s/bitcoin-s/blob/1ad402ea56718052dac4ea260c4bf8f7fbff8c55/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala#L241). If we have more confirmations than required, we should transition the utxo to the next `TxoState` if applicable.

benthecarman:
Closes #1083 

Built on top of #1206

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -162,12 +162,18 @@ object Main extends App {
     lazy val onBlock: OnBlockReceived = { block =>
       wallet.processBlock(block).map(_ => ())
     }
+    lazy val onHeaders: OnBlockHeadersReceived = { headers =>
+      wallet.updateTxoStatesFromHeader(headers.last).map(_ => ())
+    }
     if (nodeConf.isSPVEnabled) {
-      Future.successful(NodeCallbacks(onTxReceived = Seq(onTx)))
+      Future.successful(
+        NodeCallbacks(onTxReceived = Seq(onTx),
+                      onBlockHeadersReceived = Seq(onHeaders)))
     } else if (nodeConf.isNeutrinoEnabled) {
       Future.successful(
         NodeCallbacks(onBlockReceived = Seq(onBlock),
-                      onCompactFilterReceived = Seq(onCompactFilter)))
+                      onCompactFilterReceived = Seq(onCompactFilter),
+                      onBlockHeadersReceived = Seq(onHeaders)))
     } else {
       Future.failed(new RuntimeException("Unexpected node type"))
     }
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -163,7 +163,7 @@ object Main extends App {
       wallet.processBlock(block).map(_ => ())
     }
     lazy val onHeaders: OnBlockHeadersReceived = { headers =>
-      wallet.updateTxoStatesFromHeader(headers.last).map(_ => ())
+      wallet.updateUtxoPendingStates(headers.last).map(_ => ())
     }
     if (nodeConf.isSPVEnabled) {
       Future.successful(
--- a/node/src/main/scala/org/bitcoins/node/NodeCallbacks.scala
+++ b/node/src/main/scala/org/bitcoins/node/NodeCallbacks.scala
@@ -2,7 +2,7 @@ package org.bitcoins.node
 
 import org.bitcoins.core.crypto.DoubleSha256Digest
 import org.bitcoins.core.gcs.GolombFilter
-import org.bitcoins.core.protocol.blockchain.{Block, MerkleBlock}
+import org.bitcoins.core.protocol.blockchain.{Block, BlockHeader, MerkleBlock}
 import org.bitcoins.core.protocol.transaction.Transaction
 import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.db.MarkedLogger
@@ -20,14 +20,16 @@ case class NodeCallbacks(
     onCompactFilterReceived: Seq[OnCompactFilterReceived] = Seq.empty,
     onTxReceived: Seq[OnTxReceived] = Seq.empty,
     onBlockReceived: Seq[OnBlockReceived] = Seq.empty,
-    onMerkleBlockReceived: Seq[OnMerkleBlockReceived] = Seq.empty
+    onMerkleBlockReceived: Seq[OnMerkleBlockReceived] = Seq.empty,
+    onBlockHeadersReceived: Seq[OnBlockHeadersReceived] = Seq.empty
 ) {
 
   def +(other: NodeCallbacks): NodeCallbacks = copy(
     onCompactFilterReceived = onCompactFilterReceived ++ other.onCompactFilterReceived,
     onTxReceived = onTxReceived ++ other.onTxReceived,
     onBlockReceived = onBlockReceived ++ other.onBlockReceived,
-    onMerkleBlockReceived = onMerkleBlockReceived ++ other.onMerkleBlockReceived
+    onMerkleBlockReceived = onMerkleBlockReceived ++ other.onMerkleBlockReceived,
+    onBlockHeadersReceived = onBlockHeadersReceived ++ other.onBlockHeadersReceived
   )
 
   def executeOnTxReceivedCallbacks(logger: MarkedLogger, tx: Transaction)(
@@ -81,6 +83,21 @@ case class NodeCallbacks(
                 err)
           }))
   }
+
+  def executeOnBlockHeadersReceivedCallbacks(
+      logger: MarkedLogger,
+      headers: Vector[BlockHeader])(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    onBlockHeadersReceived
+      .foldLeft(FutureUtil.unit)((acc, callback) =>
+        acc.flatMap(_ =>
+          callback(headers).recover {
+            case err: Throwable =>
+              logger.error(
+                "onBlockHeadersReceived Callback failed with error: ",
+                err)
+          }))
+  }
 }
 
 object NodeCallbacks {
@@ -101,12 +118,17 @@ object NodeCallbacks {
   def onCompactFilterReceived(f: OnCompactFilterReceived): NodeCallbacks =
     NodeCallbacks(onCompactFilterReceived = Seq(f))
 
+  /** Constructs a set of callbacks that only acts on block headers received */
+  def onBlockHeadersReceived(f: OnBlockHeadersReceived): NodeCallbacks =
+    NodeCallbacks(onBlockHeadersReceived = Seq(f))
+
   /** Empty callbacks that does nothing with the received data */
   val empty: NodeCallbacks =
     NodeCallbacks(
       onTxReceived = Seq.empty,
       onBlockReceived = Seq.empty,
       onMerkleBlockReceived = Seq.empty,
-      onCompactFilterReceived = Seq.empty
+      onCompactFilterReceived = Seq.empty,
+      onBlockHeadersReceived = Seq.empty
     )
 }
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -194,6 +194,7 @@ case class DataMessageHandler(
         for {
           newApi <- chainApiF
           newSyncing <- getHeadersF
+          _ <- callbacks.executeOnBlockHeadersReceivedCallbacks(logger, headers)
         } yield {
           this.copy(chainApi = newApi, syncing = newSyncing)
         }
--- a/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
@@ -5,8 +5,13 @@ import org.bitcoins.core.bloom.{BloomFilter, BloomUpdateAll}
 import org.bitcoins.core.crypto._
 import org.bitcoins.core.currency._
 import org.bitcoins.core.hd.HDAccount
+import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.protocol.transaction.TransactionOutPoint
 import org.bitcoins.core.wallet.utxo.TxoState
+import org.bitcoins.core.wallet.utxo.TxoState.{
+  ConfirmedReceived,
+  PendingConfirmationsReceived
+}
 import org.bitcoins.wallet.api._
 import org.bitcoins.wallet.config.WalletAppConfig
 import org.bitcoins.wallet.internal._
@@ -71,7 +76,7 @@ abstract class LockedWallet
   }
 
   override def getConfirmedBalance(): Future[CurrencyUnit] = {
-    val confirmed = filterThenSum(_.blockHash.isDefined)
+    val confirmed = filterThenSum(_.state == ConfirmedReceived)
     confirmed.foreach(balance =>
       logger.trace(s"Confirmed balance=${balance.satoshis}"))
     confirmed
@@ -92,7 +97,7 @@ abstract class LockedWallet
   }
 
   override def getUnconfirmedBalance(): Future[CurrencyUnit] = {
-    val unconfirmed = filterThenSum(_.blockHash.isEmpty)
+    val unconfirmed = filterThenSum(_.state == PendingConfirmationsReceived)
     unconfirmed.foreach(balance =>
       logger.trace(s"Unconfirmed balance=${balance.satoshis}"))
     unconfirmed
@@ -161,6 +166,18 @@ abstract class LockedWallet
       utxos: Vector[SpendingInfoDb]): Future[Vector[SpendingInfoDb]] = {
     val updated = utxos.map(_.copyWithState(TxoState.Reserved))
     spendingInfoDAO.updateAll(updated)
+  }
+
+  override def updateTxoStatesFromHeader(
+      blockHeader: BlockHeader): Future[Vector[SpendingInfoDb]] = {
+    for {
+      infos <- spendingInfoDAO.findAllPendingConfirmation
+      updatedInfos <- {
+        val updatedInfoFs =
+          infos.map(info => updateUtxoConfirmedState(info, blockHeader.hashBE))
+        Future.sequence(updatedInfoFs)
+      }
+    } yield updatedInfos
   }
 }
 
--- a/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
@@ -168,7 +168,8 @@ abstract class LockedWallet
     spendingInfoDAO.updateAll(updated)
   }
 
-  override def updateTxoStatesFromHeader(
+  /** @inheritdoc */
+  def updateUtxoPendingStates(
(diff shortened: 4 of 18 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -268,7 +268,22 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
               s"Failed to update confirmation count of transaction=${transaction.txIdBE}",
               err))
 
-          updateF
+          updateF.flatMap { updated =>
```
Christewart: So this is where we can get a lot of leverage out of using ADTs. With how the current code is written you are using `if()` which doesn't give you any compiler guarantees about handling states correctly, _and_ doesn't give us any warnings if someday we need to add more types to `TxoState`. Here is a more functional and scalable approach to this:

```scala
          for {
            updated <- updateF
            confsOpt <- chainQueryApi.getNumberOfConfirmations(blockHash)
            stateChange <- {
              confsOpt match {
                case None => Future.successful(updated)
                case Some(confs) =>
                  if (confs >= walletConfig.requiredConfirmations) {
                    updated.state match {
                      case pending @ (TxoState.PendingConfirmationsReceived | TxoState.PendingConfirmationsSpent) =>
                        spendingInfoDAO.update(
                          updated.copyWithState(pending))
                      case TxoState.DoesNotExist =>
                        // ??? corrupt chain ???
                        ???
                      case TxoState.ConfirmedReceived | TxoState.ConfirmedSpent =>
                        // ??? reorg ???
                        ???
                    }
                  } else {
                    //verify that the state is at least Pending ??
                    ???
                  }
              }
            }
          } yield stateChange
```

With ADTs we can more accurately model what states are actually possible, if we decide to use them, and it gives us a clear framework for us to think in for what we possibly need to handle. When in doubt about ADTs, [listen to Jane Street](https://youtu.be/FnBPECrSC7o?t=1879)! 

Honestly, even my code is a little ugly. We should probably break out the pattern match into it's own helper method.

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -61,6 +61,9 @@ case class WalletAppConfig(
 
   lazy val discoveryBatchSize: Int = config.getInt("wallet.discoveryBatchSize")
 
+  lazy val requiredConfirmations: Int =
+    config.getInt("wallet.requiredConfirmations")
```
Christewart, line 65: require that this isn't `0`?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: Also it looks like on my first iteration of writing this code i forgot a state... just to show the power of ADTs I didn't consider `Reserved`! Doh. 

>[warn] /home/suredbits/dev/bitcoin-s/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala:279:29: match may not be exhaustive.
[warn] It would fail on the following input: Reserved
[warn]                     updated.state match {

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
Christewart: at least it shouldn't be negative, that's for sure

benthecarman:
The updating of the state after block headers wasn't as big as I expected so I added it to this PR. Should be good for review now

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -165,12 +165,19 @@ object Main extends App {
       wallet.processBlock(block)
       ()
     }
+    lazy val onHeaders: OnBlockHeadersReceived = { headers =>
+      wallet.updateTxoStatesFromHeader(headers.last)
```
Christewart: Why just the last header? Shouldn't this iterate through every header? I.e.

```scala
headers.foreach { h => 
  wallet.updateTxoStatesFromHeader(h)
}
```

I'm also worried there is going to be async problems here, as the `OnBockHeadersReceived` alias is a `Vector[BlockHeader] => Unit` while `updateTxoSatesFromHeader()` returns a `Future`, so this isn't async safe.

node/src/main/scala/org/bitcoins/node/NodeCallbacks.scala
```diff
@@ -41,12 +43,17 @@ object NodeCallbacks {
   def onCompactFilterReceived(f: OnCompactFilterReceived): NodeCallbacks =
     NodeCallbacks(onCompactFilterReceived = Seq(f))
 
+  /** Constructs a set of callbacks that only acts on compact filter received */
```
Christewart: is this comment wrong?

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -74,6 +74,9 @@ trait LockedWalletApi extends WalletApi {
     }
   }
 
+  def updateTxoStatesFromHeader(
```
Christewart: scaladoc

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -150,6 +151,18 @@ abstract class LockedWallet
     val updated = utxos.map(_.copyWithState(TxoState.Reserved))
     spendingInfoDAO.updateAll(updated)
   }
+
+  override def updateTxoStatesFromHeader(
```
Christewart: So this seems a little confusing to me, from looking at the method name, it seems that this would update all relevant utxos in the db, while looking at the implementation it only updates pending ones.

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -39,6 +39,48 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
+                }
+              case TxoState.PendingConfirmationsSpent =>
+                if (confs >= walletConfig.requiredConfirmations) {
+                  spendingInfoDAO.update(
+                    txo.copyWithState(TxoState.ConfirmedSpent))
+                } else {
+                  Future.successful(txo)
+                }
+              case TxoState.Reserved =>
+                // We should keep the utxo as reserved so it is not used in
```
Christewart, line 74: but doesn't this really mean we have seen a utxo spent on the network that we thought we had reserved? Isn't this bad? Worse case couldn't this be a hack, best case somehow how we are out of sync?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -39,6 +39,48 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
+                // a future transaction that it should not be in
+                Future.successful(txo)
+              case TxoState.ConfirmedReceived | TxoState.ConfirmedSpent =>
+                val errMsg =
+                  s"TXO ${txo.txid} was marked as confirmed but did not have a block hash, the issue has been fixed"
```
Christewart: how is this fixed? I don't see it getting updated with the hash? Also isn't this technically the wrong hash? Is this why we aren't calling `update`? Couldn't this be a reorg?

Christewart (changes requested):
This seems light on unit tests

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: Yeah, you're right, this should be changed

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
benthecarman: The pending ones are the only relevant ones here, the only thing we can get from a header is a new chain tip, which means txs only care if they now have more confirmations.

app/server/src/main/scala/org/bitcoins/server/Main.scala
benthecarman: Since we will only be updating based off height it only makes sense to grab the header with the largest height.

As for the async issues, the others do the same. I think it is okay because it is just something called for the wallet, the `Unit` return is a safety measure so the node doesn't receive any data that it doesn't want/need (at least that's what it seems like to me)

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: Yes you are right, however, it should not change if we are just receiving block headers. Will keep it the same here but need to change it in `processTransaction`

app/server/src/main/scala/org/bitcoins/server/Main.scala
Christewart: >Since we will only be updating based off height it only makes sense to grab the header with the largest height.

First off, this is an implementation detail that I definitely don't think we should rely on long term (reorgs, heaviest POW chain etc). 

Also do  we have any sort of guarantee that our internal processing of headers in the chain project is finished before this callback is called? If I remember correctly, for neutrino to update a transaction we need to: 

1. Get the block header
2. Get the compact filter header
3. Get the compact filter
4. See if any of our utxos match the filter 
5. Fetch the full block
6. Scan the full block for utxos
7. Update the the `SpendingInfoTable`

Also it seems like just a bad logical break to be ignoring _all_ headers we see except the last one. Our wallet should be built to update when we receive a single block, not a `Vector`. It's too hard to reason about when we start thinking about updating receiving sequences of blocks at the same time. We should iterate through each of those single blocks and apply the appropriate updates _for that single block_ to the wallet, and then move onto the next one. This callback breaks that. 

>As for the async issues, the others do the same. I think it is okay because it is just something called for the wallet, the Unit return is a safety measure so the node doesn't receive any data that it doesn't want/need (at least that's what it seems like to me)

Yeah, I don't like how we handle that to be honest though as it isn't really safe to use for exactly this reason. How do you know if your callback is done processing? We use async code everywhere internally.

rorp: Yes. We'll need to change callbacks to return types to `Future[]`, and use the returned values from call backs in `for`/`flatMap` statements to prevent concurrency issues.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -196,6 +196,8 @@ case class DataMessageHandler(
           logger.error(s"Error when processing headers message", err)
         }
 
+        callbacks.onBlockHeadersReceived.foreach(_.apply(headers))
```
rorp: This line should be changed to something like:
```suggestion
for {
  _ <- callbacks.onBlockHeadersReceived.foldLeft(FutureUtil.unit)(
              (acc, callback) => acc.flatMap(_ => callback(headers)))
  newApi <- chainApiF
...
```

app/server/src/main/scala/org/bitcoins/server/Main.scala
rorp: Also, `headers` can be empty. `bitcoind` returns an empty `headers` message when a fully synced node send `getheaders` command.

rorp (changes requested):

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -101,8 +102,9 @@ case class DataMessageHandler(
           Try {
             val blockFilter =
               BlockFilter.fromBytes(filter.filterBytes, filter.blockHash)
-            callbacks.onCompactFilterReceived.foreach(
-              _.apply(filter.blockHash, blockFilter))
+            callbacks.onCompactFilterReceived.foldLeft(FutureUtil.unit)(
```
rorp: I think we need all types of callback to be called synchronously, i.e. plugged into the `for` statements, like you did in line 202 for `onBlockHeadersReceived`.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -206,7 +210,8 @@ case class DataMessageHandler(
         logger.info(
           s"Received block message with hash ${msg.block.blockHeader.hash.flip}")
         Future {
-          callbacks.onBlockReceived.foreach(_.apply(msg.block))
+          callbacks.onBlockReceived.foldLeft(FutureUtil.unit)((acc, callback) =>
```
rorp: No need to wrap it in `Future` here, `foldLeft` returns a `Future` now.
```suggestion
          callbacks.onBlockReceived.foldLeft(FutureUtil.unit)((acc, callback) => acc.flatMap(_ => callback(msg.block))).map(_ => this)
```

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -220,7 +225,8 @@ case class DataMessageHandler(
           logger.trace(
             s"Transaction=${tx.txIdBE} does not belong to merkleblock, processing given callbacks")
           Future {
-            callbacks.onTxReceived.foreach(_.apply(tx))
+            callbacks.onTxReceived.foldLeft(FutureUtil.unit)((acc, callback) =>
```
rorp: Same here:
```suggestion
            callbacks.onTxReceived.foldLeft(FutureUtil.unit)((acc, callback) => acc.flatMap(_ => callback(tx))).map(_ => this)
```

rorp (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -268,7 +268,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+            confsOpt <- chainQueryApi.getNumberOfConfirmations(blockHash)
+            stateChange <- {
+              confsOpt match {
+                case None => Future.successful(updated)
+                case Some(confs) =>
+                  updated.state match {
+                    case TxoState.PendingConfirmationsReceived |
+                        TxoState.DoesNotExist =>
+                      if (confs >= walletConfig.requiredConfirmations) {
+                        spendingInfoDAO.update(
```
rorp: I think we should update both block hash and state in one database call.

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -39,6 +39,44 @@ private[wallet] trait UtxoHandling extends WalletLogger {
     spendingInfoDAO.findAllUnspentForAccount(hdAccount)
   }
 
+  protected def updateUtxoConfirmedState(
+      txo: SpendingInfoDb,
+      blockHash: DoubleSha256DigestBE): Future[SpendingInfoDb] = {
+    for {
+      confsOpt <- chainQueryApi.getNumberOfConfirmations(blockHash)
```
rorp, line 50: It looks very similar to the change in TransactionProcessing.scala. DRY it up, please.

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala
```diff
@@ -126,6 +126,16 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** Updates the `blockHash` field */
   def copyWithBlockHash(blockHash: DoubleSha256DigestBE): SpendingInfoType
 
+  state match {
+    case TxoState.ConfirmedSpent | TxoState.ConfirmedReceived =>
+      require(blockHash.isDefined,
+              "Transaction cannot be confirmed without a blockHash")
+    case TxoState.PendingConfirmationsSpent |
```
rorp: nit: `case _:TxoState => ()`

rorp (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
Christewart: From looking at the method name/documentation this isn't apparent. Can you please rename it and add a scaladoc. I.e. `updateUtxoPendingStates()`

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -284,15 +284,25 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+            s"Updating block_hash of txo=${transaction.txIdBE}, new block hash=$blockHash")
+          // update block hash
+          val txoWithHash = foundTxo.copyWithBlockHash(blockHash = blockHash)
+
+          // If the utxo was marked reserved we want to update it to spent now
+          // since it has been included in a block
+          val unreservedTxo = txoWithHash.state match {
+            case TxoState.Reserved =>
+              txoWithHash.copyWithState(TxoState.PendingConfirmationsSpent)
+            case _: TxoState =>
```
Christewart: it's generally bad practice to do `_: TxoState` as if we add another type to `TxoState` we won't get a compiler warning telling us we need to update our code here. It's better to list ouf the possible states: 

```scala
case TxoState.ConfirmedSpent | TxoState.ConfirmedReceived | ... => 
```
So now if in the future we make a `ReservedColdStorage` (for example) we will get compiler warning telling us we need to update our code.

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -125,6 +125,18 @@ case class SpendingInfoDAO()(
     }
   }
 
+  private val pendingConfStates: Set[TxoState] =
```
Christewart: nit: Maybe move this into `TxoState` in case this is useful else where?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala
```diff
@@ -167,6 +167,14 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** Updates the `blockHash` field */
   def copyWithBlockHash(blockHash: DoubleSha256DigestBE): SpendingInfoType
 
+  state match {
```
Christewart: can you move this to the top? It's confusing to me to find it in the middle of method defintions. Also can you enumerate the `TxoState` again in the pattern match so we can get an exhaustiveness check by the compiler?

PR #1181, opened 2020-02-28
Merge CI matrices, add appServer/test

### LINKED ISSUE #1161: Reduce travis CI build times
In #710 we broke out a lot of travis stuff into their on CI matrix. Initially we did this because we would have a few spurious failures fail the entire build when there were only two CI matrixes being run. I.e. something would fail in the `nodeTest` and it would be hard to tell if your `coreTest` test cases passed. 

Now it seems that builds are taking a very long time, because travis seems to rate limit by the matrix rows being run. We should probably try to consolidate more rows to reduce build time. If we reduce the rows in the build matrix, we should also be able to save on VM startup overhead and jvm warmup time etc.

### LINKED ISSUE #1180: `appServerTest/test` not being run on Travis CI
Our current CI matrix on Travis does not include execute `sbt appServerTest/test` ... it should

benthecarman:
Closes #1161 
Closes #1180

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,9 @@ language: scala
 
 env:
   matrix:
-    - TEST_COMMAND="bitcoindRpcTest/test bitcoindRpc/coverageReport bitcoindRpc/coverageAggregate bitcoindRpc/coveralls"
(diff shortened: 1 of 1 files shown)


Christewart:
We should probably merge the mac osx ones as well

PR #1182, opened 2020-02-28
TxoState Life Cycle Tests

### LINKED ISSUE #1160: utxo life cycle tests
Add integration test in the wallet to make sure we can accurately track the life cycle of a utxo through our various `TxoState`s. 

I believe we don't have such a test already. 

Eventually this test should be expanded to handle things like reorgs to make sure that our txo states are modified correctly when there is a reorg.

benthecarman:
Closes #1160 

I needed to change `NodeFundedWalletBitcoind` to use a `Wallet` instead of `UnlockedWalletApi` so I could query the internal databases for the tests

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
@@ -1,0 +1,73 @@
+package org.bitcoins.wallet
+
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.core.protocol.script.EmptyScriptPubKey
+import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
+import org.bitcoins.core.wallet.utxo.TxoState
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest.WalletWithBitcoind
+import org.scalatest.FutureOutcome
+
+class UTXOLifeCycleTest extends BitcoinSWalletTest {
+
+  behavior of "Wallet Txo States"
+
+  override type FixtureParam = WalletWithBitcoind
+
+  val testAddr: BitcoinAddress =
+    BitcoinAddress.fromString("n4MN27Lk7Yh3pwfjCiAbRXtRVjs4Uk67fG").get
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withFundedWalletAndBitcoind(test)
+  }
+
+  it should "track a utxo state change to pending spent" in { param =>
+    val WalletWithBitcoind(wallet, _) = param
+
+    for {
+      tx <- wallet.sendToAddress(testAddr,
+                                 Satoshis(3000),
+                                 SatoshisPerVirtualByte(Satoshis(3)))
+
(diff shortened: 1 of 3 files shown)


Christewart (changes requested):
so this seems incomplete, as far as as I can tell these are the only state changes being tested: 

1. `ConfirmedReceived` -> `PendingSpent` (is this even valid? Should it be `ConfirmedReceived` -> `Reserved` -> `PendingSpent`?)
2. `DoesNotExist` -> `PendingConfirmationsReceived`
3. `ConfirmedReceieved` -> `Reserved`

This is some ossible state changes.

The obvious ones are around `PendingReceived/Spent` -> `ConfirmedReceived/Spent`. (Which probably relates to #1083)

There is also a possibility to have a state change from `PendingConfirmationsReceived` -> `PendingConfirmationsSpent` 

Am I missing any other possible valid state transitions?


Also I would like an integration test (if one doesn't exist, which I don't think it does) asserting the utxo's state as we are generating blocks from a bitcoind instance. Also we need to start simulating reorgs in our test cases, but maybe that is getting out of scope of this PR and should be filed as a separate issue.

benthecarman:
> ConfirmedReceived -> PendingSpent (is this even valid? Should it be ConfirmedReceived -> Reserved -> PendingSpent?)

My thinking was that `Reserved` isn't used for every tx, only for transactions that we explicitly want to save that utxo because we declared to the other party that we will spending from. However, I should be able to add a test for `Reserved` -> `PendingSpent` and `ConfirmedReceieved` -> `Reserved`

> The obvious ones are around PendingReceived/Spent -> ConfirmedReceived/Spent. (Which probably relates to #1083)

Yes, a lot of that requires that and more work, because #1083's changes only apply when we process the transaction which will likely be when we receive it's block. It does not update utxos to confirmed on receiving further blocks.

> Also I would like an integration test (if one doesn't exist, which I don't think it does) asserting the utxo's state as we are generating blocks from a bitcoind instance. Also we need to start simulating reorgs in our test cases, but maybe that is getting out of scope of this PR and should be filed as a separate issue.

Originally tried doing that but will require a lot more as `BitcoinSWalletTest` uses a dummy chain api. Figured it would be a good final thing to tackle after all the previously mentioned things were done.

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
```diff
@@ -0,0 +1,73 @@
...
+    BitcoinAddress.fromString("n4MN27Lk7Yh3pwfjCiAbRXtRVjs4Uk67fG").get
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withFundedWalletAndBitcoind(test)
+  }
+
+  it should "track a utxo state change to pending spent" in { param =>
+    val WalletWithBitcoind(wallet, _) = param
+
+    for {
```
Christewart, line 29: nit: So one problem I have with these tests is we aren't being specific enough. What utxo are we spending? What is the initial state (I'm guessing `ConfirmedReceived`)? These should both be asserted on. Maybe this is just being too pedantic about testing, so I'll let you be the judge if you want to address this or not.

Christewart (approved):

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
benthecarman: I had this originally, but think it is better as it is now because then we don't rely on `withFundedWalletAndBitcoind` to set it up so the txs have the correct confirmations as well as, these tests are just making sure it becomes the correct state, I think it makes more sense for it to be previous state agnostic

PR #1193, opened 2020-03-03
Track addresses that correspond to dlc pub keys

### LINKED ISSUE #1188: Make DLCWallet correctly track relevant addresses
Currently if we execute a DLC we will not be able to spend any funds we receive without a rescan because the wallet does not update the address or spending info table.

benthecarman:
Closes #1188

Allows for spending after a DLC.

Fixes a bug caused the accepter to not be using the keys they saved in their database

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
@@ -184,6 +184,63 @@ private[wallet] trait AddressHandling extends WalletLogger {
     addrF
   }
 
+  def getAddress(
+      account: AccountDb,
+      chainType: HDChainType,
+      addressIndex: Int): Future[AddressDb] = {
+
+    val coinType = account.hdAccount.coin.coinType
+    val accountIndex = account.hdAccount.index
+
+    val path = account.hdAccount.purpose match {
+      case HDPurposes.Legacy =>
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -184,6 +184,63 @@ private[wallet] trait AddressHandling extends WalletLogger {
...
+        NestedSegWitHDPath(coinType, accountIndex, chainType, addressIndex)
+      case HDPurposes.SegWit =>
+        SegWitHDPath(coinType, accountIndex, chainType, addressIndex)
+    }
+
+    val pathDiff =
+      account.hdAccount.diff(path) match {
+        case Some(value) => value
+        case None =>
+          throw new RuntimeException(
```
nkohen: nit: Should probably be an `IllegalArgumentException` since this depends only on inputs (I think)

nkohen (commented):
Looks good so far, would like the new stuff to get tested though

benthecarman:
> Looks good so far, would like the new stuff to get tested though

When I made this I originally did all the work on top of the dlc execution tests and then cherry picked it to the dlc branch since it doesn't require those tests.

I rebased this on top of the tests that we're committed

benthecarman:
Rebased on top of all DLC tests and looks like I got a 100% from CI :)

Issue #942, opened 2019-12-13
WalletIntegrationTest assert exact amount of satoshis left in wallet


Christewart:
Currently in our `WalletIntegrationTest` we are not asserting the exact amount of satoshis that left the wallet, rather we are checking that we are within a given amount. 

https://github.com/bitcoin-s/bitcoin-s/blob/a811f4a6d7bebd0a8b2351766ab1acb479455586/wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala#L116-L121

We should not use ranges when asserting tests on wallet balances as all wallet transfers should be deterministic and assertable.

PR #884, opened 2019-11-21
Add comments indicating what time unit eclair sends things in for the…

Christewart:
… various rpc calls

https://acinq.github.io/eclair/#introduction

PR #1197, opened 2020-03-06
Wallet Transaction Tracking

### LINKED ISSUE #1138: tx_table in wallet to store txs we are interested in
Create a `tx_table` that stores all transactions we are interested in. Some other tables, such as `txo_spending_info` should have foreign keys into this table to make sure we have the full tx. 

One reason we need this for is using PSBTs with `BaseTransaction`, we need to verify the entire transaction as the original signature serialization algorithm does not commit to the output amount. Therefore we need the _entire_ funding transaction as an input to the PSBT so we can re-calculate the hash and verify it matches the has in the spending transactions `TransactionOutPoint`.

benthecarman:
Closes #1138 

This will add to our wallet's database the ability to track and save relevant transactions

As highlighted in #1138 this is helpful for PSBTs with legacy transaction inputs.

This is also helpful for interactive protocols (like DLCs) so we don't need to construct every possible transaction every time.


benthecarman:
This has a silent merge conflict with #1236

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -55,6 +55,11 @@ case class SatoshisPerVirtualByte(currencyUnit: CurrencyUnit)
     extends BitcoinFeeUnit
 
 object SatoshisPerVirtualByte {
+
+  def apply(satoshis: Long): SatoshisPerVirtualByte = {
```
Christewart: nit: `.fromSatoshis()`/`.fromSats()` please

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -166,6 +167,10 @@ abstract class DbCommonsColumnMappers {
     MappedColumnType
       .base[TxoState, String](_.toString, TxoState.fromString(_).get)
   }
+
+  implicit val feeUnitMapper: BaseColumnType[FeeUnit] = {
```
Christewart: So it's important to be careful with these implicits, because that means we will treat _every_ fee unit as a `SatoshisPerVirtualByte()`, which is definitely what we don't want to do. Unfortunately the only way I know how to handle this is is to not use `implicit`, and just pass in the appropriate mapper where needed, similar to what I did in https://github.com/bitcoin-s/bitcoin-s/pull/884

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTransactionTable.scala
```diff
@@ -0,0 +1,134 @@
...
+
+/** Represents a relevant transaction for the wallet that we should be keeping track of */
+case class IncomingTransactionDb(
+    txIdBE: DoubleSha256DigestBE,
+    transaction: Transaction,
+    unsignedTxIdBE: DoubleSha256DigestBE,
+    unsignedTx: Transaction,
+    numInputs: Int,
+    numOutputs: Int,
+    lockTime: UInt32) {
```
Christewart: I think we should have `incomingAmount` or something like that as well. Also thoughts on adding `wtxid: Option[DoubleSha256DigestBE]` and and `tx_type` that can be `Base`, `SegwitV0` etc? Also do we want an `outputIndex`, so we know which output is ours at the database level? 

Finally, do we want to try and support multiple outputs paying to us in the same tx (batching)? I understand if we don't want to support this for now, but please file an issue if not.

Christewart: Also, with `IncomingTransactions` we don't really necessarily have access to the `unsignedTxId` do we? That is something that is only applicable for outgoing txs no?

EDIT: Actually thinking about this more, maybe this can be a malleabillity prevention mechanism for `BaseTransaction`? I don't know, we should discuss more. We need to be _REALLY_ careful about handling malleability with `BaseTransaction`.

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
```diff
@@ -0,0 +1,162 @@
...
+    txIdBE: DoubleSha256DigestBE,
+    transaction: Transaction,
+    unsignedTxIdBE: DoubleSha256DigestBE,
+    unsignedTx: Transaction,
+    numInputs: Int,
+    numOutputs: Int,
+    lockTime: UInt32,
+    txFee: CurrencyUnit,
+    expectedFeeRate: FeeUnit,
+    actualFeeRate: FeeUnit) {
```
Christewart: I think we should add `wtxid: Option[DoubleSha256DigestBE` and `tx_type` here, also I think we should have a field called `outgoingAmount` which indicates how much money is _leaving_ our wallet. I think we should also have a field called `fundingAmount` which is the amount of money we are inputting into this tx. We should be able to have an invariant then that looks something like this: 

`require(fundingAmount =  txFee + outgoingAmount + changeAmount)` I think? 

Also, this does not support batching, which i'm fine with, but we should file an issue for it at minimum.

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
```diff
@@ -0,0 +1,162 @@
...
+                          tx.lockTime,
+                          feePaid,
+                          expectedFeeRate,
+                          SatoshisPerVirtualByte(feeRate))
+  }
+}
+
+class OutgoingTransactionTable(tag: Tag)
+    extends Table[OutgoingTransactionDb](tag, "wallet_outgoing_txs") {
+
```
Christewart, line 59: Shouldn't there be a foreign key to our utxo table with utxos we are using to fund these txs?

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
nkohen: I actually don't think it is an issue to store all `FeeUnit`s as `sats/vbyte` but you have to be careful to actually write correctly, currently `SatoshisPerKiloByte(Satoshis(1)).toLong` returns `1`. You probably need to add a `toSatsPerVByte` to `FeeUnit` interface

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTransactionTable.scala
```diff
@@ -0,0 +1,134 @@
...
+}
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.{PrimaryKey, ProvenShape}
+
+/** Represents a relevant transaction for the wallet that we should be keeping track of */
+case class IncomingTransactionDb(
+    txIdBE: DoubleSha256DigestBE,
+    transaction: Transaction,
+    unsignedTxIdBE: DoubleSha256DigestBE,
+    unsignedTx: Transaction,
```
nkohen: The scaladoc isn't enough for me to know what the difference between `transaction` and `unsignedTx`, why both?

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionDAO.scala
```diff
@@ -0,0 +1,61 @@
...
+    Seq] =
+    findByPrimaryKeys(txs.map(_.txIdBE))
+
+  def findByTxId(
+      txIdBE: DoubleSha256DigestBE): Future[Option[OutgoingTransactionDb]] = {
+    val q = table
+      .filter(_.txIdBE === txIdBE)
+
+    database.run(q.result).map {
+      case h +: Vector() =>
```
nkohen: nit: Can't this case be `case Vector(h)`?

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
```diff
@@ -0,0 +1,162 @@
...
+import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerVirtualByte}
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.{PrimaryKey, ProvenShape}
+
+/** Represents a relevant transaction for the wallet that we should be keeping track of */
+case class OutgoingTransactionDb(
+    txIdBE: DoubleSha256DigestBE,
+    transaction: Transaction,
+    unsignedTxIdBE: DoubleSha256DigestBE,
+    unsignedTx: Transaction,
```
nkohen: ditto

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
benthecarman: To have an actual enforced foreign key, we would need a column for each out point. So we could do `outPoint0`, `outPoint1`, `outPoint2`, `outPoint3`, etc, but that is pretty ugly and we have to put a cap of inputs we can put in a tx

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionDAO.scala
benthecarman: No :(

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
benthecarman: This doesn't break with batching but breaks if we don't own every input, this would break with DLCs.

testkit/src/main/scala/org/bitcoins/testkit/wallet/WalletTestUtil.scala
```diff
@@ -143,14 +149,34 @@ object WalletTestUtil {
                     scriptPubKey = wspk)
   }
 
+  def insertDummyIncomingTransaction(daos: WalletDAOs, utxo: SpendingInfoDb)(
+      implicit ec: ExecutionContext): Future[IncomingTransactionDb] = {
+    val txDb = TransactionDb(utxo.txid,
```
Christewart: nit: Please use named parameters when you have a bunch of things like this, it's much easier to read.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -33,7 +33,7 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
   override def sendToAddress(
       address: BitcoinAddress,
       amount: CurrencyUnit,
-      feeRate: FeeUnit,
+      feeRate: SatoshisPerByte,
```
Christewart: Why did this change?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -68,17 +70,35 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
   private[wallet] def processOurTransaction(
       transaction: Transaction,
+      feeRate: SatoshisPerByte,
+      inputAmount: CurrencyUnit,
       blockHashOpt: Option[DoubleSha256DigestBE]): Future[ProcessTxResult] = {
     logger.info(
       s"Processing TX from our wallet, transaction=${transaction.txIdBE} with blockHash=$blockHashOpt")
-    processTransactionImpl(transaction, blockHashOpt).map { result =>
+    for {
+      _ <- insertOutgoingTransaction(transaction, feeRate, inputAmount)
```
Christewart: is there a reason this is segregated from `processTransactionImpl()`?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -337,7 +368,9 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           logger.trace(
             s"Found $count relevant output(s) in transaction=${transaction.txIdBE}: $outputStr")
 
-          val addUTXOsFut: Future[Seq[SpendingInfoDb]] =
+          val totalIncoming = xs.map(_.output.value).sum
+
+          def addUTXOsFut(): Future[Seq[SpendingInfoDb]] =
```
Christewart: nit: Any reason we don't make this a private helper method and move it out of this method?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -337,7 +368,9 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           logger.trace(
             s"Found $count relevant output(s) in transaction=${transaction.txIdBE}: $outputStr")
 
-          val addUTXOsFut: Future[Seq[SpendingInfoDb]] =
+          val totalIncoming = xs.map(_.output.value).sum
```
Christewart: nit: Maybe rename `xs` to something more descriptive? I know you didn't name it this to start with.

wallet/src/main/scala/org/bitcoins/wallet/models/IncomingTransactionTable.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def * : ProvenShape[IncomingTransactionDb] =
+    (txIdBE, incomingAmount) <> (fromTuple, toTuple)
+
+  def primaryKey: PrimaryKey =
+    primaryKey("pk_tx", sourceColumns = txIdBE)
+
+  def fk_txId0 = {
+    val txTable = TableQuery[TransactionTable]
+    foreignKey("fk_txId0", sourceColumns = txIdBE, targetTableQuery = txTable)(
```
Christewart: can we get a better name for this like `fk_rawtx`/`fk_underlyingtx` or something like that?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala
```diff
@@ -189,6 +189,13 @@ case class SpendingInfoTable(tag: Tag)
                targetTableQuery = addressTable)(_.scriptPubKey)
   }
 
+  /** All UTXOs must have a corresponding transaction in the wallet */
+  def fk_inc_txId = {
+    val txTable = TableQuery[IncomingTransactionTable]
+    foreignKey("fk_inc_txId", sourceColumns = txid, targetTableQuery = txTable)(
```
Christewart: what does `inc` mean? `increment`?

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+  * Represents a relevant transaction for the wallet that we should be keeping track of
+  * @param txIdBE Transaction ID
+  * @param txFee fee paid by the transaction
+  * @param expectedFeeRate Fee rate the wallet expected to pay
+  * @param actualFeeRate Fee rate the transaction actually paid
+  */
+case class OutgoingTransactionDb(
+    txIdBE: DoubleSha256DigestBE,
```
Christewart, line 18: Do we want to mention what index our change output is? Also I think it would be nice to have fields like `sentAmount` as well. This goes back to our discussion of having some sort of invariant in this case classes making sure everything is accounted for.

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
```diff
@@ -0,0 +1,79 @@
...
+}
+
+object OutgoingTransactionDb {
+
+  def fromTransaction(
+      tx: Transaction,
+      expectedFeeRate: SatoshisPerByte,
+      inputAmount: CurrencyUnit): OutgoingTransactionDb = {
+    val feePaid = inputAmount - tx.outputs.map(_.value).sum
+    val feeRate = feePaid.satoshis.toLong / tx.baseSize
```
Christewart: Yikes. Is there no helper methods for calcuations like this? This seems _so_ easy to forget to update in the future when we change the `expectedFeeRate` type.

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionDAO.scala
```diff
@@ -0,0 +1,61 @@
+package org.bitcoins.wallet.models
+
+import org.bitcoins.core.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.db.{CRUD, SlickUtil}
+import org.bitcoins.wallet.config._
+import slick.jdbc.SQLiteProfile.api._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class OutgoingTransactionDAO()(
```
Christewart, line 8: rather than duplicate a lot of code across `IncomingTransactionDAO()`/`OutgoingTransactionDAO()`/`TransactionDAO()` why not make an abstract class for the similar functionality and then extend it in these things?

Christewart (commented):
Looks like we need an integration test as well?

It should test 

1. Receive funds
2. assert tx is in incoming and tx_table
3. spend funds to at least 1 address externally and 1 change output
4. assert the tx is in outgoing_tx_table and tx_table
5. Make sure the change is correct for expected and actual fee
6. Make sure that outgoing tx ends up in the incoming tx table as it pays change to the wallet.

Am I missing anything else?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoTable.scala
benthecarman: incoming

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
benthecarman: > Do we want to mention what index our change output is?

I think that is kinda recreating the spending info database, also it could cause problems if there is a tx where we own multiple outcomes (ie coinjoins, payjoins)

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: Yes we need `feeRate` and `inputAmount` and these aren't available for every call of `processTransactionImpl` but are available for every call of `processOurTransaction`

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
benthecarman: Otherwise, resolved

wallet/src/main/scala/org/bitcoins/wallet/models/OutgoingTransactionTable.scala
benthecarman: I will implement this in #1236

Christewart:
CI failing: https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/667941626#L414

Christewart:
Why are we using isCloseEnough in tests? It seems like by definition that means we are tracking things accurately. 

I have an issue open to get rid of this https://github.com/bitcoin-s/bitcoin-s/issues/942

Christewart (approved):

PR #1200, opened 2020-03-06
Add single signer docs to TxBuilder docs

### LINKED ISSUE #1196: Add documentation to website about UTXOSpendingInfoSingle and UTXOSpendingInfoFull
AFAICT there is no examples of how to use `UTXOSpendingInfoSingle` on the site. Here is where I _think_ it should be added as an example with [`TxBuilder`](https://bitcoin-s.org/docs/next/core/txbuilder). Basically how to sign a multisig using multiple `UTXOSpendingInfoSingle` where we are pretending that we have a Alice, Bob, and Carol situation.

benthecarman:
Closes #1196

### DIFF
--- a/docs/core/txbuilder.md
+++ b/docs/core/txbuilder.md
@@ -3,7 +3,7 @@ id: txbuilder
 title: TxBuilder example
 ---
 
-Bitcoin-S features a transaction buidlder that constructs and signs Bitcoin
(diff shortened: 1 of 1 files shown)


Christewart:
So what I was envisioning here is to actually _fully_ sign the multisig using the single utxo's and then show how to aggregate the signature as if this user was on separate machines.

PR #864, opened 2019-11-14
Revive scripts/ project. Bring 'create-wallet.sc' and 'chain-sync.sc'…

Christewart:
… into the scripts/ project. Also add document talking about how to run the scripts

For information on how scripts work see [`docs/scripting.md`](https://github.com/bitcoin-s/bitcoin-s/blob/7ad71efe8e4258867ec49ddff47d15ced5bc1da2/docs/scripting.md)

I want to keep adding more scripts in the future, specifically around eclair.

PR #1204, opened 2020-03-06
ChainQueryApi doc

### LINKED ISSUE #1198: Document how ChainQueryApi interacts with the wallet
https://github.com/bitcoin-s/bitcoin-s/blob/1a3b6b5b1e4eb8442dfab8b1a9faeff74418bdb0/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L23

I want to make sure we document how the `ChainQueryApi` interacts with the wallet. We need to tell users how it is responsible for updating the wallet's state, and exactly how that happens internally, maybe with an example.

benthecarman:
Closes #1198

### DIFF
--- a/docs/wallet/chain-query-api.md
+++ b/docs/wallet/chain-query-api.md
@@ -3,6 +3,27 @@ id: chain-query-api
 title: chain-query-api
 ---
 
+```scala mdoc:invisible
+import akka.actor.ActorSystem
+import org.bitcoins.core.api.ChainQueryApi
+import org.bitcoins.core.api.ChainQueryApi.FilterResponse
+import org.bitcoins.core.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.core.gcs.{FilterType, GolombFilter}
+import org.bitcoins.core.protocol.BlockStamp
+import org.bitcoins.core.protocol.blockchain.Block
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.keymanager.bip39.BIP39KeyManager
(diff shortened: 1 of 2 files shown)


docs/wallet/chain-query-api.md
```diff
@@ -0,0 +1,201 @@
...
+        }
+        .toVector
+
+      Future.sequence(filterFs)
+    }
+  }
+
+// Finally, we can initialize our wallet with our own node api
+  val wallet =
+    Wallet(keyManager = keyManager, nodeApi = nodeApi, chainQueryApi = chainApi)
```
Christewart, line 195: same thing as I said in https://github.com/bitcoin-s/bitcoin-s/pull/1205#discussion_r389860261 , just trigger one of the callbacks so someone can see how this is executed. Eventually I would like to move these to into scripts that can be run on their own. See #864 for what that would look like.

docs/wallet/chain-query-api.md
```diff
@@ -0,0 +1,207 @@
...
+want.
+
+The functions that the ChainQueryApi supports are:
+
+```scala mdoc:compile-only
+import org.bitcoins.core.api.ChainQueryApi.FilterResponse
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+import org.bitcoins.core.protocol.BlockStamp
+
+import scala.concurrent.Future
```
nkohen: nit: Move these `import`s to an `invisible` block above

nkohen: You can merge these imports with the ones below (by bringing the ones below higher)

PR #1205, opened 2020-03-06
NodeApi doc

### LINKED ISSUE #1199: Document how nodeApi interacts with the wallet
https://github.com/bitcoin-s/bitcoin-s/blob/1a3b6b5b1e4eb8442dfab8b1a9faeff74418bdb0/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L22

The nodeApi is responsibly for querying network information in the case of using neutrino with your wallet, I believe it needs to be used in the case of rescans when a filter has a match and we need to fetch a block.

benthecarman:
Closes #1199

### DIFF
--- a/docs/wallet/node-api.md
+++ b/docs/wallet/node-api.md
@@ -1,0 +1,24 @@
+---
+id: node-api
+title: node-api
+---
+
+### NodeApi
+
+The NodeApi is how the wallet project retrieves relevant node data like blocks.
+This allows the wallet for example to retrieve blocks for finding its relevant transactions.
+
+Since this is an API it can be hooked up to the `node` module of bitcoin-s but it can also be linked to
+any other implementation of your choosing. This allows you to use the bitcoin-s wallet in any schema that you
+want.
+
+The functions that the NodeApi supports are:
+
+```scala mdoc
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+
+import scala.concurrent.Future
+
+  /** Request the underlying node to download the given blocks from its peers and feed the blocks to [[org.bitcoins.node.NodeCallbacks]] */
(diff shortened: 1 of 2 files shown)


docs/wallet/node-api.md
```diff
@@ -0,0 +1,24 @@
...
+want.
+
+The functions that the NodeApi supports are:
+
+```scala mdoc
+import org.bitcoins.core.crypto.DoubleSha256DigestBE
+
+import scala.concurrent.Future
+
+  /** Request the underlying node to download the given blocks from its peers and feed the blocks to [[org.bitcoins.node.NodeCallbacks]] */
```
Christewart, line 38: So i don't really find this useful at all for documentation purposes. How do i hook this into the wallet? What is an example of an implementation? 

For example, take a look at the chain sync documentation here: https://bitcoin-s.org/docs/next/applications/chain#syncing-and-verifying-block-headers

 and the new filter sync document i've added in #1209

docs/wallet/node-api.md
```diff
@@ -0,0 +1,101 @@
...
+    override def downloadBlocks(
+        blockHashes: Vector[DoubleSha256Digest]): Future[Unit] = {
+      val blockFs = blockHashes.map(hash => bitcoind.getBlockRaw(hash))
+      Future.sequence(blockFs).map(_ => ())
+    }
+  }
+
+// Finally, we can initialize our wallet with our own node api
+  val wallet =
+    Wallet(keyManager = keyManager, nodeApi = nodeApi, chainQueryApi = chainApi)
```
Christewart, line 97: can you trigger this event so that `println(s"Received block: ${block.blockHeader.hashBE}")` is executed.

that should wrap up the example IMO.

Christewart:
needs rebase

Christewart (approved):

PR #1206, opened 2020-03-06
Change NodeCallbacks to `Future[Unit]` s

benthecarman:
Taken from #1178

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
@@ -12,6 +12,7 @@ import org.bitcoins.testkit.core.gen.{
 import org.bitcoins.testkit.util.BitcoinSUnitTest
 import org.scalacheck.Gen
 
+import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
 class MerkleBuffersTest extends BitcoinSUnitTest {
@@ -41,6 +42,7 @@ class MerkleBuffersTest extends BitcoinSUnitTest {
                 assert(txs == merkleTxs,
                        "Received TXs in callback was not the ones we put in")))
             callbackCount = callbackCount + 1
+            Future.unit
         }
 
         val merkle = MerkleBlock(block, txs.map(_.txId))
--- a/node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
@@ -3,6 +3,7 @@ package org.bitcoins.node.networking.peer
 import _root_.org.scalatest.compatible.Assertion
 import org.bitcoins.core.protocol.blockchain.{Block, MerkleBlock}
 import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.testkit.BitcoinSTestAppConfig
 import org.bitcoins.testkit.core.gen.{
@@ -12,7 +13,6 @@ import org.bitcoins.testkit.core.gen.{
 import org.bitcoins.testkit.util.BitcoinSUnitTest
 import org.scalacheck.Gen
 
-import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
 class MerkleBuffersTest extends BitcoinSUnitTest {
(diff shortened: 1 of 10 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -96,19 +97,18 @@ case class DataMessageHandler(
...
           newChainApi <- chainApi.processFilter(filter)
-        } yield {
-
-          Try {
-            val blockFilter =
-              BlockFilter.fromBytes(filter.filterBytes, filter.blockHash)
-            callbacks.onCompactFilterReceived.foreach(
-              _.apply(filter.blockHash, blockFilter))
-          } match {
+          blockFilterT = Try(
```
rorp: I think it could be rewritten without matching a `Try[]`:
```suggestion
          blockFilter <- Future(BlockFilter.fromBytes(filter.filterBytes, filter.blockHash))
          _ <- callbacks.onCompactFilterReceived.foldLeft(FutureUtil.unit)(
                (acc, callback) =>
                  acc.flatMap(_ => callback(filter.blockHash, blockFilter)))
```

node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
```diff
@@ -41,6 +42,7 @@ class MerkleBuffersTest extends BitcoinSUnitTest {
                 assert(txs == merkleTxs,
                        "Received TXs in callback was not the ones we put in")))
             callbackCount = callbackCount + 1
+            Future.successful(())
```
Christewart: nit: `FutureUtil.unit`

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -318,18 +313,23 @@ case class DataMessageHandler(
...
-  type OnCompactFilterReceived = (DoubleSha256Digest, GolombFilter) => Unit
+  type OnCompactFilterReceived =
+    (DoubleSha256Digest, GolombFilter) => Future[Unit]
+
+  /** Callback for handling a received block header */
+  type OnBlockHeadersReceived = Vector[BlockHeader] => Future[Unit]
 
   /** Does nothing */
-  def noop[T]: T => Unit = _ => ()
+  def noop[T]: T => Future[Unit] = _ => Future.successful(())
```
Christewart: nit: `FutureUtil.unit`

node-test/src/test/scala/org/bitcoins/node/networking/peer/MerkleBuffersTest.scala
benthecarman: `Future.unit` isn't available for scala 2.11.12

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
benthecarman: `Future.unit` isn't available for scala 2.11.12

benthecarman:
Add Callbacks expected behavior doc

rorp (approved):

docs/applications/node.md
```diff
@@ -11,3 +11,12 @@ that need to connect to the P2P network.
 This node is currently only released as a library, and not as a binary.
 This is because it (nor the documentation) is not deemed production
 ready. Use at your own risk, and without too much money depending on it.
+
+## Node Callback Handling
```
Christewart: i believe you need to rebase onto master and then put this in the correct file. We've moved things since this PR was opened

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -96,19 +96,19 @@ case class DataMessageHandler(
...
-            callbacks.onCompactFilterReceived.foreach(
-              _.apply(filter.blockHash, blockFilter))
-          } match {
-            case Failure(ex) =>
-              logger.error("Error processing compact filter", ex)
-            case Success(_) => ()
-          }
+          blockFilter <- Future(
+            BlockFilter.fromBytes(filter.filterBytes, filter.blockHash))
+          _ <- callbacks.onCompactFilterReceived
```
Christewart: So instead of defining this everywhere in `DataMessageHandler`, can we just create a helper method in `NodeCallbacks` that does this, and here we just call `_ <- NodeCallbacks.execute(callbacks)` or something like that?

Christewart (commented):
I think there should be a unit test that implements two callbacks on something where the first callback does `sys.error(s"Thrown error")` or whatever, and then we have a second test case that completes a promise on execution to succeed the test case. 

Basically just a unit test to make sure the first callback failing doesn't prevent the second callback from executing.

rorp:
There are a bunch of compiler errors.

PR #1208, opened 2020-03-08
Adding new SPK doc

nkohen:
Added a document to the website explaining how to add new script types to Bitcoin-S

Also added a doc explaining the differences between Full and Single Signing.


docs/core/adding-spks.md
```diff
@@ -0,0 +1,683 @@
...
+                                  output,
+                                  scriptSigF)
+  }
+}
+object ConditionalSigner extends ConditionalSigner
+```
+
+## Step 9: Add to BitcoinSigner.sign
+
+We must now add the new signing functionality from the previous step to the general-purpose signing functions by adding a new `case` for your new `ScriptPubKey` type in the `match` within `BitcoinSigner.sign`. In the case of `P2PKWithTimeout`, this looks like:
```
benthecarman: I'm not sure if this is a github bug or not but when I do `display rich diff` it shows "`ScriptPubKey` type in the `match` withing `BitcoinSigner.sign`

docs/core/adding-spks.md
```diff
@@ -0,0 +1,683 @@
...
+                                        p2pkWithTimeout,
+                                        signer,
+                                        hashType,
+                                        beforeTimeout)
+        }
+```
+
+## Step 8: Create a Signer
+
+We must now add signing functionality for our new script type withing `Signer.scala`. Once again, we have three different cases depending on your new script type.
```
benthecarman: > withing

within

docs/core/adding-spks.md
```diff
@@ -0,0 +1,683 @@
...
+[warn]       scriptPubKey: ScriptPubKey): Gen[ScriptSignature] = scriptPubKey match {
+[warn]                                                           ^
+[warn] one warning found
+```
+
+You may get these warnings for your new `ScriptSignature` type as well. These are places where the compiler expects there to be defined functionality in a pattern match where one of our new types is a possibility, but for which no functionality is defined. You must go to each of these warnings and add a `case` for the relevant new type, or add this new type to an existing case when applicable.
+
+## Step 13: Run tests and debug
+
+Lastly, once everything is compiling nicely, all that is left is to run tests and debug. While withing an `sbt` terminal session, run the following two commands to run the relevant tests:
```
benthecarman: > withing

within

docs/core/signing-transactions.md
```diff
@@ -0,0 +1,12 @@
+---
+id: spending-info
+title: Signing Transactions
+---
+
+# Signing Transactions in Bitcoin-S
+
+There are two kinds of transaction signing in Bitcoin-S, as defined in `Signer.scala`: Single Signing and Full Signing.
+
+Single signing, which is implemented by the subtypes of `SingleSigner`, consists of using a single `ECPrivateKey` (or other [Sign implementation](sign.md)) to generate a `PartialSignature` of the given transaction. This consists of an `ECDigitalSignature` and a `ECPublicKey`. This is useful any time you have a single private key and wish to get a single `ECDigitalSignature` for that private key as is the case when using [Partially Signed Bitcoin Transactions](psbts.md). In order to call `signSingle` on any instance of `SingleSigner`, you must have access to the (not signed) transaction being signed, as well as a `BitcoinUTXOSpendingInfoSingle`, which encapsulates all the information needed to sign with a single key.
```
benthecarman: > you must have access to the (not signed) transaction being signed

this sounds weird, can just do "you must have access to the unsigned transaction"

benthecarman (approved):

docs/core/adding-spks.md
```diff
@@ -0,0 +1,683 @@
...
+
+Bitcoin-S strives to have script types defined in such a way that they can be easily composed and reused. Before going through this guide and implementing a really large script template type, try to decompose your script into smaller re-usable pieces.
+
+Also remember to consider what existing pieces you can use. For example, `LockTimeScriptPubKey`s are implemented in such a way that any other `RawScriptPubKey` can be given a time lock by nesting it within a `LockTimeScriptPubKey` subtype. Likewise, `ConditionalScriptPubKey`s are built to allow any other `RawScriptPubKey` type to populate both the `OP_IF/OP_NOTIF` case and the `OP_ELSE` case.
+
+## Step 1: Create a New ScriptPubKey Trait
+
+Go to `ScriptPubKey.scala` and add a new trait:
+
+```scala
```
Christewart: So one of the nice things about mdoc is that it gives you compiler errors if your documents every become out of date. For instance, if we were to change the `RawScriptPubKey` name, how would we know we need to update this doc? That's the problem that `mdoc` solves:

https://scalameta.org/mdoc/docs/modifiers.html#compile-only

To use this, you need to use 
`scala mdoc:compile-only` rather than just `scala` when typing these things out.

now when you do `docs/run` it will compile the scala code and give you a compiler error if it isn't valid.

PR #1209, opened 2020-03-08
2020 03 08 filter sync

Christewart:
This implements `FilterSync` -- which similar to `ChainSync`. The basic idea is that given a single function `getFilterFunc: BlockHeader => Future[GolombFilter]` that allows us to query filters for a given `BlockHeader`, we can sync our internal filter data store. 

An example of this is using `BitcoindV19RpcClient`. You can implement the function like this: 

```scala
    val getFilterFunc = { case header: BlockHeader =>
      val prevFilterResultF =
        bitcoind.getBlockFilter(header.previousBlockHashBE, filterType)
      prevFilterResultF.map(_.filter)
   }   
```

We use this in our test cases to verify that this is working. This PR also adds explicit chain test fixtures for `BitcoindV19RpcClient` as it is needed so we can use the `getBlockFilter()` rpc. We cannot use older versions as the RPC is not supported.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -193,9 +193,9 @@ case class ChainHandler(
       filterHeaders: Vector[FilterHeader],
       stopHash: DoubleSha256DigestBE): Future[ChainApi] = {
 
-    val filterHeadersToCreateF = for {
+    val filterHeadersToCreateF: Future[Vector[CompactFilterHeaderDb]] = for {
       blockHeaders <- blockHeaderDAO
-        .getNChildren(stopHash, filterHeaders.size - 1)
+        .getNChildren(ancestorHash = stopHash, n = filterHeaders.size - 1)
         .map(_.sortBy(_.height))
     } yield {
       if (blockHeaders.size != filterHeaders.size) {
@@ -214,17 +214,24 @@ case class ChainHandler(
     for {
       filterHeadersToCreate <- filterHeadersToCreateF
       _ <- if (filterHeadersToCreate.nonEmpty && filterHeadersToCreate.head.height > 0) {
-        filterHeaderDAO
-          .findByHash(filterHeadersToCreate.head.previousFilterHeaderBE)
-          .map { prevHeaderOpt =>
+        val firstFilter = filterHeadersToCreate.head
+        val filterHashFOpt = filterHeaderDAO
+          .findByHash(firstFilter.previousFilterHeaderBE)
+        filterHashFOpt.map {
+          case Some(prevHeader) =>
             require(
-              prevHeaderOpt.nonEmpty,
-              s"Previous filter header does not exist: ${filterHeadersToCreate.head.previousFilterHeaderBE}")
-            require(
-              prevHeaderOpt.get.height == filterHeadersToCreate.head.height - 1,
-              s"Unexpected previous header's height: ${prevHeaderOpt.get.height} != ${filterHeadersToCreate.head.height - 1}"
+              prevHeader.height == firstFilter.height - 1,
+              s"Unexpected previous header's height: ${prevHeader.height} != ${filterHeadersToCreate.head.height - 1}"
             )
-          }
+          case None =>
+            if (firstFilter.previousFilterHeaderBE == DoubleSha256DigestBE.empty && firstFilter.height == 0) {
+              //we are ok, according to BIP157 the previous the genesis filter's prev hash should
+              //be the empty hash
+              ()
+            } else {
+              sys.error(s"Previous filter header does not exist: $firstFilter")
+            }
+        }
       } else FutureUtil.unit
       _ <- filterHeaderDAO.createAll(filterHeadersToCreate)
     } yield this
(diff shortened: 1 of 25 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -347,6 +354,33 @@ case class ChainHandler(
       height: Int): Future[Vector[CompactFilterHeaderDb]] =
     filterHeaderDAO.getAtHeight(height)
 
+  /** @inheritdoc */
+  override def getBestFilterHeader(): Future[CompactFilterHeaderDb] = {
+    //this seems realy brittle, is there a guarantee
+    //that the highest filter header count is our best filter header?
+    //
```
benthecarman: nit: remove this `//`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -347,6 +354,33 @@ case class ChainHandler(
...
+    val filterCountF = getFilterHeaderCount()
+    val ourBestFilterHeader = for {
+      count <- filterCountF
+      filterHeader <- getFilterHeadersAtHeight(count)
+    } yield {
+      //TODO: Figure out what the best way to select
+      //the best filter header is if we have competing
+      //chains. (Same thing applies to getBestBlockHash()
+      //for now, just do the dumb thing and pick the first one
```
benthecarman, line 369: Should just be to calc based off of block headers and total work

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/FilterSync.scala
```diff
@@ -0,0 +1,139 @@
...
+      getFilterFunc: BlockHeader => Future[GolombFilter])(
+      implicit ec: ExecutionContext,
+      chainAppConfig: ChainAppConfig): Future[ChainApi] = {
+    if (ourBestFilterHeader.blockHashBE == ourBestHeader.hashBE) {
+      logger.info(
+        s"Our filters are synced with our peers filters, both at blockHash=${ourBestFilterHeader.blockHashBE}")
+      Future.successful(chainApi)
+    } else {
+      logger.info(
+        s"Beginning sync for filters from filterheader=${ourBestFilterHeader} to blockheader=${ourBestHeader.hashBE}")
```
benthecarman, line 80: > ourBestFilterHeader

Should this be `ourBestFilterHeader.hashBE`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
Christewart: I filed an issue for this: https://github.com/bitcoin-s/bitcoin-s/issues/1211

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/FilterSync.scala
Christewart: No. The idea behind this sync is to to close the gap between our best filter header, and the best block header we have. Is this clear from `filter-sync.md` or should I add more about this?

I just realized I didn't write about this at all, adding now

benthecarman: I think the function makes sense. I just meant for the log you are outputting `ourBestFilterHeader.toString` and I think you should output `ourBestFilterHeader.hashBE`

Christewart: Take a look at c6ec6505e43c2d60d318303ffcd6b627800f76ae and let me know if this clears it up

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/FilterSync.scala
```diff
@@ -0,0 +1,139 @@
...
+        missing <- chainApi.getHeadersBetween(from = bestFilterBlockHeader.get,
+                                              to = ourBestHeader)
+      } yield missing
+
+      //now that we have headers that are missing filters, let's fetch the filters
+      val fetchFiltersF: Future[Vector[(BlockHeaderDb, GolombFilter)]] = for {
+        missingHeaders <- headersMissingFiltersF
+        fetchNested = missingHeaders.map { b =>
+          val filterF = getFilterFunc(b.blockHeader)
+          filterF.map(f => (b, f))
```
rorp: Filters can be pretty big, especially on mainnet. Storing all of them in the memory can cause memory issues.

chain/src/main/scala/org/bitcoins/chain/blockchain/sync/FilterSync.scala
```diff
@@ -0,0 +1,139 @@
...
+    filters.zipWithIndex.map {
+      case ((blockHeaderDb, filter), idx) =>
+        val filterHeader = if (idx == 0) {
+          FilterHeader(filterHash = filter.hash,
+                       prevHeaderHash = ourBestFilterHeader.hashBE.flip)
+        } else {
+          val prev = filters(idx - 1)._2.hash
+          FilterHeader(filterHash = filter.hash, prevHeaderHash = prev)
+        }
+        BlockFilterAggregated(filterHeader, filter, blockHeaderDb.blockHeader)
```
rorp: I think it's worth mentioning that this class can be only used with a trusted Bitcoin Core node, because it doesn't verify filters against filter header, but rather generates filter headers by its own.

docs/applications/filter-sync.md
```diff
@@ -0,0 +1,133 @@
...
+import scala.concurrent._
+
+```
+
+```scala mdoc:compile-only
+val getFilterFunc: BlockHeader => Future[FilterWithHeaderHash] = ???
+```
+
+With `getFilterFunc` given a `BlockHeader` we can find it's associated `GolombFilter` -- which is our internal repesentation
+of a BIP157 block filter.
```
nkohen, line 48: nit: BIP 158 block filter

nkohen (commented):
ACK to `filter-sync.md`

PR #1218, opened 2020-03-11
2020 03 06 wallet rescan test

Christewart:
This PR will implement unit tests for wallet rescans. 

We now have wallet unit tests that can 

1. Rescan the entire blockchain searching for funds
2. Scan the blockchain and find no funds when we expect no funds (negative test case)
3. Scan a portion of the blockchain and find a portion of the funds the original wallet had

As a by product of this, I needed to implement `ChainQueryApi`/`NodeApi` with `BitcoindRpcClient` so we have a way to sync against bitcoind. This made me realize we can't seem to actually implement `ChainQueryApi`/`NodeApi` against bitcoind without just returning trivially failed `Future` for some of the methods, which means we cannot reliably depend on those methods for wallet operation (!).

The second thing that has to be done in this PR is to give us the ability to inject a custom `ChainQueryApi`/`NodeApi` into our wallet fixtures in `BitcoinSWalletTest`. For rescans to work, we specifically need to use `BitcoindV19RpcClient` as it has support for filters, which is why there is a lot of `V19` prefixes on fixtures. 

In the future, we probably need to parameterize the wallet related to `BitcoindRpcClient` so we avoid some nasty casting I have implemented as of now.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
@@ -4,7 +4,7 @@ import org.bitcoins.core.crypto.DoubleSha256DigestBE
 import org.bitcoins.core.gcs.GolombFilter
 import org.bitcoins.core.protocol.BlockStamp
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 /**
   * This trait provides methods to query various types of blockchain data.
@@ -18,6 +18,12 @@ trait ChainQueryApi {
 
   /** Gets the hash of the block that is what we consider "best" */
   def getBestBlockHash(): Future[DoubleSha256DigestBE]
+
+  def getBestHashBlockHeight()(implicit ec: ExecutionContext): Future[Int] = for {
+    hash <- getBestBlockHash()
+    heightOpt <- getBlockHeight(hash)
+    _ = require(heightOpt.isDefined, s"Best block hash must have a height!")
+  } yield heightOpt.get
 
   /** Gets number of confirmations for the given block hash*/
   def getNumberOfConfirmations(
--- a/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
@@ -19,11 +19,12 @@ trait ChainQueryApi {
   /** Gets the hash of the block that is what we consider "best" */
   def getBestBlockHash(): Future[DoubleSha256DigestBE]
 
-  def getBestHashBlockHeight()(implicit ec: ExecutionContext): Future[Int] = for {
-    hash <- getBestBlockHash()
-    heightOpt <- getBlockHeight(hash)
-    _ = require(heightOpt.isDefined, s"Best block hash must have a height!")
-  } yield heightOpt.get
+  def getBestHashBlockHeight()(implicit ec: ExecutionContext): Future[Int] =
+    for {
+      hash <- getBestBlockHash()
+      heightOpt <- getBlockHeight(hash)
+      _ = require(heightOpt.isDefined, s"Best block hash must have a height!")
+    } yield heightOpt.get
 
   /** Gets number of confirmations for the given block hash*/
   def getNumberOfConfirmations(
--- a/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
@@ -23,7 +23,8 @@ trait ChainQueryApi {
     for {
       hash <- getBestBlockHash()
       heightOpt <- getBlockHeight(hash)
-      _ = require(heightOpt.isDefined, s"Best block hash must have a height!")
+      _ = require(heightOpt.isDefined,
+                  s"Best block hash must have a height! blockhash=$hash")
     } yield heightOpt.get
 
   /** Gets number of confirmations for the given block hash*/
--- a/docs/wallet/wallet-rescan.md
+++ b/docs/wallet/wallet-rescan.md
@@ -7,15 +7,83 @@ With [BIP157](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki) yo
 later for rescans in the case you need to restore your wallets. Our [chain](../applications/chain.md) project gives us
 an API with the ability to query for filters.
 
-In combination with a set of scripts, you can iterate through each of the filters with [`WalletApi.rescanNeutrinoWallet()`](https://github.com/bitcoin-s/bitcoin-s/blob/1a3b6b5b1e4eb8442dfab8b1a9faeff74418bdb0/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L399)
+You can rescan your wallet with filters with [`WalletApi.rescanNeutrinoWallet()`](https://github.com/bitcoin-s/bitcoin-s/blob/1a3b6b5b1e4eb8442dfab8b1a9faeff74418bdb0/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L399)
 
 
-#### Example
+### Example
 
-In this example we are going show how to rescan our bitcoin-s wallet with `CompactFilter`.
+To run this example you need to make sure you have access to a bitcoind binary. You can download this with bitcoin-s by doing
+`sbt downloadBitcoind`
 
-The first thing we need to do is spin up a
+```scala mdoc:invisible
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.wallet._
+import org.bitcoins.server.BitcoinSAppConfig
+import akka.actor.ActorSystem
+import scala.concurrent.{ExecutionContext, Future, Await}
+import scala.concurrent.duration.DurationInt
+```
 
 ```scala mdoc:compile-only
 
+//we need an actor system and app config to power this
+implicit val system: ActorSystem = ActorSystem(s"wallet-rescan-example")
+implicit val ec: ExecutionContext = system.dispatcher
+implicit val appConfig: BitcoinSAppConfig = BitcoinSTestAppConfig.getNeutrinoTestConfig()
+
+//ok now let's spin up a bitcoind and a bitcoin-s wallet with funds in it
+val walletWithBitcoindF = for {
+  w <- BitcoinSWalletTest.createWalletBitcoindNodeChainQueryApi()
+} yield w
+
+val walletF = walletWithBitcoindF.map(_.wallet)
+
+val bitcoindF = walletWithBitcoindF.map(_.bitcoind)
+
+//let's see what our initial wallet balance is
+val initBalanceF = for {
+  w <- walletF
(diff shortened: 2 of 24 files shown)


core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
```diff
@@ -19,6 +19,13 @@ trait ChainQueryApi {
   /** Gets the hash of the block that is what we consider "best" */
   def getBestBlockHash(): Future[DoubleSha256DigestBE]
 
+  def getBestHashBlockHeight()(implicit ec: ExecutionContext): Future[Int] =
+    for {
+      hash <- getBestBlockHash()
+      heightOpt <- getBlockHeight(hash)
+      _ = require(heightOpt.isDefined, s"Best block hash must have a height!")
```
benthecarman: error should include the hash

docs/wallet/wallet-rescan.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Wallet Rescans
+id: wallet-rescan
+---
+
+With [BIP157](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki) you can cache block filters locally to use
```
benthecarman, line 6: > With BIP157 you can cache block filters locally

I think it'd be better to say

> With a neutrino wallet you can cache BIP 157 block filters locally

docs/wallet/wallet-rescan.md
```diff
@@ -0,0 +1,21 @@
...
+In combination with a set of scripts, you can iterate through each of the filters with [`WalletApi.rescanNeutrinoWallet()`](https://github.com/bitcoin-s/bitcoin-s/blob/1a3b6b5b1e4eb8442dfab8b1a9faeff74418bdb0/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L399)
+
+
+#### Example
+
+In this example we are going show how to rescan our bitcoin-s wallet with `CompactFilter`.
+
+The first thing we need to do is spin up a
+
+```scala mdoc:compile-only
```
benthecarman, line 27: you didn't do an example

testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
```diff
@@ -38,6 +45,176 @@ abstract class SyncUtil {
...
+        }
+
+        val batchSize = 25
+        val batchedExecutedF = FutureUtil.batchExecute(elements = blockHashes,
+                                                       f = f,
+                                                       init = Vector.empty,
+                                                       batchSize = batchSize)
+
+        batchedExecutedF.map { _ =>
+          logger.info(s"Done fetching ${blockHashes} hashes from bitcoind")
```
benthecarman: > Done fetching ${blockHashes} hashes from bitcoind

`Done fetching ${blockHashes.length} hashes from bitcoind`

testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
```diff
@@ -38,6 +45,176 @@ abstract class SyncUtil {
...
```
benthecarman: ditto

wallet-test/src/test/scala/org/bitcoins/wallet/RescanHandlingTest.scala
```diff
@@ -0,0 +1,141 @@
...
+  behavior of "Wallet rescans"
+
+  val DEFAULT_ADDR_BATCH_SIZE = 10
+  it must "be able to discover funds that belong to the wallet using WalletApi.rescanNeutrinoWallet" in {
+    fixture: WalletWithBitcoind =>
+      val WalletWithBitcoindV19(wallet, bitcoind) = fixture
+
+      val initBalanceF = wallet.getBalance()
+
+      val rescanF = for {
```
benthecarman, line 36: useless variable

wallet-test/src/test/scala/org/bitcoins/wallet/RescanHandlingTest.scala
```diff
@@ -0,0 +1,141 @@
...
+        utxos <- utxosF
+        blockhashes = utxos.map(_.blockHash)
+        heights <- Future.sequence {
+          blockhashes.map(h =>
+            wallet.chainQueryApi.getBlockHeight(h.get).map(_.get))
+        }
+      } yield heights.min
+
+      //ok now that we have the height of the oldest utxo, let's rescan up to then
+      val rescanF = for {
```
benthecarman, line 116: useless variable

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -330,30 +329,17 @@ trait LockedWalletApi extends WalletApi {
...
-            res <- acc
-          } yield {
-            res ++ filtered
-          }
-          val newEnd = Math.max(start, endHeight - batchSize)
-          loop(start, newEnd, newAcc)
-        }
+
+
+      def fetchFiltersInRange(heightRange: Vector[Int]): Future[Vector[BlockMatchingResponse]] = {
```
benthecarman: I think this would be better as `def fetchFiltersInRange(startHeight: Int, stopHeight: Int)`

wallet-test/src/test/scala/org/bitcoins/wallet/RescanHandlingTest.scala
```diff
@@ -0,0 +1,141 @@
...
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withFundedWalletAndBitcoindV19(test)
+  }
+
+  behavior of "Wallet rescans"
+
+  val DEFAULT_ADDR_BATCH_SIZE = 10
+  it must "be able to discover funds that belong to the wallet using WalletApi.rescanNeutrinoWallet" in {
+    fixture: WalletWithBitcoind =>
+      val WalletWithBitcoindV19(wallet, bitcoind) = fixture
```
benthecarman: `bitcoind` is unused

wallet-test/src/test/scala/org/bitcoins/wallet/RescanHandlingTest.scala
```diff
@@ -0,0 +1,141 @@
...
+      }
+  }
+
+  it must "NOT discover funds that happened OUTSIDE of a certain range of block hashes" in {
+    fixture: WalletWithBitcoind =>
+      val WalletWithBitcoindV19(wallet, bitcoind) = fixture
+
+      val initBalanceF = wallet.getBalance()
+
+      val blockCountF = wallet.chainQueryApi.getBestHashBlockHeight()
```
benthecarman: unused

wallet-test/src/test/scala/org/bitcoins/wallet/RescanHandlingTest.scala
```diff
@@ -0,0 +1,141 @@
...
+                                                DEFAULT_ADDR_BATCH_SIZE)
+        balance <- newTxWallet.getBalance()
+      } yield {
+        assert(balance == amt)
+      }
+  }
+
+  it must "NOT discover funds that happened OUTSIDE of a certain range of block hashes" in {
+    fixture: WalletWithBitcoind =>
+      val WalletWithBitcoindV19(wallet, bitcoind) = fixture
```
benthecarman: `bitcoind` is unused

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -330,30 +329,17 @@ trait LockedWalletApi extends WalletApi {
...
-              endHeight)
-            filtered <- findMatches(compactFilterDbs)
-            res <- acc
-          } yield {
-            res ++ filtered
-          }
-          val newEnd = Math.max(start, endHeight - batchSize)
-          loop(start, newEnd, newAcc)
-        }
+
```
benthecarman: nit: lotta white space here

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
benthecarman: if not then you should do
```
val startHeight = heightRange.min
val endHeight = heightRange.max
```

website/sidebars.json
```diff
@@ -23,6 +23,11 @@
       "rpc/rpc-eclair",
       "rpc/rpc-bitcoind"
     ],
+    "Rs": [
+      "rpc/rpc-clients-intro",
+      "rpc/rpc-eclair",
+      "rpc/rpc-bitcoind"
+    ],
```
benthecarman: ?

benthecarman (changes requested):
I think in the future it should be done to use a gap limit rather than a address batch size.

Otherwise, some minor edits

core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
rorp: There's always a best block, so its height can't be undefined.  Maybe it makes sense to get best block header and extract the height from it?

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -40,6 +43,46 @@ private[wallet] trait RescanHandling extends WalletLogger {
...
+        _ = if (startHeight > endHeight)
+          throw InvalidBlockRange(
+            s"End position cannot precede start: $startHeight:$endHeight")
+        _ = logger.info(
+          s"Beginning to search for matches between ${startHeight}:${endHeight}")
+        range = startHeight.to(endHeight)
+        batchSize = calcGroupSize(range.length, parallelismLevel)
+        matched <- FutureUtil.batchExecute(
+          elements = range.toVector,
+          f = fetchFiltersInRange(scripts, parallelismLevel)(_),
```
rorp, line 74: I'm confused about `parallelismLevel`. For some reason it's used in several places. How many thread this code ends up creating?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -33,6 +33,13 @@ abstract class LockedWallet
       spendingInfoCount <- spendingInfoDAO.count()
     } yield addressCount == 0 && spendingInfoCount == 0
 
+  override def clearUtxosAndAddresses(): Future[LockedWallet] = {
```
rorp, line 36: I believe it should be used in `RescanHandling.rescanNeutrinoWallet()` as well.

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
Christewart: I moved methods embedded inside of `getMatchingBlocks()` to their own methods outside of this method. We were using things (like `parallelismLevel`) without explicitly passing them into the helper functions embedded inside of `getMatchingBlocks()`. Now to actually use the parallelism level we need to explicitly pass it into `fetchFiltersInRange` which in turn calls `findMatches()` where we use it calculate the batch size of attempting to match filters with `SimpleFilterMatcher`.

PR #1219, opened 2020-03-11
Added hash types to ScriptSignature Generators

### LINKED ISSUE #1217: Non of the signatures we produce in ScriptGenerators have a hash type appended to them
In general, it seems like it would be a good idea to distinguish between a vanilla digital signature and one that has a hash type appended to it. For instance, if you look at

https://github.com/bitcoin-s/bitcoin-s/blob/4dedb90ee8a3347d79db3c87e3b71de919e570a0/testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala#L70

you will see that the signatures generated there have no hash type!

nkohen:
closes #1217

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
@@ -26,6 +26,7 @@ import org.bitcoins.core.wallet.utxo.{
   P2PKWithTimeoutSpendingInfo
 }
 import org.scalacheck.Gen
+import scodec.bits.ByteVector
 
 import scala.annotation.tailrec
 import scala.concurrent.Await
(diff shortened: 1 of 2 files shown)


testkit/src/main/scala/org/bitcoins/testkit/core/gen/ScriptGenerators.scala
```diff
@@ -49,32 +50,41 @@ sealed abstract class ScriptGenerators extends BitcoinSLogger {
 
   def p2pkScriptSignature: Gen[P2PKScriptSignature] =
     for {
-      digitalSignature <- CryptoGenerators.digitalSignature
+      sig <- CryptoGenerators.digitalSignature
+      hashType <- CryptoGenerators.hashType
+      digitalSignature = ECDigitalSignature(
+        sig.bytes ++ ByteVector.fromByte(hashType.byte))
```
benthecarman, line 56: Can you not do `sig.bytes :+ hashType.byte`?

PR #1224, opened 2020-03-12
Fix the time based test cases in BitcoindV17RpcClientTest

rorp:
These test failed at certain time of the day depending on the local time zone.


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v17/BitcoindV17RpcClientTest.scala
```diff
@@ -114,7 +113,8 @@ class BitcoindV17RpcClientTest extends BitcoindRpcTest {
       addr <- client.getNewAddress
       info <- client.getAddressInfo(addr)
     } yield assert(
-      info.timestamp.exists(_.getDayOfYear == LocalDateTime.now.getDayOfYear))
+      info.timestamp.map(_.toEpochSecond).getOrElse(0L) === System
+        .currentTimeMillis() / 1000 +- 20)
```
benthecarman: You should make the `20` here a variable defined above.

Also, it may need to be bumped up, there is CI a failure

https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/661726597#L560

PR #1229, opened 2020-03-13
2020 03 13 node md

Christewart:
This implements an example of how to use the node project on our website.

### DIFF
--- a/docs/applications/node.md
+++ b/docs/applications/node.md
@@ -3,11 +3,140 @@ id: node
 title: Light Client
 ---
 
-Bitcoin-S comes bundled with a light client Bitcoin node. this client
-is capable of doing verification of some parts of the blockchain,
-and can act as a starting point for building Bitcoin applications
-that need to connect to the P2P network.
+Bitcoin-s has node module that allows you to connect to the p2p network.
 
-This node is currently only released as a library, and not as a binary.
-This is because it (nor the documentation) is not deemed production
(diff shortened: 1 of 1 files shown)


docs/applications/node.md
```diff
@@ -3,11 +3,141 @@ id: node
...
+Bitcoin-s has node module that allows you to connect to the p2p network.
+
+### Neutrino Node
+
+Bitcoin-s has experimental support for neutrino which is a new lite client proposal on the bitcoin p2p network. You can
+read more about how neutrino works [here](https://suredbits.com/neutrino-what-is-it-and-why-we-need-it/).
+
+#### Callbacks
+
+We support call backs for the following events that happen on the bitcoin p2p network:
```
benthecarman: I think using "We" should be just replaced by Bitcoin-S

docs/applications/node.md
```diff
@@ -3,11 +3,141 @@ id: node
...
+```scala mdoc:compile-only
+
+implicit val system = ActorSystem(s"node-example")
+implicit val ec = system.dispatcher
+
+//we also require a bitcoind instance to connect to
+//so let's start one
+val instance = BitcoindRpcTestUtil.instance()
+val p2pPort = instance.p2pPort
+val bitcoindF = BitcoindRpcTestUtil.startedBitcoindRpcClient(instance)
```
Christewart, line 60: this isn't the right version of bitcoind, we need to make sure we get one that supports neutrino.

PR #1232, opened 2020-03-13
Bouncy Castle Fallback

nkohen:
When `Secp256k1Context.isEnabled` is `false` we now use `bouncycastle` as a fallback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,13 @@ matrix:
         - TEST_COMMAND="test:compile"
       scala:
         - 2.11.12
+    - os: linux
+      name: "Secp256k1 Disabled Core Test"
+      env:
+        - DISABLE_SECP256K1="true"
+        - TEST_COMMAND="coreTest/test"
+      scala:
+        - 2.13.1
     - os: osx
       name: "macOS bitcoind and eclair tests"
       env:
--- a/core-test/src/test/scala/org/bitcoins/core/crypto/BouncyCastleSecp256k1Test.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/crypto/BouncyCastleSecp256k1Test.scala
@@ -20,7 +20,23 @@ class BouncyCastleSecp256k1Test extends BitcoinSUnitTest {
     }
   }
 
-  it must "add keys the same" in {
+  it must "add private keys the same" in {
+    forAll(CryptoGenerators.privateKey, CryptoGenerators.privateKey) {
+      case (priv1, priv2) =>
+        val sumWithBouncyCastle =
+          BouncyCastleUtil.addNumbers(priv1.bytes, priv2.bytes)
+        val sumWithSecp = NativeSecp256k1.privKeyTweakAdd(priv1.bytes.toArray,
+                                                          priv2.bytes.toArray)
+
+        val sumKeyWithBouncyCastle =
+          ECPrivateKey(ByteVector(sumWithBouncyCastle.toByteArray))
+        val sumKeyWithSecp = ECPrivateKey(ByteVector(sumWithSecp))
+
+        assert(sumKeyWithBouncyCastle == sumKeyWithSecp)
+    }
+  }
+
+  it must "add public keys the same" in {
     forAll(CryptoGenerators.publicKey, CryptoGenerators.privateKey) {
       case (pubKey, privKey) =>
         val sumKeyBytes =
@@ -28,7 +44,7 @@ class BouncyCastleSecp256k1Test extends BitcoinSUnitTest {
                                          privKey.bytes.toArray,
                                          true)
         val sumKeyExpected = ECPublicKey.fromBytes(ByteVector(sumKeyBytes))
-        val sumKey = pubKey.add(privKey.publicKey) // This uses Bouncy Castle
+        val sumKey = pubKey.addWithBouncyCastle(privKey.publicKey)
 
         assert(sumKey == sumKeyExpected)
     }
--- a/core-test/src/test/scala/org/bitcoins/core/crypto/BouncyCastleSecp256k1Test.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/crypto/BouncyCastleSecp256k1Test.scala
@@ -1,0 +1,83 @@
+package org.bitcoins.core.crypto
+
+import org.bitcoin.{NativeSecp256k1, Secp256k1Context}
+import org.bitcoins.testkit.core.gen.{CryptoGenerators, NumberGenerator}
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.scalacheck.Gen
+import org.scalatest.{Outcome, Succeeded}
+import scodec.bits.ByteVector
+
+class BouncyCastleSecp256k1Test extends BitcoinSUnitTest {
+
+  behavior of "CryptoLibraries"
+
+  override def withFixture(test: NoArgTest): Outcome = {
+    if (Secp256k1Context.isEnabled) {
+      super.withFixture(test)
+    } else {
+      logger.warn(s"Test ${test.name} skipped as Secp256k1 is not available.")
+      Succeeded
+    }
+  }
+
+  it must "add keys the same" in {
+    forAll(CryptoGenerators.publicKey, CryptoGenerators.privateKey) {
+      case (pubKey, privKey) =>
+        val sumKeyBytes =
+          NativeSecp256k1.pubKeyTweakAdd(pubKey.bytes.toArray,
+                                         privKey.bytes.toArray,
+                                         true)
+        val sumKeyExpected = ECPublicKey.fromBytes(ByteVector(sumKeyBytes))
+        val sumKey = pubKey.add(privKey.publicKey) // This uses Bouncy Castle
+
+        assert(sumKey == sumKeyExpected)
+    }
+  }
+
+  it must "validate keys the same" in {
+    val keyOrGarbageGen = Gen.oneOf(CryptoGenerators.publicKey.map(_.bytes),
+                                    NumberGenerator.bytevector(33))
+    forAll(keyOrGarbageGen) { bytes =>
+      assert(
+        ECPublicKey.isFullyValidWithBouncyCastle(bytes) ==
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -30,6 +30,13 @@ matrix:
         - TEST_COMMAND="test:compile"
       scala:
         - 2.11.12
+    - os: linux
```
Christewart, line 33: i don't think i want to add another CI matrix for this as it does slow down build times with an entire `scalac` compile and `jvm` warmup. Especially for something as small as this. 

I'm not sure if we should test this on CI everytime.

secp256k1jni/src/main/java/org/bitcoin/Secp256k1Context.java
```diff
@@ -46,7 +48,12 @@
   }
 
   public static boolean isEnabled() {
-     return enabled;
+      String secpDisabled = System.getenv("DISABLE_SECP256K1");
+      if (secpDisabled != null && secpDisabled.equals("true")) {
```
Christewart: probably should convert `secpDisabled` to lowercase

core-test/src/test/scala/org/bitcoins/core/crypto/ECPublicKeyTest.scala
```diff
@@ -38,15 +38,20 @@ class ECPublicKeyTest extends BitcoinSUnitTest {
...
-    forAll(CryptoGenerators.publicKey, CryptoGenerators.privateKey) {
-      case (pubKey, privKey) =>
-        val sumKeyBytes = NativeSecp256k1.pubKeyTweakAdd(pubKey.bytes.toArray,
-                                                         privKey.bytes.toArray,
-                                                         true)
-        val sumKeyExpected = ECPublicKey.fromBytes(ByteVector(sumKeyBytes))
-        val sumKey = pubKey.add(privKey.publicKey)
-
-        assert(sumKey == sumKeyExpected)
+    if (Secp256k1Context.isEnabled) {
```
Christewart: why is this not tested on bouncy castle?

Christewart (commented):
So some thoughts on structure: 

I think we pass this in as a default parameter to these methods so it's easier to test bouncy castle against secp256k1. This means we can choose at the time we call the method which one we want to use: 


```scala
     override def sign(dataToSign: ByteVector, isSecp256k1Enabled: Boolean = Secp256k1Context.isEnabled): ECDigitalSignature = {
    require(dataToSign.length == 32 && bytes.length <= 32)
    if (isSecp256k1Enabled) {
      val signature =
        NativeSecp256k1.sign(dataToSign.toArray, bytes.toArray)
      ECDigitalSignature(ByteVector(signature))
    } else {
      signWithBouncyCastle(dataToSign)
    }
  }
```

Now when testing things, you can pass in `false` if you want to test bouncy castle vs secp256k1. 

Thoughts?

benthecarman:
> So some thoughts on structure:
> 
> I think we pass this in as a default parameter to these methods so it's easier to test bouncy castle against secp256k1. This means we can choose at the time we call the method which one we want to use:
> 
> ```scala
>      override def sign(dataToSign: ByteVector, isSecp256k1Enabled: Boolean = Secp256k1Context.isEnabled): ECDigitalSignature = {
>     require(dataToSign.length == 32 && bytes.length <= 32)
>     if (isSecp256k1Enabled) {
>       val signature =
>         NativeSecp256k1.sign(dataToSign.toArray, bytes.toArray)
>       ECDigitalSignature(ByteVector(signature))
>     } else {
>       signWithBouncyCastle(dataToSign)
>     }
>   }
> ```
> 
> Now when testing things, you can pass in `false` if you want to test bouncy castle vs secp256k1.
> 
> Thoughts?

I think current implementation is better, I don't think it should be the user's concern of which library is being used unless the explicitly want one enable for all function calls.

Christewart:
> > So some thoughts on structure:
> > I think we pass this in as a default parameter to these methods so it's easier to test bouncy castle against secp256k1. This means we can choose at the time we call the method which one we want to use:
> > ```scala
> >      override def sign(dataToSign: ByteVector, isSecp256k1Enabled: Boolean = Secp256k1Context.isEnabled): ECDigitalSignature = {
> >     require(dataToSign.length == 32 && bytes.length <= 32)
> >     if (isSecp256k1Enabled) {
> >       val signature =
> >         NativeSecp256k1.sign(dataToSign.toArray, bytes.toArray)
> >       ECDigitalSignature(ByteVector(signature))
> >     } else {
> >       signWithBouncyCastle(dataToSign)
> >     }
> >   }
> > ```
> > 
> > 
> > Now when testing things, you can pass in `false` if you want to test bouncy castle vs secp256k1.
> > Thoughts?
> 
> I think current implementation is better, I don't think it should be the user's concern of which library is being used unless the explicitly want one enable for all function calls.

That was my goal by having it be a arg with a default value, if you want to use bouncy castle you can opt into using it by passing `false`, otherwise you can just call the method as normal.

Do you have any suggestions for how to make sure there isn't regressions between our bouncy castle implementation of these methods and the secp256k1 implementation? I was imagining we could have something like 


```scala

val dataToSign = .... 
val key = ECPrivateKey.freshPrivateKey()
key.sign(dataToSign) == key.sign(dataToSign, isSecp256k1Enabled = false)

```

in the unit test.

secp256k1jni/src/main/java/org/bitcoin/Secp256k1Context.java
benthecarman: maybe allow it to be `1` as well

secp256k1jni/src/main/java/org/bitcoin/Secp256k1Context.java
```diff
@@ -18,6 +18,8 @@
 
 import org.scijava.nativelib.NativeLoader;
 
+import java.util.Properties;
```
benthecarman: unused

core-test/src/test/scala/org/bitcoins/core/crypto/ECPublicKeyTest.scala
nkohen: This test essentially compares secp to bouncy castle, and if we don't have secp there isn't much of a point in doing this

benthecarman (approved):

nkohen:
After further discussion offline with @Christewart, we've decided to split up the secp and bouncy castle implementations so that they can be tested against each other, but also to keep the environment variable and the CI matrix entry for now to ensure ease of use and extensive testing for the Bouncy Castle code.

core-test/src/test/scala/org/bitcoins/core/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  it must "add keys the same" in {
+    forAll(CryptoGenerators.publicKey, CryptoGenerators.privateKey) {
+      case (pubKey, privKey) =>
+        val sumKeyBytes =
+          NativeSecp256k1.pubKeyTweakAdd(pubKey.bytes.toArray,
+                                         privKey.bytes.toArray,
+                                         true)
+        val sumKeyExpected = ECPublicKey.fromBytes(ByteVector(sumKeyBytes))
+        val sumKey = pubKey.add(privKey.publicKey) // This uses Bouncy Castle
```
Christewart: can we just add a postfix to the `add` name for all BC stuff? I.e. `addBC()`

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -200,7 +202,12 @@ sealed abstract class ExtPrivateKey
     val (il, ir) = hmac.splitAt(32)
     //should be ECGroup addition
     //parse256(IL) + kpar (mod n)
-    val tweak = NativeSecp256k1.privKeyTweakAdd(il.toArray, key.bytes.toArray)
+    val tweak = if (Secp256k1Context.isEnabled) {
```
Christewart, line 203: Is this stuff tested? this is sensitive

core/src/main/scala/org/bitcoins/core/crypto/BouncyCastleUtil.scala
```diff
@@ -0,0 +1,110 @@
...
+
+import scala.util.Try
+
+object BouncyCastleUtil {
+
+  private val curve: ECCurve = CryptoParams.curve.getCurve
+  private val G: ECPoint = CryptoParams.curve.getG
+  private val N: BigInteger = CryptoParams.curve.getN
+
+  private def number(bytes: ByteVector): BigInteger = {
```
Christewart: please rename to `getBigInteger()`

secp256k1jni/src/main/java/org/bitcoin/Secp256k1Context.java
```diff
@@ -46,7 +46,14 @@
   }
 
   public static boolean isEnabled() {
```
Christewart, line 53: Let's add a javadoc to the top of this what the env var is for, and why it is useful

core/src/main/scala/org/bitcoins/core/crypto/BouncyCastleUtil.scala
```diff
@@ -0,0 +1,110 @@
...
+    new BigInteger(1, bytes.toArray)
+  }
+
+  def addNumbers(num1: ByteVector, num2: ByteVector): BigInteger = {
+    number(num1).add(number(num2)).mod(N)
```
Christewart:
```scala
val bigIneger1 = getBigInteger(num1)
val bigInteger2 = getBigInteger(num2)
bigInteger1.add(bigInteger2).mod(N)
```

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
Christewart: Also it looks like there is an unused import in this file:

```
[warn] /home/suredbits/dev/bitcoin-s/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala:3:18: Unused import
[warn] import java.math.BigInteger
[warn]                  ^
```

PR #1236, opened 2020-03-14
Allow FeeUnit to represent decimals

benthecarman:
The main change is in `FeeUnit.scala` and `FeeRate.scala` the rest is fixing everything to be compatible.

The thinking is:

- remove `currencyUnit` because this cannot represent decimals
- add `baseAmount` this is the representation in satoshis
- fix `calc` so it correctly calculates the fee

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
@@ -532,7 +532,7 @@ object JsonReaders {
   implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeRate] {
     override def reads(json: JsValue): JsResult[BitcoinFeeRate] =
       SerializerUtil.processJsNumber[BitcoinFeeRate](num =>
-        SatoshisPerByte((num * 100000).toDouble))(json)
+        SatoshisPerByte(num * 100000.0))(json)
   }
 
   implicit object FileReads extends Reads[File] {
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
@@ -532,7 +532,7 @@ object JsonReaders {
   implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeUnit] {
     override def reads(json: JsValue): JsResult[BitcoinFeeUnit] =
       SerializerUtil.processJsNumber[BitcoinFeeUnit](num =>
-        SatoshisPerByte(Satoshis((num * 100000).toBigInt)))(json)
+        SatoshisPerByte((num * 100000).toDouble))(json)
   }
 
   implicit object FileReads extends Reads[File] {
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
@@ -25,7 +25,7 @@ import org.bitcoins.core.protocol.{
 }
 import org.bitcoins.core.script.ScriptType
 import org.bitcoins.core.script.crypto.HashType
-import org.bitcoins.core.wallet.fee.{BitcoinFeeRate, SatoshisPerByte}
+import org.bitcoins.core.wallet.fee.BitcoinsPerKiloByte
 import org.bitcoins.rpc.client.common.RpcOpts.LabelPurpose
 import org.bitcoins.rpc.jsonmodels._
 import org.bitcoins.rpc.serializers.JsonSerializers._
@@ -529,10 +529,10 @@ object JsonReaders {
   }
 
   // Currently takes in BTC/kB
-  implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeRate] {
-    override def reads(json: JsValue): JsResult[BitcoinFeeRate] =
-      SerializerUtil.processJsNumber[BitcoinFeeRate](num =>
-        SatoshisPerByte(num * 100000.0))(json)
+  implicit object BitcoinsPerKiloByteReads extends Reads[BitcoinsPerKiloByte] {
+    override def reads(json: JsValue): JsResult[BitcoinsPerKiloByte] =
+      SerializerUtil.processJsNumber[BitcoinsPerKiloByte](num =>
+        BitcoinsPerKiloByte(num))(json)
   }
 
   implicit object FileReads extends Reads[File] {
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
@@ -25,7 +25,7 @@ import org.bitcoins.core.protocol.{
 }
 import org.bitcoins.core.script.ScriptType
 import org.bitcoins.core.script.crypto.HashType
-import org.bitcoins.core.wallet.fee.{BitcoinFeeUnit, SatoshisPerByte}
+import org.bitcoins.core.wallet.fee.{BitcoinFeeRate, SatoshisPerByte}
 import org.bitcoins.rpc.client.common.RpcOpts.LabelPurpose
 import org.bitcoins.rpc.jsonmodels._
 import org.bitcoins.rpc.serializers.JsonSerializers._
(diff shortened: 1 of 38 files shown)


core-test/src/test/scala/org/bitcoins/core/wallet/builder/BitcoinTxBuilderTest.scala
```diff
@@ -60,14 +60,14 @@ class BitcoinTxBuilderTest extends BitcoinSAsyncTest {
     val listF =
       BitcoinTxBuilder(destinations = Vector.empty,
                        utxos = List(utxo),
-                       feeRate = SatoshisPerByte(1.sat),
+                       feeRate = SatoshisPerByte(1),
```
Christewart: nit: we really should just create a constant for this like

```scala
object SatoshisPerByte { 
  val one = ... 
}
```

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -1,47 +1,43 @@
...
+  def baseAmount: Double
 }
 
 /**
   * Meant to represent the different fee unit types for the bitcoin protocol
   * @see [[https://en.bitcoin.it/wiki/Weight_units]]
   */
-sealed abstract class BitcoinFeeUnit extends FeeUnit
+sealed abstract class BitcoinFeeUnit extends FeeUnit {
+  def sats: Double
```
Christewart: So I don't think it makes sense to have `sats: Double` as you can't have sats be a `Double` in the bitcoin protocol. 

If you go to other projects, and they allow you to select a fee, do they let you select `1.5` satoshis/vbyte as a option? 

There seems to be a `desiredFee` and `actualFee` or something like that.

Christewart: Thinking about this more, it seems what we don't represent well is a fee-rate, rather than an absolute fee. An absolute fee cannot be a double (at least in the sats unit). However a fee rate most definitely can.

It seems to me we are using the `FeeUnit` type for both at the moment, when it reality we maybe should have a `FeeUnit` type and a `FeeRate` type? Thoughts?

benthecarman: > If you go to other projects, and they allow you to select a fee, do they let you select 1.5 satoshis/vbyte as a option?

I know electrum and wasabi allow you to, I haven't used many other wallets.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala
```diff
@@ -69,7 +69,7 @@ object JsonSerializers {
   implicit val transactionReads: Reads[Transaction] = TransactionReads
   implicit val transactionOutPointReads: Reads[TransactionOutPoint] =
     TransactionOutPointReads
-  implicit val bitcoinFeeUnitReads: Reads[BitcoinFeeUnit] = BitcoinFeeUnitReads
+  implicit val bitcoinFeeUnitReads: Reads[BitcoinFeeRate] = BitcoinFeeUnitReads
```
Christewart: this isn't directly related to this PR, but this serializer is dangerous as it allows us to mis-type a fee unit given back from bitcoind i believe. Should this be typed as `Reads[SatoshisPerKiloByte]`

Christewart (commented):
Let's add some unit tests for the `FeeRate` stuff.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -529,10 +529,10 @@ object JsonReaders {
...
 
   // Currently takes in BTC/kB
-  implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeUnit] {
-    override def reads(json: JsValue): JsResult[BitcoinFeeUnit] =
-      SerializerUtil.processJsNumber[BitcoinFeeUnit](num =>
-        SatoshisPerByte(Satoshis((num * 100000).toBigInt)))(json)
+  implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeRate] {
+    override def reads(json: JsValue): JsResult[BitcoinFeeRate] =
+      SerializerUtil.processJsNumber[BitcoinFeeRate](num =>
+        SatoshisPerByte((num * 100000).toDouble))(json)
```
nkohen: should probably use `.toBigDecimal` no? We want to make sure we don't overflow

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonSerializers.scala
```diff
@@ -153,7 +153,7 @@ object JsonSerializers {
 
       def reads(json: JsValue): JsResult[SatoshisPerKiloByte] =
         SerializerUtil.processJsNumber(num =>
-          SatoshisPerKiloByte(Satoshis(num.toBigInt)))(json)
+          SatoshisPerKiloByte(num.toDouble))(json)
```
nkohen: ditto

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -1,60 +1,18 @@
...
-  * of a [[org.bitcoins.core.protocol.transaction.WitnessTransaction]]
-  */
-case class SatoshisPerVirtualByte(currencyUnit: CurrencyUnit)
-    extends BitcoinFeeUnit
+  final case object PerKiloByte extends FeeUnit
 
-object SatoshisPerVirtualByte {
-  val zero: SatoshisPerVirtualByte = SatoshisPerVirtualByte(CurrencyUnits.zero)
-  val one: SatoshisPerVirtualByte = SatoshisPerVirtualByte(Satoshis.one)
+  final case object Flat extends FeeUnit
```
nkohen, line 21: what does this mean? Scaladoc

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CurrencyUnitGenerator.scala
```diff
@@ -18,28 +18,28 @@ trait CurrencyUnitGenerator {
...
+  def feeRate: Gen[FeeRate] =
     Gen.oneOf(satsPerByte, satsPerKiloByte, satsPerVirtualByte)
 
-  /** Generates a FeeUnit based on the maxFee allowed for a transaction */
-  def feeUnit(maxFee: Long): Gen[FeeUnit] = {
+  /** Generates a FeeRate based on the maxFee allowed for a transaction */
+  def feeRate(maxFee: Long): Gen[FeeRate] = {
     Gen.choose(0L, maxFee / 10000L).map { n =>
-      SatoshisPerKiloByte(Satoshis(n))
+      SatoshisPerKiloByte(n.toDouble)
```
nkohen: nit: make this `maxFee / 10000.0` to avoid `toDouble` call

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeRate.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.bitcoins.core.protocol.transaction.Transaction
+
+/**
+  * This is meant to be an abstract type that represents different fee rates for
+  * blockchains
+  */
+sealed abstract class FeeRate {
+  def *(tx: Transaction): CurrencyUnit = calc(tx)
+  def calc(tx: Transaction): CurrencyUnit
+  def baseAmount: Double
```
nkohen: Is it safe for us to use `Double`? (what are the protections against overflow, does it lose precision or fail or?)

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeRate.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  override def units: FeeUnit = FeeUnit.PerVirtualByte
+}
+
+object SatoshisPerVirtualByte {
+  val zero: SatoshisPerVirtualByte = SatoshisPerVirtualByte(0)
+  val one: SatoshisPerVirtualByte = SatoshisPerVirtualByte(1)
+}
+
+case class FlatSatoshis(satoshis: Long) extends BitcoinFeeRate {
```
nkohen, line 176: scaladoc

Christewart:
Looks like CI is failing here: https://travis-ci.org/github/bitcoin-s/bitcoin-s/builds/665029427?utm_source=github_status&utm_medium=notification

benthecarman:
Error ended up being because I required the fee rate cannot be negative but some of our tests use a negative fee rate.

nkohen (commented):
LGTM but I'd still like some assurances somewhere that using `Double` everywhere isn't prone to overflow

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/serializers/JsonReaders.scala
```diff
@@ -529,10 +529,10 @@ object JsonReaders {
   }
 
   // Currently takes in BTC/kB
-  implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeUnit] {
-    override def reads(json: JsValue): JsResult[BitcoinFeeUnit] =
-      SerializerUtil.processJsNumber[BitcoinFeeUnit](num =>
-        SatoshisPerByte(Satoshis((num * 100000).toBigInt)))(json)
+  implicit object BitcoinFeeUnitReads extends Reads[BitcoinFeeRate] {
```
Christewart: Didn't we decide to change this type to `Reads[BTCPerKiloByte]` so we don't use the wrong implicit? 

https://bitcoin.org/en/developer-reference#estimatesmartfee

Also it is concerning that unit tests aren't failing when we use the wrong implicit here?

PR #1237, opened 2020-03-15
Add disclaimers to the wallet/key-manager around API breaking changes…

Christewart:
… and limitations of the current iteration of the wallet

We will be working on the wallet/key-manager project a lot for the next release of bitcoin-s, add disclaimers saying that the current projects structure will change.

### DIFF
--- a/docs/wallet/wallet.md
+++ b/docs/wallet/wallet.md
@@ -13,6 +13,10 @@ This wallet is currently only released as a library, and not as a binary.
 This is because it (nor the documentation) is not deemed production
 ready. Use at your own risk, and without too much money depending on it.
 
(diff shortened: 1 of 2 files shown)


docs/wallet/wallet.md
```diff
@@ -13,6 +13,10 @@ This wallet is currently only released as a library, and not as a binary.
 This is because it (nor the documentation) is not deemed production
 ready. Use at your own risk, and without too much money depending on it.
 
+### Disclaimer
```
benthecarman, line 16: > and surprising behavior in the current wallet.

I think this would be better "and surprising behavior from the current wallet."

PR #1247, opened 2020-03-17
2020 03 15 chainhandler getnancestor

### LINKED ISSUE #806: Filter header batch size does not match block header batch size 1959 != 1960
When starting my node and receiving filters i get this error:

https://pastebin.com/FLMxB50J

Christewart:
This PR addresses #806 

Previously when fetching headers from the database, we would just select headers between a height range and then return the headers in this range from `BlockHeaderDAO.getNChildren()`. 

This doesn't work because we can have multiple block headers at the same height. This PR does two things: 

1. Renames `getNChildren()` -> `getNAncestors()` as that is what we are actually doing
2. Calls `Blockchain.connectWalkBackwards()` on the headers returned from `BlockHeaderDAO.getNAncestors()` to validate that we can actually form a blockchain with what is returned from the database.

closes #806

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -160,18 +160,9 @@ case class BlockHeaderDAO()(
       }
     } yield {
       if (headerOpt.isDefined) {
-        val blockchains =
-          Blockchain.reconstructFromHeaders(headerOpt.get, headers)
-        require(
-          blockchains.length == 1 || blockchains.isEmpty,
-          s"Can only have one blockchain when walking backwards, got=${blockchains.length}")
-        blockchains.headOption match {
-          case Some(h) => h.toVector
-          case None    =>
-            //if we couldn't build a blockchain at all,
-            //form a trivial with the given childHash
-            Vector(headerOpt.get)
-        }
+        val connectedHeaders =
+          Blockchain.connectWalkBackwards(headerOpt.get, headers)
+        connectedHeaders.reverse
       } else {
         Vector.empty
       }
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -142,9 +142,10 @@ case class BlockHeaderDAO()(
   }
 
   /** Gets Block Headers of all children starting with the given block hash (inclusive), could be out of order */
-  def getNChildren(
+  def getNAncestors(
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
```diff
@@ -25,7 +25,7 @@ object BytesToPushOntoStack
     override val opCode = num
   }
 
-  override def operations: Seq[BytesToPushOntoStack] =
+  override val operations: Seq[BytesToPushOntoStack] =
```
Christewart, line 28: this is a quick optimization so we don't re-create `BytesToPushOntoStack` everytime we parse a script.

benthecarman: Should this be `lazy`?

Christewart: nah, it doesn't really matter `lazy` or `val`, the problem is it being a `def` so it has to be recreated _everytime_

chain/src/main/scala/org/bitcoins/chain/pow/Pow.scala
```diff
@@ -42,10 +42,16 @@ sealed abstract class Pow {
...
-                throw new RuntimeException(
-                  s"Could not find non mindiffulty block in chain! hash=${tip.hashBE.hex} height=${currentHeight}")
+              case None =>
+                config.chain match {
+                  case RegTestNetChainParams =>
+                    RegTestNetChainParams.compressedPowLimit
+                  case TestNetChainParams | MainNetChainParams =>
+                    //if we can't find a non min diffulty block, let's just fail
+                    throw new RuntimeException(
+                      s"Could not find non mindiffulty block in chain of size=${blockchain.length}! hash=${tip.hashBE.hex} height=${currentHeight}")
```
rorp: A typo in "mindiffulty" ?

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -141,20 +141,31 @@ case class BlockHeaderDAO()(
...
       }
     } yield {
-      res
+      if (headerOpt.isDefined) {
+        val connectedHeaders =
+          Blockchain.connectWalkBackwards(headerOpt.get, headers)
+        connectedHeaders.reverse
+      } else {
+        Vector.empty
+      }
```
rorp: `headerOpt.get` looks a bit brutal. Maybe`headerOpt.map(...).getOrElse(Vector.empty)`?

PR #1249, opened 2020-03-17
Multi-Outcome DLCs

nkohen:
Reworked BinaryOutcomeDLCClient into DLCClient which allows for multiple outcomes.

This PR does not implement tests for multiple outcomes, but only re-writes to create support for it in the future.


benthecarman:
lgtm so far

benthecarman (approved):

PR #1255, opened 2020-03-18
Tests for Multi-Outcome DLCs

nkohen:
Added unit and integration tests for multi-outcome DLCs

### DIFF
--- a/dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
+++ b/dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
@@ -36,6 +36,7 @@ import org.bitcoins.core.script.crypto.HashType
 import org.bitcoins.core.util.{CryptoUtil, FutureUtil}
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.core.wallet.utxo.P2WPKHV0SpendingInfo
+import org.bitcoins.dlc.DLCMessage.ContractInfo
 import org.bitcoins.rpc.BitcoindException
 import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
 import org.bitcoins.testkit.util.BitcoindRpcTest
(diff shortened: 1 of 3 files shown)


dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
```diff
@@ -672,52 +679,70 @@ class DLCClientIntegrationTest extends BitcoindRpcTest {
     }
   }
 
-  it should "be able to publish all DLC txs to Regtest for the mutual Win case" in {
-    for {
-      _ <- executeForMutualCase(outcomeWinHash, local = true)
-      _ <- executeForMutualCase(outcomeWinHash, local = false)
-    } yield succeed
+  val numOutcomesToTest: Vector[Int] = Vector(2, 8) //, 100)
```
benthecarman: > //, 100)

ping

dlc-test/src/test/scala/org/bitcoins/dlc/DLCTestUtil.scala
```diff
@@ -0,0 +1,26 @@
+package org.bitcoins.dlc
+
+import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
+
+object DLCTestUtil {
+
+  def genOutcomes(size: Int): Vector[String] = {
```
benthecarman: This should do the hashing for us since we will almost always hash the strings afterwards

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
```diff
@@ -203,9 +197,18 @@ class DLCClientIntegrationTest extends BitcoindRpcTest {
...
       val remoteVout = remoteFundingUtxos.head.outPoint.vout
 
+      val outcomes = DLCTestUtil.genOutcomes(numOutcomes)
+      val outcomeHashes =
+        outcomes.map(msg => CryptoUtil.sha256(ByteVector(msg.getBytes)).flip)
+
+      val outcomeMap =
+        outcomeHashes.zip(DLCTestUtil.genValues(numOutcomes, totalInput)).toMap
+      val otherOutcomeMap = outcomeMap.map {
+        case (hash, amt) => (hash, (totalInput - amt).satoshis)
```
benthecarman: It'd be nice this if this was it's own generator so we could just do `DLCTestUtil.genContractInfos`

benthecarman: Also will be extra nice for when we *eventually* have multiparty dlcs.

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
benthecarman: Also maybe instead of 2 & 8, do something like 2-8, incase there is a weird bug for a certain number of outcomes

benthecarman (commented):
Looks good overall, couple nits on the generators

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientIntegrationTest.scala
nkohen: I think this should be covered by unit tests, the integration tests take over a full minute as is (will add the case for 100 back in of course)

benthecarman (approved):

PR #1259, opened 2020-03-19
DLC Suredbits Client

nkohen:
Added dlc-suredbits-client project inside of app directory and implemented methods to retrieve testnet data from the Suredbits Oracle given an EclairAPI

### DIFF
--- a/app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockEclairClient.scala
+++ b/app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockEclairClient.scala
@@ -12,6 +12,7 @@ import akka.stream.scaladsl.{
   Source,
   SourceQueueWithComplete
 }
+import org.bitcoins.commons.jsonmodels.eclair._
 import org.bitcoins.core.crypto.{ECPrivateKey, ECPublicKey, Sha256Digest}
 import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
 import org.bitcoins.core.number.UInt32
@@ -24,7 +25,7 @@ import org.bitcoins.core.protocol.ln.node.NodeId
 import org.bitcoins.core.protocol.script.ScriptPubKey
 import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
-import org.bitcoins.eclair.rpc.api._
+import org.bitcoins.eclair.rpc.api.EclairApi
 import org.bitcoins.eclair.rpc.network.NodeUri
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -417,8 +418,7 @@ class MockEclairClient()(implicit system: ActorSystem)
     unsupportedFailure
 
   override def connectToWebSocket(
-      eventHandler: org.bitcoins.eclair.rpc.api.WebSocketEvent => Unit): scala.concurrent.Future[
-    Unit] = {
+      eventHandler: WebSocketEvent => Unit): scala.concurrent.Future[Unit] = {
     unsupportedFailure
   }
 
--- a/app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockEclairClient.scala
+++ b/app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockEclairClient.scala
@@ -2,16 +2,8 @@ package org.bitcoins.sbclient
 
 import java.net.InetSocketAddress
 import java.time.Instant
+import java.util.concurrent.locks.ReentrantReadWriteLock
 
-import akka.actor.ActorSystem
-import akka.stream.{ActorMaterializer, OverflowStrategy, QueueOfferResult}
-import akka.stream.scaladsl.{
-  Keep,
-  Sink,
-  SinkQueueWithCancel,
-  Source,
-  SourceQueueWithComplete
-}
 import org.bitcoins.commons.jsonmodels.eclair._
 import org.bitcoins.core.crypto.{ECPrivateKey, ECPublicKey, Sha256Digest}
 import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
@@ -23,7 +15,7 @@ import org.bitcoins.core.protocol.ln.channel.{ChannelId, FundedChannelId}
 import org.bitcoins.core.protocol.ln.currency.{LnCurrencyUnits, MilliSatoshis}
 import org.bitcoins.core.protocol.ln.node.NodeId
 import org.bitcoins.core.protocol.script.ScriptPubKey
-import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.core.util.{BitcoinSLogger, CryptoUtil}
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.eclair.rpc.api.EclairApi
 import org.bitcoins.eclair.rpc.network.NodeUri
@@ -31,30 +23,17 @@ import org.bitcoins.eclair.rpc.network.NodeUri
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.{FiniteDuration, _}
 
-class MockEclairClient()(implicit system: ActorSystem)
+class MockEclairClient()(
+    implicit override val executionContext: ExecutionContext)
     extends EclairApi
     with BitcoinSLogger {
-  implicit private val m: ActorMaterializer = ActorMaterializer.create(system)
-
-  implicit override val executionContext: ExecutionContext = system.dispatcher
-
   private def now: Instant = Instant.now
 
-  private val paymentSource: Source[
-    LnInvoice,
-    SourceQueueWithComplete[
-      LnInvoice
-    ]] =
-    Source.queue(16, OverflowStrategy.backpressure)
+  private val readWriteLock = new ReentrantReadWriteLock
+  private val readLock = readWriteLock.readLock
+  private val writeLock = readWriteLock.writeLock
 
-  private val paymentSink: Sink[LnInvoice, SinkQueueWithCancel[LnInvoice]] =
-    Sink.queue()
-
-  private val queues = paymentSource.toMat(paymentSink)(Keep.both).run()
-
-  private val paymentSender: SourceQueueWithComplete[LnInvoice] = queues._1
-
-  private val sentPaymentQueue: SinkQueueWithCancel[LnInvoice] = queues._2
+  private var lastSentPayment: Option[OutgoingPayment] = None
 
   private var paymentsById =
     Map.empty[PaymentId, Vector[OutgoingPayment]].withDefaultValue(Vector.empty)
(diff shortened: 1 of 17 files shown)


app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Client.scala
```diff
@@ -0,0 +1,130 @@
...
+    Json.reads[InvoiceAndDataResponse]
+
+  def request(
+      exchange: Exchange,
+      tradingPair: TradingPair,
+      requestType: RequestType)(
+      implicit system: ActorSystem): Future[InvoiceAndDataResponse] = {
+    implicit val ec: ExecutionContextExecutor = system.dispatcher
+
+    val prefix = "https://test.api.suredbits.com/dlc/v0"
```
benthecarman: This should probably be part of the config or at least defined somewhere higher up

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Asset.scala
```diff
@@ -0,0 +1,44 @@
...
+  def fromLowerString(str: String): Asset = {
+    val name = if (str.toLowerCase == "xbt") "btc" else str
+    val assetOpt = all.find(asset => asset.toLowerString == name)
+    assetOpt.getOrElse(OtherAsset(name.toUpperCase))
+  }
+
+  def fromUpperString(str: String): Asset = {
+    val name = if (str.toUpperCase == "XBT") "BTC" else str
+    val assetOpt = all.find(asset => asset.toUpperString == name)
+    assetOpt.getOrElse(OtherAsset(name))
```
benthecarman, line 35: Do we want it to not throw errors where we don't get an already defined `Asset`? I feel like this could be prone to mistakes.

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Client.scala
```diff
@@ -0,0 +1,130 @@
...
+      }
+  }
+
+  def makePayment(eclairApi: EclairApi, invoice: LnInvoice)(
+      implicit ec: ExecutionContext): Future[PaymentPreimage] = {
+    for {
+      payment <- eclairApi.payAndMonitorInvoice(invoice,
+                                                externalId = None,
+                                                interval = 500.milliseconds,
+                                                maxAttempts = 10)
```
benthecarman: 5 seconds might to be too short, maybe bump it up or at least make it an argument for the function.

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Asset.scala
nkohen: I think it is a nicer interface to have `TradingPair.fromString("BTCWTF")` return `UnsupportedTradingPair(Asset.BTC, OtherAsset("WTF"))` and then have the server give the error (which it does). `OtherAsset` is essentially like a `Left` in an `Either`

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Client.scala
rorp: It could be passed as a constructor parameter along with the `ActorSystem`. The latter would help to avoid issues to mixed up actor systems and/or execution contexts.

benthecarman (approved):

Christewart (commented):
Overall it looks good to me, however it would be nice to have some sort of unit tests

build.sbt
```diff
@@ -245,6 +245,11 @@ lazy val cliTest = project
     testkit
   )
 
+lazy val dlcSuredbitsClient = project
```
Christewart, line 261: Make sure you add this to `aggregate()` on `bitcoin-s` so tasks run at the root level are run in the subproject too: https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Aggregation

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockEclairClient.scala
```diff
@@ -0,0 +1,536 @@
+package org.bitcoins.sbclient
```
Christewart: This should really be put inside of `testkit`

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,43 @@
+package org.bitcoins.sbclient
```
Christewart, line 1: same thing with this, we probably want to re-use this in testkit

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/Exchange.scala
```diff
@@ -0,0 +1,119 @@
+package org.bitcoins.sbclient
```
Christewart, line 1: We should probably just open source this stuff in a nice library we don't have to update this internally and externally everytime.

nkohen: problem is internally there is a lot more to an `Exchange` than just what is here, not sure we can reuse

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
nkohen: it is now pretty dlc client specific, but should another similar thing be needed I'm sure that some super type can be factored out into testkit

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,144 @@
...
+        description = s"Mock Invoice",
+        price.toMSat
+      )
+
+    invoiceF.map { invoice =>
+      val preimage = preImage(invoice)
+
+      val dataBytes: ByteVector = ByteVector.encodeUtf8(data) match {
+        case Left(err) =>
+          throw new RuntimeException(s"Could not UTF8 encode data!", err)
```
benthecarman: nit: Could not encode to UTF-8

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,144 @@
...
+                                          port = port)
+      val _ = serverBindingP.success(bindingF)
+      bindingF
+    }
+  }
+
+  def stop(): Future[Done] = {
+    if (stoppedP.isCompleted || !serverBindingP.isCompleted) {
+      Future.failed(new RuntimeException(
+        "Cannot stop server of it has not been started or has already been stopped"))
```
benthecarman: this sentence doesn't make sense

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/RegexUtil.scala
```diff
@@ -0,0 +1,36 @@
...
+    * Util.noCaseOrRegex(Vector(A, BC, DEF)) == "(?i)(?:DEF|BC|A)".r
+    */
+  def noCaseOrRegex[T](all: Seq[T]): Regex = {
+    val orStr = all
+      .sortBy(_.toString.length)(Ordering.Int.reverse)
+      .foldLeft("") {
+        case (rx, t) =>
+          rx + t + "|"
+      }
+      .dropRight(1)
```
benthecarman: nit: init

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
nkohen: not sure what doesn't make sense about it?

benthecarman: > Cannot stop server of it has

benthecarman (approved):

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,144 @@
...
+        complete(HttpEntity(ContentTypes.`text/html(UTF-8)`, "<h1>pong</h1>"))
+      }
+    }
+  }
+
+  /**
+    * Generates a random port not in use
+    */
+  @tailrec
+  private def randomPort: Int = {
```
Christewart: Probably should just use `RpcUtil.randomPort()`

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,144 @@
...
+      case Success(value) => value
+      case Failure(_)     => randomPort
+    }
+  }
+
+  val port: Int = randomPort
+
+  val endpoint: String = s"http://localhost:$port"
+
+  private val serverBindingP = Promise[Future[Http.ServerBinding]]()
```
Christewart, line 82: Why `Promise[Future[]]`?

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
```diff
@@ -0,0 +1,144 @@
...
+      val bindingF = Http().bindAndHandle(handler = route,
+                                          interface = "localhost",
+                                          port = port)
+      val _ = serverBindingP.success(bindingF)
+      bindingF
+    }
+  }
+
+  def stop(): Future[Done] = {
+    if (stoppedP.isCompleted || !serverBindingP.isCompleted) {
```
Christewart, line 110: nit: No harm in calling `.stop()` twice?

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
```diff
@@ -0,0 +1,137 @@
...
+      payment <- eclairApi.payAndMonitorInvoice(invoice,
+                                                externalId = None,
+                                                interval = timeout / 10,
+                                                maxAttempts = 10)
+    } yield payment.status match {
+      case OutgoingPaymentStatus.Succeeded(preImage, _, _, _) => preImage
+      case OutgoingPaymentStatus.Failed(errs) =>
+        val errMsgs = errs.map(_.failureMessage).mkString(",\n")
+        throw new RuntimeException(s"Payment failed: $errMsgs")
+      case OutgoingPaymentStatus.Pending =>
```
Christewart, line 83: Hmm... do we want to to do this here? I think we should fail, but we might want a more descriptive error message

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+  }
+
+  def requestAndPay(
+      exchange: Exchange,
+      tradingPair: TradingPair,
+      requestType: RequestType,
+      eclairApi: EclairApi,
+      endpoint: String = "https://test.api.suredbits.com/dlc/v0")(
```
Christewart, line 106: maybe make this configurable in the `reference.conf` file?

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
nkohen: Because a `Future[Http.ServerBinding]` is returned when we bind the server

app/dlc-suredbits-client-test/src/test/scala/org/bitcoins/sbclient/MockServer.scala
nkohen: perhaps, but I think it is best to fail on un-intended behavior as I believe the person calling `stop` should expect the server to not already be stopped, and probably has bugs if it is. If on the other hand they explicitly want to call stop and have no expectation, they can always do `val _ = stop()`

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
nkohen: I think at this layer of abstraction, we should take in a `String` (likely without a default value in the future) but at some point, I think it would make sense to have the layer above this read an endpoint from a conf

PR #1261, opened 2020-03-20
Have DLC Wallet tests use multiple outcomes

benthecarman:


testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala
```diff
@@ -129,8 +130,24 @@ trait DLCWalletUtil {
     val walletA = fundedWalletA.wallet
     val walletB = fundedWalletB.wallet
 
+    val numOutcomes = scala.util.Random.nextInt(20) + 2
```
nkohen: I think it is probably better to just choose some number like `8`, either that or turn some `val`s into `def`s taking `numOutcomes: Int` as an input

wallet-test/src/test/scala/org/bitcoins/wallet/dlc/DLCExecutionTest.scala
```diff
@@ -30,9 +35,42 @@ class DLCExecutionTest extends BitcoinSDualWalletTest {
...
+      require(all.size == 1, "There should only be one dlc initialized")
+      all.head
+    }
+  }
+
+  def getSigs(contractInfo: ContractInfo): (
+      SchnorrDigitalSignature,
+      SchnorrDigitalSignature) = {
+
+//    implicit val ciOrdering: Ordering[(Sha256DigestBE, Satoshis)] =
```
nkohen: ping

nkohen (approved):
ACK on CI pass

PR #1268, opened 2020-03-23
CLI command to stop node

### LINKED ISSUE #1262: Add 'stop' bitcoin-s-cli command to cleanly shut things down
`bitcoin-s-cli stop` should cleanly shutdown the server.

benthecarman:
Closes #1262

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
@@ -8,12 +8,22 @@ import org.bitcoins.node.Node
 
 case class NodeRoutes(node: Node)(implicit system: ActorSystem)
     extends ServerRoute {
-  implicit val materializer = ActorMaterializer()
+  import system.dispatcher
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
 
   def handleCommand: PartialFunction[ServerCommand, StandardRoute] = {
     case ServerCommand("getpeers", _) =>
       complete {
         Server.httpSuccess("TODO implement getpeers")
+      }
+
+    case ServerCommand("stop", _) =>
+      complete {
+        node.stop().map { _ =>
+          val success = Server.httpSuccess("Node successfully stopped")
+          system.terminate()
+          success
+        }
       }
   }
 }
--- a/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
@@ -19,11 +19,11 @@ case class NodeRoutes(node: Node)(implicit system: ActorSystem)
 
     case ServerCommand("stop", _) =>
       complete {
-        node.stop().map { _ =>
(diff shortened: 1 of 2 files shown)


app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
```diff
@@ -8,12 +8,22 @@ import org.bitcoins.node.Node
...
 
   def handleCommand: PartialFunction[ServerCommand, StandardRoute] = {
     case ServerCommand("getpeers", _) =>
       complete {
         Server.httpSuccess("TODO implement getpeers")
       }
+
+    case ServerCommand("stop", _) =>
+      complete {
+        node.stop().map { _ =>
```
Christewart: So a couple things here: 

1. If `node.stop()` fails, nothing more is shutdown. I don't think we want thie behavior.
2. This `stop` command will return before the server is actually shutdown. Is this what we want? I don't think we can serve a response if we shutdown the server first, so it's a weird catch 22. I think we want a `Await.result()` somewhere, but i'm not sure where.
3. This probably should be implemented in `ServerRoute`, so this can be used across all projects. We don't want this just available in `NodeRoutes`.

rorp (commented):
I think the server should perform some security checks before stopping the node. Like verifying a master password, or trusted IP addresses, or etc.

Christewart:
> I think the server should perform some security checks before stopping the node. Like verifying a master password, or trusted IP addresses, or etc.

I think we should punt on this until we actually bake in some real authorization into bitcoin-s.

app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
benthecarman: 1. Agreed
2. If we add an `Await.result()` it needs to be outside the `complete` so the cli doesn't get a "Request failed: Future timed out after [X seconds]". However, `complete` is not a future, it runs asynchronously AFAIK so I'm not sure if this possible
3. I thought about doing this originally but not sure if it really makes sense in practicality. We have 4 different `XRoutes`: Node, Core, Chain, and Wallet. Node is the only one here that is actually an active process, the rest are just single api calls that update a database, there is no running process to actually stop.

rorp:
> I think we should punt on this until we actually bake in some real authorization into bitcoin-s.

I don't know why anybody would need it, because it's a huge security hole. But we can enable/disable in the config at least (disabled by default).

Christewart:
> > I think we should punt on this until we actually bake in some real authorization into bitcoin-s.
> 
> I don't know why anybody would need it, because it's a huge security hole. But we can enable/disable in the config at least (disabled by default).

I think this is a pretty common feature in these daemons, bitcoin core supports this: https://bitcoin.org/en/developer-reference#stop

There of course is the question of firewalls, authorization that should be used to protect this, but that should really apply for _every_ rpc command not just this one no?

benthecarman:
We already have `sendtoaddress` with no authorization, so I think this is much less of a security hole, but it does need to be addressed.

PR #1273, opened 2020-03-25
Custom fee rate for wallet sends

benthecarman:


app/cli/src/main/scala/org/bitcoins/cli/CliReaders.scala
```diff
@@ -44,6 +45,14 @@ object CliReaders {
       val reads: String => Bitcoins = str => Bitcoins(BigDecimal(str))
     }
 
+  implicit val satoshisPerVirtualByteReads: Read[SatoshisPerVirtualByte] =
+    new Read[SatoshisPerVirtualByte] {
+      val arity: Int = 1
+
+      val reads: String => SatoshisPerVirtualByte = str =>
+        SatoshisPerVirtualByte(Satoshis(BigInt(str)))
```
nkohen, line 53: doesn't this clash with your other PR that moves things to Doubles?

benthecarman: Yes

Christewart (commented):
~Shouldn't we be updating documentation somewhere to say that the fee unit we are using is sat/vbyte?~ oops looks like i missed this on the first pass.

Christewart (approved):

PR #1274, opened 2020-03-27
Broadcast DLC transactions

benthecarman:
DLC functions now broadcast the transactions by default, you have the option to do `--noBroadcast` to get the raw transaction.

I made it so you need to `getdlcfundingtx` for the tx broadcast rather than having `adddlcsigs` do it, curious on thoughts about which option is desirable.

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -263,7 +263,8 @@ object ConsoleCli {
         ),
       cmd("acceptdlcmutualclose")
         .hidden()
-        .action((_, conf) => conf.copy(command = AcceptDLCMutualClose(null)))
+        .action((_, conf) =>
+          conf.copy(command = AcceptDLCMutualClose(null, noBroadcast = false)))
         .text("Sign Mutual Close Tx for given oracle event")
         .children(
           opt[DLCMutualCloseSig]("closesig").required
@@ -272,11 +273,19 @@ object ConsoleCli {
                 case acceptClose: AcceptDLCMutualClose =>
                   acceptClose.copy(mutualCloseSig = closeSig)
                 case other => other
+              })),
+          opt[Unit]("noBroadcast").optional
+            .action((_, conf) =>
+              conf.copy(command = conf.command match {
+                case acceptClose: AcceptDLCMutualClose =>
+                  acceptClose.copy(noBroadcast = true)
+                case other => other
               }))
         ),
       cmd("getdlcfundingtx")
         .hidden()
-        .action((_, conf) => conf.copy(command = GetDLCFundingTx(null)))
+        .action((_, conf) =>
+          conf.copy(command = GetDLCFundingTx(null, noBroadcast = false)))
         .text("Returns the Funding Tx corresponding to the DLC with the given eventId")
         .children(
           opt[Sha256DigestBE]("eventid").required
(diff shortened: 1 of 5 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -760,36 +840,45 @@ object CliCommand {
       escaped: Boolean)
       extends CliCommand
 
-  case class AcceptDLCMutualClose(mutualCloseSig: DLCMutualCloseSig)
+  case class AcceptDLCMutualClose(
```
Christewart, line 880: can you just add a trait like 

```scala
trait Broadcastable { 
  def noBroadcast: Boolean
}
```

We might be able to get rid of all this ugly 

```scala
if (noBroadcast) { ... } else { ... } 
```

and refactor it into here too possibly?

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -38,6 +39,17 @@ case class WalletRoutes(wallet: UnlockedWalletApi, node: Node)(
...
+  private def handleBroadcastable(
+      tx: Transaction,
+      noBroadcast: Boolean): String = {
+    if (noBroadcast) {
+      tx.hex
+    } else {
+      node.broadcastTransaction(tx)
+      tx.txIdBE.hex
+    }
+  }
```
nkohen: have we decided to get rid of those sweep txs then? I think that's a good idea but we should make sure to do that in the `noBroadcast` case as well

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -38,6 +39,17 @@ case class WalletRoutes(wallet: UnlockedWalletApi, node: Node)(
...
   }
 
+  private def handleBroadcastable(
+      tx: Transaction,
+      noBroadcast: Boolean): String = {
+    if (noBroadcast) {
+      tx.hex
+    } else {
+      node.broadcastTransaction(tx)
+      tx.txIdBE.hex
```
nkohen: I think it is problematic that we don't propogate error messages from `node.broadcastTransaction` to the returned value from `handleBroadcastable`. It should return a `Future[String]` and return the txid inside of a map on the broadcast

nkohen: And then we need to make sure that the error gets propagated well to the cli (as opposed to just like a 5** error message), this can be tested by temporarily replacing the broadcast with a `Future.failed`

nkohen (commented):
Just a couple things to say, mostly about errors

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
Christewart: I agree with this, also why `String`? Shouldn't it at least be a hash? Types are our friends :-)

benthecarman: > I agree with this, also why String? Shouldn't it at least be a hash? Types are our friends :-)

It is either a hash or a `Transaction`

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
benthecarman: Those are still handled, `handleBroadcastable` is just called for both transactions

nkohen: I see, well in this case I actually think we should take this opportunity (or the next PR if you prefer) to get rid of the unnecessary sweep txs. We shouldn't be broadcasting frivolous txs by default :)

benthecarman: I think that is more fitting for a separate PR

nkohen:
> I made it so you need to `getdlcfundingtx` for the tx broadcast rather than having `adddlcsigs` do it, curious on thoughts about which option is desirable.

I think that instead of making `getdlcfundingtx` broadcast the funding tx, you should create a new call: `broadcastdlcfundingtx` as I think it is very confusing to have a `get` method have side effects like a broadcast, might as well split into two different calls

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
nkohen: could you open an issue?

nkohen (approved):
ACK

PR #1279, opened 2020-03-28
2020 03 28 uint32 cache

Christewart:
Check for common instances of uint32, and int32 (0,1,max etc) and use the cached object instead of creating a new one. This should reduce allocations so we can speed things up.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -332,7 +332,7 @@ object UInt32
     require(
       bytes.size <= 4,
       "UInt32 byte array was too large, got: " + BitcoinSUtil.encodeHex(bytes))
-    UInt32(bytes.toLong(false, ByteOrdering.BigEndian))
+    UInt32(bytes.toLong(signed = false, ordering = ByteOrdering.BigEndian))
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -315,13 +315,13 @@ object UInt32
             s"Cannot create ${super.getClass.getSimpleName} from $underlying")
   }
 
-  lazy val zero = UInt32(0)
-  lazy val one = UInt32(1)
+  lazy val zero = UInt32(BigInt(0))
```
nkohen, line 318: probably worth leaving a comment that this shouldn't be simplified to avoid infinite loop due to cache code

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -377,11 +391,12 @@ object Int32
             s"Cannot create ${super.getClass.getSimpleName} from $underlying")
   }
 
-  lazy val zero = Int32(0)
-  lazy val one = Int32(1)
+  val zero = Int32(BigInt(0))
```
nkohen, line 394: ditto

nkohen (commented):
Just some nits

benthecarman (approved):

PR #1280, opened 2020-03-28
Replace all usages of List inside of ScriptParser with Vector

Christewart:
This makes `ScriptParser` use a `Vector` for parsing rather than a `List`. The reason I am replacing this is because `List` does not have safe random access (`O(n)`). 

Scaladoc on List: 

```
  *  ==Performance==
  *  '''Time:''' `List` has `O(1)` prepend and head/tail access. Most other operations are `O(n)` on the number of elements in the list.
  *  This includes the index-based lookup of elements, `length`, `append` and `reverse`.
```

We should use Vector pretty much everywhere unless there is a _very_ good reason not to.


nkohen (approved):
ACK

benthecarman (approved):

Issue #1278, opened 2020-03-28
Out of memory error when syncing node on mainnet


Christewart:
When attempting to run the node on mainnet, I run out of memory. I was running with ~4GB of memory, so this shouldn't be happening 

Here is what `dmesg` is saying: 

![Screenshot from 2020-03-28 11-54-03](https://user-images.githubusercontent.com/3514957/77828772-80dfd680-70eb-11ea-8349-2b7cfd57fb9a.png)

Here is what things were looking like on visualvm before the process was killed by the OS. 

![Screenshot from 2020-03-28 11-52-47](https://user-images.githubusercontent.com/3514957/77828701-129b1400-70eb-11ea-8afb-0deb9adb959d.png)

As you can see in the upper right hand corner it looks like I have 2.5GB out of my max 4GB of heap space. 

From looking at threads, it looks like we have 2 threads running constantly, `bitcoin-s-akka.actor.default-dispatcher-7` and `chain.db-18`. I'm not sure what they are doing

![Screenshot from 2020-03-28 11-53-22](https://user-images.githubusercontent.com/3514957/77828801-ac62c100-70eb-11ea-98a8-86f543ed4d5f.png)

PR #1281, opened 2020-03-30
Reduce default num threads used for slick being a thread pool of 20 to 5

Christewart:
When looking at #1278 it appears that one problem we might be having is _to_ many threads in our database thread pools. While I doubt this is the _acutal_ problem, it seems seem logical to reduce the _large_ amount of threads that slick allocates by default to it's thread pool, which is 20!. 

See:

https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database

Says that the default number of threads allocated by slick is 20. That is way too much for our use case, even when i'm syncing filter headers it appears that 5 threads in the chain database is more than enough. 

Screenshot from 2020-03-29 20-25-39

### DIFF
--- a/db-commons/src/main/resources/db.conf
+++ b/db-commons/src/main/resources/db.conf
@@ -3,9 +3,12 @@ common = {
   profile = "slick.jdbc.SQLiteProfile$"
 
   db {
+    //for information on parameters available here see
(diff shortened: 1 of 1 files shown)


nkohen (approved):
ACK

benthecarman (approved):

Issue #1282, opened 2020-03-30
Cache utxos/addresses when calling `processCompactFilter()`


Christewart:
When downloading filters it appears that our bottleneck is currently `LockedWalletApi.processCompactFilter()`. The current implementation looks like this:

https://github.com/bitcoin-s/bitcoin-s/blob/0f899920f17bca492096eff3f3bba505c47fdb40/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L81

If we can find a way to cache `listUtxos()` and `listAddresses()` this will alleviate this bottleneck. 

Here is what visualvm looks like when syncing 

![Screenshot from 2020-03-29 20-32-28](https://user-images.githubusercontent.com/3514957/77867450-e079d700-71fc-11ea-9f96-5223c048241f.png)

PR #1283, opened 2020-03-30
Fetch addresses and utxos async in LockedWalletApi.processCompactFilt…

Christewart:
…er()

Picking some low hanging fruit on #1282 

Basically fetch utxos/addresses simultaneously, but this does not cache the across calls of `LockedWalletApi.processCompactFilter()`.


nkohen (approved):
ACK

benthecarman (approved):

PR #1285, opened 2020-03-30
Constructed simple Bitcoin-S wallet GUI

nkohen:
[doc](https://github.com/bitcoin-s/bitcoin-s/blob/afe04e7a9909bdba9e5114b489990a34c7d87b30/docs/applications/gui.md)


docs/applications/gui.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Bitcoin-S Graphical User Interface
+There is now a GUI built using [scalafx](https://www.scalafx.org/) to allow users who do not wish to use the [command line interface](cli.md) to interact with a Bitcoin-S wallet. This GUI is currently very minimal and looks something like this:
+
+![](gui-snapshot.png)
+
+At the time of writing this document, creating addresses and sending Bitcoin are the only supported wallet functionalities in the wallet's RPC API. The expansion of this interface will enable a more robust GUI and this issue is being tracked [here](https://github.com/bitcoin-s/bitcoin-s/issues/1284).
+
+### Running the GUI
+
```
benthecarman, line 15: It seems this only works with java versions 11+, maybe make a note of that

benthecarman: Also should make a note that you need `bitcoin-s-server` running

app/gui/src/main/scala/org/bitcoins/gui/dialog/SendDialog.scala
```diff
@@ -0,0 +1,58 @@
...
+      padding = Insets(20, 100, 10, 10)
+
+      var nextRow: Int = 0
+      def addRow(label: String, textField: TextField): Unit = {
+        add(new Label(label), 0, nextRow)
+        add(textField, 1, nextRow)
+        nextRow += 1
+      }
+
+      addRow("Receiving Address", addressTF)
```
benthecarman: I think "Receiving Address" is a little confusing, it can sound like we are receiving the funds. I think it would be better to just do "Address"

benthecarman (changes requested):

docs/applications/gui.md
nkohen: Def need the note about `bitcoin-s-server` but I'm running with `java 8` and it works just fine? Did you run into an issue?

benthecarman: Yeah I ran it with `java 8` and it said that the scalafx was compiled with a version of java that was 55+ and I was using a version that only supported a version that is 52+. After I installed java 11 it was fixed

nkohen: can you try with 8 again and do a clean and run again? I have seen that error before but only when trying to package the gui using graalvm. I have no issues running it personally

benthecarman: Looks like `sbt clean` fixed it :)

benthecarman (approved):

app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
```diff
@@ -0,0 +1,79 @@
...
+import scalafx.scene.control.Alert.AlertType
+import scalafx.scene.control.Alert
+import scalafx.stage.Window
+
+import scala.util.{Failure, Success}
+
+class WalletGUIModel() {
+  var taskRunner: TaskRunner = _
+
+  // Sadly, it is a Java "pattern" to pass null into
```
Christewart, line 18: :cry:

app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
```diff
@@ -0,0 +1,79 @@
...
+          }
+        )
+      case None => ()
+    }
+
+    updateBalance()
+  }
+
+  private def updateBalance(): Unit = {
+    ConsoleCli.exec(GetBalance(isSats = false)) match {
```
Christewart, line 67: Controversial!! No sats the standard? :P

build.sbt
```diff
@@ -242,6 +243,14 @@ lazy val cliTest = project
     testkit
   )
 
+lazy val gui = project
+  .in(file("app/gui"))
+  .settings(CommonSettings.prodSettings: _*)
+  .dependsOn(
+    cli,
```
Christewart: So why do we have both the `cli` and the `wallet` as a dep? Shouldn't we be using the wallet functionality via the `cli`? Is this because limitations to the current `WalletRoutes` where things we need are not exposed? If so please file an issue

app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
benthecarman: Petition for eventual config flag

nkohen: I figured I would just mimic the CLI for now (and this is its default value), in the future I imagine we can add a checkbox or dropdown or settings of some kind to allow this to be specified

Christewart (approved):

PR #1286, opened 2020-03-31
Network specific filterHeaderBatchSize

### LINKED ISSUE #1246: Neutrino node doesn't sync filters at all if we have less than 2,000 blocks (regtest)
bitcoind's chainstate: 
```
 ./bitcoin-cli -regtest getblockcount
105
```

bitcoin-s' chainstate: 

```
 ./bitcoin-s-cli getblockcount
105
```

and bitcoin-s' filtercount: 

```
./bitcoin-s-cli getfiltercount   

0
```

I believe @rorp has said that we cannot sync filters if we sync block header batches less than 2,000. 

Possibly related to #1059.

benthecarman:
Fixes #1246

Gives the ability to set a filterHeaderBatchSize for each network, this way for regtest you can set it to something much smaller and still keep the same config for other networks.

### DIFF
--- a/app/server/src/main/resources/reference.conf
+++ b/app/server/src/main/resources/reference.conf
@@ -48,6 +48,13 @@ bitcoin-s {
     chain {
         neutrino {
             filter-header-batch-size = 2000
+            filter-header-batch-size.regtest = 10
+            # You can set a network specific filter-header-batch-size
+            # by adding a trailing `.networkId` (main, test, regtest)
+            # It is recommended to keep the main and test batch size high
+            # to keep the sync time fast, however, for regtest it should be small
+            # so it does not exceed the chain size.
+
             filter-batch-size = 100
         }
     }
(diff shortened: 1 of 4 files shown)


Christewart (approved):

rorp:
But it doesn't fix the case where you have 9 blocks in the blockchain, right? Why not just set the desired batch size in the config file for each test as it's supposed to?

benthecarman:
> But it doesn't fix the case where you have 9 blocks in the blockchain, right?

Correct, but I think 10 is an okay minimum. 

>Why not just set the desired batch size in the config file for each test as it's supposed to?

I don't think any test cases are broken but what is broken is syncing a neutrino node on regtest (that has less then 2k blocks), this should fix that.

rorp:
> I don't think any test cases are broken but what is broken is syncing a neutrino node on regtest (that has less then 2k blocks), this should fix that.

Syncing in regtest is not broken, it works as expected. This behavior can be changed but changing `filter-header-batch-size` though.

rorp (approved):

Christewart:
Errors on CI and this should be documented in our example config on our website: https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/669412803#L662

PR #1287, opened 2020-03-31
GetAddress Wallet API call

### LINKED ISSUE #998: Add WalletApi 'getAddress' with HDPath that does _not_ return a new address every call
When operating a wallet with a large amount of active addresses it can be desirable to have the option to not generate a _new_ address _unless_ the previous address that has been served by the wallet has had funds sent to it. This does not leak privacy as the current address has not had money sent to it yet.

Cryptocurrency exchange's often have the problem where user's of the API generate a new address every time an API call is made. This is because the default wallet software they use only provides the ability to generate a new address, not `getaddress`.

benthecarman:
Closes #998

Adds the ability to get an address based off of it's path, not just the next generated one.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
@@ -6,7 +6,7 @@ import org.bitcoins.core.config.NetworkParameters
 import org.bitcoins.core.crypto.{DoubleSha256DigestBE, _}
 import org.bitcoins.core.currency.CurrencyUnit
 import org.bitcoins.core.gcs.{GolombFilter, SimpleFilterMatcher}
-import org.bitcoins.core.hd.{AddressType, HDAccount, HDPurpose}
+import org.bitcoins.core.hd.{AddressType, HDAccount, HDChainType, HDPurpose}
 import org.bitcoins.core.protocol.blockchain.{Block, BlockHeader, ChainParams}
 import org.bitcoins.core.protocol.script.ScriptPubKey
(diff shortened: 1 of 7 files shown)


Christewart:
I don't think this addresses all of 998? This still will keep serving "fresh" addresses even if funds have not been sent to the previous address?

nkohen (commented):
Looks good, needs tests

nkohen (approved):
ACK assuming CI

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -186,6 +186,24 @@ trait LockedWalletApi extends WalletApi with WalletLogger {
     } yield address
   }
 
+  def getUnusedAddress(addressType: AddressType): Future[BitcoinAddress]
```
Christewart, line 198: scaladocs on all of these explaining what they are meant for.

Aimeelu668:
楼主大大总结的很到位！不过自己开发的钱包都是热钱包机制，节点同步慢，归集非常麻烦，最重要的是这些私钥存在服务器上很危险，天天被黑客盯着。当时我们平台有50多万会员，为了解决钱包问题折腾了一年，最终放弃了，选择了优盾钱包（ https://www.uduncloud.com）。
通过API直接接入就行，有独立客户端操作审核、归集等，私钥也是自己在客户端上管理，服务器上不存。用了快一年，没有出现安全问题，财务都反馈说体验很好。我们也终于可以专注于业务开发了。推荐试试

PR #1297, opened 2020-04-03
GUI DLC tab

nkohen:
Replaces #1270

Adds a tab for DLCs to the wallet GUI


benthecarman:
image

This textbox should have a hint like "Number of outcomes"

image
might be nice to have generators for these as well

benthecarman:
Looks like there is some CI errors as well

benthecarman (approved):

PR #1299, opened 2020-04-04
2020 04 02 get new address queue

### LINKED ISSUE #1009: Wallet.getNewAddress() not async safe
This snippet of scala code will show that [`Wallet.getNewAddress()`](https://github.com/bitcoin-s/bitcoin-s/blob/1ad402ea56718052dac4ea260c4bf8f7fbff8c55/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala#L132) is not async safe. What I mean by this is if it is called multiple times in a row, we will get a foreign key error at the database level saying that we are attempting to violate `UNIQUE` constraint for the scriptPubKey/address relation

>2020-01-06T12:24:49 ERROR [DATABASE] Error when executing query SynchronousDatabaseAction.Fused
2020-01-06T12:24:49 ERROR [DATABASE] org.sqlite.SQLiteException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: addresses.script_pub_key)
Failed to build fixture with err=org.sqlite.SQLiteException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: addresses.script_pub_key)
2020-01-06T12:24:49 ERROR [DATABASE] Error when executing query SynchronousDatabaseAction.Fused
2020-01-06T12:24:49 ERROR [DATABASE] org.sqlite.SQLiteException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: addresses.script_pub_key)
[info] org.bitcoins.wallet.FundTransactionTest *** ABORTED ***
[info]   org.sqlite.SQLiteException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: addresses.script_pub_key)

```scala 
Vector.fill(3)(wallet.getNewAddress())
```

We need some sort of synchronization to make sure we don't serve the same address multiple times, this can lead to potentially accounting issues if this is a custodial wallet ([where multiple users are sharing the same account](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels)). 

I think this does relate to #998 over the long run, but this definitely needs to get fixed first before we can address that issue.

Christewart:
Closes #1009

Creates a background thread that process requests to `getNewAddress()` so that we can now call it multiple times and not have a race condition to the database. 

Two things still need to be addresses

- [x] A async safe buffer that stores the requests and be can be updated in async safe way.
- [x] There are new stack trace errors showing up in test logs, probably because the tale is being destroyed in tests before the `addressRequestQueue` is being drained.

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
@@ -79,7 +79,9 @@ class AddressHandlingTest extends BitcoinSWalletTest {
 
       for {
         addresses <- addressesF
-      } yield assert(addresses.distinct.length == addresses.length,
+      } yield {
+        assert(addresses.size == 10)
+        assert(addresses.distinct.length == addresses.length,
                      s"We receive an identical address!")
 
   }
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
@@ -6,6 +6,8 @@ import org.bitcoins.testkit.wallet.FundWalletUtil.FundedWallet
 import org.bitcoins.testkit.wallet.{BitcoinSWalletTest, WalletTestUtil}
 import org.scalatest.FutureOutcome
 
+import scala.concurrent.Future
+
 class AddressHandlingTest extends BitcoinSWalletTest {
   type FixtureParam = FundedWallet
 
@@ -68,4 +70,16 @@ class AddressHandlingTest extends BitcoinSWalletTest {
         assert(address2 != address3, "Must generate a new address")
       }
   }
+
+  it must "be safe to call getNewAddress multiple times in a row" in { wallet: Wallet =>
+    val addressesF = Future.sequence {
+      Vector.fill(3)(wallet.getNewAddress())
+    }
+
+    for {
+      addresses <- addressesF
+    } yield assert(addresses.distinct.length == addresses)
+
+  }
+
 }
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
@@ -73,12 +73,12 @@ class AddressHandlingTest extends BitcoinSWalletTest {
 
   it must "be safe to call getNewAddress multiple times in a row" in { wallet: Wallet =>
     val addressesF = Future.sequence {
-      Vector.fill(3)(wallet.getNewAddress())
+      Vector.fill(10)(wallet.getNewAddress())
     }
 
     for {
       addresses <- addressesF
-    } yield assert(addresses.distinct.length == addresses)
+    } yield assert(addresses.distinct.length == addresses.length,s"We receive an identical address!")
 
   }
 
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
@@ -83,7 +83,7 @@ class AddressHandlingTest extends BitcoinSWalletTest {
       } yield {
         assert(addresses.size == 10)
         assert(addresses.distinct.length == addresses.length,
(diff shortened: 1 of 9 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -82,7 +82,7 @@ trait FundTransactionHandling extends WalletLogger { self: LockedWalletApi =>
 
     val txBuilderF = for {
       addrInfosWithUtxo <- addrInfosWithUtxoF
-      change <- changeAddrF
+      change <- getNewChangeAddress(fromAccount)
```
Christewart, line 85: This is import that it is moved here, because if we fail earlier in coin selection the background thread that generates address will continue to attempt to generate an address. This means we get a bunch of exception stack traces in unit tests when we _want_ to fail funding a transaction, but we continue attempting to generate an address. 

When the address is inserted into the database, we have already torn down the wallet database which causes an exceptio saying that the address table was not found. Now with this code change, the change address isn't generated until we are certain the tx can be funded.

nkohen (approved):
ACK

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -206,4 +144,116 @@ private[wallet] trait AddressHandling extends WalletLogger {
...
+  walletThread.setName(
+    s"wallet-address-queue-${System.currentTimeMillis()}")
+  walletThread.start()
+  
+  /** A runnable that drains [[addressRequestQueue]]. Currently polls every 100ms
+    * seeing if things are in the queue. This is needed because otherwise
+    * wallet address generation is not async safe.
+    * @see https://github.com/bitcoin-s/bitcoin-s/issues/1009
+    * */
+  private case object AddressQueueRunnable extends Runnable {
```
benthecarman, line 339: On the `dlc` branch I have split up the former function `getNewAddressHelper` into 2 separate functions, one that gets the addressdb and one that writes. Would that break here / is it possible to do that?

Christewart: So I think you should maybe bring those changes upstream so I can incorporate them here: 

To answer your question, I believe the writing is the most important -- as long as the read is not used to create a new address. The errors that were caused before this PR was because there was a race condition when writing to the database.

benthecarman: https://github.com/bitcoin-s/bitcoin-s/pull/1322

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -93,67 +89,9 @@ private[wallet] trait AddressHandling extends WalletLogger {
...
-      val writeF = addressDAO.create(addressDb)
-      writeF.foreach { written =>
-        logger.debug(
-          s"Got ${chainType} address ${written.address} at key path ${written.path} with pubkey ${written.ecPublicKey}")
-      }
-
-      writeF.map(_.address)
-    }
+    val p = Promise[BitcoinAddress]
+    addressRequestQueue.add((account, chainType, p))
```
rorp, line 159: It needs a doc describing that `java.lang.IllegalStateException` can be thrown when the queue is full, and also a unit test that checks it.

wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
```diff
@@ -44,4 +46,17 @@ class AddressHandlingTest extends BitcoinSWalletTest {
...
 
+  it must "be safe to call getNewAddress multiple times in a row" in {
+    wallet: Wallet =>
+      val addressesF = Future.sequence {
+        Vector.fill(10)(wallet.getNewAddress())
+      }
+
+      for {
+        addresses <- addressesF
+      } yield assert(addresses.distinct.length == addresses.length,
```
rorp:
```suggestion
      } yield {
        assert(addresses.size == 10)
        assert(addresses.distinct.length == addresses.length,
```

rorp (changes requested):

PR #1300, opened 2020-04-04
P2WSH Signer fix + tests

benthecarman:
What the problem was when signing we would get the redeem script for p2wsh out of the wtx witness data, however, sometimes we would remove the witness data or not have the witness data when we needed it, so now we add the correct witness data from the spending info when we have it.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -196,6 +196,24 @@ case class PSBT(
 
   }
 
+  def addWitnessUTXOToInput(output: TransactionOutput, index: Int): PSBT = {
+    require(WitnessScriptPubKey.isWitnessScriptPubKey(output.scriptPubKey.asm),
+            s"Given output was not a Witness UTXO: $output")
+    require(
+      index < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $index >= ${inputMaps.size}")
+
+    val inputMap = inputMaps(index)
+    require(!inputMap.isFinalized,
+            s"Cannot update an InputPSBTMap that is finalized, index: $index")
+
+    val elements = inputMap.filterRecords(WitnessUTXOKeyId) :+ WitnessUTXO(
+      output)
+    val newInputMaps = inputMaps.updated(index, InputPSBTMap(elements))
+
+    PSBT(globalMap, newInputMaps, outputMaps)
+  }
+
   /**
     * Adds script to the indexed InputPSBTMap to either the RedeemScript
     * or WitnessScript field depending on the script and available information in the PSBT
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -60,21 +60,14 @@ sealed abstract class SignerUtils {
...
-                EmptyWitness.fromInputs(btx.inputs))
-
-            WitnessTransaction(btx.version,
-                               btx.inputs,
-                               btx.outputs,
-                               btx.lockTime,
-                               transactionWitness)
-          case wtx: WitnessTransaction => wtx
+        val wtx = {
+          val wtx = WitnessTransaction.toWitnessTx(unsignedTx)
```
nkohen: nit: rename `noWitnessWtx`

core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerIntegrationTest.scala
```diff
@@ -0,0 +1,189 @@
...
+    val programs: Vector[PreExecutionScriptProgram] =
+      tx.inputs.zipWithIndex.toVector.map {
+        case (input: TransactionInput, idx: Int) =>
+          val utxo = utxos.find(_.outPoint == input.previousOutput).get
+          createProgram(tx, idx, utxo)
+      }
+    ScriptInterpreter.runAllVerify(programs)
+  }
+
+  it must "sign inputs correctly" in {
```
nkohen: What is the difference between this and the test at the bottom of `BitcoinTxBuilderTest`?

nkohen (commented):
Just a nit and a question

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -196,6 +196,24 @@ case class PSBT(
 
   }
 
+  def addWitnessUTXOToInput(output: TransactionOutput, index: Int): PSBT = {
```
Christewart, line 205: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -494,6 +512,30 @@ case class PSBT(
     PSBT(globalMap, newInputMaps, outputMaps)
   }
 
+  def addSignatures(
```
Christewart, line 522: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -494,6 +512,30 @@ case class PSBT(
...
+    require(
+      inputIndex < inputMaps.size,
+      s"index must be less than the number of input maps present in the psbt, $inputIndex >= ${inputMaps.size}")
+    require(
+      !inputMaps(inputIndex).isFinalized,
+      s"Cannot update an InputPSBTMap that is finalized, index: $inputIndex")
+    val intersect = inputMaps(inputIndex).partialSignatures
+      .map(_.pubKey)
+      .intersect(partialSignatures.map(_.pubKey))
+    require(
```
Christewart, line 535: do we really want to fail in this case? It seems like maybe we should just ignore?

nkohen (approved):
ACK

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Changed to only throw error if the signatures are different

Issue #1185, opened 2020-03-01
Add the auto-incrementing id to slick representations of database items


benthecarman:
Should just be adding an `id: Int` to every `case class XDb` and then adding the corresponding column representation and pair functions

Refers to: https://github.com/bitcoin-s/bitcoin-s/pull/1164#discussion_r384823245

PR #1307, opened 2020-04-06
Address/UTXO Tagging

benthecarman:
First commit: The grunt of the work, this adds the ability to pass in an AddressTagDAO that will be used for tagging an address and subsequent utxos. Adds an argument to `sendToAddress` that is `newTags: Vector[AddressTag]` that will replace the current tags of the spent utxos to the new tags for the subsequent change utxo. Also adds an argument to `getNewAddress` that will tag the new address with the passed in `AddressTag`s.

Second commit: An example use case using this, creates a `StoreLocationTag` that specifies if a utxo is stored in a hot key, cold key, or deep cold key. To use it when calling `getNewAddress` they would now call it as `getNewAdress(..., tags =  Vector(HotStorage))` and `sendToAddress(..., newTags = Vector(ColdStorage))`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
@@ -1,0 +1,14 @@
+package org.bitcoins.core.wallet.utxo
+
+trait AddressTag {
+  val typeName: String
+}
+
+trait AddressTagFactory[TagType <: AddressTag] {
+
+  val all: Vector[TagType]
+
+  def fromString(str: String): Option[TagType] = {
+    all.find(tag => str.toLowerCase() == tag.toString.toLowerCase)
+  }
+}
--- a/core/src/main/scala/org/bitcoins/core/wallet/utxo/StorageLocationTag.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/utxo/StorageLocationTag.scala
@@ -1,0 +1,20 @@
+package org.bitcoins.core.wallet.utxo
+
+sealed abstract class StorageLocationTag extends AddressTag {
+  val typeName: String = "StorageLocationTag"
+}
+
+object StorageLocationTag extends AddressTagFactory[StorageLocationTag] {
+
+  /** Keys stored on a computer connected to the internet */
+  final case object HotStorage extends StorageLocationTag
+
+  /** Keys stored on a hardware wallet or other offline device */
+  final case object ColdStorage extends StorageLocationTag
+
+  /** Keys stored on a hardware wallet or other offline device locked in a safe in a distant location */
+  final case object DeepColdStorage extends StorageLocationTag
+
+  val all: Vector[StorageLocationTag] =
+    Vector(HotStorage, ColdStorage, DeepColdStorage)
+}
--- a/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
@@ -32,6 +32,8 @@ abstract class LockedWallet
   private[wallet] val outgoingTxDAO: OutgoingTransactionDAO =
     OutgoingTransactionDAO()
 
+  private[wallet] val addressTagDAOs: Vector[AddressTagDAO[_, _, _]]
+
   override def isEmpty(): Future[Boolean] =
     for {
       addressCount <- addressDAO.count()
@@ -167,14 +169,18 @@ abstract class LockedWallet
 object LockedWallet {
   private case class LockedWalletImpl(
       override val nodeApi: NodeApi,
-      override val chainQueryApi: ChainQueryApi)(
+      override val chainQueryApi: ChainQueryApi,
+      override val addressTagDAOs: Vector[AddressTagDAO[_, _, _]])(
       implicit val ec: ExecutionContext,
       val walletConfig: WalletAppConfig)
-      extends LockedWallet {}
+      extends LockedWallet
 
-  def apply(nodeApi: NodeApi, chainQueryApi: ChainQueryApi)(
+  def apply(
+      nodeApi: NodeApi,
+      chainQueryApi: ChainQueryApi,
+      addressTagDAO: Vector[AddressTagDAO[_, _, _]])(
       implicit ec: ExecutionContext,
       config: WalletAppConfig): LockedWallet =
-    LockedWalletImpl(nodeApi, chainQueryApi)
+    LockedWalletImpl(nodeApi, chainQueryApi, addressTagDAO)
 
 }
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -7,6 +7,7 @@ import org.bitcoins.core.hd._
 import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.wallet.fee.FeeUnit
+import org.bitcoins.core.wallet.utxo.AddressTag
 import org.bitcoins.keymanager.KeyManagerParams
 import org.bitcoins.keymanager.bip39.BIP39KeyManager
 import org.bitcoins.keymanager.util.HDUtil
@@ -27,14 +28,22 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
     */
   override def lock(): LockedWalletApi = {
     logger.debug(s"Locking wallet")
-    LockedWallet(nodeApi, chainQueryApi)
+    LockedWallet(nodeApi, chainQueryApi, addressTagDAOs)
   }
 
   override def sendToAddress(
       address: BitcoinAddress,
       amount: CurrencyUnit,
       feeRate: FeeUnit,
-      fromAccount: AccountDb): Future[Transaction] = {
+      fromAccount: AccountDb): Future[Transaction] =
+    sendToAddress(address, amount, feeRate, fromAccount, Vector.empty)
+
+  def sendToAddress(
+      address: BitcoinAddress,
+      amount: CurrencyUnit,
+      feeRate: FeeUnit,
+      fromAccount: AccountDb,
+      newTags: Vector[AddressTag]): Future[Transaction] = {
     logger.info(s"Sending $amount to $address at feerate $feeRate")
     val destination = TransactionOutput(amount, address.scriptPubKey)
     for {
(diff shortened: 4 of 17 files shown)


wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -32,6 +32,8 @@ abstract class LockedWallet
   private[wallet] val outgoingTxDAO: OutgoingTransactionDAO =
     OutgoingTransactionDAO()
 
+  private[wallet] val addressTagDAOs: Vector[AddressTagDAO[_, _, _]]
```
nkohen, line 35: GRRRRRRR, I don't like this

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -31,7 +32,7 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
       blockHashOpt: Option[DoubleSha256DigestBE]
   ): Future[LockedWallet] = {
     for {
-      result <- processTransactionImpl(transaction, blockHashOpt)
+      result <- processTransactionImpl(transaction, blockHashOpt, Vector.empty) //todo maybe change Vector.empty
```
nkohen, line 35: ping

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -400,6 +404,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
 
           for {
             _ <- insertIncomingTransaction(transaction, totalIncoming)
+            addTagFs = addressTagDAOs.map { dao =>
+              newTags.find(_.typeName == dao.typeName) match {
```
nkohen, line 408: I'm judging by this that only one tag for a given type can be added. I think it would be nice if this kind of thing had docs

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -400,6 +404,46 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+                      .map { out =>
+                        val address = BitcoinAddress
+                          .fromScriptPubKey(out.output.scriptPubKey,
+                                            networkParameters)
+                          .get
+                        dao.create(address,
+                                   tag.get
+                                     .asInstanceOf[AddressTagDb[_]]
+                                     .tag
+                                     .asInstanceOf[AddressTag])
```
nkohen, line 439: Looks like you might need some covariant markers

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDAO.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  override def findAll(
+      txs: Vector[DbEntryType]): Query[Table[_], DbEntryType, Seq] =
+    findByPrimaryKeys(txs.map(_.address))
+
+  def findByTag(tag: AddressTagType): Future[Vector[DbEntryType]] = {
+//    val q = table.filter(_.tag === tag) // fixme
+    val q = table
+
+    database.run(q.result).map(_.toVector).map(_.filter(_.tag == tag)) // fixme
```
nkohen, line 49: ping

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,47 @@
...
+  def tag: Rep[AddressTagType] = column[AddressTagType]("tag")(mapper)
+
+  type AddressTagTuple = (BitcoinAddress, AddressTagType)
+
+  val fromTuple: AddressTagTuple => AddressTagDbType
+
+  val toTuple: AddressTagDbType => Option[AddressTagTuple] =
+    AddressTag => Some((AddressTag.address, AddressTag.tag))
+
+  def * : ProvenShape[AddressTagDbType]
```
nkohen, line 34: I think this might be redundant as it is inherited

wallet/src/main/scala/org/bitcoins/wallet/models/StorageLocationTagTable.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.core.wallet.utxo.StorageLocationTag
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.ProvenShape
+
+case class StorageLocationTagDb(
+    address: BitcoinAddress,
+    tag: StorageLocationTag)
+    extends AddressTagDb[StorageLocationTag] {}
```
nkohen, line 11: nit: don't need {}

nkohen (commented):
I'll have to think more about structure/design but here's some initial review

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.core.wallet.utxo
+
+trait AddressTag {
+  val typeName: String
+}
+
+trait AddressTagFactory[TagType <: AddressTag] {
+
+  val all: Vector[TagType]
```
Christewart, line 9: Don't use `val` in `trait` if you can avoid it, it can cause weird initialization errors with Scala. Also you don't want to be purposeful about using `val` in traits -- do you want all inheritors not be able to define this as a method? Maye i want to read all tag types from a database?

https://stackoverflow.com/a/19642301

core/src/main/scala/org/bitcoins/core/wallet/utxo/StorageLocationTag.scala
```diff
@@ -0,0 +1,20 @@
+package org.bitcoins.core.wallet.utxo
+
+sealed abstract class StorageLocationTag extends AddressTag {
```
Christewart, line 3: So I think this is a problem -- if we want this to be extensible it seems that we actually _don't_ want this to be `sealed`. I.e. my business use case tomorrow is facilitating `LukewarmStorage` or whatever. How do I add this without coming in and modifying source code? 

If anything we maybe want to do this: 

```scala
sealed abstract class StorageLocationTag extends AddressTag {
  val typeName: String = "StorageLocationTag"
}

sealed abstract class DefaultStorageTags extends StorageLocationTag

/** Users can extend this to create their own storage tags */
abstract class UserDefinedStorageTags extends StorageLocationTag


object DefaultStorageTags extends AddressTagFactory[DefaultStorageTags] {

  /** Keys stored on a computer connected to the internet */
  final case object HotStorage extends DefaultStorageTags

  /** Keys stored on a hardware wallet or other offline device */
  final case object ColdStorage extends DefaultStorageTags

  /** Keys stored on a hardware wallet or other offline device locked in a safe in a distant location */
  final case object DeepColdStorage extends DefaultStorageTags

  val all: Vector[DefaultStorageTags] =
    Vector(HotStorage, ColdStorage, DeepColdStorage)
}
```

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -27,14 +28,22 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
...
-      fromAccount: AccountDb): Future[Transaction] = {
+      fromAccount: AccountDb): Future[Transaction] =
+    sendToAddress(address, amount, feeRate, fromAccount, Vector.empty)
+
+  def sendToAddress(
+      address: BitcoinAddress,
+      amount: CurrencyUnit,
+      feeRate: FeeUnit,
+      fromAccount: AccountDb,
+      newTags: Vector[AddressTag]): Future[Transaction] = {
```
Christewart, line 46: Why are `newTags` used here?

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDAO.scala
```diff
@@ -0,0 +1,71 @@
...
+abstract class AddressTagDAO[
+    AddressTagType <: AddressTag,
+    DbEntryType <: AddressTagDb[AddressTagType],
+    DbTable <: AddressTagTable[AddressTagType, DbEntryType]](
+    implicit val ec: ExecutionContext,
+    val appConfig: WalletAppConfig)
+    extends CRUD[DbEntryType, BitcoinAddress] {
+  import org.bitcoins.db.DbCommonsColumnMappers._
+
+  val typeName: String
```
Christewart, line 22: Why not just define the type here, i.e. 

```
def addressTag: AddressTagType

def typeName: String = addressTag.typeName
```

and then you can get rid of `create(address: BitcoinAddress, tag: AddressTag)`

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,47 @@
...
+abstract class AddressTagDb[AddressTagType <: AddressTag] {
+  def address: BitcoinAddress
+
+  def tag: AddressTagType
+}
+
+abstract class AddressTagTable[
+    AddressTagType <: AddressTag,
+    AddressTagDbType <: AddressTagDb[AddressTagType]](t: Tag, tableName: String)
+    extends Table[AddressTagDbType](t, tableName) {
```
Christewart, line 17: Any reason for not using `TableAutoInc`

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.core.wallet.utxo.AddressTag
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.{PrimaryKey, ProvenShape}
+
+abstract class AddressTagDb[AddressTagType <: AddressTag] {
+  def address: BitcoinAddress
+
+  def tag: AddressTagType
```
Christewart, line 11: Isn't this idea to have a `Vector[AddressTagType]`?

Christewart (commented):
Initial code review

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
benthecarman: This is for one db row, so there is only one. There is a Vector of AddressTagDAOs

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
benthecarman: No, we don't on any tables currently #1185

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: Adds an argument to sendToAddress that is newTags: Vector[AddressTag] that will replace the current tags of the spent utxos to the new tags for the subsequent change utxo

core/src/main/scala/org/bitcoins/core/wallet/utxo/StorageLocationTag.scala
benthecarman: `StorageLocationTag` is an example use case of extending AddressTag, this commit should not be merged. Check PR description.

PR #1308, opened 2020-04-06
Simplify `SingleSigner`

### LINKED ISSUE #1305: Simplify SingleSigner
Currently `SingleSigner` mimics `FullSigner` by going logically through a given `ScriptPubKey` until reaching a point where signing happens, it then diverges from `FullSigner`s by simply generating an `ECDigitalSignature` rather than generating a signed `TxSigComponent`.

It turns out, that this means it always ends up simply making a call to `TransactionSignatureCreator.createSig` which does not require any of the `ScriptPubKey` parsing as done by `FullSigner`s but rather can be accomplished directly with a calls to `SignerUtils` functions. This means that rather than having specific logic for every `ScriptPubKey` type on how to do `signSingle`, it is sufficient to have a single object such as `BItcoinSignSingle` with a function `signSingle` which directly computes a sig using `TransactionSignatureCreator`.

This will dramatically simplify and optimize the process of generating signatures, especially for PSBTs.

Furthermore, it will allow for us to extend the `SingleSigner` interface to create things like adaptor and Schnorr signatures, which is currently impossible to do without the proposed change.

nkohen:
Removed `SingleSigner` abstraction and replaced with a simple `signSingle` method in `SignerUtils`

Fixes #1305

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
@@ -30,6 +30,23 @@ sealed abstract class SignerUtils {
     }
   }
 
+  def signSingle(
+      spendingInfo: UTXOSpendingInfoSingle,
+      unsignedTx: Transaction,
+      isDummySignature: Boolean)(
+      implicit ec: ExecutionContext): Future[PartialSignature] = {
+    val signatureF = doSign(
+      sigComponent = sigComponent(spendingInfo, unsignedTx),
+      sign = spendingInfo.signer.signFunction,
+      hashType = spendingInfo.hashType,
+      isDummySignature = isDummySignature
+    )
+
+    signatureF.map { sig =>
+      PartialSignature(spendingInfo.signer.publicKey, sig)
+    }
+  }
+
   protected val flags: Seq[ScriptFlag] = Policy.standardFlags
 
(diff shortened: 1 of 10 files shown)


benthecarman (approved):

core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala
```diff
@@ -206,34 +186,64 @@ sealed abstract class FullSigner[-SpendingInfo <: UTXOSpendingInfoFull]
...
+        ConditionalSigner.sign(spendingInfo,
+                               unsignedTx,
+                               isDummySignature,
+                               conditional)
+      case p2wpkh: P2WPKHV0SpendingInfo =>
+        P2WPKHSigner.sign(spendingInfo, unsignedTx, isDummySignature, p2wpkh)
+      case pw2sh: P2WSHV0SpendingInfoFull =>
+        P2WSHSigner.sign(spendingInfo, unsignedTx, isDummySignature, pw2sh)
+      case _: UnassignedSegwitNativeUTXOSpendingInfo =>
+        throw new UnsupportedOperationException("Unsupported Segwit version")
```
Christewart, line 244: nit: Would be nice to add what version we got here to the error message.

Christewart (approved):

PR #1310, opened 2020-04-06
Update Secp256k1

### LINKED ISSUE #1306: Rebase onto libsecp256k1:master
Bitcoin-S currently uses a branch of libsecp256k1 from February of 2018 to execute cryptographic functions. This is out of date, and not only does that make it harder to maintain but also harder to add new functionality to (such as Schnorr and ECDSA Adaptor signatures).

We must rebase our secp256k1 library and rebuild binaries. This will also require fixing broken interfaces which include:

- `pubKeyTweakAdd` which no longer has a compression flag
- `pubKeyTweakMul` which no longer has a compression flag
- `decompress` which has been replaced with a pub key parse function

See this history for some useful commits: https://github.com/bitcoin-s/secp256k1/commits/all_schnorr

Specifically `407cd396a51c35e64fc862cb7db9fc22644fe961` which contains a squashed version of adding jni to libsecp

nkohen:
Rebased secp onto bitcoin-core's master (see https://github.com/bitcoin-s/secp256k1/pull/3) and built linux binaries

Fixes #1306


nkohen:
Note that in order for `git submodule update` (and `git clone --recursive`) to work, https://github.com/bitcoin-s/secp256k1/pull/3 must be merged

Christewart (commented):
It looks like you forgot to update `.gitmodules`.

nkohen:
I'm not sure what `.gitmodules` is about but pretty sure `.git/modules/secp256k1` has the important stuff in it as I've never had to touch `.gitmodules` to get submodule updates to register

docs/secp256k1/add-to-jni.md
```diff
@@ -0,0 +1,439 @@
...
+   unsigned char* prevArg = ...
+   unsigned char* nextArg = (unsigned char*) (prevArg + publen);
+   ```
+
+   where `publen` is a `jint` passed to this C function.
+
+   As an example that includes everything, consider the function `secp256k1_ecdsa_verify` which takes as input a `32` byte message, a variable length signature and a public key (of length `33` or `65` bytes). Our function will begin with
+
+   ```c
+   {
```
benthecarman, line 86: unnecessary `{`

docs/secp256k1/add-to-jni.md
```diff
@@ -0,0 +1,439 @@
...
+       int ret2 = secp256k1_ecdsa_signature_serialize_der(ctx, outputSer, &outputLen, &sig);
+       (void)ret2;
+     }
+   ```
+
+   As you can see, in this case we do not which to alter the value returned in `ret` if serialization fails. If we did then `ret2` would not be introduced and we would instead do `ret = [serialize]`.
+
+8. Populate return array when applicable
+
+   We now being translating our serialized results back into Java entities. If you are returning any `int`s containing meta-data (usually `ret` is included here, as are the variable lengths of outputs when applicable), you will want an `unsigned char intsarray[n]` to be already declared where `n`  is the number of pieces of meta-data. For example, in `secp256k1_ecdsa_sign`, we wish to return whether there were any errors (stored in `int ret`) and the output signature's length, `size_t outputLen`. Hence we have an `unsigned char intsarray[2]` and we populate it as follows
```
benthecarman: > We now being translating

We now begin translating?

Christewart (approved):
This LGTM assuming we get new mac binaries built.

docs/secp256k1/add-to-jni.md
nkohen: Actually I meant for it to be here as I'm trying to spell out the entire function and this seemed like the better place to include the brace than in the function signature

PR #1316, opened 2020-04-07
Dark mode GUI

### LINKED ISSUE #1314: Add dark mode to GUI
Dark mode all the things

benthecarman:
Closes #1314 

The dark-theme.css was taken from https://stackoverflow.com/a/58441758/6947285

### DIFF
--- a/app/gui/src/main/resources/themes/dark-theme.css
+++ b/app/gui/src/main/resources/themes/dark-theme.css
@@ -1,3 +1,4 @@
+/* Taken from https://stackoverflow.com/a/58441758/6947285 */
 .root {
     -fx-accent: #1e74c6;
     -fx-focus-color: -fx-accent;
--- a/app/gui/src/main/resources/themes/dark-theme.css
+++ b/app/gui/src/main/resources/themes/dark-theme.css
@@ -1,0 +1,34 @@
+.root {
+    -fx-accent: #1e74c6;
+    -fx-focus-color: -fx-accent;
+    -fx-base: #373e43;
+    -fx-control-inner-background: derive(-fx-base, 35%);
+    -fx-control-inner-background-alt: -fx-control-inner-background ;
+}
+
+.label{
+    -fx-text-fill: lightgray;
+}
+
+.text-field {
+    -fx-prompt-text-fill: gray;
+}
+
+.titulo{
+    -fx-font-weight: bold;
+    -fx-font-size: 18px;
+}
+
+.button{
+    -fx-focus-traversable: false;
+}
+
+.button:hover{
+    -fx-text-fill: white;
+}
+
+.separator *.line {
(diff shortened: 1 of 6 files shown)


app/gui/src/main/resources/themes/dark-theme.css
```diff
@@ -0,0 +1,34 @@
+.root {
```
nkohen, line 2: I think you should figure out how to comment in css and put a link to where you got this from in the file itself, just so we can find it later

app/gui/src/main/scala/org/bitcoins/gui/settings/SettingsPane.scala
```diff
@@ -0,0 +1,31 @@
...
+
+class SettingsPane {
+
+  private val themeCheckBox = new CheckBox {
+    text = "Dark Theme"
+    selected = GlobalData.defaultDarkTheme
+    onAction = new EventHandler[ActionEvent] {
+      override def handle(event: ActionEvent): Unit = {
+        if (!selected.value) {
+          stage.scene.value.getStylesheets.removeAll("/themes/dark-theme.css")
```
nkohen: I think this string deserves a val someplace

app/gui/src/main/scala/org/bitcoins/gui/WalletGUI.scala
```diff
@@ -88,7 +109,11 @@ object WalletGUI extends JFXApp {
     scene = new Scene(rootView)
   }
 
-  private val taskRunner = new TaskRunner(resultArea, glassPane, statusLabel)
+  if (GlobalData.defaultDarkTheme) {
+    stage.scene.value.getStylesheets.add("/themes/dark-theme.css")
+  }
```
nkohen: I think it would be nicer to define the `scene` as a private val above the stage and do
```
  private val scene = new Scene {
    root = rootView
    stylesheets = Seq("/themes/dark-theme.css")
  }
```

nkohen (commented):
Just a couple things

app/gui/src/main/scala/org/bitcoins/gui/TaskRunner.scala
```diff
@@ -11,7 +11,7 @@ import scalafx.scene.control.{Alert, Label}
   *
   * Copied from [[https://github.com/scalafx/ScalaFX-Tutorials/blob/master/slick-table/src/main/scala/org/scalafx/slick_table/TaskRunner.scala]]
   */
-class TaskRunner(mainView: Node, glassPane: Node, statusLabel: Label) {
+class TaskRunner(mainView: Node, glassPane: Node) {
```
nkohen, line 14: You can't make this diff without also binding all `statusLabel.text` values to `GlobalData.statusText`

PR #1317, opened 2020-04-07
Add wallet pay to many

benthecarman:


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -63,6 +63,42 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
     }
   }
 
+  override def sendToAddresses(
```
Christewart, line 66: nit: Maybe make `sendToOutputs()` that we just delegate to here? I think we should probably keep this, but it might be nice to just use outputs too.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -63,6 +63,42 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
...
+      addresses: Vector[BitcoinAddress],
+      amounts: Vector[CurrencyUnit],
+      feeRate: FeeUnit,
+      fromAccount: AccountDb): Future[Transaction] = {
```
Christewart: No `markAsReserved`?

Christewart (commented):
Looks like you are missing a unit test too

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: It is defaulted to false

Christewart: I guess what I mean is why can't it be passed into this API? Don't we want to mark the utxo as reserved possibly?

nkohen (commented):
Should probably add tests for all new endpoints, otherwise looks good

PR #1320, opened 2020-04-08
Address Tagging Attempt 2

benthecarman:
First commit: The grunt of the work, this adds an AddressTagDAO that will be used for tagging an address and subsequent utxos. Adds an argument to `sendToAddress` that is `newTags: Vector[AddressTag]` that will replace the current tags of the spent utxos to the new tags for the subsequent change utxo. Also adds an argument to `getNewAddress` that will tag the new address with the passed in AddressTags.


core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.core.wallet.utxo
+
+case class AddressTagType(name: String)
+
+case class AddressTag(tag: String, tagTypeOpt: Option[AddressTagType])
```
Christewart: Why is it `tagTypeOpt` rather than just `tagType: AddressTagType`

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
```diff
@@ -52,7 +52,8 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
                                         feeRate = SatoshisPerByte(Satoshis(3)),
                                         inputAmount = Satoshis(4000),
                                         sentAmount = Satoshis(3000),
-                                        blockHashOpt = None)
+                                        blockHashOpt = None,
+                                        Vector.empty)
```
Christewart: nit: parameter name

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -400,6 +410,20 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
 
           for {
             _ <- insertIncomingTransaction(transaction, totalIncoming)
+            prevTagDbs <- addressTagDAO.findTx(transaction.txIdBE,
+                                               networkParameters)
+            prevTags = prevTagDbs.map(tagDb =>
+              AddressTag(tagDb.tag.tag, Some(tagDb.tagType)))
```
Christewart: can we rename `AddressTag.tag` -> `AddressTag.name` ? So we don't have to do `tagDb.tag.tag` which is confusing

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDAO.scala
```diff
@@ -0,0 +1,89 @@
...
+  def findByTagType(tagType: AddressTagType): Future[Vector[AddressTagDb]] = {
+    val query = table.filter(_.tagType === tagType)
+
+    database.run(query.result).map(_.toVector)
+  }
+
+  def findTx(
+      txIdBE: DoubleSha256DigestBE,
+      network: NetworkParameters): Future[Vector[AddressTagDb]] = {
+    val infoQuery = spendingInfoTable.filter(_.txid === txIdBE)
```
Christewart: In general we should go through the `DAO` to query the table, so we don't have a layer violation.

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDAO.scala
```diff
@@ -0,0 +1,89 @@
...
+    val spendingInfosF = database.runVec(infoQuery.result)
+
+    spendingInfosF.flatMap { spendingInfos =>
+      if (spendingInfos.isEmpty) {
+        Future.successful(Vector.empty)
+      } else {
+        val firstSpk = spendingInfos.head.output.scriptPubKey
```
Christewart: Isn't this assuming we don't have multiple utxos on the same txid? Probably not safe

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -400,6 +410,20 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+            tagsToUse = prevTags
+              .filterNot(tag => newTags.contains(tag)) ++ newTags
+            newTagDbs = relevantOutsWithIdx.flatMap { out =>
+              val address = BitcoinAddress
+                .fromScriptPubKey(out.output.scriptPubKey, networkParameters)
+                .get
```
Christewart: Why `.get`? Shouldn't we just fail the future if we fail to parse the bitcoin address?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: Because we don't always have access to the `tagType`, for instance the db mapper.

nkohen: In that case I think it would be nice to have an intermediate state without the `AddressTagType` which has a `toAddressTag` method which takes in an `AddressTagType`

core/src/main/scala/org/bitcoins/core/wallet/utxo/StorageLocationTag.scala
```diff
@@ -0,0 +1,23 @@
+package org.bitcoins.core.wallet.utxo
+
+object StorageLocationTag {
+
+  /** Keys stored on a computer connected to the internet */
+  final val HotStorage =
+    AddressTag("HotStorage", Some(AddressTagType("StorageLocationTag")))
```
nkohen: nit: should probably give `"StorageLocationTag"` a `val` up top

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,14 @@
...
+
+trait AddressTagFactory[TagType <: AddressTagType] {
```
nkohen: Not sure this makes sense considering there is only one type of `AddressTagType` as the class cannot be extended.

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
nkohen: are either of these meant to be abstract classes?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.core.wallet.utxo
+
+case class AddressTagType(name: String)
```
nkohen: I don't think this type is much better than a `String` and should probably be an `ADT ` instead

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,48 @@
...
+  /** All tags must have an associated address */
+  def fk_address = {
+    val addressTable = TableQuery[AddressTable]
+    foreignKey("fk_address",
+               sourceColumns = address,
+               targetTableQuery = addressTable)(_.address)
+  }
+
+  def primaryKey: PrimaryKey =
+    primaryKey("pk_address", sourceColumns = address)
```
nkohen: Doesn't an address show up many times if it has many tags? How can it be a primary key?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: It should be safe to assume that every one can parse to a bitcoin address, it can only fail if it is not a `P2PKH`, `P2SH`, or `WitnessSPK`. Since our wallet can only receive funds from these we should never be spending from something else.

benthecarman: I don't think we ever plan on adding spending from `RawSPK`s as well

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: Sadly, no. That was the design in #1307, but that required that we have a separate table for every `AddressTagType` because the table would have no way to know how to map every type. The intention of this design is to have the `AddressTag` and `AddressTagType` just be wrappers for strings so that they can all exist in one table.

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: https://github.com/bitcoin-s/bitcoin-s/pull/1320#discussion_r406342504

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,24 @@
...
+
+sealed trait AddressTag {
+  def name: String
+  def tagType: String
+
+  def ==(at: AddressTag): Boolean = name == at.name && tagType == at.tagType
+  def !=(at: AddressTag): Boolean = !(this == at)
+}
+
+sealed trait AddressTagFactory[TagType <: AddressTagType] {
```
Christewart: It doesn't look like this is used?

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,82 @@
...
+Also, when sending you can use `fundRawTransaction` and use `fromTagOpt` to pass in an optional `AddressTag`,
+this will use only utxos associated with the `AddressTag`.
+
+### Creating your own AddressTags
+
+You can create your own custom `AddressTag`s. This allows you to tag addresses and utxos in any way that your
+application needs. To do this you are going to need to use the `CustomAddressTag`.  As an example we will create
+`AddressTag`s for storage location: `HotStorage`, `ColdStorage`, and `DeepColdStorage`.
+
+First, we will need to define the tag type. We define then define the tag name for each tag.
```
Christewart: typo

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,82 @@
...
+You will need to define a way to go to and from a `CustomAddressTag`.
+```scala mdoc:compile-only
```
Christewart: i don't think this needs to be `compile-only` as there isn't any network calls here is there?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
Christewart: Shouldn't `StorageLocationTags` extend this?

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  def fromCustom(custom: CustomAddressTag): Option[StorageLocationTag] = {
+    require(custom.tagType.isEmpty || custom.tagType == tagType)
+    fromString(custom.name)
+  }
+}
+```
+
+When using a `StorageLocationTag` now you can use them however, you please, however when passing them into
```
Christewart: Shouldn't we add an example of actually using it to receive and spend funds?

core/src/main/scala/org/bitcoins/core/wallet/utxo/example.scala
```diff
@@ -0,0 +1,40 @@
+package org.bitcoins.core.wallet.utxo
+
+trait StorageLocationTag {
```
Christewart: why don't we have a common super class? i.e. `extends AddressTag`

wallet-test/src/test/scala/org/bitcoins/wallet/AddressTagIntegrationTest.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  behavior of "Address Tag - integration test"
+
+  val feeRate: SatoshisPerByte = SatoshisPerByte(Satoshis.one)
+
+  val exampleTag: CustomAddressTag =
+    CustomAddressTag("ExampleTag", "ExampleType")
+
+  /** Checks that the given values are the same-ish, save for fee-level deviations */
+  private def isCloseEnough(
```
Christewart: isn't this defined elsewhere?

wallet-test/src/test/scala/org/bitcoins/wallet/AddressTagIntegrationTest.scala
```diff
@@ -0,0 +1,123 @@
...
+      first: CurrencyUnit,
+      second: CurrencyUnit,
+      delta: CurrencyUnit = 1000.sats): Boolean = {
```
Christewart: 1000 sats delta? :$

wallet-test/src/test/scala/org/bitcoins/wallet/models/AddressTagDAOTest.scala
```diff
@@ -0,0 +1,56 @@
...
+  behavior of "AddressTagDAO"
+
+  val exampleTag: CustomAddressTag =
+    CustomAddressTag("ExampleTag", "ExampleType")
+
+  it should "fail to insert and read an address tag into the database without a corresponding address" in {
+    daos =>
+      val tagDAO = daos.addressTagDAO
+      val addr =
+        BitcoinAddress.fromString("n4MN27Lk7Yh3pwfjCiAbRXtRVjs4Uk67fG").get
```
Christewart: Is there a reason you aren't using `TestUtil.testBitcoinAddress`?

wallet/src/main/scala/org/bitcoins/wallet/LockedWallet.scala
```diff
@@ -89,13 +91,25 @@ abstract class LockedWallet
...
-        utxo.blockHash.isDefined
+        utxo.state == ConfirmedReceived
       }
     }
 
     unspentInAccountF.map(_.foldLeft(CurrencyUnits.zero)(_ + _.output.value))
   }
 
+  override def getConfirmedBalance(tag: AddressTag): Future[CurrencyUnit] =
+    getConfirmedBalance(walletConfig.defaultAccount, tag)
```
Christewart: Hmm.... this is a good question. Do tags only apply to a specific account? I would expect them to be independent of accounts.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: I think it is a cleaner way to propogate the future by wrapping it inside of `Future.failed()` and returning that rather than just possibly throwing the exception. I believe there is a `Future.fromTry()` helper method exactly for this.

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.bitcoins.core.wallet.utxo.{AddressTag, CustomAddressTag}
+import slick.jdbc.SQLiteProfile.api._
+import slick.lifted.ProvenShape
+
+case class AddressTagDb(
+    address: BitcoinAddress,
+    tag: AddressTag,
+    tagType: String) {
+  require(
+    tag.tagType.isEmpty || tag.tagType == tagType,
```
Christewart: So I don't really understand why we have `tagType` as a parameter. Why can't we just call `tag.tagType` to get the type and remove the `tagType` parameter all together?

Christewart (commented):
I think my biggest design concern is this:

I think we want to have tags be independent of the `HDAccount` system in the wallet. Large providers may not use `HDAccounts`. We probably want to de-couple our wallet design from them and definitely not add more functionality that is behind the scenes searching based on `hdaccount+address_tag`. Thoughts?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: So, this will be used for Bitcoin-S supported address tags that we add in the future. The `StorageLocationTags` is an example for someone importing bitcoin-s and creating their own address tags.

core/src/main/scala/org/bitcoins/core/wallet/utxo/example.scala
benthecarman: This was meant to be deleted

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
benthecarman: Because sometimes (like in the db mappers) we don't have access to the tag type when we are getting the tag, so we need to be able to pass it in.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: It will still propagate to a failed future because it is within the `for`

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    fromString(custom.name)
+  }
+}
+
+sealed trait InternalAddressTagType extends AddressTagType
+sealed trait InternalAddressTag extends AddressTag
+
+trait ExternalAddressTagType extends AddressTagType
+trait ExternalAddressTag extends AddressTag
```
nkohen: This won't work, `sealed` is inherited (otherwise it wouldn't work), so you're going to have to make the top type not `sealed` (and I would suggest it go in another file). And only `InternalAddressTag[Type]` will be `sealed`

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,74 @@
...
+
+case class CustomAddressTag(name: String, tagType: String)
+    extends ExternalAddressTag
```
nkohen: I'm not sure I understand the point of having `CustomAddressTag` extend `ExternalAddressTag`? Isn't the point of having internal and external that this is no longer needed? Or perhaps you still want an internal representation for external tags in which case this should extend `InternalAddressTag` and should probably be renamed to `UnknownAddressTag` or something like this

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
nkohen: These need scaladocs, make sure to specifically make clear that `Internal` is available externally

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,74 @@
...
+
+sealed trait AddressTagType {
+  def tagType: String
+}
+
+sealed trait AddressTag {
+  def name: String
+  def tagType: String
+
+  def ==(at: AddressTag): Boolean = name == at.name && tagType == at.tagType
```
nkohen: nit: Should probably be `name == at.name && tagType.name == at.tagType.name`, that is unless you want to override equality checks in `AddressTagType` as well

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -172,6 +172,11 @@ abstract class DbCommonsColumnMappers {
     MappedColumnType
       .base[SatoshisPerByte, Long](_.toLong, SatoshisPerByte.fromLong)
   }
+
+  implicit val addressTagMapper: BaseColumnType[AddressTag] = {
+    MappedColumnType
+      .base[AddressTag, String](_.name, tag => CustomAddressTag(tag, ""))
```
nkohen: should probably give this constructor a special name rather than using `apply` as we generally don't want people doing this by accident

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+```scala mdoc:invisible
+import org.bitcoins.core.wallet.utxo._
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.core.protocol.transaction._
+import org.bitcoins.core.wallet.fee._
+import org.bitcoins.core.currency._
+import org.bitcoins.wallet.models.AccountDb
+import org.bitcoins.wallet._
+
+val ExampleAddressTag = CustomAddressTag("name", "tagType")
```
nkohen: we should probably give an example which extends `ExternalAddressTag` directly so that users have a good model no?

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+          newTags: Vector[AddressTag]) = true
+}
+```
+
+### Using AddressTags
+
+The Bitcoin-S wallet allows you to give addresses, and their associated utxos,
+a tag. These tags allow you to separate funds between utxos so you can query utxos,
+and spend from them, based off of an AddressTag. The system also allows you to create
+your own custom address tags, that will be enforced by the library.
```
nkohen, line 43: I'm a little confused, is it true that a given utxo can only have one tag?

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+
+An address tag consists of the tag name, and a tag type. We use a tag type so we can have
+tag with the same name without complications.
+
+To create an address with a tag you can use `getNewAddress` but pass in a `Vector` of `AddressTag`.
```
nkohen: nit I think `Vector[AddressTag]` is cleaner

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+It will add to the address tag database along with all the corresponding tags.
+
+```scala mdoc:silent
+wallet.getNewAddress(tags = Vector(ExampleAddressTag)) 
+```
+
+When sending with `sendToAddress` you can also a `Vector` of new `AddressTag`s that will be applied to the
+resulting change outputs. Any tags of a different tag type not included in `newTag`s will also be applied to
+the change outputs.
```
nkohen, line 57: doesn't have to be on this PR but I think it would be nice to have an RPC endpoint for adding a tag to an existing address

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+
+We will need to define the tag type, then define the tag name for each tag, as well as a way to go to and
+from a `String`. Then we define the actual tags, we are going to have a `Company`, `InsuranceFund`, and `UserId`
+tags. We are going to make the `UserId` tag special, and allow it to take in any user id so we can have a huge
+set of users but all with different ids.
+
+```scala mdoc:silent
+/** Allows to assign funds in a specific address to a user */
+trait UserIdTag extends ExternalAddressTagType {
+  val tagType: String = UserIdTags.tagType
```
nkohen: nit: `override` keyword

nkohen: same below

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+}
+
+object UserIdTags extends AddressTagFactory[UserIdTag] {
+
+  val tagType: String = "UserIdTag"
+
+  /** Funds that do not belong to any user and instead belong to the company */
+  case object Company extends ExternalAddressTag with UserIdTag {
+    val name: String = "Company"Home
```
nkohen: typo? (the ")

docs/wallet/address-tagging.md
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /** Funds in the company's insurance fund */
+  case object InsuranceFund extends ExternalAddressTag with UserIdTag {
```
nkohen, line 113: `UserIdTag` already extends `ExternalAddressTag` so this isn't needed here or on the others

testkit/src/main/scala/org/bitcoins/testkit/wallet/WalletTestUtil.scala
```diff
@@ -77,6 +77,21 @@ object WalletTestUtil {
...
+      second: CurrencyUnit,
+      delta: CurrencyUnit = 300.sats): Boolean = {
+    val diff =
+      if (first > second) {
+        first - second
+      } else if (first < second) {
+        second - first
+      } else 0.sats
+
+    diff < delta
```
nkohen: nit: `Math.abs(first - second) < delta` would be more concise

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: `tagType` is just a string

docs/wallet/address-tagging.md
benthecarman: This is invisible, just needed for the wallet api examples

docs/wallet/address-tagging.md
benthecarman: `UserIdTag` extends `ExternalAddressTagType` not `ExternalAddressTag`

docs/wallet/address-tagging.md
benthecarman: Agreed

docs/wallet/address-tagging.md
benthecarman: No, utxos can have any number of tags, just sending is only from a single tag currently to keep it simpler.

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: >  perhaps you still want an internal representation for external tags

Yes, this is the main need of this. I like `UnknownAddressTag` better so I will do that.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,43 @@
...
+sealed trait InternalAddressTagType extends AddressTagType
+
+/** An AddressTag that is native to Bitcoin-S.
+  * InternalAddressTags are still usable when using Bitcoin-S
+  * as a dependency
+  */
+sealed trait InternalAddressTag extends AddressTag
+
+/** Storage Location of the private keys associated with the address */
+sealed trait StorageLocationTag extends InternalAddressTagType {
```
Christewart: Shouldn't this be called `StorageLocationTagType`?

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -172,6 +172,11 @@ abstract class DbCommonsColumnMappers {
     MappedColumnType
       .base[SatoshisPerByte, Long](_.toLong, SatoshisPerByte.fromLong)
   }
+
+  implicit val addressTagMapper: BaseColumnType[AddressTag] = {
+    MappedColumnType
+      .base[AddressTag, String](_.name, tag => UnknownAddressTag(tag, ""))
```
Christewart: So everything from the database is of type `UnknownAddressTag`? this is not ideal :$

Christewart: At the end of the day, I think we can only provide a `addressTagMapper` of type `BaseColumnType[InternalAddressTag]`. We are going to have to require users using the external tag type to provide their own mapper that implements 

```scala
implicit val externalTagMapper: BaseColumnType[MyExternalTag] = ...
```

Christewart (changes requested):
It seems there are zero unit tests that use our `InternalAddressTag`. 

I think we should add some unit tests -- particularly in `AddressTagIntegrationTest` and the `TransactionHandlingTest`s that test our internal tags. It seems like you have just used `UnknownAddressTag` everywhere so far.

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
benthecarman: If we do that, we then need to go back to the DAO for every type. We need to have a single mapper that works for every kind of address tag.

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
nkohen: Ah I see, shouldn't it be an `AddressTagType`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/** An address tag without an unknown type, should mostly be used for parsing external address tags */
+case class UnknownAddressTag(name: String, tagType: String) extends AddressTag
+
+trait AddressTagFactory[TagType <: AddressTagType] {
+
+  def tagType: String
+
+  def all: Vector[TagType]
```
nkohen: shouldn't this be a vector of `AddressTag`s?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  def toUnknown: UnknownAddressTag = UnknownAddressTag(name, tagType)
+}
+
+/** An address tag without an unknown type, should mostly be used for parsing external address tags */
+case class UnknownAddressTag(name: String, tagType: String) extends AddressTag
+
+trait AddressTagFactory[TagType <: AddressTagType] {
+
+  def tagType: String
```
nkohen: Shouldn't this have type `TagType`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  def ==(at: AddressTag): Boolean = name == at.name && tagType == at.tagType
+  def !=(at: AddressTag): Boolean = !(this == at)
+
+  def toUnknown: UnknownAddressTag = UnknownAddressTag(name, tagType)
+}
+
+/** An address tag without an unknown type, should mostly be used for parsing external address tags */
+case class UnknownAddressTag(name: String, tagType: String) extends AddressTag
```
nkohen: shouldn't this be an `InternalAddressTag`?

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/** Storage Location of the private keys associated with the address */
+sealed trait StorageLocationTagType extends InternalAddressTagType {
+  override val tagType: String = StorageLocationTag.tagType
+}
+
+object StorageLocationTag extends AddressTagFactory[StorageLocationTagType] {
+
+  override val tagType: String = "StorageLocationTag"
+
```
nkohen: this seems backwards to me. Should be
```
sealed trait StorageLocationTag extends InternalAddressTag {
    override val tagType: AddressTagType = StorageLocationTagType
}

object StorageLocationTagType extends InternalAddressTagType {
    override val tagType: String = "StorageLocationTag"
}

object StorageLocationTag extends AddressTagFactory[StorageLocationTag] {
    override val tagType = StorageLocationTagType

    case object HotStorage extends StorageLocationTag {
        override val name: String = "HotStorage"
    }
...
}

File: db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala (hunk 34 above)
Comment (nkohen) [reply db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala]: why wouldn't we do `InternalAddressTag.fromString(tag).getOrElse(UnknownAddressTag(tag)`?

**Review by nkohen — CHANGES_REQUESTED** [review]

File: db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala (hunk 34 above)
Comment (benthecarman) [reply db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala]: We can't do a `InternalAddressTag.fromString(tag)` without knowing the tag type.

File: core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala [hunk 39]
```diff
@@ -0,0 +1,44 @@
...
+  * An tag for an address. It's name is what it is referred to as
+  * and it's tagType is its parent AddressTagType
+  */
+trait AddressTag {
+  def name: String
+  def tagType: AddressTagType
+
+  def ==(at: AddressTag): Boolean =
+    name == at.name && tagType.name == at.tagType.name
+  def !=(at: AddressTag): Boolean = !(this == at)
```
nkohen: Do we need to do this for `AddressTagType` too? If so then we won't need the call to `name` here btw

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def toUnknown: UnknownAddressTag = UnknownAddressTag(name, tagType)
```
nkohen: shouldn't this belong in `ExternalAddressTag` rather than in the top-level type? I can't imagine when it would be right to convert a known internal type to unknown

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,44 @@
...
+}
+
+trait AddressTagFactory[TagType <: AddressTag] {
```
nkohen: nit: I prefer `Tag` to `TagType` now that this extends `AddressTag`

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,44 @@
...
+  def fromUnknown(unknown: UnknownAddressTag): Option[AddressTag] = {
+    // Allow empty tagType because sometimes (like in the db mappers) we don't have access to
+    // the tagType when we are getting the tag, so we need to be able to pass it in.
+    require(
+      unknown.tagType.name.isEmpty || unknown.tagType.name == tagType.name,
+      s"If tag's tagType is defined it must be the same as tagType, got ${unknown.tagType}"
+    )
+
+    fromString(unknown.name)
+  }
```
nkohen: I don't think that this makes sense for internal Tag Factories either

core/src/main/scala/org/bitcoins/core/wallet/utxo/ExternalAddressTag.scala
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+  * Address Tag Types defined outside the library, used for other projects
+  * creating there own address tags that aren't supported by bitcoin-s
+  */
+trait ExternalAddressTagType extends AddressTagType
+
+/**
+  * Address Tags defined outside the library, used for other projects
+  * creating there own address tags that aren't supported by bitcoin-s
```
nkohen, line 17: nit: "aren't natively supported" as we do support them using this trait :) (same above)

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,77 @@
...
+/**
+  * An AddressTag that is native to Bitcoin-S.
+  * InternalAddressTags are still usable when using Bitcoin-S
+  * as a dependency
+  */
+sealed trait InternalAddressTag extends AddressTag
+
+/** An unknown address tag type, should mostly be used for parsing external address tags */
+case class UnknownAddressTagType(name: String) extends InternalAddressTagType
```
nkohen: just a thought, not sure if this is something we want, but we could add a `require` to this that makes sure `name` is not known internally? Same thing for `UnknownAddressTag` (where we would check both not just name)

nkohen: This would help us avoid multiple representations

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,77 @@
...
+
+/** An address tag without an unknown type, should mostly be used for parsing external address tags */
+case class UnknownAddressTag(
+    name: String,
+    tagType: AddressTagType = UnknownAddressTagType(""))
```
nkohen: I'm not a fan of this default type, I think users should be forced to be explicit when omitting a tagType

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,77 @@
...
+    InternalAddressTagType.fromString(unknown.tagType.name) match {
+      case Some(tagType) =>
+        tagType match {
+          case StorageLocationTagType =>
+            StorageLocationTag.fromString(unknown.name)
+        }
+      case None => None
+    }
+  }
+}
```
nkohen: When would this ever be used?

nkohen: You should have a `fromString` instead, and not use `Unknown` as an intermediate state

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -172,6 +178,17 @@ abstract class DbCommonsColumnMappers {
...
+  implicit val addressTagMapper: BaseColumnType[AddressTag] = {
+    MappedColumnType
+      .base[AddressTag, String](_.name, tag => UnknownAddressTag(tag))
+  }
+
+  implicit val addressTagTypeMapper: BaseColumnType[AddressTagType] = {
+    MappedColumnType
+      .base[AddressTagType, String](_.name,
+                                    tagType => UnknownAddressTagType(tagType))
+  }
```
nkohen: see comment above, I'd prefer if we had `fromString` methods that only returned `Unknown` if the name is not matched internally

docs/wallet/address-tagging.md
nkohen: I think it would be nicer to have `object UserIdTagType extends ExternalAddressTagType` and then `sealed trait UserIdTag extends ExternalAddressTag` and have the tags extend that

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
```diff
@@ -0,0 +1,66 @@
...
+case class AddressTagDb(
+    address: BitcoinAddress,
+    tag: AddressTag,
+    tagType: AddressTagType) {
+  require(
+    tag.tagType.name.isEmpty || tag.tagType == tagType,
+    s"If tag's tagType is defined it must be the same as tagType, got $tagType")
+
+  val unknownAddressTag: UnknownAddressTag =
+    UnknownAddressTag(tag.name, tagType)
```
nkohen: I'd prefer `internalAddressTag` which only returns `Unknown` if it is not an internal tag

nkohen (commented):
Nice! this is basically exactly what I wanted :)

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: It could but I don't see the harm of it existing here

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
benthecarman: I use `fromUnknown` because we need both the tag and the tag type to properly get an address tag, so it'd need to be `from(str1: String, str2: String)`.  I think using `UnknownAddressTag` is just a better way of doing this

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagTable.scala
benthecarman: that is essentially what `val addressTag` is. With context of replies to above comments, I think it makes sense to do it this way

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: The error messsage isn't retained when calling `.get`, compared to using `Future.fromTry()`. When debugging you will get a `NoSuchElementException` rather than a more descriptive error message.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
```diff
@@ -0,0 +1,144 @@
...
+
+/** An unknown address tag type, should mostly be used for parsing external address tags */
+case class UnknownAddressTagType(typeName: String)
+    extends InternalAddressTagType {
+  require(InternalAddressTagType.fromStringOpt(typeName).isEmpty,
+          s"This tag type is already defined, got $typeName")
+}
+
+/** An address tag without an unknown type, should mostly be used for parsing external address tags */
+case class UnknownAddressTag(tagName: AddressTagName, tagType: AddressTagType)
```
Christewart, line 38: Why do we have an `UnknownAddressTag extend InternalAddressTag`? If it's an internal tag, we should be able to define it. Only external things should have the option for `UnkonwnAddressTag` no?

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDb.scala
```diff
@@ -0,0 +1,28 @@
...
+}
+
+case class AddressTagDb(
+    address: BitcoinAddress,
+    tagName: AddressTagName,
+    tagType: AddressTagType) {
+
+  val addressTag: AddressTag = InternalAddressTag(tagName, tagType)
+
+  def ==(at: AddressTagDb): Boolean =
```
Christewart, line 18: Why don't we test for equality with `tagType`?

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -172,6 +172,17 @@ abstract class DbCommonsColumnMappers {
     MappedColumnType
       .base[SatoshisPerByte, Long](_.toLong, SatoshisPerByte.fromLong)
   }
+
+  implicit val addressTagMapper: BaseColumnType[AddressTagName] = {
```
Christewart, line 216: So anything that comes out of the database will be a `InternalAddress` type? Am I circling back to the `InternalDAO` vs `ExternalDAO` thing?

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
```diff
@@ -0,0 +1,45 @@
...
+  */
+trait AddressTag {
+  def tagName: AddressTagName
+  def tagType: AddressTagType
+
+  def ==(at: AddressTag): Boolean =
+    tagName == at.tagName && tagType == at.tagType
+  def !=(at: AddressTag): Boolean = !(this == at)
+
+  def toUnknown: UnknownAddressTag = UnknownAddressTag(tagName, tagType)
```
nkohen: I still don't think this should be available to internal tags.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
nkohen: Probably needs a better doc, but currently `UnknownAddressTag` is the internal representation of external tags.

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
nkohen: Not sure what the value of having multiple DAOs is, could you elaborate?

wallet/src/main/scala/org/bitcoins/wallet/models/AddressTagDb.scala
benthecarman: We do, we test equality of `addressTag` which uses both `tagName` and `tagType`

core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTag.scala
benthecarman: Meant to remove this

core/src/main/scala/org/bitcoins/core/wallet/utxo/InternalAddressTag.scala
Christewart: If that is the case, I have two suggestions: 

1. Rename it to `ExternalAddressTagWrapper` or something like that
2. Redefine it to take an `ExternalAddressTag` as a parameter i.e. 

```scala
case class ExternalAddressTagWrapper(external: ExternalAddressTag)
    extends InternalAddressTag {
  override val tagName: AddressTagName = external.tagName
  override val tagType: AddressTagType = external.tagType
}
```

PR #1321, opened 2020-04-08
Small DLC Improvements

benthecarman:
Small improvements to the DLC wallet


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -927,6 +927,14 @@ sealed abstract class CliCommand
 object CliCommand {
   case object NoCommand extends CliCommand
 
+  trait JsonResponse {
+    def escaped: Boolean
+  }
+
+  trait Broadcastable {
+    def noBroadcast: Boolean
+  }
```
nkohen, line 936: We should move these to `appCommons` eventually

nkohen (approved):
ACK

PR #1322, opened 2020-04-08
getNewAddressHelper Refactor

benthecarman:
These changes are incorporated in the `dlc` branch. #1299 incited adding this to the upstream branch.

This change is useful because it allows to call `getNextAvailableIndex` without needing to write that address to the database.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
@@ -89,10 +89,10 @@ private[wallet] trait AddressHandling extends WalletLogger {
     * @param account Account to generate address from
     * @param chainType What chain do we generate from? Internal change vs. external
     */
-  private def getNewAddressHelper(
+  private def getNewAddressDb(
       account: AccountDb,
       chainType: HDChainType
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -117,45 +117,56 @@ private[wallet] trait AddressHandling extends WalletLogger {
...
+      _ = logger.debug(s"Writing $addressDb to DB")
+      written <- addressDAO.create(addressDb)
+    } yield {
+      logger.debug(
+        s"Got ${chainType} address ${written.address} at key path ${written.path} with pubkey ${written.ecPublicKey}")
+      written.address
     }
   }
 
+  def getNextAvailableIndex(
```
Christewart, line 164: Was this meant to be used somewhere?

Christewart (approved):

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
benthecarman: Not yet, it is used in PTLCs and DLCs

Issue #1323, opened 2020-04-08
Filter sync optimization: CompactFilterHeaderDAO.maxHeightQuery()


Christewart:
When syncing my compact filters, the bottle neck was [`ComapctFilterDAO.maxHeightQuery`](https://github.com/bitcoin-s/bitcoin-s/blob/81b8ac7027e5e38480ac7a5d67df7a420e398bd2/chain/src/main/scala/org/bitcoins/chain/models/CompactFilterDAO.scala#L67)

![Screenshot from 2020-04-08 17-36-39](https://user-images.githubusercontent.com/3514957/78840188-ba53f400-79bf-11ea-9e86-df5a7e4574dc.png)

PR #1325, opened 2020-04-09
Make 'maxHeightQuery' vals inside of BlockHeaderDAO, CompactFilterHea…

Christewart:
…derDAO, CompactFilterDAO

Begins to address #1323 

We don't need to re-create the max height query everytime as it is a static sql query just doing a simple `SELECT` query.


nkohen (approved):
ACK

benthecarman (approved):

Issue #672, opened 2019-08-04
Get rid of hard Await.result() in P2PClient message processing


Christewart:
#655 introduced the use of `Await.result()` in `P2PClient`. Since computations in `PeerMessageReceiver` return `Future`, we need to wait for that Future to complete before we start processing the next message.

The reason we need this to be synchronous is because we could have messages that are dependent on a previous message. For instance, if we receive multiple `HeadersMessages`, to be able to construct a blockchain, we need to process them in a specific order. We can't process `headersMessage2` before `headersMessage1`  because we won't be able to connect the tips.

@nkohen proposed this solution

https://stackoverflow.com/questions/36592074/wait-on-asynchronous-future-call-before-processing-next-message-in-akka

here is a convo

https://github.com/bitcoin-s/bitcoin-s/pull/655#discussion_r308217009

PR #1326, opened 2020-04-09
Use FutureUtil.foldLeftAsync() to process messages we parsed on the p…

Christewart:
…2p network. This moves the blocking with Await.result() from inside of each message we process, to after the entire batch of messages is processed

I don't think this will have much performance improvement. It is a cleaner refactor, and a stepping stone to figuring out how to solve #672


nkohen (approved):
ACK

benthecarman (approved):

PR #1327, opened 2020-04-09
Call getFilterHeaderCount async inside of nextFilterHeaderBatchRange …

Christewart:
…while we are fetching our start height for fetching filters

Basically just moving calls outside of `for` expressions so they can be executed asynchronously.


nkohen (approved):
I'm okay with this assuming it is alright for the actual count values to change in between completion of the future and reading it (i.e. if we are short by one it is okay)

benthecarman (approved):

PR #1328, opened 2020-04-09
Move Json Reader, Writers, and Serializers to App Commons

benthecarman:


build.sbt
```diff
@@ -201,11 +205,18 @@ lazy val coreTest = project
     testkit
   )
 
+lazy val appCommons = project
+  .in(file("app/app-commons"))
```
Christewart: so in my opinion this shouldn't be in the `app/` as it isn't a standalone application like our server/cli/dlc client.

project/Deps.scala
```diff
@@ -147,6 +153,13 @@ object Deps {
 
   val chainTest = List()
 
+  def commons(scalaVersion: String) = List(
```
Christewart: Please call it `appCommons` because we do have `dbCommons` too.

Christewart (changes requested):

Christewart (approved):

PR #1330, opened 2020-04-09
[DO NOT MERGE] Rebased Secp256k1 Schnorr to be BIP 340 compliant

nkohen:
Built on top of #1310 and uses this branch of secp: https://github.com/bitcoin-s/secp256k1/pull/4

Includes full bindings to Secp256k1 Schnorr implementation, as well as a full BIP 340 implementation in Bouncy Castle, with nice types and full testing!


core/src/main/scala/org/bitcoins/core/crypto/FieldElement.scala
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  def apply(num: BigInt): FieldElement = {
+    FieldElement.fromByteArray((num % N).toByteArray)
+  }
+
+  def apply(num: BigInteger): FieldElement = {
+    FieldElement.fromByteArray(num.mod(N).toByteArray)
+  }
```
benthecarman: Could one of these call the other?

core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -0,0 +1,95 @@
...
+
+object SchnorrNonce extends Factory[SchnorrNonce] {
+
+  @tailrec
+  def fromBytes(bytes: ByteVector): SchnorrNonce = {
+    require(bytes.length <= 33,
+            s"Schnorr Nonce must be less than 33 bytes, got $bytes")
+
+    if (bytes.length == 32)
+      new SchnorrNonce(bytes)
```
benthecarman: nit: add braces

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -203,8 +203,8 @@ sealed abstract class ExtPrivateKey
     val tweak = if (Secp256k1Context.isEnabled) {
       NativeSecp256k1.privKeyTweakAdd(il.toArray, key.bytes.toArray)
     } else {
-      val sum = BouncyCastleUtil.addNumbers(key.bytes, il)
-      sum.toByteArray
+      val sum = key.fieldElement.add(FieldElement(il))
+      sum.bytes.toArray
     }
     val childKey = ECPrivateKey(ByteVector(tweak))
```
benthecarman: If `Secp256k1Context` is disabled you are doing `ByteVector` -> `Array[Byte]` -> `ByteVector`, would be better to have `tweak` just be a `ByteVector` and convert the ` NativeSecp256k1.privKeyTweakAdd` return to a `ByteVector`

core/src/main/scala/org/bitcoins/core/crypto/SchnorrDigitalSignature.scala
```diff
@@ -0,0 +1,17 @@
...
+import scodec.bits.ByteVector
+
+case class SchnorrDigitalSignature(rx: SchnorrNonce, sig: FieldElement)
+    extends NetworkElement {
+  override def bytes: ByteVector = rx.bytes ++ sig.bytes
+}
+
+object SchnorrDigitalSignature extends Factory[SchnorrDigitalSignature] {
+  override def fromBytes(bytes: ByteVector): SchnorrDigitalSignature = {
+    SchnorrDigitalSignature(SchnorrNonce(bytes.take(32)),
```
benthecarman, line 16: Should have a requirement that it is only 64 bytes

core-test/src/test/scala/org/bitcoins/core/crypto/ECPrivateKeyTest.scala
```diff
@@ -99,4 +100,16 @@ class ECPrivateKeyTest extends BitcoinSUnitTest {
...
+  it must "successfully negate itself" in {
+    forAll(CryptoGenerators.nonZeroPrivKey) { privKey =>
+      val negPrivKey = privKey.negate
+      val pubKey = privKey.publicKey
+      val negPubKey = negPrivKey.publicKey
+      assert(pubKey.bytes.tail == negPubKey.bytes.tail)
+      assert(pubKey.bytes.head != negPubKey.bytes.head)
+      assert(
+        privKey.fieldElement.add(negPrivKey.fieldElement) == FieldElement.zero)
+    }
```
benthecarman, line 112: Should there be an assertion for adding the pub keys as well?

core-test/src/test/scala/org/bitcoins/core/crypto/FieldElementTest.scala
```diff
@@ -0,0 +1,93 @@
...
+          assert(
+            fe2.subtract(fe1).toBigInteger == fe2.toBigInteger.subtract(
+              fe1.toBigInteger))
+        }
+    }
+  }
+
+  it must "wrap around correctly" in {
+    val nMinusOne = FieldElement(
+      "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140")
```
benthecarman: Should this be a member of the `FieldElement` object?

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrDigitalSignatureTest.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  it must "must not reuse R values" in {
+    forAll(CryptoGenerators.privateKey,
+           NumberGenerator.bytevector(32),
+           NumberGenerator.bytevector(32)) {
+      case (privKey, bytes1, bytes2) =>
+        val sig1 = privKey.schnorrSign(bytes1)
+        val sig2 = privKey.schnorrSign(bytes2)
+        assert(sig1.rx != sig2.rx)
+    }
```
benthecarman: It's basically impossible, but check that bytes1 and bytes2 aren't equal

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrNonceTest.scala
```diff
@@ -0,0 +1,39 @@
...
+    assertThrows[IllegalArgumentException](
+      SchnorrNonce(
+        "676f8c22de526e0c0904719847e63bda47b4eceb6986bdbaf8695db362811a"))
+
+    assertThrows[IllegalArgumentException](
+      SchnorrNonce(
+        "676f8c22de526e0c0904719847e63bda47b4eceb6986bdbaf8695db362811a010203"))
+  }
+
+  it must "fail for invalid x coordinate in" in {
```
benthecarman: nit: remove "in"

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrPublicKeyTest.scala
```diff
@@ -0,0 +1,39 @@
...
+    assertThrows[IllegalArgumentException](
+      SchnorrPublicKey(
+        "676f8c22de526e0c0904719847e63bda47b4eceb6986bdbaf8695db362811a"))
+
+    assertThrows[IllegalArgumentException](
+      SchnorrPublicKey(
+        "676f8c22de526e0c0904719847e63bda47b4eceb6986bdbaf8695db362811a010203"))
+  }
+
+  it must "fail for invalid x coordinate in" in {
```
benthecarman: nit: remove "in"

core-test/src/test/scala/org/bitcoins/core/crypto/ECPrivateKeyTest.scala
nkohen: I don't think that is necessary given that this is a test for the private keys and that the public keys are being tested for relative correctness in the above two assertions

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
```diff
@@ -0,0 +1,282 @@
...
+      secKey: ECPrivateKey,
+      auxRand: ByteVector,
+      msg: ByteVector,
+      expectedSig: SchnorrDigitalSignature): Assertion = {
+    val secpSig = secKey.schnorrSign(msg, auxRand)
+    val bouncyCastleSig = BouncyCastleUtil.schnorrSign(msg, secKey, auxRand)
+    assert(secpSig == expectedSig,
+           s"Test $index failed signing for libsecp256k1")
+    assert(bouncyCastleSig == expectedSig,
+           s"Test $index failed signing for Bouncy Castle")
```
Christewart, line 24: Just to be extra defensive: 

```assert(bouncyCastleSig == secpSig)```

this passes for me locally, but it's best to be explicit in my opinion

File: core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala [hunk 11]
```diff
@@ -0,0 +1,282 @@
...
+      msg: ByteVector,
+      sig: SchnorrDigitalSignature,
+      expectedResult: Boolean,
+      comment: String): Assertion = {
+    val secpResult = pubKey.verify(msg, sig)
+    val bouncyCastleResult = BouncyCastleUtil.schnorrVerify(msg, pubKey, sig)
+    assert(secpResult == expectedResult,
+           s"Test $index failed verification for libsecp256k1: $comment")
+    assert(bouncyCastleResult == expectedResult,
+           s"Test $index failed verification for Bouncy Castle: $comment")
```
Christewart, line 40: ```assert(bouncyCastleResult == secpResult)``` (same comment as above)

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
```diff
@@ -0,0 +1,282 @@
...
+            testSign(index, secKey, auxRand, msgBytes, schnorrSig)
+          case _ => ()
+        }
+
+        testVerify(index, pk, msgBytes, schnorrSig, result, comment)
+      case (Failure(_), _) |
+          (_, Failure(_)) => // Must be verify only test resulting in false
+        assert(secKeyOpt.isEmpty)
+        assert(auxRandOpt.isEmpty)
+        assert(!result, s"Test $index failed to parse signature: $comment")
```
Christewart, line 73: nit: `"Test $index _must_ fail to parse signature: $comment"`

right?

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
```diff
@@ -0,0 +1,282 @@
...
+    }
+  }
+
+  it must "pass the BIP 340 test-vectors" in {
```
Christewart: nit: `pass the BIP 340 test-vectors with both native secp256k1 binary and bouncy castle`

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
```diff
@@ -0,0 +1,282 @@
...
+    test(
+      index = 0,
+      secKeyOpt = Some(
```
Christewart: I think we should definitely read the CSV file that that is here rather than hard coding the test vectors into the Scala file itself. 

It's easier to maintain that way -- I can just replace the new csv file and have the latest test vectors that are associated with BIP340. With the current setup it seems _really_ hard to look through the files and tell if we need to update these test vectors or not.

core-test/src/test/scala/org/bitcoins/core/crypto/FieldElementTest.scala
```diff
@@ -0,0 +1,92 @@
...
+          fe1
+            .add(fe2)
+            .toBigInteger == fe1.toBigInteger.add(fe2.toBigInteger).subtract(N))
+    }
+  }
+
+  it must "subtract numbers correctly" in {
+    forAll(CryptoGenerators.fieldElement, CryptoGenerators.fieldElement) {
+      case (fe1, fe2) =>
+        if (fe1.toBigInteger.compareTo(fe2.toBigInteger) > 0) {
```
Christewart, line 53: can you please use intermediate `val`s so that the code is readable? This applies elsewhere in this file too.

core-test/src/test/scala/org/bitcoins/core/crypto/SchnorrDigitalSignatureTest.scala
```diff
@@ -0,0 +1,139 @@
...
+        val sig2 = privKey.schnorrSignWithNonce(message2, nonce)
+
+        // s1 = nonce + e1*privKey
+        val s1 = sig1.sig
+        // s2 = nonce + e2*privKey
+        val s2 = sig2.sig
+
+        // When signing a message you actually sign SHA256_challenge(Rx || pubKey || message)
+        val e1Bytes =
+          CryptoUtil
```
Christewart: intermediate `val`s please

core/src/main/scala/org/bitcoins/core/crypto/BouncyCastleUtil.scala
```diff
@@ -111,4 +108,80 @@ object BouncyCastleUtil {
...
+  def schnorrSignWithNonce(
+      dataToSign: ByteVector,
+      privateKey: ECPrivateKey,
+      nonceKey: ECPrivateKey): SchnorrDigitalSignature = {
+    val rx = nonceKey.schnorrNonce
+    val k = nonceKey.nonceKey.fieldElement
+    val x = privateKey.schnorrKey.fieldElement
+    val e = CryptoUtil
+      .taggedSha256(rx.bytes ++ privateKey.schnorrPublicKey.bytes ++ dataToSign,
+                    "BIP340/challenge")
```
Christewart: Can we put this as a constant some where? Maybe also cache the hash of this so it doesn't need to be re-hashed everytime?

core/src/main/scala/org/bitcoins/core/crypto/FieldElement.scala
```diff
@@ -0,0 +1,154 @@
...
+import java.math.BigInteger
+
+import org.bitcoins.core.protocol.NetworkElement
+import org.bitcoins.core.util.Factory
+import org.bouncycastle.math.ec.ECPoint
+import scodec.bits.ByteVector
+
+import scala.util.Try
+
+case class FieldElement(bytes: ByteVector) extends NetworkElement {
```
Christewart, line 18: scaladoc

Christewart (changes requested):

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
nkohen: No don't think so, `!result` means it should have succeeded

core/src/main/scala/org/bitcoins/core/crypto/FieldElement.scala
```diff
@@ -0,0 +1,160 @@
...
+    val neg = N.subtract(getBigInteger(fe.bytes))
+    FieldElement(neg)
+  }
+
+  def computePoint(fe: FieldElement): ECPoint = G.multiply(fe.toBigInteger)
+
+  /** Computes the inverse (mod M) of the input using the Euclidean Algorithm (log time)
+    * Cribbed from [[https://www.geeksforgeeks.org/multiplicative-inverse-under-modulo-m/]]
+    */
+  def computeInverse(fe: FieldElement): FieldElement = {
```
nkohen, line 130: Use `BigInteger.modInverse` instead

core-test/src/test/scala/org/bitcoins/core/crypto/BIP340Test.scala
benthecarman: The comment will show up when `result == true` though

nkohen: sure but `result == true` means the test failed in this case as it should never expect a passing check if it can't parse things

nkohen:
Was pushed to the `schnorr-dlc` branch

PR #1332, opened 2020-04-11
Require that addresses are the correct network when sending

benthecarman:
I needed to add a small change to the bitcoind rpc so it has the correct network parameters because regtest addresses also parse as testnet addresses

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -37,7 +37,7 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
       feeRate: FeeUnit,
       fromAccount: AccountDb): Future[Transaction] = {
     require(
-      address.networkParameters.p2pkhNetworkByte == networkParameters.p2pkhNetworkByte,
+      address.networkParameters.isSameNetworkBytes(networkParameters),
       s"Cannot send to address on other network, got ${address.networkParameters}"
     )
     logger.info(s"Sending $amount to $address at feerate $feeRate")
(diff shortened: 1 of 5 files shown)


Christewart (approved):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -37,8 +37,9 @@ sealed abstract class Wallet extends LockedWallet with UnlockedWalletApi {
       feeRate: FeeUnit,
       fromAccount: AccountDb): Future[Transaction] = {
     require(
-      address.networkParameters == networkParameters,
-      s"Cannot send to address on other network, got ${address.networkParameters}")
+      address.networkParameters.p2pkhNetworkByte == networkParameters.p2pkhNetworkByte,
```
Christewart: isn't this only going to work for `p2pkh`? Wont' we need this for `p2sh` and `segwit` as well?

Christewart (approved):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: The reason for using `p2pkhNetworkByte` is because on testnet and regtest they are equal. When parse an address ie `2N4WCUEPY3Y52PVBn5Wgz5c1RoMqmbUGXri`, it will be parsed as a testnet address but it is also valid on regtest. So if we were just comparing `networkParameters` then it'd be marked invalid even though it is not. So, we compare `p2pkhNetworkByte` because they are equal across testnet and regtest. We could have used `p2sh` or `segwit` but it doesn't really matter, we are just trying to compare the `networkParameters` between if it is valid on mainnet or test chains(testnet and regtest)

Christewart: So why don't we just make a nice helper method inside in `NetworkParameters` that says `isSameNetworkBytes()` or something?

benthecarman: That is a good idea

PR #1338, opened 2020-04-15
Prevent wallet from initating losing DLCs 

benthecarman:
Should protect users from accidentally screwing themselves.


nkohen (approved):
ACK

PR #1340, opened 2020-04-15
Verify Signatures given to DLCWallet

benthecarman:
I needed to rework some of the `DLCClient` parts so I can get the unsigned versions of the transactions. Also, a bug wound found in the `TransactionSignatureChecker`, that will be added to master along with tests. Another thing that was added was a verify input function for PSBTs, this was only used for the funding inputs because we should have all the necessary data to be able to create a valid input there, however, for the CETs we would have to produce signatures of our own so we only verify the other party's signature instead of the whole input.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -82,6 +82,22 @@ case class PSBT(
     PSBT(global, inputs, outputs)
   }
 
+  def finalizeInput(index: Int): Try[PSBT] = {
+    require(index >= 0 && index < inputMaps.size,
+            s"Index must be within 0 and the number of inputs, got: $index")
+    val inputMap = inputMaps(index)
+    if (inputMap.isFinalized) {
+      Success(this)
+    } else {
+      inputMap.finalize(transaction.inputs(index)).map { finalizedInputMap =>
+        val newInputMaps =
+          inputMaps.updated(index, finalizedInputMap)
+
+        PSBT(globalMap, newInputMaps, outputMaps)
+      }
+    }
+  }
+
   /** Finalizes this PSBT if possible, returns a Failure otherwise
     * @see [[https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer]]
     */
@@ -522,6 +538,51 @@ case class PSBT(
       inputMaps.updated(inputIndex, InputPSBTMap(newElements))
 
     PSBT(globalMap, newInputMaps, outputMaps)
+  }
+
+  def verifyFinalizedInput(index: Int): Boolean = {
+    val inputMap = inputMaps(index)
+    require(inputMap.isFinalized, "Input must be finalized to verify")
+
+    val wUtxoOpt = inputMap.witnessUTXOOpt
+    val utxoOpt = inputMap.nonWitnessOrUnknownUTXOOpt
+
+    val newInput = {
+      val input = transaction.inputs(index)
+      val scriptSigOpt = inputMap.finalizedScriptSigOpt
+      val scriptSig =
+        scriptSigOpt.map(_.scriptSig).getOrElse(EmptyScriptSignature)
+      TransactionInput(input.previousOutput, scriptSig, input.sequence)
+    }
+
+    val tx = transaction.updateInput(index, newInput)
+
+    wUtxoOpt match {
+      case Some(wUtxo) =>
+        inputMap.finalizedScriptWitnessOpt match {
+          case Some(scriptWit) =>
+            val wtx = {
+              val wtx = WitnessTransaction.toWitnessTx(transaction)
+              wtx.updateWitness(index, scriptWit.scriptWitness)
+            }
+            val output = wUtxo.witnessUTXO
+
+            ScriptInterpreter.verifyInputScript(wtx, index, output)
+          case None =>
+            false
+        }
+      case None =>
+        utxoOpt match {
+          case Some(utxo) =>
+            val input = tx.inputs(index)
+            val output =
+              utxo.transactionSpent.outputs(input.previousOutput.vout.toInt)
+
+            ScriptInterpreter.verifyInputScript(tx, index, output)
+          case None =>
+            false
+        }
+    }
   }
 
   /**
--- a/dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
@@ -31,6 +31,7 @@ import org.bitcoins.commons.jsonmodels.dlc.{
 import scodec.bits.ByteVector
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
 /** This case class allows for the construction and execution of
   * Discreet Log Contracts between two parties.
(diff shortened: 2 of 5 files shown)


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -524,6 +524,51 @@ case class PSBT(
...
+      case Some(wUtxo) =>
+        inputMap.finalizedScriptWitnessOpt match {
+          case Some(scriptWit) =>
+            val wtx = {
+              val wtx = WitnessTransaction.toWitnessTx(transaction)
+              wtx.updateWitness(index, scriptWit.scriptWitness)
+            }
+            val output = wUtxo.witnessUTXO
+
+            ScriptInterpreter.verifyInputScript(wtx, index, output)
```
nkohen, line 570: Do we run into issues if there is a `p2wsh` elsewhere which has redeem script witnesses that aren't included?

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
```diff
@@ -238,6 +239,39 @@ case class DLCClient(
...
+          val idx = index + remoteTweak
+
+          psbt
+            .addWitnessUTXOToInput(remoteFundingInputs(index).output, idx)
+            .addSignatures(sigs, idx)
+            .inputMaps(idx)
+            .finalize(psbt.transaction.inputs(idx)) match {
+            case Success(inputMap) =>
+              val newInputMaps = psbt.inputMaps.updated(idx, inputMap)
+              val finalized = psbt.copy(inputMaps = newInputMaps)
```
nkohen: I think this should just be an added method to `PSBT` called `finalizeInput` which takes an index and returns a new PSBT

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
```diff
@@ -238,6 +239,39 @@ case class DLCClient(
...
+
+              ret && finalized.verifyFinalizedInput(idx)
```
nkohen: should probably make this whole thing an `if` statement on `ret` rather than putting it at the bottom so that you don't do a ton of work if the first one fails

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -376,11 +376,70 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
+                             dlcOffer,
+                             dlcAccept,
+                             fundingInputs,
+                             outcomeSigs)
+    } yield {
+      val correctNumberOfSigs = sign.cetSigs.outcomeSigs.size == client.outcomes.size
+
+      correctNumberOfSigs && sign.cetSigs.outcomeSigs.foldLeft(true) {
+        case (ret, (outcome, sig)) =>
+          ret && client.verifyCETSig(outcome, sig)
```
nkohen, line 394: ditto above for `ret`, and also you could do `sign.cetSigs.outcomeSigs.foldLeft(correctNumberOfSigs) { ...`

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
```diff
@@ -502,28 +571,41 @@ abstract class DLCWallet extends LockedWallet with UnlockedWalletApi {
...
+      outcomeSigDbs: Vector[DLCCETSignatureDb]): Future[(DLCClient, SetupDLC)] = {
 
-      setupF.map((client, _))
-    }
+    clientFromDb(dlcDb, dlcOffer, dlcAccept, fundingInputs, outcomeSigDbs)
+      .flatMap { client =>
+        val outcomeSigs = outcomeSigDbs.map(_.toTuple).toMap
+
+        val setupF = if (dlcDb.isInitiator) {
+          // TODO: Note that the funding tx in this setup is not signed
```
nkohen, line 589: not sure what you mean here

benthecarman: This comment was there previously, is a note that we could maybe tell the user that if you're the initiator you don't have the other party's funding sigs.

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
benthecarman: This shouldn't be needed because scala has short circuiting implemented, so if the first part of the `&&` is false, it won't evaluate the rest. Also with `||`, if the first part is `true` then it won't evaluate the rest.

wallet/src/main/scala/org/bitcoins/wallet/DLCWallet.scala
benthecarman: This shouldn't be needed because scala has short circuiting implemented, so if the first part of the && is false, it won't evaluate the rest. Also with ||, if the first part is true then it won't evaluate the rest.

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
benthecarman: From the docs
image

benthecarman: actually seems `foldLeft` does not use short-circuiting however, still applies to other comments

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Will take a look and make a PR on master if there is

benthecarman: Looks like it should be good in PSBTs

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
nkohen: Scala does have short circuiting but that only applies to the second argument to `&&`, not to all of the code that constructs `finalized` and `idx` here (analogous statement elsewhere). So unless you want to make all the `val`s `lazy`, which I think would be a little weird to read, I'd suggest using `if`

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
```diff
@@ -238,6 +239,38 @@ case class DLCClient(
...
+    remoteSigs.zipWithIndex
+      .foldLeft(true) {
+        case (ret, ((outPoint, sigs), index)) =>
+          if (ret) {
+            require(psbt.transaction.inputs(index).previousOutput == outPoint,
+                    "Adding signature for incorrect input")
+
+            val idx = index + remoteTweak
+
+            // TODO: add funding witness and redeem scripts
```
nkohen, line 260: is there maybe an issue filed for this TODO and the one below?

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
```diff
@@ -370,12 +428,22 @@ case class DLCClient(
...
+    WitnessTransaction(
+      TransactionConstants.validLockVersion,
+      Vector(fundingInput),
+      outputs.filter(_.value >= Policy.dustThreshold),
+      timeouts.contractMaturity.toUInt32,
+      TransactionWitness(Vector(P2WSHWitnessV0(fundingSPK)))
     )
+  }
+
+  /** Constructs CET given sig*G, the funding tx's UTXOSpendingInfo and payouts */
```
nkohen: nit: signed or unsigned?

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
benthecarman: Yes, I moved this to an if because of that

dlc/src/main/scala/org/bitcoins/dlc/DLCClient.scala
benthecarman: https://github.com/bitcoin-s/bitcoin-s/issues/1344

nkohen (approved):
ACK

PR #1341, opened 2020-04-16
TransactionSignatureChecker bug fix and tests

benthecarman:


core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -18,6 +20,32 @@ import scala.annotation.tailrec
...
+    checkSignature(txSignatureComponent, PartialSignature(pubKey, signature))
+
+  def checkSignature(
+      txSignatureComponent: TxSigComponent,
+      partialSignature: PartialSignature): TransactionSignatureCheckerResult = {
+    checkSignature(txSignatureComponent,
+                   txSignatureComponent.output.scriptPubKey.asm.toList,
+                   partialSignature.pubKey,
+                   partialSignature.signature,
+                   Policy.standardFlags)
```
nkohen: If this is the default value might as well leave it off

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -18,6 +20,32 @@ import scala.annotation.tailrec
...
+
+  def checkSignature(
+      txSignatureComponent: TxSigComponent,
+      script: Seq[ScriptToken],
+      partialSignature: PartialSignature): TransactionSignatureCheckerResult =
+    checkSignature(txSignatureComponent,
+                   script,
+                   partialSignature.pubKey,
+                   partialSignature.signature,
+                   Policy.standardFlags)
```
nkohen: ditto

core-test/src/test/scala/org/bitcoins/core/crypto/TransactionSignatureCheckerTest.scala
```diff
@@ -1,8 +1,229 @@
...
 
 /**
   * Created by chris on 2/29/16.
   */
-class TransactionSignatureCheckerTest extends BitcoinSUnitTest {}
+class TransactionSignatureCheckerTest extends BitcoinSUnitTest {
+
+  behavior of "TransactionSignatureChecker"
+
+  it must "validate a P2PK sig" in {
```
nkohen, line 36: for all of these validation tests you should add invalidation tests or asserts for if something small is changed

nkohen (approved):
ACK assuming CI passes

Christewart (approved):

PR #1342, opened 2020-04-16
Add low R signing

### LINKED ISSUE #1312: Implement low R signing
Here is where this was implemented in bitcoin-core: https://github.com/bitcoin/bitcoin/pull/13666

Essentially the idea is to deterministically retry nonces until the `R` value is small enough, resulting in signatures that are always under 72 bytes.

We should probably implement this and make it available through a flag in `ECPrivateKey.sign`.

nkohen:
[Introduces signing with entropy to the JNI](https://github.com/bitcoin-s/secp256k1/pull/7).

Also introduces and tests low R signing as was introduced to bitcoin-core [here](https://github.com/bitcoin/bitcoin/pull/13666).

Closes #1312


core-test/src/test/scala/org/bitcoins/core/crypto/SignTest.scala
```diff
@@ -1,34 +1,56 @@
...
+
+      sigF1.flatMap { sig1 =>
+        assert(pubKey.verify(bytes, sig1))
+        assert(sig1.bytes.length <= 70) // This assertion fails if Low R is not used
+
+        sigF2.map { sig2 =>
+          assert(sig1.bytes == sig2.bytes)
+          assert(sig1 == sig2)
+        }
+      }
```
benthecarman: nit:
```scala
for {
	sig1 <- sigF1
	sig2 <- sigF2
} yield {
	assert(pubKey.verify(bytes, sig1))
	assert(sig1.bytes.length <= 70) // This assertion fails if Low R is not used
	assert(sig1.bytes == sig2.bytes)
	assert(sig1 == sig2)
}
```

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -26,27 +27,84 @@ trait Sign {
...
+  def signWithEntropyFunction: (
+      ByteVector,
+      ByteVector) => Future[ECDigitalSignature]
+
+  def signWithEntropyFuture(
+      bytes: ByteVector,
+      entropy: ByteVector): Future[ECDigitalSignature] =
+    signWithEntropyFunction(bytes, entropy)
+
+  private def signLowRFuture(bytes: ByteVector, startAt: UInt32)(
```
benthecarman: `@tailrec`

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
```diff
@@ -26,27 +27,84 @@ trait Sign {
...
+    val sig1F: Future[ECDigitalSignature] =
+      signWithEntropyFunction(bytes, startBytes)
+    val sig2F: Future[ECDigitalSignature] =
+      signWithEntropyFunction(bytes, startPlusOneBytes)
+    sig1F.flatMap { sig1 =>
+      if (sig1.bytes.length <= 70) {
+        Future.successful(sig1)
+      } else {
+        sig2F.flatMap { sig2 =>
+          if (sig2.bytes.length <= 70) {
```
benthecarman: In the bitcoin-core PR they have this
```
// In DER serialization, all values are interpreted as big-endian, signed integers. The highest bit in the integer indicates
// its signed-ness; 0 is positive, 1 is negative. When the value is interpreted as a negative integer, it must be converted
// to a positive value by prepending a 0x00 byte so that the highest bit is 0. We can avoid this prepending by ensuring that
// our highest bit is always 0, and thus we must check that the first byte is less than 0x80.
```

I believe this would be more efficient so we should probably use this

benthecarman (changes requested):
It might be useful to make a `ECDummyLowRSignature` as well

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
nkohen: actually it isn't tail recursive sadly due to the fact that the call is made within a `map`

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
nkohen: Thing is if you look here: https://github.com/bitcoin-s/secp256k1/pull/7/files#diff-0857e942f42747dfba3759e3319abf4fR135

We DER serialize in libsecp256k1 before returning through the JNI so we don't have access to the compact version. So it would be less efficient to do the processing to get this byte

core/src/main/scala/org/bitcoins/core/crypto/Sign.scala
benthecarman: :(

benthecarman (approved):

benthecarman (approved):

benthecarman (commented):
This looks good to me, however, to implement this into the wallet will be tricky with how it is. Currently when we get signers, they will use only use the `signFunction` and not the `signLowR` function. 

Maybe it would be worth creating a `LowRSign` type that extends `Sign` and defaults to low r signing.

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -24,28 +24,85 @@ trait Sign {
...
+      bytes: ByteVector,
+      entropy: ByteVector): Future[ECDigitalSignature] =
+    signWithEntropyFunction(bytes, entropy)
+
+  private def signLowRFuture(bytes: ByteVector, startAt: Long)(implicit
+      ec: ExecutionContext): Future[ECDigitalSignature] = {
+    val startBytes = ByteVector.fromLong(startAt).padLeft(32)
+    val startPlusOneBytes = ByteVector.fromLong(startAt + 1).padLeft(32)
+
+    val sig1F: Future[ECDigitalSignature] =
```
Christewart: Hmm, so it appears the low r signing stuff won't play well with rate limiting schemes in a key manager? If I understand correctly the idea is just to keep generating signatures until you get one below 70 bytes?

Christewart: Perhaps we should keep this simple for now and just generate one signature and check it and then recurse rather than generating two signatures in parallel?

Christewart (approved):

PR #1343, opened 2020-04-17
Increased test coverage by 0.97%

nkohen:
It is really annoying that the old version of scala gets such low coverage compared to 2.13, so I increased coverage because pretty much all feature branches are failing coverage on old scala versions.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
@@ -3,7 +3,7 @@ package org.bitcoins.core.util
 import org.bitcoins.core.number.{UInt5, UInt8}
 import org.bitcoins.core.protocol.BtcHumanReadablePart
 import org.bitcoins.core.protocol.ln.LnHumanReadablePart
-import scodec.bits.{BitVector, ByteVector}
+import scodec.bits.ByteVector
 
 import scala.annotation.tailrec
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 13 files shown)


benthecarman (approved):

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -143,17 +143,10 @@ sealed abstract class Bech32 {
           } else {
             val byte = Bech32.charset.indexOf(h.toLower).toByte
 
-            if (byte >= 0 && byte < 32) {
```
Christewart, line 146: why did this get removed? Do we have a guarantee that the byte is in the right range?

nkohen: Yes: `Bech32.charset.contains(h.toLower)` is above and provides this guarantee

nkohen: Which is why this was never being covered

benthecarman: If `h.toLower` is not in the charset, it will return `-1`. So then this check would fail but now there would be no check

nkohen: no, if `h.toLower` is not in the charset, then the check above this one will throw an `IllegalArgumentException`

PR #1351, opened 2020-04-21
Add sendrawtransaction CLI and Api commands

benthecarman:


app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
```diff
@@ -25,5 +27,17 @@ case class NodeRoutes(node: Node)(implicit system: ActorSystem)
...
       }
+
+    case ServerCommand("sendrawtransaction", arr) =>
+      SendRawTransaction.fromJsArr(arr) match {
+        case Failure(exception) =>
+          reject(ValidationRejection("failure", Some(exception)))
+        case Success(SendRawTransaction(tx)) =>
+          complete {
+            node.broadcastTransaction(tx).map { _ =>
+              Server.httpSuccess(s"Broadcast ${tx.txIdBE}")
```
nkohen: nit: Broadcast -> Broadcasted

PR #1348, opened 2020-04-20
Rescan from account

benthecarman:
Needed for wallet account birthdays

PR #1353, opened 2020-04-21
Add wallet creation time for rescans

### LINKED ISSUE #1313: Give Wallets Birthdays
Give wallets a birthday for when they are created/first receive funds, this will allow us to set a better start block for when doing wallet rescans.

benthecarman:
Built on top of #1348 
Closes #1313

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
@@ -41,6 +41,7 @@ trait ChainQueryApi {
       startHeight: Int,
       endHeight: Int): Future[Vector[FilterResponse]]
 
+  /** Gets the block height of the closest block to the given time */
   def epochSecondToBlockHeight(time: Long): Future[Int]
 
 }
--- a/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
@@ -41,6 +41,8 @@ trait ChainQueryApi {
       startHeight: Int,
       endHeight: Int): Future[Vector[FilterResponse]]
 
+  def epochSecondToBlockHeight(time: Long): Future[Int]
+
 }
 
 object ChainQueryApi {
--- a/core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
@@ -1,0 +1,9 @@
+package org.bitcoins.core.util
+
+import java.time.{ZoneId, ZonedDateTime}
+
+object TimeUtil {
+
+  def currentEpochSecond: Long =
+    ZonedDateTime.now(ZoneId.of("UTC")).toEpochSecond
+}
--- a/core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
@@ -1,9 +1,16 @@
 package org.bitcoins.core.util
 
-import java.time.{ZoneId, ZonedDateTime}
+import java.time.Instant
 
 object TimeUtil {
 
-  def currentEpochSecond: Long =
-    ZonedDateTime.now(ZoneId.of("UTC")).toEpochSecond
+  def now: Instant = Instant.now
+
+  /** Returns the current timestamp in milliseconds */
+  def currentEpochMs: Long = now.toEpochMilli
+
+  /** Returns the current timestamp in seconds */
+  def currentEpochSecond: Long = {
+    now.getEpochSecond
+  }
 }
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
@@ -1,6 +1,7 @@
 package org.bitcoins.keymanager
 
 import java.nio.file.{Files, Path}
+import java.time.Instant
 import java.util.NoSuchElementException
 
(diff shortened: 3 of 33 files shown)


key-manager/src/main/scala/org/bitcoins/keymanager/bip39/BIP39KeyManager.scala
```diff
@@ -62,6 +63,9 @@ case class BIP39KeyManager(
 object BIP39KeyManager extends BIP39KeyManagerCreateApi with BitcoinSLogger {
   val badPassphrase = AesPassword.fromString("changeMe").get
 
+  private def getCurrentTimeUTC: Long =
```
Christewart: Let's move this to `BitcoinSUtil` if it doesn't exist there yet, or some sort of `TimeUtil` object in core.

Christewart: Also this should probably documented somewhere that we use `creationTime` to be seconds

testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
```diff
@@ -84,7 +84,7 @@ abstract class SyncUtil extends BitcoinSLogger {
           case BlockStamp.BlockHeight(height) =>
             Future.successful(height)
           case BlockStamp.BlockTime(_) =>
-            Future.failed(new RuntimeException(s"Cannot query by block time"))
+            Future.successful(0)
```
Christewart: Why is this changed? Bitcoin core doesn't allow you to query by block time AFAIK?

testkit/src/main/scala/org/bitcoins/testkit/wallet/WalletTestUtil.scala
```diff
@@ -84,11 +86,15 @@ object WalletTestUtil {
     HDAccount(coin = HDCoin(purpose, HDCoinType.Testnet), index = 1)
   }
 
-  def firstAccountDb = AccountDb(freshXpub(), defaultHdAccount)
+  def firstAccountDb: AccountDb =
+    AccountDb(freshXpub(),
+              defaultHdAccount,
+              ZonedDateTime.now(ZoneId.of("UTC")).toEpochSecond)
```
Christewart: Use our new `TimeUtil.nowSeconds()` that I suggested above :-)

Christewart (changes requested):
I think you didn't add unit tests for rescanning with wallet creation time?

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -37,6 +37,7 @@ sealed trait WalletApi {
 
   val nodeApi: NodeApi
   val chainQueryApi: ChainQueryApi
+  val creationTime: Long
```
Christewart: This should really be typed as `java.time.Instant` -- avoid using `Long` to represent a time at all costs as it is _extremely_ easy to get confused with handling time.

key-manager/src/main/scala/org/bitcoins/keymanager/bip39/BIP39KeyManager.scala
```diff
@@ -23,9 +23,9 @@ import scala.util.{Failure, Success, Try}
 case class BIP39KeyManager(
     private val mnemonic: MnemonicCode,
     kmParams: KeyManagerParams,
-    private val bip39PasswordOpt: Option[String])
+    private val bip39PasswordOpt: Option[String],
+    birthday: Long)
```
Christewart: Same thing applies here, let's make sure not to use `Long` to represent time.

core/src/main/scala/org/bitcoins/core/api/ChainQueryApi.scala
```diff
@@ -41,6 +41,8 @@ trait ChainQueryApi {
       startHeight: Int,
       endHeight: Int): Future[Vector[FilterResponse]]
 
+  def epochSecondToBlockHeight(time: Long): Future[Int]
```
Christewart, line 45: scaladoc

core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
```diff
@@ -0,0 +1,9 @@
+package org.bitcoins.core.util
+
+import java.time.{ZoneId, ZonedDateTime}
+
+object TimeUtil {
+
```
Christewart, line 6:
```scala
object TimeUtil {
  
  def now: Instant = Instant.now
  
  /** Returns the current timestamp in milliseconds */
  def currentEpochMs: Long = now.toEpochMilli
  
  /** Returns the current timestamp in seconds */
  def currentEpochSecond: Long = {
    now.getEpochSecond
  }
}
```

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -111,13 +114,22 @@ object WalletStorage {
...
-      (String, String, String)] = jsonE.flatMap { json =>
+      (String, String, String, Long)] = jsonE.flatMap { json =>
       logger.trace(s"Read encrypted mnemonic JSON: $json")
+      val creationTimeNum = Try(json(CREATION_TIME).num.toLong) match {
+        case Success(value) =>
+          value
+        case Failure(err) if err.isInstanceOf[NoSuchElementException] =>
+          // If no CREATION_TIME is set, we set date to start of bitcoin-s wallet proj
+          // default is Block 555,990 block time 2018-12-28
+          1546042867L
```
Christewart: Can we put this into some util file with a nice name? I don't really like this magic number, and we may want to use it else where.

PR #1356, opened 2020-04-22
Fix sendrawtransaction output again woops

benthecarman:
This one line got me good. Fixed again, should actually be good now


benthecarman:
Is very trivial so just merging

PR #1363, opened 2020-04-23
Batch processing compact filters

### LINKED ISSUE #1282: Cache utxos/addresses when calling `processCompactFilter()`

benthecarman:
Fixes #1282 

The change here is to process filters in batches when we are syncing. Specifically, we process the filters after we have accumulated `chainConfig.filterBatchSize` filters. This also required changing `OnCompactFilterReceived` to `OnCompactFiltersReceived`.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -22,7 +22,7 @@ import scala.concurrent.{ExecutionContext, Future}
 case class DataMessageHandler(
     chainApi: ChainApi,
     callbacks: NodeCallbacks,
-    receivedFilterCount: Int = 0,
+    currentFilterBatch: Vector[CompactFilterMessage] = Vector.empty,
     syncing: Boolean = false)(
     implicit ec: ExecutionContext,
     appConfig: NodeAppConfig,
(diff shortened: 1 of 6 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -22,7 +22,7 @@ import scala.concurrent.{ExecutionContext, Future}
 case class DataMessageHandler(
     chainApi: ChainApi,
     callbacks: NodeCallbacks,
-    receivedFilterCount: Int = 0,
+    currentFilterBatch: Vector[CompactFilterMessage] = Vector.empty,
```
Christewart, line 28: Add a scaladoc for what this is for

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -322,8 +327,8 @@ object DataMessageHandler {
   type OnTxReceived = Transaction => Future[Unit]
 
   /** Callback for handling a received compact block filter */
-  type OnCompactFilterReceived =
-    (DoubleSha256Digest, GolombFilter) => Future[Unit]
+  type OnCompactFiltersReceived =
+    (Vector[DoubleSha256Digest], Vector[GolombFilter]) => Future[Unit]
```
Christewart: Shouldn't this be `Vector[(DoubleSha256Digest, GolombFilter)]`? The way this is implemented currently means that the two vector's sizes could be different

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -75,37 +78,41 @@ case class DataMessageHandler(
...
           }
           newChainApi <- chainApi.processFilter(filter)
-          blockFilter <- Future(
-            BlockFilter.fromBytes(filter.filterBytes, filter.blockHash))
-          _ <- callbacks.executeOnCompactFilterReceivedCallbacks(
-            logger,
-            filter.blockHash,
-            blockFilter)
+          // If we are not syncing or our filter batch is full, process the filters
+          _ <- if (!syncing || batchSizeFull) {
```
rorp, line 104: Initially I implemented compact filters sync this. And I reverted it because 1) it didn't really speed up the sync process; 2) the node was able to sync only a multiple of `batchSizeFull` filter. So if the batch size is 100 and we initially synced 101 filters, we'll process the the 102nd filter only in ~approximately  16 hours, when the next batch is filled up.

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -99,12 +104,14 @@ trait LockedWalletApi extends WalletApi with WalletLogger {
...
-          nodeApi.downloadBlocks(Vector(blockHash))
-        }
+      processFs = blockFilters.map {
+        case (blockHash, blockFilter) =>
+          val matcher = SimpleFilterMatcher(blockFilter)
+          if (matcher.matchesAny(scriptPubKeys.toVector.map(_.asmBytes))) {
+            nodeApi.downloadBlocks(Vector(blockHash))
+          } else FutureUtil.unit
       }
+      _ <- Future.sequence(processFs)
```
rorp: Do we care about the block order here? `Future.sequence` doesn't guarantee the order.

Also, what is error recovery strategy here?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
benthecarman: That's why we do the check on `syncing`. Then it will check every filter immediately after it initially catches up to the chain tip

Christewart (approved):

PR #1365, opened 2020-04-23
 Ignore immature coinbase outputs when funding a transaction 

### LINKED ISSUE #1357: Don't use coinbase outputs for 100 blocks
Bitcoin-S wallet currently has no recognition of coinbase outputs and so addresses for these outputs are sometimes chosen on Regtest for funding transactions (such as when using `SendToAddress`) and these are then rejected by bitcoind as they are not legal. We should have a tag for addresses that are coinbase outputs and our coin selection should not consider these outputs until they are legal (depth 100+)

benthecarman:
Built on #1364 so I could change `FundTransactionHandlingTest` to `fundWalletWithBitcoind` without needing to rewrite the tests.

Fixes #1357

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutPoint.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutPoint.scala
@@ -24,7 +24,7 @@ case class TransactionOutPoint(txId: DoubleSha256Digest, vout: UInt32)
   def ==(outPoint: TransactionOutPoint): Boolean =
     txId == outPoint.txId && vout == outPoint.vout
 
-  def !==(outPoint: TransactionOutPoint): Boolean =
+  def !=(outPoint: TransactionOutPoint): Boolean =
     !(this == outPoint)
 }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionOutPoint.scala
```diff
@@ -20,6 +20,12 @@ case class TransactionOutPoint(txId: DoubleSha256Digest, vout: UInt32)
 
   override def toString: String =
     s"TransactionOutPoint(${txIdBE.hex}:${vout.toBigInt})"
+
+  def ==(outPoint: TransactionOutPoint): Boolean =
+    txId == outPoint.txId && vout == outPoint.vout
+
+  def !==(outPoint: TransactionOutPoint): Boolean =
```
Christewart: why `!==` and not `!=`?

nkohen (approved):

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -58,7 +62,24 @@ trait FundTransactionHandling extends WalletLogger { self: LockedWalletApi =>
...
+      confFs = coinbaseUtxos.map(
+        utxo =>
+          chainQueryApi
+            .getNumberOfConfirmations(utxo.blockHash.get)
+            .map((utxo, _)))
+      confs <- Future.sequence(confFs)
+      immatureCoinbases = confs
+        .filter {
+          case (_, confsOpt) =>
+            confsOpt.isDefined && confsOpt.get > 100
```
Christewart: Let's `100` into `Consensus.coinbaseMaturity` or something like that

Christewart (changes requested):

Christewart (approved):

PR #1366, opened 2020-04-23
Re-enable NeutrinoNodeWithWalletTest for Linux

benthecarman:
Test will be disabled for mac and windows because we do not have a bitcoin core binaries that have neutrino enabled

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -13,12 +13,10 @@ import org.bitcoins.testkit.node.NodeUnitTest.NeutrinoNodeFundedWalletBitcoind
 import org.bitcoins.testkit.node.{NodeTestUtil, NodeUnitTest}
 import org.bitcoins.testkit.wallet.BitcoinSWalletTest
 import org.bitcoins.wallet.api.UnlockedWalletApi
-import org.bitcoins.wallet.config.WalletAppConfig
-import org.scalatest.{DoNotDiscover, FutureOutcome}
+import org.scalatest.FutureOutcome
 
 import scala.concurrent.{Future, Promise}
 
-@DoNotDiscover
 class NeutrinoNodeWithWalletTest extends NodeUnitTest {
 
   /** Wallet config with data directory set to user temp directory */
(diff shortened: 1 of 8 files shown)


node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -29,9 +28,13 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
...
 
   def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    withNeutrinoNodeFundedWalletBitcoind(test,
-                                         callbacks,
-                                         Some(BitcoindVersion.Experimental))
+    if (isLinux) {
+      withNeutrinoNodeFundedWalletBitcoind(test,
+                                           callbacks,
+                                           Some(BitcoindVersion.Experimental))
+    } else {
```
Christewart: Can you add a comment here why this is being done so in the future we remember?

testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
```diff
@@ -50,6 +50,15 @@ trait BaseAsyncTest
 
   override lazy val timeLimit: Span = 5.minutes
 
+  lazy val isLinux: Boolean =
```
Christewart: nit: Can you move these into `BitcoinSUtil` so we can use them elsewhere too?

Christewart (approved):

testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
Christewart: Actually you should probably just use `scala.util.Properties.isMac` or `Properties.isLinux` etc

rorp (changes requested):
In this version the unit test won't be run on any OS other than Linux. It should be Linux-only in CI environment, and it should work for any supported OSes otherwise.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -28,12 +29,17 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
   override type FixtureParam = NeutrinoNodeFundedWalletBitcoind
 
   def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    if (isLinux) {
+    // We need to disable the test on non-linux CI runs
+    // because we do not have a mac binary of the BIP 157
+    // compatible version of bitcoin core
+    val prop = System.getProperty("TEST_COMMAND")
+    val isCI = prop != null && prop.nonEmpty
```
rorp: It could be useful in other unit test, consider to move it in a utility class.

PR #1367, opened 2020-04-23
Fix isMinimalPush to not look for OP_1Negate when pushing 0x4f(79)

### LINKED ISSUE #1361: PSBTTest CI failure
Screenshot from 2020-04-23 06-44-48

Saw this on #1359 which doesn't touch anything related to PSBT

benthecarman:
`isMinimalPush` was incorrectly expecting `OP_1NEGATE` when we were pushing 0x4f (79 in decimal) onto the stack to be valid. This was caused because we were checking `OP_1NEGATE`'s `opcode` not its actual `underlying` value.

Was able to delete this entire case
```
case size if (size == 1 && token.bytes.head == OP_1NEGATE.opCode) =>
            pushOp == OP_1NEGATE
```
because it should be captured by
```
case scriptNumOp: ScriptNumberOperation =>
        scriptNumOp == pushOp
```

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/util/BitcoinScriptUtilTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/util/BitcoinScriptUtilTest.scala
@@ -183,6 +183,20 @@ class BitcoinScriptUtilTest extends BitcoinSUnitTest {
       false)
   }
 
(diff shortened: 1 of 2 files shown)


Christewart (approved):
Nice job hunting this down Ben!

PR #1375, opened 2020-04-26
Re-add SchnorrNonce.freshNonce(), for DLC doc. Add mdoc:to-string to …

Christewart:
…dlc doc to make sure the doc is compiled so we have failures on CI if the API changes

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
@@ -88,8 +88,4 @@ object SchnorrNonce extends Factory[SchnorrNonce] {
   def apply(xCoor: FieldElement): SchnorrNonce = {
     SchnorrNonce(xCoor.bytes)
   }
-
-  def freshNonce(): SchnorrNonce = {
-    ECPrivateKey.freshPrivateKey.schnorrNonce
-  }
 }
--- a/core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
@@ -90,7 +90,6 @@ object SchnorrNonce extends Factory[SchnorrNonce] {
   }
 
   def freshNonce(): SchnorrNonce = {
-    val xcordBytes = ECPrivateKey.freshPrivateKey.publicKey.bytes.tail
-    fromBytes(xcordBytes)
+    ECPrivateKey.freshPrivateKey.schnorrNonce
   }
 }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/crypto/SchnorrNonce.scala
```diff
@@ -88,4 +88,9 @@ object SchnorrNonce extends Factory[SchnorrNonce] {
   def apply(xCoor: FieldElement): SchnorrNonce = {
     SchnorrNonce(xCoor.bytes)
   }
+
+  def freshNonce(): SchnorrNonce = {
+    val xcordBytes = ECPrivateKey.freshPrivateKey.publicKey.bytes.tail
+    fromBytes(xcordBytes)
```
nkohen: nit: `ECPrivateKey.freshPrivateKey.schnorrNonce`

docs/wallet/dlc.md
```diff
@@ -63,8 +64,8 @@ val loseHash = CryptoUtil.sha256(ByteVector("LOSE".getBytes)).flip
 
 (pubKey.bytes ++ rValue.bytes).toHex
 (winHash.bytes ++ Satoshis(100000).bytes ++ loseHash.bytes ++ Satoshis.zero.bytes).toHex
-Schnorr.signWithNonce(winHash.bytes, privKey, nonce).hex
-Schnorr.signWithNonce(loseHash.bytes, privKey, nonce).hex
+NativeSecp256k1.schnorrSignWithNonce(winHash.bytes.toArray, privKey.bytes.toArray, nonce.bytes.toArray)
+NativeSecp256k1.schnorrSignWithNonce(loseHash.bytes.toArray, privKey.bytes.toArray, nonce.bytes.toArray)
```
nkohen: nit: `privKey.schnorrSignWithNonce(_Hash.bytes, nonce)`

docs/wallet/dlc.md
```diff
@@ -48,7 +48,8 @@ f8758d7f03a65b67b90f62301a3554849bde6d00d50e965eb123398de9fd6ea7af05f01f1ca852cf
 
 Note: if you wish to setup your own oracle for testing, you can do so by pasting the following into the `sbt core/console`:
 
-```scala
+```scala mdoc:to-string
+import org.bitcoin._
```
nkohen: This won't be needed given change below

docs/wallet/dlc.md
Christewart: So I don't think this works as these methods take a `ECPRivateKey` and not a `SchnorrNonce` 

```scala
  def schnorrSignWithNonce(
      dataToSign: ByteVector,
      nonce: ECPrivateKey): SchnorrDigitalSignature
```

PR #1379, opened 2020-04-27
Remove Unlocked vs Locked Wallet distinction

benthecarman:
Removes the locked vs unlocked wallet api distinction, this is being done in favor of it being added/changed to part of the key manager. If a user wishes to keep this distinction they should be able to recreate the old `LockedWalletApi` and `UnlockedWalletApi` interfaces.


Christewart (approved):

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -115,18 +115,19 @@ object Main extends App {
...
       nodeApi: Node,
       chainQueryApi: ChainQueryApi,
-      bip39PasswordOpt: Option[String]): Future[UnlockedWalletApi] = {
+      bip39PasswordOpt: Option[String]): Future[WalletApi] = {
     hasWallet().flatMap { walletExists =>
       if (walletExists) {
         logger.info(s"Using pre-existing wallet")
-        val locked = LockedWallet(nodeApi, chainQueryApi)
 
-        // TODO change me when we implement proper password handling
```
nkohen, line 124: nit: keep the todo

nkohen (approved):

PR #1380, opened 2020-04-27
Crypto Project Refactor

nkohen:
Move `core.crypto` into a project called `crypto` that has even less in scope.


nkohen:
Appears I need to rebase as something got in today while I was working on this.

crypto/src/main/scala/org/bitcoins/crypto/NetworkElement.scala
```diff
@@ -1,6 +1,5 @@
-package org.bitcoins.core.protocol
+package org.bitcoins.crypto
```
benthecarman, line 1: Maybe a better abstraction would be to keep `NetworkElement` in core.protocol and create a `Serializable` for crypto, then have `NetworkElement` extend `Serializable`

benthecarman (approved):

Christewart (commented):
So one thing I can see that is a problem the `.travis.yml` file hasn't been modified to add the new `cryptoTest/test` command that needs to be run? Also we should probably add a coverage minimum to `crypto` as well since it's a sensitive module. 

Also i'm fine doing this in another PR, but we need to add a doc on the website for the new module and add the command to add the new jar file in a build.sbt example

Christewart (approved):

PR #1385, opened 2020-04-30
Create logger config for other libraries to configure bitcoin-s logging

### LINKED ISSUE #699: Make a distinction between logging when running standalone binaries vs. using modules as libraries
There's two ways of using Bitcoin-S: by running compiled binaries or including specific modules as libraries. Logging should be handled differently in these cases. 

When running as a binary: 
We should log to the users data directory, following our own logging convention/structure. 

When running as a library: 
We should not provide any logging configuration, and leave that to the user through their preferred way of configuring Logback. 

One way of handling this would be to default to "library"-logging, and then somehow flip a switch in main methods of the binaries.

benthecarman:
Closes #699

This makes it so the Loggers no longer require an `AppConfig`, they now require a `LoggerConfig`. This allows it so people using an bitcoin-s as a library do not need an entire `AppConfig` set and can instead pass in a `LoggerConfig` that they create themselves.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/HttpLogger.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/HttpLogger.scala
@@ -1,11 +1,12 @@
 package org.bitcoins.server
 
-import org.bitcoins.db.{AppConfig, AppLoggers, MarkedLogger}
+import org.bitcoins.db.{AppLoggers, LoggerConfig, MarkedLogger}
 
 /** Exposes access to the HTTP RPC server logger */
 private[bitcoins] trait HttpLogger {
   private var _logger: MarkedLogger = _
-  protected[bitcoins] def logger(implicit config: AppConfig) = {
+  protected[bitcoins] def logger(
+      implicit config: LoggerConfig): MarkedLogger = {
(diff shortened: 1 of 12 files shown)


rorp (approved):

Christewart (changes requested):
So this is not what I had in mind when reworking logging. 

By convention, [logback]() libraries are controlled by a logback.xml file. That is what I want to have happen, NOT having the logging controlled solely by `reference.conf`. 

If we can have both as a solution `reference.conf` and `logback.xml` that is great, but I'm guessing that is going to lead to clashing of logging.

For instance, in our `reference.conf` if I have `wallet = OFF` (wallet logging turned off), but in our `logback.xml` if I have: 
```
 <logger name="org.bitcoins.wallet" level="DEBUG"/>
```
which one should take precedence? Should wallet logging be set to `DEBUG`/`OFF`? Basically, if we want to keep both files, we need to have one file read logging configuration from the other. I would be fine with taking logging out of `reference.conf` though and just using logback.xml similar to what eclalir does: https://github.com/acinq/eclair#logging

http://logback.qos.ch/manual/configuration.html

PR #1386, opened 2020-04-30
Implement KeyManagerLogger

benthecarman:
Built on top of #1465 so CI passes.

Previously, we had defined the `KeyHandling` `LoggerKind` but never used it. Now the `KeyManager` should use the correct logger.


Christewart (commented):
You need to make this configurable from `bitcoin-s.conf` files and add an example to the website

PR #1387, opened 2020-05-01
Add rpcport configuration options in both bitcoin-s.conf and as a com…

Christewart:
…mand line parameter

You can now set `rpcport` with the command line like `--rpcport 12345` or in the `bitcoin-s.conf` file with `bitcoin-s.server.rpcport=12345`

The rpc command line option takes precedent over the `bitcoin-s.conf` if they are both used.

You can also use the `--rpcport` option on the `bitcoin-s-cli`. Unfortunately there is a limitation. You need to pass in the arg _after_ the command you are going to run: `bitcoin-s-cli getblockcount --rpcport=12345`. 

This format will NOT work:

```
bitcoin-s-cli --rpcport=12345 getblockcount
```


nkohen (approved):
I've been looking everywhere and didn't find anything that can help put options before commands, lgtm

PR #1389, opened 2020-05-01
End to End DLC with SbClient

benthecarman:
Starts up bitcoind, eclair, and bitcoin-s server enforcing they have compatible configs. As well as adds rpc commands for sb client.

This should not be merged in it's current state. It currently will always require that elcair and bitcoind be started with every bitcon-s server. This should be a config option or separate type of server.

### DIFF
--- a/app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
+++ b/app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
@@ -6,6 +6,7 @@ import akka.http.scaladsl.model.{HttpRequest, Uri}
 import akka.stream.Materializer
 import akka.util.ByteString
 import org.bitcoins.commons.jsonmodels.eclair.OutgoingPaymentStatus
+import org.bitcoins.commons.jsonmodels.sbclient._
 import org.bitcoins.crypto.{
   AesCrypt,
   AesEncryptedData,
@@ -18,8 +19,51 @@ import org.bitcoins.core.protocol.ln.{LnInvoice, PaymentPreimage}
 import org.bitcoins.eclair.rpc.api.EclairApi
 import play.api.libs.json.{JsError, JsSuccess, JsValue, Json, Reads}
 
-import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
+
+case class SbClient(
+    eclairApi: EclairApi,
+    endpoint: String = "https://test.api.suredbits.com/dlc/v0") {
+
+  def getPublicKey(exchange: Exchange, pair: TradingPair)(implicit
+      system: ActorSystem): Future[SchnorrPublicKey] = {
+    val uri =
+      s"$endpoint/${exchange.toLongString}/${pair.toLowerString}/PublicKey"
+
+    import system.dispatcher
+
+    SbClient.rawRestCall(uri).map(SchnorrPublicKey.fromHex)
+  }
+
+  def requestRValueAndPay(
+      exchange: Exchange,
+      tradingPair: TradingPair,
+      eclairApi: EclairApi)(implicit
+      system: ActorSystem): Future[ECPublicKey] = {
+    val dataF =
+      SbClient.requestAndPay(exchange = exchange,
+                             tradingPair = tradingPair,
+                             requestType = RequestType.RValue,
+                             eclairApi = eclairApi,
+                             endpoint = endpoint)
+    dataF.map(ECPublicKey.fromHex)(system.dispatcher)
+  }
+
+  def requestLastSigAndPay(
+      exchange: Exchange,
+      tradingPair: TradingPair,
+      eclairApi: EclairApi)(implicit
+      system: ActorSystem): Future[SchnorrDigitalSignature] = {
+    val dataF =
+      SbClient.requestAndPay(exchange = exchange,
+                             tradingPair = tradingPair,
+                             requestType = RequestType.LastSig,
+                             eclairApi = eclairApi,
+                             endpoint = endpoint)
+    dataF.map(SchnorrDigitalSignature.fromHex)(system.dispatcher)
+  }
+}
 
 object SbClient {
 
@@ -35,19 +79,6 @@ object SbClient {
           .map(payload => payload.decodeString(ByteString.UTF_8)))
   }
 
-  def getPublicKey(
-      exchange: Exchange,
-      pair: TradingPair,
-      endpoint: String = "https://test.api.suredbits.com/dlc/v0")(implicit
-      system: ActorSystem): Future[SchnorrPublicKey] = {
-    val uri =
-      s"$endpoint/${exchange.toLongString}/${pair.toLowerString}/PublicKey"
-
-    import system.dispatcher
-
-    rawRestCall(uri).map(SchnorrPublicKey.fromHex)
-  }
-
   def restCall(uri: Uri)(implicit system: ActorSystem): Future[JsValue] = {
     rawRestCall(uri).map(Json.parse)(system.dispatcher)
   }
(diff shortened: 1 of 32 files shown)


app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -127,6 +145,133 @@ object Main extends App {
...
+      eclair.instance.zmqConfig.get.rawBlock == bitcoind.instance.zmqConfig.rawBlock,
+      s"Eclair's bitcoind zmq raw block must match bitcoind's, ${eclair.instance.zmqConfig.get.rawBlock} != ${bitcoind.instance.zmqConfig.rawBlock}"
+    )
+    require(
+      eclair.instance.zmqConfig.get.rawTx == bitcoind.instance.zmqConfig.rawTx,
+      s"Eclair's bitcoind zmq raw tx must match bitcoind's, ${eclair.instance.zmqConfig.get.rawTx} != ${bitcoind.instance.zmqConfig.rawTx}"
+    )
+
+    def start: Future[EclairBitcoindPair] = {
+      val f: Runnable = new Runnable {
```
Christewart: I don't think this is needed, can't we just do `eclair.start()`?

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -127,6 +145,133 @@ object Main extends App {
     }
   }
 
+  // todo move me
+  case class EclairBitcoindPair(
+      eclair: EclairRpcClient,
+      bitcoind: BitcoindRpcClient) {
```
Christewart: probably should `extend StartStop`

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -127,6 +145,133 @@ object Main extends App {
...
+      val f: Runnable = new Runnable {
+        override def run(): Unit = {
+          eclair.start()
+          ()
+        }
+      }
+
+//      val startEclairF = eclair.start()
+      f.run()
+      val startBitcoindF = bitcoind.start()
```
Christewart: We probably want to make sure bitcoind is started _before_ eclair, as eclair needs to bind to bitcoind for blockchain information.

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -127,6 +145,133 @@ object Main extends App {
...
+      for {
+        startedBitcoind <- startBitcoindF
+      } yield EclairBitcoindPair(eclair, startedBitcoind)
+    }
+
+    def stop: Future[EclairBitcoindPair] = {
+      val stoppedEclairF = eclair.stop()
```
Christewart: Do want to to stop these in parallel? We probably want to make sure eclair shuts down _completely_ before we stop bitcoind. This way we know that bitcoind can receive all information that eclair sends on shutdown.

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -127,6 +145,133 @@ object Main extends App {
...
+      val stoppedBitcoindF = bitcoind.stop()
+      for {
+        stoppedEclair <- stoppedEclairF
+        stoppedBitcoind <- stoppedBitcoindF
+      } yield EclairBitcoindPair(stoppedEclair, stoppedBitcoind)
+    }
+  }
+
+  val binaryDirectory =
```
Christewart: It seems most of this code is copied from somewhere? Any way we can put it some where so we don't repeat ourselves in both places?

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/config/ConfigUtil.scala
```diff
@@ -13,6 +13,13 @@ object ConfigUtil {
     }
   }
 
+  def getString(config: Config, path: String): Option[String] = {
+    Try(config.getString(path)) match {
```
Christewart: You can do `Try(...).toOption`

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/config/EclairInstance.scala
```diff
@@ -122,13 +145,42 @@ object EclairInstance {
 
     val rpcUri: URI = new URI(s"http://$rpcHost:$rpcPort")
 
+    val bitcoindRpcHost =
```
Christewart: Nice! This would be nice to have in master.

app/server/src/main/scala/org/bitcoins/server/Main.scala
benthecarman: I tried that orginally and it would not work, it only worked after running it in a new thread

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
```diff
@@ -117,8 +140,7 @@ object SbClient {
       tradingPair: TradingPair,
       requestType: RequestType,
       eclairApi: EclairApi,
-      endpoint: String = "https://test.api.suredbits.com/dlc/v0")(
-      implicit system: ActorSystem): Future[String] = {
+      endpoint: String)(implicit system: ActorSystem): Future[String] = {
```
nkohen, line 150: I think that it makes sense to get rid of `endpoint` here and use the class one as it would be weird to have an `SbClient(endpoint1).requestAndPay(endpoint2)`

app/gui/src/main/scala/org/bitcoins/gui/WalletGUI.scala
```diff
@@ -69,10 +70,17 @@ object WalletGUI extends JFXApp {
...
+      override def handle(event: ActionEvent): Unit = model.updateBalance()
+    }
+  }
+
   private val buttonBar = new HBox {
-    children = Seq(getNewAddressButton, sendButton)
+    children = Seq(updateBalanceButton, getNewAddressButton, sendButton)
     alignment = Pos.Center
-    spacing <== width / 2
+    spacing <== width / 3
```
nkohen: nit: Probably best to do `width / children().length`

app/gui/src/main/scala/org/bitcoins/gui/sbclient/GlobalSbClientData.scala
```diff
@@ -0,0 +1,3 @@
+package org.bitcoins.gui.sbclient
+
+object GlobalSbClientData {}
```
nkohen: Reason for having this?

app/server/src/main/scala/org/bitcoins/server/EclairBitcoindPair.scala
```diff
@@ -0,0 +1,88 @@
...
+    eclair.instance.zmqConfig.get.rawBlock == bitcoind.instance.zmqConfig.rawBlock,
+    s"Eclair's bitcoind zmq raw block must match bitcoind's, ${eclair.instance.zmqConfig.get.rawBlock} != ${bitcoind.instance.zmqConfig.rawBlock}"
+  )
+  require(
+    eclair.instance.zmqConfig.get.rawTx == bitcoind.instance.zmqConfig.rawTx,
+    s"Eclair's bitcoind zmq raw tx must match bitcoind's, ${eclair.instance.zmqConfig.get.rawTx} != ${bitcoind.instance.zmqConfig.rawTx}"
+  )
+
+  override def start(): Future[EclairBitcoindPair] = {
+    val f: Runnable = new Runnable {
```
nkohen: probably worth a comment

nkohen (commented):
Looks like there is a merge conflict too

app/dlc-suredbits-client/src/main/scala/org/bitcoins/sbclient/SbClient.scala
benthecarman: This is in the `SbClient` `object`, I think it makes sense to keep it there and have the ability to pass in an endpoint, this makes the tests much cleaner and it can be reused for other things in the future (ie I imagine we could just rename it to PAID api client)

nkohen: ah didn't realize it was in the object

nkohen (approved):
LGTM assuming you've tested it locally.

PR #1398, opened 2020-05-05
Add option to correctly use logback config

### LINKED ISSUE #1239: Verify logging actually works if modules are used as libraries
In #640 we added the ability to adjust logging in a conf file for modules. However i'm unsure how well this actually works if you use a single module instead of the entire project. You can find an example of how this is configured in our [`reference.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/e598d4c37e8a721086b4347b6632433efc31e853/testkit/src/main/resources/reference.conf#L5) file in the testkit project. 

In general, I find our logging configuration _very_ hard to work with for development

#### [`AppLoggers`](https://github.com/bitcoin-s/bitcoin-s/blob/e598d4c37e8a721086b4347b6632433efc31e853/db-commons/src/main/scala/org/bitcoins/db/AppLoggers.scala#L19)

`AppLoggers` is meant to be extended by modules (such as [`WalletLogger`](https://github.com/bitcoin-s/bitcoin-s/blob/a6e21fe43b0e34f22cd3a839e225e814fa6f6550/wallet/src/main/scala/org/bitcoins/wallet/WalletLogger.scala) for the wallet project) so that can have more finely controlled logging configuration over modules. The problem with this is that `logback-classic` already solves this problem with package scoping IMO. 

Our [`BitcoinSLogger`](https://github.com/bitcoin-s/bitcoin-s/blob/839d520206023fa1075a65e19392d9f9b72e3a62/core/src/main/scala/org/bitcoins/core/util/BitcoinSLogger.scala) trait is a traditional logback logger which allows us to control things as normal in the `logback-test.xml` file (you can see config below). This is used in the `core` and `core-test` project and is what we should revert to IMO. See the `Things we need to test` section for why. 

#### Log configuration files
We have 4 logging files you need to worry about _just_ for test cases. 

1. [`testkit/reference.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/resources/reference.conf#L5) - this controls both our [bitcoin-s logging for modules](https://bitcoin-s.org/docs/contributing#logging-when-working-on-bitcoin-s-tests) _and_ [akka logging for test cases](https://bitcoin-s.org/docs/contributing#akka-logging)
2. [`core-test/logback-test.xml`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core-test/src/test/resources/logback-test.xml#L4) - effectively an ON/OFF switch for logging, this is useless in my opinion and just makes things more confusing. We should delete IMO.
3. [`core/common-logback.xml`](https://github.com/bitcoin-s/bitcoin-s/blob/839d520206023fa1075a65e19392d9f9b72e3a62/core/src/main/resources/common-logback.xml) - this is where logging controls actually occur for `core-test` that use `BitcoinSLogger`. 

### Things we need to test 

1. What happens if we are only using one of these modules (i.e. we are using wallet, but _not_ using node). Does this play nice?

2. How do we adjust logging at runtime? 

With logback, you can modify the logging file at runtime and it will adjust your logging level while the application is running. Do we support this behavior? 

http://logback.qos.ch/manual/configuration.html#autoScan

3. If the user's application uses another logging format, such as `logback-classic` 

http://logback.qos.ch/manual/configuration.html#automaticConf

I'd like to hear from others to see if they think this is misguided or things you have experienced as well :+1:

benthecarman:
Fixes #1239

This is essentially a redo of #1086. That PR did fix the problem but introduced more if we want other libraries to be able to use our logger. This PR fixes it properly by having the logger name be prefixed by `org.bitcoins` so every log comes from the same parent logger. This then adds small functionality so we can still use our own logger levels from the `bitcoin-s.conf` file.

This PR also adds a `useLogbackConf` option to the `AppConfig` to determine when to use the logback config. It is defaulted to `false` in the bitcoin-s library so we can set our own with `BitcoinSAppConfig`.


Christewart:
> This PR adds a `useLogbackConf` option to the `AppConfig` to determine when to use the logback config. It is defaulted to `false` in the bitcoin-s library so we can set our own with `BitcoinSAppConfig`.

So I actually think the default should be `true`. We should turn it to false in any application that we package (i.e. `cli`, `app-server` etc). Also, should we have a setting in our `reference.conf` file called `bitcoin-s.logging.logback` that can be set to `true` that does the same thing? 

If we do set `logback=true`, can we start the server with 

```bashrc
./bin/bitcoin-s-server -Dlogback.configurationFile=/path/to/logback 
```

db-commons/src/main/scala/org/bitcoins/db/package..scala
```diff
@@ -29,6 +29,15 @@ package object db {
         None
       }
     }
+
+    /** Returns the boolean at key or the given default value */
```
Christewart, line 33: just noticed this with this file name, but it appears there is a `..scala` (double `..`)

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
```diff
@@ -74,8 +75,10 @@ object NodeAppConfig {
   /** Constructs a node configuration from the default Bitcoin-S
     * data directory and given list of configuration overrides.
     */
-  def fromDefaultDatadir(confs: Config*)(
+  def fromDefaultDatadir(useLogbackConf: Boolean, confs: Config*)(
```
Christewart, line 78: We should really create an 

```scala

trait AppConfigFactory[T <: AppConfig] { 
  def fromDefaultDataDir(...)(...): T
}
```

testkit/src/main/resources/reference.conf
```diff
@@ -3,6 +3,9 @@ bitcoin-s {
     network = regtest # regtest, testnet3, mainnet
 
     logging {
+        # Ignore bitcoin-s logging config and use a logback config
+        logback = false
```
Christewart: So I think we should default this to `true` in testkit, because by definition if you are using `testkit` you are doing tests, which should be configured by a `logback-test.xml`

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
benthecarman: Agreed, I think that is better suited for a separate PR though

benthecarman: #1451

PR #1399, opened 2020-05-05
PSBT from Transaction

benthecarman:


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -770,6 +770,36 @@ object PSBT extends Factory[PSBT] {
...
+      BaseTransaction(tx.version, unsignedInputs, tx.outputs, tx.lockTime)
+
+    val globalMap = GlobalPSBTMap(
+      Vector(GlobalPSBTRecord.UnsignedTransaction(unsignedTx)))
+    val inputMaps = {
+      tx match {
+        case btx: BaseTransaction =>
+          btx.inputs.map(InputPSBTMap.fromTransactionInput).toVector
+        case wtx: WitnessTransaction =>
+          wtx.witness.witnesses.map(InputPSBTMap.fromWitness)
```
nkohen: Does this cover the `P2SH(P2WSH)` case correctly?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
 object InputPSBTMap extends PSBTMapFactory[InputPSBTRecord, InputPSBTMap] {
   import org.bitcoins.core.psbt.InputPSBTRecord._
 
+  def fromWitness(witness: ScriptWitness): InputPSBTMap = {
+    witness match {
+      case p2wpkh: P2WPKHWitnessV0 =>
+        val sig = PartialSignature(p2wpkh.pubKey, p2wpkh.signature)
```
nkohen: you need to check that there is a signature first, it is possible to have a p2wpkh witness that only has the pubkey ... sadly we don't have separate types for these two things

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+        InputPSBTMap(Vector(sig))
+      case p2wsh: P2WSHWitnessV0 =>
+        val sigs = p2wsh.pubKeys.zip(p2wsh.signatures).map {
```
nkohen: same as above, this may not have the signatures yet. either way, you also want to put the witness script record in

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
 object InputPSBTMap extends PSBTMapFactory[InputPSBTRecord, InputPSBTMap] {
   import org.bitcoins.core.psbt.InputPSBTRecord._
 
+  def fromWitness(witness: ScriptWitness): InputPSBTMap = {
```
nkohen, line 621: this (and others) doesn't seem to add the sighash to the PSBTInputMap

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+    }
+  }
+
+  @tailrec
+  def fromScriptSig(scriptSignature: ScriptSignature): InputPSBTMap = {
+    scriptSignature match {
+      case p2pkh: P2PKHScriptSignature =>
+        val sig = PartialSignature(p2pkh.publicKey, p2pkh.signature)
+        InputPSBTMap(Vector(sig))
+      case p2sh: P2SHScriptSignature =>
```
nkohen, line 661: don't we want the redeem script in this case?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+        val sigs = p2sh.publicKeys.zip(p2sh.signatures).map {
+          case (key, sig) =>
+            PartialSignature(key, sig)
+        }
+        InputPSBTMap(sigs.toVector)
+      case cond: ConditionalScriptSignature =>
+        fromScriptSig(cond.nestedScriptSig)
+      case _: CLTVScriptSignature | _: CSVScriptSignature |
```
nkohen: no partial signatures?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -621,6 +621,49 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+          _: MultiSignatureScriptSignature | _: NonStandardScriptSignature |
+          _: P2PKScriptSignature | TrivialTrueScriptSignature |
+          EmptyScriptSignature =>
+        InputPSBTMap.empty
+    }
+  }
+
+  /**  */
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -94,6 +99,13 @@ sealed abstract class P2WSHWitnessV0 extends ScriptWitnessV0 {
     RawScriptPubKey.fromBytes(cmpct.bytes ++ stack.head)
   }
 
+  lazy val pubKeys: Seq[ECPublicKey] = {
+    val pubKeys: Seq[ScriptToken] = redeemScript.asm
+      .filter(_.isInstanceOf[ScriptConstant])
+      .filterNot(_.isInstanceOf[ScriptNumberOperation])
+    pubKeys.map(k => ECPublicKey(k.hex))
+  }
```
nkohen: This seems unsafe, not all script constants are going to be public keys (hashes for example), you're going to want to check `ECPublicKey.isFullyValid` as well

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: It should, those qualify as a `WitnessTransaction`

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
benthecarman: I don't think we have a way to get the pubkeys for these types

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -94,6 +99,13 @@ sealed abstract class P2WSHWitnessV0 extends ScriptWitnessV0 {
     RawScriptPubKey.fromBytes(cmpct.bytes ++ stack.head)
   }
 
+  lazy val pubKeys: Seq[ECPublicKey] = {
+    val pubKeys: Seq[ScriptToken] = redeemScript.asm
+      .filter(_.isInstanceOf[ScriptConstant])
+      .filterNot(_.isInstanceOf[ScriptNumberOperation])
+    pubKeys.map(k => ECPublicKey(k.hex)).filter(_.isFullyValid)
```
Christewart, line 105: Why the `.filter(_.isFullyValid)` here? Is it to try and get rid of other things that may be a `ScriptConstant`, but not a `ECPublicKey`?

benthecarman: Yes

nkohen: yup

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
nkohen: I see :/ so we are able to recover public keys from signatures in ECDSA (using `CryptoUtil.recoverPublicKey`) but it will return two possible candidates, only one of which will be the correct key. Any way we can use this to not lose information here?

Also, for `LockTimeScriptSignature`s, you should be able to call `fromScriptSig` on the nested script sig right?

benthecarman: To use `CryptoUtil.recoverPublicKey` we need the message they are signing, for that we need the previous output :/

PR #1400, opened 2020-05-06
Input info Refactor

nkohen:
Replaced `UTXOSpendingInfo` with `InputInfo` and `UTXOInfo`.

TODO
- [x] Scaladocs
- [x] Doc changes


core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+  override val p2pkhPreImage: Option[ECPublicKey] = None
+
+  def redeemScript: ScriptPubKey
+
+  override val redeemScriptOpt: Option[ScriptPubKey] = Some(redeemScript)
+
+  override def scriptPubKey: P2SHScriptPubKey = P2SHScriptPubKey(redeemScript)
+
+  def nestedInputInfo: InputInfo
+}
```
benthecarman: Can we get some invariant so we make sure the hash of the redeem script matches the output

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+
+case class P2SHNoNestInputInfo(
```
benthecarman: Can this be called `P2SHNonSegwitInputInfo`, I feel that makes more sense

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+    scriptPubKey: WitnessScriptPubKey,
+    scriptWitness: ScriptWitness,
+    conditionalPath: ConditionalPath)
+    extends InputInfo {
+  override def p2pkhPreImage: Option[ECPublicKey] = None
+
+  override def redeemScriptOpt: Option[ScriptPubKey] = None
+
+  override def scriptWitnessOpt: Option[ScriptWitness] = Some(scriptWitness)
+}
```
benthecarman: Can this also get some invariant so we make sure the hash of the witness script matches the output

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+  P2WSHWitnessV0,
+  RawScriptPubKey,
+  ScriptPubKey,
+  ScriptWitness,
+  ScriptWitnessV0,
+  UnassignedWitnessScriptPubKey,
+  WitnessCommitment,
+  WitnessScriptPubKey,
+  WitnessScriptPubKeyV0
+}
```
benthecarman: nit: `org.bitcoins.core.protocol.script._`

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+
+  def output: TransactionOutput = {
+    TransactionOutput(amount, scriptPubKey)
+  }
+
+  def redeemScriptOpt: Option[ScriptPubKey]
+
+  def scriptWitnessOpt: Option[ScriptWitness]
+
+  def p2pkhPreImage: Option[ECPublicKey]
```
benthecarman: what's the reason for not naming just `pubkey`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -31,7 +38,9 @@ import scala.util.{Failure, Success, Try}
 /**
   * Created by chris on 12/26/15.
   */
-sealed abstract class ScriptPubKey extends Script
+sealed abstract class ScriptPubKey extends Script {
+  def pubKeysFor(inputInfo: InputInfo): Vector[ECPublicKey]
```
benthecarman: I think it'd make more sense to just have a `ScriptPubKey.pubKeys` and then put all the invariants inside of the `InputInfo`s

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -666,11 +673,11 @@ object InputPSBTMap extends PSBTMapFactory[InputPSBTRecord, InputPSBTMap] {
   }
 
   /** Constructs a full (ready to be finalized) but unfinalized InputPSBTMap
-    * from a UTXOSpendingInfoFull, the corresponding PSBT's unsigned transaction,
+    * from a NewSpendingInfoFull, the corresponding PSBT's unsigned transaction,
     * and if this is a non-witness spend, the transaction being spent
     */
-  def fromUTXOSpendingInfo(
-      spendingInfo: UTXOSpendingInfoFull,
+  def fromNewSpendingInfo(
```
benthecarman: I don't understand the `New` here, why not just `fromSpendingInfo`

benthecarman (changes requested):

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -0,0 +1,446 @@
...
+  WitnessScriptPubKeyV0
+}
+import org.bitcoins.core.protocol.transaction.{
+  TransactionOutPoint,
+  TransactionOutput
+}
+import org.bitcoins.core.script.crypto.HashType
+import org.bitcoins.crypto.{ECPublicKey, Sign}
+
+sealed trait InputInfo {
```
Christewart, line 20: So I think part of the reason we keep having design issues is this trait where we bundle in everything as Options. I think we need to break this into smaller pieces so that we can then make this a proper ADT without all `Options` (this many options in one trait usually indicates a design flaw). Here some pseudocode of what I was thinkikng: 

```scala
//this is needed for every utxo we spend, so we make this it's own case class
//and use a has-a relationship rather than a "is-a" relationship
case class Output(outPoint: TransactionOutPoint, output: TransactionOutput) {
  def scriptPubKey: ScriptPubKey = output.scriptPubKey

  def amount: CurrencyUnit = output.value
}

//here is where we start getting into ADTs
//for encoding things we need to "satisfy" the script
sealed trait InputInfo {
  def output: Output
}

case class BaseP2SHInputInfo(redeemScript: RawScriptPubKey,
                             override val output: Output) extends InputInfo

case class RawWitSPKInputInfo(witnessScript: WitnessScriptPubKey,
                              override val output: Output) extends InputInfo

case class P2PKHInputInfo(publicKey: ECPublicKey, override val output: Output) extends InputInfo

case class P2SHWitSPKInputInfo(redeemScript: WitnessScriptPubKey, scriptWitness: ScriptWitness, output: Output) extends InputInfo

//....
```

The unfortunately thing with this is `Output` doesn't give you information about the `ScriptPubKey` nested inside of it

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
Christewart: I think I agree with Ben, it's unfortunate but all `ScriptPubKey`s do not have a public key in the raw script itself. It's confusing to me that you use this method for both cases of 

1. The public key is emebedeed in the spk itself -- which means this doesn't use `inputInfo` at all.
2. The public key is not embedded in the spk itself, for that case we read it from `InputInfo`. 

FWIW I think the `publicKey` or `pubKeys` should only be available on scripts that have the public key embedded directly in the spk itself.

If you want to move this into a companion object like we did yesterday for `scriptWitnessOpt` and `redeemScriptOpt` I would be ok with that, but I don't think it should be in the interface here.

Christewart: You might even be able to create a cleaner interface to use if you move it into companion objects: https://github.com/Christewart/bitcoin-s-core/tree/2020-05-spk-refactor-suggestion

core/src/main/scala/org/bitcoins/core/wallet/utxo/UTXOInfo.scala
```diff
@@ -0,0 +1,80 @@
...
+          s"Cannot have signers that do not sign for one of $keysToSignFor")
+
+  def amount: CurrencyUnit = inputInfo.amount
+  def output: TransactionOutput = inputInfo.output
+  def outPoint: TransactionOutPoint = inputInfo.outPoint
+  def conditionalPath: ConditionalPath = inputInfo.conditionalPath
+}
+
+object UTXOInfo {
+  type Any = UTXOInfo[InputInfo]
```
Christewart: nit: I would prefer not to use type aliases unless the names are really long. These seem reasonable to me

core/src/main/scala/org/bitcoins/core/wallet/utxo/ConditionalPath.scala
```diff
@@ -0,0 +1,44 @@
...
+  *
+  * If you over-specify a path, such as giving a condition where none is needed,
+  * then the remaining over-specified path will be ignored.
+  *
+  * For example, if you wanted to spend a ConditionalScriptPubKey(P2PK1, P2PK2)
+  * (which looks like OP_IF <P2PK1> OP_ELSE <P2PK2> OP_ENDIF) with the P2PK1 case,
+  * then you would construct a ConditionalSpendingInfo using nonNestedTrue as your
+  * ConditionalPath. Otherwise if you wanted to use P2PK2 you would use nonNestedFalse.
+  */
+sealed trait ConditionalPath {
```
Christewart, line 13: I don't love the names for the subtypes of `ConditionalPath` so feel free to change, but 

```
sealed trait ConditionalPath

sealed abstract class Leaf extends ConditionalPath

sealed abstract class Node extends ConditionalPath {
  def head: Boolean
}

object ConditionalPath {
  case object NoConditionsLeft extends Leaf

  case class ConditionTrue(nextCondition: ConditionalPath)
      extends Node {
    override val head: Boolean = true
  }
  case class ConditionFalse(nextCondition: ConditionalPath)
      extends Node {
    override val head: Boolean = false
  }
```

is what this should probably look like

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
nkohen: I think in the future this will become the field for hash pre-images needed in general (e.g. HTLC pre-image)

Christewart (commented):
I think InputInfo and UTXOInfo should probably have their names switched:

1. A utxo is an output, which is why I think it would make sense to have that be the thing that contains the outpoint, and output. 
2. InputInfo to me indicates that this is the thing that is is spending a utxo, which is what should have signers , hashtype  (anything that is needed to produce an input)

Currently we have `UtxoInfo(InputInfo())` -- when creating an input you need to have reference a utxo to which is a "has-a" relationship -- a input info has a utxo info `InputInfo(UtxoInfo())`.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
Christewart: There needs to be a lengthy scaladoc on this and why it can't be pushed into subtypes like everything else that is specific to scripts (redeemScript, scriptWitness). It also needs to needs to be renamed probably: `p2pkhPreimage` doesn't make much sense for an HTLC preimage like you are talking about. Also, is this field used for `p2wpkh`? Also we should append the `Opt` suffix to indicate it is an `Option`.

core/src/main/scala/org/bitcoins/core/wallet/utxo/ConditionalPath.scala
```diff
@@ -0,0 +1,50 @@
...
+  def headOption: Option[Boolean]
```
Christewart, line 14: Let's move this into a companion object if you feel like this is necessary (similar to `scriptWitnessOpt`, `receemScriptOpt`, same argument, `NoCondition.headOption()` represents an illegal state).

nkohen: No illegal states here, even the standard library has `Nil.headOption`

benthecarman (approved):

PR #1404, opened 2020-05-08
2020 05 08 gui sb background color

Christewart:
Screenshot from 2020-05-08 12-57-00


benthecarman:
This doesn't need to be built on top of #1402

PR #1405, opened 2020-05-08
Wallet send from outpoints

benthecarman:
Useful for coin selection

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -4,7 +4,11 @@ import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.CliReaders._
 import org.bitcoins.core.config.NetworkParameters
 import org.bitcoins.core.currency._
-import org.bitcoins.core.protocol.transaction.{EmptyTransaction, Transaction}
+import org.bitcoins.core.protocol.transaction.{
+  EmptyTransaction,
+  Transaction,
+  TransactionOutPoint
+}
 import org.bitcoins.core.protocol.{BitcoinAddress, BlockStamp}
 import org.bitcoins.core.psbt.PSBT
 import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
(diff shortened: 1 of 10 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -505,6 +561,12 @@ object CliCommand {
       amount: Bitcoins,
       satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
       extends CliCommand
+  case class SendFromOutPoints(
+      outPoints: Vector[TransactionOutPoint],
+      destination: BitcoinAddress,
+      amount: Bitcoins,
+      satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
```
Christewart, line 568: Do we want this to be available on segwit only? Isn't `sats/vbyte` only available on segwit outpoints?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -205,6 +207,59 @@ abstract class Wallet
     } yield updatedInfos
   }
 
+  private def finishSend(txBuilder: BitcoinTxBuilder): Future[Transaction] = {
```
Christewart, line 213: scaladoc of what this method is doing

wallet-test/src/test/scala/org/bitcoins/wallet/WalletSendingTest.scala
```diff
@@ -140,4 +141,27 @@ class WalletSendingTest extends BitcoinSWalletTest {
       sendToAddressesF
     }
   }
+
+  it should "" in { fundedWallet =>
```
Christewart: name for the test case

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: No, on non-segwit txs it will just be the same as sats/byte

Christewart: Ok, what about mixed outpoints? I.e. non segwit and segwit utxos? We probably need to come up with a coherent model for this case -- it's hard to think about!

benthecarman: This is just handled by `FeeUnit.calc` should be handled correctly already

Christewart: Ok, so when doing our API, does fee unit of `sat/vbyte` or `sat/byte` matter? Can you pay a fee too large or too small of a fee depending on that unit, and how we interpret it?

Obviously if we were to interpret the number as `SatsPerKiloByte` will effect your fee rate, but the difference between `sats/vbyte` & `sats/byte` won't?

benthecarman: It's handled correctly, for `sats/vbyte` it does `tx.vsize * fee rate`, for `sats/byte` it does `tx.baseSize * fee rate`

Christewart (approved):

PR #1406, opened 2020-05-08
Add listSpentAddresses call

benthecarman:


wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
```diff
@@ -114,4 +114,21 @@ class AddressHandlingTest extends BitcoinSWalletTest {
       }
   }
 
+  it must "get the correct spent addresses" in { fundedWallet: FundedWallet =>
+    val wallet = fundedWallet.wallet
+
+    for {
```
Christewart, line 120: So if I understand this test case, this is checking that the output is _not_ listed as spent before we have called `.processTransaction()` right?

Do we have a test case that actually checks it is listed as spent when it shows up on the network?

benthecarman: No, `processTransaction` is called in `sendToAddress`, so this is testing that after we process a tx (from us or network) that it correctly lists the addresses as spent.

Christewart (approved):

PR #1408, opened 2020-05-08
Add list unused addresses call

benthecarman:
Addresses that have never ever received funds


Christewart (changes requested):
I think we are missing a unit test in the wallet project for this

benthecarman:
> I think we are missing a unit test in the wallet project for this

https://github.com/bitcoin-s/bitcoin-s/pull/1408/commits/90bc407766d3a50bb099b1796b138b8afee2a043#diff-bffb479aef0517e77121a1e849f44973R117

Christewart (approved):

PR #1409, opened 2020-05-09
Wallet Send with coin selection algorithim

benthecarman:


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -505,6 +559,12 @@ object CliCommand {
       amount: Bitcoins,
       satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
       extends CliCommand
+  case class SendWithAlgo(
+      destination: BitcoinAddress,
+      amount: Bitcoins,
+      satoshisPerVirtualByte: Option[SatoshisPerVirtualByte],
```
nkohen: nit: Should probably be called `feeOpt` right?

nkohen (approved):
Just one nit

PR #1413, opened 2020-05-11
2020 05 11 bech32 addr tostring

Christewart:
Make `Bech32Address.toString` give us `bcrt1qq6w6pu6zq90az9krn53zlkvgyzkyeglzukyepf` rather than `Bech32Address(bcrt1qq6w6pu6zq90az9krn53zlkvgyzkyeglzukyepf)`


benthecarman (approved):
Does this need to be fixed for the other address types?

Christewart:
> Does this need to be fixed for the other address types?

Nope, for some reason it was override on the subtype.

PR #1417, opened 2020-05-12
Add ability to make OP_RETURN commitments

benthecarman:
The OP_RETURN commitment should probably be given a proper `ScriptPubKey` type, but was just hacking this together for funzies.

Proof it works as well: 
https://blockstream.info/testnet/tx/7080d81c2943167380337689664a325ded8ae4ff62c377f18da976fb3516e421?expand

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -348,7 +348,7 @@ object ConsoleCli {
             .action((feeRate, conf) =>
               conf.copy(command = conf.command match {
                 case opReturnCommit: OpReturnCommit =>
-                  opReturnCommit.copy(satoshisPerVirtualByte = Some(feeRate))
+                  opReturnCommit.copy(feeRateOpt = Some(feeRate))
                 case other => other
               }))
         ),
@@ -677,7 +677,7 @@ object CliCommand {
   case class OpReturnCommit(
       message: String,
       hashMessage: Boolean,
-      satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
+      feeRateOpt: Option[SatoshisPerVirtualByte])
       extends CliCommand
   case object GetNewAddress extends CliCommand
   case object GetUtxos extends CliCommand
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -319,6 +319,39 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("opreturncommit")
+        .action((_, conf) =>
+          conf.copy(command = OpReturnCommit("", hashMessage = false, None)))
+        .text("Send money to the given address")
+        .children(
+          arg[String]("message")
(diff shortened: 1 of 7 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -505,6 +543,11 @@ object CliCommand {
       amount: Bitcoins,
       satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
       extends CliCommand
+  case class OpReturnCommit(
+      message: String,
+      hashMessage: Boolean,
+      satoshisPerVirtualByte: Option[SatoshisPerVirtualByte])
```
nkohen: nit: should probably be `feeOpt`

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -477,6 +477,29 @@ class RoutesSpec
 
     }
 
+    "make an OP_RETURN commitment" in {
+
+      val message = "Never gonna give you up, never gonna let you down"
```
nkohen, line 708: nice :)

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -262,6 +266,31 @@ abstract class Wallet
...
+      CryptoUtil.sha256(ByteVector(message.getBytes)).bytes
+    } else {
+      if (message.length > 32) {
+        throw new IllegalArgumentException(
+          s"Message cannot be greater than 32 characters, it should be hashed, got $message")
+      } else ByteVector(message.getBytes)
+    }
+
+    val asm = Seq(OP_RETURN,
+                  BytesToPushOntoStack(messageToUse.size),
```
nkohen: this won't work for small things (think 1 bit), you should use the calculate push op method in script util

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -262,6 +266,31 @@ abstract class Wallet
...
+      message: String,
+      hashMessage: Boolean,
+      feeRate: FeeUnit,
+      fromAccount: AccountDb): Future[Transaction] = {
+    val messageToUse = if (hashMessage) {
+      CryptoUtil.sha256(ByteVector(message.getBytes)).bytes
+    } else {
+      if (message.length > 32) {
+        throw new IllegalArgumentException(
+          s"Message cannot be greater than 32 characters, it should be hashed, got $message")
```
nkohen: Are you not allowed to push larger messages?

benthecarman: I believe 40 bytes is the max, figured 32 bytes is fine for this so we have a clear distinction between hashes.

If you want to push bigger messages checkout BSV :P

benthecarman: Actually looks like it's 83 bytes

https://github.com/bitcoin/bitcoin/blob/master/src/script/standard.h#L33

benthecarman: I agree then, this should be changed

PR #1426, opened 2020-05-14
TxBuilder Refactor

nkohen:
Built on top of #1400 

Replaces monolithic and hard to adapt `TxBuilder` with three layers:
1) `RawTxBuilder` - Responsible for collecting `Vector`s of inputs and outputs
2) `RawTxFinalizer` - Responsible for taking completed lists of inputs and outputs and constructing unsigned transactions (fee considerations and change happens here)
3) `RawTxSigner` - Responsible for signing finalized transactions

TODO: Move TxBuilder tests into other files


core-test/src/test/scala/org/bitcoins/core/wallet/builder/TxBuilderTest.scala
```diff
@@ -1,41 +1,31 @@
...
-    TxBuilder
+    Transaction
       .isValidFeeRange(estimatedFee, actualFee, feeRate)
       .isFailure must be(true)
 
-    TxBuilder
+    Transaction
       .isValidFeeRange(actualFee, estimatedFee, feeRate)
       .isFailure must be(true)
   }
```
benthecarman: Should this test be moved to `TransactionTest` now?

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
 }
 
-sealed abstract class BaseTransaction extends Transaction {
-  override def bytes = RawBaseTransactionParser.write(this)
-  override def weight = byteSize * 4
+object Transaction extends Factory[Transaction] {
+  override def fromBytes(bytes: ByteVector): Transaction = {
+    Try(WitnessTransaction(bytes)) match {
+      case Success(wtx) => wtx
+      case Failure(_)   => BaseTransaction(bytes)
```
benthecarman: Should replace this with what @Christewart did in #1431

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
+      sanityAmountChecks(forSigned, inputInfos, expectedFeeRate, tx)
+    }
+  }
+
+  def sanityDustCheck(tx: Transaction): Try[Unit] = {
+    if (tx.outputs
+          .filterNot(_.scriptPubKey.asm.contains(OP_RETURN))
+          .map(_.value)
+          .exists(_ < Policy.dustThreshold)) {
+      TxBuilderError.OutputBelowDustThreshold
```
benthecarman: Would be nice if it had the the output(s) that are below the dust limit

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -156,10 +367,32 @@ sealed abstract class WitnessTransaction extends Transaction {
...
+    * [[https://github.com/bitcoin/bitcoin/blob/e8cfe1ee2d01c493b758a67ad14707dca15792ea/src/primitives/transaction.h#L282-L287s]]
+    */
+  override def bytes: ByteVector = {
+    val versionBytes = version.bytes.reverse
+    val inputBytes = BytesUtil.writeCmpctSizeUInt(inputs)
+    val outputBytes = BytesUtil.writeCmpctSizeUInt(outputs)
+    val witnessBytes = witness.bytes
+    val lockTimeBytes = lockTime.bytes.reverse
+    //notice we use the old serialization format if all witnesses are empty
+    // https://github.com/bitcoin/bitcoin/blob/e8cfe1ee2d01c493b758a67ad14707dca15792ea/src/primitives/transaction.h#L276-L281
```
benthecarman: nit: spacing

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
+package org.bitcoins.core.wallet.builder
+
+import org.bitcoins.core.number.{Int32, UInt32}
+import org.bitcoins.core.protocol.script.{
```
benthecarman: nit: Press Ctrl + Alt + O

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
benthecarman: This is a current problem, but it'd nice to be fixed here. Currently, when we do 
```
Transaction.fromBytes(EmptyTransaction.bytes)
```
It will return a `BaseTransaction` not an `EmptyTransaction`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+    TransactionInput,
+    Vector[TransactionInput]] = Vector.newBuilder
+
+  private val outputsBuilder: mutable.ReusableBuilder[
+    TransactionOutput,
+    Vector[TransactionOutput]] = Vector.newBuilder
+
+  private var lockTime: UInt32 = TransactionConstants.lockTime
+
+  private var finalizerOpt: Option[RawTxFinalizer] = None
```
benthecarman: Are these being mutable going to cause problems?

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+    */
+  def calcLockTime(utxos: Seq[InputSigningInfo[InputInfo]]): Try[UInt32] = {
+    def computeNextLockTime(
+        currentLockTimeOpt: Option[UInt32],
+        locktime: Long): Try[UInt32] = {
+      val lockTime =
+        if (locktime > UInt32.max.toLong || locktime < 0) {
+          TxBuilderError.IncompatibleLockTimes
+        } else Success(UInt32(locktime))
+      lockTime.flatMap { l: UInt32 =>
```
benthecarman: nit: better name than `l`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -0,0 +1,169 @@
...
+      if (finalizedTx.outputs.size == expectedOutputs.size) {
+        false
+      } else {
+        //the extra output should be the changeOutput
+        !(finalizedTx.outputs.size == (expectedOutputs.size + 1) &&
+          finalizedTx.outputs.map(_.scriptPubKey).contains(changeSPK))
+      }
+    val spendingTxOutPoints = finalizedTx.inputs.map(_.previousOutput)
+    val hasExtraOutPoints =
+      !spendingTxOutPoints.forall(expectedOutPoints.contains)
```
benthecarman: Doesn't handle the case where `expectedOutPoints.size > spendingTxOutPoints.size`

Should just need to add a `&& expectedOutPoints.size == spendingTxOutPoints.size`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  def txBuilderFrom(
+      outputs: Seq[TransactionOutput],
+      utxos: Seq[InputSigningInfo[InputInfo]],
+      feeRate: FeeUnit,
+      changeSPK: ScriptPubKey): RawTxBuilder = {
+    val inputs =
+      InputSigningInfo.calcSequenceForInputs(utxos, Policy.isRBFEnabled)
```
benthecarman: I don't think RBF should be enabled by default. It is not good for privacy can cause double spend issues for txs before they are confirmed

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -0,0 +1,169 @@
...
+import org.bitcoins.core.policy.Policy
+import org.bitcoins.core.protocol.script.ScriptPubKey
+import org.bitcoins.core.protocol.transaction._
+import org.bitcoins.core.wallet.fee.FeeUnit
+import org.bitcoins.core.wallet.utxo.{InputInfo, InputSigningInfo}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Success, Try}
+
+sealed trait RawTxFinalizer {
```
benthecarman, line 32: It would be nice if we could nest finalizers, so we can use many at once without needing a type for every combination.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -419,7 +447,9 @@ case class UnassignedSegwitNativeInputInfo(
     scriptWitness: ScriptWitness,
     conditionalPath: ConditionalPath,
     pubKeys: Vector[ECPublicKey])
-    extends InputInfo
+    extends InputInfo {
+  override def requiredSigs: Int = pubKeys.length
```
benthecarman, line 451: I don't know the correct solution to this but I think this could cause problems

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
```diff
@@ -29,6 +41,102 @@ sealed trait InputSigningInfo[+InputType <: InputInfo] {
...
+    * A transaction needs a valid sequence number to spend a OP_CHECKSEQUENCEVERIFY script.
+    * See BIP68/112 for more information
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki]]
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki]]
+    */
+  private def solveSequenceForCSV(scriptNum: ScriptNumber): UInt32 =
+    if (LockTimeInterpreter.isCSVLockByBlockHeight(scriptNum)) {
+      val blocksPassed = scriptNum.toLong & TransactionConstants.sequenceLockTimeMask.toLong
+      val sequence = UInt32(blocksPassed)
+      sequence
```
benthecarman: nit: don't need to assign to a `val`

benthecarman (commented):
Looks really good!

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+class RawTxBuilder {
+  private var version: Int32 = TransactionConstants.validLockVersion
+
+  private val inputsBuilder: mutable.Builder[
+    TransactionInput,
+    Vector[TransactionInput]] = Vector.newBuilder
+
+  private val outputsBuilder: mutable.Builder[
+    TransactionOutput,
+    Vector[TransactionOutput]] = Vector.newBuilder
```
benthecarman, line 49: I think the tx builder should have some distinction from change outputs and outputs we are sending to. This will make things simpler for reasoning things like changing the fee rate of the tx or combining 2 tx builders, etc

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -77,6 +77,19 @@ object TxSigComponent {
         BaseTxSigComponent(transaction, inputIndex, output, flags)
     }
   }
+
+  def getScriptWitness(
+      txSigComponent: TxSigComponent): Option[ScriptWitness] = {
```
Christewart: Probably should be a return type of `Option[ScriptWitnessV0]` as you are treating `EmptyScriptWitness` as `None`.

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
Christewart: Let's leave this for now and just rebase onto #1431 when that gets merged

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
-  override def weight = byteSize * 4
+object Transaction extends Factory[Transaction] {
+  override def fromBytes(bytes: ByteVector): Transaction = {
+    Try(WitnessTransaction(bytes)) match {
+      case Success(wtx) => wtx
+      case Failure(_)   => BaseTransaction(bytes)
+    }
+  }
+
+  def addDummySigs(utx: Transaction, inputInfos: Vector[InputInfo])(
```
Christewart: scaladoc, something that would be nice to add to the doc is if it adds signatures to both witness and non-witness transactions.

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
+                             utx.outputs,
+                             utx.lockTime,
+                             wit)
+      }
+    }
+  }
+
+  /** Runs various sanity checks on a transaction */
+  def sanityChecks(
+      forSigned: Boolean,
```
Christewart: maybe `isSigned`? I think that's what you mean here?

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
+        Await.result(addDummySigs(tx, inputInfos), 5.seconds)
+      }
+
+      val actualFee = creditingAmount - spentAmount
+      val estimatedFee = expectedFeeRate * expectedTx
+      val feeResult =
+        isValidFeeRange(estimatedFee, actualFee, expectedFeeRate)
+
+      if (!forSigned && feeResult.isFailure) {
+        println(s"Before: $tx")
```
Christewart: ping

core/src/main/scala/org/bitcoins/core/wallet/builder/FinalizedTxWithSigningInfo.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def sign(expectedFeeRate: FeeUnit)(
+      implicit ec: ExecutionContext): Future[Transaction] = {
+    RawTxSigner.sign(this, expectedFeeRate)
+  }
+
+  def sign(
+      expectedFeeRate: FeeUnit,
+      invariants: (
+          Seq[ScriptSignatureParams[InputInfo]],
```
Christewart: `Vector` rather than `Seq` (I'm guessing I introduced this originally?)

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  def setLockTime(lockTime: UInt32): this.type = {
+    this.lockTime = lockTime
+    this
+  }
+}
+
+object RawTxBuilder {
+
```
Christewart, line 147: Let's add a `newBuilder` method similar to `Vector.newBuilder`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+  def apply(): RawTxBuilder = new RawTxBuilder()
+
+  /**
+    * Sets the ScriptSignature for every input in the given transaction to an EmptyScriptSignature
+    * as well as sets the witness to an EmptyWitness
+    * @param tx Transaction to empty signatures
+    * @return Transaction with no signatures
+    */
+  def emptyAllScriptSigs(tx: Transaction): Transaction = {
```
Christewart: Move to `TxUtil` object

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
```diff
@@ -29,6 +41,102 @@ sealed trait InputSigningInfo[+InputType <: InputInfo] {
...
+      sequence | TransactionConstants.sequenceLockTimeTypeFlag
+    }
+
+  /**
+    * This helper function calculates the appropriate sequence number for each transaction input.
+    * [[CLTVScriptPubKey]] and [[CSVScriptPubKey]]'s need certain sequence numbers on the inputs
+    * to make them spendable.
+    * See BIP68/112 and BIP65 for more info
+    */
+  def calcSequenceForInputs(
```
Christewart: Maybe make a file called `InputUtil` and move it in there?

core/src/main/scala/org/bitcoins/core/util/BytesUtil.scala
```diff
@@ -0,0 +1,53 @@
+package org.bitcoins.core.util
+
+import org.bitcoins.core.number.UInt64
+import org.bitcoins.core.protocol.CompactSizeUInt
+import org.bitcoins.crypto.{CryptoBytesUtil, Factory, NetworkElement}
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+trait BytesUtil extends CryptoBytesUtil {
```
Christewart, line 10: nit: Just do `has-a` instead of `is-a` with `CryptoBytesUtil`

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -108,40 +118,241 @@ sealed abstract class Transaction extends NetworkElement {
...
+                             inputs,
+                             utx.outputs,
+                             utx.lockTime,
+                             wit)
+      }
+    }
+  }
+
+  /** Runs various sanity checks on a transaction */
+  def sanityChecks(
```
Christewart: `TxUtil` file

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
Christewart: Maybe move to a `TxUtil` file?

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  def clear(): Unit = {
+    version = TransactionConstants.validLockVersion
+    inputsBuilder.clear()
+    outputsBuilder.clear()
+    lockTime = TransactionConstants.lockTime
+  }
+
+  def addInput(input: TransactionInput): this.type = {
```
Christewart, line 77: so are there any requirements onthese `TransactionInput`? For instance, do they need to have `EmptyScriptSignature` be the scriptSig? Can the ScriptSignature have the redeemScript in the case case of p2sh?

Christewart: Depending on the answer above, we don't seem to have the ability to add any `ScriptWitness` to the transaction. Should we have a `+=(tuple: (TransactionInput,ScriptWitness)) method?

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
```diff
@@ -0,0 +1,263 @@
...
+    finalizerOpt match {
+      case None =>
+        Future.failed(
+          new RuntimeException(
+            "Cannot call result with no RawTxFinalizer specified"))
+      case Some(finalizer) => result(finalizer)
+    }
+  }
+
+  def result(finalizer: RawTxFinalizer)(
```
Christewart: I think this method should just be `def result(): BaseTransaction` and then `RawTxFinalizer` can just take a `BaseTransaction` as input. This keeps the `Future` out of this API as well.

Christewart (commented):
Overall very nice, I think we should make `RawTxFinalizer` depend on `RawTxBuilder` rather than having `RawTxBuilder` take in a finalizer. This keeps the `Future`s out of `RawTxBuilder`. It would be nice to have scaladocs on what are expected out of `TransactionInput` in `RawTxBuilder` as well: Do they have to have a `EmptyScriptSignature`, do the `ScriptSignature`s redeemScripts need to be in them? What about `scriptWitness` associated with the `TransactionInput` in the case of segwit?

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: I'm open to the argument but my goal in this PR is to replace existing functionality, so this likely belongs on a different PR

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
nkohen: Currently used only for fee estimation so better over than under, in the future this could change but generally speaking we only kind of support UnassignedSegwit anyway...

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxBuilder.scala
nkohen: Will add docs once things are finalized but the outputs here are all non-change so this should work out +1

core/src/main/scala/org/bitcoins/core/util/BytesUtil.scala
nkohen: I personally think that for `Util` files it makes sense to use `is-a` as everything is static functions anyway. Also the goal in doing it this way is to allow users to call all the `BytesUtil` methods that the `crypto` project needed as if they were still defined in the `core` project (like before). I.e. this un-breaks our API.

PR #1431, opened 2020-05-15
Remove the clause in Transction.fromHex() where we throw in the case …

Christewart:
…of a witness transaction, do something smarter and look at bytes in the byte vector

https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-id

~Previously we would throw inside of `RawWitnessTransactionParser` if it was not a valid witness tx, which can be expensive. Now inside of `Transaction.fromBytes()` we look at the bytes at index 4 and 5 to see if it is a witness tx~

Unfortunately we cannot entirely get rid of the try/catch, we can just reduce the number of times it happens by adding a preliminary for bytes at index 4 and 5. This is because unsigned base transactions with 0 inputs and 1 output can "appear" as a segwit tx. See the [`TransactionTest`](https://github.com/bitcoin-s/bitcoin-s/blob/01d89df1b7c6bc4b1594406d54d5e6019705c654/core-test/src/test/scala/org/bitcoins/core/protocol/transaction/TransactionTest.scala#L88) test case added in this PR. If you want to fiddle with it yourself, you can try removing the check and run the `bitcoindRpcTest/test` cases and see the failures.

See: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-id


benthecarman:
Looks like this is causing some CI failures

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -175,14 +164,23 @@ sealed abstract class WitnessTransaction extends Transaction {
...
+  override def fromBytes(bytes: ByteVector): Transaction = {
+    //see BIP141 for marker/flag bytes
+    //https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-id
+    if (bytes(4) == WitnessTransaction.marker && bytes(5) == WitnessTransaction.flag) {
+      //this throw/catch is _still_ necessary for the case where we have unsigned base transactions
+      //with zero inputs and 1 output which is serialized as "0001" at bytes 4 and 5.
+      //these transactions will not have a script witness associated with them making them invalid
+      //witness transactions (you need to have a witness to be considered a witness tx)
+      //see: https://github.com/bitcoin-s/bitcoin-s/blob/01d89df1b7c6bc4b1594406d54d5e6019705c654/core-test/src/test/scala/org/bitcoins/core/protocol/transaction/TransactionTest.scala#L88
+      try {
```
nkohen, line 176: as this try stuff only applies on base txs with no inputs and one output, can't we explicitly check 8 bytes ahead for a cmpct size uint (8 bytes ahead because we want to skip the satoshi value in the output) and then check to see if the size of the tx is `5 + currency_unit_size + cmpct_uint_size + cmpct_uint_result + lock_time_size` and if it is then interpret it as a base transaction, otherwise a witness transaction: no `try` necessary?

nkohen (commented):
Looks good, but I do think we can get rid of the `try` entirely without too much of a headache

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
Christewart: https://github.com/bitcoin-s/bitcoin-s/issues/1442#issue-620451122

PR #1434, opened 2020-05-16
Try and improve reliability of UpdateBloomFilterTest, also add getMem…

Christewart:
…PoolEntryOpt() that returns an Option indicating if a tx was in the mempool

An attempt to fix this test case 

Screenshot from 2020-05-16 10-25-54

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
@@ -89,7 +89,7 @@ class MempoolRpcTest extends BitcoindRpcTest {
     } yield succeed
   }
 
-  it must "failed to find a mempool entry" in {
+  it must "fail to find a mempool entry" in {
     val resultF = for {
       (client, _) <- clientsF
       txid = DoubleSha256Digest.empty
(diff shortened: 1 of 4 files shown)


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
```diff
@@ -87,6 +89,28 @@ class MempoolRpcTest extends BitcoindRpcTest {
     } yield succeed
   }
 
+  it must "failed to find a mempool entry" in {
```
benthecarman: nit: failed -> fail

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MempoolRpcTest.scala
```diff
@@ -87,6 +89,28 @@ class MempoolRpcTest extends BitcoindRpcTest {
...
+      txid = DoubleSha256Digest.empty
+      result <- client.getMemPoolEntry(txid)
+    } yield {
+      result
+    }
+
+    recoverToSucceededIf[BitcoindException](resultF)
+  }
+
+  it must "failed to find a mempool entry and return None" in {
```
benthecarman: nit: failed -> fail

benthecarman (approved):

PR #1435, opened 2020-05-17
2020 05 17 optimizations

Christewart:
Optimize `ScriptOperations` which was using `def` for pre-defined script opts, and use `Vector[]` as the collection that contains script operations rather than `Seq`. `Seq` can sometimes be implemented as a `List` which has some O(n) behavior

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
@@ -125,7 +125,7 @@ object P2PKHScriptSignature extends ScriptFactory[P2PKHScriptSignature] {
              _: ScriptConstant,
              _: BytesToPushOntoStack,
              z: ScriptConstant) =>
-      if (ECPublicKey.isFullyValid(z.bytes)) true
+      if ((z.bytes.length == 33 || z.bytes.length == 65) && ECPublicKey.isFullyValid(z.bytes)) true
       else !P2SHScriptSignature.isRedeemScript(z)
     case _ => false
   }
--- a/core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
@@ -53,16 +53,9 @@ trait ScriptOperationFactory[T <: ScriptOperation] extends BitcoinSLogger {
 
   /** Finds a [[org.bitcoins.core.script.ScriptOperation ScriptOperation]] from a given [[scala.Byte Byte]]. */
   @inline final def fromByte(byte: Byte): T = {
-    var idx = 0
-    while (idx < operations.length) {
-      val op = operations(idx)
-      if (op.toByte == byte) {
-        return op
-      } else {
-        idx += 1
-      }
-    }
-    sys.error(s"Could not find opcode for byte=${byte}")
+    ScriptOperation.map.get(byte).getOrElse {
+      sys.error(s"Could not find opcode for byte=${byte}")
+    }.asInstanceOf[T]
   }
 
   def fromBytes(bytes: ByteVector): Option[T] = {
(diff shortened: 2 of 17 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -125,7 +125,7 @@ object P2PKHScriptSignature extends ScriptFactory[P2PKHScriptSignature] {
              _: ScriptConstant,
              _: BytesToPushOntoStack,
              z: ScriptConstant) =>
-      if (ECPublicKey.isFullyValid(z.bytes)) true
+      if ((z.bytes.length == 33 || z.bytes.length == 65) && ECPublicKey.isFullyValid(z.bytes)) true
```
benthecarman, line 128: Are there any old p2pkh that have pubkeys that are different sizes?

core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
```diff
@@ -52,8 +52,17 @@ trait ScriptOperationFactory[T <: ScriptOperation] extends BitcoinSLogger {
...
+    var idx = 0
+    while (idx < operations.length) {
+      val op = operations(idx)
+      if (op.toByte == byte) {
+        return op
+      } else {
+        idx += 1
+      }
+    }
+    sys.error(s"Could not find opcode for byte=${byte}")
```
benthecarman: I think a better way to optimize this would be to have a `Map[Byte, ScriptOperation]` so then we can just do `map.get(byte)`

core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
```diff
@@ -72,9 +81,19 @@ trait ScriptOperationFactory[T <: ScriptOperation] extends BitcoinSLogger {
...
+   * There is an optimization here by moving popular opcodes
+   * to the front of the vector so when we iterate through it,
+   * we are more likely to find the op code we are looking for
+   * sooner */
+  final override val operations: Vector[ScriptOperation] = {
+    Vector(OP_DUP, OP_HASH160, OP_EQUALVERIFY, OP_CHECKSIG) ++ //p2pkh
+    Vector(BytesToPushOntoStack.push20Bytes, BytesToPushOntoStack.push33Bytes,
+      BytesToPushOntoStack.push32Bytes) ++ //popular push op codes
+    Vector(OP_HASH160, OP_EQUAL) ++ //p2sh
+    Vector(OP_0, OP_1, OP_CHECKMULTISIG) //multisig
```
benthecarman: need a `++` at the end of this line

core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
```diff
@@ -72,9 +81,19 @@ trait ScriptOperationFactory[T <: ScriptOperation] extends BitcoinSLogger {
...
     ScriptNumberOperation.operations ++
-      Seq(OP_FALSE, OP_PUSHDATA1, OP_PUSHDATA2, OP_PUSHDATA4, OP_TRUE) ++
+      Vector(OP_FALSE, OP_PUSHDATA1, OP_PUSHDATA2, OP_PUSHDATA4, OP_TRUE) ++
```
benthecarman, line 84: `OP_FALSE` and `OP_TRUE` are the same as `OP_1` and `OP_0`, no? those are already included in the list

core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
Christewart: so this is actually an interesting question, this has the overhead of a hash function rather than just iterating through a Vector which modern CPUs are really good at doing. I would suspect the `map.get()` operation is slower although I have no hard data for this.

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
nkohen: Could you file an issue with your benchmark/hot paths findings for block parsing @Christewart ?

nkohen (approved):
Looks good

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
Christewart: #1443

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/script/ScriptOperationFactory.scala
Christewart: It's pretty clear from profiling locally the Map solution is superior
Screenshot from 2020-05-18 15-21-52

PR #1436, opened 2020-05-17
Add hardcoded jvmopts file that starts with a 2g heap

Christewart:
We have had builds failing with test case failure on `BlockTest` 

Screenshot from 2020-05-17 16-24-52

Without an obvious regression as far as I can tell


benthecarman (approved):

nkohen:
I don't have anything against this if we need it in the future but I did some profiling of sbt while compiling the project and it looks like we never even get close to 750 MB even with relatively low GC so this seems like it might be a little over-kill

PR #1446, opened 2020-05-19
Implemented (lax) signature parsing in Bitcoin-S

nkohen:
This allows us to finally update bouncy castle!

The current version is literally just me mimicking  (`var`s, `Array`s and all) the code here line for line: https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.cpp#L16-L165

I will attempt to clean this up and scala-fy it tomorrow, though I've already attempted it once and tbh I don't fully understand the business with the first bit of length bytes (I know it has to do with leading zeroes appended to r and s values, but need to get a better grasp on this), which makes it quite hard to re-write.

That said, at the time I did not actually have a working integration and now that I do it should be easier to write a new function that (should do) does the same thing and compare.

I still haven't been able to access the Gemini fork and see what is going on over there but hopefully this can help on their end!

EDIT:
I have now implemented it in a clean and readable way that is still efficient using iterators!

### DIFF
--- a/crypto/src/main/scala/org/bitcoins/crypto/DERSignatureUtil.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/DERSignatureUtil.scala
@@ -60,7 +60,7 @@ sealed abstract class DERSignatureUtil {
     * throws an exception if the given sequence of bytes is not a DER encoded signature
     */
   def decodeSignature(bytes: ByteVector): (BigInt, BigInt) = {
-    DERSignatureUtil.parseDERLax2(bytes) match {
+    DERSignatureUtil.parseDERLax(bytes) match {
       case Some((r, s)) => (r, s)
       case None         => (0, 0)
     }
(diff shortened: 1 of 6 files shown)


crypto/src/main/scala/org/bitcoins/crypto/DERSignatureUtil.scala
```diff
@@ -235,6 +205,108 @@ sealed abstract class DERSignatureUtil {
     require(DERSignatureUtil.isLowS(sigLowS))
     sigLowS
   }
+
+  /** Scala implementation of https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.cpp#L16-L165
```
benthecarman, line 209: nit: instead of using master, use the current HEAD commit so if there are changes it doesn't make this comment outdated

crypto/src/main/scala/org/bitcoins/crypto/DERSignatureUtil.scala
```diff
@@ -235,6 +205,108 @@ sealed abstract class DERSignatureUtil {
...
+      } else None
+    }
+
+    def nextByteMustBe(requiredByte: Byte): Option[Unit] = {
+      nextOption().flatMap { nextByte =>
+        if (nextByte == requiredByte) {
+          Some(())
+        } else None
+      }
+    }
```
benthecarman, line 239: Why not have this return a boolean

nkohen: So that it returns (breaks out of the `for`) below

Christewart (commented):
Is there any test cases from bitcoin core that be ported off to specifically test this stuff? It would be nice to have a few sanity unit tests at least that can directly test this

PR #1448, opened 2020-05-19
Fix allowing BytesToPushOntoStack(0)

### LINKED ISSUE #1445: BytesPushOntoStack.zero is not a valid bytes to push onto stack
This was found on #1435 

We currently construct a `BytesToPushOntoStack(0)` which is not a valid `BytesToPushOntoStack`:

https://github.com/bitcoin-s/bitcoin-s/blob/9a7b90c13283fbce34662838ea48282a7c662bc8/core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala#L29

You see the valid ranges for `BytesToPushOntoStack` here: https://en.bitcoin.it/wiki/Script

>N/A | 1-75 | 0x01-0x4b | (special) | data | The next opcode bytes is data to be pushed onto the stack
-- | -- | -- | -- | -- | --

The key thing here is that the opcode range is `1-75` rather than `0-75`. I would have fixed this on #1435 but it seems that [`BitcoinScriptUtil.calcPushOp()`](https://github.com/bitcoin-s/bitcoin-s/blob/9a7b90c13283fbce34662838ea48282a7c662bc8/core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala#L230) depends on this quiet heavily. 

If we change the range too 

```scala
  override val operations: Vector[BytesToPushOntoStack] = {
    (for { i <- 1 to 75 } yield BytesToPushOntoStackImpl(i)).toVector
  }
```

We get a bunch of errors. This was found in #1435 because we key the `scriptOpMap` by `byte` which gives us duplicate keys for the `opcode=0`

1. `OP_0` 
2. `BytesToPushOntoStack(0)`

The work around is moving `ScriptNumberOperation` to the end of the `ScriptOperationFactory.operations` vector, which seems to put `OP_0` rather than `BytesToPushOntoStack()` in the `scriptOpMap`

benthecarman:
Fixes #1445 

As described in #1445, `BytesToPushOnStack(0)` should not be valid but we were pushing `OP_0` when this occurred. This fixes that problem. However, after fixing that it caused some of our tests to fail, namely anywhere that we were putting an `EmptyDigitalSignature` into a script signature. This was a problem because when we check if it is valid script sig, we check that it has the correct `BytesToPushOnStack`, however, now when pushing `BytesToPushOnStack(0)` we no longer add anything to the stack. Luckily, this was easily remedied by replacing these instances with an `EmptyScriptSignature`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
@@ -20,6 +20,7 @@ object BytesToPushOntoStack
     /*  //see the 'Constants; section in https://en.bitcoin.it/wiki/Script
       require(num >= -1 && num <= 75, "A valid script number is between 1 and 75, the number passed in was: " + num)*/
     require(num >= 1, "BytesToPushOntoStackImpl cannot be less than 1")
+    require(num <= 75, "BytesToPushOntoStackImpl cannot be greater than 75")
     override val opCode = num
   }
 
--- a/core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
@@ -11,12 +11,6 @@ trait BytesToPushOntoStack extends ScriptOperation
 object BytesToPushOntoStack
     extends ScriptOperationFactory[BytesToPushOntoStack] {
 
-  /**
-    * Represents that zero bytes need to be pushed onto the stack
-    * this really means we need to push an empty byte vector on the stack
-    */
-  lazy val zero: BytesToPushOntoStack = apply(0)
-
   lazy val push33Bytes = operations(33)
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/bitcoins/core/script/constant/BytesToPushOntoStack.scala
```diff
@@ -25,12 +19,12 @@ object BytesToPushOntoStack
       extends BytesToPushOntoStack {
     /*  //see the 'Constants; section in https://en.bitcoin.it/wiki/Script
       require(num >= -1 && num <= 75, "A valid script number is between 1 and 75, the number passed in was: " + num)*/
-    require(num >= 0, "BytesToPushOntoStackImpl cannot be negative")
+    require(num >= 1, "BytesToPushOntoStackImpl cannot be less than 1")
```
Christewart, line 22: nit: probably should add `<=75` too

Christewart (approved):

nkohen (approved):

PR #1449, opened 2020-05-19
2020 05 19 improve test performance

Christewart:
Two more test case nuisances that are causing CI builds to fail: 

1. `NetworkPayloadTest` -- I allow the property based tests to be run in parallel to try and improve performance
2. `BIP32PathTest` -- this is old code, so I reduce the number of iterations on property based tests from `100` -> `20`.


core-test/src/test/scala/org/bitcoins/core/crypto/bip32/BIP32PathTest.scala
```diff
@@ -16,7 +16,7 @@ import scala.util.{Success, Try}
 class BIP32PathTest extends BitcoinSUnitTest {
 
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
-    generatorDrivenConfigNewCode
+    generatorDriveConfigOldCode
```
benthecarman, line 19: Is this safe? Maybe a better idea might be to split `BIP32PathTest` into 2 separate files, one with ones we running 100 times, and ones we want running 20 times.

core-test/src/test/scala/org/bitcoins/core/protocol/blockchain/BlockTest.scala
```diff
@@ -43,4 +45,14 @@ class BlockTest extends BitcoinSUnitTest {
     val time = timeBlockParsing(Block.fromHex(lines))
     assert(time <= 15000)
   }
+
+  it must "have serialization symmetry" in {
+    forAllAsync(BlockchainElementsGenerator.block) { block =>
+      Future {
```
nkohen: This seems like a pattern, perhaps we should make a `forAllInParallel` which takes in a `T => Assertion` and wraps the call in a `Future` rather than having users do it. This will hopefully avoid mix-ups in the future as if something becomes future-ified it will have to be adapted with more care. E.g.
```
forAllInParallel(BlockChainElementsGenerator.block) { block =>
  val result = Block(block.hex) == block
  if (!result) logger.warn("block.hex: " + block.hex)
  assert(result)
}
```

PR #1450, opened 2020-05-19
Small optimization for ScriptOperationFactory.operations

benthecarman:


core/src/main/scala/org/bitcoins/core/script/constant/StackPushOperationFactory.scala
```diff
@@ -21,7 +21,7 @@ trait StackPushOperationFactory {
     * @return
     */
   private def operations =
```
Christewart: This should be made a `val` do we don't have to run the comptuation associated with building `operations` everytime we call it.

PR #1452, opened 2020-05-20
Implement best block hash correctly

### LINKED ISSUE #1211: Implement `getBestBlockHash()` and `getBestFilterHeader()` correctly
Currently our `ChainHandler.getBestBlockHash()` and `ChainHandler.getBestFilterHeader()` are not implemented correctly. We just do the dumb thing and return the first header we see at at the tallest height. 

We need to implement the ability to calculate the heaviest chain: 

https://github.com/bitcoin-s/bitcoin-s/blob/d9881dd1539b93982dc464243c6354408f9d0696/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala#L111

https://github.com/bitcoin-s/bitcoin-s/pull/1209#discussion_r389825430

https://bitcoin.org/en/glossary/block-chain

benthecarman:
Closes #1211 

This has all tests passing but is having problems when running a `bitcoin-s-server` that previous did not have `chainWork` in its database. 

It takes around 4 hours to calculate work for an old db.


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -486,6 +453,101 @@ case class ChainHandler(
     }
     loop(Future.successful(to), Vector.empty)
   }
+
+  def isMissingChainWork: Future[Boolean] = {
+    for {
+      first100 <- blockHeaderDAO.getBetweenHeights(0, 100)
+      first100MissingWork = first100.nonEmpty && first100.exists(
+        _.chainWork == BigInt(0))
```
nkohen, line 466: won't this be problematic if someone starts to calculate chain work but doesn't finish?

nkohen (approved):

PR #1454, opened 2020-05-21
Fix fee rate calculation bugs, add tests

benthecarman:
Our fee rates were calculating like everything was sats/vbyte.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
@@ -10,7 +10,7 @@ import org.bitcoins.core.protocol.transaction.Transaction
 sealed abstract class FeeUnit {
   def currencyUnit: CurrencyUnit
   def *(tx: Transaction): CurrencyUnit = calc(tx)
-  def calc(tx: Transaction): CurrencyUnit = Satoshis(tx.vsize * toLong)
+  def calc(tx: Transaction): CurrencyUnit
   def toLong: Long = currencyUnit.satoshis.toLong
 }
 
@@ -25,17 +25,23 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
   def toSatPerKb: SatoshisPerKiloByte = {
     SatoshisPerKiloByte(currencyUnit.satoshis * Satoshis(1000))
   }
+
+  override def calc(tx: Transaction): CurrencyUnit =
+    Satoshis(tx.byteSize * toLong)
 }
 
 object SatoshisPerByte {
   def fromLong(sats: Long): SatoshisPerByte = SatoshisPerByte(Satoshis(sats))
 }
 
+/**
+  * KiloBytes here are defined as 1000 bytes.
+  */
 case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
     extends BitcoinFeeUnit {
 
   def toSatPerByte: SatoshisPerByte = {
-    val conversionOpt = (currencyUnit.toBigDecimal * 0.001).toBigIntExact
+    val conversionOpt = (currencyUnit.toBigDecimal / 1000).toBigIntExact
     conversionOpt match {
       case Some(conversion) =>
         val sat = Satoshis(conversion)
@@ -45,8 +51,10 @@ case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
         throw new RuntimeException(
           s"Failed to convert sat/kb -> sat/byte for ${currencyUnit}")
     }
-
   }
+
+  override def calc(tx: Transaction): CurrencyUnit =
+    Satoshis((tx.byteSize * toLong / 1000))
 }
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -25,17 +25,23 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
...
 
+/**
+  * KiloBytes here are defined as 1000 bytes.
+  */
 case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
     extends BitcoinFeeUnit {
 
   def toSatPerByte: SatoshisPerByte = {
-    val conversionOpt = (currencyUnit.toBigDecimal * 0.001).toBigIntExact
+    val conversionOpt = (currencyUnit.toBigDecimal / 1000).toBigIntExact
```
Christewart: Is there precision issues here?

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -45,8 +51,10 @@ case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
         throw new RuntimeException(
           s"Failed to convert sat/kb -> sat/byte for ${currencyUnit}")
     }
-
   }
+
+  override def calc(tx: Transaction): CurrencyUnit =
+    Satoshis((tx.byteSize * toLong / 1000))
```
Christewart: Same as above, what's the behavior with precision here? Is this what we want?

Christewart: As an example: 

```scala
scala> 101 * 100 / 1000
val res3: Int = 10

scala> 101 * 100 * 0.001
val res4: Double = 10.1
```

Your implementation silently rounds the `0.1`, which is scary. My implementation presreves the `0.1` which would end up in a runtime exception when passing it to `Satoshis()`

benthecarman: I think for the `toSatPerByte` it should be changed back, however, for the others we don't want it throw a runtime exception when calculating a fee for anything that isn't a multiple of 1000

benthecarman: Curious on @nkohen's thoughts as he is the one who prompted me to make this change

Christewart: The general rule of thumb that I think we should follow is: 

Implicitly rounding decimal places in money handling libraries is a BAD idea. Be explicit and throw and make the end user _explicitly_ handle that case. This may mean we need to offer `RoundingMode` in the API.

Perhaps we should look to see how bitcoin core does this?

benthecarman: image

benthecarman: https://github.com/bitcoin/bitcoin/blob/master/src/policy/feerate.cpp#L23

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
nkohen: I think using `Exact` means it throws if there were to be any precision issue?

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -25,28 +25,45 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
   def toSatPerKb: SatoshisPerKiloByte = {
     SatoshisPerKiloByte(currencyUnit.satoshis * Satoshis(1000))
   }
+
+  override def calc(tx: Transaction): CurrencyUnit =
+    Satoshis(tx.byteSize * toLong)
```
nkohen, line 30: Is this correct behavior for if `tx` is a `WitnessTransaction`? It might be but could you double check?

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -25,28 +25,45 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
...
+    val conversionOpt = (currencyUnit.toBigDecimal / 1000.0).toBigIntExact
     conversionOpt match {
       case Some(conversion) =>
         val sat = Satoshis(conversion)
         SatoshisPerByte(sat)
 
       case None =>
         throw new RuntimeException(
-          s"Failed to convert sat/kb -> sat/byte for ${currencyUnit}")
+          s"Failed to convert exactly sat/kb -> sat/byte for ${currencyUnit}")
```
nkohen: nit: I'd prefer `s"Failed to convert sat/kb -> sat/byte (loss of precision)  for $currencyUnit"`

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
benthecarman: It is. I am trying to find another library with multiple types for their fee rates but can't find one to prove it you though

benthecarman: There is a test that tests the expected behavior that witness txs have different fees for sats/byte vs sats/vbyte

PR #1459, opened 2020-05-22
Attempt to optimize block tests more to avoid timeouts

Christewart:
More optimizaitons to reduce time it's takes to parse a block

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.core.number
 
 import org.bitcoins.core.util.{BytesUtil, NumberUtil}
-import org.bitcoins.crypto.{Factory, NetworkElement}
+import org.bitcoins.crypto.{CryptoBytesUtil, Factory, NetworkElement}
 import scodec.bits.{ByteOrdering, ByteVector}
 
 import scala.util.{Failure, Success, Try}
@@ -144,7 +144,6 @@ sealed abstract class UInt64 extends UnsignedNumber[UInt64] {
   override def apply: A => UInt64 = UInt64(_)
   override def andMask = 0xFFFFFFFFFFFFFFFFL
 
-  private val Z = "0"
   /**
     * Converts a [[BigInt]] to a 8 byte hex representation.
     * [[BigInt]] will only allocate 1 byte for numbers like 1 which require 1 byte, giving us the hex representation 01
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -1,7 +1,7 @@
 package org.bitcoins.core.number
 
 import org.bitcoins.core.util.{BytesUtil, NumberUtil}
-import org.bitcoins.crypto.{Factory, NetworkElement}
+import org.bitcoins.crypto.{CryptoBytesUtil, Factory, NetworkElement}
```
nkohen, line 4: nit: We should use `BytesUtil` instead of `CryptoBytesUtil` everywhere but in `Crypto`

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -326,14 +323,31 @@ object MultiSignatureScriptSignature
...
+      case Nil => true
+      case h +: t =>
+        if (h.isInstanceOf[ScriptConstant] || StackPushOperationFactory
+          .isPushOperation(h)) {
+          isPushOpsOrScriptConstants(t)
+        } else {
+          false
+        }
+    }
+  }
```
nkohen: I think I prefer the following if there's no reason not to
```asm.forall(token => token.isInstanceOf[ScriptConstant] || StackPushOperationFactory.isPushOperation(token))```

File: crypto/src/main/scala/org/bitcoins/crypto/CryptoBytesUtil.scala [hunk 3]
```diff
@@ -63,17 +66,22 @@ trait CryptoBytesUtil {
...
     * for instance, ints are required to be 4 bytes. If the number is just 1
     * it will only take 1 byte. We need to pad the byte with an extra 3 bytes so the result is
     * 00000001 instead of just 1.
     */
-  private def addPadding(charactersNeeded: Int, hex: String): String = {
-    val paddingNeeded = charactersNeeded - hex.length
-    val padding = Vector.fill(paddingNeeded)("0")
-    val paddedHex = padding.mkString + hex
-    paddedHex
+  final def addPadding(charactersNeeded: Int, hex: String): String = {
```
nkohen: nit: `charactersNeeded -> paddingNeeded` as it is no longer how many total characters are needed

crypto/src/main/scala/org/bitcoins/crypto/CryptoBytesUtil.scala
```diff
@@ -63,17 +66,22 @@ trait CryptoBytesUtil {
...
+    val builder = new StringBuilder
+    var counter = 0
+    while (counter < charactersNeeded) {
+      builder.append(Z)
+      counter+=1
+    }
```
nkohen:
```
(0 until charactersNeeded).foreach { _ => builder.append(Z) }
```
is more scala-like

nkohen (commented):
Mostly nits

crypto/src/main/scala/org/bitcoins/crypto/CryptoBytesUtil.scala
Christewart: it also requires allocation of a `Range` which slows us down

PR #1461, opened 2020-05-24
Minor Fixes

nkohen:
Fixed sequence computation for P2PKWithTimeout and RawTxFinalizer composition discovered during DLC rebase.


benthecarman:
Are the secp changes intentional?

PR #1463, opened 2020-05-25
Create a simplified version of the WalletApi.unmarkUTXOsAsReserved() …

Christewart:
…that just takes in a tx and scans outpoints if they are in our wallet, also move the mark/unmark methods out of Wallet.scala and into UtxoHandling.scala

Builds off of #1458 

1. Adds `unmarkUTXOsAsReserved` that takes in a tx and then scans outpoints looking for ones that are in our wallet, and then un-reserves them. 
2. Moves the Utxo methods out of `Wallet.scala` and into `UtxoHandling.scala`


benthecarman (approved):
I think UTXOHandling needs scalafmt, otherwise LGTM

PR #1465, opened 2020-05-26
Increase Key Manager test coverage

benthecarman:
Needed for #1386


key-manager-test/src/test/scala/org/bitcoins/keymanager/util/HdUtilTest.scala
```diff
@@ -35,7 +35,7 @@ class HdUtilTest extends KeyManagerUnitTest {
       HDUtil.getXpubVersion(HDPurposes.NestedSegWit, TestNet3) == NestedSegWitTestNet3Pub)
 
     assertThrows[IllegalArgumentException] {
-      HDUtil.getXpubVersion(HDPurpose(-1), TestNet3)
+      HDUtil.getXpubVersion(HDPurpose(1), TestNet3)
```
benthecarman, line 38: These we're causing an `IllegalArgumentException` from `BIP32Node` not from `HDUtil`

PR #1470, opened 2020-05-26
Fee Rate Api Support

### LINKED ISSUE #1163: Implement ability to fetch fee rates from APIs
Create a `FeeFetcher` interface in the core project, and then implement a few fee provider APIs and then allow a fee provider to be specific in our config file

benthecarman:
Closes #1163 

Preliminary api support for fetching fee rates. Added https://bitcoiner.live/ to start off with because it had a simple api.


Christewart (commented):
So I think we should actually breakout the fee providers into a separate module. I don't want to add a bunch of dependencies into the wallet project. Then the wallet can just take a `FeeRateApi` as a parameter to the `WalletApi`

I also think the `FeeRateApi` should be moved into the `core` project so that it is available for everyone to implement. The `FeeRateApi` should not have any implemented methods

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -97,13 +97,12 @@ case class WalletRoutes(wallet: WalletApi, node: Node)(
...
-            val feeRate =
-              satoshisPerVirtualByteOpt.getOrElse(SatoshisPerByte(100.satoshis))
-            wallet.sendToAddress(address, bitcoins, feeRate).map { tx =>
-              node.broadcastTransaction(tx)
-              Server.httpSuccess(tx.txIdBE)
-            }
+            wallet
+              .sendToAddress(address, bitcoins, satoshisPerVirtualByteOpt)
+              .map { tx =>
+                node.broadcastTransaction(tx)
```
Christewart: We should probably `.flatMap { _ => Server.httpSuccess(tx.txIdBe) }`

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -156,14 +150,14 @@ case class WalletRoutes(wallet: WalletApi, node: Node)(
...
-            val feeRate =
-              satoshisPerVirtualByteOpt.getOrElse(SatoshisPerByte(100.satoshis))
-            wallet.makeOpReturnCommitment(message, hashMessage, feeRate).map {
-              tx =>
+            wallet
+              .makeOpReturnCommitment(message,
+                                      hashMessage,
+                                      satoshisPerVirtualByteOpt)
+              .map { tx =>
                 node.broadcastTransaction(tx)
```
Christewart: same here

fee-provider/src/main/scala/org/bitcoins/feeprovider/BitcoinerLiveFeeRateProvider.scala
```diff
@@ -0,0 +1,35 @@
...
+case class BitcoinerLiveFeeRateProvider(minutes: Int)(
+    implicit override val system: ActorSystem)
+    extends CachedHttpFeeRateApi {
+  private val bitcoinerLiveValidMinutes =
+    Vector(30, 60, 120, 180, 360, 720, 1440)
+  require(
+    bitcoinerLiveValidMinutes.contains(minutes),
+    s"$minutes is not a valid selection, must be from $bitcoinerLiveValidMinutes")
+
+  override def uri: Uri =
```
Christewart: nit: `override val`

fee-provider/src/main/scala/org/bitcoins/feeprovider/HttpFeeRateApi.scala
```diff
@@ -0,0 +1,74 @@
...
+      )
+  }
+}
+
+abstract class HttpFeeRateApi extends FeeRateApi {
+  implicit val system: ActorSystem
+
+  def uri: Uri
+
+  def converter: String => Try[FeeUnit]
```
Christewart: lt's just make this a normal method 

```scala
def converter(str: String): Try[FeeUnit] = ...
```

fee-provider/src/main/scala/org/bitcoins/feeprovider/HttpFeeRateApi.scala
```diff
@@ -0,0 +1,74 @@
...
+      .makeApiCall(uri)
+      .flatMap(ret => Future.fromTry(converter(ret)))(system.dispatcher)
+  }
+}
+
+abstract class CachedHttpFeeRateApi extends HttpFeeRateApi {
+
+  private var cachedFeeRateOpt: Option[(FeeUnit, Instant)] = None
+
+  val cacheDuration: Duration = Duration.ofMinutes(5)
```
Christewart: Probably should be private, and also configurable in a 
```
bitcoin-s { 
  fee-provider { 
    cache-duration = 5.minutes 
  }
}
```

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -444,10 +446,27 @@ trait WalletApi extends WalletLogger {
       outPoints: Vector[TransactionOutPoint],
       address: BitcoinAddress,
       amount: CurrencyUnit,
-      feeRate: FeeUnit): Future[Transaction] = {
+      feeRateOpt: Option[FeeUnit],
+      fromAccount: AccountDb): Future[Transaction] = {
+    for {
+      feeRate <- feeRateOpt match {
```
Christewart: Let's refactor this `feeRateOpt match { ... } ` pattern into a helper method somewhere

Christewart (commented):
It looks like ther is no unit tests with `BitcoinLiveFeeRateProvider`. Also like I said online it would be nice to have a dynamic fee test inside of `WalletIntegrationTest`.

fee-provider/src/main/scala/org/bitcoins/feeprovider/HttpFeeRateApi.scala
benthecarman: I think having it public is better, it allows other people using the library to set their own `cacheDuration`

PR #1473, opened 2020-05-26
Rework Standard Finalizer (into pieces)

nkohen:
Made the standard non-interactive tx finalizer simple using composition of its (reusable) pieces

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
@@ -80,94 +80,49 @@ case object FilterDustFinalizer extends RawTxFinalizer {
   }
 }
 
-/** A finalizer which takes as parameters an input info for each input, as well
-  * as a fee rate and change scriptpubkey and adds a change output resulting in
-  * the expected fee (after fee estimation).
+/** A finalizer who's Future fails if its sanity checks are not passed,
+  * otherwise it does nothing.
   */
-case class NonInteractiveWithChangeFinalizer(
+case class SanityCheckFinalizer(
     inputInfos: Vector[InputInfo],
-    feeRate: FeeUnit,
-    changeSPK: ScriptPubKey)
+    expectedOutputs: Vector[TransactionOutput],
+    expectedFeeRate: FeeUnit,
+    changeSPKs: Vector[ScriptPubKey] = Vector.empty)
     extends RawTxFinalizer {
   override def buildTx(txBuilderResult: RawTxBuilderResult)(
       implicit ec: ExecutionContext): Future[Transaction] = {
-    val RawTxBuilderResult(version, inputs, outputs, lockTime) = txBuilderResult
+    val tx = txBuilderResult.toBaseTransaction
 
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -80,94 +80,49 @@ case object FilterDustFinalizer extends RawTxFinalizer {
...
-        false
-      } else {
-        //the extra output should be the changeOutput
-        !(finalizedTx.outputs.size == (expectedOutputs.size + 1) &&
-          finalizedTx.outputs.map(_.scriptPubKey).contains(changeSPK))
-      }
+    val hasExtraOutputs = !finalizedTx.outputs.forall(
+      output =>
+        expectedOutputs.contains(output) || changeSPKs.contains(
+          output.scriptPubKey))
```
benthecarman: nit: could optimize this by doing:

```scala
val hasExtraOutputs = !finalizedTx.outputs.diff(expectedOutputs).forall(
      output => changeSPKs.contains(output.scriptPubKey))
```

that way you don't need to check the `expectedOutputs` twice

benthecarman: Also, there is on check that every `changeSpk` is included

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -182,6 +137,106 @@ object NonInteractiveWithChangeFinalizer {
...
+/** A finalizer which performs fee estimation and adds a
+  * change output resulting in the expected fee.
+  */
+case class ChangeFinalizer(
+    inputInfos: Vector[InputInfo],
+    feeRate: FeeUnit,
+    changeSPK: ScriptPubKey)
+    extends RawTxFinalizer {
+  override def buildTx(txBuilderResult: RawTxBuilderResult)(
+      implicit ec: ExecutionContext): Future[Transaction] = {
```
benthecarman, line 157: Should add dummy witness data for `txBuilderResult` that will result in witness txs.

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -182,6 +137,106 @@ object NonInteractiveWithChangeFinalizer {
...
+  }
+}
+
+/** A finalizer which adds only the witness data included in the
+  * wtxid (pubkey in P2WPKH and redeem script in p2sh).
+  *
+  * Note that when used in composition with other finalizers,
+  * if AddWitnessDataFinalizer is not last then its effects
+  * will be reversed since witness data is not currently kept
+  * between finalizers during compisition.
```
benthecarman: Should make an issue for this

benthecarman: > compisition

composition

benthecarman (approved):

PR #1475, opened 2020-05-26
Add more optimizations, MultSignatureScriptPubKey.isMultiSignatureScr…

Christewart:
…iptPubKey() fails fast if it's not a valid multisig spk, also use more caching inside of Number's (UInt64, UInt32, etc)

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/protocol/ln/ShortChannelIdTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/protocol/ln/ShortChannelIdTest.scala
@@ -36,7 +36,7 @@ class ShortChannelIdTest extends BitcoinSUnitTest {
       .fromHumanReadableString("0x0x65536")
     an[IllegalArgumentException] must be thrownBy ShortChannelId
       .fromHumanReadableString("0x0x-1")
-    an[NoSuchElementException] must be thrownBy ShortChannelId
+    an[IllegalArgumentException] must be thrownBy ShortChannelId
       .fromHumanReadableString("1x1x1x1")
     ShortChannelId.fromHumanReadableString("cafebabe") must be(
       ShortChannelId.fromHex("cafebabe"))
--- a/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberType.scala
@@ -82,7 +82,7 @@ sealed abstract class Number[T <: Number[T]]
     }
   }
 
-  override def bytes: ByteVector = BytesUtil.decodeHex(hex)
+  override lazy val bytes: ByteVector = BytesUtil.decodeHex(hex)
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -228,39 +228,47 @@ object MultiSignatureScriptPubKey
...
+              op.isInstanceOf[ScriptConstant] || op
+                .isInstanceOf[BytesToPushOntoStack] || op.isInstanceOf[ScriptNumber] || op == OP_CHECKMULTISIG ||
+                op == OP_CHECKMULTISIGVERIFY)
+
+          val result = isNotEmpty && containsMultiSigOp && hasRequiredSignatures &&
+            hasMaximumSignatures && isStandardOps
+          result
+        case (Success(_), Failure(_)) => false
+        case (Failure(_), Success(_)) => false
+        case (Failure(_), Failure(_)) => false
```
benthecarman, line 266: could condense these to one `case`

core-test/src/test/scala/org/bitcoins/core/protocol/ln/ShortChannelIdTest.scala
```diff
@@ -36,7 +36,7 @@ class ShortChannelIdTest extends BitcoinSUnitTest {
       .fromHumanReadableString("0x0x65536")
     an[IllegalArgumentException] must be thrownBy ShortChannelId
       .fromHumanReadableString("0x0x-1")
-    an[NoSuchElementException] must be thrownBy ShortChannelId
+    an[IllegalArgumentException] must be thrownBy ShortChannelId
```
benthecarman, line 39: Is indicative on an unintended change?

Christewart: This is caused from the change in `CryptoBytesUtil.decodeHex()` which now uses `ByteVector.fromValidHex()`. A `ShortChannelId` is not valid hex, which is why a `IllegalArgumentException` gets thrown.

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -379,7 +385,22 @@ object UInt64
...
   def apply(num: BigInt): UInt64 = UInt64Impl(num)
+
+  private def checkCached(num: Long): UInt64 = {
+    if (num == 0) zero
+    else if (num == 1) one
+    else if (num == 25) twentyFive
+    else if (num == 23) twentyThree
+    else if (num == 34) thirtyFour
+    else if (num == 22) twentyTwo
+    else if (num == 105) oneHundredFive
```
benthecarman, line 401: This might be done _slightly_ faster with a map

PR #1481, opened 2020-05-28
Implement Bitcoind as ChainQueryApi & NodeApi

### LINKED ISSUE #1207: Make `BitcoindRpcClient` trait extend `ChainQueryApi`
This will mean bitcoind can be a dropin replacement for our chain project from the perspective of anything that uses the `ChainQueryApi`, like our wallet project.

benthecarman:
Closes #1207

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -86,10 +86,7 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
   }
 
   /** Gets the number of compact filters in the database */
-  override def getFilterCount: Future[Int] =
-    Future.failed(
-      throw new UnsupportedOperationException(
-        s"bitcoind ${instance.getVersion} does not support block filters"))
+  override def getFilterCount: Future[Int] = ???
 
   /** Returns the block height of the given block stamp */
   override def getHeightByBlockStamp(blockStamp: BlockStamp): Future[Int] =
@@ -105,8 +102,7 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
 
   override def getFiltersBetweenHeights(
       startHeight: Int,
-      endHeight: Int): Future[Vector[ChainQueryApi.FilterResponse]] =
-    Future.successful(Vector.empty)
+      endHeight: Int): Future[Vector[ChainQueryApi.FilterResponse]] = ???
 
   /** Gets the block height of the closest block to the given time */
   override def epochSecondToBlockHeight(time: Long): Future[Int] =
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -3,8 +3,17 @@ package org.bitcoins.rpc.client.common
 import java.io.File
 
 import akka.actor.ActorSystem
+<<<<<<< HEAD
 import org.bitcoins.core.api.FeeRateApi
 import org.bitcoins.core.wallet.fee.FeeUnit
+||||||| merged common ancestors
+=======
+import org.bitcoins.core.api.{ChainQueryApi, NodeApi}
+import org.bitcoins.core.protocol.BlockStamp
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.core.util.FutureUtil
+import org.bitcoins.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+>>>>>>> Implement Bitcoind as ChainQueryApi & NodeApi
 import org.bitcoins.rpc.client.v16.BitcoindV16RpcClient
 import org.bitcoins.rpc.client.v17.BitcoindV17RpcClient
 import org.bitcoins.rpc.client.v18.BitcoindV18RpcClient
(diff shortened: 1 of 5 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -44,6 +53,56 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
...
+  }
+
+  /** Gets number of confirmations for the given block hash */
+  override def getNumberOfConfirmations(
+      blockHash: DoubleSha256DigestBE): Future[Option[Int]] = {
+    getBlockHeader(blockHash).map(header => Some(header.confirmations))
+  }
+
+  /** Gets the number of compact filters in the database */
+  override def getFilterCount: Future[Int] =
```
Christewart: Not be implemented here and then implemented in the specific version types no?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -44,6 +53,56 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
...
+      new UnsupportedOperationException(
+        "epochSecondToBlockHeight is not supported by bitcoind"))
+
+  // Node Api
+
+  override def broadcastTransaction(transaction: Transaction): Future[Unit] =
+    sendRawTransaction(transaction).map(_ => ())
+
+  override def downloadBlocks(
+      blockHashes: Vector[DoubleSha256Digest]): Future[Unit] = FutureUtil.unit
```
Christewart, line 114: Can we not retrieve full blocks from disk?

benthecarman: We can, but that won't actually do anything for the callback because it needs to come from the network to be passed through the data message handler, I might be able to fix this actually. Will take a second look.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
benthecarman: It is overridden for the versions that support it.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
benthecarman: So, it's not actually possible to call or set the wallet callbacks without having the wallet actually present. Maybe this should be changed to return a `Future[Vector[Block]]`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
Christewart: It's bad idea to override in the super type. For instance, are you going to remember to update this for V20 & V21? You aren't going to get a compiler warning -- which makes it _super_ easy to forget.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
Christewart: Yeah, I think we should consider re-working this as it doesn't make much sense. Maybe file an issue and we will go with this implementation for now?

Christewart: https://github.com/bitcoin-s/bitcoin-s/issues/1497

benthecarman: Oh, I made #1487

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -58,6 +66,52 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(
...
+  }
+
+  /** Gets number of confirmations for the given block hash */
+  override def getNumberOfConfirmations(
+      blockHash: DoubleSha256DigestBE): Future[Option[Int]] = {
+    getBlockHeader(blockHash).map(header => Some(header.confirmations))
+  }
+
+  /** Gets the number of compact filters in the database */
+  override def getFilterCount: Future[Int] = ???
```
Christewart, line 84: Why is this implemented here?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v19/BitcoindV19RpcClient.scala
```diff
@@ -36,6 +39,23 @@ class BitcoindV19RpcClient(override val instance: BitcoindInstance)(
...
+        .map { height =>
+          for {
+            hash <- getBlockHash(height)
+            filter <- getBlockFilter(hash, FilterType.Basic)
+          } yield FilterResponse(filter.filter, hash, height)
+        }
+        .toVector)
+  }
+
+  override def getFilterCount: Future[Int] = getBlockCount
```
Christewart, line 57: I don't think this is right, is there really no `getfiltercount` in any bitcoind version?

benthecarman: No, after getting a block it'll calculate the filter immediately so there should never be a point where they are different.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
benthecarman: So `BitcoindRpcClient` doesn't need to be `abstract`

PR #1499, opened 2020-06-02
Testkit wallet with bitcoind uses bitcoind as api

benthecarman:
Needed to fix `BitcoindV19RpcClient`'s `getFiltersBetweenHeights` to use batching so it doesn't overload akka with requests.

After this was able to verify that the neutrino coinbases issues do not exist, at least in tests.

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
@@ -1,17 +1,17 @@
 package org.bitcoins.testkit.chain
 
 import org.bitcoins.chain.blockchain.sync.FilterWithHeaderHash
+import org.bitcoins.commons.jsonmodels.bitcoind.GetBlockFilterResult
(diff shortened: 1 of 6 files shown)


testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala
```diff
@@ -45,75 +45,41 @@ abstract class SyncUtil extends BitcoinSLogger {
       }
   }
 
-  def getChainQueryApi(bitcoindV19RpcClient: BitcoindV19RpcClient)(
-      implicit ec: ExecutionContext): ChainQueryApi = {
+  def getTestChainQueryApi(bitcoind: BitcoindRpcClient): ChainQueryApi = {
```
Christewart, line 48: Cant this just be deleted now? Or simply do 

```scala

def getTestChainQueryApi(bitcoind: BitcoindRpcClient): ChainQueryApi = {
bitcoind
}

```
?

PR #1500, opened 2020-06-02
DLC TableView and DLCStatus ADT

nkohen:
Introduced a `DLCStatus` state machine as well as adding a table-view to the GUI which displays some relevant information from these statuses to the user.


app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+      case _: Confirmed       => "CONFIRMED"
+      case _: CloseOffered    => "CLOSE OFFERED"
+      case _: Closed          => "CLOSED"
+      case _: Claiming        => "CLAIMING"
+      case _: Claimed         => "CLAIMED"
+      case _: Penalized       => "PENALIZED"
+      case _: RemoteClaiming  => "REMOTE CLAIMING"
+      case _: RemoteClaimed   => "REMOTE CLAIMED"
+      case _: RemotePenalized => "REMOTE PENALIZED"
+      case _: Refunded        => "REFUNDED"
```
benthecarman: Why not just give the `DLCStatus` trait a `statusString` function

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -161,9 +164,107 @@ class DLCPane(glassPane: VBox) {
...
+    }
+
+    val initiatorCol = new TableColumn[DLCStatus, String] {
+      text = "Initiator"
+      prefWidth = 80
+      cellValueFactory = { status =>
+        val str = if (status.value.isInitiator) {
+          "YES"
+        } else {
+          "NO"
```
benthecarman, line 193: nit: I prefer `Yes` and `No` but do what you want

benthecarman (approved):

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
benthecarman: Also pref title case over all caps

nkohen: My reasoning was that layer/responsibility-wise this seems like more of a front-end/client/user utility than something that belongs to the elements of the ADT.

nkohen: (this was originally something I wrote in the `DLCPane` but this seemed like a better place)

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
nkohen: I'll probably leave it all caps for now because that's what I did with status as well, but open to changing this in the future of course

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
Christewart: > Why not just give the `DLCStatus` trait a `statusString` function

Or just redefine `toString`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+               oracleSig,
+               cet)
+    }
+
+    def toRemoteClaiming(cet: Transaction): RemoteClaiming = {
+      RemoteClaiming(eventId, isInitiator, offer, accept, sign, fundingTx, cet)
+    }
+  }
+
+  case class CloseOffered(
```
Christewart: scaladocs would be useful, so this is for the case where we are aborting and offer?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+
+  case class Claiming(
+      eventId: Sha256DigestBE,
+      isInitiator: Boolean,
+      offer: DLCOffer,
+      accept: DLCAccept,
+      sign: DLCSign,
+      fundingTx: Transaction,
+      oracleSig: SchnorrDigitalSignature,
+      cet: Transaction)
```
Christewart: Isn't this required to be a `WitnessTransaction`? Let's tighten the type if we can. The same applies to other CETs here I think. I think the funding transaction can be non-witness tx though right?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+      extends AcceptedDLCStatus
+
+  case class Refunded(
+      eventId: Sha256DigestBE,
+      isInitiator: Boolean,
+      offer: DLCOffer,
+      accept: DLCAccept,
+      sign: DLCSign,
+      fundingTx: Transaction,
+      refundTx: Transaction)
```
Christewart: witness tx too?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+      isInitiator: Boolean,
+      offer: DLCOffer,
+      accept: DLCAccept,
+      sign: DLCSign,
+      fundingTx: Transaction,
+      closeSig: DLCMutualCloseSig,
+      closeTx: Transaction)
+      extends AcceptedDLCStatus
+
+  case class Claiming(
```
Christewart: Is `Claiming` equivalent to `ExecutingCET`?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -0,0 +1,295 @@
...
+      offer: DLCOffer,
+      accept: DLCAccept)
+      extends AcceptedDLCStatus {
+
+    def toSigned(sign: DLCSign): Signed = {
+      Signed(eventId, isInitiator, offer, accept, sign)
+    }
+  }
+
+  case class Signed(
```
Christewart: So at this state either party can broadcast the DLC and have it confirmed on the chain right? It's probably a good idea to have this state, but we need to be careful when handling it as I think it is effectively equivalent to confirmed. 

Bitcoin exchanges use this semantic whenever they sign transactions internally because as soon as you authorize a transaction with digital signatures, it can be broadcast by anyone and have funds sent from your wallet. Until this is double spent, or entirely purged from your system (very hard to do!), these funds are at risk of leaving your wallet. 

If you issue multiple `Signed` to your counter party, and they don't use the same utxos, they can broadcast both of them and steal money from you. I guess there isn't an actionable item here, but more of a warning. 

Perhaps it's worth creating a higher level type in the ADT representing when funds are sign by both parties and are in flight. Is that what `AcceptedDLCStatus` is?

benthecarman:
replaced by #2203

PR #1504, opened 2020-06-03
Document LN Data types

### LINKED ISSUE #525: Add showcase for decoding a Lightning Invoice on bitcoin-s.org

benthecarman:
Closes #525

### DIFF
--- a/docs/core/lightning-network.md
+++ b/docs/core/lightning-network.md
@@ -15,7 +15,7 @@ import scala.util._
 import scodec.bits._
 ```
 
-```scala mdoc:compile-only
+```scala mdoc:to-string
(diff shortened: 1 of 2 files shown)


docs/core/lightning-network.md
```diff
@@ -0,0 +1,29 @@
...
+```scala mdoc:invisible
+import org.bitcoins.core.protocol.ln._
+import org.bitcoins.core.protocol.ln.currency._
+import org.bitcoins.core.protocol.ln.node._
+import org.bitcoins.crypto._
+import scala.util._
+import scodec.bits._
+```
+
+```scala mdoc:compile-only
```
Christewart: So I don't think we want `compile-only` here, I think we want  `to-string` so we get nice comments on the doc that look like this

```
// this generates a random private key
val privkey = ECPrivateKey()
// privkey: ECPrivateKey = Masked(ECPrivateKeyImpl)
val pubkey = privkey.publicKey
// pubkey: org.bitcoins.crypto.ECPublicKey = ECPublicKey(033ee1b9566f60866af0aae5b02cc5148b91a1cf3d3b3f0b35c9f6368b3b3a9012)

val segwitAddress = {
    // see https://bitcoin.org/en/glossary/pubkey-script
    // for reading resources on the details of scriptPubKeys
    // pay-to-witness-pubkey-hash scriptPubKey V0
    val scriptPubKey = P2WPKHWitnessSPKV0(pubkey)
    Bech32Address(scriptPubKey, TestNet3)
}
// segwitAddress: Bech32Address = tb1qaftdnhe2svfu40kg64nf4nv99kjan9phpuaggw

println(segwitAddress.toString)
// tb1qaftdnhe2svfu40kg64nf4nv99kjan9phpuaggw
```

https://bitcoin-s.org/docs/next/core/addresses

PR #1506, opened 2020-06-03
Wallet API DLC Status Endpoints (the hairy version)

nkohen:
Built on top of #1500

Introduces `getdlcs` and `getdlc` wallet api endpoints


benthecarman:
replaced by #2203

PR #1511, opened 2020-06-04
Optimize and Clean up WalletRpcTest

benthecarman:
`be able to import an address` would sometimes fail because the receiver would not have seen the block yet, now if it hasn't seen it we will publish the block from that client as well.

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
@@ -4,17 +4,17 @@ import java.io.File
 import java.util.Scanner
 
 import org.bitcoins.commons.jsonmodels.bitcoind.RpcOpts
+import org.bitcoins.commons.jsonmodels.bitcoind.RpcOpts.AddressType
+import org.bitcoins.core.crypto.ECPrivateKeyUtil
 import org.bitcoins.core.currency.{Bitcoins, CurrencyUnit, Satoshis}
-import org.bitcoins.core.number.{Int64, UInt32}
+import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.protocol.script.ScriptSignature
 import org.bitcoins.core.protocol.transaction.{
   TransactionInput,
   TransactionOutPoint
 }
-import org.bitcoins.core.protocol.{BitcoinAddress, P2PKHAddress}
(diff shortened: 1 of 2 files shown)


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
```diff
@@ -357,9 +357,13 @@ class WalletRpcTest extends BitcoindRpcTest {
   }
 
   it should "be able to import an address" in {
+
+    val address = Bech32Address
```
Christewart, line 361: Why have this hard coded?

benthecarman: So we don't need to fetch an address from one of the clients, one less thing that can fail

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -852,7 +865,14 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
       txid <- fundMemPoolTransaction(sender, address, amount)
       addr <- sender.getNewAddress
       blockHash <- sender.generateToAddress(1, addr).map(_.head)
-      _ <- hasSeenBlock(receiver, blockHash)
+      seenBlock <- hasSeenBlock(receiver, blockHash)
+      _ <- if (seenBlock) {
+        FutureUtil.unit
+      } else {
+        sender
+          .getBlockRaw(blockHash)
```
Christewart, line 873: How does this work compared to `getBlock()` which in turn is querying `bitcoind`? Why will `getBlockRaw` succeed while `getBlock()` doesn't?

benthecarman: `getBlock` returns information about the block (a `GetBlockResult`) vs `getBlockRaw` returns an actual `Block`

Christewart: Why would `getBlockRaw` work and `getBlock` wouldn't work in the same bitcoind instance

benthecarman: The problem wasn't that they were failing, the problem was that the `receiver` had not seen the block, so we use `getBlockRaw` on the `sender` and give it to the `receiver` manually

PR #1515, opened 2020-06-04
Attempt to fix BitcoindV19RpcClientTest from losing connection

benthecarman:
In CI this would sometimes give
```
should check to see if the utxoUpdate input has been updated *** FAILED *** (10 milliseconds)
[info]   akka.stream.StreamTcpException: The connection closed with error: Connection reset
```
hopefully this fixes

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
@@ -101,10 +101,11 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
 
     val psbt =
       "cHNidP8BACoCAAAAAAFAQg8AAAAAABepFG6Rty1Vk+fUOR4v9E6R6YXDFkHwhwAAAAAAAA=="
-    val updatedF =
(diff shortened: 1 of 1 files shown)


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v19/BitcoindV19RpcClientTest.scala
```diff
@@ -101,10 +101,11 @@ class BitcoindV19RpcClientTest extends BitcoindRpcTest {
 
     val psbt =
       "cHNidP8BACoCAAAAAAFAQg8AAAAAABepFG6Rty1Vk+fUOR4v9E6R6YXDFkHwhwAAAAAAAA=="
-    val updatedF =
-      clientF.flatMap(client => client.utxoUpdatePsbt(psbt, Seq(descriptor)))
 
-    updatedF.map { result =>
+    for {
+      (client, _) <- clientPairF
```
Christewart, line 106: How is this different than the previous version? it seems logically equivalent to me

benthecarman: Using `clientPairF` instead of `client` was hoping that might fix.

Issue #1508, opened 2020-06-04
Deadlock on appServer/run after upgrading to akka 2.6


Christewart:
**Describe the bug**
We deadlock and never return from `Http.bindAndHandle()` inside of `Server.start()`

https://github.com/bitcoin-s/bitcoin-s/blob/95fb2b7e24b3de448597a04caded69eb86f9fa11/app/server/src/main/scala/org/bitcoins/server/Server.scala#L82

**To Reproduce**
It should be reproducible off of 8d21cd137824fe30648cedf6a1c7936a1eae9245 when akka 2.6 was introduced

Christewart:
So it seems that the threads in akka `default-dispatcher` and `internal-dispatcher` are slowly dying after the application is started. I'm not sure why this is the case at the moment 
![Screenshot from 2020-06-04 08-29-07](https://user-images.githubusercontent.com/3514957/83762835-99192800-a63d-11ea-8878-314d30c09f66.png)

Christewart:
This looks like the NPE is an upstream issue in akka: https://github.com/akka/akka-http/issues/3221#issuecomment-644774799 

https://github.com/akka/akka/pull/29244

PR #1521, opened 2020-06-05
Add call to HttpEntity.Empty to avoid NPE in akka

Christewart:
This is a workaround for #1508

The server still binds, but this doesn't fix the root cause of https://github.com/akka/akka-http/issues/3221 which is a NPE exception thrown in binding and handling the server


Christewart:
This is included in #1521

PR #1537, opened 2020-06-09
2020 06 09 string factory

Christewart:
This adds a common StringFactory trait that can be used to deserialize a string to a type T. This is should be used for everything in our code base so we don't have a mismatch of how things are deserialized from a String.

### DIFF
--- a/crypto/src/main/scala/org/bitcoins/crypto/StringFactory.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/StringFactory.scala
@@ -1,0 +1,20 @@
+package org.bitcoins.crypto
+
+import scala.util.Try
+
+/** A common factory trait that can be re-used to deserialize a string to a type t */
+trait StringFactory[T] {
+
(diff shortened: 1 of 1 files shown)


benthecarman:
I think this can be moved to core, there isn't anything in crypto that should be extending this as everything there should be just using `Factory` because they are encoded in hex

crypto/src/main/scala/org/bitcoins/crypto/StringFactory.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  /** Treis to parse a string to type t, returns None if failure */
+  def fromStringOpt(string: String): Option[T] = {
+    fromStringT(string).toOption
+  }
+
+  /** Tries to parsea  string to type t, returns [[scala.util.Failure]] if the fails */
+  def fromStringT(string: String): Try[T] = {
+    Try(fromString(string))
+  }
```
nkohen, line 19: https://github.com/bitcoin-s/bitcoin-s/pull/1538#discussion_r437526845

PR #1538, opened 2020-06-09
2020 06 09 addr invoice string factory

Christewart:
This is built on #1537 

This implements `StringFactory` with `Address` and `LnInvoice` to get a uniform interface.


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -355,18 +370,23 @@ object P2SHAddress extends AddressFactory[P2SHAddress] {
...
-  def apply(value: String): Try[BitcoinAddress] = fromString(value)
-
-  override def fromString(value: String): Try[BitcoinAddress] = {
-    P2PKHAddress
-      .fromString(value)
-      .orElse(P2SHAddress.fromString(value))
-      .orElse(Bech32Address.fromString(value))
-      .orElse(Failure(new IllegalArgumentException(
-        s"Could not decode the given value to a BitcoinAddress, got: $value")))
+  def apply(value: String): BitcoinAddress = fromString(value)
```
benthecarman, line 373: If we're going to do this it should go on the `StringFactory`

app-commons/src/main/scala/org/bitcoins/commons/serializers/JsonSerializers.scala
```diff
@@ -599,7 +599,7 @@ object JsonSerializers {
   implicit def mapAddressesByLabelReads: Reads[
     Map[BitcoinAddress, LabelResult]] =
     Reads.mapReads[BitcoinAddress, LabelResult](s =>
-      JsSuccess(BitcoinAddress.fromString(s).get))
+      JsSuccess(BitcoinAddress.fromStringT(s).get))
```
nkohen: nit: `fromString`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
```diff
@@ -78,7 +78,7 @@ trait WalletRpc { self: Client =>
       List(JsString(accountOrLabel)) ++ addressType.map(Json.toJson(_)).toList
 
     bitcoindCall[BitcoinAddress]("getnewaddress", params).map(addr =>
-      BitcoinAddress.fromScriptPubKey(addr.scriptPubKey, instance.network).get)
+      BitcoinAddress.fromScriptPubKeyT(addr.scriptPubKey, instance.network).get)
```
nkohen: nit: replace T with get and just use the direct call

core/src/main/scala/org/bitcoins/core/protocol/AddressFactory.scala
```diff
@@ -2,26 +2,31 @@ package org.bitcoins.core.protocol
...
 
   /** Same as fromString, but throws the exception */
-  def fromStringExn(str: String): T = fromString(str) match {
-    case Success(addr) => addr
-    case Failure(exn)  => throw exn
+  def fromStringExn(str: String): T = fromString(str)
+
+
+  @deprecated(s"Use fromScriptPubKeyT", since="2020-06-09")
+  def fromScriptPubKey(spk: ScriptPubKey, np: NetworkParameters): Try[T] = {
```
nkohen: shouldn't this just be changed to match the normal pattern that we have now where this returns a `T` and we can just rework all calls to it?

crypto/src/main/scala/org/bitcoins/crypto/StringFactory.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  /** Treis to parse a string to type t, returns None if failure */
+  def fromStringOpt(string: String): Option[T] = {
+    fromStringT(string).toOption
+  }
+
+  /** Tries to parse string to type t, returns [[scala.util.Failure]] if the fails */
+  def fromStringT(string: String): Try[T] = {
+    Try(fromString(string))
+  }
```
nkohen: This seems backwards to me, I think that users should implement this and that `fromString` should default to `fromStringT.getOrElse(throw something)`

nkohen: specifically it should probably throw the failure:
```
def fromString(string: String): T = {
  fromStringT(string) match {
    case Success(t) => t
    case Failure(err) => throw err
  }
}

File: eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala [hunk 6]
```diff
@@ -807,7 +807,7 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
     }
 
     paymentRequestF.map { paymentRequest =>
-      val i = LnInvoice.fromString(paymentRequest.serialized).get
+      val i = LnInvoice.fromStringT(paymentRequest.serialized).get
```
nkohen: nit: `fromString`

crypto/src/main/scala/org/bitcoins/crypto/StringFactory.scala
Christewart: My thinking here is the default implementation -- that must be implemented if you want parsing functionality -- is the version that throws an exception. Since this is the "lowest common denominator" I figured this would be best left unimplemented. If you want to totally ignore the safe APIs you can. 

If you want to use safe APIs that return `Try` or `Option` you can still override the respective methods and implement `fromString` as you have said.

Issue #1530, opened 2020-06-07
Rework signing logic to take full funding transaction and hash it before signing


Christewart:
### The bug

Both segwit and pre-segwit transactions are vulnerable to overpaying fees. The only way to protect against this is by hashing the funding transction and comparing the txid in the outpoint to the _real_ funding txid before allowing a digital signature to be created. 

It was believed this was fixed in [BIP143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#motivation) with this: 

>The algorithm does not involve the amount of Bitcoin being spent by the input. This is usually not a problem for online network nodes as they could request for the specified transaction to acquire the output value. For an offline transaction signing device ("cold wallet"), however, the unknowing of input amount makes it impossible to calculate the exact amount being spent and the transaction fee. To cope with this problem a cold wallet must also acquire the full transaction being spent, which could be a big obstacle in the implementation of lightweight, air-gapped wallet. By including the input value of part of the transaction digest, a cold wallet may safely sign a transaction by learning the value from an untrusted source. In the case that a wrong value is provided and signed, the signature would be invalid and no funding might be lost. [5]

The key insight to this attack re-using the victim utxo's outpoints across multiple signing rounds with fake values. In BIP143, the signature only commits to _this utxos_ value, not _all_ utxos values. So you can re-use valid signatures that won't be invalidated even if other amounts change in other utxos used by this transaction.

If you don't _validate_ (by hashing the _all_ funding txs) that _all_ of the outpoints we are signing for are _real outputs_ on _real transactions_, an attacker can use real outpoints with incorrect values in conjunction with real outpoints with _real_ values, and then re-use the valid digital signature for the _real_ output with the correct value with another valid digital signature from another victim utxo to overpay fees.

This attack probably won't get caught by whitelisting policies or rate limiting policies as you can do this with as few as 2 signing rounds to cause loss of funds. You may be even to circumvent "amount based" rate limiting policies as this attack allows you to lie about the amounts you are signing for. If you have an exclusive locking scheme on utxos, your wallet software may prevent loss of funds here assuming that the attacker doesn't have access to modify that lock.

https://blog.trezor.io/details-of-firmware-updates-for-trezor-one-version-1-9-1-and-trezor-model-t-version-2-3-1-1eba8f60f2dd

### The fix

We need to hash all funding transactions before signing them and verify the hash in the spending transaction's outpoints match the funding transactions txid.

PR #1540, opened 2020-06-09
PSBT Segwit Vulnerability fix

benthecarman:
Related to #1530 
Should probably wait to merge this until the BIP 174 changes are posted and merged.

This might break the DLC branch


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -1047,6 +1047,8 @@ sealed trait WitnessScriptPubKey extends ScriptPubKey {
   def witnessVersion = WitnessVersion(asm.head)
 }
 
+sealed trait BIP143VulnerableScriptPubKey extends WitnessScriptPubKey
```
nkohen: Probably worth a doc here explaining the vulnerability

nkohen: Also is there any reason to introduce this trait rather than just using `WitnessScriptPubKeyV0`?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -607,6 +603,74 @@ case class PSBT(
...
 
+                  Success(tx)
+              }
+            case (None, wtx: WitnessTransaction) =>
+              utxoOpt match {
+                case Some(utxo) =>
+                  val input = tx.inputs(index)
+                  val output = utxo.transactionSpent.outputs(
+                    input.previousOutput.vout.toInt)
+                  val txSigComponent = output.scriptPubKey match {
```
nkohen: this block really seems like it deserves its own method in a different file

nkohen: btw, I suggest that this method take `Policy.standardFlags` as a param rather than hard coding it (just to make the future easier)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -607,6 +603,74 @@ case class PSBT(
...
+                  val inputResult = ScriptInterpreter.run(
+                    PreExecutionScriptProgram(txSigComponent))
+                  if (inputResult == ScriptOk) {
+                    Success(tx)
+                  } else {
+                    println("============")
+                    println(output.scriptPubKey)
+                    println(
+                      (inputMap.witnessScriptOpt, inputMap.redeemScriptOpt))
+                    println("============")
```

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -575,27 +575,35 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+          }
         } else {
           elements
         }
-      case _: P2PKHScriptPubKey | _: P2PKScriptPubKey |
-          _: P2PKWithTimeoutScriptPubKey | _: MultiSignatureScriptPubKey |
-          EmptyScriptPubKey | _: LockTimeScriptPubKey |
-          _: NonStandardScriptPubKey | _: WitnessCommitment |
-          _: ConditionalScriptPubKey =>
+      case _: BIP143VulnerableScriptPubKey | _: P2PKHScriptPubKey |
```
nkohen: I'm not convinced this new case is needed (should still be exhaustive without), also refer to my not liking this trait above)

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -575,27 +575,35 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
   private def changeToWitnessUTXO(
       transactionOutput: TransactionOutput): InputPSBTMap = {
     val newElements = transactionOutput.scriptPubKey match {
-      case _: WitnessScriptPubKey =>
+      case rs: WitnessScriptPubKey
+          if !rs.isInstanceOf[BIP143VulnerableScriptPubKey] =>
```
nkohen: oof, using `if` in `match` `case`es actually breaks exhaustiveness checks :(

nkohen: This should probably just go straight below as it is never possible

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -575,27 +575,35 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
       case _: P2SHScriptPubKey =>
-        if (redeemScriptOpt.isDefined && redeemScriptOpt.get.redeemScript
-              .isInstanceOf[WitnessScriptPubKey]) {
-          filterRecords(NonWitnessUTXOKeyId) :+ WitnessUTXO(transactionOutput)
+        if (redeemScriptOpt.isDefined) {
+          val redeemScript = redeemScriptOpt.get.redeemScript
+          if (redeemScript
+                .isInstanceOf[WitnessScriptPubKey] && !redeemScript
+                .isInstanceOf[BIP143VulnerableScriptPubKey]) {
+            filterRecords(NonWitnessUTXOKeyId) :+ WitnessUTXO(transactionOutput)
```
nkohen: this case isn't possible, no need to complicate things

core/src/main/scala/org/bitcoins/core/psbt/PSBTRecord.scala
```diff
@@ -152,6 +152,10 @@ object InputPSBTRecord extends Factory[InputPSBTRecord] {
 
   case class WitnessUTXO(witnessUTXO: TransactionOutput)
       extends InputPSBTRecord {
+//    require(
+//      !witnessUTXO.scriptPubKey.isInstanceOf[BIP143VulnerableScriptPubKey],
+//      "This UTXO is vulnerable to the BIP143 vulnerability, use NonWitnessOrUnknownUTXO instead"
+//    )
```
nkohen: Is there a reason this is commented out?

nkohen (commented):
So generally speaking I think that there's no need to over-engineer with taproot in mind as so much is going to need to change when that time comes and I don't think we'll end up doing it the way that is done here. Because of this, I propose we merge the `WitnessScriptPubKeyV0` type with the `BIP143VulnerableScriptPubKey` as they are the same and take the code simplifications this gives us for now rather than complicating code in ways we'll need to change later anyways

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
benthecarman: I think this is fine to keep for now, it should never be ran currently because there is only `WitnessScriptPubKeyV0`s, and come taproot we'll just have to add this back in

benthecarman:
Looks like `SignerTest`'s `sign p2wsh inputs correctly when provided no witness data` is broken, it is trivially fixed in #1560, probably best to close this and merge that PR with this built under it. 

It fails because we are using `                psbt.addWitnessUTXOToInput(spendInfo.output, idx)` and it throws an error for having a witness utxo with no prev tx. We don't have access to the prev tx here but will in the new PR.

benthecarman:
Closed in place of #1560

PR #1541, opened 2020-06-09
Migrated DLC message serialization from ujson to play json

nkohen:
It is an open question whether we want to go this route of using play json or if instead we wish to overhaul our current use of ujson to use nice composable pieces like is encouraged by play.


Christewart:
This is a pre-cursor to #1610 , we will pick this up off of master in v0.5.0

Issue #1532, opened 2020-06-08
Wallet callbacks


Christewart:
Should we add wallet callbacks in v0.4.0? If so, what do we need to add callbacks on?

PR #1542, opened 2020-06-10
Callback rework

benthecarman:
Part of larger work for #1532

Creates a `Callback` type and a `CallbackHandler` that cleans up the execution for us. Reworks the `NodeCallbacks` to use these.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/Callback.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/Callback.scala
@@ -4,17 +4,21 @@ import org.bitcoins.core.util.{FutureUtil, SeqWrapper}
 import org.slf4j.Logger
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
 
+/** A function to be called in response to an event */
 trait Callback[T] {
   def apply(param: T): Future[Unit]
 }
 
+/** A function with two parameters to be called in response to an event */
 trait CallbackBinary[T1, T2] extends Callback[(T1, T2)] {
   def apply(param1: T1, param2: T2): Future[Unit]
 
   override def apply(param: (T1, T2)): Future[Unit] = apply(param._1, param._2)
 }
 
+/** A function with three parameters to be called in response to an event */
 trait CallbackTernary[T1, T2, T3] extends Callback[(T1, T2, T3)] {
   def apply(param1: T1, param2: T2, param3: T3): Future[Unit]
 
@@ -22,6 +26,7 @@ trait CallbackTernary[T1, T2, T3] extends Callback[(T1, T2, T3)] {
     apply(param._1, param._2, param._3)
 }
 
+/** Manages a set of callbacks, should be used to manage execution and logging if needed */
 case class CallbackHandler[C, T <: Callback[C]](
     name: String,
     override val wrapped: IndexedSeq[T])
@@ -35,13 +40,11 @@ case class CallbackHandler[C, T <: Callback[C]](
 
   def execute(param: C, recoverFunc: Throwable => Unit = _ => ())(
       implicit ec: ExecutionContext): Future[Unit] = {
-    wrapped
-      .foldLeft(FutureUtil.unit)((acc, callback) =>
-        acc.flatMap(_ =>
-          callback(param).recover {
-            case err: Throwable =>
-              recoverFunc(err)
-          }))
+    FutureUtil.foldLeftAsync((), wrapped)((_, callback) =>
+      callback(param).recover {
+        case NonFatal(err) =>
+          recoverFunc(err)
+      })
   }
 
   def execute(logger: Logger, param: C)(
--- a/core/src/main/scala/org/bitcoins/core/api/Callback.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/Callback.scala
@@ -1,0 +1,53 @@
+package org.bitcoins.core.api
+
+import org.bitcoins.core.util.{FutureUtil, SeqWrapper}
+import org.slf4j.Logger
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait Callback[T] {
+  def apply(param: T): Future[Unit]
+}
+
+trait CallbackBinary[T1, T2] extends Callback[(T1, T2)] {
+  def apply(param1: T1, param2: T2): Future[Unit]
+
+  override def apply(param: (T1, T2)): Future[Unit] = apply(param._1, param._2)
+}
+
+trait CallbackTernary[T1, T2, T3] extends Callback[(T1, T2, T3)] {
+  def apply(param1: T1, param2: T2, param3: T3): Future[Unit]
+
+  override def apply(param: (T1, T2, T3)): Future[Unit] =
+    apply(param._1, param._2, param._3)
+}
+
+case class CallbackHandler[C, T <: Callback[C]](
+    name: String,
+    override val wrapped: IndexedSeq[T])
+    extends SeqWrapper[T] {
+
+  def ++(other: CallbackHandler[C, T]): CallbackHandler[C, T] = {
+    require(name == other.name,
+            "Cannot combine callback handlers with different names")
+    CallbackHandler(name, wrapped ++ other.wrapped)
+  }
+
+  def execute(param: C, recoverFunc: Throwable => Unit = _ => ())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    wrapped
+      .foldLeft(FutureUtil.unit)((acc, callback) =>
+        acc.flatMap(_ =>
+          callback(param).recover {
+            case err: Throwable =>
+              recoverFunc(err)
+          }))
+  }
+
+  def execute(logger: Logger, param: C)(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    val recoverFunc = (err: Throwable) =>
+      logger.error(s"$name Callback failed with error: ", err)
+    execute(param, recoverFunc)
+  }
+}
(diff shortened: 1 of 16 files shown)


node/src/main/scala/org/bitcoins/node/NodeCallbacks.scala
```diff
@@ -1,133 +1,164 @@
...
 
   /** Empty callbacks that does nothing with the received data */
   val empty: NodeCallbacks =
-    NodeCallbacks(
-      onTxReceived = Seq.empty,
-      onBlockReceived = Seq.empty,
-      onMerkleBlockReceived = Seq.empty,
-      onCompactFiltersReceived = Seq.empty,
-      onBlockHeadersReceived = Seq.empty
+    apply(Vector.empty, Vector.empty, Vector.empty, Vector.empty, Vector.empty)
```
Christewart: Let's not use `apply()` like this as it is super confusing. Just use the proper name `NodeCallbacks()`

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,53 @@
+package org.bitcoins.core.api
+
+import org.bitcoins.core.util.{FutureUtil, SeqWrapper}
+import org.slf4j.Logger
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait Callback[T] {
```
Christewart, line 10: We need scaladocs on these

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,53 @@
...
+}
+
+trait CallbackTernary[T1, T2, T3] extends Callback[(T1, T2, T3)] {
+  def apply(param1: T1, param2: T2, param3: T3): Future[Unit]
+
+  override def apply(param: (T1, T2, T3)): Future[Unit] =
+    apply(param._1, param._2, param._3)
+}
+
+case class CallbackHandler[C, T <: Callback[C]](
```
Christewart, line 30: scaladoc

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  def ++(other: CallbackHandler[C, T]): CallbackHandler[C, T] = {
+    require(name == other.name,
+            "Cannot combine callback handlers with different names")
+    CallbackHandler(name, wrapped ++ other.wrapped)
+  }
+
+  def execute(param: C, recoverFunc: Throwable => Unit = _ => ())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    wrapped
```
Christewart: Can this be `FutureUtil.foldLeftAsync` ? There maybe we need to add a `FutureUtil` variant that allows you to add `recover`? It seems like a useful utility function

Christewart: Thinking about this more, if we have 3 callbacks `f1` `f2` and `f3` and `f1` fails (or hangs forever), we don't want `f2` and `f3` to fail. We need to use a `foreach()` rather than a `foldLeft` and we probably don't want to accumulate on the callbacks

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,53 @@
...
+      .foldLeft(FutureUtil.unit)((acc, callback) =>
+        acc.flatMap(_ =>
+          callback(param).recover {
+            case err: Throwable =>
```
Christewart: this pattern match case needs to be `scala.util.control.NonFatal`

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,53 @@
...
+  override def apply(param: (T1, T2, T3)): Future[Unit] =
+    apply(param._1, param._2, param._3)
+}
+
+case class CallbackHandler[C, T <: Callback[C]](
+    name: String,
+    override val wrapped: IndexedSeq[T])
+    extends SeqWrapper[T] {
+
+  def ++(other: CallbackHandler[C, T]): CallbackHandler[C, T] = {
```
Christewart, line 35: I'm confusing, we are nesting callback handlers? Don't we just want to add callbacks to callback handlers?

benthecarman: This is just a way to combine both handlers, not nest them

core/src/main/scala/org/bitcoins/core/api/Callback.scala
benthecarman: If that happens `f2` and `f3` will still execute because of the `recover`

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,56 @@
...
+
+/** A function with two parameters to be called in response to an event */
+trait CallbackBinary[T1, T2] extends Callback[(T1, T2)] {
+  def apply(param1: T1, param2: T2): Future[Unit]
+
+  override def apply(param: (T1, T2)): Future[Unit] = apply(param._1, param._2)
+}
+
+/** A function with three parameters to be called in response to an event */
+trait CallbackTernary[T1, T2, T3] extends Callback[(T1, T2, T3)] {
```
rorp: Why wouldn't we name it `Callback3` similar to how it's done in the standard library: https://www.scala-lang.org/api/current/scala/Function3.html ?

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  def ++(other: CallbackHandler[C, T]): CallbackHandler[C, T] = {
+    require(name == other.name,
+            "Cannot combine callback handlers with different names")
+    CallbackHandler(name, wrapped ++ other.wrapped)
+  }
+
+  def execute(param: C, recoverFunc: Throwable => Unit = _ => ())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    FutureUtil.foldLeftAsync((), wrapped)((_, callback) =>
```
Christewart: There is still a problem here with using `foldLeftAsync()`. I'd be curious to here @rorp 's opinion on this:

Let's say we have two callbacks defined as this: 

```scala
  val f1 = new Callback[Int] {
    def apply(param: Int): Future[Unit] = {
      Thread.sleep(1000000) //sleep 1,000 seconds for whatever reason (simulated delay)
      Future.unit
    }
  }
  
  val f2 = new Callback[Int] {
    def apply(param: Int): Future[Unit] = {
      //no delay here, execute right away
      Future.unit
    }
  }
```

We will now wait for 1,000 seconds on `f1` before we start executing `f2`. Is this really the behavior we want?

benthecarman: I see your point now, maybe have a CallbackHandler and a separate SynchronousCallbackHandler

core/src/main/scala/org/bitcoins/core/api/Callback.scala
Christewart: Scaladoc then

core/src/main/scala/org/bitcoins/core/api/Callback.scala
rorp: I'm not really understand why use `Future`s this way. If we want to wait for 1000 seconds we need to put `sleep()` inside a `Future`: 

```scala
  val f1 = new Callback[Int] {
    def apply(param: Int): Future[Unit] = Future {
      Thread.sleep(1000000) //sleep 1,000 seconds for whatever reason (simulated delay)
    }
  }
```

Christewart: I think we should just say there are no execution guarantees when passing in the callbacks. You can bake in ordering if you want, by doing something like this: 

```scala
  val f3 = new Callback[Int] {
    def apply(param: Int): Future[Unit] = {
      //since i care about ordering, i need to enforce ordering in this callback
      //since it's important to call f1 before f2
      //what's even better is you can use your own recover techniques here.
      
      f1(param).flatMap(_ =>  f2(param))
    }
  }
```

rorp: > I think we should just say there are no execution guarantees when passing in the callbacks. 

I still don't understand what's the issue here. `foldLeftAsync()` in fact has execution guarantees -- it executes `Future`s one by one. If one doesn't want the `Future`s to be executed one by one should simply not wrap the code in `Future`s or not return them outside `Callback.apply()`.

I think that's what we should say.

core/src/main/scala/org/bitcoins/core/api/Callback.scala
```diff
@@ -0,0 +1,59 @@
...
+  def execute(param: C, recoverFunc: Throwable => Unit = _ => ())(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    val executeFs = wrapped.map { callback =>
+      callback(param).recover {
+        case NonFatal(err) =>
+          recoverFunc(err)
+      }
+    }
+
+    Future.sequence(executeFs).map(_ => ())
```
rorp, line 55: What do you mean by `execute syncronously`? `Future.sequence` executes futures in parallel asynchronously unlike `FutureUtil.sequentially` or `FutureUtil.foldLeftAsync`.

Christewart (approved):

rorp (approved):

PR #1543, opened 2020-06-10
Wallet callbacks

### LINKED ISSUE #1532: Wallet callbacks

benthecarman:
Built on top of #1542

Closes #1532 

Adds 4 types of callbacks:

- OnTransactionProcessed: when we process a tx

- OnTransactionBroadcast: when we broadcast a tx

- OnReservedUtxos: When a utxo becomes reserved or unreserved (might need a better name)

- OnNewAddressGenerated: when we generate a new address that did not exist in our database before

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
@@ -2,7 +2,6 @@ package org.bitcoins.wallet
 
 import org.bitcoins.core.api.{Callback, CallbackHandler}
 import org.bitcoins.core.protocol.transaction.Transaction
-import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.wallet.models.{AddressDb, SpendingInfoDb}
 import org.slf4j.Logger
 
@@ -77,9 +76,6 @@ object WalletCallbacks {
       onNewAddressGenerated = onNewAddressGenerated ++ other.onNewAddressGenerated
     )
   }
-
-  /** Does nothing */
-  def noop[T]: T => Future[Unit] = _ => FutureUtil.unit
 
   /** Constructs a set of callbacks that only acts on processed transaction */
   def onTransactionProcessed(f: OnTransactionProcessed): WalletCallbacks =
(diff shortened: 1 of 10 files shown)


wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
```diff
@@ -0,0 +1,128 @@
...
+    override def +(other: WalletCallbacks): WalletCallbacks = copy(
+      onTransactionProcessed = onTransactionProcessed ++ other.onTransactionProcessed,
+      onTransactionBroadcast = onTransactionBroadcast ++ other.onTransactionBroadcast,
+      onReservedUtxos = onReservedUtxos ++ other.onReservedUtxos,
+      onNewAddressGenerated = onNewAddressGenerated ++ other.onNewAddressGenerated
+    )
+  }
+
+  /** Does nothing */
+  def noop[T]: T => Future[Unit] = _ => FutureUtil.unit
```
Christewart: Isn't this defined elsewhere?

wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
```diff
@@ -0,0 +1,128 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * Callbacks for responding to events in the wallet.
+  * The appropriate callback is executed whenever the wallet finishes,
+  * the corresponding function.
+  */
+trait WalletCallbacks {
+
+  def onTransactionProcessed: CallbackHandler[
```
Christewart, line 18: Is there a reason these are `Transaction` and the other things are `AddressDb` and `SpendingInfoDb` (two callbacks is `Db` the others are not)

Christewart (commented):
We need to add a section to the website for wallet callbacks like this with node: https://bitcoin-s.org/docs/next/node/node#callbacks

wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
benthecarman: It's defined in `NodeCallbacks` I could probably move it to `Callbacks` so they can both use it

wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
benthecarman: Not particularly, we could give the `TransactionDb` but it doesn't have any information you can't calculate from just the `Transaction`

Christewart: I think we should be consistent with whatever we do, either `Db` or no `Db`.

PR #1552, opened 2020-06-12
Fixed all walletTest threads not being closed

nkohen:


benthecarman (approved):

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -439,6 +440,7 @@ object BitcoinSWalletTest extends WalletLogger {
       wallet <- BitcoinSWalletTest.createWallet2Accounts(bitcoind,
                                                          bitcoind,
                                                          extraConfig)
+      _ = wallet.stopWalletThread()
```
Christewart, line 443: Why is this here? Shouldn't this be in `destroyWallet`?

benthecarman: We create the wallet again so it can have callbacks after this, so this wallet will never passed to destroy wallet

Christewart: Ugh, so it seems this thread management should really happen in `WalletAppConfig` as well then? It seems like it is _so_ easy to forget to do this when a wallet is destroyed

PR #1555, opened 2020-06-15
Make it so wallet fixtures take a bip39PasswordOpt as a paramter

Christewart:
Previously we would create a bip39 password internally inside of [`createDefaultWallet()`](https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2020-06-14-walletfixtures-bip39?expand=1#diff-0fb6ac004fe1e550b7c13258d7d0706cL408). The problem with this approach is the caller of the fixture has no idea what the password is, so it shouldn't be able to spend from the wallet.  

As far as I can tell, this hasn't mattered yet as the `BIP39KeyManager` seems to retain the password and doesn't ask for it to served by the wallet for spending from an address. We need to look into this further to understand why this is the case.

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
@@ -6,8 +6,8 @@ import org.bitcoins.core.protocol.script.EmptyScriptPubKey
 import org.bitcoins.core.protocol.transaction.TransactionOutput
 import org.bitcoins.core.wallet.fee.{SatoshisPerByte, SatoshisPerVirtualByte}
 import org.bitcoins.core.wallet.utxo.TxoState
-import org.bitcoins.testkit.wallet.BitcoinSWalletTest
-import org.bitcoins.testkit.wallet.BitcoinSWalletTest.{
+import org.bitcoins.testkit.wallet.{
+  BitcoinSWalletTest,
   WalletWithBitcoind,
   WalletWithBitcoindRpc
 }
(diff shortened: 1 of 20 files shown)


benthecarman (commented):
Needs `scalafmt`

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
```diff
@@ -24,7 +24,7 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
       .fromString("bcrt1qlhctylgvdsvaanv539rg7hyn0sjkdm23y70kgq")
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    withFundedWalletAndBitcoind(test)
+    withFundedWalletAndBitcoind(test, getBIP39PasswordOpt())
```
benthecarman, line 27: What is the use case for being able to give a specific bip39 password for a certain test? It would make more sense to just call `getBIP39PasswordOpt` in the fixture (in this case `withFundedWalletAndBitcoind`)

Christewart: The problem is your test case can't access it then if it is generated inside your fixture. I thought about making it a default argument for the methods 

```scala
def withFundedWalletAndBitcoind(test: OneArgAsyncTest, bip39PasswordOpt: Option[String] = getBIP39PasswordOpt()) = ...
```

but I think it is too easy to forget to pass in the bip39 password when you need it.

benthecarman (approved):

PR #1556, opened 2020-06-15
2020 06 14 wallet root accounts

Christewart:
This is built on top of #1555 

If [`createRootAccount()`](https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2020-06-14-wallet-root-accounts?expand=1#diff-fe3b2d50794290699b4ac0929ec80201L446) is called twice, and the `xpub`s are the same, we currently throw an exception since we are using `accountDAO.create()`. This is not ideal as when starting up the wallet we may want to create IFF the account does not exist.

This PR now checks the `xpub` in the database and see's if it matches the xpub we are attempting to generate. If the `xpub`s are the same, we simply return the account in the db already rather than throwing an exception. 

If the `xpub`s are different, we return an exception indicating we are probably using a different key-manager rather than the one that was used for entries in the `AccountTable`. In the future, we probably need a way to indicate what account corresponds to what key manager (say, the master private key fingerprint for BIP32).

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
@@ -20,11 +20,12 @@ import org.scalatest.compatible.Assertion
 import scala.concurrent.Future
 
 class WalletUnitTest extends BitcoinSWalletTest {
+  private val bip39PasswordOpt: Option[String] = getBIP39PasswordOpt()
 
   override type FixtureParam = Wallet
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome =
-    withNewWallet(test, getBIP39PasswordOpt())
+    withNewWallet(test, bip39PasswordOpt)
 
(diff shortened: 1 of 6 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
```diff
@@ -166,4 +167,26 @@ class WalletUnitTest extends BitcoinSWalletTest {
...
+  }
+
+  it must "be able to detect an incompatible key manager with a wallet" in {
+    wallet: Wallet =>
+      recoverToSucceededIf[RuntimeException] {
+        Wallet.initialize(wallet, bip39PasswordOpt).flatMap { _ =>
+          //use a BIP39 password to make the key-managers different
+          Wallet.initialize(
+            wallet,
+            Some("random-password-to-make-key-managers-different"))
```
benthecarman, line 188: nit: Could do something like `Some(bip39PasswordOpt.getOrElse("") ++ "extra")` to assure we can't get same password

PR #1560, opened 2020-06-15
Rework signing logic to take full funding transaction

### LINKED ISSUE #1530: Rework signing logic to take full funding transaction and hash it before signing

benthecarman:
Built on top of #1540 

Fixes #1530


core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxSignerTest.scala
```diff
@@ -110,6 +80,7 @@ class RawTxSignerTest extends BitcoinSAsyncTest {
     val utxo = ScriptSignatureParams(
       InputInfo(
         outPoint = outPoint,
+        creditingTx,
```
nkohen: nit: Named arg for this and the ones below

core-test/src/test/scala/org/bitcoins/core/wallet/builder/StandardNonInteractiveFinalizerTest.scala
```diff
@@ -113,6 +114,7 @@ class StandardNonInteractiveFinalizerTest extends BitcoinSAsyncTest {
     val utxo = ScriptSignatureParams(
       InputInfo(
         outPoint = outPoint,
+        creditingTx,
```
nkohen: ditto

core-test/src/test/scala/org/bitcoins/core/wallet/utxo/InputInfoTest.scala
```diff
@@ -34,6 +28,26 @@ class InputInfoTest extends BitcoinSAsyncTest {
 
   behavior of "InputInfo"
 
+  it must "fail given no a prevTransaction that doesn't match the outPoint" in {
```
nkohen: nit: grammar

core-test/src/test/scala/org/bitcoins/core/wallet/utxo/InputSigningInfoTest.scala
```diff
@@ -36,6 +30,7 @@ class InputSigningInfoTest extends BitcoinSUnitTest {
     val outPoint = TransactionOutPoint(creditingTx.txId, UInt32.zero)
     val inputInfo = InputInfo(
       outPoint = outPoint,
+      creditingTx,
```
nkohen: ditto

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -834,12 +829,12 @@ object PSBT extends Factory[PSBT] {
...
 
   /**
-    * Wraps a Vector of pairs of NewSpendingInfos and the Transactions whose outputs are spent.
+    * Wraps a Vector of pairs of ScriptSignatureParams and the Transactions whose outputs are spent.
     * Note that this Transaction is only necessary when the output is non-segwit.
     */
   case class SpendingInfoAndNonWitnessTxs(
       infoAndTxOpts: Vector[
-        (ScriptSignatureParams[InputInfo], Option[BaseTransaction])]) {
+        (ScriptSignatureParams[InputInfo], Option[Transaction])]) {
```
nkohen: why the type change?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -852,7 +847,7 @@ object PSBT extends Factory[PSBT] {
 
     def map[T](func: (
         ScriptSignatureParams[InputInfo],
-        Option[BaseTransaction]) => T): Vector[T] = {
+        Option[Transaction]) => T): Vector[T] = {
```

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -911,6 +906,10 @@ object PSBT extends Factory[PSBT] {
     val outputMaps = unsignedTx.outputs.map(_ => OutputPSBTMap.empty).toVector
 
     Future.sequence(inputMapFs).map { inputMaps =>
+      require(inputMaps.zip(unsignedTx.inputs).forall {
+        case (map, txIn) =>
+          map.nonWitnessOrUnknownUTXOOpt.get.transactionSpent.txId == txIn.previousOutput.txId
+      })
```
nkohen: Should this be covered in PSBT maybe?

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -20,6 +16,13 @@ import org.bitcoins.crypto.{ECPublicKey, NetworkElement, Sign}
 sealed trait InputInfo {
   def outPoint: TransactionOutPoint
 
+  def prevTransaction: Transaction
```
nkohen: I think that this belongs in `InputSigningInfo` rather than here because the previous transaction should not concern a party that isn't signing for this UTXO

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -223,34 +243,40 @@ object RawInputInfo {
...
           s"Currently unsupported ScriptPubKey $scriptPubKey")
     }
   }
 }
 
-case class EmptyInputInfo(outPoint: TransactionOutPoint, amount: CurrencyUnit)
+case class EmptyInputInfo(
+    outPoint: TransactionOutPoint,
+    prevTransaction: Transaction,
+    amount: CurrencyUnit)
```
nkohen: Since `EmptyInputInfo` is already only for tests you should just override `prevTransaction` to be `EmptyTransaction` in the body rather than taking in a param that could conflict with `outPoint`

nkohen: Also, this becomes somewhat irrelevant once you look at my comment above so :man_shrugging:

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -102,13 +96,18 @@ trait FundTransactionHandling extends WalletLogger { self: WalletApi =>
...
+    val addrInfosWithUtxoF: Future[
+      Vector[(SpendingInfoDb, Transaction, AddressInfo)]] =
       for {
         selectedUtxos <- selectedUtxosF
-        addrInfoOptF = selectedUtxos.map { utxo =>
-          val addrInfoOptF = getAddressInfo(utxo)
-          //.get should be safe here because of foreign key at the database level
-          addrInfoOptF.map(addrInfoOpt => (utxo, addrInfoOpt.get))
+        prevTxFs = selectedUtxos.map(utxo =>
+          transactionDAO.findByOutPoint(utxo.outPoint).map(_.get.transaction))
```
nkohen: nit: could have some better error handling and readability

nkohen (commented):
My main comment is that `prevTransaction` does not belong in `InputInfo`

nkohen:
As per our discussion on slack, `prevTransaction == EmptyTransaction ||` will be needed in the outpoint invariant when moving `prevTransaction` from `InputInfo` to `InputSigningInfo`

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: because now witness txs need the previous tx

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: because now witness txs need the previous tx

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
nkohen: Ah I see, in that case the name of the case class is going to have to change too :P

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -573,35 +562,41 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
-    ECSignatureParams(inputInfo, signer, hashType)
+    ECSignatureParams(inputInfo, txVec.head.transactionSpent, signer, hashType)
   }
 
   private def changeToWitnessUTXO(
       transactionOutput: TransactionOutput): InputPSBTMap = {
     val newElements = transactionOutput.scriptPubKey match {
-      case _: WitnessScriptPubKey =>
+      case rs: WitnessScriptPubKey if !rs.isInstanceOf[WitnessScriptPubKeyV0] =>
         filterRecords(NonWitnessUTXOKeyId) :+ WitnessUTXO(transactionOutput)
```
nkohen: This exhaustive-check-breaking `if` clause can be avoided by simply moving this case beneath the case that checks for `WitnessScriptPubKeyV0`

Christewart (commented):
So here is my concern with how this PR is designed: 

The place that signature generation actually occurs is [`TransactionSignatureCreator.createSig()`](https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureCreator.scala#L39-L75). It's called in [`SignerUtils.doSign()`](https://github.com/bitcoin-s/bitcoin-s/blob/6ca09c714d7fdb777b986ed8d0746af8ebda1acd/core/src/main/scala/org/bitcoins/core/wallet/signer/Signer.scala#L35) here. 

There doesn't seem to be any checks added inside of `TransactionSignatureCreator.createSig()`. Moreover, it seems the `TxSigComponent` class does not add the funding transaction to the data structure. There are a couple thoughts I have here

1. is `TxSigComponent` still the right parameter to `TransactionSignatureCreator.createSig()`?
2. We should executes security checks for this attack at the signing level, if we don't refactor `TransactionSignatureCreator.createSig()`, we need to execute this check right before signing so we are _sure_ that it is still correct.

Looking at the scaladoc on `TxSigComponent`, here is what I wrote back in the day: 

>Represents a transaction whose input is being checked against the spending conditions of a

So it sounds like i intended this to be used for script verification, and not signature creation (although it can be used for both). 

Thoughts @nkohen ?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -582,26 +571,32 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+          }
         } else {
           elements
         }
-      case _: P2PKHScriptPubKey | _: P2PKScriptPubKey |
-          _: P2PKWithTimeoutScriptPubKey | _: MultiSignatureScriptPubKey |
-          EmptyScriptPubKey | _: LockTimeScriptPubKey |
-          _: NonStandardScriptPubKey | _: WitnessCommitment |
-          _: ConditionalScriptPubKey =>
+      case _: WitnessScriptPubKeyV0 | _: P2PKHScriptPubKey |
```
nkohen, line 655: This new case will never be reached (`WitnessScriptPubKey` case needs to move down)

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -816,12 +773,12 @@ object PSBT extends Factory[PSBT] {
   }
 
   /**
-    * Wraps a Vector of pairs of NewSpendingInfos and the Transactions whose outputs are spent.
-    * Note that this Transaction is only necessary when the output is non-segwit.
+    * Wraps a Vector of pairs of ScriptSignatureParams and the Transactions whose outputs are spent.
+    * Note that this Transaction is only necessary when the output if it is pre-segwitV1.
     */
-  case class SpendingInfoAndNonWitnessTxs(
+  case class SpendingInfoAndPrevTransactionTxs(
```
nkohen: nit: `SpendingInfoAndPrevTxs`?

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -671,6 +721,103 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
     }
     ScriptInterpreter.runAllVerify(programs)
   }
+
+  def verifyPSBTInputScript(
```
nkohen, line 718: can you not use the TxSigComponent method that was added for any of this?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/CreditingTxGen.scala
```diff
@@ -59,6 +59,27 @@ sealed abstract class CreditingTxGen {
               conditionalOutput)
   }
 
+  def nonWitOutput: Gen[ScriptSignatureParams[InputInfo]] = {
```
nkohen: do you mean raw output? (no p2sh)

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
benthecarman: Possibly but will require a bunch of work to get a valid input info from a signed tx + input map. Maybe better for a separate PR.

core-test/src/test/scala/org/bitcoins/core/crypto/TransactionSignatureSerializerTest.scala
```diff
@@ -402,4 +405,117 @@ class TransactionSignatureSerializerTest extends BitcoinSAsyncTest {
...
+    }
+  }
+
+  it should "have old and new calculateScriptForSigning functions agree" in {
+    forAllAsync(CreditingTxGen.inputsAndOutputs(),
+                ScriptGenerators.scriptPubKey) {
+      case ((creditingTxsInfo, destinations), (changeSPK, _)) =>
+        val fee = SatoshisPerVirtualByte(Satoshis(100))
+
+        StandardNonInteractiveFinalizer
```
Christewart: Please use intermediate `val` rather than just chaining maps, it's much more readable. 

```scala
        val unsignedTxF = StandardNonInteractiveFinalizer
          .txFrom(outputs = destinations,
                  utxos = creditingTxsInfo,
                  feeRate = fee,
                  changeSPK = changeSPK)
          
        val correctScriptsF = unsignedTxF.map { spendingTx =>
          creditingTxsInfo.flatMap { signInfo =>
            signInfo.signers.map { _ =>
              val txSigComponent =
                TxSigComponent(signInfo.inputInfo, spendingTx)

              val oldScript =
                BitcoinScriptUtil.calculateScriptForSigning(
                  txSigComponent,
                  txSigComponent.output.scriptPubKey.asm)

              val newScript =
                BitcoinScriptUtil.calculateScriptForSigning(
                  spendingTx,
                  signInfo,
                  signInfo.output.scriptPubKey.asm)

              oldScript == newScript
            }
          }
        }

        correctScriptsF.map(x => assert(x.forall(_ == true)))
```

it allows you to also give names to the intermediate `val` to express to the reader of your code what your intention is (`correctScripts`, `unsignedTxF` etc).

This pattern applies a few other places in this PR.

core-test/src/test/scala/org/bitcoins/core/psbt/PSBTUnitTest.scala
```diff
@@ -250,31 +256,38 @@ class PSBTUnitTest extends BitcoinSAsyncTest {
...
     assert(spendingInfo.hashType == HashType.sigHashAll)
-    assert(InputInfo.getRedeemScript(spendingInfo.inputInfo).isEmpty)
+    assert(InputInfo.getRedeemScript(spendingInfo.inputInfo).isDefined)
     assert(
       InputInfo
-        .getScriptWitness(spendingInfo.inputInfo)
-        .contains(P2WSHWitnessV0(RawScriptPubKey.fromAsmHex(
-          "5221029da12cdb5b235692b91536afefe5c91c3ab9473d8e43b533836ab456299c88712103372b34234ed7cf9c1fea5d05d441557927be9542b162eb02e1ab2ce80224c00b52ae"))))
+        .getRedeemScript(spendingInfo.inputInfo)
+        .get == MultiSignatureScriptPubKey(
```
Christewart: intermediate `val`s are your friend, we aren't trying to minimize LOC here, readability over terseness.

core-test/src/test/scala/org/bitcoins/core/wallet/signer/SignerTest.scala
```diff
@@ -186,6 +155,46 @@ class SignerTest extends BitcoinSAsyncTest {
     }
   }
 
+  it should "have old and new createSig functions agree" in {
```
Christewart: this isn't testing `createSig()` actually? It seems to be testing the deprecated version of `doSign()` vs the new version of `doSign()`. It may indirectly test it, but we aren't explicitly calling it so let's name the test case appropriately.

Christewart: Also please use intermediate `val`s to make this code more readable (see comment above)

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -253,6 +254,233 @@ sealed abstract class TransactionSignatureSerializer {
...
+    * [[https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki BIP143]].
+    * [[https://github.com/bitcoin/bitcoin/blob/f8528134fc188abc5c7175a19680206964a8fade/src/script/interpreter.cpp#L1113]]
+    */
+  def serializeForSignature(
+      spendingTransaction: Transaction,
+      signingInfo: InputSigningInfo[InputInfo],
+      hashType: HashType): ByteVector = {
+    val idx = spendingTransaction.inputs.zipWithIndex
+      .find(_._1.previousOutput == signingInfo.outPoint)
+      .get
```
Christewart: can we format a nice error message when the outpoint doesn't exist rather than just a `NoSuchElementException? For instance, it would be nice to have the outpoint that _doesn't_ exist in the transaction

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -44,6 +46,62 @@ sealed abstract class TxSigComponent {
 
 object TxSigComponent {
 
+  private def inputIndex(inputInfo: InputInfo, tx: Transaction): UInt32 = {
+    tx.inputs.zipWithIndex
+      .find(_._1.previousOutput == inputInfo.outPoint) match {
+      case Some((_, index)) => UInt32(index)
+      case None =>
+        throw new IllegalArgumentException(
+          "Transaction did not contain expected input.")
```
Christewart: put the outpoint in the error message, let's try to give useful error messages with input given by the user :-)

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -44,6 +46,62 @@ sealed abstract class TxSigComponent {
...
+    }
+  }
+
+  def apply(
```
Christewart, line 49: Why don't we just get away from the carte-blanche `apply()` methods? 

Can't we do something like 

`TxSigComponent.fromWitnessTx()` (i'm guessing this case we can get a more specific type on `InputInfo`)

`TxSigComponent.fromBaseTransaction()` (maybe we can type this with a `RawInputInfo`? 

and then `TxSigComponent.fromBaseTransactionP2SH()` (not sure how this one would look like)

We have some really nasty pattern matching going on (using `_: ScriptPubKey`, which is equivalent to just punting ADTs) which is needed because we aren't narrowing our scope with the parameters to these methods. 

At minimum, (if we don't take my suggestion above) we need to enumerate all of our SPK types instead of using `case _: ScriptPubKey` so we retain pattern match exhaustiveness when we add new scriptpubkey types (say, `WitnessScriptPubKeyV1`)

core/src/main/scala/org/bitcoins/core/protocol/transaction/TxUtil.scala
```diff
@@ -322,4 +304,26 @@ object TxUtil {
       Success(())
     }
   }
+
+  def addWitnessData(
```
Christewart, line 310: scaladoc

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -203,8 +204,12 @@ case class PSBT(
           inputMap.redeemScriptOpt.isDefined && WitnessScriptPubKey
             .isWitnessScriptPubKey(
               inputMap.redeemScriptOpt.get.redeemScript.asm)
+        val notBIP143Vulnerable =
```
Christewart, line 207: is this right? If i understand this, it applies to every script that exists in the bitcoin ecosystem no?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
Christewart: Why is transaction an `Option[]` here? Also it might be better to define this type this way: 

```scala
case class SpendingInfoAndPrevTx(scriptSigParams: ScriptSignatureParams[InputInfo], txOpt: Option[Transaction])
```
and then you can just use `Vector[SpendingInfoAndPrevTx]` everywhere

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -573,35 +562,41 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
-        filterRecords(NonWitnessUTXOKeyId) :+ WitnessUTXO(transactionOutput)
       case _: P2SHScriptPubKey =>
-        if (
-          redeemScriptOpt.isDefined && redeemScriptOpt.get.redeemScript
-            .isInstanceOf[WitnessScriptPubKey]
-        ) {
-          filterRecords(NonWitnessUTXOKeyId) :+ WitnessUTXO(transactionOutput)
+        if (redeemScriptOpt.isDefined) {
+          val redeemScript = redeemScriptOpt.get.redeemScript
+          if (
```
Christewart, line 644: what is this doing? a comment would be nice

core/src/main/scala/org/bitcoins/core/psbt/PSBTRecord.scala
```diff
@@ -157,6 +151,11 @@ object InputPSBTRecord extends Factory[InputPSBTRecord] {
 
   case class WitnessUTXO(witnessUTXO: TransactionOutput)
```
Christewart, line 152: So is this defined in BIP174? and is it inherently insecure? If so should we even support it?

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -482,6 +484,54 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
         script
     }
 
+  def calculateScriptForSigning(
+      spendingTransaction: Transaction,
+      signingInfo: InputSigningInfo[InputInfo],
+      script: Seq[ScriptToken]): Seq[ScriptToken] = {
+    // .get should be safe
+    val (_, idx) = spendingTransaction.inputs.zipWithIndex
```
Christewart: Would should really define this utility method somewhere :P

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
Christewart: I'm with Nadav, it's pretty hard to read this. It would be useful to define helper methods to delegate to and also be more specific types in the parameters so it doesn't have to be monolithic. Simple helper methods like

```
  private def verifyWitnessPSBTInputScript(
      wtx: WitnessTransaction,
      inputPSBTMap: InputPSBTMap,
      index: Int,
      flags: Seq[ScriptFlag] = Policy.standardFlags): Try[Transaction] = {
      //..handle witness inputs here
  }
```

go a long way for improving readability. If you want to address it on a separate PR i'm fine with that, but file an issue please

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
```diff
@@ -15,9 +16,24 @@ import org.bitcoins.crypto.Sign
   */
 sealed trait InputSigningInfo[+InputType <: InputInfo] {
   def inputInfo: InputType
+  def prevTransaction: Transaction
   def hashType: HashType
   def signers: Vector[Sign]
 
+  // If using EmptyTransaction we are testing or dummy signing
+  require(
+    prevTransaction == EmptyTransaction || outPoint.txId == prevTransaction.txId,
```
Christewart, line 25: is there a way the `prevTransaction == EmptyTransaction` invariant can be used to exploit the signature serialization scheme?

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -253,6 +254,233 @@ sealed abstract class TransactionSignatureSerializer {
...
+    */
+  def hashForSignature(
+      spendingTransaction: Transaction,
+      signingInfo: InputSigningInfo[InputInfo],
+      hashType: HashType): DoubleSha256Digest = {
+    val inputIndexOpt = spendingTransaction.inputs.zipWithIndex
+      .find(_._1.previousOutput == signingInfo.outPoint)
+      .map(_._2)
+
+    if (inputIndexOpt.isEmpty) {
```
Christewart, line 462: I would prefer to move this into `serializeForSignature()` as it can be used indepdently of `hashForSignature()`. Basically this check should be moved to the lowest common denominator for signing things (serialization for signature, not hashing).

Christewart (changes requested):
A couple higher level thoughts:

Do we want to use `prevTransaction` of `fundingTransaction` ? It seems the variable name `fundingTransaction` is what is used by offchain protocols, it would make sense in my opinion to be consistent with that

It also seems like we aren't really using the type parameter for `InputSigningInfo[]` any where besides Signer.scala . Is this intentional? We basically do `InputSigningInfo[InputInfo]` everywhere which kind of defeats the point of the type parameter. 

Screenshot from 2020-06-24 07-33-08

core/src/main/scala/org/bitcoins/core/psbt/PSBTRecord.scala
benthecarman: Yeah it will still be included in BIP174 so I think it's best to continue to support it.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: It is for the following if statement, we check if it is a witness output, then we also need to check if it's V0. So if it's non-segwit or SegwitV0, we use full tx, otherwise we use just the output (segwitV1+)

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
benthecarman: This needs to stay here unless we know the preimage to `errorHash`

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
benthecarman: Did this but by type of input info rather than tx. Reason being we could have a witness tx, but the input we are checking is not a segwit input.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: We use `Option` here because some won't need txs (future segwit versions). We could remove the `Option` until then but I don't see the point in that as we will just have to redo the work again.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
benthecarman: I added
 ```
    require(
      signingInfo.prevTransaction != EmptyTransaction,
      "prevTransaction can only be an EmptyTransaction when dummy signing")
```
to `serializeForSignature`, this way an error will be thrown if we are doing a real signature, not a dummy signature or one that expects an error hash.

File: core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala [hunk 29]
```diff
@@ -44,6 +46,130 @@ sealed abstract class TxSigComponent {
...
+          unsignedWtx
+      }
+    }
+
+    WitnessTxSigComponent(wtx, UInt32(idx), inputInfo.output, flags)
+  }
+
+  def fromWitnessInput(
+      inputInfo: UnassignedSegwitNativeInputInfo,
+      unsignedTx: Transaction,
```
Christewart, line 97: Doesn't this have to be a `WitnessTransaction`?

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -44,6 +46,130 @@ sealed abstract class TxSigComponent {
...
+          noWitnessWtx.updateWitness(idx, scriptWitness)
+      }
+    }
+
+    WitnessTxSigComponent(wtx, UInt32(idx), inputInfo.output, flags)
+  }
+
+  def fromWitnessInput(
+      inputInfo: P2SHNestedSegwitV0InputInfo,
+      unsignedTx: Transaction,
```
Christewart, line 107: same here

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
Christewart: I guess my personal preference in these situations is to avoid using the `Option[]` type as it further complicates things, rather use an ADT to represent the different states we can have (getting rid of the `Option`!) and then in the future we can extend the type ADT to include segwit v1

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
Christewart: I think you can now use `TxUtil.inputIndex()`?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -528,13 +535,54 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
...
+    val witVec = getRecords(WitnessUTXOKeyId)
+    val txVec = getRecords(NonWitnessUTXOKeyId)
+
+    val output = if (witVec.size == 1) {
+      witVec.head.witnessUTXO
+    } else if (txVec.size == 1) {
+      val tx = txVec.head.transactionSpent
+      tx.outputs(txIn.previousOutput.vout.toInt)
+    } else {
+      throw new UnsupportedOperationException(
```
nkohen: Should probably be `IllegalArgument` or just `Runtime` exception

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -545,22 +593,35 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
           .isInstanceOf[P2WPKHWitnessSPKV0] || redeemScriptOpt.exists(
           _.isInstanceOf[P2WPKHWitnessSPKV0])
       ) {
-        Some(P2WPKHWitnessV0(signer.publicKey))
+        require(preImages.size == 1 && preImages.head.isInstanceOf[ECPublicKey],
+                "P2WPKHWitnessV0 must have it's public key as a pre-image")
+        Some(P2WPKHWitnessV0(preImages.head.asInstanceOf[ECPublicKey]))
```
nkohen: A `match` on the head would probably be more idiomatic

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -110,6 +112,53 @@ object InputInfo {
...
+  def getHashPreImagesAndConditionalPath(
+      signedTransaction: Transaction,
+      inputIndex: Int): (Vector[NetworkElement], ConditionalPath) = {
+
+    val txIn = signedTransaction.inputs(inputIndex)
+
+    @tailrec
+    def getPreImagesAndCondPath(
+        scriptSignature: ScriptSignature,
+        prevCond: ConditionalPath = ConditionalPath.NoCondition): (
```
nkohen: More efficient to take a `Vector[Boolean]` (default to `empty`) and only convert back outside (rather than doing in and out of the type on every iteration)

nkohen (commented):
Just a couple nits

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
benthecarman: No, because sometimes we'll sometimes have unsigned txs that are considered `BaseTransaction`s. This shouldn't cause any problems because in the these functions we always make it a `WitnessTransaction` and set the witness data if needed.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
benthecarman: Was able to just remove this all together actually. Since our `ScriptSignatureParams` have a prexTx on them we don't need this. Cleaned a lot of things up :)

PR #1561, opened 2020-06-15
Have BroadcastTransactionTest rebroadcast on failure

benthecarman:
Changed the test so it tries to broadcast twice if the first attempt fails

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
@@ -70,7 +70,18 @@ class BroadcastTransactionTest extends NodeUnitTest {
       bitcoindBalancePreBroadcast <- balanceF
       _ <- node.broadcastTransaction(tx)
       _ <-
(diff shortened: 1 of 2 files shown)


node-test/src/test/scala/org/bitcoins/node/BroadcastTransactionTest.scala
```diff
@@ -63,8 +63,18 @@ class BroadcastTransactionTest extends NodeUnitTest {
 
       bitcoindBalancePreBroadcast <- balanceF
       _ <- node.broadcastTransaction(tx)
-      _ <- TestAsyncUtil.awaitConditionF(() => hasSeenTx(tx),
-                                         duration = 1.second)
+      _ <- TestAsyncUtil
```
Christewart: So can this cause other test cases to fail if we can't reliably broadcast transactions to bitcoind?

benthecarman: Yes, but on most other test cases we just broadcast the tx from the bitcoind node so we don't need to worry about that.

Issue #1559, opened 2020-06-15
Wallet Thread Management Consolidation


nkohen:
Currently, every wallet has an address queue handling thread and a slick database thread pool. It is important to kill these threads when stopping a wallet. Currently, the slick thread pool is handled by the `WalletAppConfig` while the address queue thread is handled directly by the Wallet.

In particular, if someone wishes to use a `Wallet` instance to create another, they must ensure that the first `Wallet`'s address thread gets killed. @Christewart [mentioned](https://github.com/bitcoin-s/bitcoin-s/pull/1552#pullrequestreview-430171205) that perhaps we should consider moving this thread up to `WalletAppConfig` too. Regardless of what is done though, it is certainly a good idea to move this thread handling code up to a higher level (not a per-Wallet instance level)

PR #1562, opened 2020-06-15
Rescan when UTXOs don't have associated transactions

### LINKED ISSUE #1275: Handling of adding new foreign keys to the SpendingInfo Table
In #1197 we are adding a foreign key to the SpendingInfo table that will cause old spending infos to be invalid for not having a record for it's foreign key. The best way to handle this is would be to do a rescan on start up if there is a foreign key error on the table. However, a rescan can be expensive and time consuming. If you have other ideas how best ways to handle this and similar issues in the future, please share.

### LINKED ISSUE #1558: Have Wallet rescan if there exists SpendingInfoDbs with no corresponding entry in the TransactionDb
Once we have a fix for #1530 the funding tx for each input will be needed to sign each input. If there is an older wallet with UTXOs that existed before the tx database they would unable to spend it, therefore we should rescan until we have all the funding txs for all of our spendable utxos

benthecarman:
Fixes #1558
Fixes #1275

Downloading the block should be sufficient here because we will process every tx in the block and any tx relevant to us should be added to our transaction database.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -111,6 +111,7 @@ abstract class Wallet
     } yield {
       stopWalletThread()
     }
+    ()
   }
 
   override def broadcastTransaction(transaction: Transaction): Future[Unit] =
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -88,12 +88,20 @@ abstract class Wallet
     } yield hasTxs
   }
 
-  for {
-    utxos <- utxosWithMissingTx
-    blockHashes = utxos.flatMap(_.blockHash.map(_.flip))
-    // Download the block the tx is from so we process the block and subsequent txs
-    _ <- nodeApi.downloadBlocks(blockHashes.distinct)
-  } yield ()
+  protected def downloadMissingUtxos: Future[Unit] =
+    for {
+      utxos <- utxosWithMissingTx
+      blockHashes = utxos.flatMap(_.blockHash.map(_.flip))
+      // Download the block the tx is from so we process the block and subsequent txs
+      _ <- nodeApi.downloadBlocks(blockHashes.distinct)
+    } yield ()
+
+  override def start(): Future[Unit] = {
+    for {
+      _ <- walletConfig.start()
+      _ <- downloadMissingUtxos
+    } yield ()
+  }
 
   override def stop(): Unit = {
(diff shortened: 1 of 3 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -64,6 +64,26 @@ abstract class Wallet
...
+      }
+    } yield hasTxs
+  }
+
+  for {
+    utxos <- utxosWithMissingTx
+    blockHashes = utxos.flatMap(_.blockHash.map(_.flip))
+    // Download the block the tx is from so we process the block and subsequent txs
+    _ <- nodeApi.downloadBlocks(blockHashes.distinct)
+  } yield ()
```
rorp: I think it should be in a separate initialization method (`start`, `init`, etc).

rorp (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: That would be nice, however, the wallet does not have a `start` function, and `init` is only called when we are creating a new wallet.

Christewart:
Now that #1598 is in, can we move this stuff into the `WalletAppConfig.start()`? I don't think this should be in `Wallet.scala`, are we keeping this around for the long term?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -75,6 +75,34 @@ abstract class Wallet
...
+
+  protected def downloadMissingUtxos: Future[Unit] =
+    for {
+      utxos <- utxosWithMissingTx
+      blockHashes = utxos.flatMap(_.blockHash.map(_.flip))
+      // Download the block the tx is from so we process the block and subsequent txs
+      _ <- nodeApi.downloadBlocks(blockHashes.distinct)
+    } yield ()
+
+  override def start(): Future[Unit] = {
```
Christewart, line 99: This isn't what I had in mind, can't this go into the `WalletAppConfig`?

benthecarman: No, because the `WalletAppConfig` doesn't have access to any of the DAOs or `nodeApi`

Christewart: :cry: , I guess if we do merge this though we can move the Thread inside of `AddressHandling.scala` into `Wallet.scala` and create/destroy it in the start/stop methods?

Christewart:
Also now fixes #1559

PR #1563, opened 2020-06-15
Added compile scope scalafmtCheck to travis runs

nkohen:
This was broken on #1023 where it was replaced with `test:scalafmtCheck` ... which only runs for test scope, both commands are needed


benthecarman (approved):

Christewart:
Needs rebase

PR #1575, opened 2020-06-16
Added DLC json data structures

nkohen:


benthecarman (approved):

benthecarman:
Should this wait on #1541

nkohen:
> Should this wait on #1541

Nope, we probably will close that PR in favor of refactoring ujson usage instead of using play framework

PR #1576, opened 2020-06-16
Fix WalletAppConfig.hasWallet for Postgres

### LINKED ISSUE #1554: WalletAppConfig.hasWallet() does not work for postgres
The current implementation: 

https://github.com/bitcoin-s/bitcoin-s/blob/master/app/server/src/main/scala/org/bitcoins/server/Main.scala#L156

We just detect if there are files on disk. This will not work for postgres, we need a different way to detect if postgres is setup with a wallet correctly.

benthecarman:
Fixes #1554

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
@@ -136,14 +136,20 @@ case class WalletAppConfig(
   }
 
   /** Checks if the following exist
-    *  1. A wallet exists
-    *  2. seed exists
+    *  1. A seed exists
+    *  2. wallet exists
     *  3. The account exists */
-  def hasWallet()(implicit ec: ExecutionContext): Future[Boolean] = {
-    val walletDB = dbPath.resolve(dbName)
-    val hdCoin = defaultAccount.coin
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -136,14 +136,20 @@ case class WalletAppConfig(
...
-    val hdCoin = defaultAccount.coin
-    if (Files.exists(walletDB) && seedExists()) {
-      AccountDAO()(ec, this).read((hdCoin, 0)).map(_.isDefined)
+  private def hasWallet()(
+      implicit walletConf: WalletAppConfig,
+      ec: ExecutionContext): Future[Boolean] = {
+    if (walletConf.seedExists()) {
+      val hdCoin = walletConf.defaultAccount.coin
+      val walletDB = walletConf.dbPath resolve walletConf.dbName
+      if (walletConf.driverName == "postgresql" || Files.exists(walletDB)) {
```
Christewart, line 148: @rorp is this enough to deteremine that postgres is enabled and the wallet is configured correctly?

rorp: Yes, I think so.

PR #1578, opened 2020-06-16
Cap the amount of threads an actor system can spin up to 2 in tests

benthecarman:


Christewart:
It looks like we need to configure `default-blocking-io-dispatcher` and `internal-dispatcher` to have a max amount of threads too

Screenshot from 2020-06-16 15-57-20

Christewart:
Doesn't look like it worked on bd924ff 

Screenshot from 2020-06-17 07-01-21

testkit/src/main/resources/reference.conf
```diff
@@ -120,5 +119,31 @@ akka {
...
+            # Since every test suite currently
+            # has it's own actor system, we should limit the number of threads in the system
+            fork-join-executor {
+                # minimum of one thread for tests
+                parallelism-min = 1
+
+                # maximum of 2 threads for tests
+                parallelism-max = 2
+            }
+            default-blocking-io-dispatcher {
```
Christewart: It looks like the problem here is that this configuration is not defind for `default-blocking-io-dispatcher`? 

```
    default-blocking-io-dispatcher {
      type = "Dispatcher"
      executor = "thread-pool-executor"
      throughput = 1
      thread-pool-executor {
        fixed-pool-size = 16
      }
    }
```

We probably need to use the `fixed-pool-size` config

https://doc.akka.io//docs/akka/current/general/configuration-reference.html

benthecarman:
Looks like that fixed it :D

testkit/src/main/resources/reference.conf
```diff
@@ -120,5 +119,48 @@ akka {
...
+                parallelism-max = 2
+                fixed-pool-size = 1
+            }
+            thread-pool-executor {
+                fixed-pool-size = 1
+            }
+        }
+        default-blocking-io-dispatcher {
+            # minimum of one thread for tests
+            parallelism-min = 1
```
Christewart: I don't think `parallelism-min` and `parallelism` max are valid under `default-blocking-io-dispatcher` , so let's remove them. I think we also should have at least 2 threads in the `fixed-pool-size` as this is for blocking tasks that may take awhile, this will at least let us have two blocking tasks done as the same time.

https://doc.akka.io//docs/akka/current/general/configuration-reference.html

PR #1582, opened 2020-06-17
Add tests that NodeCallbacks are executed

benthecarman:
Tests that all the node callbacks are properly executed.

Needed to add a `dummyChainApi` to make the tests faster, as well as,avoid complications with the node not being correctly synced for the test

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
@@ -3,17 +3,15 @@ package org.bitcoins.node.networking.peer
 import org.bitcoins.core.currency._
 import org.bitcoins.core.gcs.{FilterType, GolombFilter}
 import org.bitcoins.core.p2p._
+import org.bitcoins.core.protocol.CompactSizeUInt
 import org.bitcoins.core.protocol.blockchain.{Block, BlockHeader, MerkleBlock}
 import org.bitcoins.core.protocol.transaction.Transaction
-import org.bitcoins.core.protocol.{BitcoinAddress, CompactSizeUInt}
 import org.bitcoins.crypto.DoubleSha256Digest
 import org.bitcoins.node._
-import org.bitcoins.rpc.client.common.BitcoindVersion
-import org.bitcoins.rpc.client.v19.BitcoindV19RpcClient
 import org.bitcoins.server.BitcoinSAppConfig
 import org.bitcoins.testkit.BitcoinSTestAppConfig
 import org.bitcoins.testkit.node.NodeUnitTest
-import org.bitcoins.testkit.node.fixture.SpvNodeConnectedWithBitcoind
+import org.bitcoins.testkit.node.fixture.SpvNodeConnectedWithBitcoindV19
 import org.scalatest.FutureOutcome
 
 import scala.concurrent.{Future, Promise}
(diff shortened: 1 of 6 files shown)


node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
```diff
@@ -0,0 +1,177 @@
...
+  /** Wallet config with data directory set to user temp directory */
+  implicit override protected def config: BitcoinSAppConfig =
+    BitcoinSTestAppConfig.getSpvWithEmbeddedDbTestConfig(pgUrl)
+
+  override type FixtureParam = SpvNodeConnectedWithBitcoind
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome =
+    withSpvNodeConnectedToBitcoind(test, Some(BitcoindVersion.V19))
+
+  private val junkAddress: BitcoinAddress =
```
Christewart: how many junk addresses are you creating? lol

node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
```diff
@@ -0,0 +1,177 @@
...
+        sender <- spv.peerMsgSenderF
+
+        hash <- bitcoind.generateToAddress(blocks = 1, junkAddress).map(_.head)
+        header <- bitcoind.getBlockHeaderRaw(hash)
+
+        payload = HeadersMessage(CompactSizeUInt.one, Vector(header))
+
+        callback: OnBlockHeadersReceived = (_: Vector[BlockHeader]) => {
+          Future {
+            resultP.success(true)
```
Christewart: in our assertions can you make `resultP` the type of the parameter of the callback. This way we can complete it with the parameter for the callback and we can guarantee that the same thing we expected to be passed in is the thing that the callback is executing on? It's just a bit of a sanity check

node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
```diff
@@ -0,0 +1,177 @@
...
+        dataMessageHandler = DataMessageHandler(dummyChainApi, callbacks)
+        _ <- dataMessageHandler.handleDataPayload(payload, sender)
+        result <- resultP.future
+      } yield assert(result)
+  }
+
+  it must "verify OnCompactFilterReceived callbacks are executed" in {
+    param: FixtureParam =>
+      val SpvNodeConnectedWithBitcoind(spv, rpc) = param
+      val bitcoind = rpc.asInstanceOf[BitcoindV19RpcClient]
```
Christewart: nit, it maybe worth making a `SpvNodeConnectedWithBitcoindV19` fixture as this test code is super brittle (what happens when we bump to V20?, or downgrade to V18?)

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -60,6 +71,99 @@ trait NodeUnitTest extends BitcoinSFixture with EmbeddedPg {
 
   lazy val bitcoindPeerF = startedBitcoindF.map(NodeTestUtil.getBitcoindPeer)
 
+  val dummyChainApi: ChainApi = new ChainApi {
```
Christewart: Maybe could be called the `genesisChainApi`? This might be useful else where. It seems to me all of the information in this chainApi would work with a blockchain on the genesis block?

node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
benthecarman: It's the same one I keep copying over, maybe should move it into the util file xD

Christewart (approved):

PR #1589, opened 2020-06-18
Pulled down dlc CLI code

nkohen:
This returns `400 Bad Request` when a request is sent to the server.


benthecarman:
Should we include the wallet routes?

benthecarman (approved):

PR #1590, opened 2020-06-18
DLC GUI Tab

nkohen:
Added stuff to a new package in the existing GUI and made a new tab for DLCs


benthecarman:
Is it possible to add not add the `dlc` package here, there shouldn't be merge conflicts that way

Issue #1586, opened 2020-06-18
Using `updateAll` on a `CRUDAutoInc` table causes a constraint violation


benthecarman:
It seems that `updateAll` is broken for our auto inc tables, going back to tested all the way back to `38db570e8f4` PR #827, and the issue still occurs. It is weird because using `update` works just fine and that function just calls `updateAll` with a single item. To reproduce this test case should suffice, it should be placed in the `SpendingInfoDAOTest`

```
  it must "update all " in { daos =>
    val WalletDAOs(_, addressDAO, spendingInfoDAO, _, _, _) = daos

    for {
      account <- daos.accountDAO.create(WalletTestUtil.firstAccountDb)
      addr <- addressDAO.create(getAddressDb(account))
      u1 = sampleLegacyUTXO(addr.scriptPubKey)
      _ <- insertDummyIncomingTransaction(daos, u1)
      utxo1 <- daos.utxoDAO.create(u1)

      u2 = WalletTestUtil.sampleSegwitUTXO(addr.scriptPubKey)
      _ <- insertDummyIncomingTransaction(daos, u2)
      utxo2 <- daos.utxoDAO.create(u2)

      utxos = Vector(utxo1, utxo2)
      changed = utxos.map(_.copyWithState(TxoState.DoesNotExist))
      updated <- spendingInfoDAO.updateAll(changed)
    } yield assert(updated == changed)
  }
```

it will fail on the `updateAll` line giving this error

```
must update all *** FAILED *** (3 seconds, 406 milliseconds)
[info]   org.sqlite.SQLiteException: [SQLITE_CONSTRAINT]  Abort due to constraint violation (UNIQUE constraint failed: txo_spending_info.id)
[info]   at org.sqlite.core.DB.newSQLException(DB.java:1010)
[info]   at org.sqlite.core.DB.newSQLException(DB.java:1022)
[info]   at org.sqlite.core.DB.execute(DB.java:861)
[info]   at org.sqlite.core.DB.executeUpdate(DB.java:902)
[info]   at org.sqlite.jdbc3.JDBC3PreparedStatement.executeUpdate(JDBC3PreparedStatement.java:98)
[info]   at slick.jdbc.LoggingPreparedStatement.$anonfun$executeUpdate$5(LoggingStatement.scala:170)
[info]   at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.scala:17)
[...]

PR #1596, opened 2020-06-19
Temporary fix for update all on CRUDAutoInc Tables

benthecarman:
Addresses #1586

This fixes that issue but not in a great way, what this does is does each update individually which is not optimal. I could not find a fix for this, possibly an issue with `update` in Jdbc because `upsert` works just fine.

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/CRUDAutoInc.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/CRUDAutoInc.scala
@@ -24,6 +24,8 @@ abstract class CRUDAutoInc[T <: DbRowAutoInc[T]](implicit
     safeDatabase.runVec(actions)
   }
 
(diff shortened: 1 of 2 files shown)


db-commons/src/main/scala/org/bitcoins/db/CRUDAutoInc.scala
```diff
@@ -23,6 +24,12 @@ abstract class CRUDAutoInc[T <: DbRowAutoInc[T]](implicit
     safeDatabase.runVec(actions)
   }
 
+  override def updateAll(ts: Vector[T]): Future[Vector[T]] = {
+    FutureUtil.foldLeftAsync(Vector.empty[T], ts) { (accum, t) =>
```
Christewart, line 30: Can you link to the issue that you have filed here so there is some context given

Christewart (approved):

PR #1598, opened 2020-06-19
AppConfig Start Interface

benthecarman:


Christewart (changes requested):
This needs to use the `StartStop` interface

Christewart (approved):

PR #1605, opened 2020-06-24
ScriptSignature from P2WSH Witness

nkohen:
Added method to pull script signature out of P2WSH witness

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
@@ -1,7 +1,11 @@
 package org.bitcoins.core.protocol.script
 
 import org.bitcoins.core.protocol.CompactSizeUInt
-import org.bitcoins.core.serializers.script.RawScriptWitnessParser
+import org.bitcoins.core.script.constant.{OP_0, ScriptNumberOperation}
+import org.bitcoins.core.serializers.script.{
+  RawScriptWitnessParser,
+  ScriptParser
+}
 import org.bitcoins.core.util.{BitcoinScriptUtil, BytesUtil}
 import org.bitcoins.crypto.{
   ECDigitalSignature,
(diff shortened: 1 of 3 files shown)


benthecarman (approved):
ACK on CI passing

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -106,6 +110,21 @@ sealed abstract class P2WSHWitnessV0 extends ScriptWitnessV0 {
     relevantStack.map(ECDigitalSignature.fromBytes)
   }
 
+  // Note that this is not guaranteed to work for non-standard script signatures
+  lazy val scriptSignature: ScriptSignature = {
```
Christewart, line 114: Should this be an `Option[ScriptSignature]`? 

This throws on the case where the `stack` is empty (`.tail`)

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -106,6 +110,21 @@ sealed abstract class P2WSHWitnessV0 extends ScriptWitnessV0 {
...
+    val asm = stack.toVector.tail.reverse.flatMap { bytes =>
+      if (bytes.isEmpty) {
+        Vector(OP_0)
```
Christewart, line 117: Why `OP_0` if there is nothing on the stack?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -106,6 +110,21 @@ sealed abstract class P2WSHWitnessV0 extends ScriptWitnessV0 {
...
+      } else if (bytes.length == 1 && bytes.head <= 16 && bytes.head >= -1) {
```
Christewart, line 118: Maybe just

```scala
val numOpt = bytes.headOption.flatMap(byte => ScriptNumberOperation.fromNumber(byte))

...

else if (numOpt.isDefined)) { 
  Vector(numOpt.get)
} 
...

nkohen: I personally like how it is because it clearly enumerates all cases

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
nkohen: Not nothing on the stack, an empty element on the stack (a Script zero)

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
nkohen: Upon further evaluation, I believe that this actually does work in all cases, and the thing that doesn't work is the apply method below which takes in a ScriptSignature (which will fail to act as expected if there's anything other than push ops). So no `Option`, but you're right that we should `require` that there be a redeem script (i.e. `stack.nonEmpty`) at the top of this trait

PR #1607, opened 2020-06-25
Only fail CallbackTest if promise is not completed 

### LINKED ISSUE #1604: CallbackTest failure on CI
https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/701453171#L660

I can't seem to find where the error occurs in the log though

Screenshot from 2020-06-24 06-28-05

it is concerning that the log linked above is not causing any failures on CI as far as I can tell?

From what I can tell it's not related to `CallbackTest`?

benthecarman:
Closes #1604

I think previously, the test would "pass" but then after the timeout expired the it would attempt to complete the promise and throw an error, causing the test to fail after the fact.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/api/CallbackTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/api/CallbackTest.scala
@@ -1,6 +1,5 @@
 package org.bitcoins.core.api
 
-import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.testkit.util.BitcoinSAsyncTest
 import org.scalatest.Assertion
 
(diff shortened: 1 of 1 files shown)


benthecarman:
Looks like this made it worse :/

core-test/src/test/scala/org/bitcoins/core/api/CallbackTest.scala
```diff
@@ -17,13 +17,19 @@ class CallbackTest extends BitcoinSAsyncTest {
 
     val f1: Callback[Unit] = _ => {
       Thread.sleep(testTimeout.toMillis)
```
Christewart: You might want to use use `system.scheduler` to schedule the promise to be completed at `testTimeout` rather than using `Thread.sleep`. I'm not sure if this is related or not, but after merging #1578 maybe there can be some thread starvation going on with `Thread.sleep()`?

core-test/src/test/scala/org/bitcoins/core/api/CallbackTest.scala
```diff
@@ -16,14 +16,21 @@ class CallbackTest extends BitcoinSAsyncTest {
...
 
     val f1: Callback[Unit] = _ => {
-      Thread.sleep(testTimeout.toMillis)
-      promise.complete(fail("2nd callback did not start before timeout"))
+      system.scheduler.scheduleOnce(testTimeout) {
+        if (!promise.isCompleted) {
+          promise.complete(fail("2nd callback did not start before timeout"))
+        }
+      }
       FutureUtil.unit
```
Christewart: Now this isn't actually blocking anything, you need to return 

`promise.future`

PR #1608, opened 2020-06-25
Attempt to fix NPE in forAllAsync

### LINKED ISSUE #1536: NPE on PSBT construct and sign test case
https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/696429090#L1675

Screenshot from 2020-06-09 07-42-36

benthecarman:
Closes #1536


Christewart:
I don't see how this would fix this? A NPE exception wouldn't get fixed by just calling `.result()` twice

PR #1611, opened 2020-06-25
Node broadcast tx twice test

benthecarman:


benthecarman:
Looks like this doesn't work on postgres, seems it might be indicative of more issues with `CRUDAutoInc`

Christewart:
> Looks like this doesn't work on postgres, seems it might be indicative of more issues with `CRUDAutoInc`

Perhaps we should just write unit tests for `CRUD`/`CRUDAutoInc` against Postgres/sqlite rather than try to debug indirectly?

PR #1618, opened 2020-06-26
CRUD Test suite + updateAll improvements

### LINKED ISSUE #1616: Add CRUDTest & CRUDAutoIncTest to test functionality on each database we support
From this comment:

https://github.com/bitcoin-s/bitcoin-s/pull/1611#issuecomment-650128425

We really need tests that directly test all functionality in `CRUD`/`CRUDAutoInc` _without_ testing other functionality related to a specific DAO. Hopefully these will save us a lot of time in the future of trying to figure out "why things are working on sqlite, but not on Posgres". We should be able to do this for all database drivers we support. As of the time of this writing, we support postgre and sqlite

benthecarman:
Closes #1616 

Adds tests for the `CRUD` interface.

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/db/TestDAOFixture.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/db/TestDAOFixture.scala
@@ -94,15 +94,8 @@ sealed trait TestDAOFixture
       _ = assert(created)
 
       _ = testDAO.deleteAll()
-      secondRead0 <- testDAO.read(id = "abc")
-      secondRead1 <- testDAO.read(id = "abc1")
-      secondRead2 <- testDAO.read(id = "abc2")
-      secondRead3 <- testDAO.read(id = "abc3")
-    } yield {
-      val read2 =
-        Vector(secondRead0, secondRead1, secondRead2, secondRead3).flatten
-      read2 == Vector.empty
-    }
+      all <- testDAO.findAll()
+    } yield all.isEmpty
   }
 
   def testUpsert(testDAO: TestDAO): Future[Boolean] = {
--- a/testkit/src/main/scala/org/bitcoins/testkit/db/TestDAOFixture.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/db/TestDAOFixture.scala
@@ -61,7 +61,7 @@ sealed trait TestDAOFixture
   def testCreate(testDAO: TestDAO): Future[Boolean] = {
     for {
       _ <- testDAO.create(testDb)
-      read <- testDAO.read(id = "abc")
+      read <- testDAO.read(testDb.pk)
     } yield read.contains(testDb)
   }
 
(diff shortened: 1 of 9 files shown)


db-commons/src/main/resources/db.conf
```diff
@@ -51,4 +51,13 @@ bitcoin-s {
       url = "jdbc:sqlite:"${bitcoin-s.chain.db.path}${bitcoin-s.chain.db.name}
     }
   }
+
+  test = ${common}
```
Christewart: This probably should not be in `src/main`, the configuration should be in `test`

testkit/src/main/resources/postgresql/test/migration/V1__test_db_baseline.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE test_table (pk TEXT NOT NULL,data TEXT NOT NULL,constraint pk_pk primary key(pk));
```
Christewart: I don't think we need migrations for this do we? This is strictly for test cases so I think we can just call create the table directly with Slick using `DbManagement.createTable()`?

Christewart (commented):
So does this iteration strictly only test postgres or does it test sqlite as well? If it tests both, where exactly is the code that differentiates which database driver we are testing?

benthecarman:
> So does this iteration strictly only test postgres or does it test sqlite as well? If it tests both, where exactly is the code that differentiates which database driver we are testing?

This is similar to how our other tests work and needs `PG_ENABLED="1"` to be postgres. I would like to make it test both without having to do that but haven't fully figured that out yet.

testkit/src/main/resources/postgresql/test/migration/V1__test_db_baseline.sql
benthecarman: I think this is preferred, this way we can test that the migrations are correctly hooked up to the tests

Christewart: This is just test code isn't it? We don't need migrations for test code IMO as we can just re-create the test environment

testkit/src/main/scala/org/bitcoins/testkit/db/TestDAOFixture.scala
```diff
@@ -0,0 +1,175 @@
...
+                                       TestDb("abc3", hex"0053"))
+
+  val updatedDb: TestDb = testDb.copy(data = hex"0000")
+
+  val updatedDbs: Vector[TestDb] = testDbs.map(_.copy(data = hex"0000"))
+
+  def testCreate(testDAO: TestDAO): Future[Boolean] = {
+    for {
+      _ <- testDAO.create(testDb)
+      read <- testDAO.read(id = "abc")
```
Christewart: nit: `testDb.pk` no?

Christewart (approved):

PR #1621, opened 2020-07-01
Override correct equals functions

benthecarman:
TIL overriding `==` is incorrect and should instead override the `equals` function.


Christewart:
Looks like coverage is failing for some reason? https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/704051439#L670

PR #1631, opened 2020-07-07
Move BIP 143 invariant for PSBTs to allow signed transactions

benthecarman:
This moves the check for the psbt input being vulnerable to BIP 143. This is useful because it will allow PSBTs that are signed as well as single input transactions to use `WitnessUTXO`. Single input txs are not vulnerable and PSBTs that are already signed are not be vulnerable. 

Needed to fix some test cases to be compatible.


Christewart:
Tests fail https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/705655149#L1659

Christewart (approved):

PR #1638, opened 2020-07-07
Add docker setup docs

matthewjablack:
This PR adds docker-setup.md for a quick and easy way to setup bitcoin-s with Docker.


benthecarman:
This is great! However, the bitcoind and eclair conf are only used when running the end-to-end dlc branch. Those sections should be moved into a separate section and denote what they are intended for.

Also, this should be added to the sidebar. You should be able to add it in `website/sidebars.json` under getting setup

PR #1643, opened 2020-07-08
Run only docs CI for PRs that start with Docs:

benthecarman:
We currently have a rule for the CI that it won't run any if it starts with `Docs:`, however, since we test the docs in the CI we never use this. This makes it so when we have a PR's commit message starts with `Docs:` it will now just run the docs CI matrix.

Remake of #1642, it accidentally got closed and wouldn't let me reopen it :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,19 +134,19 @@ before_script:
 stages:
   - name: docs
     if:
-      commit_message =~ /^Docs:/ AND NOT
+      commit_message =~ /(?i)^docs:/ AND NOT
       ((branch = master AND type = push) OR (tag IS present))
     # don't run tests on merge builds, just publish library
     # and website
   - name: compile
     if:
-      commit_message !~ /^Docs:/ AND NOT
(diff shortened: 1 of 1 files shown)


Christewart:
This is cool, i'm guessing it is case sensitive? I.e `dOcs:` won't work? Not that big of a deal, but I could definitely see myself doing 'docs:' 

Also is this update to date? https://bitcoin-s.org/docs/next/contributing-website#ci

benthecarman:
> This is cool, i'm guessing it is case sensitive? I.e `dOcs:` won't work? Not that big of a deal, but I could definitely see myself doing 'docs:'

Can probably make it case insensitive 

> Also is this update to date? [bitcoin-s.org/docs/next/contributing-website#ci](https://bitcoin-s.org/docs/next/contributing-website#ci)

LGTM, still doesn't run tests, now just checks compilation.

benthecarman:
Now case insensitive :)

PR #1651, opened 2020-07-08
Fix length of BigIntMapper

### LINKED ISSUE #1645: Primary Key constraint failed block_headers.hash
**Describe the bug**
When starting my bitcoin-s node on e3938544edd4e841ee4e317fd433865d1f8a1cb0 (although I don't think this is introduced here) it stops syncing due to a primary key error

If you look at the logs below, there is this suspicious error message: 

>2020-07-08T18:19:48 ERROR [P2P] Cannot send a message to our peer when we are not connected! payload=GetDataMessage(inventoryCount=1, inventories=Vector(Inventory(MsgBlock,DoubleSha256Digest(b639ca945ef5833f8254988969906555654d27da080015000724210...) peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333)

I wonder if this fails to send, and causes us to sync from the beginning? 

I have synced my node before, but for some reason this is my best block hash? 

>2020-07-08T18:20:23 INFO [P2P] Started node, best block hash 000000000eb141f16b227064410339bd623efe96e0a17be9301ac35984456f4c at height 14442, with 14442 filter headers and 14442 filers

You can also see i these headers in my database: 

>sqlite> select COUNT(*) from block_headers;
1723239

**Logs**
https://pastebin.com/stDHMup1

benthecarman:
Closes #1645 

For mainnet block 88443 it's chain work is encoded to `0xfffef2bf0566ab`
and for block 260000 is encoded to `0x01253721228459eac00c`, so sqlite sees 88443 as larger because it starts with 0xff even though block 260000's chain work is longer.

The solution here is to enforce that they are all 32 bytes long so sqlite can't confuse them and prioritize ones with shorter lengths.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala
@@ -7,6 +7,7 @@ import org.bitcoins.crypto.DoubleSha256DigestBE
 import org.bitcoins.testkit.chain.{
   BlockHeaderHelper,
   ChainDbUnitTest,
+  ChainTestUtil,
   ChainUnitTest
 }
 import org.scalatest.FutureOutcome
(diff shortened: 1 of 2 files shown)


benthecarman:
Added a test, made sure this failed beforehand but does not after this change

Christewart:
Can you add a description what the fix is for context? Closing the issue and saying you added a test case doesn't really explain what the problem was in the implementation

benthecarman:
> Can you add a description what the fix is for context? Closing the issue and saying you added a test case doesn't really explain what the problem was in the implementation

Added to the PR desc

Christewart (approved):

PR #1657, opened 2020-07-10
Remove the akka boiler plate needed to get around issue 1508

### LINKED ISSUE #1508: Deadlock on appServer/run after upgrading to akka 2.6

Christewart:
Closes #1508 

This was fixed upstream in akka 

https://github.com/akka/akka-http/issues/3221#issuecomment-644774799

https://github.com/akka/akka/pull/29244

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -27,8 +27,6 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 object Main extends App with BitcoinSLogger {
 
   private def runMain(): Unit = {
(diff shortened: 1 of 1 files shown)


benthecarman (approved):
Works for me locally

PR #1658, opened 2020-07-10
Sign transactions when built through fundRawTransaction

benthecarman:
Previously, we were building the transaction without signing it, this fixes it to sign it.


Christewart:
Bitcoin core does not sign transactions when using `fundrawtransaction` so i don't think we should either: https://developer.bitcoin.org/reference/rpc/fundrawtransaction.html#fundrawtransaction

benthecarman:
> Bitcoin core does not sign transactions when using `fundrawtransaction` so i don't think we should either: [developer.bitcoin.org/reference/rpc/fundrawtransaction.html#fundrawtransaction](https://developer.bitcoin.org/reference/rpc/fundrawtransaction.html#fundrawtransaction)

Sounds good, this broke a bunch of tests as well.

PR #1660, opened 2020-07-11
Create ShuffleFinalizer

### LINKED ISSUE #1648: Create `ShuffleFinalizer`
Would be nice to have a `RawTxFinalizer` that shuffles the inputs and outputs ordering. This should improve privacy for users as well as could make our testing slightly more robust.

benthecarman:
Closes #1648

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
@@ -60,6 +60,26 @@ trait RawTxFinalizer {
   }
 }
 
+trait FinalizerFactory[T <: RawTxFinalizer] {
+
+  def txBuilderFrom(
+      outputs: Seq[TransactionOutput],
+      utxos: Seq[InputSigningInfo[InputInfo]],
+      feeRate: FeeUnit,
+      changeSPK: ScriptPubKey): RawTxBuilderWithFinalizer[T]
+
+  def txFrom(
+      outputs: Seq[TransactionOutput],
+      utxos: Seq[InputSigningInfo[InputInfo]],
+      feeRate: FeeUnit,
+      changeSPK: ScriptPubKey)(implicit
+      ec: ExecutionContext): Future[Transaction] = {
+    val builderF = Future(txBuilderFrom(outputs, utxos, feeRate, changeSPK))
+
+    builderF.flatMap(_.buildTx())
+  }
+}
+
 /** A trivial finalizer that does no processing */
 case object RawFinalizer extends RawTxFinalizer {
 
@@ -249,7 +269,8 @@ case class StandardNonInteractiveFinalizer(
   }
 }
 
-object StandardNonInteractiveFinalizer {
+object StandardNonInteractiveFinalizer
+    extends FinalizerFactory[StandardNonInteractiveFinalizer] {
 
   def txBuilderFrom(
       outputs: Seq[TransactionOutput],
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -344,3 +358,33 @@ object SubtractFeeFromOutputsFinalizer {
...
+
+  override def buildTx(txBuilderResult: RawTxBuilderResult)(implicit
+      ec: ExecutionContext): Future[Transaction] = {
+    ShuffleInputsFinalizer
+      .andThen(ShuffleOutputsFinalizer)
+      .buildTx(txBuilderResult)
+  }
+}
+
+case object ShuffleInputsFinalizer extends RawTxFinalizer {
```
Christewart, line 414: Can you add a scaladoc for what he purpose of this is? Same thing with `ShuffleOutputsFinalizer`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -227,7 +227,8 @@ case class AddWitnessDataFinalizer(inputInfos: Vector[InputInfo])
 case class StandardNonInteractiveFinalizer(
     inputInfos: Vector[InputInfo],
     feeRate: FeeUnit,
-    changeSPK: ScriptPubKey)
+    changeSPK: ScriptPubKey,
+    otherFinalizerOpt: Option[RawTxFinalizer] = None)
```
Christewart: This seems like a hack to me, since we are still pre-0.4 do we want to change design to support this?

benthecarman: Yeah, wasn't sure on the best way to do this. I'm guessing @nkohen will have the strongest opinion on it.

nkohen: So how we move forward here depends on the answer to the question
```
Should the "standard" finalizer shuffle or not?
```
Regardless of the answer to this question there should be two separate finalizers: one which does the shuffling and one which does not. The only question is which should be called standard and which should be called `[Shuffled | Unshuffled]NonInteractiveFinalizer`. Due to the add witness component of each composition won't be possible but this is not really an issue as it just means that there will be 10ish lines of code duplicated, none of which is logic (all construction).

TLDR: Get rid of the Option and make two different finalizers, the "standard" one gets to keep the companion object.

P.S. My vote is to leave standard as is and introduce a new one called shuffled until the PR that actually changes how we handle things internally.

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -280,6 +280,67 @@ object StandardNonInteractiveFinalizer {
...
+      expectedFeeRate = feeRate,
+      changeSPKs = Vector(changeSPK))
+
+    val addWitnessData = AddWitnessDataFinalizer(inputInfos)
+
+    addChange
+      .andThen(sanityCheck)
+      .andThen(addWitnessData)
+      .andThen(ShuffleFinalizer)
+      .buildTx(txBuilderResult)
```
nkohen: `addWitnessData` *MUST* happen last otherwise its effects are dropped (as per its scaladoc).
`sanityCheck` *SHOULD* happen second to last so that we can sanity check the inputs and outputs post-shuffling.
Thus, this should be
```
addChange
    .andThen(ShuffleFinalizer)
    .andThen(sanityCheck)
    .andThen(addWitnessData)
    .buildTx(txBuilderResult)
```

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -280,6 +280,67 @@ object StandardNonInteractiveFinalizer {
...
+  }
+}
+
+object ShufflingNonInteractiveFinalizer {
```
nkohen: I think we should have a trait for this kind of `object` which has `txBuilderFrom` and `txFrom` with these same parameters.

nkohen (commented):
Just a couple things

nkohen (approved):

PR #1662, opened 2020-07-12
Implemenet BitGo fee provider

benthecarman:


fee-provider/src/main/scala/org/bitcoins/feeprovider/BitGoFeeRateProvider.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.Uri
+import org.bitcoins.commons.jsonmodels.wallet.BitGoResult
+import org.bitcoins.commons.serializers.JsonSerializers._
+import org.bitcoins.core.wallet.fee.SatoshisPerKiloByte
+import play.api.libs.json.{JsError, JsSuccess, Json}
+
+import scala.util.{Failure, Success, Try}
+
+case class BitGoFeeRateProvider(blockTargetOpt: Option[Int])(implicit
```
Christewart, line 15: Can you link to bitgo's documentation there so that we don't have to hunt it down in the future when we need to look at it?

PR #1663, opened 2020-07-12
Random Coin Selection

benthecarman:
Intended for if someone wants to use bitcoin-s for research on coin selection algos


Christewart (changes requested):
Can you add some sort of base line unit test to make sure this works? I understand it's hard since it's random, it also worries me there are zero tests

Christewart (approved):

PR #1664, opened 2020-07-12
Fix CoinSelector for KiloX fee rates

benthecarman:
There was a bug in the `CoinSelector` when we are using a fee rate that is either `SatoshisPerKiloByte` or `SatoshisPerKW`. This was caused because at this line:

https://github.com/benthecarman/bitcoin-s/blob/2d8ad5b14e9767bd8c6c70ff95dafd49200e1197/wallet/src/main/scala/org/bitcoins/wallet/api/CoinSelector.scala#L59

we would multiply by `feeRate.currencyUnit`, this causes problems for anything where in our calculation we have to divide by X (in this case 1000) and overestimate the fee that each input is adding.

To fix this, I reworked the `FeeUnit` api to hopefully prevent this, now we take in a function that is what we use for the tx calc and a separate function for multiplying. This way the user will have to set a the scale factor in the multiply function and to correctly calculate the tx fee, so their `feeRate * Long` will be correct as well.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
@@ -12,9 +12,7 @@ sealed abstract class FeeUnit {
   final def *(cu: CurrencyUnit): CurrencyUnit = this * cu.satoshis.toLong
   final def *(int: Int): CurrencyUnit = this * int.toLong
 
-  final def *(long: Long): CurrencyUnit =
-    if (scaleFactor == 1) currencyUnit * long
-    else currencyUnit * long / Satoshis(scaleFactor)
+  final def *(long: Long): CurrencyUnit = Satoshis(toLong * long / scaleFactor)
   def *(tx: Transaction): CurrencyUnit = calc(tx)
 
   /** The coefficient the denominator in the unit is multiplied by,
--- a/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
@@ -12,14 +12,18 @@ sealed abstract class FeeUnit {
   final def *(cu: CurrencyUnit): CurrencyUnit = this * cu.satoshis.toLong
   final def *(int: Int): CurrencyUnit = this * int.toLong
 
-  /** Should multiply currencyUnit by long along with a scale factor if necessary.
-    * This will be used when calculating the fee for a transaction.
-    *
-    * This should ONLY be overridden if the FeeUnit's denominator is NOT a multiple of 1.
-    *  ie sats/kilobyte
-    */
-  def *(long: Long): CurrencyUnit = currencyUnit * long
+  final def *(long: Long): CurrencyUnit =
+    if (scaleFactor == 1) currencyUnit * long
+    else currencyUnit * long / Satoshis(scaleFactor)
   def *(tx: Transaction): CurrencyUnit = calc(tx)
+
+  /** The coefficient the denominator in the unit is multiplied by,
+    * for example sats/kilobyte -> 1000
+    */
+  def scaleFactor: Long
+
+  require(scaleFactor > 0,
+          s"Scale factor cannot be less than or equal to 0, got $scaleFactor")
 
   /** Takes the given transaction returns a size that will be used for calculating the fee rate.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -9,8 +9,14 @@ import org.bitcoins.core.protocol.transaction.Transaction
   */
 sealed abstract class FeeUnit {
   def currencyUnit: CurrencyUnit
+  def *(cu: CurrencyUnit): CurrencyUnit = this * cu.satoshis.toLong
+  def *(int: Int): CurrencyUnit = this * int.toLong
+  def *(long: Long): CurrencyUnit = currencyUnit * long
   def *(tx: Transaction): CurrencyUnit = calc(tx)
-  def calc(tx: Transaction): CurrencyUnit
+  def txSizeForCalc(tx: Transaction): Long
```
Christewart, line 29: What does this do? Can't we just use tx.weight?

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -9,8 +9,14 @@ import org.bitcoins.core.protocol.transaction.Transaction
   */
 sealed abstract class FeeUnit {
   def currencyUnit: CurrencyUnit
+  def *(cu: CurrencyUnit): CurrencyUnit = this * cu.satoshis.toLong
+  def *(int: Int): CurrencyUnit = this * int.toLong
```
Christewart: What unit is `int` here? Aren't you just moving the original problem you had into this? 

Same with `def *(long: Long)`

Christewart (commented):
So this is sort of re-purposing what `FeeUnit` is. Now we can compute "raw" fees with `FeeUnit` that aren't necessarily associated with a transaction?

Maybe I'm just not understanding how `CoinSelector` works, but why don't we just build a transaction in `CoinSelector` with `RawTxBuilder`? Shouldn't that give us a more accurate fee estimate? 

At the end of the day, if @nkohen is ok with changing `FeeUnit` to work like this then I am ok with it.

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
benthecarman: No, that's only for per-vbyte fee units. For a per-byte we need to use byteSize

benthecarman:
> So this is sort of re-purposing what `FeeUnit` is. Now we can compute "raw" fees with `FeeUnit` that aren't necessarily associated with a transaction?

Yes

> Maybe I'm just not understanding how `CoinSelector` works, but why don't we just build a transaction in `CoinSelector` with `RawTxBuilder`? Shouldn't that give us a more accurate fee estimate?

It essentially already does that but uses some short cuts like having a `bytesSoFar` variable to not need to calculate the full transaction.

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -9,8 +9,14 @@ import org.bitcoins.core.protocol.transaction.Transaction
...
+  def *(long: Long): CurrencyUnit = currencyUnit * long
```
nkohen: This method absolutely needs a detailed scaladoc because it is the only non-abstract method that needs to be overridden sometimes. You have to make sure its super clear that people only override this method and not others. Also make clear what the responsibility is here wrt units.

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
nkohen, line 29: scaladoc

nkohen (commented):
I think the documentation required for this change to happen is not there but otherwise seems fine to me (probably need some tests too)

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
Christewart: This seems like a big footgun, we should probably leave this as abstract in the `FeeUnit` and implement in our subtypes as I know I probably won't remember to override this next time we need to modify this. At least if it's abstract the compiler will yell at me

Christewart (approved):

nkohen (approved):

Issue #1581, opened 2020-06-17
org.postgresql.util.PSQLException: The connection attempt failed 


Christewart:
https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/699299783#L696

![Screenshot from 2020-06-17 08-50-42](https://user-images.githubusercontent.com/3514957/84906521-abe91f00-b077-11ea-9e83-3ea306484ead.png)

benthecarman:
https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/708393865#L842

PR #1667, opened 2020-07-13
Disable parallelExecution

rorp:


benthecarman (approved):

Christewart (changes requested):
So this seems to fix the #1581, however this does increase CI times. Here is the latest CI run on this PR:

Screenshot from 2020-07-14 06-44-43

and here is the latest CI run on #1664 

Screenshot from 2020-07-14 07-08-45

It seems that it is the `bitcoind+eclair` test suite that has increased in time the most with this setting (+5 minutes!). There is no needed to set `parallelExecution in Test := false` in those test suites, as they do not touch any of the database code.

PR #1669, opened 2020-07-14
Bump scala 2.12 compiler to 2.12.12

Christewart:
https://github.com/scala/scala/releases/tag/v2.12.12


benthecarman (changes requested):
Should update the versions for the travis builds as well

benthecarman (approved):

PR #1670, opened 2020-07-14
Segwit Signing Bug Fix

nkohen:
Various signing improvements and a fix for badly rebuilt segwit scripts when signing

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
@@ -49,13 +49,13 @@ case object WitnessVersion0 extends WitnessVersion {
           //need to check if the hashes match
           val stackTop = scriptWitness.stack.head
           val stackHash = CryptoUtil.sha256(stackTop)
-          if (stackHash != Sha256Digest(witnessProgram.head.bytes)) {
+          val witnessHash = Sha256Digest(witnessProgram.head.bytes)
+          if (stackHash != witnessHash) {
             logger.debug(
               "Witness hashes did not match Stack hash: " + stackHash)
             logger.debug("Witness program: " + witnessProgram)
-            Failure(
-              new IllegalArgumentException(
-                "Witness hash did not match stack hash"))
+            Failure(new IllegalArgumentException(
+              s"Witness hash $witnessHash did not match stack hash $stackHash"))
           } else {
             val compactSizeUInt =
               CompactSizeUInt.calculateCompactSizeUInt(stackTop)
@@ -69,9 +69,8 @@ case object WitnessVersion0 extends WitnessVersion {
         logger.error("Witness: " + scriptWitness)
         logger.error("Witness program: " + witnessProgram)
         //witness version 0 programs need to be 20 bytes or 32 bytes in size
-        Failure(
-          new IllegalArgumentException(
-            "Witness program had invalid length for version 0"))
+        Failure(new IllegalArgumentException(
+          s"Witness program had invalid length (${programBytes.length}) for version 0, must be 20 or 30: $witnessProgram"))
     }
   }
 
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
@@ -2,9 +2,10 @@ package org.bitcoins.core.protocol.script
 
 import org.bitcoins.core.protocol.CompactSizeUInt
 import org.bitcoins.core.script.constant._
-import org.bitcoins.core.script.result._
 import org.bitcoins.core.util.{BitcoinSLogger, BytesUtil}
 import org.bitcoins.crypto.{CryptoUtil, Sha256Digest, Sha256Hash160Digest}
+
+import scala.util.{Failure, Success, Try}
 
 /**
   * Created by chris on 11/10/16.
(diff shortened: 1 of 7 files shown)


benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
```diff
@@ -61,13 +53,14 @@ case object WitnessVersion0 extends WitnessVersion {
             logger.debug(
               "Witness hashes did not match Stack hash: " + stackHash)
             logger.debug("Witness program: " + witnessProgram)
-            Right(ScriptErrorWitnessProgramMisMatch)
+            Failure(
+              new IllegalArgumentException(
+                "Witness hash did not match stack hash"))
```
Christewart: nit: add both hashes to the error message

core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
```diff
@@ -76,7 +69,9 @@ case object WitnessVersion0 extends WitnessVersion {
         logger.error("Witness: " + scriptWitness)
         logger.error("Witness program: " + witnessProgram)
         //witness version 0 programs need to be 20 bytes or 32 bytes in size
-        Right(ScriptErrorWitnessProgramWrongLength)
+        Failure(
+          new IllegalArgumentException(
+            "Witness program had invalid length for version 0"))
```
Christewart: nit: add the length (possibly the actual program) to the error message

core/src/main/scala/org/bitcoins/core/protocol/transaction/TxUtil.scala
```diff
@@ -19,55 +19,55 @@ import scala.util.{Failure, Success, Try}
```
Christewart, line 19: Is everything just a refactor in this file?

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
```diff
@@ -220,7 +220,8 @@ case class InputPSBTMap(elements: Vector[InputPSBTRecord])
         redeemScriptOpt.isDefined && redeemScriptOpt.get.redeemScript
           .isInstanceOf[WitnessScriptPubKeyV0]
 
-      isNativeV0 || isP2SHV0
+      // Only unsafe if it doesn't already have signatures
+      (isNativeV0 || isP2SHV0) && partialSignatures.isEmpty
```
Christewart, line 224: So this means if somebody else provided a signature we assume that it is safe? I.e. they are responsible for checking other outpoints on the tx?

core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala
```diff
@@ -436,10 +436,46 @@ sealed abstract class ScriptInterpreter extends BitcoinSLogger {
       else evaluated
     }
 
+    def rebuildV0(witness: ScriptWitness, program: Seq[ScriptToken]): Either[
+      (Seq[ScriptToken], ScriptPubKey),
```
Christewart: Ok, so since we are touching this, I think we should also do `Either[ScriptError,(Seq[ScriptToken],ScriptPubKey)]` (right-bias) return type as in Scala 2.12 they made `Either` right-biased

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -575,15 +571,13 @@ trait BitcoinScriptUtil extends BitcoinSLogger {
...
-    scriptEither match {
-      case Left((_, scriptPubKey)) =>
+  private def parseScriptTry(scriptT: Try[ScriptPubKey]): Seq[ScriptToken] =
+    scriptT match {
+      case Success(scriptPubKey) =>
         logger.debug(
           "Script pubkey asm inside calculateForSigning: " + scriptPubKey.asm)
         scriptPubKey.asm
-      case Right(_) => Nil //error
+      case Failure(err) => throw err
```
Christewart, line 580: :tada:

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
benthecarman: This is per-input, so if the input is signed it won't need to be checked, but if you go to sign another it will still need to be checked

core/src/main/scala/org/bitcoins/core/protocol/transaction/TxUtil.scala
nkohen: Yup, just made things easier to reason about

core/src/main/scala/org/bitcoins/core/psbt/PSBTMap.scala
nkohen: Yup, when signing one must always do the extra verification, but if you wanted to verify signatures using a PSBT, you shouldn't be forced to do such verification

nkohen: so for example you could do
```
psbt.fromTransaction(utx).addSignatures(sigs).addWitnessUTXO(wutxo)
```
and then validate the sigs there

nkohen: meanwhile
```
psbt.fromTransaction(utx).addWitnessUTXO(wutxo).sign(0, key)
```
will still fail

Christewart (approved):

PR #1673, opened 2020-07-15
Enforce unique outpoints for SpendingInfoDb

benthecarman:


db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
```diff
@@ -58,7 +58,7 @@ class DbManagementTest extends BitcoinSAsyncTest with EmbeddedPg {
                                           dbConfig(ProjectType.Wallet))
     val walletDbManagement = createWalletDbManagement(walletAppConfig)
     val result = walletDbManagement.migrate()
-    val expected = if (walletAppConfig.driverName == "postgresql") 2 else 5
+    val expected = if (walletAppConfig.driverName == "postgresql") 3 else 6
```
Christewart, line 61: nit: We may want to factor this magic number out into a file somewhere?

Christewart (approved):

PR #1674, opened 2020-07-15
Add ProcessBlockTest

benthecarman:
Didn't end up finding my issue, but this should at least be added to master


Christewart:
There is this failure that looks suspicious: 

https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/708393865#L822

PR #1676, opened 2020-07-15
Unreserve spent utxos

benthecarman:
Previously, a utxo wouldn't be marked unreserved when it was spent, this fixes it so once it has been confirmed in a block it will.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -186,8 +186,21 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           for {
             outputsBeingSpent <-
               spendingInfoDAO.findOutputsBeingSpent(transaction)
+
+            // unreserved outputs now they are in a block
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -186,8 +186,21 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           for {
             outputsBeingSpent <-
               spendingInfoDAO.findOutputsBeingSpent(transaction)
+
+            // unreserved outputs now they are in a block
+            outputsToUse = blockHashOpt match {
```
Christewart, line 191: Hmm, so we need to unreserve, and switch to `TxoState.ConfirmedSpent`?

benthecarman: We either will in the end set it to `ConfirmedSpent` or `PendingConfirmationsSpent`. However, we set it to `PendingConfirmationsReceived` just to unreserve it and then let `markAsPendingSpent` mark it accordingly

Christewart (approved):

PR #1680, opened 2020-07-17
Shuffle Finalizer Tests + bug fix

benthecarman:
After trying to implement the ShuffleFinalizer in the wallet I found some bugs, I fixed them and added tests to make sure that doesn't happen again!

Also found a bug in `isValidFeeRange` where if `estimatedFee` and `actualFee` were both `0` it would fail, so I fixed that as well.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxFinalizerTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxFinalizerTest.scala
@@ -4,7 +4,13 @@ import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.protocol.script.{EmptyScriptSignature, ScriptPubKey}
 import org.bitcoins.core.protocol.transaction._
+import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.crypto.DoubleSha256DigestBE
+import org.bitcoins.testkit.core.gen.{
+  CreditingTxGen,
+  CurrencyUnitGenerator,
+  ScriptGenerators
+}
 import org.bitcoins.testkit.util.BitcoinSAsyncTest
 import org.scalatest.Assertion
 
@@ -12,6 +18,9 @@ import scala.concurrent.Future
 import scala.util.Random
 
 class RawTxFinalizerTest extends BitcoinSAsyncTest {
+
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
 
   behavior of "RawTxFinalizer"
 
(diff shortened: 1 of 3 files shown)


core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxFinalizerTest.scala
```diff
@@ -131,6 +140,54 @@ class RawTxFinalizerTest extends BitcoinSAsyncTest {
...
+        val txBuilder = RawTxBuilder() ++= inputs ++= outputs
+        val finalized = txBuilder.setFinalizer(ShuffleInputsFinalizer)
+        val txsF =
+          FutureUtil.foldLeftAsync(Vector.empty[Transaction], 0 to 20) {
+            (accum, _) => finalized.buildTx().map(_ +: accum)
+          }
+        txsF.map(txs =>
+          assert(
+            inputs.size <= 1 || txs.exists(
+              _.inputs.map(_.previousOutput) != inputs.map(_.outPoint))))
```
nkohen, line 155: If the number of inputs is small, won't this assertion have a decent chance of failing?

core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxFinalizerTest.scala
```diff
@@ -131,6 +140,54 @@ class RawTxFinalizerTest extends BitcoinSAsyncTest {
...
+    forAllAsync(CreditingTxGen.inputsAndOutputs()) {
+      case (inputs, outputs) =>
+        val txBuilder = RawTxBuilder() ++= inputs ++= outputs
+        val finalized = txBuilder.setFinalizer(ShuffleOutputsFinalizer)
+        val txsF =
+          FutureUtil.foldLeftAsync(Vector.empty[Transaction], 0 to 20) {
+            (accum, _) => finalized.buildTx().map(_ +: accum)
+          }
+        txsF.map(txs =>
+          assert(outputs.size <= 1 || txs.exists(_.outputs != outputs)))
```
nkohen, line 169: diito

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -319,19 +319,25 @@ case class ShufflingNonInteractiveFinalizer(
...
-      .buildTx(txBuilderResult)
+      val shuffledInputInfos = tempTxBuilderResult.inputs
+        .map(input => inputInfos.find(_.outPoint == input.previousOutput).get)
+
+      val sanityCheck =
+        SanityCheckFinalizer(inputInfos = shuffledInputInfos,
+                             expectedOutputSPKs =
+                               tempTxBuilderResult.outputs.map(_.scriptPubKey),
+                             expectedFeeRate = feeRate,
+                             changeSPKs = Vector(changeSPK))
```
nkohen: Why did this change need to happen? Do any of the sanity checks depend on order?

core-test/src/test/scala/org/bitcoins/core/wallet/builder/RawTxFinalizerTest.scala
benthecarman: that's why we generate 20 of them

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
benthecarman: Fixed, turns out don't need this after the bug fix

Issue #1655, opened 2020-07-10
 Cannot send a message to our peer when we are not connected!


Christewart:
**Describe the bug**
I am starting my neutrino node on  8b85751b461d2ec4151f12265ae4fb2cf9fe40c9 and I run into this error message: 

### First attempt at running
>./bitcoin-s-server
2020-07-10T13:41:53 INFO [HTTP] Initialized chain api
2020-07-10T13:41:54 INFO [WALLET] Using pre-existing wallet
2020-07-10T13:41:55 INFO [P2P] Sending getdata=GetDataMessage(inventoryCount=1, inventories=Vector(Inventory(MsgBlock,DoubleSha256Digest(b639ca945ef5833f8254988969906555654d27da080015000724210...) to peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333)
2020-07-10T13:41:55 DEBUG [P2P] Sending msg=getdata to peer=neutrino.testnet3.suredbits.com/34.210.154.40:18333
2020-07-10T13:41:55 ERROR [P2P] Cannot send a message to our peer when we are not connected! payload=GetDataMessage(inventoryCount=1, inventories=Vector(Inventory(MsgBlock,DoubleSha256Digest(b639ca945ef5833f8254988969906555654d27da080015000724210...) peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333)
2020-07-10T13:41:55 INFO [P2P] Starting node
2020-07-10T13:41:57 INFO [P2P] Attempting to connect to peer=neutrino.testnet3.suredbits.com/34.210.154.40:18333
2020-07-10T13:41:57 DEBUG [P2P] Tcp connection to: neutrino.testnet3.suredbits.com/34.210.154.40:18333
2020-07-10T13:41:57 DEBUG [P2P] Local: /192.168.1.187:45212
2020-07-10T13:41:57 INFO [P2P] Connection established with peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333)
2020-07-10T13:41:57 DEBUG [P2P] Sending msg=version to peer=neutrino.testnet3.suredbits.com/34.210.154.40:18333
2020-07-10T13:41:57 INFO [P2P] We've been disconnected by Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) command=PeerClosed
[...]

PR #1682, opened 2020-07-17
Start server before sync

benthecarman:
Before we would need to wait to be fully synced to start the rpc server, this is annoying because we can't use something like `getblockcount` to see the progress. This fixed that.


Christewart (approved):
This may be part of the failures I've detailed in #1655 (how the block count I sent to bitcoind is 0) but it doesn't seem to fix all of them.

Issue #1284, opened 2020-03-30
Bitcoin-S Wallet API Design Discussion


nkohen:

nkohen:
Current interface is:

- `getbalance`
- `getnewaddress`
- `sendtoaddress`
- `rescan`

nkohen:
Low hanging fruit (because they are already implemented in `WalletApi` but not `WalletRoutes`):

- separate calls for confirmed and unconfirmed balance
- `getUTXOs`
- `getAddresses`
- `getAccounts`
- `isEmpty`
- `getAddressInfo`
- `getNewAccount`

Christewart:
So I think we need to get rid of `LockedWallet` and `UnlockedWallet`. If anything, we should rename the APIs to be something that requires a `KeyManager` to use, and something that does not require a key manager to use. 

For instance, listing all addresses in our current wallet should not require a `KeyManager`. However, sending a transaction should require a `KeyManager`. 

An interesting case is generating fresh addresses, if we are using BIP32 it appears that we would not need a key manager to generate a new address, however if we _arent_ using BIP32, I think we would need a `KeyManager`

nkohen:
I think another related possibility is to have an ADT which has `KeyManager` as a subtype and `PublicKeyManager` as another for things like generating addresses (or perhaps `PublicKeyManager` should be a supertype of `KeyManager`). This way the wallet can just have a key or pubkey manager and act accordingly

benthecarman:
Would be nice to have APIs for users to be able to send by doing their own coin control. Also to go with this would be good to have a send max/whole utxo(s).

Kenny-zhao:
区块链企业钱包开放平台优盾钱包API接口官方文档：[https://www.uduncloud.com/gateway-interface](https://www.uduncloud.com/gateway-interface) 以供参考

Christewart:
So since we are getting close to releasing v0.4 I thought I would put some more thought into this: 

Currently our [`WalletApi`](https://github.com/bitcoin-s/bitcoin-s/blob/7abca65ac458c327726ef7077fbc55fe9a2f4777/wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala#L45) interface contains too much stuff in my opinion. I'm especially concerned with us having clashes with the `WalletApi` and key manager work that we would like to do in v0.5. Here are my current thoughts:

[...]

PR #1693, opened 2020-07-22
Create HDWalletApi

benthecarman:
Part of #1284 

Separates the `WalletApi` into `HDWalletApi`. This new trait has any of the calls that had was associated with an account. This should allow us to delegate key manager types better.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -92,7 +92,7 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
           confirmedBalance <- wallet.getConfirmedBalance()
           unconfirmedBalance <- wallet.getUnconfirmedBalance()
           addresses <- wallet.listAddresses()
-          utxos <- wallet.listUtxos()
+          utxos <- wallet.listUtxos(wallet.walletConfig.defaultAccount)
         } yield {
           (expectedConfirmedAmount == confirmedBalance) &&
           (expectedUnconfirmedAmount == unconfirmedBalance) &&
@@ -175,7 +175,7 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
 
       for {
         addresses <- wallet.listAddresses()
-        utxos <- wallet.listUtxos()
+        utxos <- wallet.listUtxos(wallet.walletConfig.defaultAccount)
         _ = assert(addresses.size == 6)
         _ = assert(utxos.size == 3)
 
@@ -189,14 +189,14 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
             .sendToAddress(address, TestAmount)
 
         addresses <- wallet.listAddresses()
-        utxos <- wallet.listUtxos()
+        utxos <- wallet.listUtxos(wallet.walletConfig.defaultAccount)
         _ = assert(addresses.size == 7)
         _ = assert(utxos.size == 3)
 
         _ <- wallet.clearAllUtxosAndAddresses()
 
         addresses <- wallet.listAddresses()
-        utxos <- wallet.listUtxos()
+        utxos <- wallet.listUtxos(wallet.walletConfig.defaultAccount)
         _ = assert(addresses.isEmpty)
         _ = assert(utxos.isEmpty)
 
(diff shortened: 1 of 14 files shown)


wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -162,7 +162,7 @@ case class WalletAppConfig(
       chainQueryApi: ChainQueryApi,
       feeRateApi: FeeRateApi,
       bip39PasswordOpt: Option[String])(implicit
-      ec: ExecutionContext): Future[WalletApi] = {
+      ec: ExecutionContext): Future[HDWalletApi] = {
```
nkohen, line 165: should probably be renamed `createHDWallet`

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -193,7 +193,7 @@ object WalletAppConfig
       feeRateApi: FeeRateApi,
       bip39PasswordOpt: Option[String])(implicit
       walletConf: WalletAppConfig,
-      ec: ExecutionContext): Future[WalletApi] = {
+      ec: ExecutionContext): Future[HDWalletApi] = {
```
nkohen, line 197: ditto

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -92,7 +92,7 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
           confirmedBalance <- wallet.getConfirmedBalance()
           unconfirmedBalance <- wallet.getUnconfirmedBalance()
           addresses <- wallet.listAddresses()
-          utxos <- wallet.listUtxos()
+          utxos <- wallet.listUtxos(wallet.walletConfig.defaultAccount)
```
nkohen: this pattern probably deserves a method

nkohen (approved):

wallet/src/main/scala/org/bitcoins/wallet/api/HDWalletApi.scala
```diff
@@ -0,0 +1,519 @@
...
+import scala.concurrent.Future
+
+/**
+  * API for the wallet project.
+  *
+  * This wallet API is BIP44 compliant.
+  *
+  * @see [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki BIP44]]
+  */
+trait HDWalletApi extends WalletApi {
```
Christewart, line 26: Is there any way we can minimize the number of overloaded methods in here? There is like 4 different versions of `sendFromOutPoints`, `sendWithAlgo` etc

wallet/src/main/scala/org/bitcoins/wallet/api/HDWalletApi.scala
```diff
@@ -0,0 +1,519 @@
...
+  def listAddresses(account: HDAccount): Future[Vector[AddressDb]]
+
+  def listSpentAddresses(account: HDAccount): Future[Vector[AddressDb]]
+
+  def listFundedAddresses(
+      account: HDAccount): Future[Vector[(AddressDb, CurrencyUnit)]]
+
+  def listUnusedAddresses(account: HDAccount): Future[Vector[AddressDb]]
+
+  def clearUtxosAndAddresses(account: HDAccount): Future[WalletApi]
```
Christewart: Probably should return a `HDWalletApi` rather than `WalletApi`?

wallet/src/main/scala/org/bitcoins/wallet/api/HDWalletApi.scala
```diff
@@ -0,0 +1,519 @@
...
+  /**
+    * Tries to create a new account in this wallet. Fails if the
+    * most recent account has no transaction history, as per
+    * BIP44
+    *
+    * @see [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account BIP44 account section]]
+    */
+  def createNewAccount(
+      hdAccount: HDAccount,
+      keyManagerParams: KeyManagerParams): Future[Wallet]
```
Christewart: return type should probably be `Future[HDWalletApi]`?

wallet/src/main/scala/org/bitcoins/wallet/api/HDWalletApi.scala
benthecarman: Would like to but we can't set default arguments because the function is overloaded.

Christewart (approved):

Issue #1696, opened 2020-07-25
scala.concurrent.Future behavior between Scala 2.13 and Scala 2.12


Christewart:
In light of https://github.com/scala/bug/issues/12089 it appears that the behavior of `Future` has changed substantially between Scala 2.12 and 2.13. 

The TLDR is that nested Futures won't necessarily run in parallel. This is highlighted with `Future.sequence` where each future will run synchronously after one another rather than in parallel. Other things that were of note in this thread seems to be the use of `scala.concurrent.blocking` which says: 

```
Used to designate a piece of code which potentially blocks, allowing the current [[BlockContext]] to adjust
   *  the runtime's behavior.
   *  Properly marking blocking code may improve performance or avoid deadlocks.
```

It seems that this should ALSO be used for pieces of code that can take a long time to run. If we don't use `blocking` for code that runs for long periods of time it can cause thread starvation. I'm thinking rescans is where this can be useful, and our `SafeDatabase.runVec()` which runs all of our blocking JDBC database code.

PR #1697, opened 2020-07-25
2020 07 25 optimize recalc chainwork

Christewart:
Built on top of #1710 

The primary purpose of this PR is to optimize chain work recalculation.

The strategy for this PR was to exploit the fact that sqlite allows multiple parallel reads at a time. Now we create smaller batches of reads. This PR takes our `batchSize` and then breaks it out into `numBatches` to read from the database. 

Unfortunately, according to this [stackoverflow post](https://stackoverflow.com/a/23350768/967713) sqlite seems to only allow concurrent writes if have you a special configuration. Therefore I do not write to the database in a concurrent fashion on this PR. This means the bottleneck is writing to the database. 

### Usage of `scala.concurrent.blocking`

In #1696 I hint at using `scala.concurrent.blocking` to notify the runtime that we might be running a `Future` that will take a long time to complete. I've added `scala.concurrent.blocking` in two places on this PR:

1. Inside of [`SafeDatabase.runVec()`](https://github.com/bitcoin-s/bitcoin-s/commit/9b7e9ec4b6258d4c06005576df48081f4e59d4b3#diff-ccdce0133b19da94e06016c74f8af488R2050) as this is where are blocking jdbc database calls are
2. ~Around the result of parallel fetches from the database reading headers [here](https://github.com/bitcoin-s/bitcoin-s/commit/9b7e9ec4b6258d4c06005576df48081f4e59d4b3#diff-4ec0be00e8ef523c47423cedac277783R571)~ (with #1698 this is a big speedup, so this usage of `scala.concurrent.blocking` doesn't seem to make a performance difference and doesn't cause thread starvation when i remove it)

This will mean a unique thread will get spun up to actually execute these which will keep them from being run directly on our main akka dispatcher. I'm not 100% confident that this is used being correctly so I'm open to suggestions.

### Writing to the database synchronously

While we read small batches in parallel from the database, it isn't possible to write small batches to the database in parallel as [sqlite does not support parallel writes out of the box](https://stackoverflow.com/a/23350768/967713). This means this is the current bottleneck of our implementation.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -675,12 +675,8 @@ case class ChainHandler(
   private def calculateChainWorkGenesisBlock(
       genesisHeader: BlockHeaderDb): Future[BlockHeaderDb] = {
     val expectedWork = Pow.getBlockProof(genesisHeader.blockHeader)
-    if (genesisHeader.chainWork != expectedWork) {
-      val genesisWithWork = genesisHeader.copy(chainWork = expectedWork)
-      blockHeaderDAO.update(genesisWithWork)
-    } else {
-      Future.successful(genesisHeader)
-    }
+    val genesisWithWork = genesisHeader.copy(chainWork = expectedWork)
+    blockHeaderDAO.update(genesisWithWork)
   }
 }
 
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -18,6 +18,7 @@ import org.bitcoins.crypto.{
   DoubleSha256DigestBE
 }
 
+import scala.annotation.tailrec
 import scala.concurrent._
 
 /**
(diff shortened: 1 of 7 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -502,87 +503,158 @@ case class ChainHandler(
...
-          val updated = accum :+ newHeader
-          // updated the latest batch
-          blockHeaderDAO
-            .upsertAll(updated.takeRight(batchSize))
-            .flatMap(_ =>
-              loop(
-                remainingHeaders.tail,
-                updated.takeRight(batchSize)
-              ))
+  /** Creates [[numBatches]] of requqests to the database fetching [[batchSize]] headers
```
benthecarman: > requqests

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -502,87 +503,158 @@ case class ChainHandler(
...
+    } else {
+      val header = remainingHeaders.head
+
+      val currentChainWork = {
+        accum.lastOption.map(_.chainWork) match {
+          case Some(prevWork) =>
+            prevWork
+          case None =>
+            // this should be the case where the accum is
+            //empty, so this header is the last one we hvae
```
benthecarman: > hvae

Christewart (commented):
Seem to work for me on #1697

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -500,87 +503,161 @@ case class ChainHandler(
...
+  def recalculateChainWork: Future[ChainHandler] = {
+    logger.info("Calculating chain work for previous blocks")
+
+    val maxHeightF = blockHeaderDAO.maxHeight
+    val startHeightF = blockHeaderDAO.getLowestNoWorkHeight
+    val startF = for {
+      startHeight <- startHeightF
+      headers <- {
         if (startHeight == 0) {
-          Future.successful(Vector.empty)
```
benthecarman, line 576: We want to be able to recalc on the gensis header, if you have an old db it will be `0`

benthecarman:
Looks like git history got messed up too

Christewart:
> Looks like git history got messed up too

Should be fixed, all work is in 49cf938cbaccb9bf9d5306c93cad13a43d77a7fa

benthecarman:
Will need to take a closer look but there is something wrong with the logic, previous chain work on block 1781240 was
```
000000000000000000000000000000000000000000000001b6e6917ea82bd99b5d
``` 
after doing a chainwork on this pr it was
```
000000000000000000000000000000000000000000000001b7bf310ad303a93e66
```

Also, comparing to bitcoind they are not the same:
block 1781616:
this pr:
```
0000000000000000000000000000000000000000000001b93224d24d2da8d919
```
bitcoind:
```
0000000000000000000000000000000000000000000001b85b16f421222130cc
```

Christewart:
> Will need to take a closer look but there is something wrong with the logic, previous chain work on block 1781240 was
> 
> ```
> 000000000000000000000000000000000000000000000001b6e6917ea82bd99b5d
> ```
> 
> after doing a chainwork on this pr it was
> 
> ```
> 000000000000000000000000000000000000000000000001b7bf310ad303a93e66
> ```
> 
> Also, comparing to bitcoind they are not the same:
> block 1781616:
> this pr:
> 
> ```
> 0000000000000000000000000000000000000000000001b93224d24d2da8d919
> ```
> 
> bitcoind:
> 
> ```
> 0000000000000000000000000000000000000000000001b85b16f421222130cc
> ```

So i expect this is something to do with the parallelization logic. I reduced `numBatches` to 1 in 65f423c as with #1698 it doesn't seem like the database is as big of a bottleneck. Even weirder, I think I am misunderstanding something as it doesn't seem like we have multiple database threads running at the same time when i up `numBatches`, at least from what I can tell on visualvm.

To be clear, `upsertAll()` is still taking all of our time, but as I said earlier in this thread it appears we can't concurrently write to sqlite :-(. Postgres would go much faster!

Screenshot from 2020-07-26 08-17-52

Christewart:
So looking into the wrong numbers bug a bit more: 

It seems like we recalc all headers just fine up at least height 1,500,000

>sqlite> select hash from block_headers where height = 1500000 ;
0000000000049a6b07f91975568dc96bb1aec1a24c6bdadb21eb17c9f1b7256f
sqlite> select hex("chain_work")from block_headers where height = 1500000;
0000000000000000000000000000000000000000000000010BD369E987B8FE260D

bitcoind output: 

> bitcoin-cli getblockheader 0000000000049a6b07f91975568dc96bb1aec1a24c6bdadb21eb17c9f1b7256f
{
  "hash": "0000000000049a6b07f91975568dc96bb1aec1a24c6bdadb21eb17c9f1b7256f",
  "confirmations": 281728,
  "height": 1500000,
  "version": 536870912,
  "versionHex": "20000000",
  "merkleroot": "a707c2a95e5cce2d24cda092d87df99deba85fd1c3e0e614f0b4dc296c33a52c",
  "time": 1555736999,
  "mediantime": 1555736991,
  "nonce": 1374035913,
  "bits": "1b3f7280",
  "difficulty": 1032.905033092196,
  "chainwork": "00000000000000000000000000000000000000000000010bd369e987b8fe260d",
  "nTx": 2,
  "previousblockhash": "000000000004c98f51d68a4837215014760d79c091dd5ce95cc1013e5e19780d",
  "nextblockhash": "000000000000c9982537a5535dc6f9c08db80b6f72496c55f5e17bc6998ab6c3"
}


however as you state there seems to be some sort of corruption that happens later in the calculation which seems extremely bizarre to me. How can we calculate 1.5M header's chain work correctly and then botch it later on?

Here is the results for header at height 1,750,000: 

>sqlite> select hash from block_headers where height = 1750000 ;
0000000023b901d13c0dcaa89e9db08a0f81a6c5d50ebcd2cf4e2614465e7822
sqlite> select hex("chain_work") from block_headers where height = 1750000 ;
0000000000000000000000000000000000000000000000015BB802B04B4B2744AC


bitcoind: 
> bitcoin-cli getblockheader 0000000023b901d13c0dcaa89e9db08a0f81a6c5d50ebcd2cf4e2614465e7822
{
  "hash": "0000000023b901d13c0dcaa89e9db08a0f81a6c5d50ebcd2cf4e2614465e7822",
  "confirmations": 31728,
  "height": 1750000,
  "version": 536870912,
  "versionHex": "20000000",
  "merkleroot": "ce38c8f7739d5f8450b1701a00bec5fe4657469b835f26a68cec31a1c8841e69",
  "time": 1591469994,
  "mediantime": 1591469969,
  "nonce": 1411754007,
  "bits": "1c3fffc0",
  "difficulty": 4,
  "chainwork": "00000000000000000000000000000000000000000000015bb802b04a4b2644ab",
  "nTx": 1,
  "previousblockhash": "000000000ce6b4652f29a10ee400b5b166765680fc8b04f432346bec7aabdd07",
  "nextblockhash": "00000000129bbd6e527c37eed2ecf941fa86f03ef1a786dd0488ec1ba2ff5b74"
}

Christewart:
I'm guessing this problem is being caused like this because I have two headers at `1580960`

> select hash from block_headers where height = 1580960 ;
000000000003e674fd8ad391f508613c386e9019848b4120b9c82cb0e65b35be
000000006999656106c726515ccfc34d160a5fa299ddb6bb278598b2feefaa7e

So it seems that this is the incorrect piece of logic: 

https://github.com/bitcoin-s/bitcoin-s/pull/1697/files#diff-4ec0be00e8ef523c47423cedac277783R543

We can't just arbitrarily search for heights in the database when we need to actually make sure headers reference each other.

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -500,87 +503,185 @@ case class ChainHandler(
...
+          val genesisHeaderF = blockHeaderDAO.getAtHeight(0)
+          genesisHeaderF.flatMap { h =>
+            require(h.length == 1, s"Shoudl only have one genesis header!")
```
benthecarman: > Shoudl

Should

benthecarman:
I tried adding this test to make sure having multiple headers at a height doesn't cause the wrong chain work calc, but it is failing. The test (put in `ChainHandlerTest`):

```
  it must "properly recalculate chain work with duplicate heights" in {
    tempHandler: ChainHandler =>
      val duplicate = BlockHeader(
        version = Int32.one,
        previousBlockHash = ChainTestUtil.blockHeader562462.hash,
        merkleRootHash = DoubleSha256Digest.empty,
        time = UInt32.zero,
        nBits = UInt32.zero,
        nonce = UInt32.zero
      )

      val chain1 = Vector(
        BlockHeaderDbHelper.fromBlockHeader(3,
                                            BigInt(0),
                                            ChainTestUtil.blockHeader562464),
        BlockHeaderDbHelper.fromBlockHeader(2,
                                            BigInt(0),
                                            ChainTestUtil.blockHeader562463),
        BlockHeaderDbHelper.fromBlockHeader(1,
                                            BigInt(0),
                                            ChainTestUtil.blockHeader562462)
      )

      val chain2 = Vector(
        BlockHeaderDbHelper.fromBlockHeader(2, BigInt(0), duplicate),
        BlockHeaderDbHelper.fromBlockHeader(1,
                                            BigInt(0),
                                            ChainTestUtil.blockHeader562462)
      )

      val headersWithNoWork = (chain1 ++ chain2).distinct

      val chains = Vector(Blockchain(chain1), Blockchain(chain2))

      val chainHandler = tempHandler.copy(blockchains = chains)

      for {
        _ <- chainHandler.blockHeaderDAO.createAll(headersWithNoWork)
        isMissingWork <- chainHandler.isMissingChainWork
        _ = assert(isMissingWork)
        newHandler <- chainHandler.recalculateChainWork
        duped <-
          newHandler.blockHeaderDAO.findByHash(duplicate.hashBE).map(_.get)
        best <- newHandler.getBestBlockHeader()
      } yield {
        assert(duped.chainWork == chain2.foldLeft(genesis.chainWork)((work, headerDb) =>
          work + Pow.getBlockProof(headerDb.blockHeader)))

        assert(best.chainWork == chain1.foldLeft(genesis.chainWork)((work, headerDb) =>
          work + Pow.getBlockProof(headerDb.blockHeader)))
      }
  }
```

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -500,87 +503,185 @@ case class ChainHandler(
...
+        _ = headersToCalc.headOption.map { h =>
+          logger.info(
+            s"Recalculating chain work... current height: ${h.height} maxHeight=${maxHeight}")
+        }
+        headersWithWork = {
+          headersToCalc.map { chain =>
+            calcChainWork(remainingHeaders = chain.headers.sortBy(_.height),
+                          accum = Vector.empty,
+                          lastHeaderWithWorkInDb = lastHeader)
+          }.flatten
```
benthecarman: can just use `flatMap`

Christewart:
We should probably add documentation on the `--force-recalc` flag (I propose we rename to `--force-recalc-chainwork`) https://bitcoin-s.org/docs/next/config/configuration

Also that would mean we add the ability configure it via `bitcoin-s.chain`. Also we should probably consider documenting all of our command line flags somwhere which I don't think is done?

benthecarman:
> Also that would mean we add the ability configure it via `bitcoin-s.chain`. 

I don't think this makes sense, you shouldn't need to touch your config file every time you run the application. If that holds true, then the only reason you'd have the recalc chain work option enabled in your config file is because you want to run it every time which makes no practical sense. 

> Also we should probably consider documenting all of our command line flags somwhere which I don't think is done?

Good for a separate PR

> We should probably add documentation on the `--force-recalc` flag (I propose we rename to `--force-recalc-chainwork`) [bitcoin-s.org/docs/next/config/configuration](https://bitcoin-s.org/docs/next/config/configuration)

Can go with PR above ^

Christewart:
> > Also that would mean we add the ability configure it via `bitcoin-s.chain`.
> 
> I don't think this makes sense, you shouldn't need to touch your config file every time you run the application. If that holds true, then the only reason you'd have the recalc chain work option enabled in your config file is because you want to run it every time which makes no practical sense.

I think bitcoin core allows you to do this with `-reindex` which the same thing applies. I guess I think what doesn't make sense is not having continuity between the config file and command line options. We don't have this as it exists, but I think it's something we should strive for. Basically you should be able to configure something from the command line OR from the config file, with the command line taking precedent

Christewart:
I ran chainwork recalc on  2c6bbc4226d95abec96e677bfbbe7928659cc4f3 and got the correct answer for the latest header `000000007e26a03221c0705e3a2ccee73b931b7a336a6cdafd1deb0ddb199b39` and `height=1781689`

The chain work: `000000000000000000000000000000000000000000000001b87e6b1fa84ad9a134`

chain/src/main/scala/org/bitcoins/chain/config/ChainAppConfig.scala
```diff
@@ -98,6 +98,10 @@ case class ChainAppConfig(
   lazy val filterBatchSize: Int =
     config.getInt(s"${moduleName}.neutrino.filter-batch-size")
 
+  lazy val forceRecalcChainWork: Boolean =
+    config.getBooleanOrElse(s"$moduleName.neutrino.filter-batch-size",
```
Christewart: You didn't change `filter-batch-size` in the string

Christewart: Also we should probably at least add it to the example configurations in this PR otherwise it is soooo easy to forget.

PR #1703, opened 2020-07-27
Create getBlockchainsBetweenHeights for BlockHeaderDAO

### LINKED ISSUE #1701: Implement `BlockHeaderDAO.getBlockchainsBetweenHeights()`
This feature is needed for https://github.com/bitcoin-s/bitcoin-s/pull/1697#issuecomment-664377596 in #1697 

It is very tempting to use our [`BlockHeaderDAO.getBetweenHeights()`](https://github.com/bitcoin-s/bitcoin-s/blob/6a1e4aea0163fc1a15c8b0a415db959bf9f357bb/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala#L195) method for fetching a blockchain, but this is NOT what the method does. 

We should implement a `BlockHeaderDAO.getBlockchainsBetweenHeights()` to make sure that we return headers that are actually a real blockchain rather just a set of headers between two given heights.

benthecarman:
Closes #1701

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -343,35 +343,10 @@ case class BlockHeaderDAO()(implicit
       } else {
         val headersByHeight = headers.groupBy(_.height).toVector
         val sortedHeaders = headersByHeight.sortBy(_._1).reverse.map(_._2)
-        val chains = makeChains(sortedHeaders, Vector.empty)
-        chains.map(Blockchain(_))
-      }
-    }
-  }
-
-  @tailrec
-  private def makeChains(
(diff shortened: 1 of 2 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -334,6 +334,48 @@ case class BlockHeaderDAO()(implicit
...
+        val headersByHeight = headers.groupBy(_.height).toVector
+        val sortedHeaders = headersByHeight.sortBy(_._1).reverse.map(_._2)
+        val chains = makeChains(sortedHeaders, Vector.empty)
+        chains.map(Blockchain(_))
+      }
+    }
+  }
+
+  @tailrec
+  private def makeChains(
```
Christewart: Could you just use `BaseBlockchain.reconstructFromHeaders()`?

PR #1708, opened 2020-07-27
Move Node type of out Wallet API

benthecarman:
Part of #1284 

Built on #1707 because there would be a lot of merge conflicts

This moves any call out of the wallet api that had to do with if it was a neutrino node or a spv node. Creates separate types for each.

Creates types for HDWallets that are neutrino or spv, and a `AnyHDWallet` that is both neutrino and spv (maybe can come up with a better name).

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/api/NeutrinoWalletApi.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/api/NeutrinoWalletApi.scala
@@ -9,7 +9,7 @@ import org.bitcoins.wallet.api.NeutrinoWalletApi.BlockMatchingResponse
 
 import scala.concurrent.{ExecutionContext, Future}
 
-trait NeutrinoWalletApi extends WalletApi {
+trait NeutrinoWalletApi { self: WalletApi =>
 
   /**
     * Processes the give block, updating our DB state if it's relevant to us.
--- a/wallet/src/main/scala/org/bitcoins/wallet/api/NeutrinoWalletApi.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/api/NeutrinoWalletApi.scala
@@ -1,0 +1,101 @@
+package org.bitcoins.wallet.api
+
+import org.bitcoins.core.gcs.GolombFilter
+import org.bitcoins.core.protocol.BlockStamp
+import org.bitcoins.core.protocol.blockchain.Block
+import org.bitcoins.core.protocol.script.ScriptPubKey
+import org.bitcoins.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.wallet.api.NeutrinoWalletApi.BlockMatchingResponse
+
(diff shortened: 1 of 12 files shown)


wallet/src/main/scala/org/bitcoins/wallet/api/NeutrinoWalletApi.scala
```diff
@@ -0,0 +1,99 @@
...
+import org.bitcoins.core.gcs.GolombFilter
+import org.bitcoins.core.protocol.BlockStamp
+import org.bitcoins.core.protocol.blockchain.Block
+import org.bitcoins.core.protocol.script.ScriptPubKey
+import org.bitcoins.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.wallet.api.NeutrinoWalletApi.BlockMatchingResponse
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait NeutrinoWalletApi extends WalletApi {
```
Christewart: So should this really `extends WalletApi` or should it require a `WalletApi` to be injected into this trait with [scala's self type annotation](https://www.baeldung.com/scala/self-type-annotation), i.e. 

```scala
trait NeutrinoWalletApi { self : WalletApi => ... } 
```

As I think we have discussed online, it doesn't seem like the neutrino APIs are directly related to the wallet, rather they are a synchronization mechanism for _any_ wallet type.

wallet/src/main/scala/org/bitcoins/wallet/api/SpvWalletApi.scala
```diff
@@ -0,0 +1,26 @@
...
+import scala.concurrent.Future
+
+/**
+  * API for the wallet project.
+  *
+  * This wallet API is BIP44 compliant.
+  *
+  * @see [[https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki BIP44]]
+  */
+trait SpvWalletApi extends WalletApi {
```
Christewart: Same is true here, I believe. Also perhaps we want to move these into the core project eventually too? :/

wallet/src/main/scala/org/bitcoins/wallet/api/WalletApi.scala
```diff
@@ -447,10 +354,11 @@ trait WalletApi extends WalletLogger {
...
-object WalletApi {
+trait NeutrinoHDWalletApi extends HDWalletApi with NeutrinoWalletApi
 
-  case class BlockMatchingResponse(
-      blockHash: DoubleSha256DigestBE,
-      blockHeight: Int)
+trait SpvHDWalletApi extends HDWalletApi with SpvWalletApi
 
-}
+trait AnyHDWalletApi
```
Christewart, line 338: Ok, i'm confused as to what the purpose of this is. Can we get a nice description for a scaladoc?

Christewart (changes requested):

nkohen (approved):
LGTM Assuming CI

Christewart (approved):

PR #1710, opened 2020-07-28
Fix getBlockchainsBetweenHeights

benthecarman:
Fixed to use `Blockchain.connectWalkBackwards` to connect every possible chain 

Fixed the test as well to make sure that `chains` are non-empty and that all of the expected chains are found.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -334,43 +334,47 @@ case class BlockHeaderDAO()(implicit
     headersF.map(headers => Blockchain.fromHeaders(headers.reverse))
   }
 
+  @tailrec
+  private def loop(
+      chains: Vector[Blockchain],
+      allHeaders: Vector[BlockHeaderDb]): Vector[Blockchain] = {
+    val usedHeaders = chains.flatMap(_.headers).distinct
+    val diff = allHeaders.filter(header =>
+      !usedHeaders.exists(_.hashBE == header.hashBE))
+    if (diff.isEmpty) {
+      chains
+    } else {
+      val sortedDiff = diff.sortBy(_.height)(Ordering.Int.reverse)
+
+      val newChainHeaders =
+        Blockchain.connectWalkBackwards(sortedDiff.head, allHeaders)
+      val newChain = Blockchain(
+        newChainHeaders.sortBy(_.height)(Ordering.Int.reverse))
+      loop(chains :+ newChain, allHeaders)
+    }
+  }
+
   /** Retrieves a blockchain with the best tip being the given header */
   def getBlockchainsBetweenHeights(from: Int, to: Int)(implicit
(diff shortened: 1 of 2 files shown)


benthecarman:
Running the chainwork recalc with this patch and the prev iteration both didn't fix the resulting chainwork to be correct. Need to dig into if the problem is with this pr or in #1697

rorp (approved):

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -339,14 +339,38 @@ case class BlockHeaderDAO()(implicit
       ec: ExecutionContext): Future[Vector[Blockchain]] = {
     getBetweenHeights(from = from, to = to).map { headers =>
       if (headers.map(_.height).distinct.size == headers.size) {
```
Christewart, line 361: Some preliminary optimziations as `getBlockchainsBetweenHeights()` seems like it can have some pretty pathogical behavior in the case of forks (especially large ranges of heights!) 

```scala

  @tailrec
  private def loop(
      chains: Vector[Blockchain],
      allHeaders: Vector[BlockHeaderDb]): Vector[Blockchain] = {
    val usedHeaders = chains.flatMap(_.headers).distinct
    val diff = allHeaders.filter(header =>
      !usedHeaders.exists(_.hashBE == header.hashBE))
    if (diff.isEmpty) {
      chains
    } else {
      val sortedDiff = diff.sortBy(_.height).reverse

      val newChainHeaders =
        Blockchain.connectWalkBackwards(sortedDiff.head, allHeaders)
      val newChain = Blockchain(newChainHeaders.sortBy(_.height).reverse)
      loop(chains :+ newChain, allHeaders)
    }
  }

  /** Retrieves a blockchain with the best tip being the given header */
  def getBlockchainsBetweenHeights(from: Int, to: Int)(implicit
      ec: ExecutionContext): Future[Vector[Blockchain]] = {
    getBetweenHeights(from = from, to = to).map { headers =>
      val distinctHeights = headers.map(_.height).distinct
      if (distinctHeights.size == headers.size) {
        val headersByHeight = headers.sortBy(_.height)
        Vector(Blockchain.fromHeaders(headersByHeight.reverse))
      } else {
        val headersByHeight: Vector[(Int, Vector[BlockHeaderDb])] =
          headers.groupBy(_.height).toVector
        val tips: Vector[BlockHeaderDb] = {
          headersByHeight.sortBy(_._1).last._2
        }
        val allHeaders = headers

        val chains = tips.map { tip =>
          Blockchain
            .connectWalkBackwards(tip, allHeaders)
            .sortBy(_.height)
            .reverse
        }
        val init = chains.map(Blockchain(_)).distinct

        loop(init, allHeaders).distinct
      }
    }
  }
```

benthecarman: Applied these plus some others

Christewart (approved):

PR #1711, opened 2020-07-28
Fixed getblockchaininfo for v19

### LINKED ISSUE #1709: Could not parse getblockchaininfo on v0.19.0.1 on bitcoind rpc
This happened on our suredbits-api when spinning up a regtest node to back our api. 

>[info] 2020-07-28 07:59:01,902 ERROR Main.scala:254 - Failed to start the suredbits api Could not parse JsResult: {"chain":"regtest","blocks":101,"headers":101,"bestblockhash":"722a93adc0cbf1f444dd2d67542e3e938478d5ef60ca567b36fc02ba2eb8da06","difficulty":4.656542373906925E-10,"mediantime":1589158464,"verificationprogress":1,"initialblockdownload":true,"chainwork":"00000000000000000000000000000000000000000000000000000000000000cc","size_on_disk":30575,"pruned":false,"softforks":{"bip34":{"type":"buried","active":false,"height":500},"bip66":{"type":"buried","active":false,"height":1251},"bip65":{"type":"buried","active":false,"height":1351},"csv":{"type":"buried","active":false,"height":432},"segwit":{"type":"buried","active":true,"height":0},"testdummy":{"type":"bip9","bip9":{"status":"defined","start_time":0,"timeout":9223372036854775807,"since":0},"active":false}},"warnings":""}! Error: Error when parsing result of 'getblockchaininfo': {"obj.softforks":[{"msg":["error.expected.jsarray"],"args":[]}],"obj.bip9_softforks":[{"msg":["error.path.missing"],"args":[]}]}!

nkohen:
Fixes #1709

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
@@ -52,7 +52,23 @@ case class GetBlockWithTransactionsResult(
     nextblockhash: Option[DoubleSha256DigestBE])
     extends BlockchainResult
 
-case class GetBlockChainInfoResult(
+sealed trait GetBlockChainInfoResult extends BlockchainResult {
+  def chain: NetworkParameters
+  def blocks: Int
+  def headers: Int
+  def bestblockhash: DoubleSha256DigestBE
+  def difficulty: BigDecimal
+  def mediantime: Int
+  def verificationprogress: BigDecimal
+  def initialblockdownload: Boolean
+  def chainwork: String // How should this be handled?
+  def size_on_disk: Long
+  def pruned: Boolean
+  def pruneheight: Option[Int]
+  def warnings: String
+}
+
+case class GetBlockChainInfoResultPreV19(
     chain: NetworkParameters,
     blocks: Int,
     headers: Int,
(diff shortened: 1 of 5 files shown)


Christewart (approved):

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
```diff
@@ -52,7 +52,23 @@ case class GetBlockWithTransactionsResult(
...
+sealed trait GetBlockChainInfoResult extends BlockchainResult {
+  def chain: NetworkParameters
+  def blocks: Int
+  def headers: Int
+  def bestblockhash: DoubleSha256DigestBE
+  def difficulty: BigDecimal
+  def mediantime: Int
+  def verificationprogress: BigDecimal
+  def initialblockdownload: Boolean
+  def chainwork: String // How should this be handled?
```
benthecarman, line 64: nit: maybe now a good time to change this to a big int

benthecarman (approved):

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
nkohen: So this is unrelated to db stuff that's been going on and is about the fact that the bitcoind rpc returns a string with quotes around it (presumably to maintain precision). This can either be handled by changing the `Reads[BigInt]` to accept either a number or a `String` or to do something custom with just this `case class`, either way I think it would be better in a different PR

PR #1715, opened 2020-07-28
Optimize getBestFilter functions

### LINKED ISSUE #1700: CompactFilterHeaderDAO.getBestFilterHeader and CompactFilterDAO.getBestFilter are exceptionally slow.
It takes exponential time to sync compact header in  because iterate `CompactFilterHeaderDAO.getBestFilterHeader` and `CompactFilterDAO.getBestFilter` iterate over whole blockchain several times.

Most likely it happens because those methods use complex client-side computations along with database-side queries, so slick cannot generate performant SQL.

### LINKED ISSUE #1719: Make chain_work column use VARCHAR sqlite data type rather tha VARBINARY
As @rorp states on slack, it is strange we use VARCHAR for everything besides the chain_work column in the headers table. Originally @benthecarman thought we would be able to get around weird issues with padding the uint256 number in the database to be able to correctly interpret it coming out of the db

Unfortunately it didn't seem to fix these issues which is why we needed #1651

benthecarman:
Closes #1700 
Closes #1719 

Reduces the number of database reads needed from 2 -> 1. I think this is the most optimal sql query we can use to get the filter with the most work. 

This is the resulting sql query 
``` 
select s13."hash", s13."filter_type", s13."bytes", s13."height", s13."block_hash" from "cfilters" s13, "block_headers" s14 where (s14."chain_work" = (select max(s21."chain_work") from "cfilters" s20, "block_headers" s21 where s20."block_hash" = s21."hash")) and (s13."block_hash" = s14."hash") limit 1
```

### DIFF
--- a/chain/src/main/resources/sqlite/chain/migration/V4__chain_work_to_varchar.sql
+++ b/chain/src/main/resources/sqlite/chain/migration/V4__chain_work_to_varchar.sql
@@ -1,0 +1,9 @@
+-- This block changes the chain work column from a VARBINARY to a VARCHAR
+CREATE TABLE "block_headers_temp" ("height" INTEGER NOT NULL,"hash" VARCHAR(254) PRIMARY KEY NOT NULL,"version" INTEGER NOT NULL,"previous_block_hash" VARCHAR(254) NOT NULL,"merkle_root_hash" VARCHAR(254) NOT NULL,"time" INTEGER NOT NULL,"n_bits" INTEGER NOT NULL,"nonce" INTEGER NOT NULL,"hex" VARCHAR(254) NOT NULL, "chain_work" VARCHAR(33) NOT NULL DEFAULT "0000000000000000000000000000000000000000000000000000000000000000");
+INSERT INTO "block_headers_temp" SELECT "height", "hash", "version", "previous_block_hash", "merkle_root_hash", "time", "n_bits", "nonce", "hex", hex("chain_work") FROM "block_headers";
+DROP TABLE "block_headers";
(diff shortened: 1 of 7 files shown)


nkohen (approved):

Christewart (approved):

rorp (changes requested):
It doesn't fix full table scans for `MAX(chain_work)` and `WHERE chain_work = ?`

benthecarman:
Added an index for the chain work, should be much faster now

chain/src/main/resources/sqlite/chain/migration/V4__chain_work_to_varchar.sql
```diff
@@ -0,0 +1,9 @@
+-- This block changes the chain work column from a VARBINARY to a VARCHAR
+CREATE TABLE "block_headers_temp" ("height" INTEGER NOT NULL,"hash" VARCHAR(254) PRIMARY KEY NOT NULL,"version" INTEGER NOT NULL,"previous_block_hash" VARCHAR(254) NOT NULL,"merkle_root_hash" VARCHAR(254) NOT NULL,"time" INTEGER NOT NULL,"n_bits" INTEGER NOT NULL,"nonce" INTEGER NOT NULL,"hex" VARCHAR(254) NOT NULL, "chain_work" VARCHAR(33) NOT NULL DEFAULT "0000000000000000000000000000000000000000000000000000000000000000");
```
Christewart, line 2: I think we need postgres too?

rorp (approved):
👍

PR #1722, opened 2020-07-29
Have RawTxSigner use low R signing

benthecarman:
Doing this will have the wallet use low r signing, this will save the user on on average 1 byte per tx.

Needed to change the psbt unit test to use low r signing as well to pass.


nkohen (approved):
ACK but we probably can't merge this until macOS binary for secp is built and pushed into master

Christewart (approved):

PR #1723, opened 2020-07-30
Fix Postgres types

rorp:
`VARCHAR(254)` is not enough to hold most of the compact filters.

Also the type of `wallet_address_tags.address` does not match the type of `addresses.address`.


benthecarman (approved):
ACK

benthecarman (commented):
You'll need to up the number of migrations in the `DbManagementTest`, otherwise ACK

benthecarman (approved):

Christewart (approved):

PR #1726, opened 2020-07-30
ChainHandler: cache block and filter header counts

rorp:
https://github.com/bitcoin-s/bitcoin-s/pull/1715 speeds up the sync process a lot, but filter sync is still too slow for Postgres.
The block count and the filter count don't change (or change once in 10 minutes) during the filter sync, so it makes sense to cache them.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -44,12 +44,20 @@ case class ChainHandler(
     extends ChainApi
     with ChainVerificationLogger {
 
+  @volatile private var blockCount: Option[Int] = None
+  @volatile private var filterHeaderCount: Option[Int] = None
+
   /** @inheritdoc */
   override def getBlockCount(): Future[Int] = {
-    logger.debug(s"Querying for block count")
-    blockHeaderDAO.bestHeight.map { height =>
-      logger.debug(s"getBlockCount result: count=$height")
-      height
+    cachedBlockCount match {
+      case Some(count) => Future.successful(count)
+      case None =>
+        logger.debug(s"Querying for block count")
+        blockHeaderDAO.bestHeight.map { height =>
+          logger.debug(s"getBlockCount result: count=$height")
+          cacheBlockCount(Some(height))
+          height
+        }
     }
   }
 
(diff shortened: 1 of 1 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -251,6 +259,7 @@ case class ChainHandler(
       val maxHeight = filterHeadersToCreate.maxBy(_.height)
       logger.info(
         s"Processed filters headers from height=${minHeight.height} to ${maxHeight.height}. Best hash=${maxHeight.blockHashBE}")
+      cacheFilterHeaderCount(None)
```
benthecarman, line 266: You need to do this for the block count and filter count as well

rorp: Every time `ChainHandler.processHeaders` processes a new block header it creates a new instance of `ChainHandler`: https://github.com/bitcoin-s/bitcoin-s/blob/master/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala#L113. So there's no need in resetting the block count cache.

benthecarman: Should update the filter count cache still

Christewart: > Every time `ChainHandler.processHeaders` processes a new block header it creates a new instance of `ChainHandler`: https://github.com/bitcoin-s/bitcoin-s/blob/master/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala#L113. So there's no need in resetting the block count cache.

This is an important design goal I had in mind, if we can "encapsulate" state changes by returning new interface types (in this case `Future[ChainApi]`) we can get away with internal mutability. However in certain parts of our codebase we discard the returned `ChainApi` ([here with wallet/node](https://github.com/bitcoin-s/bitcoin-s/blob/930f184dca69a3aa7512fc92a3db6833c5bfac6d/app/server/src/main/scala/org/bitcoins/server/Main.scala#L119-L120)). I really like the pattern of returning new `Api` types to signify a "state change" of the internal API, but I guess I haven't been explicit about it as a design goal.

Christewart:
CI failing: https://travis-ci.org/github/bitcoin-s/bitcoin-s/builds/713442829

PR #1728, opened 2020-07-30
Improved block parsing time by approx. 33%

nkohen:
Every time we encounter a P2PKH script signature in a block we were triggering a key decompression to validate the key, which was not necessary in part because typing script is already an imperfect thing, and also because other validation that script signatures should go through fail and so there is no need to fail so early as it costs so much time.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
@@ -126,14 +126,14 @@ object P2PKHScriptSignature extends ScriptFactory[P2PKHScriptSignature] {
   def isP2PKHScriptSig(asm: Seq[ScriptToken]): Boolean =
     asm match {
       case Seq(_: BytesToPushOntoStack,
-               _: ScriptConstant,
+               maybeSig: ScriptConstant,
                _: BytesToPushOntoStack,
-               z: ScriptConstant) =>
+               maybeKey: ScriptConstant) =>
         if (
-          (z.bytes.length == 33 || z.bytes.length == 65) && ECPublicKey
-            .isFullyValid(z.bytes)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -126,14 +126,14 @@ object P2PKHScriptSignature extends ScriptFactory[P2PKHScriptSignature] {
...
-               _: ScriptConstant,
+               maybeSig: ScriptConstant,
                _: BytesToPushOntoStack,
-               z: ScriptConstant) =>
+               maybeKey: ScriptConstant) =>
         if (
-          (z.bytes.length == 33 || z.bytes.length == 65) && ECPublicKey
-            .isFullyValid(z.bytes)
+          (maybeKey.bytes.length == 33 || maybeKey.bytes.length == 65) &&
+          (maybeSig.bytes.length > 68 && maybeSig.bytes.length < 73)
```
Christewart, line 134: Just double checking, shoudl these be inclusive or exclusive? i.e. 

```scala
(maybeSig.bytes.length >= 68 && maybeSig.bytes.length <= 73)
```

nkohen: exclusive, 68 is way less than is really probable and 72 is the max sig size

benthecarman: 72 bytes is the upper bound 

https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#der-encoding

I am not sure about the lower bound

benthecarman (approved):

Christewart (approved):

PR #1751, opened 2020-08-04
Add mempool.space as a fee provider

### LINKED ISSUE #1747: Create Mempool.space FeeProvider
Mempool.space recently released a new fee estimator, would be nice to use

https://mempool.space/about

https://mempool.space/api/v1/fees/recommended

benthecarman:
Closes #1747


Christewart:
We have no ability currently to configure fee proviers from the bitcoin-s.conf file right? For instance to pick which one you want. Perhaps we should add this as a feature for v0.5

benthecarman:
> We have no ability currently to configure fee proviers from the bitcoin-s.conf file right? For instance to pick which one you want. Perhaps we should add this as a feature for v0.5

Yeah I agree

PR #1767, opened 2020-08-06
Add funding and closing txids to DLCDb

benthecarman:
Added columns for the funding and closing txIds, these should be useful for tracking the state of the dlc as well as fetching relevant data.


benthecarman:
This is no longer needed

PR #1771, opened 2020-08-06
Add synchronous version of StartStop

Christewart:
Make the existing `StartStop` return `T` and make a `StartStopAsync` interface that returns `Future[T]`


core/src/main/scala/org/bitcoins/core/util/StartStopAsync.scala
```diff
@@ -0,0 +1,10 @@
+package org.bitcoins.core.util
+
+import scala.concurrent.Future
+
+/** Provide a uniform trait to start/stop a service asynchrously. For synchronous starts
+  * and stops please see [[StartStop]] */
+trait StartStopAsync[T] {
```
benthecarman: Couldn't this just be
```
trait StartStopAsync[T] extends StartStop[Future[T]]
```

benthecarman (approved):

PR #1775, opened 2020-08-06
Use safeRewind for schnorr and adaptor sig functions

benthecarman:
Same as #1774 but for the `dlc` branch


nkohen:
The target branch for this should probably be `dlc-crypto`

benthecarman:
> The target branch for this should probably be `dlc-crypto`

Fixed

PR #1777, opened 2020-08-06
Better return message for bitcoin-s-cli stop

### LINKED ISSUE #1768: Better return message for bitcoin-s-cli stop
This is what gets returned when it works:
```
$ ./bitcoin-s-cli stop
Connection refused (Connection refused)
```

benthecarman:
Closes #1768 

image

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
@@ -5,6 +5,7 @@ import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server._
 import org.bitcoins.node.Node
 
+import scala.concurrent.duration.DurationInt
 import scala.util.{Failure, Success}
 
 case class NodeRoutes(node: Node)(implicit system: ActorSystem)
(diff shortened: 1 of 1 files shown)


app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
```diff
@@ -22,7 +23,14 @@ case class NodeRoutes(node: Node)(implicit system: ActorSystem)
         val nodeStopping = node.stop().map { _ =>
           Server.httpSuccess("Node shutting down")
         }
-        system.terminate()
+
+        val shutdownRunnable = new Runnable {
```
Christewart, line 27: Why is this needed? Otherwise we terminate the system before propagating the http success?

Christewart (approved):

app/server/src/main/scala/org/bitcoins/server/NodeRoutes.scala
benthecarman: Yeah, we would queue up the message to send, but then shutdown. This way we give ourselves enough time to send the message

PR #1782, opened 2020-08-07
Fix for Wallet confirmed states

### LINKED ISSUE #1770: Wallet state incorrect if you pair an old wallet with a new chain db
So I accidentally deleted my chaindb on testnet, which meant that I had to do IBD to get all headers and filters. I synced the _new_ chaindb with an _existing_ wallet with the expected outcome of ending up with the exact same balance when I get done with IBD.

After completing IBD, here is what my wallet `txo_spending_info` table contains

>sqlite> select * from txo_spending_info;
1|2ada5ad20661496580dfde529d0f7228db03728479f2fc8ef2306a4ca89ab80001000000|1976a914e712987d28becb69e3fcf33f02e4e6d34c6e199088ac|1815403|m/44'/1'/0'/0/1|||00b89aa84c6a30f28efcf279847203db28720f9d52dedf8065496106d25ada2a|000000000021240700150008da274d6555659069899854823f83f55e94ca39b6|PendingConfirmationsReceived

The key thing here is the state of the utxo which is `PendingConfirmationsReceived`. It should be in the `Confirmed` state as it was in a testnet block that was mined quite sometime ago. [Here](https://blockstream.info/testnet/address/n2akfR5UdUan78gwaiGCzsBLhrvY4J57nY) is a link to a block explorer for the address.

### LINKED ISSUE #1780: getconfirmedbalance not working
This is the address on the testnet blockchain: https://blockstream.info/testnet/address/tb1qufh4hy0nse2mn46tsvzs0yp2thew9736fvat06

```
./bitcoin-s-cli getutxos
2ada5ad20661496580dfde529d0f7228db03728479f2fc8ef2306a4ca89ab80001000000 1815403 sats
0644f13e2623cd3020ca38ed97e10fbe54f5dbfa0022da90262b83bfe0b4f7c700000000 30000000 sats
b8a89c1c4bb9b744baf640199fa1bf25e1bedb9d6fc8d1102220801b805604d101000000 20000000 sats
c171dee042da941d568b1c516047d84a6e0aadb8d6d34fe2bf8f1b11ca2eee4d00000000 10000000 sats

➜  bin git:(master) ✗ ./bitcoin-s-cli getconfirmedbalance
0E-8 BTC
➜  bin git:(master) ✗ ./bitcoin-s-cli getbalance
0.60000000 BTC
```

I would expect after that many confirms that `getconfirmedbalance` and `getbalance would return the same thing.

benthecarman:
Fixes #1780 
Fixes #1770 

There were 2 bugs here:

1) When we received a block, we would not try to process the header causing the callback to not be executed sometimes, as well as then not adding the header to the db

2) When the wallet tried to update the confirmed state of a txo it would check on the number of confirmations of the lastest block header (always 1), not the one that the transaction is associated with.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -8,7 +8,6 @@ import org.bitcoins.crypto.DoubleSha256DigestBE
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.models.BroadcastAbleTransactionDAO
 import org.bitcoins.node.{NodeCallbacks, P2PLogger}
-import org.sqlite.SQLiteException
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -236,19 +235,18 @@ case class DataMessageHandler(
           s"Received block message with hash ${block.blockHeader.hash.flip}")
 
         val newApiF = {
-          // if we've already processed this header it will fail, but we should still try
-          try {
-            for {
-              processedApi <- chainApi.processHeader(block.blockHeader)
-              _ <- callbacks.executeOnBlockHeadersReceivedCallbacks(
-                logger,
-                Vector(block.blockHeader))
-            } yield processedApi
-
-          } catch {
-            case _: SQLiteException =>
-              Future.successful(chainApi)
-          }
+          chainApi
+            .getHeader(block.blockHeader.hashBE)
+            .flatMap { headerOpt =>
+              if (headerOpt.isEmpty) {
+                for {
+                  processedApi <- chainApi.processHeader(block.blockHeader)
+                  _ <- callbacks.executeOnBlockHeadersReceivedCallbacks(
+                    logger,
+                    Vector(block.blockHeader))
+                } yield processedApi
+              } else Future.successful(chainApi)
+            }
         }
 
         for {
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -8,6 +8,7 @@ import org.bitcoins.crypto.DoubleSha256DigestBE
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.models.BroadcastAbleTransactionDAO
 import org.bitcoins.node.{NodeCallbacks, P2PLogger}
+import org.sqlite.SQLiteException
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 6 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -230,11 +231,34 @@ case class DataMessageHandler(
...
+        val block = msg.block
         logger.info(
-          s"Received block message with hash ${msg.block.blockHeader.hash.flip}")
-        callbacks
-          .executeOnBlockReceivedCallbacks(logger, msg.block)
-          .map(_ => this)
+          s"Received block message with hash ${block.blockHeader.hash.flip}")
+
+        val newApiF = {
+          // if we've already processed this header it will fail, but we should still try
```
Christewart: Why don't we just check if we have it in our DB? I guess this doesn't _necessarily_ mean that `.processHeader()` completed sucessfully but it seems better than this no? 

The problem with the current implementation is postgres does not throw `SQLiteException` :-)

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -119,14 +118,20 @@ private[wallet] trait UtxoHandling extends WalletLogger {
                   // We should keep the utxo as reserved so it is not used in
                   // a future transaction that it should not be in
                   txo
-                case TxoState.ConfirmedReceived | TxoState.ConfirmedSpent =>
+                case TxoState.DoesNotExist | TxoState.ConfirmedReceived |
```
Christewart, line 121: Why did `TxoState.DoesNotExist` get moved here?

benthecarman: `DoesNotExist` is not currently implemented, but from my understanding it will be used if a something is double spent and another txo takes it place in the chain.

I think the block getting confirmations shouldn't change it's state from DNE, we should have proper handling of if the block has been reorged back in or something.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
benthecarman: That's a better idea

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
Christewart: The intention here is that it's useful for L2 protocols like Lightning, for instance you know what the funding outpoint should be, but it hasn't actually been broadcast on the network. Same thing can apply for HTLC success txs that spend a commitment transaction. 

That should probably be the scaladoc on `TxoState.DoesNotExist` as the current one kind of sucks.

benthecarman: Well the same logic applies here as well. If it doesn't exist yet we shouldn't be updating it to be confirmed or not

Christewart (approved):

PR #1789, opened 2020-08-08
Use database to calculate number of confirmations

### LINKED ISSUE #1780: getconfirmedbalance not working

benthecarman:
Fixes #1780 

Context: https://github.com/bitcoin-s/bitcoin-s/issues/1780#issuecomment-670966280

This should not be an performance hit to anything too meaningful as `getNumberOfConfirmations` is only used for a few wallet functions

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -443,18 +443,16 @@ case class ChainHandler(
       case Some(blockHeight) =>
         for {
           tips <- blockHeaderDAO.chainTips
-          chains <- blockHeaderDAO.getBlockchainsBetweenHeights(
-            blockHeight,
-            // max by height so we get the most chains since all tips will have same work
-            tips.maxBy(_.height).height)
-        } yield {
-          val confs = chains.flatMap { chain =>
-            val inChain = chain.exists(_.hashBE == blockHash) && chain.exists(
-              _.hashBE == chain.tip.hashBE)
-            if (inChain) Some(chain.tip.height - blockHeight + 1)
-            else None
+          getNAncestorsFs = tips.map { tip =>
+            blockHeaderDAO.getNAncestors(tip.hashBE, tip.height - blockHeight)
           }
-          confs.maxOption
+          ancestorChains <- Future.sequence(getNAncestorsFs)
+        } yield {
+          ancestorChains.flatMap { chain =>
+            if (chain.last.hashBE == blockHash) {
+              Some(chain.head.height - blockHeight + 1)
+            } else None
+          }.maxOption
         }
     }
   }
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -448,11 +448,15 @@ case class ChainHandler(
           }
           ancestorChains <- Future.sequence(getNAncestorsFs)
         } yield {
-          ancestorChains.flatMap { chain =>
+          val confs = ancestorChains.flatMap { chain =>
             if (chain.last.hashBE == blockHash) {
               Some(chain.head.height - blockHeight + 1)
             } else None
-          }.maxOption
+          }
+
+          if (confs.nonEmpty) {
+            Some(confs.max)
+          } else None
         }
     }
   }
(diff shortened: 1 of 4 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -442,11 +442,9 @@ case class ChainHandler(
...
       case None => FutureUtil.none
       case Some(blockHeight) =>
-        for {
-          tipHash <- getBestBlockHash()
-          tipHeightOpt <- getBlockHeight(tipHash)
-        } yield {
-          tipHeightOpt.map(tipHeight => tipHeight - blockHeight + 1)
+        blockHeaderDAO.chainTips.map { tips =>
+          val bestHeader = tips.maxBy(_.chainWork)
+          Some(bestHeader.height - blockHeight + 1)
```
Christewart: This doesn't seem right. We don't know if the block is apart of these various chains do we? The block we are talking about could be a part of Chain A, but not Chain B?

benthecarman: I'm not sure how we'd solve that without linking headers all the way back

Christewart (changes requested):
This doesn't seem to work for me: 

>➜  bin git:(7aef728b3a) ✗ ./bitcoin-s-cli getblockcount
1806673
➜  bin git:(7aef728b3a) ✗ ./bitcoin-s-cli getconfirmedbalance
0E-8 BTC
➜  bin git:(7aef728b3a) ✗ ./bitcoin-s-cli getblockcount
1806674
➜  bin git:(7aef728b3a) ✗ ./bitcoin-s-cli getconfirmedbalance
0E-8 BTC

Christewart:
Still nothing

>➜  bin git:(33e1166a9a) ✗ ./bitcoin-s-cli getblockcount
1806675
➜  bin git:(33e1166a9a) ✗ ./bitcoin-s-cli getconfirmedbalance
0E-8 BTC
➜  bin git:(33e1166a9a) ✗ ./bitcoin-s-cli getblockcount
1806681
➜  bin git:(33e1166a9a) ✗ ./bitcoin-s-cli getconfirmedbalance
0E-8 BTC

benthecarman:
I did a bunch of txs on this branch today while testing for the release and they are all marked as confirmed in my db. Unsure what the cause could be for @Christewart

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -442,10 +442,21 @@ case class ChainHandler(
...
         } yield {
-          tipHeightOpt.map(tipHeight => tipHeight - blockHeight + 1)
+          val confs = ancestorChains.flatMap { chain =>
+            if (chain.last.hashBE == blockHash) {
+              Some(chain.head.height - blockHeight + 1)
+            } else None
+          }
+
+          if (confs.nonEmpty) {
+            Some(confs.max)
```
Christewart, line 458: Is this right? Don't we want to use the "heaviest" confirmation rather than the highest number?

benthecarman: We are using `blockHeaderDAO.chainTips` which means all the tips will have equal work. More than likely everything in `confs` will be equal, but we have to pick one, so for some reason if a block is in a chain equal work chain but longer, then I think that's the better chain to pick.

Christewart (approved):

PR #1793, opened 2020-08-10
Send correct version message on node start up

benthecarman:
The `timestamp` field was incorrectly sending the current epoch in milliseconds instead of seconds. It also will now send a better user agent string, and the correct starting block height.

I am still not sure on how to send the synced headers and blocks message, it is not in the version message: https://developer.bitcoin.org/reference/p2p_networking.html#version

Before:
image

After:
image


Christewart:
Cool! What tool are you using to get this info? Is that in the bitcoin core gui?

benthecarman:
> Cool! What tool are you using to get this info? Is that in the bitcoin core gui?

Yeah, if you go to `Window` > `Peers` it gives info about them

Christewart (approved):

Christewart (approved):

PR #1801, opened 2020-08-11
Make Akka log to file correctly

### LINKED ISSUE #1738: akka logs are not added to bitcoin-s.log file
Akka logs are critical for server/node operations. We should append these logs to bitcoin-s.log so we have all relevant information for debugging.

benthecarman:
Fixes #1738 

Adds `akka-slf4j` as a dependency. My intelliJ ended up formatting the whole `Deps` file too

This fixes 2 things:

1) The akka config in a `bitcoin-s.conf` file was not being used

2) The akka logs are now correctly formatted and logged to file

It seems this also might have fixed something with the `BitcoinSLogger`

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/Main.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/Main.scala
@@ -1,6 +1,6 @@
 package org.bitcoins.server
 
-import java.nio.file.Paths
+import java.nio.file.{Path, Paths}
 
 import akka.actor.ActorSystem
 import akka.dispatch.Dispatchers
@@ -14,6 +14,7 @@ import org.bitcoins.chain.models.{
   CompactFilterHeaderDAO
 }
 import org.bitcoins.core.Core
+import org.bitcoins.core.config.{BitcoinNetworks, MainNet, RegTest, TestNet3}
 import org.bitcoins.core.util.{BitcoinSLogger, FutureUtil, NetworkUtil}
 import org.bitcoins.db.AppConfig
 import org.bitcoins.feeprovider.BitcoinerLiveFeeRateProvider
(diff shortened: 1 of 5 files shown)


nkohen (approved):
LGTM Assuming it works

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -28,21 +29,41 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+    val datadir: Path = {
+      val lastDirname = network match {
+        case MainNet  => "mainnet"
+        case TestNet3 => "testnet3"
+        case RegTest  => "regtest"
       }
+      datadirPath.resolve(lastDirname)
+    }
+
+    System.setProperty("bitcoins.log.location",
```
Christewart, line 58: So i'm guessing this isn't going to work if we use the `node` project as a library?

Christewart (approved):

app/server/src/main/scala/org/bitcoins/server/Main.scala
benthecarman: The bug was independent of the `node` project since it takes in its own `ActorSystem`. If someone else if using the `node` module in their project they'll have their own `logback.xml`  and `ActorSystem` to configure for however they want to do logging.

Christewart:
Works for me! Nice job @benthecarman ! 

>2020-08-12T12:50:34 INFO [Slf4jLogger] Slf4jLogger started
2020-08-12T12:50:34 INFO [ActorSystemImpl] Akka logger started
2020-08-12T12:50:34 DEBUG [BitcoinSLogger$] Resolved bitcoin-s config:

PR #1803, opened 2020-08-11
DLC Signature Tests

nkohen:
Added negative test cases for signature verification and a test case for computing an oracle signature in `DLCStatus.RemoteClaimed` (fixed bugs)

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
@@ -5,6 +5,7 @@ import org.bitcoins.commons.jsonmodels.dlc.DLCMessage.{
   DLCOffer,
   DLCSign
 }
+import org.bitcoins.core.policy.Policy
 import org.bitcoins.core.protocol.script.P2WSHWitnessV0
 import org.bitcoins.core.protocol.transaction.{Transaction, WitnessTransaction}
 import org.bitcoins.crypto.{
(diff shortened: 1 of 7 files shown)


app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -185,29 +186,33 @@ object DLCStatus {
           cetSig: ECDigitalSignature): SchnorrDigitalSignature = {
         val sigPubKey = oraclePubKey.computeSigPoint(msg.bytes, preCommittedR)
         val possibleOracleS =
-          sigPubKey.extractAdaptorSecret(adaptorSig, cetSig).fieldElement
+          sigPubKey
+            .extractAdaptorSecret(adaptorSig,
+                                  ECDigitalSignature(cetSig.bytes.dropRight(1)))
```
benthecarman: > `.dropRight(1)`

`.init`

benthecarman (approved):
Test looks good to me. Only way I'd improve them is to have the `numOutcomes` be a generated value and the bit for `flipAtIndex` be randomly generated as well.

However, not too important

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
```diff
@@ -337,4 +336,187 @@ class DLCClientTest extends BitcoinSAsyncTest {
...
+    val (toFlip, back) = backWithToFlip.splitAt(1)
+    front ++ toFlip.xor(ByteVector.fromByte(1)) ++ back
+  }
+
+  private def flipBit(fundingSigs: FundingSignatures): FundingSignatures = {
+    val (firstOutPoint, sigs) = fundingSigs.head
+    val badSigBytes = flipAtIndex(sigs.head.signature.bytes, 60)
+    val badSig = ECDigitalSignature(badSigBytes)
+    val badSigs =
+      sigs.tail.prepended(PartialSignature(sigs.head.pubKey, badSig))
```
benthecarman: `prepended` is not available for 2.12

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
```diff
@@ -337,4 +336,187 @@ class DLCClientTest extends BitcoinSAsyncTest {
 
     Future.sequence(testFs).flatMap(_ => executeRefundCase(numOutcomes))
   }
+
+  private def flipAtIndex(bytes: ByteVector, byteIndex: Int): ByteVector = {
```
benthecarman: Can we move these functions to somewhere in the `testkit`, that way we can use them in other places. These would be useful for testing the wallet is correctly handling invalid sigs as well

benthecarman (approved):

PR #1810, opened 2020-08-12
Revert scalaTestPlus to 3.2.0.0

benthecarman:
To configure the logging in the test cases you should use the `level` config.

2 things here:

1) scalaTestPlus 3.2.1.0 was causing the test cases being ran not to output, reverting the update fixed this.

2) Modules that rely on an appconfig that did not have a reference.conf file were using some fallback config that we could not easily configure. This adds a reference.conf to each testing module so it can easily configured.

### DIFF
--- a/project/Deps.scala
+++ b/project/Deps.scala
@@ -9,7 +9,7 @@ object Deps {
     val scalaTest = "3.2.0"
 
     val scalaTestPlus =
-      "3.2.1.0" //super annoying... https://oss.sonatype.org/content/groups/public/org/scalatestplus/
+      "3.2.0.0" //super annoying... https://oss.sonatype.org/content/groups/public/org/scalatestplus/
     val slf4j = "1.7.30"
     val spray = "1.3.5"
     val zeromq = "0.5.2"


Christewart:
>     2. Modules that rely on an appconfig that did not have a reference.conf file were using some fallback config that we could not easily configure. This adds a reference.conf to each testing module so it can easily configured.

I don't think this is right. The testkit's `reference.conf` is on the classpath. Since all of the `test` modules depend on `testkit` it should be on the classpath. I actually think this is a bit more difficult to configure logging as we now have 4 different reference.conf files rather than 1. 

>The idea is that libraries and frameworks should ship with a reference.conf in their jar. Applications should provide an application.conf, or if they want to create multiple configurations in a single JVM, they could use ConfigFactory.load("myapp") to load their own myapp.conf.

Christewart:
> > ```
> > 2. Modules that rely on an appconfig that did not have a reference.conf file were using some fallback config that we could not easily configure. This adds a reference.conf to each testing module so it can easily configured.
> > ```
> 
> I don't think this is right. The testkit's `reference.conf` is on the classpath. Since all of the `test` modules depend on `testkit` it should be on the classpath. I actually think this is a bit more difficult to configure logging as we now have 4 different reference.conf files rather than 1.
> 
> https://github.com/lightbend/config#standard-behavior
> 
> > The idea is that libraries and frameworks should ship with a reference.conf in their jar. Applications should provide an application.conf, or if they want to create multiple configurations in a single JVM, they could use ConfigFactory.load("myapp") to load their own myapp.conf.

I think #1814 fixes 2 now. We should downgrade the scalatestplus version though

Christewart (approved):

Issue #1809, opened 2020-08-12
Log file getting deleted and not cached


Christewart:
I was running 1ec10dd2cc and reporting to @benthecarman if my utxo states were updated and sent him my log file with nearly 2,000 logs in it. I came back and checked the log file later and there is now 2 lines of logs in it. This is probably related to #1805 

>2020-08-12T20:14:02 DEBUG [chain-verification] No prev found for BlockHeaderDb(1670872,DoubleSha256BDigestBE(000000000021240700150008da274d6555659069899854823f83f55e94ca39b6),Int32Impl(536870912),DoubleSha256BDigestBE(0000000000324f82435e02560aab6907c52ba059778afdac815f42d991eee9c2),DoubleSha256BDigestBE(478a1407f8333d77c148d186240e2c92b788420a96b0e0731c4076e4cef76b9f),UInt32Impl(1584970932),UInt32Impl(486604799),UInt32Impl(480148405),00000020c2e9ee91d9425f81acfd8a7759a02bc50769ab0a56025e43824f3200000000009f6bf7cee476401c73e0b0960a4288b7922c0e2486d148c1773d33f807148a47b4bc785effff001db57b9e1c,6060848556455348433979) hashBE=DoubleSha256BDigestBE(000000000021240700150008da274d6555659069899854823f83f55e94ca39b6)
2020-08-12T20:14:02 INFO [WALLET] 4 txos are now confirmed!

I've attached the original log file that I had before 
[bitcoin-s.log](https://github.com/bitcoin-s/bitcoin-s/files/5065569/bitcoin-s.log)

benthecarman:
The logs taking a bit could be due to the fact our logging is very inefficient (#1717) + the addition of the async logger for akka logs.

PR #1816, opened 2020-08-13
Remove default false for useLogbackConf

### LINKED ISSUE #1813: bitcoin-s.logging.logback does not work
I'm just trying to toggle settings inside of the [`testkit/reference.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/resources/reference.conf#L6-L7) while using it with our test cases. 

For instance, if I set `bitcoin-s.logging.logback = false` and the configure `bitcoin-s.logging.level=OFF` I am still getting logs when running `coreTest/testOnly *TransactionTest*`

You can see an example of this here: https://github.com/Christewart/bitcoin-s-core/commit/8a803f1221a3e89bcf253ccc0ead1a2da00ba9b8 if you want to reproduce locally

benthecarman:
Fixes #1813 

In #1398 I had set it all of the `AppConfig`s to override the `useLogbackConf` option to `false`. The reasoning was so we can use our own settings, however, I'm not sure why I did that, the `logback` option should just cover that for us. This removes the defaulting to false and just has it read from the config file.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
@@ -22,9 +22,9 @@ import org.bitcoins.db.AppConfig
 case class BitcoinSAppConfig(
     private val directory: Path,
     private val confs: Config*)(implicit ec: ExecutionContext) {
-  val walletConf: WalletAppConfig = WalletAppConfig(directory, false, confs: _*)
-  val nodeConf: NodeAppConfig = NodeAppConfig(directory, false, confs: _*)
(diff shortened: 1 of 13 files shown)


Christewart:
This seems to fix #1809

benthecarman:
I did a bunch of txs on this branch today while testing for the release and they are all marked as confirmed in my db. Unsure what the cause could be for @Christewart

Christewart:
Logging is going to need to be turned down to get this PR passing CI

https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/717686469#L1224

Christewart (approved):

PR #1819, opened 2020-08-13
Fix server parsing for send to address

benthecarman:
Our cli command had the `noBroadcast` option but the server never parsed it. This fixes that along with its subsequent tests.


Christewart:
What happens when `noBroadcast` is used?

benthecarman:
> What happens when `noBroadcast` is used?

It returns the raw tx bytes instead of broadcasting and returning the txId

Christewart:
> > What happens when `noBroadcast` is used?
> 
> It returns the raw tx bytes instead of broadcasting and returning the txId

Is it stored anywhere? What is the use case?

benthecarman:
The wallet will still process it so it will be stored in the tx_table. Usecase is if you want to broadcast it with some other means or if you want to do something with it beforehand. We originally added this with the `dlc` branch so we could get the full tx so we could use `decoderawtransaction` on it without having to lookup the tx on our node first.

Christewart (approved):

PR #1823, opened 2020-08-13
Remove unnecessary parallelization in Wallet

benthecarman:
As @rorp discovered, we have lots of future that are parallelized in the wallet unnecessarily and can cause database errors due to SQLite only allowing a single database connection.

I went through the wallet and tried to fix every one I found.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -138,11 +138,9 @@ abstract class Wallet
   override def processCompactFilters(
       blockFilters: Vector[(DoubleSha256Digest, GolombFilter)]): Future[
     Wallet] = {
-    val utxosF = listUtxos()
-    val addressesF = listAddresses()
     for {
-      utxos <- utxosF
-      addresses <- addressesF
+      utxos <- listUtxos()
+      addresses <- listAddresses()
       scriptPubKeys =
         utxos.flatMap(_.redeemScriptOpt).toSet ++ addresses
           .map(_.scriptPubKey)
@@ -222,16 +220,15 @@ abstract class Wallet
       yield {
         val filtered = utxos
           .filter(predicate)
-          .map {
-            case txo: SpendingInfoDb =>
-              txo.state match {
-                case TxoState.PendingConfirmationsReceived |
-                    TxoState.ConfirmedReceived =>
-                  txo.output.value
-                case TxoState.Reserved | TxoState.PendingConfirmationsSpent |
-                    TxoState.ConfirmedSpent | TxoState.DoesNotExist =>
-                  CurrencyUnits.zero
-              }
+          .map { txo =>
+            txo.state match {
+              case TxoState.PendingConfirmationsReceived |
+                  TxoState.ConfirmedReceived =>
+                txo.output.value
+              case TxoState.Reserved | TxoState.PendingConfirmationsSpent |
+                  TxoState.ConfirmedSpent | TxoState.DoesNotExist =>
+                CurrencyUnits.zero
+            }
           }
 
         filtered.fold(0.sats)(_ + _)
@@ -239,24 +236,22 @@ abstract class Wallet
   }
 
   override def getConfirmedBalance(): Future[CurrencyUnit] = {
-    val confirmed = filterThenSum(_.state == ConfirmedReceived)
-    confirmed.foreach(balance =>
-      logger.trace(s"Confirmed balance=${balance.satoshis}"))
-    confirmed
+    filterThenSum(_.state == ConfirmedReceived).map { balance =>
+      logger.trace(s"Confirmed balance=${balance.satoshis}")
+      balance
+    }
   }
 
   override def getConfirmedBalance(account: HDAccount): Future[CurrencyUnit] = {
-    val allUnspentF = spendingInfoDAO.findAllUnspent()
-    val unspentInAccountF = for {
-      allUnspent <- allUnspentF
+    for {
+      allUnspent <- spendingInfoDAO.findAllUnspent()
     } yield {
-      allUnspent.filter { utxo =>
+      val confirmedUtxos = allUnspent.filter { utxo =>
         HDAccount.isSameAccount(utxo.privKeyPath.path, account) &&
         utxo.state == ConfirmedReceived
       }
+      confirmedUtxos.foldLeft(CurrencyUnits.zero)(_ + _.output.value)
     }
-
-    unspentInAccountF.map(_.foldLeft(CurrencyUnits.zero)(_ + _.output.value))
   }
 
   override def getConfirmedBalance(tag: AddressTag): Future[CurrencyUnit] = {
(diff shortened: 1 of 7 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -135,11 +135,9 @@ abstract class Wallet
   override def processCompactFilters(
       blockFilters: Vector[(DoubleSha256Digest, GolombFilter)]): Future[
     Wallet] = {
-    val utxosF = listUtxos()
-    val addressesF = listAddresses()
     for {
-      utxos <- utxosF
-      addresses <- addressesF
+      utxos <- listUtxos()
```
Christewart, line 142: This is a performance regression for #1282 / #1283

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -531,14 +523,15 @@ abstract class Wallet
...
   override def createNewAccount(kmParams: KeyManagerParams): Future[Wallet] = {
-    val lastAccountOptF = accountDAO
-      .findAll()
-      .map(_.filter(_.hdAccount.purpose == kmParams.purpose))
-      .map(_.sortBy(_.hdAccount.index))
-      // we want to the most recently created account,
-      // to know what the index of our new account
-      // should be.
-      .map(_.lastOption)
+    def lastAccountOptF =
```
Christewart: make this another private method at the top level of the class rather than embedding inside of `createNewAccount`. Maybe it's useful elsewhere in the future.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -569,10 +562,10 @@ abstract class Wallet
       }
     }
     val newAccountDb = AccountDb(xpub, hdAccount)
-    val accountCreationF = accountDAO.create(newAccountDb)
-    accountCreationF.map(created =>
-      logger.debug(s"Created new account ${created.hdAccount}"))
-    accountCreationF.map(_ => this)
+    accountDAO.create(newAccountDb).map { created =>
```
Christewart, line 576: Things like this don't really help in my opinion, I don't think our solution should be "let's NOT use intermediate `val`s to avoid this problem!" This makes code unreadable.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -613,16 +606,14 @@ object Wallet extends WalletLogger {
...
 
     //see if we already have this account in our database
     //Three possible cases:
     //1. We have nothing in our database, so we need to insert it
     //2. We already have this account in our database, so we do nothing
     //3. We have this account in our database, with a DIFFERENT xpub. This is bad. Fail with an exception
     //   this most likely means that we have a different key manager than we expected
-    val accountOptF = accountDAO.read(account.coin, account.index)
-    accountOptF.flatMap {
+    wallet.accountDAO.read(account.coin, account.index).flatMap {
```
Christewart, line 627: again, I don't think the solution is to NOT use intermediate `val`

Christewart (changes requested):
I'm mostly a NACK on this PR. The majority of changes aren't changing any existing behavior and are making things harder to read. We care about _parallel_ behavior, this is pattern is not an example of parallel behavior, at least as I understand @rorp 

```scala

val future1: Future[Int] = foo()
future1.map { int => //do something } 
```

is equivalent to 

```scala
foo().map { int => //do something }
```

expect the latter pattern becomes _extremely_ hard to read and code review as you scale. i.e.

```scala
foo().map(int => //do something).map(_ => //do something else).map(_=> //and again)
```

rather than 

```scala
val future1 = foo()
val future2 = future1.map(_ => //do something) 
val future3 = future2.map(_ => //do something else)
val future4 = future3.map(_ => //and again)
```

@rorp thoughts?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
Christewart: Also according to sqlite you can safely read concurrently from a sqlite database, however writing is where the problem comes into play if I understand @rorp correctly

https://stackoverflow.com/a/4060838/967713

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
rorp: It is

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
rorp: We don't need intermediate vals in most of the cases, because they create the unnecessary parallelization

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
rorp: 👍

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
rorp: That fixes potential SQLITE_BUSY on production size databases. To fix the bottleneck (if it actually exists) the join should be done on the database level.

benthecarman: Since we now process filters in batches, it shouldn't be as much a concern

rorp:
> The majority of changes aren't changing any existing behavior and are making things harder to read. 

Some of the changes are not changing the logic for sure, but they do improve readability.
Classic example: https://github.com/bitcoin-s/bitcoin-s/pull/1823/files#diff-fe3b2d50794290699b4ac0929ec80201R243 one line instead of two makes the difference.

> ```scala
> val future1 = foo()
> val future2 = future1.map(_ => //do something) 
> val future3 = future2.map(_ => //do something else)
> val future4 = future3.map(_ => //and again)
> ```

I find it unreadable because there's to much information noise here. Also in our code base those `do something`s tend to be a couple of screens long, and hurt readability even more.

I find this much easier to read.
```scala
for().map {
   doSomething()
   doSomethingElse()
   andAgain()
}
```
And again overuse of Future instantiations leads to undesirable parallelization.

rorp (approved):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
Christewart: >That fixes potential SQLITE_BUSY on production size databases. To fix the bottleneck (if it actually exists) the join should be done on the database level.

SQLITE_BUSY only occurs on writes, there is no point in reducing performance for reads: https://sqlite.org/rescode.html#busy

wallet/src/main/scala/org/bitcoins/wallet/internal/RescanHandling.scala
```diff
@@ -210,17 +210,15 @@ private[wallet] trait RescanHandling extends WalletLogger {
...
       Executors.newFixedThreadPool(Runtime.getRuntime.availableProcessors() * 2)
 
-    val blocksF = for {
+    for {
       blocks <- getMatchingBlocks(
         scripts = scriptPubKeys,
         startOpt = startOpt,
         endOpt = endOpt)(ExecutionContext.fromExecutor(threadPool))
     } yield {
+      threadPool.shutdown()
```
Christewart, line 221: This does not shutdown the thread pool on failure cases.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -166,75 +166,57 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
-                processNewIncomingTx(transaction, blockHashOpt, newTags)
-                  .map(_.toVector)
-
-              case txos: Vector[SpendingInfoDb] =>
-                FutureUtil
-                  .sequentially(txos)(txo =>
-                    processExistingIncomingTxo(transaction, blockHashOpt, txo))
-                  .map(_.toVector)
-            }
+    def incomingTxoFut: Future[Vector[SpendingInfoDb]] =
```
Christewart: Let's just move these helper methods out of `processTransactionImpl()` now since they aren't `val` anymore.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -166,75 +166,57 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
 
-        aggregateFut
-      }
+    for {
+      incoming <- incomingTxoFut
+      outgoing <- outgoingTxFut
+      _ <- walletCallbacks.executeOnTransactionProcessed(logger, transaction)
     } yield {
-      aggregate
+      ProcessTxResult(incoming.toList, outgoing.toList)
```
Christewart: Since we are touching all of this, let's modify `ProcessTxResult` to be `Vector[]` rather than `List[]`

Christewart (changes requested):

PR #1824, opened 2020-08-14
Fix address tag issue where we weren't adding tags to an address, we …

### LINKED ISSUE #1815: AddressHandlingTest failure in drop address tag test case
https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/717648179#L4524

Screenshot from 2020-08-13 12-39-19

Christewart:
fixes #1815 

Previously we would use `upsert()` for `AddressTagDAO.tagAddress()`. This effectively means we we cannot have more than one tag per address (we will overwrite existing tags with `upsert()`). This PR uses `create()` rather than `upsert()` now to get the unit test passing.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
@@ -410,7 +410,8 @@ private[wallet] trait AddressHandling extends WalletLogger {
       address: BitcoinAddress,
       tag: AddressTag): Future[AddressTagDb] = {
     val addressTagDb = AddressTagDb(address, tag)
-    addressTagDAO.upsert(addressTagDb)
+    val f = addressTagDAO.create(addressTagDb)
+    f
   }
 
   def getAddressTags(address: BitcoinAddress): Future[Vector[AddressTagDb]] = {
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -410,7 +410,8 @@ private[wallet] trait AddressHandling extends WalletLogger {
       address: BitcoinAddress,
       tag: AddressTag): Future[AddressTagDb] = {
     val addressTagDb = AddressTagDb(address, tag)
-    addressTagDAO.upsert(addressTagDb)
+    val f = addressTagDAO.create(addressTagDb)
+    f
```
benthecarman, line 414: From my understanding, this is what we are trying to get away from akin to #1823, with this `val` we may start the future before we want to

Christewart: Let's leave this aside for now and continue this on #1823 , is there a reason we were using `upsert()` rather than `create`?

benthecarman: The reason I did that originally was in case someone tried to tag the address with the same tag twice, then they wouldn't get duplicates

benthecarman: I think this should be fine though

Christewart: Ok, but then why not enforce with this with a primary key at the database level of `(BitcoinAddress,Tag)`? This unit test does not pass otherwise because with the use of `upsert()` an address can have at most one tag

benthecarman: Yeah, that's probably best. I forget why I didn't do that originally but that makes the most sense to me

Issue #1829, opened 2020-08-14
Optimize `BaseBlockchain.connectWalkBackwards` out of O(n^2) behavior


Christewart:
https://github.com/bitcoin-s/bitcoin-s/issues/1780#issuecomment-674195866

One of the problems that is causing #1780 is [`BaseBlockchain.connectWalkBackwards()`](https://github.com/bitcoin-s/bitcoin-s/blob/master/chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala#L248) is a bottleneck.

Specifically this line: 

https://github.com/bitcoin-s/bitcoin-s/blob/86f68b3b622a4531748169cc475a5df52ac4c23e/chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala#L248

Some ideas of optimizing: 

1. Remove header from `ancestors` when it's in the chain (therefore are traversal gets shorter as we walk backwards)
2. Maybe avoid using `.find()` for just a raw while loop? 
3. Parallel search on the collection?

benthecarman:
Can probably do something where we do `groupBy(_.height)` and then only search for the `previousBlockHashBE` at the expected height.

PR #1832, opened 2020-08-14
Fix get missing utxos

benthecarman:
On `wallet.start()` we would call `downloadMissingUtxos` to get the txs associated with out outputs. However, when we'd process these txs we would see we already have the utxo and not add the tx to our database becuase we only add it in `processNewIncomingTx`. This fixes that by changing `processExistingIncomingTxo` to try and add the tx to the database if we are processing it's block a second time.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -92,7 +92,7 @@ abstract class Wallet
 
   private def utxosWithMissingTx: Future[Vector[SpendingInfoDb]] = {
     for {
-      utxos <- spendingInfoDAO.findAllUnspent()
+      utxos <- spendingInfoDAO.findAll()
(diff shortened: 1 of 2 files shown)


Christewart:
This in conjunction with #1829 fixes #1780 ! 

>➜  bin git:(ef8cb35eeb) ✗ ./bitcoin-s-cli getbalance
0.60000000 BTC
➜  bin git:(ef8cb35eeb) ✗ ./bitcoin-s-cli getconfirmedbalance
0.60000000 BTC

:tada: :tada: :tada:

Christewart (commented):
So things look good to me, but is there any test cases we can add to prevent regressions in the future?

benthecarman:
> So things look good to me, but is there any test cases we can add to prevent regressions in the future?

I don't think so, I had to edit my db schema to be able to reproduce

Christewart (approved):

PR #1833, opened 2020-08-14
2020 08 14 issue 1829

### LINKED ISSUE #1829: Optimize `BaseBlockchain.connectWalkBackwards` out of O(n^2) behavior

### LINKED ISSUE #1830: Cache BlockHeader.hash rather than re-computing everytime we access it
This is another bottleneck coming out of #1829 , when using [`BaseBlockchain.connectWalkBackwards`](https://github.com/bitcoin-s/bitcoin-s/blob/86f68b3b622a4531748169cc475a5df52ac4c23e/chain/src/main/scala/org/bitcoins/chain/blockchain/BaseBlockChain.scala#L248) we re-compute the block's hash everytime we look at a block header, which means we have to serialize it over and over again.

We should just cache the hash

Screenshot from 2020-08-14 13-58-58

Christewart:
fixes #1829 
fixes #1830 

Takes @benthecarman suggestion to optimize `BaseBlockchain.connectWalkbackwards()` for large setse of headers by creating a `Map` indexed by header height. 

This PR also caches the result of `BlockHeader.hash` after the first call so we don't need to re-serialize a header everytime we iterate over it looking for a hash

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/blockchain/BlockHeader.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/blockchain/BlockHeader.scala
@@ -116,7 +116,7 @@ sealed trait BlockHeader extends NetworkElement {
   def nonce: UInt32
 
   /** Returns the block's hash in the protocol level little endian encoding */
-  def hash: DoubleSha256Digest = CryptoUtil.doubleSHA256(bytes)
+  lazy val hash: DoubleSha256Digest = CryptoUtil.doubleSHA256(bytes)
 
   /**
     * Returns the block hash in big endian format, this is useful for rpc
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -597,14 +597,20 @@ class ChainHandlerTest extends ChainDbUnitTest {
             getHeaderF.map(headerOpt =>
               assert(headerOpt.contains(expectedBlockHeaderDb)))
           val newAccum = accum.:+(assertionF)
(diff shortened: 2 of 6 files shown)


core/src/main/scala/org/bitcoins/core/protocol/blockchain/BlockHeader.scala
```diff
@@ -116,7 +116,7 @@ sealed trait BlockHeader extends NetworkElement {
   def nonce: UInt32
 
   /** Returns the block's hash in the protocol level little endian encoding */
-  def hash: DoubleSha256Digest = CryptoUtil.doubleSHA256(bytes)
+  lazy val hash: DoubleSha256Digest = CryptoUtil.doubleSHA256(bytes)
 
   /**
     * Returns the block hash in big endian format, this is useful for rpc
```
benthecarman, line 122: maybe make `bytes` and `hashBE` lazy val as well

benthecarman (approved):

Issue #1825, opened 2020-08-14
BitcoinPowTest reject execution exception on CI


Christewart:
First seen on #1816 although I believe the PR is completely unrelated 

https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/717749023#L613

![Screenshot from 2020-08-14 08-45-45](https://user-images.githubusercontent.com/3514957/90255906-8e7fc980-de0a-11ea-9eef-f6189838dfab.png)

Christewart:
Now that I think about this the root cause might be #1281 as we reduced the number of reads that can be going on at a given time, we cap the number of threads to 5 in this PR.

PR #1798, opened 2020-08-11
Refactor wallet DB

rorp:

Issue #1769, opened 2020-08-06
Database file locked, using bitcoin-s-cli while syncing filters


Christewart:
**Describe the bug**
While doing IBD I wanted to check out the state of my database. Here is what I executed in terms of cli commands

>➜  bin git:(master) ✗ ./bitcoin-s-cli getblockcount
1805050
➜  bin git:(master) ✗ ./bitcoin-s-cli getfilterheadercount
1805047
➜  bin git:(master) ✗ ./bitcoin-s-cli getfiltercount
626599

You can see the `getfiltercount` request being received at the very top of the logs I pasted. This seems to cause the database to become locked and throw an exception. This stops the node from operating causing IBD to fail. You can see that I actually got a valid return for `getfiltercount`, however there may have been a write attempt to the database simultaneously that caused the node to fail.

>2020-08-06T15:56:47 DEBUG [chain-verification] getFilterCount result: count=626599
[INFO] [08/06/2020 10:56:47.440] [bitcoin-s-akka.actor.default-dispatcher-14] [akka.actor.ActorSystemImpl(bitcoin-s)] http-rpc-server: Response for
  Request : HttpRequest(HttpMethod(POST),http://localhost:9999/,List(Timeout-Access: <function1>, Accept-Encoding: gzip, deflate, User-Agent: Java/1.8.0_242, Host: localhost:9999, Accept: text/html, image/gif, image/jpeg, */*;q=0.2, */*;q=0.2, Connection: keep-alive),HttpEntity.Strict(application/json,83 bytes total),HttpProtocol(HTTP/1.1))
  Response: Complete(HttpResponse(200 OK,List(),HttpEntity.Strict(application/json,30 bytes total),HttpProtocol(HTTP/1.1)))
[ERROR] [08/06/2020 10:56:47.472] [bitcoin-s-akka.actor.internal-dispatcher-25] [akka://bitcoin-s/user/P2PClient$-1596728123902] [SQLITE_BUSY]  The database file is locked (database is locked)
org.sqlite.SQLiteException: [SQLITE_BUSY]  The database file is locked (database is locked)
        at org.sqlite.core.DB.newSQLException(DB.java:1010)
        at org.sqlite.core.DB.newSQLException(DB.java:1022)
        at org.sqlite.core.DB.throwex(DB.java:987)
        at org.sqlite.core.DB.exec(DB.java:185)
        at org.sqlite.SQLiteConnection.commit(SQLiteConnection.java:404)
[...]

Issue #1724, opened 2020-07-30
Database queue fills and throws an exception when syncing filters checkpoints


Christewart:
After running chainwork recalculation  as documetned here, I got a database error when starting to sync the node after running chain work recalc: https://github.com/bitcoin-s/bitcoin-s/pull/1697#issuecomment-666581022

The key thing from the error message below is

> Task slick.basic.BasicBackend$DatabaseDef$$anon$3@7334dd54 rejected from slick.util.AsyncExecutor$$anon$1$$anon$2@45c456b3[Running, pool size = 5, active threads = 5, queued tasks = 1000, completed tasks = 2760]

I believe slick has a default of `1000` tasks to be queue for running on the database before it throws and exception, which happens here. Why do we end up 1,000 queued tasks and we can't process them all on startup? Are we not batching upserting of compact filters?

>2020-07-30T18:14:40 DEBUG [P2P] Parsed 1 message(s) from bytes: NetworkMessage(NetworkHeader(TestNet3,cfcheckpt,UInt32Impl(57028),ByteVector(4 bytes, 0x40555c33)), CompactFilterCheckPointMessage(filterType=Basic, stopHash=DoubleSha256Digest(399b19db0deb1dfdda6c6a337a1b933be7ce2c3a5e70c02132a0267e00000000), filterHeaders=DoubleSha256Digest(22353b4ea01c7b93e55558f3e47489f6a518e8d570e0867bac3f56d013ccfd20)...DoubleSha256Digest(0bad5ec75288b82e8a441f2abcc6cd7ff41735684b073411f0971f0ca90f5095)))
2020-07-30T18:14:40 DEBUG [P2P] Unaligned bytes after this: 8132
2020-07-30T18:14:40 DEBUG [P2P] Received message=cfcheckpt from peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) state=Normal
[...]

PR #1840, opened 2020-08-16
Make database thread pool of size 1 to remove concerns of concurrent …

Christewart:
…access to the database

This uses the [`numThreads`](https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database) inside of scala slicks configuration to gurantee synchronous access to datatabase. This is -- in my opinion -- a superior alternative to #1823 as it actually fixes the `SQLITE_BUSY` problem by making it impossible to happen.

This allows for us to write Scala code and not worry about concurrent access to the sqlite database anymore, it also _preserves_ concurrent access to databases like postgres that support multiple connections to the database. 

You can see the single threads for `wallet.db-1` and `chain.db-1` 

Screenshot from 2020-08-16 09-56-51

In the future inside of `db.conf` we should really have different configurations for sqlite and postgres so that we can preserve safe concurrent accesses for the database in postgres. 

More generally, this is the only way to solve the `SQLITE_BUSY` problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO. 

```scala
//3rd party library 
val getbalance1 = wallet.getBalance()
val getbalance2 = wallet.getBalance()
```

With this PR, since the thread pool is of size 1, this two calls will have to be synchronized by the single thread that can access the database. 

This does not solve the problem hinted at by #1825 

>java.util.concurrent.RejectedExecutionException: Task slick.basic.BasicBackend$DatabaseDef$$anon$3@118ce568 rejected from slick.util.AsyncExecutor$$anon$1$$anon$2@31e3b8fb[Running, pool size = 5, active threads = 5, queued tasks = 1000, completed tasks = 2435]

With this PR, we can still overwhelm the queue. There is no way to solve as far as I can tell, you need to upgrade the performance of your database or switch to something like postgres that allows for concurrent accesses. This error can _also_ occur on postgres since at the end of the day you always hammer the database with arbitrary sql queries. You can adjust the [`queueSize`](https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database) configuration option to have a bigger queue if this is needed for your application

>queueSize (Int, optional, default: 1000): The size of the queue for database actions which cannot be executed immediately when all threads are busy. Beyond this limit new actions fail immediately. Set to 0 for no queue (direct hand-off) or to -1 for an unlimited queue size (not recommended).

### DIFF
--- a/db-commons/src/main/resources/db.conf
+++ b/db-commons/src/main/resources/db.conf
@@ -10,7 +10,7 @@ common = {
     username = ""
     password = ""
 
-    numThreads = 5 # default num threads is 20, which is way too much
+    numThreads = 1 # default num threads is 20, which is way too much
     #maxConnections = 1
     # as long as we're on SQLite there's no point
     # in doing connection pooling
--- a/db-commons/src/main/resources/db.conf
+++ b/db-commons/src/main/resources/db.conf
@@ -10,7 +10,9 @@ common = {
     username = ""
     password = ""
 
-    numThreads = 1 # default num threads is 20, which is way too much
+    # this needs to be set to 1 for SQLITE as it does not support concurrent database operations
+    # see: https://github.com/bitcoin-s/bitcoin-s/pull/1840
+    numThreads = 1
     #maxConnections = 1
     # as long as we're on SQLite there's no point
     # in doing connection pooling


rorp (commented):
It simply reintroduces https://github.com/bitcoin-s/bitcoin-s/issues/1724 in other places.
https://github.com/bitcoin-s/bitcoin-s/pull/1823 is the right fix. It should be applied in all sub-projects, not only for the wallet

Christewart:
> It simply reintroduces #1724 in other places.
> #1823 is the right fix. It should be applied in all sub-projects, not only for the wallet

#1823 is not a fix for `SQLITE_BUSY`

>More generally, this is the only way to solve the SQLITE_BUSY problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO.

With regards to #1724 / #1825 the PR #1823 does not solve it either. You cannot reason about what third party users of the API are going to call that my exhaust the queue. It's impossible to solve in general

See my original post for details of why, and how it can be mitigated

rorp:
> More generally, this is the only way to solve the `SQLITE_BUSY` problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO.
> 
> ```scala
> //3rd party library 
> val getbalance1 = wallet.getBalance()
> val getbalance2 = wallet.getBalance()
> ```

That means we should say in t documentation something like `if you experience SQLITE_BUSY exception, consider to refactor your code in this way, because such and such. And in general don't assign futures to local variables to avoid multi-threading side effects, unless you 100% need to`:

```scala
for {
  getbalance1 <- wallet.getBalance()
  getbalance2 <- wallet.getBalance()
} yield ...
```

Christewart:
> > More generally, this is the only way to solve the `SQLITE_BUSY` problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO.
> > ```scala
> > //3rd party library 
> > val getbalance1 = wallet.getBalance()
> > val getbalance2 = wallet.getBalance()
> > ```
> 
> That means we should say in t documentation something like `if you experience SQLITE_BUSY exception, consider to refactor your code in this way, because such and such. And in general don't assign futures to local variables to avoid multi-threading side effects, unless you 100% need to`:
> 
> ```scala
> for {
>   getbalance1 <- wallet.getBalance()
>   getbalance2 <- wallet.getBalance()
> } yield ...
> ```

This still doesn't really work. My example was trivial, however in practice things will be more complex. For instance, what if you have  webserver that allows people to query the balance of their wallet, it's hard to synchronize that in your application. We even support this, you could just call `bitcoin-s-cli getbalance` in twice in a row. The best way to solve this is at the database level with slick 

https://scala-slick.org/doc/3.3.1/database.html#database-thread-pool

rorp:
> > > More generally, this is the only way to solve the `SQLITE_BUSY` problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO.
> > > ```scala
> > > //3rd party library 
> > > val getbalance1 = wallet.getBalance()
> > > val getbalance2 = wallet.getBalance()
> > > ```
> > 
> > 
> > That means we should say in t documentation something like `if you experience SQLITE_BUSY exception, consider to refactor your code in this way, because such and such. And in general don't assign futures to local variables to avoid multi-threading side effects, unless you 100% need to`:
> > ```scala
> > for {
> >   getbalance1 <- wallet.getBalance()
> >   getbalance2 <- wallet.getBalance()
> > } yield ...
> > ```
> 
> This still doesn't really work. 

Judging by https://github.com/bitcoin-s/bitcoin-s/pull/1823 and https://github.com/bitcoin-s/bitcoin-s/pull/1798 it does work.

Christewart:
> > > > More generally, this is the only way to solve the `SQLITE_BUSY` problem. Since bitcoin-s is a library, there is no way to guarantee a caller of a method that accesses the database does not do it twice in a row, so the solution proposed in #1823 isn't really satisfactory IMO.
> > > > ```scala
> > > > //3rd party library 
> > > > val getbalance1 = wallet.getBalance()
> > > > val getbalance2 = wallet.getBalance()
> > > > ```
> > > 
> > > 
> > > That means we should say in t documentation something like `if you experience SQLITE_BUSY exception, consider to refactor your code in this way, because such and such. And in general don't assign futures to local variables to avoid multi-threading side effects, unless you 100% need to`:
> > > ```scala
> > > for {
> > >   getbalance1 <- wallet.getBalance()
> > >   getbalance2 <- wallet.getBalance()
> > > } yield ...
> > > ```
> > 
> > 
> > This still doesn't really work.
> 
> Judging by #1823 and #1798 it does work.

How so? The user can still issue arbitrary requests and a [thread pool of size 5](https://github.com/bitcoin-s/bitcoin-s/blob/26f47d269c9ecb40dd69c498c04b0b1f47c9c0d3/db-commons/src/main/resources/db.conf#L13) will still issue them to the database. Since we have a thread pool of size 5 still, this means we can still encounter a `SQLITE_BUSY` exception.

Christewart:
#1769 is an example of this being done in the wild, i tried to look at my filter count while IBD was going on

PR #1847, opened 2020-08-17
Implement basic TLV functionality

nkohen:
Introduced basic TLV types (Ping, Pong, Error, Unknown)

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/number/UInt16Test.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/number/UInt16Test.scala
@@ -1,0 +1,202 @@
+package org.bitcoins.core.number
+
+import org.bitcoins.testkit.core.gen.NumberGenerator
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.scalacheck.Gen
+import scodec.bits.ByteVector
+
+import scala.util.Try
+
+class UInt16Test extends BitcoinSUnitTest {
+
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
+
+  behavior of "UInt16"
+
+  it must "create the number zero as an unsigned 16 bit integer" in {
+    val zero = UInt16(ByteVector(0x0.toByte))
+    assert(zero.toInt == 0)
+  }
+
+  it must "create the max number for an unsigned byte" in {
+    val maxByteValue = UInt16(ByteVector(0xff.toByte))
+    assert(maxByteValue.toInt == 255)
+  }
+
+  it must "create the number 256" in {
+    val uInt16 = UInt16(ByteVector(0x01.toByte, 0x00.toByte))
+    assert(uInt16.toInt == 256)
+  }
+
+  it must "create the number 65535" in {
+    val uInt16 = UInt16(ByteVector(0xff.toByte, 0xff.toByte))
+    assert(uInt16.toInt == 65535)
+  }
+
+  it must "have the correct maximum number representation for UInt16" in {
+    assert(UInt16.max.toInt == 65535)
+    assert(UInt16.max.hex == "ffff")
+  }
+
+  it must "fail to create the number 65536" in {
+    assertThrows[IllegalArgumentException] {
+      UInt16(ByteVector(0x01.toByte, 0x0.toByte, 0x0.toByte))
+    }
+
+    assertThrows[IllegalArgumentException] {
+      UInt16(65536)
+    }
+  }
+
+  it must "throw an exception if we try and create a UInt16 with a negative number" in {
+    assertThrows[IllegalArgumentException] {
+      UInt16(-1)
+    }
+  }
+
+  it must "throw an exception if we try and create a UInt16 with more than 2 bytes" in {
+    assertThrows[IllegalArgumentException] {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -0,0 +1,140 @@
+package org.bitcoins.core.protocol.tlv
+
+import org.bitcoins.core.number.UInt16
+import org.bitcoins.core.protocol.BigSizeUInt
+import org.bitcoins.crypto.{Factory, NetworkElement}
+import scodec.bits.ByteVector
+
+sealed trait TLV extends NetworkElement {
+  def tpe: BigSizeUInt
```
benthecarman, line 10: type

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -0,0 +1,140 @@
...
+  override def fromTLVValue(value: ByteVector): PongTLV = {
+    val numIgnored = UInt16(value.take(2))
+    val ignored = value.drop(2).take(numIgnored.toLong)
+
+    PongTLV.forIgnored(ignored)
+  }
+
+  def forIgnored(ignored: ByteVector): PongTLV = {
+    new PongTLV(ignored)
+  }
```
benthecarman, line 147: Are there more types we should define?

core-test/src/test/scala/org/bitcoins/core/number/UInt16Test.scala
```diff
@@ -0,0 +1,202 @@
...
+        val result = num1.toInt + num2.toInt
+        if (result <= UInt16.max.toInt) {
+          assert(num1 + num2 == UInt16(result.toInt))
+        } else {
+          assert(Try(num1 + num2).isFailure)
+        }
+    }
+  }
+
+  it must "subtract zero correctly" in {
```
benthecarman, line 108: maybe add a test that it fails to subtracting from 0 
ie `UInt16.zero - UInt16(10)`

core-test/src/test/scala/org/bitcoins/core/protocol/tlv/TLVTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    forAll(TLVGen.pingTLV) { ping =>
+      assert(PingTLV(ping.bytes) == ping)
+    }
+  }
+
+  "PongTLV" must "have serialization symmetry" in {
+    forAll(TLVGen.pongTLV) { pong =>
+      assert(PongTLV(pong.bytes) == pong)
+    }
+  }
```
benthecarman, line 43: Should add tests that a `PongTLV` has an `Pong` type, etc

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -0,0 +1,140 @@
...
+
+    require(
+      bytes.length >= prefixSize + length.num.toLong,
+      s"Length specified was $length but not enough bytes in ${bytes.drop(prefixSize)}")
+
+    val value = bytes.drop(prefixSize).take(length.num.toLong)
+
+    (tpe, length, value)
+  }
+
```
benthecarman, line 42: It'd be nice to have a list of `knownTypes`

benthecarman (changes requested):

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
nkohen: reserved word, in scala the convention is to use `tpe` for this reason

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
nkohen: Do you mean for TLV? We can certainly add more later but these are more or less just to test things out and define conventions.

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: Ah yeah, forgot

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
rorp: &#96;type&#96;

rorp: or tlvType

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -0,0 +1,142 @@
...
+  override def bytes: ByteVector = {
+    tpe.bytes ++ length.bytes ++ value
+  }
+}
+
+object TLV extends Factory[TLV] {
+
+  def decodeTLV(bytes: ByteVector): (BigSizeUInt, BigSizeUInt, ByteVector) = {
```
Christewart: I would say we should just wrap this in a return type with a nice name and fields

I.e. `DecodeTLVResult` or something like that.

PR #1848, opened 2020-08-18
Create KeyManager Api & move SpendingInfoDb to core

benthecarman:
Part of #1284 

Originally tried just moving `SpendingInfoDb` to `core`. However I discovered we needed the `KeyManager`. So I created `KeyManagerApi` and `BIP39KeyManagerApi` along with it.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerApi.scala
@@ -1,0 +1,15 @@
+package org.bitcoins.core.api.keymanager
+
+import org.bitcoins.core.crypto.ExtPublicKey
+import org.bitcoins.core.hd.{HDAccount, HDPath}
+import org.bitcoins.crypto.Sign
+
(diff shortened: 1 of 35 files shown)


core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerApi.scala
```diff
@@ -0,0 +1,15 @@
+package org.bitcoins.core.api.keymanager
+
+import org.bitcoins.core.crypto.ExtPublicKey
+import org.bitcoins.core.hd.{HDAccount, HDPath}
+import org.bitcoins.crypto.Sign
+
+import scala.util.Try
+
+sealed trait KeyManagerApi
```
Christewart: I don't think we want to seal this, as other users may want to extend this to create their own API.

Christewart (changes requested):
Website failing CI: https://travis-ci.org/github/bitcoin-s/bitcoin-s/jobs/718818211

Christewart (approved):

PR #1849, opened 2020-08-18
Move TransactionDb to Core

benthecarman:
Part of #1284


core/src/main/scala/org/bitcoins/core/wallet/transactions/TransactionDb.scala
```diff
@@ -1,16 +1,9 @@
-package org.bitcoins.wallet.models
+package org.bitcoins.core.wallet.transactions
```
Christewart: same here, since it's part of the API we should probably have it in

`org.bitcoins.core.api.wallet` I think?

Christewart (approved):

PR #1850, opened 2020-08-18
Move AddressTagDb to Core

benthecarman:
Part of #1284


core/src/main/scala/org/bitcoins/core/wallet/utxo/AddressTagDb.scala
```diff
@@ -1,12 +1,6 @@
-package org.bitcoins.wallet.models
+package org.bitcoins.core.wallet.utxo
```
Christewart: Since these are part of the API shouldn't they be in 

`org.bitcoins.core.api.wallet` ?

Christewart (approved):

PR #1867, opened 2020-08-21
Only use appenders with no Logback conf

### LINKED ISSUE #1852: When used as a library bitcoin-s disables all logging for the host application
I'm using the latest master in https://github.com/rorp/eclair/tree/bitcoin-s-integration and noticed that once bitcoin-s gets initialized Eclair does not log a single line into it log file.

benthecarman:
Fixes #1852 

Previously our appenders would override the logback conf's appenders even if we wanted to use the logback conf instead of our internal ones. This changes it to only use them if we our config says to.

Fixed a typo as well.


Christewart:
Did you test with this with eclair?

benthecarman:
> Did you test with this with eclair?

Yes

Christewart (approved):

PR #1868, opened 2020-08-21
Start calling appConfig.stop() in destruction fixture code

Christewart:
Basically after running tests, if you are watching visualvm, you will see there is a _ton_ of slick thread pools still running. I'm sure after running this for long periods of time this will cause our OS to stall or at least have sbt crash

This is just from looking at `nodeTest/test`

Screenshot from 2020-08-21 10-27-12

and after 

Screenshot from 2020-08-21 11-21-23

You can see i still don't have them all cleaned up

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
@@ -349,9 +349,10 @@ object NodeUnitTest extends P2PLogger {
   def destroyNode(node: Node)(implicit
       config: BitcoinSAppConfig,
       ec: ExecutionContext): Future[Unit] = {
-    node
-      .stop()
-      .flatMap(_ => ChainUnitTest.destroyAllTables())
+    for {
+      _ <- ChainUnitTest.destroyAllTables()
+      _ <- node.stop()
+    } yield ()
   }
 
   def destroyNodeConnectedWithBitcoind(
@@ -445,6 +446,7 @@ object NodeUnitTest extends P2PLogger {
     val destroyedF = for {
       _ <- destroyNode(fundedWalletBitcoind.node)
       _ <- BitcoinSWalletTest.destroyWalletWithBitcoind(walletWithBitcoind)
+      _ <- appConfig.walletConf.stop()
     } yield ()
 
     destroyedF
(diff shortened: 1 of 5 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -117,22 +117,23 @@ abstract class Wallet
         } else FutureUtil.unit
     } yield ()
 
-  override def start(): Future[Unit] = {
+  override def start(): Future[WalletApi] = {
```
benthecarman: nit: `Future[Wallet]`

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -117,22 +117,23 @@ abstract class Wallet
...
       _ <- walletConfig.start()
       _ <- downloadMissingUtxos
     } yield {
       startWalletThread()
+      this
     }
   }
 
-  override def stop(): Unit = {
+  override def stop(): Future[WalletApi] = {
```
benthecarman: nit: `Future[Wallet]`

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -349,9 +349,10 @@ object NodeUnitTest extends P2PLogger {
   def destroyNode(node: Node)(implicit
       config: BitcoinSAppConfig,
       ec: ExecutionContext): Future[Unit] = {
-    node
-      .stop()
-      .flatMap(_ => ChainUnitTest.destroyAllTables())
+    for {
+      _ <- ChainUnitTest.destroyAllTables()
+      _ <- node.stop()
+    } yield ()
```
benthecarman, line 355: These changed order, is that intended?

Christewart: Yes, `node.stop()` also calls `chainAppConfig.stop()` since there is a hard dependency between the two. However if we stop the chain thread pool before we drop the tables, we will get an error.

https://github.com/bitcoin-s/bitcoin-s/pull/1868/files#diff-16bb4997cf1e0749ed67b9939c98c958R155

benthecarman (approved):

PR #1869, opened 2020-08-21
Correcly use reference.conf file

benthecarman:
Previously, our reference.conf file wasn't being used because it was in `app/server/resources` and `AppConfig` looks for it in `db-commons/resources/`. This moves it to it's correct location.

This revealed a bug with the `filter-header-batch-size` config option where it would not be able to read the default value because having the network specific made it an OBJECT type thus overwriting it. This changes it to have a `default` option instead.

Fixed some typos as well.


db-commons/src/main/resources/reference.conf
```diff
@@ -50,7 +50,7 @@ bitcoin-s {
 
     chain {
         neutrino {
-            filter-header-batch-size = 2000
+            filter-header-batch-size.default = 2000
```
Christewart, line 53: I think we need to update documentation on the website

https://bitcoin-s.org/docs/next/config/configuration#example-configuration-file

and probably `testkit/reference.conf` too

Christewart (changes requested):

db-commons/src/main/resources/reference.conf
benthecarman: Fixed

Christewart (approved):

PR #1872, opened 2020-08-21
Simplify Transaction Broadcast

benthecarman:
Using `sendInventoryMessage` will only send the txId and require the other peer to request it before they fully receive the transaction. This changes it to `sendTransactionMessage` and send the transaction initially.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/Node.scala
+++ b/node/src/main/scala/org/bitcoins/node/Node.scala
@@ -207,13 +207,21 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
         logger.error(s"Error when writing broadcastable TX to DB", exception)
       case Success(written) =>
         logger.debug(
-          s"Wrote tx=${written.transaction.txIdBE} to broadcastable table")
+          s"Wrote tx=${written.transaction.txIdBE.hex} to broadcastable table")
     }
 
     for {
       _ <- addToDbF
       peerMsgSender <- peerMsgSenderF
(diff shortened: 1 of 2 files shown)


node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -212,8 +212,8 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
     for {
       _ <- addToDbF
       peerMsgSender <- peerMsgSenderF
-      _ = logger.info(s"Sending out inv for tx=${transaction.txIdBE}")
-      _ <- peerMsgSender.sendInventoryMessage(transaction)
+      _ = logger.info(s"Sending out txmessage for tx=${transaction.txIdBE}")
+      _ <- peerMsgSender.sendTransactionMessage(transaction)
```
Christewart, line 225: Seems like this is a privacy leak? We cool with this for now?

https://developer.bitcoin.org/reference/p2p_networking.html#tx

benthecarman: Hmm I didn't know this. Maybe not the best idea then? On one hand we are already revealing that we are a Bitcoin-S node through our user agent and it is relatively easy to find out that we are only broadcasting our own txs, on the other hand, we should probably do the _correct_ thing for when we want to implement a mempool.

Maybe create 2 different broadcasting functions, `broadcastTransaction` and `broadcastTransactionLazy`, using an `inv` message and a `tx` message respectively. So we can have support for both and use the lazy version for now.

Thoughts?

Christewart (approved):

Issue #1874, opened 2020-08-21
Remove AppConfig.initialize() and move all code into AppConfig.start()


Christewart:
As the title says, I think this is more intuitive? I don't think we need both `.initialize()` and `.start()`

https://github.com/bitcoin-s/bitcoin-s/blob/master/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala#L37-L40

Issue #1871, opened 2020-08-21
Check scalatest fixture destruction code for resource leaks


Christewart:
Basically we need to make sure we are calling `appConfig.stop()` when destroying fixtures, otherwise we will end up having a bunch of threads laying around that are never used that will cause your machine to crash eventually if you run enough tests.

We have gotten away with these resource leaks for awhile because [we have just been spinning up new jvms](https://github.com/bitcoin-s/bitcoin-s/blob/master/project/CommonSettings.scala#L115-L120) for `chainTest`/`walletTest`, and then when that JVM gets destroyed, the threads automatically get cleaned up afterwards.

However, with the `nodeTest` we do not fork the jvm, thus the threads do _not_ get cleaned up by virtue of shutting down the JVM. This is what #1868 is trying to solve in the node project. 

It appears that we may also be seeing some of this in #1844 -- specifically 75549ca . 

We probably want to consider at least writing unit tests for the `ChainAppConfig`, `WalletAppConfig`, `NodeAppConfig` to make sure they are cleaning up after themselves correctly. Not sure what the best way is to maek sure we don't re-introduce resource leaks into our fixture code though.

PR #1875, opened 2020-08-21
2020 08 21 clean broadcast dao

Christewart:
Addresses parts of #1874 and #1871 

1. Extends `StartStopAsync` in BitcoinSAppConfig()
2. Adds `CachedAppConfig` trait 
3. Makes sure `P2PClientTest` cleans up after itself!

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NodeAppConfigTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NodeAppConfigTest.scala
@@ -9,6 +9,8 @@ import org.bitcoins.core.config.MainNet
 import ch.qos.logback.classic.Level
 import java.nio.file.Files
 
+import scala.concurrent.Await
+
 class NodeAppConfigTest extends BitcoinSAsyncTest {
   val tempDir = Files.createTempDirectory("bitcoin-s")
 
@@ -25,6 +27,7 @@ class NodeAppConfigTest extends BitcoinSAsyncTest {
     val mainnetConf = ConfigFactory.parseString("bitcoin-s.network = mainnet")
     val mainnet: NodeAppConfig = withOther.withOverrides(mainnetConf)
     assert(mainnet.network == MainNet)
+    withOther.stop().map(_ => succeed)
   }
 
   it must "be overridable with multiple levels" in {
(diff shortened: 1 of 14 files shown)


testkit/src/main/scala/org/bitcoins/testkit/fixtures/NodeDAOFixture.scala
```diff
@@ -18,13 +19,19 @@ trait NodeDAOFixture extends NodeUnitTest {
...
+                },
+                destroy = () => destroyAppConfig(nodeConfig))(test)
+  }
+
+  private def destroyAppConfig(nodeConfig: NodeAppConfig): Future[Unit] = {
+    for {
+      _ <- nodeConfig.dropAll()
+      _ <- nodeConfig.stop()
+    } yield ()
+  }
```
benthecarman, line 36: Maybe add this to `AppConfig`

testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
```diff
@@ -43,7 +43,7 @@ trait BaseAsyncTest
     * Needed because the default execution context will become overloaded
     * if we do not specify a unique execution context for each suite
     */
-  implicit override lazy val executionContext: ExecutionContext =
+  implicit override def executionContext: ExecutionContext =
```
benthecarman, line 46: Why change this?

testkit/src/main/scala/org/bitcoins/testkit/fixtures/NodeDAOFixture.scala
Christewart: I don't think we should add the `dropAll()` to AppConfig as if I was actually using the library that would be _very_ unexpected to me -- dropping all data on `.stop()`

testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
Christewart: I was getting NullPointerExceptions to do with scala's initialization of traits and sub traits with `val`. With `def` these go away

node-test/src/test/scala/org/bitcoins/node/NodeAppConfigTest.scala
```diff
@@ -25,6 +27,7 @@ class NodeAppConfigTest extends BitcoinSAsyncTest {
     val mainnetConf = ConfigFactory.parseString("bitcoin-s.network = mainnet")
     val mainnet: NodeAppConfig = withOther.withOverrides(mainnetConf)
     assert(mainnet.network == MainNet)
+    withOther.stop().map(_ => succeed)
```
benthecarman: I think you need to stop `mainnet` as well

benthecarman (approved):

PR #1877, opened 2020-08-22
2020 08 22 chaintest cleanup

Christewart:
Built on top of #1875 

Addresses part of #1871 

This starts to clean up resources used in the chain project, it doesn't get everything yet, but it's a good start.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/config/ChainAppConfigTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/config/ChainAppConfigTest.scala
@@ -10,6 +10,8 @@ import org.bitcoins.testkit.chain.ChainUnitTest
 import org.bitcoins.testkit.util.FileUtil
 import org.scalatest.FutureOutcome
 
+import scala.concurrent.Await
+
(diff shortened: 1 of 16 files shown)


chain-test/src/test/scala/org/bitcoins/chain/config/ChainAppConfigTest.scala
```diff
@@ -81,7 +83,11 @@ class ChainAppConfigTest extends ChainUnitTest {
   }
 
   override def afterAll: Unit = {
-
     FileUtil.deleteTmpDir(chainAppConfig.baseDatadir)
+    val config1StopF = chainAppConfig.stop()
+    val config2StopF = config.stop()
+    val stopF = config1StopF.flatMap(_ => config2StopF)
```
benthecarman: Why not
```
val stopF = for {
	_ <- chainAppConfig.stop()
	_ <- config.stop()
} yield ()
```

benthecarman (changes requested):

benthecarman (approved):

PR #1878, opened 2020-08-22
Fix async bug with destruction of fixtures

Christewart:
This was found in #1877 but I don't want it to get bogged down in review for other PRs.

Basically we were dropping the future on the ground when destructing fixtures. This could link to async bugs between `afterAll()` executing and clashing with fixture destruction. We should always have fixture destruction finish before we start executing `afterAll()`


testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala
```diff
@@ -36,14 +36,21 @@ trait BitcoinSFixture extends BitcoinSAsyncFixtureTest {
           FutureOutcome.failed(err).toFuture
       }
 
-    val futOutcome: FutureOutcome = new FutureOutcome(outcomeF)
-
-    val result: FutureOutcome = futOutcome.onOutcomeThen { _ =>
-      fixtureF.flatMap(f => destroy(f))
```
Christewart, line 42: here is where we we just drop this expression on the ground

benthecarman (approved):

PR #1885, opened 2020-08-24
Move all remaining wallet db representations

benthecarman:
Part of #1284 

Moves all the remaining wallet db representations to core. I also ran optimize imports on the `wallet` and `walletTest` projects to hopefully cut down on the number of rebase errors in the future.


nkohen (approved):
ACK

PR #1891, opened 2020-08-24
Apply string factory to a bunch of easy things

Christewart:
~Apply `StringFactory` trait to some easy things~

I ended up applying this to everything in the project, since we have pretty much broken every API in v0.4 we might as well break these as well and hopefully have stability in v0.5

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelectionAlgo.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelectionAlgo.scala
@@ -1,9 +1,11 @@
 package org.bitcoins.core.api.wallet
 
+import org.bitcoins.crypto.StringFactory
+
 /** Represents the various ways the wallet can do coin selection */
 sealed abstract class CoinSelectionAlgo
 
-object CoinSelectionAlgo {
+object CoinSelectionAlgo extends StringFactory[CoinSelectionAlgo] {
 
   /** Randomly selects utxos until it has enough to fund the desired amount,
     * should only be used for research purposes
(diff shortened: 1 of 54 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelectionAlgo.scala
```diff
@@ -32,7 +34,13 @@ object CoinSelectionAlgo {
...
 
-  def fromString(str: String): Option[CoinSelectionAlgo] = {
+  override def fromStringOpt(str: String): Option[CoinSelectionAlgo] = {
     all.find(state => str.toLowerCase() == state.toString.toLowerCase)
   }
+
+  override def fromString(string: String): CoinSelectionAlgo = {
+    val algoOpt = fromStringOpt(string)
+    algoOpt.getOrElse(
+      sys.error(s"Could not find coin selection algorithm=${string}"))
```
benthecarman, line 44: why `sys.error` over `IllegalArgumentExeception` or `RuntimeExeception`?

benthecarman (commented):
lgtm otherwise

core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelectionAlgo.scala
Christewart: sys.error is a runtime exception, it's the more "Scala way" to do it. I think there is a marginal benefit for using `sys.error()` because other runtimes -- such as scalajs -- have an easier way to decouple from jvm specific feature (`java.lang.RuntimeException`) and replace this call with whatever javascript throws for exceptions. :man_shrugging: . I've been trying to use that more.

Here's the definition:

```scala
package object sys {
  /** Throw a new RuntimeException with the supplied message.
   *
   *  @return   Nothing.
   */
  def error(message: String): Nothing = throw new RuntimeException(message)
```

EDIT: So _someday_ we may have an easier time refactoring to get bitcoin-s to work on javascript :-) https://www.scala-js.org/

benthecarman (approved):

PR #1895, opened 2020-08-25
Updated setup and dlc docs

nkohen:


docs/getting-setup.md
```diff
@@ -136,6 +136,18 @@ and then finally test that `Secp256k1Context.isEnabled()` as in Step 2.
...
+
+```bashrc
+sbt gui/universal:stage
+```
+
+to build and then the following (once setup is complete) to run:
+
+```bashrc
+./app/gui/target/universal/stage/bin/bitcoin-s-gui
+```
```
benthecarman, line 149: I think it's easier to just have them use `sbt gui/run` that's what I always do. Then they also don't need to make sure they build a new one every update

benthecarman (approved):

docs/wallet/dlc.md
```diff
@@ -179,45 +168,15 @@ where the `eventid` is in all but the messages other than the DLC Offer message,
...
 
-#### Mutual Close
+#### Execute
 
-Upon receiving an oracle signature, either party can initiate a mutual close with
+Upon receiving an oracle signature, you can execute the DLC unilaterally with
 
 ```bashrc
-./app/cli/target/graalvm-native-image/bitcoin-s-cli initdlcmutualclose --eventid [eventid] --oraclesig [sig] --escaped
+./app/cli/target/graalvm-native-image/bitcoin-s-cli executedlc --eventid [eventid] --oraclesig [sig]
```
Christewart: So I think we should take out stuff about `graalvm-native-image` for now as you need to install extra stuff and me building it did not work. It would be simpler if he just uses the standard jar.

Christewart (approved):

PR #1896, opened 2020-08-25
Clean up calls in Node and Chain

benthecarman:
Nothing glaringly bad, just some basic clean up

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -355,7 +355,7 @@ case class ChainHandler(
     blockHeaderDAO.getAtHeight(height)
 
   /** @inheritdoc */
-  override def getFilterHeaderCount(): Future[Int] = {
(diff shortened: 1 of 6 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -367,7 +355,7 @@ case class ChainHandler(
     blockHeaderDAO.getAtHeight(height)
 
   /** @inheritdoc */
-  override def getFilterHeaderCount: Future[Int] = {
+  override def getFilterHeaderCount(): Future[Int] = {
```
Christewart: nit: Since these don't modify the database we shouldn't use `()`. This should be saved for _writes_ to the database.

Christewart (changes requested):

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
benthecarman: fixed

Christewart (approved):

PR #1897, opened 2020-08-25
Add an ability to one Postgres database for all sub-projects

### LINKED ISSUE #1883: Postgres database structure
Currently we create 3 separate databases for postgres -- which we introduced in #1315 -- similar to sqlite

1. chaindb
2. walletdb
3. nodedb

However, when using a stand alone server that can support connections pools and also supports authentication we should probably have all tables inside of database perhaps divided by schema. 

We also then don't have to spin 3 indpendent postgres database instances for tests.

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/1883

### DIFF
--- a/docs/config/configuration.md
+++ b/docs/config/configuration.md
@@ -230,9 +230,11 @@ bitcoin-s {
     common {
         profile = "slick.jdbc.PostgresProfile$"
         db {
+            driver = org.postgresql.Driver
             url = "jdbc:postgresql://localhost:5432/database"
             username = "user"
             password = "topsecret"
+            numThreads = 5
         }
     }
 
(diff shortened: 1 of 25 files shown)


docs/config/configuration.md
```diff
@@ -239,6 +239,8 @@ bitcoin-s {
 }
 ```
 
+The database driver will create a separate SQL namespace for each sub-project: `chain`, `node` and `wallet`.
```
Christewart, line 252: I think more on this page needs to be updated

https://bitcoin-s.org/docs/next/config/configuration#database-configuration

For example, do you need `username` and `password` for each module in the configuration now? 

Can't you just say 

```
bitcoin-s.chain.profile = bitcoin-s.common.profile
bitcoin-s.chain.db = bitcoin-s.common.db
bitcoin-s.wallet.profile = bitcoin-s.common.profile
bitcoin-s.wallet.db = bitcoin-s.common.db 
```

now?

Christewart (changes requested):
So the code looks good to me, but don't we need to re-write migrations?

Christewart (approved):

PR #1898, opened 2020-08-25
DLC TLVs

nkohen:
Introduces TLVs for `DLCOffer`, `DLCAccept`, and `DLCSign` as specified in https://github.com/discreetlogcontracts/dlcspecs/blob/cf7e49e6a852c99e076777fbb96142e626cac641/Protocol.md

### DIFF
--- a/crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
@@ -121,6 +121,16 @@ object ECDigitalSignature extends Factory[ECDigitalSignature] {
     }
   }
 
+  def fromFrontOfBytes(bytes: ByteVector): ECDigitalSignature = {
+    val sigWithExtra = fromBytes(bytes)
+    val sig = fromRS(sigWithExtra.r, sigWithExtra.s)
+
+    require(bytes.startsWith(sig.bytes),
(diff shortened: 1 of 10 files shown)


benthecarman:
doesn't compile

crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
```diff
@@ -121,6 +121,16 @@ object ECDigitalSignature extends Factory[ECDigitalSignature] {
     }
   }
 
+  def fromFrontOfBytes(bytes: ByteVector): ECDigitalSignature = {
+    val sigWithExtra = fromBytes(bytes)
+    val sig = fromRS(sigWithExtra.r, sigWithExtra.s)
+
+    require(bytes.startsWith(sig.bytes),
+            s"Received weirdly encoded signature at beginning of $bytes")
+
```
benthecarman, line 130: Should you require that `sig == sigWithExtra`?

benthecarman (commented):
lgtm otherwise

crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
nkohen: no, `sigWithExtra` has all the bytes

PR #1900, opened 2020-08-25
Send get filters message if we haven't cached any yet

### LINKED ISSUE #1059: Filter headers doesn't continue syncing if you kill node during IBD of filters
If you stop the node during IBD of filters, and restart your node, you will not start syncing filter headers again when you restart the node. This means that your node is only partially synced with no way to restart the sync process.

benthecarman:
Fixes #1059

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -209,7 +209,13 @@ case class DataMessageHandler(
                        "which is less than max. This means we are synced,",
                        "not requesting more.")
                     .mkString(" "))
-                if (appConfig.nodeType == NodeType.NeutrinoNode && !syncing)
+                // If we are in neutrino mode, we might need to start fetching filters and their headers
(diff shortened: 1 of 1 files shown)


Christewart:
Needs rebase

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -209,7 +209,9 @@ case class DataMessageHandler(
                        "which is less than max. This means we are synced,",
                        "not requesting more.")
                     .mkString(" "))
-                if (appConfig.isNeutrinoEnabled && !syncing)
+                if (
```
Christewart, line 216: since this is kind of a nasty expression, can you just put a comment above the `if` saying what this checking in plain english?

Christewart: maybe link to the issue in source code too?

benthecarman:
Rebased and added comment

Christewart (approved):

PR #1902, opened 2020-08-25
Create DatabaseDriver ADT instead of booleans

benthecarman:


Christewart (approved):
Oooh, speaking my language now! I really like this :-)

Christewart:
Maybe we should wait for #1897 to get in though as it might affect this PR?

PR #1907, opened 2020-08-26
Remove AppConfig.initialize() in favor of AppConfig.start()

### LINKED ISSUE #1874: Remove AppConfig.initialize() and move all code into AppConfig.start()

Christewart:
fixes #1874

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
@@ -10,7 +10,7 @@ import org.bitcoins.core.util.{BitcoinSLogger, StartStopAsync}
 import slick.basic.DatabaseConfig
 import slick.jdbc.JdbcProfile
 
-import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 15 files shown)


db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -10,7 +10,7 @@ import org.bitcoins.core.util.{BitcoinSLogger, StartStopAsync}
 import slick.basic.DatabaseConfig
 import slick.jdbc.JdbcProfile
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{Future}
```
benthecarman, line 13: nit: Don't need `{}`

benthecarman (approved):

PR #1921, opened 2020-08-27
Warn if peer does not support services we need

### LINKED ISSUE #1917: Report error in logs if the bitcoind peer is not neutrino compatible
As reported by Nicolas Dorier, we should check service bits (?) of the node we are connecting to to ensure that they are neutrino compatible. If they are not, we _atleast_ log that they do not have neturino support and probably disconnect the client (?)

benthecarman:
Closes #1917

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
@@ -19,7 +19,7 @@ import org.bitcoins.node.networking.peer.PeerMessageReceiverState.{
   Normal,
   Preconnection
 }
-import org.bitcoins.node.{NodeCallbacks, P2PLogger}
+import org.bitcoins.node.{NodeCallbacks, NodeType, P2PLogger}
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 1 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -163,6 +163,21 @@ class PeerMessageReceiver(
           case good: Initializing =>
             val newState = good.withVersionMsg(versionMsg)
 
+            nodeAppConfig.nodeType match {
+              case NodeType.NeutrinoNode =>
+                if (!versionMsg.services.nodeCompactFilters) {
```
Christewart, line 169: Is there a point in continuing node operations if neutrino isn't enabled? Would it be nicer just to fail fast and get a nice error message?

benthecarman: Maybe once we have peer discovery, for now though I agree, it'd be nice to just fail entirely

Christewart: Maybe file a future issue for this?

Christewart (approved):

PR #1926, opened 2020-08-27
Implement getBestFilterHeader based on a number of block headers that…

### LINKED ISSUE #1919: Filter headers failing to sync
This morning i'm trying to sync my node and failing to sync my filter headers. This was run on  267cf06fa7a033e83313883de6511b673b5913b5  . If you look at the logs it does appear i sync block headers ok

This is where things stop happening with my node:

>2020-08-27T13:51:24 DEBUG [P2P] Unaligned bytes after this: 0
2020-08-27T13:51:24 DEBUG [P2P] Received message=cfheaders from peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) state=Normal 
2020-08-27T13:51:24 INFO [P2P] Got 386 compact filter header hashes
2020-08-27T13:52:48 INFO [P2P] Stopping node
2020-08-27T13:52:48 WARN [P2P] Cannot disconnect client that is not connected to 

As you can see, there is about ~1.5 minutes of nothing happening until I kill my node

Logs: 
https://pastebin.com/RjUi7yDE

Christewart:
… can be passed in as a parameter. These headers can be used to indicate what your current best chain is

My proposals to fix #1919 

Basically, I created

```scala
def getBestFilterHeaderForHeaders(
      headers: Vector[BlockHeaderDb]): Future[Option[CompactFilterHeaderDb]]
```

This allows you to pass in `headers` that are part of your best chain, and we will join the filter header table and block header table looking for the filter header with the max chain work. 

I have not tested this out locally yet to see if this fixes my problem.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -384,9 +384,13 @@ case class ChainHandler(
 
     for {
       filterHeaders <- filterHeadersOptF
-      maxHeightOpt = filterHeaders.flatten.maxByOption(_.height)
     } yield {
-      maxHeightOpt
+      val flattened = filterHeaders.flatten
+      if (flattened.isEmpty) None
+      else {
+        Some(flattened.maxBy(_.height))
+      }
+
     }
 
   }
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -374,7 +374,21 @@ case class ChainHandler(
 
   /** @inheritdoc */
   override def getBestFilterHeader(): Future[Option[CompactFilterHeaderDb]] = {
-    filterHeaderDAO.getBestFilterHeader
+    val bestFilterHeadersInChains = blockchains.map { blockchain =>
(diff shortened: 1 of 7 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -374,7 +374,21 @@ case class ChainHandler(
 
   /** @inheritdoc */
   override def getBestFilterHeader(): Future[Option[CompactFilterHeaderDb]] = {
-    filterHeaderDAO.getBestFilterHeader
+    val bestFilterHeadersInChains = blockchains.map { blockchain =>
```
Christewart: For this to be a truely general solution, we will probably need some sort of recursive step here to look at previous headers if we zero matches in the `blockchains` we have in memory.

Christewart: Thinking about this more, i think we can do something clever with `compactFilterHeaderDAO.maxHeight` to avoid the need to endlessly iterate through the blockchain.

benthecarman (commented):
This makes sense to me, did it fix your issue?

Christewart:
> This makes sense to me, did it fix your issue?

It does not yet! :-(. Still working.

benthecarman (approved):

Christewart:
1d0d8d612d56fe2b6e527ca016ee85d41c5597d1 fixes my issue! :tada:

PR #1932, opened 2020-08-28
Cache DoubleSha256Digest.hex

Christewart:
I'm syncing a testnet node and watching visualvm, apparently `DoubleSha256Digest.hex` is a host spot during IBD, let's just cache the hex representation to avoid this:

Screenshot from 2020-08-28 15-12-53


benthecarman:
Should just do this with every `HashDigest`

PR #1938, opened 2020-08-31
Update callbacks for LN support

rorp:


wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
```diff
@@ -15,6 +16,10 @@ import scala.concurrent.{ExecutionContext, Future}
   */
 trait WalletCallbacks {
 
+  def onBlockTransactionProcessed: CallbackHandler[
```
Christewart: Can you add a scaladoc on when this is called? This is done after all "internal" wallet processing is done right?

wallet/src/main/scala/org/bitcoins/wallet/WalletCallbacks.scala
```diff
@@ -53,6 +67,9 @@ trait WalletCallbacks {
 }
 
 /** Callback for handling a processed transaction */
+trait OnBlockTransactionProcessed
```
Christewart: What is different between `OnBlockTransactionProcessed` and `OnTransactionProcessed`

Christewart (changes requested):
The wallet callbacks need to be updated: https://bitcoin-s.org/docs/next/wallet/wallet-callbacks#callbacks

and we need to add a section on the website for chain callbacks

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -46,6 +47,8 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
 
   val peer: Peer
 
+  val initialSyncDone: Option[Promise[Done]]
```
Christewart: I don't understand what is happening here. Can you add documentation to it? It's also concerning that this is public and can be completed completed by anyone. Shouldn't `DataMessageHandler` just have the reference to the `Promise` and everyone else that is using this would just have a `initSyncDoneF: Option[Future[Done]]`?

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/Node.scala
rorp: This is an initial sync latch,  eg:
https://github.com/rorp/eclair/blob/bitcoin-s-integration/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala#L148

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -176,6 +180,9 @@ case class DataMessageHandler(
         Future.successful(this)
       case HeadersMessage(count, headers) =>
         logger.info(s"Received headers message with ${count.toInt} headers")
+        if (count.toInt == 0) {
+          Try(initialSyncDone.map(_.success(Done)))
+        }
```
benthecarman: If we're doing IBD and finish downloading block headers this will signal that we are done syncing even though we have no filter headers or filters

benthecarman:
Looks like formatting is incorrect as well

benthecarman:
Still has formatting issues

testkit/src/main/resources/reference.conf
```diff
@@ -6,7 +6,7 @@ bitcoin-s {
         # Ignore bitcoin-s logging config and use a logback config
         logback = true
 
-        level = OFF # trace, debug, info, warn, error, off
+        level = debug # trace, debug, info, warn, error, off
```
benthecarman: should move back to off

Issue #1910, opened 2020-08-26
Entering DLC with Ben ends up having Cannot have signers that do not sign for one of error


Christewart:
Error executing command:
Request failed: requirement failed: Cannot have signers that do not sign for one of Vector(ECPublicKey(02eea4707c7f47eba5b95d9a57c3120fc85dbdeea408d6edd262403d2f5f5836e3))

I got when attempting to generate and offer for a DLC 

This occurred on #1899 and commit 3d7ffb5d1c6ee73ef894ed51bf417d0a0d28d6dc

Christewart:
This occurred again on after doing a full rescan of the wallet on   61b2f443e30337acc45925a74f87eb1af3a006f1. Note that it's a _different_ public key that is causing the error this time which is even more concerning

```
Error executing command:
Request failed: requirement failed: Cannot have signers that do not sign for one of Vector(ECPublicKey(0220cc2033ec002d595dad4a9228b93eae658949747c0e3b63e6ea9aeee8d81a45))
```

Here are the logs

https://pastebin.com/EF8jiim9


EDIT:

This seems to apply to just normal "spends" from the wallet
````
[info] 2020-08-31T21:13:20 INFO [HTTP] Unhandled error in server:
[info] java.lang.IllegalArgumentException: requirement failed: Cannot have signers that do not sign for one of Vector(ECPublicKey(029c1ff414edd7b913f0822f26c72d56f195791636fcc0e8b5cae1c034051a8580))
[info]  at scala.Predef$.require(Predef.scala:340)
[info]  at org.bitcoins.core.wallet.utxo.InputSigningInfo.$init$(InputSigningInfo.scala:39)
[info]  at org.bitcoins.core.wallet.utxo.ScriptSignatureParams.<init>(InputSigningInfo.scala:65)
[info]  at org.bitcoins.core.api.wallet.db.SpendingInfoDb.toUTXOInfo(SpendingInfoDb.scala:200)
[info]  at org.bitcoins.core.api.wallet.db.SpendingInfoDb.toUTXOInfo$(SpendingInfoDb.scala:186)
[info]  at org.bitcoins.core.api.wallet.db.SegwitV0SpendingInfo.toUTXOInfo(SpendingInfoDb.scala:30)
[info]  at org.bitcoins.core.api.wallet.db.SpendingInfoDb.toUTXOInfo(SpendingInfoDb.scala:183)
[...]

PR #1944, opened 2020-08-31
Only mark utxos as reserved on successful coin selection

benthecarman:
Previous we could fail to select the utxos (like in #1910) but still reserve the utxos. This changes it to be the last thing we do so we avoid having reserved utxos even when we failed to select them

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
@@ -103,15 +103,11 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
     val selectedUtxosF: Future[Vector[SpendingInfoDb]] =
       for {
         walletUtxos <- utxosF
-        //currently just grab the biggest utxos
         utxos = CoinSelector.selectByAlgo(coinSelectionAlgo = coinSelectionAlgo,
                                           walletUtxos = walletUtxos,
                                           outputs = destinations,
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -147,6 +143,9 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
 
         }
       }
+      _ <-
+        if (markAsReserved) markUTXOsAsReserved(addrInfosWithUtxo.map(_._1))
```
Christewart, line 148: Even here things can fail later on in the `for` expression. We probably want to do something like 

```scala

 val resultF = for { ... } yield { ... }

resultF.recoverWith { _ => 
  //unreserve utxos since we failed to sign
  //any other cleanup things?
}

resultF
```

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
benthecarman: Would like to be able to do that but we don't have access to which utxos we selected outside of the future.

Christewart:
That doesn't seem ideal? Why is that?

benthecarman:
> That doesn't seem ideal? Why is that?

nvm I was wrong

Christewart (approved):

PR #1946, opened 2020-09-01
Rename HDChainType to HDChangeType

### LINKED ISSUE #1945: Rename HDChainType -> HDChangeType
I believe this is a typo? 

https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki

benthecarman:
Fixes #1945


Christewart:
I think i was wrong with #1945, it does actually have other semantics of 'chain' i believe?

benthecarman:
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#change

So it looks like it specified as `external chain` and `internal chain`, however, these chains refer to if it is change or not.

Imo, it is fine either way, which generally means do not make the change.

PR #1948, opened 2020-09-01
Validate key manager matches account db on wallet startup

benthecarman:
This should not allow the wallet to startup without having your account db and key manager match.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -117,9 +117,34 @@ abstract class Wallet
         } else FutureUtil.unit
     } yield ()
 
+  private def checkRootAccount: Future[Unit] = {
+    val coinType = HDUtil.getCoinType(keyManager.kmParams.network)
+    val coin =
+      HDCoin(purpose = keyManager.kmParams.purpose, coinType = coinType)
+    val account = HDAccount(coin = coin, index = 0)
+    // safe since we're deriving from a priv
+    val xpub = keyManager.deriveXPub(account).get
+
+    accountDAO.read(account.coin, account.index).flatMap {
+      case Some(account) =>
+        if (account.xpub != xpub) {
+          val errorMsg =
+            s"Divergent xpubs for account=$account. Existing database xpub=${account.xpub}, key manager's xpub=$xpub. " +
+              s"It is possible we have a different key manager being used than expected, key manager=$keyManager"
+          Future.failed(new RuntimeException(errorMsg))
+        } else {
+          FutureUtil.unit
+        }
+      case None =>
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -117,9 +117,34 @@ abstract class Wallet
         } else FutureUtil.unit
     } yield ()
 
+  private def checkRootAccount: Future[Unit] = {
```
Christewart, line 120: This silently fails for me. The error does not get propogated for some reason.

Christewart: Ok this is very strange, it _is_ failing for me locally, and i finally got it to spit out an error message with this custom code. However I don't understand why the error is propogating back upwards in the call stack. Do you have any idea? Could you try and "trivially fail" (sys.error(s"trivial fail!")) and see if you get the error to propagate? 

```scala
    accountDAO.read(account.coin, account.index).flatMap {
      case Some(account) =>
        logger.info(s"account=$account")
        if (account.xpub != xpub) {
          val errorMsg =
            s"Divergent xpubs for account=$account. Existing database xpub=${account.xpub}, key manager's xpub=$xpub. " +
              s"It is possible we have a different key manager being used than expected, key manager=$keyManager"
          logger.error(errorMsg)
          Future.failed(new RuntimeException(errorMsg))
        } else {
          logger.info(s"Xpubs matched!")
          FutureUtil.unit
        }
      case None =>
        logger.info(s"No accounts found!!")
        val errorMsg = s"Missing root xpub for account $account in database"
        Future.failed(new RuntimeException(errorMsg))
    }
```

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: #1950 should fix this

Christewart: I don't think so, we have a `.failed()` callback registered already don't we at the very end of Main.scala

```
    startFut.failed.foreach { err =>
      logger.error(s"Error on server startup!", err)
    }
```

benthecarman: I've tested and it does work. Not sure why the `logger.error` does not appear, maybe because our `AppConfig` has been stopped and we can't log?

benthecarman: Either way, discussion should occur on that PR

Christewart (approved):

PR #1950, opened 2020-09-01
Throw errors that occurr in startup

benthecarman:
Since `runMain()` returns `Unit` the application would just stall on an error. This ensures that we print the stack trace of the error and exit

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/ServerRunTest.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/ServerRunTest.scala
@@ -1,0 +1,34 @@
+package org.bitcoins.server
+
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.util.BitcoinSAsyncTest
+
+import scala.concurrent.Future
+import scala.reflect.io.Directory
+
+class ServerRunTest extends BitcoinSAsyncTest {
+
+  // Note: on this test passing it will output a stack trace
+  // because runMain calls err.printStackTrace()
+  it must "throw errors" in {
+    val datadir = BitcoinSTestAppConfig.tmpDir()
+    val directory = new Directory(datadir.toFile)
+
+    val args =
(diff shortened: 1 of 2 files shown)


app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -193,6 +193,8 @@ object Main extends App with BitcoinSLogger {
     }
     startFut.failed.foreach { err =>
       logger.error(s"Error on server startup!", err)
+      err.printStackTrace()
```
Christewart, line 195: Can we add a unit test for this in some way? It seems that what we need?

Continuing discussion here: https://github.com/bitcoin-s/bitcoin-s/pull/1948#discussion_r481614347

benthecarman: I'm not sure how we would, we can't really make tests for the server's `Main`

Christewart: `runMain()` is a method, you should be able to pass in bad configs somehow? We would have to make it non-private which i am ok with

benthecarman: Make `runMain()` take in `args` so we can test ^

app/server-test/src/test/scala/org/bitcoins/server/ServerRunTest.scala
```diff
@@ -0,0 +1,34 @@
...
+    val datadir = BitcoinSTestAppConfig.tmpDir()
+    val directory = new Directory(datadir.toFile)
+
+    val args =
+      Vector("--datadir", datadir.toAbsolutePath.toString)
+
+    // Use Exception because different errors can occur
+    recoverToSucceededIf[Exception] {
+      val runMainF = Main.runMain(args)
+      val deleteDirF = Future {
```
Christewart, line 23: Why is this inside a `Future`?

Christewart (approved):

Christewart (approved):

PR #1955, opened 2020-09-02
Create MultiSignatureScriptPubKey.toBIP67MultiSignatureScriptPubKey

benthecarman:
Eventually maybe we can create a separate type for `BIP67MultiSignatureScriptPubKey`


Christewart (commented):
So I do have a few concerns with this, namely we need to make sure we _carefully_ handle this in the wallet as re-ordering the public keys changes script hashes which changes addresses etc. I guess it isn't really the focus of this PR, but I wanted to make sure I write this somewhere.

benthecarman:
> So I do have a few concerns with this, namely we need to make sure we _carefully_ handle this in the wallet as re-ordering the public keys changes script hashes which changes addresses etc. I guess it isn't really the focus of this PR, but I wanted to make sure I write this somewhere.

Yeah, I agree, I think making this it's own type will help mitigate this but not entirely

Christewart:
> > So I do have a few concerns with this, namely we need to make sure we _carefully_ handle this in the wallet as re-ordering the public keys changes script hashes which changes addresses etc. I guess it isn't really the focus of this PR, but I wanted to make sure I write this somewhere.
> 
> Yeah, I agree, I think making this it's own type will help mitigate this but not entirely

Maybe we can add a flag on the parameters for `MultiSignatureScriptPubKey` like `isBIP67: Boolean` where we can automatically convert them to compatible keys internally, or default it to compatible keys? Idk

PR #1959, opened 2020-09-02
Optimize findClosestToTime

benthecarman:
Found this while looking through for places to use `_.maxByOption`. Here we are loading in every header before a given time, then doing the computation. Instead changed it to do that in the SQL query

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -207,10 +207,14 @@ case class BlockHeaderDAO()(implicit
     table.filter(header => header.height >= from && header.height <= to).result
   }
 
-  def findAllBeforeTime(time: UInt32): Future[Vector[BlockHeaderDb]] = {
-    val query = table.filter(_.time < time)
+  def findBeforeTime(time: UInt32): Future[Option[BlockHeaderDb]] = {
+    val beforeTime = table.filter(_.time < time)
 
-    database.run(query.result).map(_.toVector)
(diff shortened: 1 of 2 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -223,7 +227,13 @@ case class BlockHeaderDAO()(implicit
 
     opt.flatMap {
       case None =>
-        findAllBeforeTime(time).map(_.maxBy(_.time))
+        findBeforeTime(time).flatMap {
+          case None =>
+            Future.failed(
+              new IllegalStateException("No block headers in database."))
```
Christewart: Is this really an `IllegalState`? Probably should just be a `RuntimeException`

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -207,10 +207,14 @@ case class BlockHeaderDAO()(implicit
     table.filter(header => header.height >= from && header.height <= to).result
   }
 
-  def findAllBeforeTime(time: UInt32): Future[Vector[BlockHeaderDb]] = {
-    val query = table.filter(_.time < time)
+  def findBeforeTime(time: UInt32): Future[Option[BlockHeaderDb]] = {
```
Christewart: Probably should be rename to something like `findClosestBeforeTime()` or something like that? Otherwise I would think `findBeforeTime()` would return all headers before a specific time

Christewart (changes requested):

Christewart (approved):

PR #1962, opened 2020-09-02
 Rework ChainHandler.nextBlockHeadersBatchRange() to make sure the stopHash we return is in the best chain we have

### LINKED ISSUE #1919: Filter headers failing to sync

Christewart:
Attempts to fix #1919 

Previously inside of `ChainHandler.nextBlockHeaderBatchRange` we wouldn't check that the `stopHash` we were returning is _actually_ part of our best blockchain. This fixes this problem. I'm wondering if this problem happens with `ChainHandler.nextFilterHeaderBatchRange()` we as well, but i haven't investigated.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
@@ -474,6 +474,54 @@ class ChainHandlerTest extends ChainDbUnitTest {
       }
   }
 
+  it must "generate the correct range of block filters if a header is reorged" in {
+    chainHandler: ChainHandler =>
+      val reorgFixtureF = buildChainHandlerCompetingHeaders(chainHandler)
+      val chainHandlerF = reorgFixtureF.map(_.chainApi)
+      val newHeaderBF = reorgFixtureF.map(_.headerDb1)
+      val newHeaderCF = reorgFixtureF.map(_.headerDb2)
+      val batchSize = 100
+
+      //two competing headers B,C built off of A
+      //so just pick the first headerB to be our next block header batch
+      val assert1F = for {
+        chainHandler <- chainHandlerF
+        newHeaderB <- newHeaderBF
+        blockHeaderBatchOpt <- chainHandler.nextBlockHeaderBatchRange(
+          prevStopHash = ChainTestUtil.regTestGenesisHeaderDb.hashBE,
+          batchSize = batchSize)
+        count <- chainHandler.getBlockCount()
+      } yield {
+        assert(count == 1)
+        assert(blockHeaderBatchOpt.isDefined)
+        val Some((height, hash)) = blockHeaderBatchOpt
+        assert(newHeaderB.hash == hash)
+        assert(newHeaderB.height == height)
+      }
+
+      //now let's build a new block header ontop of C and process it
+      //when we call chainHandler.nextBlockHeaderBatchRange it
+      //should be C's hash instead of B's hash
+      for {
+        _ <- assert1F
+        chainHandler <- chainHandlerF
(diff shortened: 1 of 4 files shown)


chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -474,6 +474,54 @@ class ChainHandlerTest extends ChainDbUnitTest {
...
+        count <- chainHandler.getBlockCount()
+      } yield {
+        assert(count == 1)
+        assert(blockHeaderBatchOpt.isDefined)
+        val Some((height, hash)) = blockHeaderBatchOpt
+        assert(newHeaderB.hash == hash)
+        assert(newHeaderB.height == height)
+      }
+
+      //now let's build a new block header ontop of C and process it
```
benthecarman, line 502: nit: 
> ontop

on top

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -147,21 +147,51 @@ case class ChainHandler(
...
-        throw UnknownBlockHeight(s"Unknown header height ${stopHeight}"))
-    } yield {
-      if (startHeight > stopHeight)
-        None
-      else
-        Some((startHeight, stopBlock.hashBE.flip))
-    }
+      height <- startHeightF
+      inMemoryBlockchains =
+        blockchains.filter(_.find(_.height == height).isDefined)
```
benthecarman, line 154: can replace `_.find(..).isDefined` with `_.exists`

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -147,21 +147,51 @@ case class ChainHandler(
...
+            Future.successful(hashHeightOpt)
+          } else {
+            //fetch from database
+            val chainsF = blockHeaderDAO.getBlockchainsBetweenHeights(
+              from = height,
+              to = height + batchSize)
+
+            val hashHeightOptF = for {
+              chains <- chainsF
+            } yield getBestChainAtHeight(height, chains)
```
benthecarman, line 169: shouldn't this `height` be what we had before?

```
stopHeight = if (startHeight - 1 + batchSize > blockCount) blockCount
			 else startHeight - 1 + batchSize
```

core/src/main/scala/org/bitcoins/core/api/chain/db/BlockHeaderDb.scala
```diff
@@ -27,6 +27,8 @@ case class BlockHeaderDb(
 
     blockHeader
   }
+
+  lazy val hash: DoubleSha256Digest = hashBE.flip
```
benthecarman, line 31: Should look for other places we can do this

PR #1972, opened 2020-09-04
Add chainHash to ChainParams

benthecarman:
So we don't need to use `chain.genesisBlock.blockHeader.hash` every time we wan the genesis block hash. Will be nice for DLCs/LN that have `chain_hash` defined as this as well


core/src/main/scala/org/bitcoins/core/protocol/blockchain/ChainParams.scala
```diff
@@ -48,6 +42,10 @@ sealed abstract class ChainParams {
   /** The Genesis [[org.bitcoins.core.protocol.blockchain.Block Block]] in the blockchain. */
   def genesisBlock: Block
 
+  def chainHash: DoubleSha256Digest = genesisBlock.blockHeader.hash
```
Christewart: Can we call it `genesisHash`? `chainHash` is a bit ambiguous as a blockchain as a bunch of hashes

Christewart (approved):

PR #1977, opened 2020-09-04
Use randomized fee rates for wallet tests

### LINKED ISSUE #1976: Remove hard coded fee units from test cases
We aren't really testing robustness of our fee handling code because we have our fee unit hard coded at 3 sat/vybte

https://github.com/bitcoin-s/bitcoin-s/blob/a1e0fe1aa4ae3bc8b38194c1200a4db36b3a23ce/testkit/src/main/scala/org/bitcoins/testkit/wallet/DLCWalletUtil.scala#L145

We should be using a random generator here like `CurrencyUnitGenerator.feeUnit`

Christewart:
Fixes #1976 in master

Basically we should have test cases that have as input variable fees

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/db/OutgoingTransactionDb.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/db/OutgoingTransactionDb.scala
@@ -39,14 +39,16 @@ object OutgoingTransactionDb {
       s"sentAmount ($sentAmount) cannot be greater than the amount the wallet input ($inputAmount)")
 
     val feePaid = inputAmount - totalOutput
-    val feeRate = feePaid.satoshis.toLong / tx.baseSize.toDouble
+
+    // Calculate fee rate in satoshis per byte
+    val feeRate = feePaid.satoshis.toLong / tx.byteSize.toDouble
     OutgoingTransactionDb(
       tx.txIdBE,
       inputAmount,
       sentAmount,
       feePaid,
       expectedFee,
-      SatoshisPerByte.fromLong(feeRate.toLong)
+      SatoshisPerByte.fromLong(Math.round(feeRate))
     )
   }
 }
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
@@ -8,31 +8,31 @@ abstract class FeeUnitGen {
 
   def satsPerByte: Gen[SatoshisPerByte] = {
     for {
-      curr <- fee
+      curr <- realisticFee
     } yield SatoshisPerByte(curr)
   }
 
   def satsPerKiloByte: Gen[SatoshisPerKiloByte] = {
     for {
-      curr <- fee
+      curr <- realisticFee
     } yield SatoshisPerKiloByte(curr * 1000)
   }
 
   def satsPerKiloWeight: Gen[SatoshisPerKW] = {
     for {
-      curr <- fee
+      curr <- realisticFee
     } yield SatoshisPerKW(curr * 1000)
   }
 
   def satsPerVirtualByte: Gen[SatoshisPerVirtualByte] = {
     for {
-      curr <- fee
+      curr <- realisticFee
     } yield SatoshisPerVirtualByte(curr)
   }
 
   /** Choose between different sets of Satoshi sizes so we get a better distribution */
-  private def fee: Gen[Satoshis] = {
-    Gen.oneOf(realisticFee, highFee, exorbitantFee)
+  private def realisticFee: Gen[Satoshis] = {
+    Gen.oneOf(lowFee, highFee, exorbitantFee)
   }
 
   /** Max fee between 101 - 100 sats. So we can have
@@ -40,7 +40,7 @@ abstract class FeeUnitGen {
     * 2. 100 sats/vbyte
     * 3. 100 sats/kb
     */
-  private def realisticFee: Gen[Satoshis] = {
+  private def lowFee: Gen[Satoshis] = {
     Gen
       .choose(0, 100)
       .map(Satoshis(_))
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
@@ -1,0 +1,88 @@
+package org.bitcoins.testkit.core.gen
+
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.wallet.fee._
+import org.scalacheck.Gen
+
+abstract class FeeUnitGen {
+
+  def satsPerByte: Gen[SatoshisPerByte] = {
+    for {
+      curr <- fee
+    } yield SatoshisPerByte(curr)
+  }
+
+  def satsPerKiloByte: Gen[SatoshisPerKiloByte] = {
+    for {
+      curr <- fee
+    } yield SatoshisPerKiloByte(curr * 1000)
+  }
+
+  def satsPerKiloWeight: Gen[SatoshisPerKW] = {
+    for {
+      curr <- fee
+    } yield SatoshisPerKW(curr * 1000)
+  }
+
+  def satsPerVirtualByte: Gen[SatoshisPerVirtualByte] = {
+    for {
+      curr <- fee
+    } yield SatoshisPerVirtualByte(curr)
+  }
+
+  /** Choose between different sets of Satoshi sizes so we get a better distribution */
+  private def fee: Gen[Satoshis] = {
+    Gen.oneOf(realisticFee, highFee, exorbitantFee)
+  }
+
+  /** Max fee between 101 - 100 sats. So we can have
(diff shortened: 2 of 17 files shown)


nkohen:
Shouldn't there be assertions in these tests that feerates are being used?

benthecarman:
Every wallet made from a Wallet fixture will have a random fee rate generator

https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala#L378

Christewart:
> Every wallet made from a Wallet fixture will have a random fee rate generator
> 
> https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala#L378

>Shouldn't there be assertions in these tests that feerates are being used?

Do these exist?

benthecarman:
Doesn't look like we have assertions for the fee rates being used.

Christewart:
> Doesn't look like we have assertions for the fee rates being used.

So i think i'm going to have hand off this work to you but here are my thoughts on how we can do this:

https://github.com/Christewart/bitcoin-s-core/commit/4c26c24e20a68bf7cc512bdaa98710e4aa163337#diff-0f6aec5eb46481a2ae9ae7221a2ef713R98

Does this seem right/reasonable to you?

testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /** Max fee between 101 - 100 sats. So we can have
+    * 1. 100 sats/byte
+    * 2. 100 sats/vbyte
+    * 3. 100 sats/kb
+    */
+  private def realisticFee: Gen[Satoshis] = {
+    Gen
+      .choose(0, 100)
```
nkohen, line 45: I think we should make this a larger upper bound (maybe 1000?). Realistic should also include the improbable cases that are still realistic.

nkohen: at least this is what we mean by it so far in this codebase, I think this generator should be renamed to normalFee or lowFee and a separate generator should be made for realisitic which goes 0 - 1k

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -375,14 +377,15 @@ object BitcoinSWalletTest extends WalletLogger {
...
 
-  private[testkit] class RandomFeeProvider extends FeeRateApi {
-    private val rnd = new scala.util.Random(TimeUtil.now.toEpochMilli)
-    private val start = 2
-    private val end = 100
+  private[bitcoins] class RandomFeeProvider extends FeeRateApi {
+    // Useful for tests
+    var lastFeeRate: Option[FeeUnit] = None
 
     def getFeeRate: Future[FeeUnit] = {
```
nkohen: I assume this should have an override? Otherwise it shouldn't be a Future

wallet-test/src/test/scala/org/bitcoins/wallet/AddressTagIntegrationTest.scala
```diff
@@ -77,6 +75,7 @@ class AddressTagIntegrationTest extends BitcoinSWalletTest {
           .map(unconfirmed => assert(unconfirmed == valueFromBitcoind))
 
       account <- wallet.getDefaultAccount()
+      feeRate <- wallet.feeRateApi.getFeeRate
```
nkohen: nit: if the wallet API has a fee rate API on it already it might be nice to implement `getFeeRate` on the wallet as this looks kinda funny and not as user-friendly

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -375,14 +377,15 @@ object BitcoinSWalletTest extends WalletLogger {
     }
   }
 
-  private[testkit] class RandomFeeProvider extends FeeRateApi {
-    private val rnd = new scala.util.Random(TimeUtil.now.toEpochMilli)
-    private val start = 2
-    private val end = 100
+  private[bitcoins] class RandomFeeProvider extends FeeRateApi {
+    // Useful for tests
+    var lastFeeRate: Option[FeeUnit] = None
```
nkohen, line 382: Will this be problematic with Async tests? Or are we guaranteed to have one per fixture/test

core/src/main/scala/org/bitcoins/core/api/wallet/db/OutgoingTransactionDb.scala
```diff
@@ -39,14 +39,16 @@ object OutgoingTransactionDb {
       s"sentAmount ($sentAmount) cannot be greater than the amount the wallet input ($inputAmount)")
 
     val feePaid = inputAmount - totalOutput
-    val feeRate = feePaid.satoshis.toLong / tx.baseSize.toDouble
+
+    // Calculate fee rate in satoshis per byte
+    val feeRate = feePaid.satoshis.toLong / tx.byteSize.toDouble
```
Christewart, line 44: do we really not have some sort of centralized place to do this? It scares me when we just have one off fee rate calculations.

testkit/src/main/scala/org/bitcoins/testkit/core/gen/p2p/ControlMessageGenerator.scala
```diff
@@ -32,8 +32,10 @@ object ControlMessageGenerator {
 
   def feeFilterMessage: Gen[FeeFilterMessage] = {
     for {
-      fee <- CurrencyUnitGenerator.feeUnit.suchThat(
-        !_.isInstanceOf[SatoshisPerVirtualByte])
+      fee <- FeeUnitGen.feeUnit.suchThat(feeRate =>
```
Christewart: nit: It's usually a bad idea to use `.suchThat()` as it will cause generators to fail if it's hard to satisfy the predicate (in this case, `!feeRate.isInstanceOf[...] && ...`)

wallet-test/src/test/scala/org/bitcoins/wallet/WalletIntegrationTest.scala
```diff
@@ -110,28 +111,28 @@ class WalletIntegrationTest extends BitcoinSWalletTest {
...
-                                     outgoingTx.get.actualFee,
-                                     3.satoshi))
+        WalletTestUtil.isFeeRateCloseEnough(outgoingTx.get),
+        s"Actual fee=${outgoingTx.get.feeRate} expectedFeeRate=$feeRate"
+      )
+      balancePostSend <- wallet.getBalance()
+    } yield {
+      assert(outgoingTx.get.expectedFee == feeRate.calc(signedTx))
+      assert(
+        outgoingTx.get.expectedFee === outgoingTx.get.actualFee +- feeRate.currencyUnit, // could be off by one due to rounding
```
Christewart, line 122: off by one unit in fee measurement? I.e. `+- 40 sat/vbyte` if `40 sat/vbyte` was your original fee estimate?

benthecarman: No, in the measurement of the actual fee paid, `expectedFee` and `actualFee` are `CurrencyUnit` values. So `+- 40 sats` in the fee paid if the we used a fee rate of `40 sat/byte`

core/src/main/scala/org/bitcoins/core/api/wallet/db/OutgoingTransactionDb.scala
benthecarman: Agreed, I think it is out of scope for this PR however, but will file an issue to do so.

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
benthecarman: We should be guaranteed one fixture be test case

testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
benthecarman: I think this is just a poor naming of the function. I have a function above called `fee` that uses this as well as `highFee` and `exorbitantFee`. That way it gives a more even distribution from 0-1000 sats/byte.

benthecarman: Renamed to hopefully be more clear

core/src/main/scala/org/bitcoins/core/api/wallet/db/OutgoingTransactionDb.scala
benthecarman: #1989 #1990

nkohen (approved):

Christewart:
I guess since this was my PR originally I can't approve it on github, but :+1:

PR #1990, opened 2020-09-09
Create FeeUnitFactory and functions to calculate tx fee rates

### LINKED ISSUE #1989: Add ability to calculate a `FeeUnit` from a `Transaction`
Currently, to calculate a fee rate for a transaction we have to do it manually, it would be nice and much safer if we had a way to take in a transaction and the funding amount and return a given `FeeUnit`.

Context: https://github.com/bitcoin-s/bitcoin-s/pull/1977#discussion_r485668105

benthecarman:
Closes #1989

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
@@ -15,10 +15,12 @@ sealed abstract class FeeUnit {
   final def *(long: Long): CurrencyUnit = Satoshis(toLong * long / scaleFactor)
   def *(tx: Transaction): CurrencyUnit = calc(tx)
 
+  def factory: FeeUnitFactory[FeeUnit]
+
   /** The coefficient the denominator in the unit is multiplied by,
     * for example sats/kilobyte -> 1000
     */
-  def scaleFactor: Long
+  def scaleFactor: Long = factory.scaleFactor
 
   require(scaleFactor > 0,
           s"Scale factor cannot be less than or equal to 0, got $scaleFactor")
@@ -26,14 +28,14 @@ sealed abstract class FeeUnit {
   /** Takes the given transaction returns a size that will be used for calculating the fee rate.
     * This is generally the denominator in the unit, ie sats/byte
     */
-  def txSizeForCalc(tx: Transaction): Long
+  def txSizeForCalc(tx: Transaction): Long = factory.txSizeForCalc(tx)
 
   /** Calculates the fee for the transaction using this fee rate, rounds down satoshis */
   final def calc(tx: Transaction): CurrencyUnit = this * txSizeForCalc(tx)
   def toLong: Long = currencyUnit.satoshis.toLong
 }
 
-trait FeeUnitFactory[T <: FeeUnit] {
+trait FeeUnitFactory[+T <: FeeUnit] {
 
   /** The coefficient the denominator in the unit is multiplied by,
     * for example sats/kilobyte -> 1000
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -45,13 +70,38 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
     SatoshisPerKiloByte(currencyUnit.satoshis * Satoshis(1000))
   }
 
-  override def txSizeForCalc(tx: Transaction): Long = tx.byteSize
+  override def txSizeForCalc(tx: Transaction): Long =
+    SatoshisPerByte.txSizeForCalc(tx)
 
-  override def scaleFactor: Long = 1
+  override def scaleFactor: Long = SatoshisPerByte.scaleFactor
```
nkohen: Seeing as this is the common pattern, I think you should have the top-level trait have an abstract def `FeeUnitFactory` and default impls like this ^ defined at the top level

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -45,13 +70,38 @@ case class SatoshisPerByte(currencyUnit: CurrencyUnit) extends BitcoinFeeUnit {
...
+  override def txSizeForCalc(tx: Transaction): Long = tx.byteSize
+
+  override def calc(
+      inputAmount: CurrencyUnit,
+      tx: Transaction): SatoshisPerByte = {
+    val feePaid = inputAmount - tx.totalOutput
+    val feeRate = feePaid.satoshis.toLong / txSizeForCalc(tx).toDouble
+
+    fromLong(Math.round(feeRate))
+  }
```
nkohen: why is this here? This seems to be equivalent to the thing it overrides?

benthecarman: Woops, missed in my cleanup

nkohen (approved):

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -15,24 +15,53 @@ sealed abstract class FeeUnit {
...
+  /** Takes the given transaction returns a size that will be used for calculating the fee rate.
+    * This is generally the denominator in the unit, ie sats/byte
+    */
+  def txSizeForCalc(tx: Transaction): Long
+
+  /** Creates an instance T where the value given is the numerator of the fee unit
+    */
+  def fromLong(long: Long): T
+
+  def calc(inputAmount: CurrencyUnit, tx: Transaction): T = {
```
Christewart: What exactly is `inputAmount` here? A scala doc would be nice. You don't have an index if it's a specific input. If it's for all inputs we probably should rename to `totalInputAmount`

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -15,24 +15,53 @@ sealed abstract class FeeUnit {
...
+    val feePaid = inputAmount - tx.totalOutput
+    val feeRate = feePaid.satoshis.toLong / txSizeForCalc(tx).toDouble
+
+    fromLong(Math.round(feeRate * scaleFactor))
```
Christewart, line 67: Should document the rounding style in the scaladoc as well

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -93,12 +153,25 @@ case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
...
+    SatoshisPerVirtualByte
 }
 
-object SatoshisPerVirtualByte {
+object SatoshisPerVirtualByte extends FeeUnitFactory[SatoshisPerVirtualByte] {
+
+  /** The coefficient the denominator in the unit is multiplied by,
+    * for example sats/kilobyte -> 1000
+    */
+  override def scaleFactor: Long = 1
```
Christewart: nit: `val`

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -80,9 +122,27 @@ case class SatoshisPerKiloByte(currencyUnit: CurrencyUnit)
...
+  override def factory: FeeUnitFactory[SatoshisPerKiloByte] =
+    SatoshisPerKiloByte
+}
+
+object SatoshisPerKiloByte extends FeeUnitFactory[SatoshisPerKiloByte] {
 
+  /** The coefficient the denominator in the unit is multiplied by,
+    * for example sats/kilobyte -> 1000
+    */
   override def scaleFactor: Long = 1000
```
Christewart: nit `val`

Christewart (commented):
Looks good overall!

Christewart (approved):

Christewart (approved):

PR #1991, opened 2020-09-09
Happy Path Test Vectors!

nkohen:
Generated and added tests checking happy path test vectors.


dlc-test/src/test/scala/org/bitcoins/dlc/DLCTestVectorTest.scala
```diff
@@ -0,0 +1,44 @@
...
+}
+import org.bitcoins.testkit.util.BitcoinSAsyncTest
+import org.scalacheck.Gen
+
+import scala.concurrent.Future
+
+class DLCTestVectorTest extends BitcoinSAsyncTest {
+  behavior of "DLCTestVectors"
+
+  val randomTestVectorGen: Gen[Future[DLCTestVector]] =
```
benthecarman, line 17: This could be moved to `DLCTestVectorGen`

dlc-test/src/test/scala/org/bitcoins/dlc/DLCTestVectorTest.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  it should "have serialization symmetry" in {
+    forAllAsync(randomTestVectorGen) { testVectorF =>
+      testVectorF.map { testVector =>
+        assert(DLCTestVector.fromJson(testVector.toJson).contains(testVector))
+      }
+    }
+  }
+
+  it should "pass dlc_tests" in {
```
benthecarman: super nit: it is called `dlc_tests` here but the file is named `dlc_test`

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVectorGen.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.{File, PrintWriter}
+
+import play.api.libs.json._
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.io.Source
+
+object DLCTestVectorGen {
+  //import DLCTLVGen._
```
benthecarman: ping

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVectorGen.scala
```diff
@@ -0,0 +1,89 @@
...
+  import DLCTxGen._
+
+  implicit private val ec: ExecutionContext = ExecutionContext.global
```
benthecarman, line 13: Instead of using this why not just make `regenerateTestFile()` instead be `regenerateTestFile()(implict ec: ExecutionContext)`

benthecarman (commented):
Should some of this be moved to `testkit`?

dlc-test/src/test/scala/org/bitcoins/dlc/DLCTestVectorTest.scala
nkohen: `DLCTestVectorGen` is not in testkit or test scope so that's not the best place for it. Eventually I do plan on moving it to testkit though, but not until I have more types of `DLCTestVector` and have hashed out how I'm going to distinguish them and such

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVectorGen.scala
nkohen: I want this the be the object actually used for generating test vectors in "Main" scope (which is what I did to make the json file, in the sbt console)

nkohen: tbh in the future this will likely have a Main branch out of it

benthecarman: Oh okay, makes sense

benthecarman (approved):

PR #1997, opened 2020-09-10
Sort DLC funding keys lexicographically

benthecarman:
To comply with https://github.com/discreetlogcontracts/dlcspecs/pull/75


nkohen (approved):
ACK assuming CI passes

PR #1969, opened 2020-09-03
Resolve issues with reorgs and syncing filters

Christewart:
This pulls over test cases from #1962 

fixes #1965 

Basically it tests `ChainHandler.nextBlockHeaderRange()` in the case of reorgs. 

EDIT: 

So here is what is all done in this PR

1. Removes the insertion of the genesis filter header and the genesis filter in the test fixtures
2. Due to this, we create a new test fixture called `withChainHandlerGenesisFilter()` for uses where we want the gensis filter header and filter inserted
3. Makes all of our syncing code work with syncing the genesis filter header and filter from our data source
4. Adds unit tests to make sure filter header reorgs work.

PR #2003, opened 2020-09-11
Introduce 'FilterSyncMarker' to ChainApi, make it clearier what exact…

Christewart:
…ly the (Int,DoubleSha256Digest) tuple is returned from ChainApi.nextBlockHeaderRange()

This creates `FilterSyncMarker` class and adds it to the `ChainApi`. 

This would have been _really_ nice to have in #1969 , it's easy for me to confuse raw `Int` and `DoubleSha256Digest` without named fields. Now we have nice fields to remind us what the int/hash stands for.

This also moves `ChainApi` out of the package `org.bitcoins.core.api.chain.db` to `org.bitcoins.core.api.chain`

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -6,8 +6,8 @@ import akka.http.scaladsl.model.ContentTypes._
 import akka.http.scaladsl.server.ValidationRejection
 import akka.http.scaladsl.testkit.ScalatestRouteTest
 import org.bitcoins.core.Core
+import org.bitcoins.core.api.chain.ChainApi
 import org.bitcoins.core.api.wallet.{AddressInfo, CoinSelectionAlgo}
-import org.bitcoins.core.api.chain.db.ChainApi
 import org.bitcoins.core.api.wallet.db.{
   AccountDb,
   AddressDb,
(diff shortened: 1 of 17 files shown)


benthecarman (commented):
LGTM, but tbh don't love the name `FilterSyncMarker`. I think it should should at least be something like `BlockMarker` or `BlockHeightMarker` or something as this can be used for more than syncing filters.

benthecarman:
Also looks like docs are not compiling

Christewart:
> LGTM, but tbh don't love the name `FilterSyncMarker`. I think it should should at least be something like `BlockMarker` or `BlockHeightMarker` or something as this can be used for more than syncing filters.

I agree that the name sucks, what I don't about your proposals is that it loses the context that the class is useful for -- syncing filters. The two names you proposed don't retain that context and as far as I know there isn't another part of the code base that would find this class useful?

benthecarman:
> the context that the class is useful for -- syncing filters

imo you're kinda painting yourself into a corner here. I think the name should be representative of what the data it represents, not what it is currently used for.

Christewart:
> > the context that the class is useful for -- syncing filters
> 
> imo you're kinda painting yourself into a corner here. I think the name should be representative of what the data it represents, not what it is currently used for.

Yeah, this is my intention here. 

The context matters when dealing with `height` and `hash` (a height of what? a hash of what? is it the `endHeight` or the `startHeight`, is it the `prevBlockHash`, or is it a `endBlockHash`). What I'm attempting to do with this helper class is to clarify all of those questions, because I found it hard to keep it straight in my head as traversing the code base. 

The [BIP157 protocol](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#cfheaders) dictates that these are all `startHeight` and `stopHash`. 

Perhaps I should add a scaladoc ontop of `FilterSyncMarker` that indicates this?

benthecarman:
> Perhaps I should add a scaladoc ontop of FilterSyncMarker that indicates this?

Yeah that's probably better

benthecarman (approved):

PR #2005, opened 2020-09-11
Rename parameter names from 'duration' -> 'interval' just like our Ac…

Christewart:
…torSystem.schedule() names the parameter

Previously the parameter for `AsyncUtil.retryUntilSatisfiedF` had a parameter named `duration`. This was _very_ misleading. What this parameter really is a `interval` in between times to evaluate the predicate passed in as a parameter `conditionF` 

```scala
conditionF: () => Future[Boolean]
```

So `conditionF` gets called every `interval`.


benthecarman (approved):
Strong ACK from me, this always confused me

PR #2007, opened 2020-09-11
Make BlockHeader, BlockHeaderDb have better toStrings

Christewart:
This gives us prefixes on hashes in the `BlockHeader` and `BlockHeaderDb` `toString` methods. 

Now you will see things like `BlockHeader(hashBE=1234...` rather than just `BlockHeader(1234...` which doesn't give you a clear indication which hash this is in the `toString()`. Is it the prev block hash, merkle root hash, block hash etc?

This removes that ambiguity when looking at logs.


benthecarman:
Should make all the hashes `_.hex`

benthecarman (approved):

PR #2009, opened 2020-09-11
Make ChainApi.getHeadersBetween() be inclusive on the 'from' parameter

### LINKED ISSUE #1996: Make ChainApi.getHeadersBetween() return inclusive 'from' and 'to'
Currently we _exclude_ the `from` parameter for `ChainApi.getHeadersBetween()` 

This makes it hard to use this method for cases of using the genesis block, as you can't specify a parent for the genesis block I'm running up against this in #1969 when attempting to sync filters for the genesis block. 

We should change the behavior of this method to _include_ the `from` _and_ `to` headers in range of block headers returned

https://github.com/bitcoin-s/bitcoin-s/blob/47fb9e0530d07105cdd1472246b5ef612a36a6ce/core/src/main/scala/org/bitcoins/core/api/chain/db/ChainApi.scala#L147-L150

Here is where i specifically need it: 

https://github.com/bitcoin-s/bitcoin-s/blob/47fb9e0530d07105cdd1472246b5ef612a36a6ce/chain/src/main/scala/org/bitcoins/chain/blockchain/sync/FilterSync.scala#L99

Christewart:
fixes #1996

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -626,7 +626,7 @@ case class ChainHandler(
         currentF: Future[BlockHeaderDb],
         accum: Vector[BlockHeaderDb]): Future[Vector[BlockHeaderDb]] = {
       currentF.flatMap { current =>
-        if (current == from) {
+        if (current.hashBE == from.hashBE) {
           Future.successful(current +: accum)
         } else {
           val nextOptF = getHeader(current.previousBlockHashBE)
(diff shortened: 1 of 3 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -626,7 +626,7 @@ case class ChainHandler(
         currentF: Future[BlockHeaderDb],
         accum: Vector[BlockHeaderDb]): Future[Vector[BlockHeaderDb]] = {
       currentF.flatMap { current =>
-        if (current.previousBlockHashBE == from.hashBE) {
+        if (current == from) {
```
benthecarman: Should still compare hashes, it'll be faster.

Should do `current.hashBE == from.hashBE`

benthecarman (approved):
ACK on CI passing. Maybe should run a IBD with this

PR #2027, opened 2020-09-17
Add test to use llyod's info

benthecarman:
Moved some helper functions from the execution tests so we can setup and execute in other files.


nkohen (approved):
ack

PR #2029, opened 2020-09-17
Add ImmatureCoinbase TxoState

benthecarman:


wallet-test/src/test/scala/org/bitcoins/wallet/ProcessBlockTest.scala
```diff
@@ -36,4 +42,42 @@ class ProcessBlockTest extends BitcoinSWalletTest {
...
+      _ = assert(startingUtxos.isEmpty)
+      addr <- wallet.getNewAddress()
+      hashes <- bitcoind.generateToAddress(101, addr)
+      blocks <- FutureUtil.sequentially(hashes)(bitcoind.getBlockRaw)
+      _ <- FutureUtil.sequentially(blocks)(wallet.processBlock)
+      utxos <- wallet.listUtxos()
+      balance <- wallet.getBalance()
+    } yield {
+      assert(utxos.size == 101)
+      assert(balance == Bitcoins(50 * 48) + Bitcoins(25 * 53))
```
Christewart: So i don't really understand what these magic numbers are here other than the `50`, but I don't think this unit test is correct. It seems the bitcoin-core wallet doesn't change the balance until the coinbase tx is actually valid to be spent

>//starting balance
> ./bitcoin-cli -regtest getbalance
> 6088.99982620

> //get a new address
>./bitcoin-cli -regtest getnewaddress
bcrt1qhlsyvrn76s282dp07mshryy4sqsflt9a3cchtz

>//send money to that address
> ./bitcoin-cli -regtest generatetoaddress 1 bcrt1qhlsyvrn76s282dp07mshryy4sqsflt9a3cchtz
[
  "62d1c6c2c2332aa0f718187aa5dba4c570fe2dfa0ca621542125c64c2ee2d35a"
]

>//balance hasn't changed!!
>./bitcoin-cli -regtest getbalance
6088.99982620

I think this should be our behavior too? I believe ours is different?

benthecarman: The `Bitcoins(50 * 48) + Bitcoins(25 * 53)` comes from, start at block height 101, then halving occurs at block 150, so we have 48 50 btc blocks, and then 53 25 btc blocks

Christewart:
Ok. Was that meant to address my comment then too about bitcoin core's wallet balance being different?

benthecarman:
> Ok. Was that meant to address my comment then too about bitcoin core's wallet balance being different?

No, I made the these tests trying to find it and figured may as well pr them in. Still would like to find the bug

benthecarman:
Changed to add a `TxoState` for Immature Coinbases. This should filter it out of our balance as well as help debug issues with coinbase txos.

core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
```diff
@@ -14,6 +14,8 @@ object TxoState extends StringFactory[TxoState] {
   /** Means that no funds have been sent to this utxo EVER */
   final case object DoesNotExist extends TxoState
 
+  final case object ImmatureCoinbase extends TxoState
```
Christewart, line 20: This seems like  a great idea! I would add a scaladoc to it refering to what it is exactly with a link to some external documentation too

Christewart (commented):
Unless i am missing this, we are still missing a unit test that explicitly checks we aren't adding the coinbase utxo balance to the wallet until it's confirmed?

Christewart (approved):

PR #2031, opened 2020-09-18
2020 09 18 btchrp stringfactory

Christewart:
This PR does two things:

1. Makes `Bech32.checkHrpValidity()` return a string  rather than a `Bech32HumanReadablePart`
2. Makes `BtcHumanReadablePart` extend `StringFactory`

`1` is important because as we scale out to more `Bech32` applications (such as DLC) we don't want to have to modify the core code everytime. Now it's the responsibility of the caller of this code to place the returned `String` into the correct subclass of `Bech32HumanReadablePart`. 

I experienced this problem with working with a bech32 oracle address format.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
@@ -108,8 +108,8 @@ sealed abstract class Bech32 {
 
     hrpT.flatMap { chars =>
       val str = chars.mkString
-      val lnT = LnHumanReadablePart(str)
-      val btcT = BtcHumanReadablePart(str)
+      val lnT = LnHumanReadablePart.fromStringT(str)
+      val btcT = BtcHumanReadablePart.fromStringT(str)
 
       lnT
         .orElse(btcT)
--- a/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/Bech32.scala
@@ -261,6 +261,18 @@ sealed abstract class Bech32 {
     }
   }
 
+  def splitToHrpAndData[T <: Bech32HumanReadablePart](
+      bech32: String,
+      factory: StringFactory[T]): Try[(T, Vector[UInt5])] = {
+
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -75,7 +73,7 @@ sealed abstract class Bech32 {
     * [[https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32 BIP173]]
     * rules
     */
-  def checkHrpValidity(hrp: String): Try[Bech32HumanReadablePart] = {
+  def checkHrpValidity(hrp: String): Try[String] = {
```
nkohen, line 77: I personally think this would be nice to have (at least as an overloaded option)
```
def checkHrpValidity(hrp: String, factories: Vector[StringFactory[Bech32HumanReadablePart]]]): Try[Bech32HumanReadablePart]
```
as this seems like an unnecessary loss of type seeing as we have a non-sealed abstract class that should be returned in most cases.
(note that this is going to require you make `StringFactory` co-variant (i.e. `[T]` -> `[+T]`) but it should already be that and I'm shocked that this hasn't previously caused issues).

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -210,8 +202,7 @@ sealed abstract class Bech32 {
     *      [[https://github.com/sipa/bech32/blob/master/ref/python/segwit_addr.py#L62 this function]]
     *      by Sipa
     */
-  def splitToHrpAndData(
-      bech32: String): Try[(Bech32HumanReadablePart, Vector[UInt5])] = {
+  def splitToHrpAndData(bech32: String): Try[(String, Vector[UInt5])] = {
```
nkohen, line 214: ditto here

nkohen (commented):
Generally looks fine but I don't think we should get rid of the types if we don't have to

nkohen (approved):

PR #2034, opened 2020-09-18
ECDSA Adaptor Signatures for master branch

nkohen:
Brought down ecdsa adaptor signatures implemented in scala from the `dlc-crypto` branch

### DIFF
--- a/crypto-test/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
+++ b/crypto-test/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
@@ -2,7 +2,11 @@ package org.bitcoins.crypto
 
 import org.bitcoins.core.config.{MainNet, RegTest, TestNet3}
 import org.bitcoins.core.crypto.ECPrivateKeyUtil
-import org.bitcoins.testkit.core.gen.{ChainParamsGenerator, CryptoGenerators}
+import org.bitcoins.testkit.core.gen.{
+  ChainParamsGenerator,
+  CryptoGenerators,
+  NumberGenerator
+}
 import org.bitcoins.testkit.util.BitcoinSUnitTest
 
 class ECPrivateKeyTest extends BitcoinSUnitTest {
@@ -117,4 +121,20 @@ class ECPrivateKeyTest extends BitcoinSUnitTest {
     }
   }
 
+  it must "correctly execute the ecdsa single signer adaptor signature protocol" in {
+    forAll(CryptoGenerators.privateKey,
+           CryptoGenerators.privateKey,
+           NumberGenerator.bytevector(32)) {
+      case (privKey, adaptorSecret, msg) =>
+        val adaptorSig = privKey.adaptorSign(adaptorSecret.publicKey, msg)
+        assert(
+          privKey.publicKey
+            .adaptorVerify(msg, adaptorSecret.publicKey, adaptorSig))
+        val sig = adaptorSecret.completeAdaptorSignature(adaptorSig)
+        val secret =
+          adaptorSecret.publicKey.extractAdaptorSecret(adaptorSig, sig)
+        assert(secret == adaptorSecret)
+        assert(privKey.publicKey.verify(msg, sig))
+    }
+  }
 }
--- a/crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
@@ -37,6 +37,11 @@ object BouncyCastleUtil {
       .getOrElse(false)
   }
 
+  def pubKeyTweakMul(pubKey: ECPublicKey, tweak: FieldElement): ECPublicKey = {
+    val tweakedPoint = pubKey.toPoint.multiply(tweak.toBigInteger)
+    ECPublicKey.fromPoint(tweakedPoint, pubKey.isCompressed)
+  }
+
   def decompressPublicKey(publicKey: ECPublicKey): ECPublicKey = {
     if (publicKey.isCompressed) {
       val point = decodePoint(publicKey.bytes)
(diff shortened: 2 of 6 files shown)


crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -215,3 +220,224 @@ object BouncyCastleUtil {
     }
   }
 }
+
+object AdaptorStuff {
```
Christewart, line 224: Can we give it a better name? `AdaptorUtil`?

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -215,3 +220,224 @@ object BouncyCastleUtil {
...
+
+  def adaptorSign(
+      privateKey: ECPrivateKey,
+      adaptorPoint: ECPublicKey,
+      dataToSign: ByteVector): ECAdaptorSignature = {
+    // Include dataToSign and adaptor in nonce derivation
+    val hash = CryptoUtil.sha256(dataToSign ++ serializePoint(adaptorPoint))
+    val k = DLEQStuff.dleqNonceFunc(hash.bytes,
+                                    privateKey.fieldElement,
+                                    "ECDSAAdaptorNon")
```
Christewart, line 249: Is this a special tag you added? If so should this be documented or at least factored out into a constant?

crypto/src/main/scala/org/bitcoins/crypto/ECAdaptorSignature.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  require(ECPublicKey.isFullyValidWithBouncyCastle(tweakedNonce.bytes),
+          s"Tweaked nonce (R) must be a valid public key: $tweakedNonce")
+  require(ECPublicKey.isFullyValidWithBouncyCastle(untweakedNonce.bytes),
+          s"Untweaked nonce (R') must be a valid public key: $untweakedNonce")
+}
+
+object ECAdaptorSignature extends Factory[ECAdaptorSignature] {
+
+  def fromBytes(bytes: ByteVector): ECAdaptorSignature = {
```
Christewart, line 31: nit `override`

Christewart (approved):
Basically just nits about names and code structure (moving things out of single files). But that can be fixed in later PRs

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: Jonas made this

nkohen: (it'll be in Lloyd's spec)

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -215,3 +220,224 @@ object BouncyCastleUtil {
...
+      hash: ByteVector,
+      fe: FieldElement,
+      algoName: String): FieldElement = {
+    val kBytes =
+      CryptoUtil.taggedSha256(fe.bytes ++ hash, algoName).bytes
+    FieldElement(kBytes)
+  }
+
+  def dleqChallengeHash(
+      algoName: String,
```
benthecarman, line 369: Maybe make these an ADT?

crypto-test/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
```diff
@@ -117,4 +121,20 @@ class ECPrivateKeyTest extends BitcoinSUnitTest {
...
+      case (privKey, adaptorSecret, msg) =>
+        val adaptorSig = privKey.adaptorSign(adaptorSecret.publicKey, msg)
+        assert(
+          privKey.publicKey
+            .adaptorVerify(msg, adaptorSecret.publicKey, adaptorSig))
+        val sig = adaptorSecret.completeAdaptorSignature(adaptorSig)
+        val secret =
+          adaptorSecret.publicKey.extractAdaptorSecret(adaptorSig, sig)
+        assert(secret == adaptorSecret)
+        assert(privKey.publicKey.verify(msg, sig))
```
benthecarman, line 137: Maybe add a check that `sig == privKey.sign(msg)`

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -215,3 +220,224 @@ object BouncyCastleUtil {
...
+
+  /** Proves that the DLOG_G(R') = DLOG_Y(R) (= fe)
+    * For a full description, see https://cs.nyu.edu/courses/spring07/G22.3220-001/lec3.pdf
+    */
+  def dleqProve(
+      fe: FieldElement,
+      adaptorPoint: ECPublicKey,
+      algoName: String): (FieldElement, FieldElement) = {
+    // (fe*G, fe*Y)
+    val (p1, p2) = dleqPair(fe, adaptorPoint)
```
benthecarman, line 391: nit: name these instead `feG, feY`

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -215,3 +220,224 @@ object BouncyCastleUtil {
...
+      s: FieldElement,
+      e: FieldElement,
+      p1: ECPublicKey,
+      adaptor: ECPublicKey,
+      p2: ECPublicKey): Boolean = {
+    val r1 = p1.tweakMultiply(e.negate).add(s.getPublicKey)
+    val r2 = p2.tweakMultiply(e.negate).add(adaptor.tweakMultiply(s))
+    val challengeHash = dleqChallengeHash(algoName, adaptor, r1, r2, p1, p2)
+
+    challengeHash == e.bytes
```
benthecarman, line 441: Do we need checks here that nothing is 0/point of infinity

benthecarman:
With this can we get rid of the `dlc-crypto` and `schnorr-dlc` branches?

benthecarman:
Should this update secp binaries?

Christewart:
I think the idea was to get this into master so that @rorp has this stuff available to him for PTLC experiments on bitcoin-s/eclair. Since we haven't resolved our secp binaries problem in eclair, we have to use all bouncy castle.

crypto-test/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
nkohen: I don't think this will be true due to nonce selection being different

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: I believe these will be the names used in the spec (as it is what the C impl does) but I could be wrong

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: no points here, just a hash and a number (points are hashed but that should be covered by comparing to number)

nkohen:
I think generally speaking we are keeping secp binaries out of master and on `dlc-crypto` until they become stable (and just have bouncy until then)

crypto/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: That would restrict library user use cases as they wouldn't be able to extend this

PR #2036, opened 2020-09-20
Add Schnorr Signing to Sign interface

benthecarman:


crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -51,6 +51,11 @@ trait Sign {
       entropy: ByteVector): Future[ECDigitalSignature] =
     signWithEntropyFunction(bytes, entropy)
 
+  def schnorrSignFunction: ByteVector => Future[SchnorrDigitalSignature]
```
Christewart, line 54: This really feels like it should be some sort of ADT as a subclass of `Sign`. Then we can have a `ECDSASign` and a `SchnorrSign`. Thoughts?

benthecarman: NACK from me. There shouldn't be anything that can sign for ecdsa and not schnorr or vice versa.

Christewart:
Well of course that's true, the more important question is "What does the
user of the library want us to sign with?" We should allow them to specify
which one with an ADT. Some contexts it is apparent (for instance p2pkh
scripts vs taproot scripts) other contexts its not at all apparent
(signmesage for instance). Its always best in these cases to let users
specify so we as library developers don't have to guess.

nkohen:
I do worry about old hardware wallets

PR #2039, opened 2020-09-21
DLCFundingSignatures now accepts P2WSH, P2SH(Segwit)

nkohen:
Added support for counter-parties to use non-P2WPKH funding inputs.

Not in this PR because they come together in the next: tests using non-P2WPKH, correct fee computation


benthecarman (approved):
Should have tests but if they are coming in the next pr I guess it's okay with p2wpkh tests passing currently

PR #2040, opened 2020-09-21
Fixed PSBT.verifyFinalizedInput

nkohen:
If an input map has a NonWitnessUTXO it should be able to find the witness spk (as NonWitnessUTXO is meant to be used for witness UTXOs because of the BIP 143 vuln)


nkohen:
The unit test added fails on master

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -565,8 +565,21 @@ case class PSBT(
...
     val utxoOpt = inputMap.nonWitnessOrUnknownUTXOOpt
+    val wUtxoOpt = inputMap.witnessUTXOOpt match {
+      case Some(wutxo) => Some(wutxo)
+      case None =>
+        utxoOpt match {
+          case Some(utxo) =>
+            val output = utxo.transactionSpent.outputs(
+              transaction.inputs(index).previousOutput.vout.toInt)
+            output.scriptPubKey match {
+              case _: NonWitnessScriptPubKey => None
```
benthecarman: What about for p2sh(segwit)

benthecarman (approved):

PR #2041, opened 2020-09-21
Only update txo confirmed state if there is a change

benthecarman:
Previously, `updateUtxoConfirmedStates` would return every `SpendingInfoDb` that it was given with the updated ones changed. This made its logs incorrect as well as caused some side effects where we would update or not update things correctly.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
@@ -90,8 +90,8 @@ private[wallet] trait UtxoHandling extends WalletLogger {
   }
 
   protected def updateUtxoConfirmedState(
-      txo: SpendingInfoDb): Future[SpendingInfoDb] = {
-    updateUtxoConfirmedStates(Vector(txo)).map(_.head)
+      txo: SpendingInfoDb): Future[Option[SpendingInfoDb]] = {
+    updateUtxoConfirmedStates(Vector(txo)).map(_.headOption)
   }
 
   protected def updateUtxoConfirmedStates(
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -106,45 +106,47 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
               s"Given txos exist in block (${blockHash.hex}) that we do not have! $txos")
-            txos
+            Vector.empty
           case Some(confs) =>
-            txos.map { txo =>
+            txos.flatMap { txo =>
               txo.state match {
                 case TxoState.ImmatureCoinbase =>
                   if (confs > Consensus.coinbaseMaturity) {
                     if (confs >= walletConfig.requiredConfirmations) {
```
Christewart, line 115: So with coinbase txs we need 100 confirmations + the require confirmations? I guess htis does seem right otherwise we can have a reorg happen that causes the coinbase utxo be pending again. I hadn't though of this corner case before :/

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -106,45 +106,47 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
+                      Some(txo.copyWithState(TxoState.ConfirmedReceived))
                     } else {
-                      txo.copyWithState(TxoState.PendingConfirmationsReceived)
+                      Some(
+                        txo.copyWithState(
+                          TxoState.PendingConfirmationsReceived))
                     }
                   } else {
-                    txo
+                    None
```
Christewart, line 123: Don't  we need some sort of logic here to handle the reorg case? I.e. if we had pending confirmations, and then a reorg happened, we need to go back to immature coinbase?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -321,7 +323,10 @@ private[wallet] trait UtxoHandling extends WalletLogger {
         spendingInfoDAO.updateAllSpendingInfoDb(mempoolUtxos)
       // update the confirmed ones
       updatedBlockUtxos <- updateUtxoConfirmedStates(utxosInBlocks)
-      updated = updatedMempoolUtxos ++ updatedBlockUtxos
+      pendingConf = utxosInBlocks.filterNot(utxo =>
```
Christewart, line 330: could you add a coment here of what exactly is going on? I'm not sure ifI understand. Maybe a better named for `updatedBlockUtxos` would be `updatedConfirmedUtxos`?

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: This is just incase someone sets `walletConfig.requiredConfirmations` greater than 100 `Consensus.coinbaseMaturity`

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: A shorter chain won't cause reorg so it's not possible for a reorg to bring us back to immature. However, it is possible that our block is reorged and our immature coinbase will no longer be part of the chain. We should be safe from this unless there is over a 101 block reorg because if our coinbase was reorged at say 3 confs it would just stay stuck as `ImmatureCoinbase`

Christewart (approved):

PR #2042, opened 2020-09-21
Implemented generalized DLC fee computation

nkohen:
Also implemented LnMessage wrapper around TLVs.


core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
-    val filterDust = FilterDustFinalizer
-
-    addFutureFee
-      .andThen(subtractFee)
-      .andThen(filterDust)
-      .buildTx(txBuilderResult)
+  private val commonFutureFee = if (sharedFutureFee % 2 == 0) {
+    Satoshis(sharedFutureFee / 2)
+  } else {
+    Satoshis(sharedFutureFee / 2 + 1)
```
benthecarman: I think for these you can just do `Satoshis(Math.ceil(sharedFutureFee / 2.0))`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -401,6 +400,24 @@ case class AddFutureFeeFinalizer(
   }
 }
 
+case class SubtractFromOutputFinalizer(spk: ScriptPubKey, subAmt: CurrencyUnit)
```
benthecarman, line 403: With this can we remove `SubtractFeeFromOutputsFinalizer`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
+    offerInputs: Vector[DualFundingInput],
+    offerPayoutSPK: ScriptPubKey,
+    offerChangeSPK: ScriptPubKey,
+    acceptInputs: Vector[DualFundingInput],
+    acceptPayoutSPK: ScriptPubKey,
+    acceptChangeSPK: ScriptPubKey,
+    feeRate: SatoshisPerVirtualByte,
+    fundingSPK: ScriptPubKey)
     extends RawTxFinalizer {
+  private val sharedFutureFee = (feeRate * 125).satoshis.toLong
```
benthecarman: can this `125` be a constant, it is not clear what it is for

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
-  * from the change outputs. This finalizer filters dust outputs.
-  *
-  * Note: spendingVBytes may be 0
-  */
-case class P2WPKHDualFundingTxFinalizer(
-    spendingVBytes: Long,
-    feeRate: FeeUnit,
-    fundingSPK: ScriptPubKey)
-    extends RawTxFinalizer {
+  private val sharedFundingFee = (feeRate * 53).satoshis.toLong
```
benthecarman: can this `53` be a constant, it is not clear what it is for

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
+
+  private val commonFundingFee = if (sharedFundingFee % 2 == 0) {
+    Satoshis(sharedFundingFee / 2)
+  } else {
+    Satoshis(sharedFundingFee / 2 + 1)
+  }
```
benthecarman: ditto Math.ceil comment

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
+  private def computeFees(
+      inputs: Vector[DualFundingInput],
+      payoutSPK: ScriptPubKey,
+      changeSPK: ScriptPubKey): (CurrencyUnit, CurrencyUnit) = {
+    val futureFee = feeRate * payoutSPK.asmBytes.length + commonFutureFee
+
+    val inputWeight =
+      inputs.foldLeft(1L) { // Start at 1 to cover witness header
+        case (weight, DualFundingInput(scriptSignature, maxWitnessLen)) =>
+          weight + 4 * (41 + scriptSignature.asm.length) + maxWitnessLen.toInt
```
benthecarman: can this 41 get a constant or a comment

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -462,99 +479,85 @@ object SubtractFeeFromOutputsFinalizer {
...
+      }
+    val outputWeight = 4 * (9 + changeSPK.asm.length)
```
benthecarman: same with this 9

benthecarman (changes requested):
Imo this should have tests before merging

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: Technically yes but I actually think it could be useful in the future so I'm not planning to

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: Once I've updated the spec I'll be commenting with links to it

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: ditto

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: ditto

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: ditto

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestVector.scala
```diff
@@ -109,15 +111,18 @@ object ValidTestInputs {
   }
 }
 
-// TODO: Add realOutcome, oracleSignature, offerSignedCET, acceptSignedCET fields
+case class DLCTransactions(
```
benthecarman, line 120: Why not use `SetupDLC` here?

nkohen: a) The extra adaptor signatures are redundant
b) simpler
c) wanted something I could freely change within the test vector code that wasn't mingled with other external stuff, I don't view this as particularly stable but if in the future it truly doesn't change I'd be open to switching to something like `SetupDLC` and using outcomes and such. Though I will note that part of what's unstable is I haven't generated negative test vectors yet and those may end up changing this to contain options and such

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
```diff
@@ -208,11 +208,88 @@ class DLCClientTest extends BitcoinSAsyncTest {
...
+      val closingOutput = closingTx.outputs.map(_.value).sum
+      val actualClosingFee = closingInput - closingOutput
+
+      def feeRateBounds(
+          tx: Transaction,
+          actualFee: CurrencyUnit,
+          missingOutputBytes: Long = 0): (Double, Double) = {
+        val vbytesLower = Math.ceil(tx.weight / 4.0) + missingOutputBytes
+        val vbytesUpper =
+          vbytesLower + 2 // Have to add 2 to account for two parties rounding up
```
benthecarman: Maybe in the spec we should say, only offerer or only acceptor rounds up

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
```diff
@@ -208,11 +208,88 @@ class DLCClientTest extends BitcoinSAsyncTest {
...
+
+        // Upper VBytes => Lower fee rate
+        val lowerBound = actualFee.satoshis.toLong / vbytesUpper
+        // Lower VBytes => Upper fee rate
+        val upperBound = actualFee.satoshis.toLong / vbytesLower
+        (lowerBound, upperBound)
```
benthecarman: what about rounding here

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
nkohen: eh
1) it is more complicated than just adding the vbyte, currently the spec has a nice almost simple computation and I don't want to add if/elses to it :P
2) this means one party pays some percentage of a vbyte more than the other on shared fields

nkohen: plus, re 2: both parties should be willing to go up so might as well

dlc-test/src/test/scala/org/bitcoins/dlc/DLCClientTest.scala
nkohen: not sure what you mean?

nkohen: (both returned values are Double's btw so shouldn't have a rounding issue?)

benthecarman: You're doing division here with 2 longs, so there will be loss of precision. Maybe the `upperBound` should be in a `Math.ceil`?

nkohen: the denominators are Doubles not Longs (Math.ceil returns a double)

benthecarman (approved):

benthecarman (approved):

benthecarman (approved):

PR #2048, opened 2020-09-22
Change requesting filters log to use big endian hashes

benthecarman:
Should make copy pasting from here easier

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
@@ -198,7 +198,7 @@ case class PeerMessageSender(client: P2PClient)(implicit conf: NodeAppConfig)
       res <- filterSyncMarkerOpt match {
         case Some(filterSyncMarker) =>
           logger.info(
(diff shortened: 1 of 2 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
```diff
@@ -198,7 +198,7 @@ case class PeerMessageSender(client: P2PClient)(implicit conf: NodeAppConfig)
       res <- filterSyncMarkerOpt match {
         case Some(filterSyncMarker) =>
           logger.info(
-            s"Requesting compact filters from=${filterSyncMarker.startHeight} to=${filterSyncMarker.stopBlockHash}")
+            s"Requesting compact filters from=${filterSyncMarker.startHeight} to=${filterSyncMarker.stopBlockHash.flip.hex}")
```
Christewart: Maybe we should just re-work the `FilterSyncMarker.toString`? and get rid of `filterSyncMarker.startheight` and `filterSyncMakrer.stopBlockHash` here and in the log below?

Christewart (changes requested):

Christewart (approved):

PR #2053, opened 2020-09-23
Add missing settings for native packager 

benthecarman:
https://sbt-native-packager.readthedocs.io/en/latest/formats/debian.html#required-settings

### DIFF
--- a/app/server/server.sbt
+++ b/app/server/server.sbt
@@ -10,7 +10,6 @@ mainClass := Some("org.bitcoins.server.Main")
 
 graalVMNativeImageOptions ++= Seq(
   "-H:EnableURLProtocols=http",
-
   "-H:+ReportExceptionStackTraces",
   // builds a stand-alone image or reports a failure
   "--no-fallback",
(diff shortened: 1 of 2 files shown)


Christewart (approved):
Nit: This isn't for native image packaging, just using the native packager. I can see how that is confusing though :/ :/ :/

app/server/server.sbt
```diff
@@ -6,6 +6,8 @@ Compile / fork := true
 
 libraryDependencies ++= Deps.server(scalaVersion.value)
 
+version := "0.4.0"
```
Christewart: Can you check that this doesn't interfere with `sbt-cli-release`, i believe there is a requirement that the version number isn't set for our binaries to automatically publish to sonatype? https://github.com/olafurpg/sbt-ci-release

benthecarman: Yep, had to remove this. I guess when doing the windows binary we have to temporarily set this. It might automatically do it correctly when we have it at the v0.4.0 tag. I guess we'll see.

benthecarman: Looks like it will set it properly when we are on the tag

https://github.com/dwijnand/sbt-dynver#detail

PR #2060, opened 2020-09-25
Create initial BitcoindV20RpcClient

benthecarman:
Part of #1529 

Initial creating of necessary types and getting old tests passing. Following PRs will add support for changed RPC command


nkohen (approved):
ACK

PR #2062, opened 2020-09-26
Let RawTxSigner dummy sign transactions

benthecarman:
Needed for #2051 

This lets us build dummy transactions and sign transactions with dummy signatures

PR #2063, opened 2020-09-26
Add ability to fully spend utxos

### LINKED ISSUE #2051: Add spend utxos function to Wallet
#2049 shows that it is hard to correctly spend an entire utxo or to empty an entire wallet. It would be beneficial if you could spend an entire utxo or set of utxos and just have fees subtracted from the resulting output.

We could make a function that takes in a set of outpoints and sends their entire balance minus fees to a given address.

benthecarman:
Built on top of #2062 

Closes #2051 

Needed to create `LowRDummyECDigitialSignature` so we can accurately predict tx/signature sizes when low r signing

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/WalletSendingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/WalletSendingTest.scala
@@ -6,11 +6,12 @@ import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.transaction.TransactionOutput
 import org.bitcoins.core.script.constant.{BytesToPushOntoStack, ScriptConstant}
 import org.bitcoins.core.script.control.OP_RETURN
-import org.bitcoins.core.wallet.fee.SatoshisPerByte
+import org.bitcoins.core.wallet.fee._
 import org.bitcoins.core.wallet.utxo.TxoState
 import org.bitcoins.crypto.CryptoUtil
 import org.bitcoins.testkit.core.gen.{CurrencyUnitGenerator, FeeUnitGen}
 import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest.RandomFeeProvider
 import org.bitcoins.testkit.wallet.FundWalletUtil.FundedWallet
 import org.scalatest.{Assertion, FutureOutcome}
 import scodec.bits.ByteVector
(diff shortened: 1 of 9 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -389,6 +390,58 @@ abstract class Wallet
...
+      _ = require(diff.isEmpty,
+                  s"Not all OutPoints belong to this wallet, diff $diff")
+      spentUtxos =
+        utxoDbs.filterNot(utxo => TxoState.receivedStates.contains(utxo.state))
+      _ = require(
+        spentUtxos.isEmpty,
+        s"Some out points given have already been spent, ${spentUtxos.map(_.outPoint)}")
+
+      prevTxFs = utxoDbs.map(utxo =>
+        transactionDAO.findByOutPoint(utxo.outPoint).map(_.get.transaction))
```
Christewart: Why not just return the tuple here so we don't have to worry about zipping incorrectly below? 

I.e. 

```scala
utxoTxTuple <- Future.sequence { 
    utxoDbs.map(utxo =>
        transactionDAO.findByOutPoint(utxo.outPoint)
        .map(txDb => (utxo, txDb.get.transaction))
}
```

wallet-test/src/test/scala/org/bitcoins/wallet/WalletSendingTest.scala
```diff
@@ -210,6 +211,51 @@ class WalletSendingTest extends BitcoinSWalletTest {
...
+
+      // +- range in case of rounding or unexpected signature sizes
+      assert(
+        actualFeeRate.toLong === expectedFeeRate.toLong +- range,
+        s"Expected fee rate: $expectedFeeRate, inputs: $numInputs input: $inputAmount, tx bytes: ${tx.byteSize} " +
+          s"vsize: ${tx.vsize} weight ${tx.weight}"
+      )
+
+      assert(tx.outputs.size == 1)
+      assert(tx.outputs.head.scriptPubKey == testAddress.scriptPubKey)
```
Christewart, line 255: Probably need an assertion on the amount in the output too?

Christewart (approved):

wallet-test/src/test/scala/org/bitcoins/wallet/WalletSendingTest.scala
benthecarman: It is by the fee rate assertion

Christewart (approved):

PR #2064, opened 2020-09-27
Create DLC Oracle module

benthecarman:


benthecarman:
Do not merge, lots have changes coming soon :)

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
```diff
@@ -0,0 +1,243 @@
...
+import org.bitcoins.core.crypto.ExtKeyVersion.SegWitMainNetPriv
+import org.bitcoins.core.crypto.{BIP39Seed, ExtPrivateKey}
+import org.bitcoins.crypto._
+import org.bitcoins.keymanager.{DecryptedMnemonic, EncryptedMnemonic}
+import org.slf4j.LoggerFactory
+import scodec.bits.ByteVector
+
+import scala.util.{Failure, Success, Try}
+
+object SeedStorage {
```
Christewart: Is there a reason we don't just depend on the key manager project and use that code? I think key manager is zero dep

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/EventDAO.scala
```diff
@@ -0,0 +1,71 @@
+package org.bitcoins.dlc.oracle.storage
+
+import org.bitcoins.crypto.{FieldElement, SchnorrNonce}
+import org.bitcoins.db.{AppConfig, CRUD, DbCommonsColumnMappers, SlickUtil}
+import org.bitcoins.dlc.oracle.SigningVersion
+import slick.lifted.{ForeignKeyQuery, ProvenShape}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class EventDAO()(implicit
```
Christewart, line 10: So in my opinion we should have a unique DAO for each `SigningVersion` as we might not know what the requirements are for signing? This reminds me of the situation we have with spending info table (what we need to sign a segwit v1 utxo is very different than a segwit v0 output) -- should we apply this here as well?

Christewart (commented):
Also don't forget to add the tests to the Ci matrix

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/EventDAO.scala
benthecarman: We shouldn't need a DAO for each `SigningVersion`. Since the `EventDAO` doesn't contain the outcomes instead the `EventOutcomeDAO` does we can easily add on more fields if needed without effecting older outcomes. If we really needed to we could make a `EventOutcomeV2DAO` but I think how it stands currently is forwards compatible enough.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
benthecarman: The reason I did this is so I can get direct access to the `ExtPrivateKey`. Using whats in the km project, we don't have direct access to get `ExtPrivateKey` and thus can't get the keys on the exact paths I want.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/RValueDb.scala
```diff
@@ -0,0 +1,35 @@
...
+
+case class RValueDb(
+    nonce: SchnorrNonce,
+    label: String,
+    purpose: HDPurpose,
+    accountCoin: HDCoinType,
+    accountIndex: Int,
+    chainType: Int,
+    keyIndex: Int) {
+
```
Christewart, line 15: Shouldn't there be an invariant here saying that it `keyIndex` is hardened?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
Christewart: Hmm... ok this is a bit worrying for reasons discussed this morning (https://github.com/discreetlogcontracts/dlcspecs/issues/90). I think one thing that would make me feel better about people not foot gunning this and just using `ExtPrivateKey` directly and deriving a bunch of non hardened keys would be creating a new type -- let's say `ExtKeyHardened` that allows ONLY hardened derivation. 

...also is it necessarily to copy paste everything over? We should probably at least attempt to re-use some code so we don't have to maintain two different things that do more or less the exact same thing. Also it's very sensitive code, so the less copy-paste the better I would feel about this. Then when we make chagnes to the key manager project this release we won't forget about this project.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
```diff
@@ -0,0 +1,243 @@
...
+      seedPath: Path,
+      passphrase: AesPassword,
+      bip39PasswordOpt: Option[String]): ExtPrivateKey = {
+    val mnemonic = decryptMnemonicFromDisk(seedPath, passphrase).mnemonicCode
+    val seed = bip39PasswordOpt match {
+      case Some(pw) =>
+        BIP39Seed.fromMnemonic(mnemonic = mnemonic, password = pw)
+      case None =>
+        BIP39Seed.fromMnemonic(mnemonic = mnemonic,
+                               password = BIP39Seed.EMPTY_PASSWORD)
```
Christewart: Should be a parameter

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    seed.toExtPrivateKey(SegWitMainNetPriv)
```
Christewart: ExtKeyPrivVersion should be a parameter to `getPrivateKeyFromDisk()` or readable from the appConfig

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
```diff
@@ -0,0 +1,243 @@
...
+        }
+      }
+
+    decryptedEither match {
+      case Left(err)    => sys.error(err.toString)
+      case Right(value) => value
+    }
+  }
+
+  def getPrivateKeyFromDisk(
```
Christewart: is there any way we can limit scope on this ? Although I guess we have the same problem in the key manager project currently with `WalletStorage.decryptMnemonicFromDisk()`. In fact, can't you just re-use the code since we expose that as a public api currently?

Christewart (changes requested):

Christewart (commented):
Also i get this is re-using a bunch of code that already is well tested but the code is very sensitive so it would be good to have some tests similuting the entire flow from creating an event to publishing it. We shoudl also add an issue to document this on the website (doesn't have to be done on this PR)

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
benthecarman: It is, it's an option, and this is what is used if it is given as `None`

Christewart: I'm blind :man_facepalming:

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/SeedStorage.scala
benthecarman: I like this idea of a `ExtKeyHardened`, however, I think I will do that in a separate PR onto `master`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/storage/RValueDb.scala
benthecarman: It is always hardened here

Christewart: Then why not add the invariant? They we know for sure it's never hardened and can't be misused by someone accidentally in the future

benthecarman: There is no invariant to add, `Int` is neither hardened nor unhardened. We can only tell with a `UInt32`

Christewart: Is it because `keyIndex` is relative to the `ExtKey.hardenedIdx` ?

benthecarman:
```
val hardenedIdx = UInt32(NumberUtil.pow2(31).toLong)

case class BIP32Node(index: Int, hardened: Boolean) {
  require(index >= 0, s"BIP32 node index must be positive! Got $index")

  /**
    * Converts this node to a BIP32 notation
    * unsigned 32 bit integer
    */
  def toUInt32: UInt32 =
    if (hardened) ExtKey.hardenedIdx + UInt32(index.toLong)
    else UInt32(index)
}
```

Christewart (approved):
So i'm cool with merging with the understanding that we file an issue for 

1. `ExtKeyHardened` and using that here
2. Providing a safe API for generating rValue

Issue #1918, opened 2020-08-27
Wallet via bitcoind rpc rather than bitcoin-s p2p


Christewart:
Talking with Nicolas Dorier this morning he said it would be _very_ nice to have way to use a DLC wallet with just a normal bitcoind rpc client _and_ a _pruned_ bitcoind rather than p2p neutrino stuff. We should investigate and comment here with what it will take to get this done

NicolasDorier:
The way we deal with rescan with BTCPay Server is the following:

Our UTXO indexer, NBXPlorer only starts synching from the time the node is fully synched.
It does not attempt to scan past block, it only starts from where the node is.

If a user wants to rescan, we do it as a one time operation via scantxoutset

PR #2076, opened 2020-09-29
Create Util functions for wallets with a bitcoind backend

benthecarman:
Just creates some preliminary functions. Nothing hooked up yet.

Part of #1918

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -1,51 +1,41 @@
 package org.bitcoins.server
 
-import org.bitcoins.core.api.chain.ChainQueryApi
-import org.bitcoins.core.api.chain.ChainQueryApi.FilterResponse
 import org.bitcoins.core.api.node.NodeApi
-import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.core.protocol.transaction.Transaction
 import org.bitcoins.core.util.{BitcoinSLogger, FutureUtil}
-import org.bitcoins.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.crypto.DoubleSha256Digest
 import org.bitcoins.rpc.client.common.BitcoindRpcClient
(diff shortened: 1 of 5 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -0,0 +1,99 @@
...
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.core.util.{BitcoinSLogger, FutureUtil}
+import org.bitcoins.crypto.{DoubleSha256Digest, DoubleSha256DigestBE}
+import org.bitcoins.rpc.client.common.BitcoindRpcClient
+import org.bitcoins.wallet.Wallet
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Useful utilities to use in the wallet project for syncing things against bitcoind */
+object BitcoindRpcBackendUtil extends BitcoinSLogger {
```
Christewart, line 13: Can we just move this into the `bitcoindRpc` project so we can share this between here and https://github.com/bitcoin-s/bitcoin-s/blob/fa388c744847081788a98352c304a377e815904c/testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala#L51

app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -322,6 +323,32 @@ object Main extends App with HttpLogger {
       start
     }
   }
+
+  def createWalletWithBitcoindCallbacks(
```
Christewart: nit: Maybe move this out into a separate file and possibly add a test case for it? Maybe this could go into your `BitcoindRpcBackendUtil`

Christewart (changes requested):

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: Then the bitcoind rpc project will need to depend on wallet

Christewart (approved):

Christewart (approved):

PR #2077, opened 2020-09-29
Create BitcoindRpcAppConfig

### LINKED ISSUE #2018: Make bitcoin-s.bitcoind-rpc configuration
We should be able to configure a remote connection to a bitcoind node. Something that looks like 

```
bitcoin-s { 
  bitcoind-rpc { 
     rpcuser=...
     rpcpassword = ...
  }
}
```

I believe we want to use [`rpcconnect`](https://github.com/bitcoin/bitcoin/blob/06dbbe76dd027e924eb95ffc4b409daffaefe61b/share/examples/bitcoin.conf#L121)? to allow for remote connections to a node? 

In general, we should probably conform to the [`bitcoin.conf`](https://github.com/bitcoin/bitcoin/blob/master/share/examples/bitcoin.conf) key names where possible.

We should probably advise that users use ssh tunneling to securely connect to their remote bitcoind instance too?

benthecarman:
Closes #2018

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
@@ -1,16 +1,15 @@
 package org.bitcoins.server
 
-import com.typesafe.config.Config
-import org.bitcoins.wallet.config.WalletAppConfig
-import org.bitcoins.node.config.NodeAppConfig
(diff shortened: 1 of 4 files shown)


Christewart (changes requested):
Everything LGTM except that documentation needs to be updated

Christewart (approved):

PR #2079, opened 2020-09-29
Calculate HRP from network param instead of the inverse

benthecarman:
This fixes it so when creating a `Bech32Address` given a `NetworkParameter`it will be correctly set for the internal structure. Previously, if we created an address with `SigNet` as the network the  hrp would be `tb` so its network would then be `Testnet3`. Now, it will save the `SigNet` network param and calculate the hrp to `tb` from the network param, thus making sure our `SigNet` network param is correct.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/Address.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/Address.scala
@@ -79,7 +79,7 @@ sealed abstract class P2SHAddress extends BitcoinAddress {
   */
 sealed abstract class Bech32Address extends BitcoinAddress {
 
-  val hrp: BtcHumanReadablePart = BtcHumanReadablePart(networkParameters)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -79,12 +79,10 @@ sealed abstract class P2SHAddress extends BitcoinAddress {
   */
 sealed abstract class Bech32Address extends BitcoinAddress {
 
-  def hrp: BtcHumanReadablePart
+  val hrp: BtcHumanReadablePart = BtcHumanReadablePart(networkParameters)
```
nkohen: make this lazy just for safety since this is an abstract class, initialization stuff can get weird

nkohen (approved):
lgtm with one thing

PR #2087, opened 2020-09-30
Make bitcoind extend chain api

benthecarman:
Part of #1918

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -3,10 +3,18 @@ package org.bitcoins.rpc.client.common
 import java.io.File
 
 import akka.actor.ActorSystem
-import org.bitcoins.core.api.chain.ChainQueryApi
+import org.bitcoins.core.api.chain.db.{
+  BlockHeaderDb,
+  CompactFilterDb,
+  CompactFilterHeaderDb
+}
+import org.bitcoins.core.api.chain.{ChainApi, ChainQueryApi, FilterSyncMarker}
 import org.bitcoins.core.api.feeprovider.FeeRateApi
 import org.bitcoins.core.api.node.NodeApi
+import org.bitcoins.core.gcs.FilterHeader
+import org.bitcoins.core.p2p.CompactFilterMessage
 import org.bitcoins.core.protocol.BlockStamp
+import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.protocol.transaction.Transaction
 import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.core.wallet.fee.FeeUnit
@@ -39,7 +47,7 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
     extends Client
     with FeeRateApi
     with NodeApi
-    with ChainQueryApi
+    with ChainApi
     with BlockchainRpc
     with MessageRpc
     with MempoolRpc
@@ -68,7 +76,7 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
             new RuntimeException("Unexpected error when getting fee rate"))
       }
     }
-  // Chain Query Api
+  // Chain Api
 
   /** Gets the height of the given block */
   override def getBlockHeight(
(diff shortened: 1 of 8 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -112,6 +120,82 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
...
+
+  override def getBestBlockHeader(): Future[BlockHeaderDb] =
+    for {
+      hash <- getBestBlockHash
+      header <- getBlockHeader(hash)
+    } yield header.blockHeaderDb
+
+  override def nextBlockHeaderBatchRange(
+      prevStopHash: DoubleSha256DigestBE,
+      batchSize: Int): Future[Option[FilterSyncMarker]] = FutureUtil.none
```
Christewart: Please fail explicitly in these scenarios so it doesn't result in hours of debugging in the future :-). I.e.

```scala
Future.failed(new RuntimeException(s"Bitcoind chainApi doesn't allow you fetch block header batch range"))
```

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -112,6 +120,82 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
...
+  override def processFilters(
+      message: Vector[CompactFilterMessage]): Future[ChainApi] =
+    Future.successful(this)
+
+  override def processCheckpoints(
+      checkpoints: Vector[DoubleSha256DigestBE],
+      blockHash: DoubleSha256DigestBE): Future[ChainApi] =
+    Future.successful(this)
+
+  override def getFilterHeaderCount(): Future[Int] = ???
```
Christewart, line 182: Since we just using bitcoind in this scenario, and bitcoind generates the filters, it's probably safe to assume that `filterHeaderCount == blockHeaderCount`? So you might be able to just use `getBlockCount()` behind the scenes? I would put this reasoning as a scaladoc so we don't forget in the future.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -112,6 +120,82 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
...
+
+  override def getFilterHeadersAtHeight(
```
Christewart, line 184: ditto as above

Christewart (commented):
I'm realizing that this PR probably isn't done, but left some preliminary feedback i guess.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
benthecarman: These are `???` because on versions without filter support we want to return an error instead.

benthecarman: Context for how it was done last time: https://github.com/bitcoin-s/bitcoin-s/pull/2087

Christewart: Ohhh, didn't think about filter support only being post v18 i think? Can we do something with versions to actually support this in newer versions?

benthecarman: Yeah, it's overridden in those

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v20/BitcoindV20RpcClient.scala
```diff
@@ -62,6 +63,36 @@ class BitcoindV20RpcClient(override val instance: BitcoindInstance)(implicit
...
+      filter <- getBlockFilter(hash, FilterType.Basic)
+    } yield Vector(filter.filterDb(height))
+  }
+
+  // Cannot get the filter headers from the RPC
+
+  override def getFilterHeadersAtHeight(
+      height: Int): Future[Vector[CompactFilterHeaderDb]] = FutureUtil.emptyVec
+
+  override def getBestFilterHeader(): Future[Option[CompactFilterHeaderDb]] =
```
Christewart: Doesn't bitcoind v20 support filters?

Christewart: While they might not have a direct rpc for query `getbestfiltereheader` we can synthetically implement this with a call to `getbestheader` can't we? and then retrieving that filter from bitcoind?

benthecarman: We can't get filter headers from the rpc

Christewart: Did they remove `getblockfilter` in v20? Otherwise can't you emulate

https://github.com/bitcoin-s/bitcoin-s/blob/fa388c744847081788a98352c304a377e815904c/testkit/src/main/scala/org/bitcoins/testkit/chain/SyncUtil.scala#L38-L37

Christewart: Wow I don't see it.... why did they get rid of it? :/ 

https://developer.bitcoin.org/reference/rpc/index.html#blockchain-rpcs

benthecarman: There is still `getFilter` but there is no `getFilterHeader`

Christewart: Ok I think you've convinced me it can't be done. 

So can we just error on these cases too? Silent failures lead to hours of debugging :-)

Christewart (approved):

PR #2088, opened 2020-09-30
Bitcoind backend on server start up 

### LINKED ISSUE #1918: Wallet via bitcoind rpc rather than bitcoin-s p2p

benthecarman:
Closes #1918

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala
@@ -7,6 +7,9 @@ import java.nio.file._
 import com.typesafe.config.Config
 import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.db._
+import org.bitcoins.node.NodeType
+import org.bitcoins.node.config.NodeAppConfig
+import org.bitcoins.rpc.client.common.BitcoindRpcClient
 import org.bitcoins.rpc.config._
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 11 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala
```diff
@@ -29,7 +32,16 @@ case class BitcoindRpcAppConfig(
...
 
-  override def start(): Future[Unit] = FutureUtil.unit
+  lazy val nodeConf: NodeAppConfig = NodeAppConfig(directory, confs: _*)
+
+  override def start(): Future[Unit] = {
+    nodeConf.nodeType match {
+      case NodeType.BitcoindBackend =>
+        client.start().map(_ => ())
+      case NodeType.SpvNode | NodeType.NeutrinoNode | NodeType.FullNode =>
+        FutureUtil.unit
```
Christewart, line 43: Shouldn't this just be an error?

node/src/main/scala/org/bitcoins/node/NodeType.scala
```diff
@@ -6,21 +6,29 @@ sealed abstract class NodeType {
...
     override def shortName: String = "neutrino"
   }
 
-  final case object SpvNode extends NodeType {
+  final case object SpvNode extends InternalImplementationNodeType {
     override def shortName: String = "spv"
   }
 
-  val all: Vector[NodeType] = Vector(FullNode, NeutrinoNode, SpvNode)
+  final case object BitcoindBackend extends ExternalImplementationNodeType {
```
Christewart, line 26: Hmm, i'm not sure if I agree with this being considered a `NodeType`. It's more of a source for chain information. Would you consider the block stream API another node type as well if we were to implement that?

I guess `BitcoindRpcClient` does implement `NodeApi` though so :man_shrugging:.

Christewart (approved):
Basically what I ended up leaving is nits, feel free to ignore if you want to get it merged

node/src/main/scala/org/bitcoins/node/NodeType.scala
benthecarman: I agree this is kinda bastardizing the `NodeType` but probably a better solution can be done in the future.

PR #2094, opened 2020-09-30
Add basic DLC Oracle

### LINKED ISSUE #2096: Choosing a signing key for a `DLCOracle`
Currently, the `DLCOracle` is using an `ExtPrivateKeyHardened` for its seed. This makes it safer when getting nonces as they all need to be from hardened paths to be secure.

However, since we are using a `ExtPrivateKeyHardened` it makes it so we cannot use a standard derivation path for the signing key. This is a problem is the user wants to stake funds under this signing key. This problem arises because when they then want to be able to spend these funds. Ideally the user could enter their seed into a wallet and it can handle the spending, however using a `ExtPrivateKeyHardened` will not permit that because we will no longer be using a standard derivation path.

I see there are 4 options:

- Continue using `ExtPrivateKeyHardened` and have our signing key be on a non-standard path and provide a way to export the private key.
	- Many wallets support importing a private key so it wouldn't be a huge deal to spend it later

- Stop using `ExtPrivateKeyHardened` and trust that we implement it correctly to only use hardened paths.
	- On the PR there is already checks for this https://github.com/bitcoin-s/bitcoin-s/pull/2094/files#diff-7f6f022a7b4ab34bb53e4ae817becbe9R81

- Use a separate seed for signing
	- This makes backups harder and can confuse users

- Make the `DLCOracle` able to spend natively
	- imo this is the worst option as then the oracle needs to become chain aware and require a lot more overhead.

### LINKED ISSUE #2099: EventDb doesn't store public key associated with an event
Two reasons this should be stored

1. You shouldn't have to access the private key to read information about an event. In other words there should be a "watch-only" mode here.
2. The on going debate here: https://github.com/discreetlogcontracts/dlcspecs/issues/93

benthecarman:
Closes #2096 
Closes #2099

### DIFF
--- a/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
+++ b/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
@@ -4,32 +4,9 @@ import java.sql.SQLException
 
 import org.bitcoins.core.util.TimeUtil
 import org.bitcoins.crypto._
-import org.bitcoins.testkit.BitcoinSTestAppConfig.tmpDir
-import org.bitcoins.testkit.fixtures.BitcoinSFixture
-import org.bitcoins.testkit.util.FileUtil
-import org.scalatest.FutureOutcome
+import org.bitcoins.testkit.fixtures.DLCOracleFixture
 
-import scala.concurrent.Future
-
-class DLCOracleTest extends BitcoinSFixture {
-
-  override type FixtureParam = DLCOracle
-
-  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    val builder: () => Future[DLCOracle] = () => {
-      val conf = DLCOracleAppConfig(tmpDir())
-      conf.initialize(AesPassword.fromString("bad"), None)
-    }
-
-    val destroy: DLCOracle => Future[Unit] = dlcOracle => {
-      val conf = dlcOracle.conf
-      conf.dropAll().flatMap { _ =>
-        FileUtil.deleteTmpDir(conf.baseDatadir)
-        conf.stop()
-      }
-    }
-    makeDependentFixture(builder, destroy = destroy)(test)
-  }
+class DLCOracleTest extends DLCOracleFixture {
 
   val testOutcomes: Vector[String] = (0 to 10).map(_.toString).toVector
 
@@ -75,8 +52,9 @@ class DLCOracleTest extends BitcoinSFixture {
       } yield {
         assert(rValDbOpt.isDefined)
         val rValDb = rValDbOpt.get
-        val hash = CryptoUtil.sha256(
-          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.eventName))
+        val hash = CryptoUtil.taggedSha256(
+          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.eventName),
+          "DLC/Commitment")
         assert(
           dlcOracle.publicKey.verify(hash.bytes, rValDb.commitmentSignature))
       }
@@ -125,5 +103,17 @@ class DLCOracleTest extends BitcoinSFixture {
       val dummyNonce = SchnorrNonce(ECPublicKey.freshPublicKey.bytes.tail)
       recoverToSucceededIf[RuntimeException](
         dlcOracle.signEvent(dummyNonce, "testOutcomes"))
+  }
+
+  it must "fail to sign an outcome that doesn't exist" in {
+    dlcOracle: DLCOracle =>
+      recoverToSucceededIf[RuntimeException] {
+        for {
+          eventDb <-
+            dlcOracle.createNewEvent("test", TimeUtil.now, testOutcomes)
+          signedEventDb <-
+            dlcOracle.signEvent(eventDb.nonce, "not a real outcome")
+        } yield ()
+      }
   }
 }
--- a/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
+++ b/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
@@ -70,7 +70,7 @@ class DLCOracleTest extends BitcoinSFixture {
         assert(rValDbOpt.isDefined)
         val rValDb = rValDbOpt.get
         val hash = CryptoUtil.sha256(
-          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.label))
+          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.eventName))
         assert(
           dlcOracle.publicKey.verify(hash.bytes, rValDb.commitmentSignature))
       }
--- a/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
+++ b/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
@@ -1,0 +1,121 @@
+package org.bitcoins.dlc.oracle
+
+import java.sql.SQLException
+
+import org.bitcoins.crypto._
+import org.bitcoins.testkit.BitcoinSTestAppConfig.tmpDir
+import org.bitcoins.testkit.fixtures.BitcoinSFixture
+import org.bitcoins.testkit.util.FileUtil
+import org.scalatest.FutureOutcome
+
+import scala.concurrent.Future
+
+class DLCOracleTest extends BitcoinSFixture {
+
+  override type FixtureParam = DLCOracle
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val builder: () => Future[DLCOracle] = () => {
+      val conf = DLCOracleAppConfig(tmpDir())
+      conf.initialize(AesPassword.fromString("bad"), None)
+    }
+
+    val destroy: DLCOracle => Future[Unit] = dlcOracle => {
+      val conf = dlcOracle.conf
+      conf.dropAll().flatMap { _ =>
+        FileUtil.deleteTmpDir(conf.baseDatadir)
+        conf.stop()
+      }
+    }
+    makeDependentFixture(builder, destroy = destroy)(test)
+  }
+
+  val testOutcomes: Vector[String] = (0 to 10).map(_.toString).toVector
+
+  behavior of "DLCOracle"
+
+  it must "correctly initialize" in { dlcOracle: DLCOracle =>
+    assert(dlcOracle.conf.seedExists())
+  }
+
+  it must "start with no events" in { dlcOracle: DLCOracle =>
+    dlcOracle.listEventDbs().map { events =>
+      assert(events.isEmpty)
+    }
+  }
+
+  it must "start with no pending events" in { dlcOracle: DLCOracle =>
+    dlcOracle.listPendingEventDbs().map { events =>
+      assert(events.isEmpty)
+    }
+  }
+
+  it must "create a new event and list it with pending" in {
+    dlcOracle: DLCOracle =>
+      for {
+        testEventDb <- dlcOracle.createNewEvent("test", testOutcomes)
+        pendingEvents <- dlcOracle.listPendingEventDbs()
+      } yield {
+        assert(pendingEvents.size == 1)
+        assert(pendingEvents.contains(testEventDb))
+      }
+  }
+
+  it must "create a new event with a valid commitment signature" in {
+    dlcOracle: DLCOracle =>
+      for {
+        testEventDb <- dlcOracle.createNewEvent("test", testOutcomes)
+        rValDbOpt <- dlcOracle.rValueDAO.read(testEventDb.nonce)
+      } yield {
+        assert(rValDbOpt.isDefined)
+        val rValDb = rValDbOpt.get
+        val hash = CryptoUtil.sha256(
+          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.label))
+        assert(
+          dlcOracle.publicKey.verify(hash.bytes, rValDb.commitmentSignature))
+      }
+  }
+
+  it must "create multiple events with different names" in {
+    dlcOracle: DLCOracle =>
+      for {
+        _ <- dlcOracle.createNewEvent("test", testOutcomes)
+        _ <- dlcOracle.createNewEvent("test1", testOutcomes)
+      } yield succeed
+  }
+
+  it must "fail to create multiple events with the same name" in {
+    dlcOracle: DLCOracle =>
+      recoverToSucceededIf[SQLException] {
+        for {
+          _ <- dlcOracle.createNewEvent("test", testOutcomes)
+          _ <- dlcOracle.createNewEvent("test", testOutcomes)
+        } yield ()
+      }
+  }
+
+  it must "create and sign a event" in { dlcOracle: DLCOracle =>
+    val outcome = testOutcomes.head
+    for {
+      eventDb <- dlcOracle.createNewEvent("test", testOutcomes)
+      sig <- dlcOracle.signEvent(eventDb.nonce, outcome)
+      outcomeDbs <- dlcOracle.eventOutcomeDAO.findByNonce(eventDb.nonce)
+      outcomeDb = outcomeDbs.find(_.message == outcome).get
+      signedEvent <- dlcOracle.eventDAO.read(eventDb.nonce)
+    } yield {
+      assert(signedEvent.isDefined)
+      assert(signedEvent.get.attestationOpt.contains(sig.sig))
+      assert(dlcOracle.publicKey.verify(outcomeDb.hashedMessage.bytes, sig))
+      assert(
+        SchnorrDigitalSignature(signedEvent.get.nonce,
+                                signedEvent.get.attestationOpt.get) == sig)
+    }
+  }
+
+  it must "fail to sign an event that doesn't exist" in {
+    dlcOracle: DLCOracle =>
+      val dummyNonce = SchnorrNonce(ECPublicKey.freshPublicKey.bytes.tail)
+      recoverToSucceededIf[RuntimeException](
+        dlcOracle.signEvent(dummyNonce, "testOutcomes"))
+  }
+}
--- a/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
+++ b/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
@@ -98,11 +98,13 @@ class DLCOracleTest extends BitcoinSFixture {
     val outcome = testOutcomes.head
     for {
       eventDb <- dlcOracle.createNewEvent("test", testOutcomes)
-      sig <- dlcOracle.signEvent(eventDb.nonce, outcome)
+      signedEventDb <- dlcOracle.signEvent(eventDb.nonce, outcome)
       outcomeDbs <- dlcOracle.eventOutcomeDAO.findByNonce(eventDb.nonce)
       outcomeDb = outcomeDbs.find(_.message == outcome).get
       signedEvent <- dlcOracle.eventDAO.read(eventDb.nonce)
     } yield {
+      val sig = signedEventDb.sigOpt.get
+
       assert(signedEvent.isDefined)
       assert(signedEvent.get.attestationOpt.contains(sig.sig))
       assert(dlcOracle.publicKey.verify(outcomeDb.hashedMessage.bytes, sig))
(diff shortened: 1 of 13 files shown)


dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+      val outcomesByNonce = outcomes.groupBy(_.nonce)
+      eventDbs.map(db => Event(db, outcomesByNonce(db.nonce)))
+    }
+  }
+
+  def createNewEvent(
+      label: String,
+      outcomes: Vector[String]): Future[EventDb] = {
+    for {
+      indexOpt <- rValueDAO.maxKeyIndex
```
Christewart, line 126: So I do worry about the usage of `maxKeyIndex`. There is nothing that protects against two separate threads calling `createNewEvent()` at the same time. If I understand correctly, this would lead to an r value being re-used? 

I think it would be wiser to have `rValueDAO` be responsible for retrieving _and_ incrementing the `maxKeyIndex`. We should then write test cases to make sure we can't get `rValueDAO` to return the same index on separte calls.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+      eventDbs <- eventDAO.findAll()
+      outcomes <- eventOutcomeDAO.findAll()
+    } yield {
+      val outcomesByNonce = outcomes.groupBy(_.nonce)
+      eventDbs.map(db => Event(db, outcomesByNonce(db.nonce)))
+    }
+  }
+
+  def createNewEvent(
+      label: String,
```
Christewart: I'm not sure where you want to address this, but in the future we will need to take a `pubKey` and `expiration` as part of #2099 and #2101 . I understand that you don't think it's wise to provide the abililty to rotate keys, but just also from the oracle explorer perspective it makes it _much_ easier to just store the pubkey in the database so I can just read it out of the database instead of having to touch private key material to get the public key.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+        case Some(value) => value + 1
+        case None        => 0
+      }
+
+      path = getPath(index)
+      nonce = getKValue(label, path).schnorrNonce
+
+      hash =
+        CryptoUtil.sha256(nonce.bytes ++ CryptoUtil.serializeForHash(label))
+      commitmentSig = signingKey.schnorrSign(hash.bytes)
```
Christewart, line 140: What exactly is `commitmentSig`?  I was under the assumption it was the final signature published, but it seems that it is just affirming that you actually own the private key? This seems like something that can be replaced if we put signatures on oracle addresses? https://github.com/discreetlogcontracts/dlcspecs/issues/99

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+
+      _ <- rValueDAO.create(rValueDb)
+      eventDb <- eventDAO.create(eventDb)
+      _ <- eventOutcomeDAO.createAll(eventOutcomeDbs)
+    } yield eventDb
+  }
+
+  def signEvent(
+      nonce: SchnorrNonce,
+      outcome: String): Future[SchnorrDigitalSignature] = {
```
Christewart: nit: I would maybe return the `EventDb` rather than `SchnorrDigitalSignature`. Realistically users of the API are going to probably want to know more information about the Event (and the `EventDb` should now contain `SchnorrDigitalSignature`).

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+object DLCOracle {
+
+  def apply(
+      mnemonicCode: MnemonicCode,
+      password: AesPassword,
+      bip39PasswordOpt: Option[String] = None)(implicit
+      conf: DLCOracleAppConfig): DLCOracle = {
+    val decryptedMnemonic = DecryptedMnemonic(mnemonicCode, TimeUtil.now)
+    val encrypted = decryptedMnemonic.encrypt(password)
+    WalletStorage.writeMnemonicToDisk(conf.seedPath, encrypted)
```
Christewart: Do we want to write to disk _everytime_ the DlcOracle() constructor is called?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+  private val PURPOSE = 585
+
+  private val rValueAccount: HDAccount = {
+    val coin = HDCoin(HDPurpose(PURPOSE), HDCoinType.fromNetwork(conf.network))
+    HDAccount(coin, 0)
+  }
+
+  private val chainIndex = 0
+
+  private val signingKey: ECPrivateKey = {
```
Christewart: Probably should be a `def` so `signingKey` isn't loaded into memory until it actually needs to be accessed.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,206 @@
...
+import org.bitcoins.core.protocol.Bech32Address
+import org.bitcoins.core.protocol.script.P2WPKHWitnessSPKV0
+import org.bitcoins.core.util.TimeUtil
+import org.bitcoins.crypto._
+import org.bitcoins.dlc.oracle.storage._
+import org.bitcoins.keymanager.{DecryptedMnemonic, WalletStorage}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class DLCOracle(extPrivateKey: ExtPrivateKeyHardened)(implicit
```
Christewart: `private val extPrivateKey: ExtPrivateKeyHardened`

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -0,0 +1,121 @@
...
+                                signedEvent.get.attestationOpt.get) == sig)
+    }
+  }
+
+  it must "fail to sign an event that doesn't exist" in {
+    dlcOracle: DLCOracle =>
+      val dummyNonce = SchnorrNonce(ECPublicKey.freshPublicKey.bytes.tail)
+      recoverToSucceededIf[RuntimeException](
+        dlcOracle.signEvent(dummyNonce, "testOutcomes"))
+  }
```
Christewart, line 106: So I would like to see a test case that calls `createEvent()` twice in parallel a verify that the same `nonce` is NEVER returned. In general, we should always make sure the database configuration for the DLC oracle project has `numThreads=1` to make sure that there is no concurrent database accesses. We should use a `require()` inside of `DLCOracleAppConfig` to enforce this.

Christewart (changes requested):
Also I think @nkohen should review this as it's very sensitive code.

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
Christewart: Issue filed for the latter half of this issue: https://github.com/bitcoin-s/bitcoin-s/issues/2107

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: The `commitmentSig` is the oracle signing their R value to prove that the oracle is infact going to sign for this event / R value

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: I think it's not very vulnerable to that. The db has a primary key for the nonce so it should fail if you try to make a second with the same, also the r value is tweaked based on the label so it's extra safe there.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: To be more clear they aren't signing the R value but instead 
```scala
CryptoUtil.sha256(nonce.bytes ++ CryptoUtil.serializeForHash(label))
```

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Imo I think it makes more sense for the dlc explorer to use a different db schema than a oracle implementation

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
```diff
@@ -61,8 +65,41 @@ case class DLCOracleAppConfig(
...
+      val entropy = MnemonicCode.getEntropy256Bits
+      val mnemonicCode = MnemonicCode.fromEntropy(entropy)
+      val decryptedMnemonic = DecryptedMnemonic(mnemonicCode, TimeUtil.now)
+      val encrypted = decryptedMnemonic.encrypt(password)
+      WalletStorage.writeMnemonicToDisk(seedPath, encrypted)
+    }
+
+    val key =
+      WalletStorage.getPrivateKeyFromDisk(seedPath,
+                                          SegWitMainNetPriv,
```
nkohen, line 91: is this param not used or else is there any way to ensure that mainnet is really the intended target?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
Christewart: >Imo I think it makes more sense for the dlc explorer to use a different db schema than a oracle implementation

This may be true, but it still doesn't address the problem of having to have access to the private key to serve oracle information. I think it's a bad idea to require access to the private key to give people an oracle address for instance. 

Also I think we both agree expirations are a good idea in the oracle database ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/Event.scala
```diff
@@ -0,0 +1,73 @@
...
+  def eventName: String
+
+  /** The total number of outcomes */
+  def numOutcomes: Long = outcomes.size
+
+  /** The version of signing for this event */
+  def signingVersion: SigningVersion
+
+  /** A signature by the oracle of the hash of nonce */
+  def commitmentSignature: SchnorrDigitalSignature
```
nkohen: Should probably sign the eventName with the nonce

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  // 585 is a random one I picked, unclaimed in https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+  private val PURPOSE = 585
+
+  private val rValueAccount: HDAccount = {
+    val coin = HDCoin(HDPurpose(PURPOSE), HDCoinType.fromNetwork(conf.network))
+    HDAccount(coin, 0)
+  }
+
+  private val chainIndex = 0
```
nkohen: what is the purpose of this value?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+//    val coin = HDCoin(HDPurposes.SegWit, HDCoinType.fromNetwork(conf.network))
+//    val account = HDAccount(coin, 0)
+//    val chain = HDChain(HDChainType.External, account)
+//    HDAddress(chain, 0)
+//  }
+
+  // 585 is a random one I picked, unclaimed in https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+  private val PURPOSE = 585
+
+  private val rValueAccount: HDAccount = {
```
nkohen: might not be the best name as this account also seems to contain the signing key

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  private def signingKey: ECPrivateKey = {
+    val accountIndex = rValueAccount.index
+    val coin = rValueAccount.coin
+    val purpose = coin.purpose
+    // Use a different chain index so we don't reuse a key
+    val chain = chainIndex + 1
+
+    val path = BIP32Path.fromString(
+      s"m/${purpose.constant}'/${coin.coinType.toInt}'/$accountIndex'/$chain'/0'")
```
nkohen: should call `getPath`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private val chainIndex = 0
+
+  private def signingKey: ECPrivateKey = {
+    val accountIndex = rValueAccount.index
+    val coin = rValueAccount.coin
+    val purpose = coin.purpose
+    // Use a different chain index so we don't reuse a key
+    val chain = chainIndex + 1
```
nkohen: don't follow this, it is always 1?

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -0,0 +1,123 @@
...
+import org.scalatest.FutureOutcome
+
+import scala.concurrent.Future
+
+class DLCOracleTest extends BitcoinSFixture {
+
+  override type FixtureParam = DLCOracle
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    val builder: () => Future[DLCOracle] = () => {
```
Christewart: nit: Maybe move this into testkit? It's nice to have test utilities available to you in other projects

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+  private def getKValue(rValDb: RValueDb): ECPrivateKey =
+    getKValue(rValDb.eventName, rValDb.path)
+
+  private def getKValue(label: String, path: BIP32Path): ECPrivateKey = {
+    require(path.forall(_.hardened),
+            s"Cannot use a BIP32Path with unhardened nodes, got $path")
+    val priv = extPrivateKey.deriveChildPrivKey(path).key
+    val hash =
+      CryptoUtil.sha256(
+        priv.schnorrNonce.bytes ++ CryptoUtil.serializeForHash(label))
```
nkohen: should probably be tagged

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+      index = indexOpt match {
+        case Some(value) => value + 1
+        case None        => 0
+      }
+
+      path = getPath(index)
+      nonce = getKValue(eventName, path).schnorrNonce
+
+      hash =
+        CryptoUtil.sha256(nonce.bytes ++ CryptoUtil.serializeForHash(eventName))
```
nkohen: should be tagged

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+      rValueDb = RValueDbHelper(nonce = nonce,
+                                eventName = eventName,
+                                account = rValueAccount,
+                                chainType = chainIndex,
+                                keyIndex = index,
+                                commitmentSignature = commitmentSig)
+
+      eventDb = EventDb(nonce, eventName, outcomes.size, Mock, None)
+      eventOutcomeDbs = outcomes.map { outcome =>
+        val hash = CryptoUtil.sha256(outcome)
```
nkohen: should be tagged

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  def signEvent(nonce: SchnorrNonce, outcome: String): Future[EventDb] = {
+    for {
+      rValDbOpt <- rValueDAO.read(nonce)
+      rValDb = rValDbOpt match {
+        case Some(value) => value
+        case None =>
+          throw new RuntimeException(
+            s"Nonce not found from this oracle ${nonce.hex}")
+      }
```
nkohen: nit: should be
```
rValDb <- rValDbOpt match {
  case Some(value) => Future.successful(value)
  case None => Future.failed(new RuntimeException(s"Nonce not found for this oracle ${nonce.hex}")
}
```
same below

benthecarman: I think these are functionally the same because it is already in a `flatMap`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
benthecarman: We just needed to have a private key version, using this one because the address we will staking under will be mainnet native segwit

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: It needed to be specified for the bip32 path, figured I'd make it a `val` instead of writing `0` in the string

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/Event.scala
benthecarman: Already was, will fix scaladoc

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: For nonces it is 0, for the signing key it is 1

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Fixed, was only for R values originally, but had to change because of `ExtPrivateKeyHardened`

Related #2096

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: If we use `getPath` it'll use the r value chain index instead of the signing key one.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Implemented anyways

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
nkohen: This is always 1 as it is the signing key, the comment could be much more clear as this does nothing to prevent nonce reuse which is the main reuse concern. Additionally chainIndex could use a comment as it is not easy to tell why it is the way it is from code alone

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
nkohen: functionally the same but throw should be avoided when possible as it is terribly inefficient

benthecarman: TIL

nkohen (approved):

Christewart (approved):

PR #2095, opened 2020-09-30
Added Schnorr signature point computation test vectors

nkohen:
and their corresponding generators and tests.


dlc/src/main/scala/org/bitcoins/dlc/testgen/SchnorrSigPointTestVector.scala
```diff
@@ -0,0 +1,99 @@
...
+    pubKey: SchnorrPublicKey,
+    pubNonce: SchnorrNonce,
+    signature: SchnorrDigitalSignature,
+    sigPoint: ECPublicKey)
+    extends TestVector {
+  require(signature.sig.getPublicKey == sigPoint,
+          s"Signature ($signature) does not match Signature Point ($sigPoint)")
+
+  def privKey: ECPrivateKey = inputs.privKey
+  def privNonce: ECPrivateKey = inputs.privKey
```
benthecarman: this should be `inputs.privNonce`

dlc/src/main/scala/org/bitcoins/dlc/testgen/SchnorrSigPointTestVectorGen.scala
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    Future.successful(
+      Vector(
+        generateTest,
+        generateTest,
+        generateTest,
+        generateTest,
+        generateTest
+      )
```
benthecarman: nit: use a `Vector.fill` here

benthecarman (changes requested):

dlc/src/main/scala/org/bitcoins/dlc/testgen/SchnorrSigPointTestVector.scala
nkohen: oops

benthecarman (approved):

PR #2097, opened 2020-09-30
Allow any HDCoinType

benthecarman:
Previously we could only represent a `HDCoinType` of 0 or 1, this changes it so we can represent any `HDCoinType`. This will be useful for things like [DLC Oracles](https://github.com/discreetlogcontracts/dlcspecs/issues/102) or if someone wants to use shitcoins >:(

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/hd/HDPathTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/hd/HDPathTest.scala
@@ -70,10 +70,12 @@ class HDPathTest extends BitcoinSUnitTest {
   behavior of "HDCoinType"
 
   it must "correctly represent Bitcoin and Testnet coins" in {
-    HDCoinType.fromInt(0) must be(HDCoinType.Bitcoin)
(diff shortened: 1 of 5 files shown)


Christewart (approved):
Approving, but I would personally like to see something like 

```scala
sealed trait HDCoinType { 
  def toInt: Int
}

object HDCoinType { 
  final case object Bitcoin extends HDCoinType
  final case object Testnet extends HDCoinType
  case class UnknownCoinType(int: Int) extends HDCoinType
}
```
which allows us to pattern match more effectively in my opinion. Not sure why we didn't do this in the first place. 

Feel free to ignore though.

benthecarman:
@Christewart I originally did it like that and it required a much bigger refactor so I decided to do it like this. Can change it to typed way if you'd prefer

Christewart:
> @Christewart I originally did it like that and it required a much bigger refactor so I decided to do it like this. Can change it to typed way if you'd prefer

Let's just get it in for now and consider refactoring later on https://github.com/bitcoin-s/bitcoin-s/issues/2112

PR #2100, opened 2020-10-01
Made TLV scripts UInt16 prefixed and added hash pre-images to dlc_test.json

nkohen:


dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestUtil.scala
```diff
@@ -21,11 +21,11 @@ import scodec.bits.ByteVector
 
 object DLCTestUtil {
 
-  def genOutcomes(size: Int): Vector[Sha256Digest] = {
+  def genOutcomes(size: Int): Vector[(String, Sha256Digest)] = {
```
benthecarman, line 24: It'd be nice if this had it's own type

benthecarman (approved):

PR #2102, opened 2020-10-01
NFC normalization

nkohen:
Introduced `CryptoUtil` string normalization and string hashing.

### DIFF
--- a/crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
@@ -14,16 +14,32 @@ import scodec.bits.{BitVector, ByteVector}
   */
 trait CryptoUtil {
 
+  def normalize(str: String): String = {
+    java.text.Normalizer.normalize(str, java.text.Normalizer.Form.NFC)
+  }
+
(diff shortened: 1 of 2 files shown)


crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
```diff
@@ -14,18 +14,34 @@ import scodec.bits.{BitVector, ByteVector}
   */
 trait CryptoUtil {
 
+  def normalize(str: String): String = {
+    java.text.Normalizer.normalize(str, java.text.Normalizer.Form.NFC)
+  }
+
+  def serializeForHash(str: String): ByteVector = {
+    ByteVector(normalize(str).getBytes("UTF-8"))
+  }
```
benthecarman, line 23: Should these be private?

benthecarman (approved):
ack on ci

crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
nkohen: I'd prefer that they not be (in case I want to generate test vectors in the future for example, and also for debugging in the future)

PR #2103, opened 2020-10-01
Add BIP45 Multisig Purpose

benthecarman:
https://github.com/bitcoin/bips/blob/master/bip-0045.mediawiki#Purpose


nkohen (commented):
Looks good but it would be nice to have tests (like static test vectors, surely other wallets have these lying around someplace?)

benthecarman:
Added tests and fixed CI

Christewart (approved):

Issue #1949, opened 2020-09-01
All commands should return [] rather than empty white space if it's empty


Christewart:
```
./bitcoin-s-cli getutxos

```

I'm about to rescan my wallet, i wanted to confirm that it's empty, so i ran `getutxos` and it doesn't really return anything. 

It would be nice if it returned something like:

```
./bitcoin-s-cli getutxos
[]
```

to indicate we have an empty vector of utxos

PR #2105, opened 2020-10-01
Add DLC Oracle Server Endpoints

### LINKED ISSUE #2084: Make API endpoints for the oracle module
We need to 

1. Create an event
1a. RValue generation
1b. Specify static or dynamic pubkey for an event
2. Sign for an event
3. Endpoint for retrieving signatures for an event
4. Lists events that need to be signed

benthecarman:
Built on top of #2094

Closes #2084 

Adds basic endpoints for interacting with the oracle

TODO: 

- [x] Docs

### DIFF
--- a/app/cli/src/main/resources/logback.xml
+++ b/app/cli/src/main/resources/logback.xml
@@ -1,0 +1,1 @@
+<configuration />
--- a/app/oracle-server/oracle-server.sbt
+++ b/app/oracle-server/oracle-server.sbt
@@ -1,0 +1,27 @@
+name := "bitcoin-s-oracle-server"
+
+// Ensure actor system is shut down
+// when server is quit
+Compile / fork := true
+
+libraryDependencies ++= Deps.oracleServer
+
+mainClass := Some("org.bitcoins.oracle.server.Main")
+
+graalVMNativeImageOptions ++= Seq(
+  "-H:EnableURLProtocols=http",
+  "-H:+ReportExceptionStackTraces",
+  // builds a stand-alone image or reports a failure
+  "--no-fallback",
+  // without this, we get complaints about Function3
+  // I'm not sure why, though...
+  "--initialize-at-build-time=scala.Function3",
+  "--report-unsupported-elements-at-runtime",
+  "--verbose"
+)
+
+packageSummary := "A DLC Oracle"
+
+packageDescription := "A basic DLC oracle that allows you to commit to events and sign them"
+
+enablePlugins(JavaAppPackaging, GraalVMNativeImagePlugin)
--- a/app/oracle-server/src/main/scala/org/bitcoins/oracle/server/Main.scala
+++ b/app/oracle-server/src/main/scala/org/bitcoins/oracle/server/Main.scala
@@ -1,0 +1,124 @@
+package org.bitcoins.oracle.server
+
+import java.nio.file.{Path, Paths}
+
+import akka.actor.ActorSystem
+import com.typesafe.config.ConfigFactory
+import org.bitcoins.core.config._
+import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.db._
+import org.bitcoins.dlc.oracle.DLCOracleAppConfig
+import org.bitcoins.keymanager.bip39.BIP39KeyManager
+import org.bitcoins.server.Server
+
+import scala.concurrent.ExecutionContext
+import scala.util.Properties
+
+object Main extends App with BitcoinSLogger {
+
+  val argsWithIndex = args.zipWithIndex
+
+  val dataDirIndexOpt = {
+    argsWithIndex.find(_._1.toLowerCase == "--datadir")
+  }
+
+  val datadirPathOpt = dataDirIndexOpt match {
+    case None => None
+    case Some((_, dataDirIndex)) =>
+      val str = args(dataDirIndex + 1)
+      val usableStr = str.replace("~", Properties.userHome)
+      Some(Paths.get(usableStr))
+  }
+
+  val configIndexOpt = {
+    argsWithIndex.find(_._1.toLowerCase == "--conf")
+  }
+
+  val baseConfig = configIndexOpt match {
+    case None =>
+      val configPath =
+        datadirPathOpt.getOrElse(AppConfig.DEFAULT_BITCOIN_S_DATADIR)
+      AppConfig.getBaseConfig(configPath)
+    case Some((_, configIndex)) =>
+      val str = args(configIndex + 1)
+      val usableStr = str.replace("~", Properties.userHome)
+      val path = Paths.get(usableStr)
+      ConfigFactory.parseFile(path.toFile).resolve()
+  }
+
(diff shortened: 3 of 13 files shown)


app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
```diff
@@ -0,0 +1,116 @@
...
+      GetEvent.fromJsArr(arr) match {
+        case Failure(exception) =>
+          reject(ValidationRejection("failure", Some(exception)))
+        case Success(GetEvent(nonce)) =>
+          complete {
+            oracle.getEvent(nonce).map {
+              case Some(completed: CompletedEvent) =>
+                Server.httpSuccess(completed.signature.hex)
+              case None | Some(_: PendingEvent) =>
+                Server.httpSuccess("[]")
```
Christewart, line 111: Probably should separate these cases return more useful messages in these cases.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/Event.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.bitcoins.commons.jsonmodels.dlc.SigningVersion
+import org.bitcoins.crypto.{FieldElement, SchnorrDigitalSignature, SchnorrNonce}
+import org.bitcoins.dlc.oracle.storage._
+
+sealed trait Event {
+  def nonce: SchnorrNonce
+  def label: String
+  def numOutcomes: Long
+  def signingVersion: SigningVersion
+  def commitmentSignature: SchnorrDigitalSignature
```
Christewart: A scaladoc would be nice to indicate what the difference between `commitmentSignature` and `attestation` is. The latter is apparent to me but i'm not sure what this is.

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -483,7 +484,71 @@ object OpReturnCommit extends ServerJsonModels {
...
+        Try {
+          val label = labelJs.str
+          val outcomes = outcomesJs.arr.map(_.str).toVector
+
+          CreateEvent(label, outcomes)
+        }
+      case Nil =>
+        Failure(
+          new IllegalArgumentException("Missing label and outcome arguments"))
+      case other =>
```
Christewart, line 514: i think the only other possibility here is `head :: Nil | head :: tail` isn't it? It's always nice to try and retain pattern match exhaustiveness.

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -483,7 +484,71 @@ object OpReturnCommit extends ServerJsonModels {
             s"Bad number of arguments: ${other.length}. Expected: 3"))
     }
   }
+}
+
+// Oracle Models
+
+case class CreateEvent(label: String, outcomes: Vector[String])
```
Christewart: Why do you call it `label` rather than `eventName`? I prefer the latter if you are indifferent.

Christewart (commented):
I gave some preliminary review, but I want to re-review after #2094 gets in. Also we should be adding documentation on this PR with examples of 

1. Using bitcoin-s-cli
2. Using something like `curl` so that it's easy to code against the API from another programming language

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/Event.scala
benthecarman: Done in #2094

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
benthecarman: Done in #2094

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
benthecarman: Was trying to follow #1949 

Imo for apis like this we don't really want human readable returns

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
benthecarman: Those will have the same error message, so I think it is fine as is. We do this for every other `ServerJsonModel` as well

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
benthecarman: Maybe instead we should return an error

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/Main.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.db._
+import org.bitcoins.dlc.oracle.DLCOracleAppConfig
+import org.bitcoins.keymanager.bip39.BIP39KeyManager
+import org.bitcoins.server.Server
+
+import scala.concurrent.ExecutionContext
+import scala.util.Properties
+
+object Main extends App with BitcoinSLogger {
```
Christewart, line 17: nit: So I would like to start renaming `Main` -> `DlcOracleServerMain` (and the same thing applies to other `Main` in our projects). It's just easier to search for rather than looking through 10 `Main` in a intellij :P. Feel free to ignore on this PR though, just a general thought i've had lately.

app/oracle-server/oracle-server.sbt
```diff
@@ -0,0 +1,27 @@
+name := "bitcoin-s-oracle-server"
+
+// Ensure actor system is shut down
+// when server is quit
+Compile / fork := true
```
Christewart, line 5: nit: I don't think this stuff is needed, and i definitely don't think we need to fork for all compile tasks. You can read more about this here: https://www.scala-sbt.org/1.x/docs/Forking.html#Enable+forking

app/cli/src/main/resources/logback.xml
```diff
@@ -0,0 +1 @@
+<configuration />
```
Christewart, line 1: nit: Should we just get rid of this ? Or maybe put a example hin here?

Christewart (approved):
LGTM contingent on docs being added in a separate PR, is that the plan right now?

app/cli/src/main/resources/logback.xml
benthecarman: This is to disable logging in the cli project. Without this we would get logs from parsing some of the arguments which can be confusing to the user

app/oracle-server/oracle-server.sbt
benthecarman: I just copied from the app server's

PR #2109, opened 2020-10-02
Have scala-steward ignore scalafmt updates

benthecarman:
Closes #2138 
Closes #2198 

This should stop scala-steward from creating PRs to update the scalafmt version


Christewart:
I would prefer to have us at least notified when an update is made, and then make the decision to incorporate it. If we ignore (any) dependency we won't get those updates and I have to go manually check, which is a PITA.

Issue #2086, opened 2020-09-30
AppConfig.getBaseConfig() does not seem to use application.conf correctly


Christewart:
Using the dlc oracle as a library, i've specified this `bitcoin-s` configuration in my applications `application.conf` file. 
```
bitcoin-s {
  network = regtest //testnet3, mainnet
  common {
    profile = "slick.jdbc.PostgresProfile$"
    db {
      url = "jdbc:postgresql://localhost:5432/oracle"
      driver = "org.postgresql.Driver"
      username = ""
      password = ""
      numThreads = 1 # I worry about async issues with signing and r values, so let's keep 1 thread
    }
  }

   oracle.profile = ${bitcoin-s.common.profile}
   oracle.db = ${bitcoin-s.common.db}
}
```

Here is the error message I am getting from scala playframework

>com.typesafe.config.ConfigException$UnresolvedSubstitution: reference.conf @ jar:file:/home/suredbits/.cache/coursier/v1/https/repo1.maven.org/maven2/com/typesafe/play/play-server_2.13/2.8.2/play-server_2.13-2.8.2.jar!/reference.conf: 90: Could not resolve substitution to a value: ${play.server.dir}

Which doesn't really make sense as i'm not messing around with play configuration at all.

Looking deep down the stack trace I see 

```
    com.typesafe.config.impl.SimpleConfig.resolve(SimpleConfig.java:41)
     org.bitcoins.db.AppConfig$.getBaseConfig(AppConfig.scala:229)
     org.bitcoins.db.AppConfig.baseConfig$lzycompute(AppConfig.scala:133)
     org.bitcoins.db.AppConfig.baseConfig(AppConfig.scala:132)
     org.bitcoins.db.AppConfig.config$lzycompute(AppConfig.scala:137)
     org.bitcoins.db.AppConfig.config(AppConfig.scala:136)
     org.bitcoins.db.AppConfig.chain$lzycompute(AppConfig.scala:120)
     org.bitcoins.db.AppConfig.chain(AppConfig.scala:119)
     org.bitcoins.db.AppConfig.network$lzycompute(AppConfig.scala:126)
     org.bitcoins.db.AppConfig.network(AppConfig.scala:126)
     org.bitcoins.db.AppConfig.<init>(AppConfig.scala:150)
     org.bitcoins.dlc.oracle.DLCOracleAppConfig.<init>(DLCOracleAppConfig.scala:17)
```

[...]

PR #2113, opened 2020-10-02
Move configuration for sqlite into the 'bitcoin-s' key

Christewart:
This starts to address #2086 

We shouldn't have configurations we are supporting live outside of the `bitcoin-s` configuration name space for HOCON

This also deletes a redundant `db.conf` file that was found in `db-commons-test`


benthecarman (approved):

benthecarman:
postgres tests are failing

PR #2117, opened 2020-10-04
Use Llyod's Oracle recommendations on commitment Signature

benthecarman:
Llyod recommended to me to rename the `commitmentSignature` to `announcementSignature` which I think is better. Also to add the `maturityTime` to the announcement pre-image.

### DIFF
--- a/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
+++ b/dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
@@ -11,6 +11,7 @@ import org.bitcoins.crypto._
 import org.bitcoins.dlc.oracle.storage._
 import org.bitcoins.testkit.core.gen.ChainParamsGenerator
 import org.bitcoins.testkit.fixtures.DLCOracleFixture
+import scodec.bits.ByteVector
 
 class DLCOracleTest extends DLCOracleFixture {
 
@@ -88,7 +89,7 @@ class DLCOracleTest extends DLCOracleFixture {
     }
   }
 
-  it must "create a new event with a valid commitment signature" in {
+  it must "create a new event with a valid announcement signature" in {
     dlcOracle: DLCOracle =>
       for {
         testEventDb <-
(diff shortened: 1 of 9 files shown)


dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -98,10 +99,13 @@ class DLCOracleTest extends DLCOracleFixture {
         assert(rValDbOpt.isDefined)
         val rValDb = rValDbOpt.get
         val hash = CryptoUtil.taggedSha256(
-          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(rValDb.eventName),
-          "DLCv0/Commitment")
+          rValDb.nonce.bytes ++ CryptoUtil.serializeForHash(
+            rValDb.eventName) ++ ByteVector.fromLong(
+            testEventDb.maturationTime.getEpochSecond),
```
nkohen, line 104: I'm not sure I'm happy with using a non-string serialization mixed with a string one, you should string append the long rather than the number's bytes

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -138,16 +139,17 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
       nonce = getKValue(eventName, path, signingVersion).schnorrNonce
 
       hash = CryptoUtil.taggedSha256(
-        nonce.bytes ++ CryptoUtil.serializeForHash(eventName),
-        signingVersion.commitmentTag)
-      commitmentSig = signingKey.schnorrSign(hash.bytes)
+        nonce.bytes ++ CryptoUtil.serializeForHash(eventName) ++ ByteVector
+          .fromLong(maturationTime.getEpochSecond),
```
nkohen, line 143: ditto

nkohen (approved):
ACK because mock

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
benthecarman: Discussed on slack - Conclusion: fine for now, fix when we have a clear spec for the announcement signature

PR #2121, opened 2020-10-05
2020 10 05 redo config

### LINKED ISSUE #2086: AppConfig.getBaseConfig() does not seem to use application.conf correctly

Christewart:
This attempts to fix #2086 

Basically this moves us to using the top level config for resolving things (`bitcoin-s`) rather than diving into the `bitcoin-s` namespace. 

Here is what I do in this PR

1. When resolving things, use absolute paths rather than relative (`bitcoin-s.chain.filterBatchSize` rather than `chain.filterBatchSize`)
2. Make `AppConfig.config` protected so that everyone doesn't have access to it.
3. Make things `lazy` in the `AppConfig` so configuration resolution is not eager. This means we don't resolve the configuration too early when using bitcoin-s as a library (I think this is the root cause of #2086, still testing though).


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.13
+sbt.version=1.4.0
```
benthecarman: Is this needed?

Christewart: No sorry testing out the new version :P

benthecarman (changes requested):
Need to change the `BitcoindRpcAppConfig` to use absolute paths

benthecarman:
Looks like this fixes #2124

Christewart:
> Need to change the `BitcoindRpcAppConfig` to use absolute paths

It's concerning this didn't cause CI failures :raised_eyebrow:

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
```diff
@@ -45,16 +45,17 @@ case class DLCOracleAppConfig(
...
+    for {
+      _ <- super.start()
+    } yield {
+      if (Files.notExists(datadir)) {
+        Files.createDirectories(datadir)
+      }
+      val numMigrations = {
+        migrate()
+      }
+      logger.info(s"Applied $numMigrations to the wallet project")
```
benthecarman: > Applied $numMigrations to the wallet project

this is not the wallet

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -82,33 +82,38 @@ case class WalletAppConfig(
...
+    config.getInt("bitcoin-s.wallet.requiredConfirmations")
 
   override def start(): Future[Unit] = {
-    logger.debug(s"Initializing wallet setup")
+    for {
+      _ <- super.start()
+      //moved out of main class body because we need lazy behavior when loading configs
+      _ = require(
+        requiredConfirmations >= 1,
+        s"requiredConfirmations cannot be less than 1, got: $requiredConfirmations")
```
benthecarman: I think it'd be cleaner to ust change the definition of `requiredConfirmations` to
```scala

lazy val requiredConfirmations: Int = {
    confs = config.getInt("wallet.requiredConfirmations")
	require(
        confs >= 1,
        s"requiredConfirmations cannot be less than 1, got: $confs")
	confs
}

```

benthecarman (changes requested):

benthecarman:
> It's concerning this didn't cause CI failures

We should probably make a `BitcoindRpcAppConfigTest`

benthecarman (approved):

benthecarman:
`DLCOracleAppConfig` needs its server conf to have the absolute path as well

PR #2128, opened 2020-10-05
Increase DLC Oracle test coverage

benthecarman:
Found a bug where if our public key was an odd value it would generate the wrong staking address due to the BIP 340 tie breaker


testkit/src/main/scala/org/bitcoins/testkit/fixtures/DLCOracleDAOFixture.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.concurrent.Future
+
+case class DLCOracleDAOs(
+    rValueDAO: RValueDAO,
+    eventDAO: EventDAO,
+    outcomeDAO: EventOutcomeDAO)
+
+trait DLCOracleDAOFixture extends BitcoinSFixture with EmbeddedPg {
+
+  implicit protected def config: DLCOracleAppConfig
```
Christewart: Is there a reason you don't just move the implementation up here? It seems like you are use the same implementation in all places? 

```scala
BitcoinSTestAppConfig.getDLCOracleWithEmbeddedDbTestConfig(pgUrl)
```

Christewart (approved):

PR #2134, opened 2020-10-06
Update slack links

benthecarman:
Someone emailed me saying the link was expired


Christewart:
These expire every 30 days as far as I know :-(. Maybe we should put a calendar invite out there to remind us?

Christewart (approved):

PR #2137, opened 2020-10-06
Remove requirement for ZMQ with bitcoind backend

benthecarman:


app/server/src/main/scala/org/bitcoins/server/Main.scala
```diff
@@ -219,9 +220,35 @@ object Main extends App with HttpLogger {
...
 
-        zmq = BitcoindRpcBackendUtil.createZMQWalletCallbacks(wallet)
+        blockCount <- bitcoind.getBlockCount
+        // Create callbacks for processing new blocks
+        _ = bitcoindRpcConf.zmqPortOpt match {
+          case Some(_) =>
+            val zmq = BitcoindRpcBackendUtil.createZMQWalletCallbacks(wallet)
+            zmq.start()
+          case None =>
+            var prevCount = blockCount
```
Christewart: This should really be moved out of `Main.scala` and it should be unit tested in the wallet project?

Christewart (changes requested):

benthecarman:
Moved polling to the util file and created a test

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -118,4 +122,32 @@ object BitcoindRpcBackendUtil extends BitcoinSLogger {
...
+      bitcoind: BitcoindRpcClient,
+      startCount: Int,
+      interval: FiniteDuration = 10.seconds)(implicit
+      system: ActorSystem,
+      ec: ExecutionContext): Cancellable = {
+    var prevCount = startCount
+    system.scheduler.scheduleWithFixedDelay(0.seconds, interval) { () =>
+      {
+        bitcoind.getBlockCount.flatMap { count =>
+          if (prevCount != count) {
```
Christewart: Is this always a safe assumption? Or do we need to individually consider the cases

1. prevCount < count
2. prevcount > count
3. prevCount == count

I'm particularily concerned about 2. Is this possible to hit on testnet?

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -118,4 +122,32 @@ object BitcoindRpcBackendUtil extends BitcoinSLogger {
...
+            val range = prevCount.to(count)
+
+            val hashFs =
+              range.map(bitcoind.getBlockHash(_).map(_.flip))
+            prevCount = count
```
Christewart: I'm not sure if we want to adjust `prevCount` so early. What if `hashFs` fail for whatever reason? 

However this is a competing concern with my suggested change, if `prevCount` is not adjusted until everything is successful, is `10.seconds` enough of an interval to make sure we don't just run this `Runnable`again with the same parameters as last time? I think there should be a note in the code about not making the interval too aggressive.

Also logging in this runnable would be EXTREMELY valuable so that we can see errors happening if there are any. Basically a log everytime it's attempting to run and maybe a log on success/failure?

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -118,4 +122,32 @@ object BitcoindRpcBackendUtil extends BitcoinSLogger {
       }
     }
   }
+
+  def startBitcoindBlockPolling(
```
Christewart, line 135: scaladoc please

wallet-test/src/test/scala/org/bitcoins/wallet/BitcoindBlockPollingTest.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.bitcoins.testkit.fixtures.BitcoinSFixture
+import org.bitcoins.testkit.util.BitcoinSAsyncTest
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.testkit.{BitcoinSTestAppConfig, EmbeddedPg}
+import org.bitcoins.wallet.config.WalletAppConfig
+
+import scala.concurrent.Await
+import scala.concurrent.duration.DurationInt
+
+class BitcoindBlockPollingTest extends BitcoinSAsyncTest with EmbeddedPg {
```
Christewart, line 16: Any reason you don't just use the `BitcoinSWalletTest` test trait? I believe that has everything you need?

Christewart (changes requested):

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: 2 is technically possible but extremely unlikely. Can add a case for it though

wallet-test/src/test/scala/org/bitcoins/wallet/BitcoindBlockPollingTest.scala
benthecarman: Because the wallets in there are automatically setup to process new blocks. We want to test that our polling will find the new blocks

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: Added logs and docs about the interval.

Christewart: Do you have any thoughts on other things? I.e. moving where we update `prevCount`?

benthecarman: I think where `prevCount` is set is correct. I'd rather it be before the futures so we don't need to wait to receive and process all the blocks. Like you said then we might double up and request a set twice.

Christewart: yes but then you never realize you received funds in that case if `hashesF` fails. We probably need utilize some sort of `future.recover()` to reset `prevCount` to the original value in the case that something fails. We should probably also make `prevCount` an `AtomicReference[Int]` to be more async safe.

benthecarman: Did both of those

Christewart (approved):

PR #2139, opened 2020-10-06
Full support for non-p2wpkh DLC funding inputs

nkohen:
with passing unit tests


core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
```diff
@@ -86,6 +93,28 @@ case class ScriptSignatureParams[+InputType <: InputInfo](
...
+                         UInt32.zero)),
+      Vector(TransactionOutput(Satoshis.zero, EmptyScriptPubKey)),
+      UInt32.zero
+    )
+
+    val maxWitnessLenF = BitcoinSigner
+      .sign(this, unsignedTx = dummyTx, isDummySignature = true)
+      .map(_.transaction)
+      .map {
+        case wtx: WitnessTransaction  => wtx.witness.head.byteSize.toInt
```
benthecarman, line 112: instead of using `_.head` here shouldn't we be using the `witness(output.vout)`

benthecarman (changes requested):

benthecarman (approved):

PR #2141, opened 2020-10-07
Refactor Mains to use common BitcoinSRunner

### LINKED ISSUE #2116: Refactor commonalties between Main's into a single trait
In #2105 we introduce another Main.scala. From what I can tell there is a lot of duplication between the 2. The configuration should be factored out so we don't need to remember to update it in two places.

benthecarman:
Fixes #2116

`BitcoinSServerMain` is a little funky so we can test it


app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
```diff
@@ -0,0 +1,95 @@
...
+      val str = args(configIndex + 1)
+      val usableStr = str.replace("~", Properties.userHome)
+      val path = Paths.get(usableStr)
+      ConfigFactory
+        .parseFile(path.toFile)
+        .resolve()
+  }
+
+  lazy val configDataDir: Path = Paths.get(
+    baseConfig.getStringOrElse("bitcoin-s.datadir",
```
Christewart, line 72: I don't have anything actionable on this for now, but we do want to make sure we are compatible with the resolution hiearchy in the type safe config specification. I listed it out on #2147

Christewart: Actually thinking about this more, is there a reason we don't move 

1. configDataDir
2. datadirPath
3. network 

Could these all be made available via a `BitcoinSAppConfig` in this trait?

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -256,12 +177,13 @@ object Main extends App with HttpLogger {
   }
 
   //start everything!
-  val run = runMain(args.toVector)
-
-  run.failed.foreach { err =>
-    logger.error(s"Failed to startup server!", err)
-    sys.exit(1)
-  }(scala.concurrent.ExecutionContext.Implicits.global)
+  lazy val run: Unit = {
```
Christewart, line 181: Why is this a `lazy val` ? Maybe make this a `def run(): Unit` ?

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -27,91 +22,17 @@ import org.bitcoins.wallet.Wallet
...
-        case MainNet  => "mainnet"
-        case TestNet3 => "testnet3"
-        case RegTest  => "regtest"
-        case SigNet   => "signet"
-      }
-      datadirPath.resolve(lastDirname)
-    }
 
-    System.setProperty("bitcoins.log.location", datadir.toAbsolutePath.toString)
+class Main(override val args: Array[String]) extends BitcoinSRunner {
```
Christewart: I would really rather not call this or the object below `Main` as it starts to get confusing when searching in intellij. Maybe call this `BitcoinSServerMain`?

Christewart: you can always specify the `Main` that gets run in the build file here: https://github.com/bitcoin-s/bitcoin-s/blob/b1635a8f70b5a50cfd4c00c9781587cc0935127f/app/server/server.sbt#L9-L8

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -377,6 +298,10 @@ object Main extends App with HttpLogger {
   }
 }
 
+object Main extends App {
```
Christewart: Like I said above, i would prefer to name this something else.

Christewart (changes requested):

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
benthecarman: Made it a lazy val so we can't accidentally start the server twice

app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
benthecarman: 1) We need all of these so we can properly set the log location.

2) We only use a `BitcoinSAppConfig` in the app server, the oracle server uses a `DLCOracleAppConfig`

Christewart (approved):

PR #2143, opened 2020-10-07
Small improvements on FundTransactionHandling

benthecarman:
After #1986 this scaladoc was wrong for this function.

Also the logic we were using to build the `txBuilder` and `finalizer` were the exact same as using the `ShufflingNonInteractiveFinalizer.txBuilderFrom` function.


Christewart (approved):

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -138,20 +129,13 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
           logger.info(s"UTXO $index details: ${utxo.output}")
       }
 
-      val inputs =
```
Christewart, line 141: So basically we are moving from doing this explicitly to use the built in code inside of `ShufflingNonInteractiveFinalizer`?

benthecarman: Yes, these exact lines are already in there

Issue #2147, opened 2020-10-07
Write configuration resolution test cases for AppConfig


Christewart:
According to the [typesafe config docs](https://github.com/lightbend/config#standard-behavior), this is the order of 

>
    1. system properties
    2. application.conf (all resources on classpath with this name)
    3. application.json (all resources on classpath with this name)
    4. application.properties (all resources on classpath with this name)
    5. reference.conf (all resources on classpath with this name)

I don't think we honor this at the moment -- specifically with System properties. We should write sanity unit test cases in `dbCommonsTest` to make sure we behave as expected. 

Here is what I had in mind 

```scala
      //set system property
      System.setProperty("bitcoin-s.sqlite.profile",
                         "slick.jdbc.PostgresProfile$")
      val chainConfig = ChainAppConfig(dataDir)
      val assertF = for {
        //resolve the config
        _ <- chainConfig.start()
      } yield {
        //should be instance of postgresprofile as that is what i set the
        //system property to be
        assert(chainConfig.slickDbConfig.profile.isInstanceOf[PostgresProfile])
      }
```

Unfortunately we can't test this with `bitcoin-s.datadir` until #2114  is complete.

benthecarman:
An easier check for the system properties would be to set something like `bitcoin-s.wallet.requiredConfirmations`

PR #2148, opened 2020-10-07
2020 10 06 explicit classloader

Christewart:
This _seems_ to get thing working for me on the sbapi side of things. It's a partial fix to #2086 

The big things here are 

1. Using an explicit class loader (this is recommend by typesafe config [here](https://github.com/lightbend/config#debugging-your-configuration))
2. The usage of `ConfigFactory.defaultOverrides()` -- this should bring in system properties into the configuration resolution, but this is not tested yet. It should be tested more thoroughly on #2147

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
@@ -133,7 +133,12 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
     logger.debug(s"Resolved bitcoin-s config:")
     logger.debug(finalConfig.asReadableJson)
 
-    val resolved = finalConfig.resolve()
+    val resolved = {
+      ConfigFactory
+        .defaultOverrides()
+        .withFallback(finalConfig)
+        .resolve()
+    }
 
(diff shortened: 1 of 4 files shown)


db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -156,8 +161,10 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
 
   lazy val slickDbConfig: DatabaseConfig[JdbcProfile] = {
     Try {
-      DatabaseConfig.forConfig[JdbcProfile](path = s"bitcoin-s.$moduleName",
-                                            config = config)
+      val c = DatabaseConfig.forConfig[JdbcProfile](path =
+                                                      s"bitcoin-s.$moduleName",
+                                                    config = config)
+      c
```
benthecarman, line 167: what's the point of this val

benthecarman (approved):

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
Christewart: I was printing stuff related to it and i guess i forgot to remove it.

PR #2151, opened 2020-10-08
Add ability to create ranged oracle events

benthecarman:
After #2141 I needed to refactor some stuff in the oracle server

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -1015,6 +1015,57 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("createrangedevent")
+        .action((_, conf) =>
+          conf.copy(command = CreateRangedEvent("", Instant.MIN, 0, 0, 0)))
+        .text("Registers an oracle event with a range of outcomes")
+        .children(
+          arg[String]("label")
+            .text("Label for this event")
+            .required()
+            .action((label, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(eventName = label)
+                case other => other
+              })),
+          arg[Instant]("maturationtime")
+            .text("The earliest expected time an outcome will be signed, given in epoch second")
+            .required()
+            .action((time, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(maturationTime = time)
+                case other => other
+              })),
+          arg[Int]("start")
+            .text("The first possible outcome number")
+            .required()
+            .action((start, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(start = start)
+                case other => other
+              })),
+          arg[Int]("stop")
+            .text("The last possible outcome number")
+            .required()
+            .action((stop, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(stop = stop)
+                case other => other
+              })),
+          arg[Int]("step")
+            .text("The increment between each outcome")
+            .required()
+            .action((step, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(step = step)
+                case other => other
+              }))
+        ),
       cmd("getevent")
         .action((_, conf) => conf.copy(command = GetEvent(null)))
         .text("Get an event's details")
@@ -1282,6 +1333,13 @@ object ConsoleCli {
         RequestParam(
           "createevent",
           Seq(up.writeJs(label), up.writeJs(time), up.writeJs(outcomes)))
+      case CreateRangedEvent(eventName, time, start, stop, step) =>
+        RequestParam("createrangedevent",
+                     Seq(up.writeJs(eventName),
+                         up.writeJs(time),
+                         up.writeJs(start),
+                         up.writeJs(stop),
+                         up.writeJs(step)))
       case SignEvent(nonce, outcome) =>
         RequestParam("signevent", Seq(up.writeJs(nonce), up.writeJs(outcome)))
       case GetSignature(nonce) =>
@@ -1565,6 +1623,15 @@ object CliCommand {
       maturationTime: Instant,
       outcomes: Seq[String])
       extends CliCommand
+
+  case class CreateRangedEvent(
+      eventName: String,
+      maturationTime: Instant,
+      start: Int,
+      stop: Int,
+      step: Int)
+      extends CliCommand
+
   case class SignEvent(nonce: SchnorrNonce, outcome: String) extends CliCommand
   case class GetSignature(nonce: SchnorrNonce) extends CliCommand
 }
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -1017,7 +1017,7 @@ object ConsoleCli {
         ),
       cmd("createrangedevent")
         .action((_, conf) =>
-          conf.copy(command = CreateRangedEvent("", Instant.MIN, 0, 0, 0)))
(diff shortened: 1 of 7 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1015,6 +1015,56 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("createrangedevent")
+        .action((_, conf) =>
+          conf.copy(command = CreateRangedEvent("", Instant.MIN, 0, 0, 1)))
+        .text("Registers an oracle event with a range of outcomes")
+        .children(
+          arg[String]("label")
```
Christewart: any reason to use "label" here rather than "eventName"? I would much prefer "eventName".

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -119,6 +119,30 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     }
   }
 
+  def createNewRangedEvent(
```
Christewart, line 127: How do you handle a step where the last step is not uniform? I.e. 

start=0
step=5
stop=11

This should be well documented.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -119,6 +119,30 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      eventName: String,
+      maturationTime: Instant,
+      start: Int,
+      stop: Int,
+      step: Int = 1): Future[EventDb] = {
+    if (step == 0) {
```
Christewart, line 133: `step <= 0`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -119,6 +119,30 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      Future.failed(new IllegalArgumentException("Cannot have a step of 0"))
+    } else {
+      if (step > 0) {
```
Christewart, line 136: If you add the comment above, I don't think this is needed anymore.

Also, why must `start < step `? Can't you have 

start=10
step=1
stop=20

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -119,6 +119,30 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     }
   }
 
+  def createNewRangedEvent(
+      eventName: String,
+      maturationTime: Instant,
+      start: Int,
+      stop: Int,
+      step: Int = 1): Future[EventDb] = {
```
Christewart, line 132: I would just add requires upfront saying

```scala
require(start >= 0)
require(stop >= start)
require(step > 0)
```

With appropriate error messages

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -81,6 +81,56 @@ class DLCOracleTest extends DLCOracleFixture {
...
+          dlcOracle.createNewRangedEvent(eventName, time, start, stop, step)
+        eventOpt <- dlcOracle.getEvent(testEventDb.nonce)
+      } yield {
+        assert(eventOpt.isDefined)
+        val event = eventOpt.get
+
+        assert(event.isInstanceOf[PendingEvent])
+        assert(event.eventName == eventName)
+        assert(event.outcomes == expectedOutcomes)
+        assert(event.numOutcomes == expectedOutcomes.size)
```
Christewart: nit: It might be nice to assert a hard coded value here, the answer should be 40 right?

docs/oracle/oracle-server.md
```diff
@@ -56,6 +56,13 @@ For more information on how to use our built in `cli` to interact with the serve
   - `label` - Label for this event
   - `maturationtime` - The earliest expected time an outcome will be signed, given in epoch second
   - `outcomes` - Possible outcomes for this event
+- `createrangedevent` `label` `maturationtime` `start` `stop` `step` - Registers an oracle event with a range of outcomes
+  - `label` - Label for this event
```
Christewart: same thing applies here with  `label` -> `eventName`

Christewart (changes requested):
This is not related to this PR, but we should really create a 

```scala
trait DLCOracleApi { } 
```

In the core project.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: negative step is okay ie 

start = 100
stop = 0
step = -1

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: I don't see the reason in disallowing descending values or negative values

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: That is what was on the other one, can fix

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: see comments above and below

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -81,6 +81,85 @@ class DLCOracleTest extends DLCOracleFixture {
...
 
+  it must "create a ranged event and get its details" in {
+    dlcOracle: DLCOracle =>
+      val time = TimeUtil.now
+      val eventName = "ranged"
+      val start = -100
+      val stop = 100
+      val step = 5
+
+      val expectedOutcomes = start.to(stop, step).map(_.toString)
```
nkohen, line 92: ahhhhhhhhh do we know how we feel about encodings for numeric events? I could imagine an argument that since they transport in TLVs as `int32` perhaps they should be serialized as such when hashing? Either way we need to be extra careful to make sure we match whatever spec is going to do on this

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -81,6 +81,85 @@ class DLCOracleTest extends DLCOracleFixture {
...
+
+  it must "create a ranged event with a non-uniform step" in {
+    dlcOracle: DLCOracle =>
+      val time = TimeUtil.now
+      val eventName = "ranged"
+      val start = 0
+      val stop = 11
+      val step = 5
+
+      val expectedOutcomes = Vector("0", "5", "10")
```
nkohen, line 121: ditto

nkohen (commented):
Looks good, just one concern about serialization

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
benthecarman: Yeah agreed. This is mostly just waiting on the actual spec

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Ended up going with this in #2162

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
benthecarman: Now signs the int32's bytes in #2162

PR #2153, opened 2020-10-08
2020 10 08 issue 2147

### LINKED ISSUE #2147: Write configuration resolution test cases for AppConfig

Christewart:
This fixes #2147 

This adds two test cases: 

1. A test case to make sure the config gets resolved correctly by substituting variables in the config
2. That system properties override default configuration specification

### DIFF
--- a/db-commons-test/src/test/scala/org/bitcoins/db/AppConfigTest.scala
+++ b/db-commons-test/src/test/scala/org/bitcoins/db/AppConfigTest.scala
@@ -1,12 +1,7 @@
-package org.bitcoins.testkit.db
-
-import java.nio.file.{Path, Paths}
-import java.util.Properties
+package org.bitcoins.db
 
 import com.typesafe.config.ConfigFactory
 import org.bitcoins.core.config.TestNet3
-import org.bitcoins.db.AppConfig
-import org.bitcoins.server.BitcoinSAppConfig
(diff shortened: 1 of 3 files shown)


benthecarman (approved):
ACK on CI

PR #2155, opened 2020-10-08
Make tx an argument for decoderawtransaction

benthecarman:
previously you would need to do `bitcoin-s-cli decoderawtransaction --tx 0200000...`

now you can do `bitcoin-s-cli decoderawtransaction 0200000...`


nkohen (approved):

Christewart (commented):
Is there any documentation that needs to be updated? I don't see this even listed on our server endpoints?

https://bitcoin-s.org/docs/next/applications/server#server-endpoints

PR #2157, opened 2020-10-08
Add Wallet State Descriptors

benthecarman:
This creates `WalletStateDescriptor`s. These are used to add any data about the state of the wallet to the db.

The first being `SyncHeight` which is just a block height and hash. This is used to keep track of what was the last block processed by the wallet.

This is useful for when we are using the bitcoind backend so we know which block we missed while offline so we can then process them.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -20,6 +20,38 @@ import scala.util.{Failure, Success}
 /** Useful utilities to use in the wallet project for syncing things against bitcoind */
 object BitcoindRpcBackendUtil extends BitcoinSLogger {
 
+  def catchupWalletToBitcoind(bitcoind: BitcoindRpcClient, wallet: Wallet)(
+      implicit ec: ExecutionContext): Future[Unit] = {
+    for {
+      bitcoindHeight <- bitcoind.getBlockCount
+      walletStateOpt <- wallet.getSyncHeight()
+      _ <- walletStateOpt match {
+        case None => FutureUtil.unit // New wallet, nothing to sync
+        case Some(syncHeight) =>
+          if (syncHeight.height > bitcoindHeight) {
+            Future.failed(
+              new RuntimeException("Bitcoind is not caught up to wallet"))
+          } else {
+            val blockRange = syncHeight.height.to(bitcoindHeight).tail
+
+            val func: Vector[Int] => Future[Unit] = { range =>
+              val hashFs =
+                range.map(bitcoind.getBlockHash(_).map(_.flip))
+              for {
+                hashes <- Future.sequence(hashFs)
+                _ <- wallet.nodeApi.downloadBlocks(hashes)
+              } yield ()
+            }
+
+            FutureUtil.batchExecute(elements = blockRange.toVector,
+                                    f = func,
+                                    init = Vector.empty,
+                                    batchSize = 25)
+          }
+      }
+    } yield ()
+  }
+
   def createWalletWithBitcoindCallbacks(
       bitcoind: BitcoindRpcClient,
       wallet: Wallet)(implicit ec: ExecutionContext): Wallet = {
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -20,34 +20,63 @@ import scala.util.{Failure, Success}
 /** Useful utilities to use in the wallet project for syncing things against bitcoind */
 object BitcoindRpcBackendUtil extends BitcoinSLogger {
 
+  /** Has the wallet process all the blocks it has not seen up until bitcoind's chain tip */
   def catchupWalletToBitcoind(bitcoind: BitcoindRpcClient, wallet: Wallet)(
       implicit ec: ExecutionContext): Future[Unit] = {
+
+    def doSync(walletHeight: Int, bitcoindHeight: Int): Future[Unit] = {
+      if (walletHeight > bitcoindHeight) {
+        Future.failed(new RuntimeException(
+          s"Bitcoind and wallet are in incompatible states, " +
+            s"wallet height: $walletHeight, bitcoind height: $bitcoindHeight"))
+      } else {
+        val blockRange = walletHeight.to(bitcoindHeight).tail
+
+        logger.info(s"Syncing ${blockRange.size} blocks")
+
+        val func: Vector[Int] => Future[Unit] = { range =>
+          val hashFs =
+            range.map(bitcoind.getBlockHash(_).map(_.flip))
+          for {
+            hashes <- Future.sequence(hashFs)
+            _ <- wallet.nodeApi.downloadBlocks(hashes)
+          } yield ()
+        }
+
+        FutureUtil
+          .batchExecute(elements = blockRange.toVector,
+                        f = func,
+                        init = Vector.empty,
+                        batchSize = 25)
+          .map(_ => ())
+      }
+    }
+
     for {
       bitcoindHeight <- bitcoind.getBlockCount
       walletStateOpt <- wallet.getSyncHeight()
       _ <- walletStateOpt match {
-        case None => FutureUtil.unit // New wallet, nothing to sync
-        case Some(syncHeight) =>
-          if (syncHeight.height > bitcoindHeight) {
-            Future.failed(
-              new RuntimeException("Bitcoind is not caught up to wallet"))
-          } else {
-            val blockRange = syncHeight.height.to(bitcoindHeight).tail
-
-            val func: Vector[Int] => Future[Unit] = { range =>
-              val hashFs =
-                range.map(bitcoind.getBlockHash(_).map(_.flip))
-              for {
-                hashes <- Future.sequence(hashFs)
-                _ <- wallet.nodeApi.downloadBlocks(hashes)
-              } yield ()
+        case None =>
+          for {
+            utxos <- wallet.listUtxos()
+            lastConfirmedOpt = utxos.filter(_.blockHash.isDefined).lastOption
+            _ <- lastConfirmedOpt match {
+              case None => FutureUtil.unit
+              case Some(utxo) =>
+                for {
+                  heightOpt <- bitcoind.getBlockHeight(utxo.blockHash.get)
+                  _ <- heightOpt match {
+                    case Some(height) =>
+                      logger.info(
+                        s"Last utxo occurred at block $height, syncing from there")
+                      doSync(height, bitcoindHeight)
+                    case None => FutureUtil.unit
+                  }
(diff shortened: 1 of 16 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -54,15 +54,21 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
             newWallet
           }
       }
-    res.onComplete(failure =>
+    resF.onComplete(failure =>
       signalBlockProcessingCompletion(block.blockHeader.hash, failure))
-    res.foreach(_ =>
+    resF.foreach(_ =>
```
Christewart: I think these log messages should be done after the `updateSyncHeight` call is successful, so that the logs are atomic. 

As currently stands, you can block processing work, but updating sync height failing. From looking at the logs I would think everything succeeded.

This does open a broader question of what to do in the case of sync height failing to update, but block processing works. Can anything nasty happen there?

wallet/src/main/scala/org/bitcoins/wallet/models/WalletStateDescriptorDAO.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class WalletStateDescriptorDb(
+    tpe: WalletStateDescriptorType,
+    descriptor: WalletStateDescriptor) {
+  require(descriptor.descriptorType == tpe)
+}
+
+case class WalletStateDescriptorDAO()(implicit
```
Christewart, line 22: So the plan is to throw them all in the same table as strings? I would prefer to have some sort of structure in the database so I can query for simple things height or block hash.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,38 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 /** Useful utilities to use in the wallet project for syncing things against bitcoind */
 object BitcoindRpcBackendUtil extends BitcoinSLogger {
 
+  def catchupWalletToBitcoind(bitcoind: BitcoindRpcClient, wallet: Wallet)(
```
Christewart: scaladoc

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,38 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+      implicit ec: ExecutionContext): Future[Unit] = {
+    for {
+      bitcoindHeight <- bitcoind.getBlockCount
+      walletStateOpt <- wallet.getSyncHeight()
+      _ <- walletStateOpt match {
+        case None => FutureUtil.unit // New wallet, nothing to sync
```
Christewart: What about old wallets?

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,38 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+        case Some(syncHeight) =>
+          if (syncHeight.height > bitcoindHeight) {
+            Future.failed(
+              new RuntimeException("Bitcoind is not caught up to wallet"))
```
Christewart: This is a little bit more troubling than bitcoind not being caught up? For instance, are you on the right network? Maybe there was a huge reorg (testnet?). I don't think this case is as simple as "we are ahead of bitcoind"

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: is this method even tested anywhere?

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/models/WalletStateDescriptorDAO.scala
benthecarman: I used strings so it can be more extensible in the future. So if we wanted to add a descriptor that had data about something that doesn't have a height and block hash, we can easily do without needing to make a new table.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,67 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+    for {
+      bitcoindHeight <- bitcoind.getBlockCount
+      walletStateOpt <- wallet.getSyncHeight()
+      _ <- walletStateOpt match {
+        case None =>
+          for {
+            utxos <- wallet.listUtxos()
+            lastConfirmedOpt = utxos.filter(_.blockHash.isDefined).lastOption
+            _ <- lastConfirmedOpt match {
+              case None => FutureUtil.unit
```
Christewart, line 64: So if we shutdown the bitcoin-s wallet in a state that only has unconfirmed utxos, we won't be able to sync ever?

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,67 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+          .batchExecute(elements = blockRange.toVector,
+                        f = func,
+                        init = Vector.empty,
+                        batchSize = 25)
+          .map(_ => ())
+      }
+    }
+
+    for {
+      bitcoindHeight <- bitcoind.getBlockCount
```
Christewart, line 56: Why are we doing things by height here rather than hashes? I'm a little skeptical that this is reorg safe.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -16,6 +16,67 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 /** Useful utilities to use in the wallet project for syncing things against bitcoind */
 object BitcoindRpcBackendUtil extends BitcoinSLogger {
 
+  /** Has the wallet process all the blocks it has not seen up until bitcoind's chain tip */
+  def catchupWalletToBitcoind(bitcoind: BitcoindRpcClient, wallet: Wallet)(
```
Christewart: nit: We use the terminology `sync` elsewhere, if you don't have a strong preference I would prefer to try and be consistent: `syncWalletToBitcoind()`

wallet-test/src/test/scala/org/bitcoins/wallet/BitcoindBackendTest.scala
```diff
@@ -0,0 +1,81 @@
...
+      _ <- bitcoind.sendToAddress(addr, amountToSend)
+      _ <- bitcoind.generateToAddress(6, addr)
+
+      // assert wallet hasn't seen it yet
+      firstBalance <- wallet.getBalance()
+      _ = assert(firstBalance == Satoshis.zero)
+
+      // Set sync height
+      _ <-
+        wallet.stateDescriptorDAO.updateSyncHeight(header.hashBE, header.height)
```
Christewart, line 67: I'm confused, why are we _externally_ calling `updateSyncHeight()` here? Shouldn't this be set internally when we call `catchupWalletToBitcoind()`. Does this have to be done because of the comment I left in the case of syncing the wallet that only has unconfirmed utxos when the bitcoin-s wallet is shutdown?

Christewart (changes requested):

wallet-test/src/test/scala/org/bitcoins/wallet/BitcoindBackendTest.scala
benthecarman: If this wasn't set it would be considered a new wallet and not try to catch up. This is just so we are able to test the catch up sequence.

benthecarman: If this wasn't set it would be considered a new wallet and not try to catch up. This is just so we are able to test the catch up sequence.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: The thing is if we do use hashes and there is a discrepancy we don't really have a way to resolve it because we only know 1 hash. Since we are using bitcoind for the backend we should trust that it is handling reorgs for us.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: Yeah I guess, not sure the best way to solve this, only real solution would be to rescan every block.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: The problem exists in either situation. With a hash we can atleast then query bitcoind and see if that hash is in the best chain.

If the hash is not in the best chain, can we go back and look through the block hashes that are in the [`SpendingInfoTable`](https://github.com/bitcoin-s/bitcoin-s/blob/6ce7d3d7019ca11eafe347728a5d0dc83647c692/wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala#L490-L489) looking to see which is the last one apart of the best chain?

We then would need to go and "revert the state" of all utxos that were confirmed in blocks after the `block_hash` we find in the best chain right?

Christewart: >We then would need to go and "revert the state" of all utxos that were confirmed in blocks after the block_hash we find in the best chain right?

Hmm... I'm not sure what "revert the state" means here though. It could be mean 

`PendingConfirmation` <- `Confirmed`
 `PendingConfirmationsSpent` <- `ConfirmedSpent`

Heck, it could even mean 

`DoesNotExist` <- `PendingConfirmationsSpent` (!)

benthecarman: Maybe we can use scantxoutset when there are discrepancies?

or maybe just in general

https://developer.bitcoin.org/reference/rpc/scantxoutset.html

Christewart: I think we agreed to table this discussion for now (as it applies any backend we have). #2179 is the issue to track these problems

Christewart (approved):

PR #2160, opened 2020-10-09
Fix npe exception on oracleServer/run

Christewart:
Fixes a `NullPointerException` on startup that was introduced in #2141


nkohen (approved):
ACK

PR #2149, opened 2020-10-07
Oracle Announcement TLVs

benthecarman:

PR #2162, opened 2020-10-09
Use New Oracle TLVs in DLCOracle

benthecarman:
Built on top of #2185 

This modifies the DLCOracle to use new oracle descriptors in its implementation and db.


dlc-oracle/src/main/resources/postgresql/oracle/migration/V1__dlc_oracle_baseline.sql
```diff
@@ -13,22 +13,14 @@ CREATE TABLE r_values
...
-    pubkey          TEXT      NOT NULL,
-    event_name      TEXT      NOT NULL UNIQUE,
-    num_outcomes    INTEGER   NOT NULL,
-    signing_version TEXT      NOT NULL,
-    maturation_time TIMESTAMP NOT NULL,
-    attestation     TEXT,
+    nonce                TEXT      NOT NULL,
+    pubkey               TEXT      NOT NULL,
+    event_name           TEXT      NOT NULL UNIQUE,
+    event_descriptor_tlv TEXT      NOT NULL,
```
Christewart: we can no longer modify existing migrations. Write new ones.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -193,21 +234,29 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+                  "Outcome given is not a possible outcome for this event")
+          CryptoUtil.serializeForHash(outcome)
+        case rangeV0: RangeEventDescriptorV0TLV =>
+          val int32 = Int32(outcome.toLong)
+          require(rangeV0.outcomes.contains(int32),
+                  "Outcome given is not a possible outcome for this event")
+          int32.bytes
+        case _: ExternalEventDescriptorV0TLV =>
+          throw new IllegalArgumentException(
+            "Unable to handle external event descriptors")
```
Christewart: nit: Add the external tlv to the exception?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/Event.scala
```diff
@@ -29,61 +28,168 @@ sealed trait Event {
...
+
+  def announcementTLV: OracleAnnouncementTLV = {
+    eventTLV match {
+      case v0TLV: OracleEventV0TLV =>
+        OracleAnnouncementV0TLV(announcementSignature, v0TLV)
+    }
+  }
+}
+
+sealed trait PendingEvent extends Event
```
Christewart: nit: How would you feel about renaming the `Event` superclass here to something like `OracleEvent` just so it's more unique when seraching through intellij. 

I feel like every library has an `Event` type.

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -4,18 +4,22 @@ import java.sql.SQLException
...
+import org.bitcoins.testkit.core.gen.{ChainParamsGenerator, TLVGen}
 import org.bitcoins.testkit.fixtures.DLCOracleFixture
-import scodec.bits.ByteVector
 
 class DLCOracleTest extends DLCOracleFixture {
 
-  val testOutcomes: Vector[String] = (0 to 10).map(_.toString).toVector
+  val testOutcomes: Vector[String] = Vector("sunny", "windy", "rainy", "cloudy")
+
+  val testDescriptor: EventDescriptorTLV = TLVGen.eventDescriptorTLV.sampleSome
```
Christewart: You need to add range descriptors to the `TLVGen.eventDescriptorTLV` generator. Currently the range stuff is not tested at all. Also you should really make this a `def` because otherwise its a single static value in `testDescriptor` rather than a randomly generated value everytime it is called.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -95,37 +95,80 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+    *
+    * If the final step would be higher than stop, it will not go past it.
+    */
+  def createNewRangedEvent(
+      eventName: String,
+      maturationTime: Instant,
+      start: Int32,
+      stop: Int32,
+      step: UInt16): Future[EventDb] = {
+    require(stop >= start, s"Start ($start) must be after Stop ($stop)")
```
Christewart: Shouldn't there be a require on `maturationTime` to be in the future as well? or is this done else where ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -138,10 +181,13 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
-        nonce.bytes ++ CryptoUtil.serializeForHash(eventName) ++ ByteVector
-          .fromLong(maturationTime.getEpochSecond),
-        signingVersion.announcementTag)
+      epoch = UInt32(maturationTime.getEpochSecond)
+
+      eventTLV =
+        OracleEventV0TLV(publicKey, nonce, epoch, descriptorTLV, eventName)
+
+      hash =
+        CryptoUtil.taggedSha256(eventTLV.bytes, signingVersion.announcementTag)
```
Christewart: nit: I think we should really refactor the tagged stuff into a util method so people like me don't forget to add the tag in the future :-)

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/storage/EventDAOTest.scala
```diff
@@ -38,12 +39,14 @@ class EventDAOTest extends DLCOracleDAOFixture {
     0,
     SchnorrDigitalSignature(nonce, FieldElement.one))
 
+  val descriptor: EventDescriptorTLV = TLVGen.eventDescriptorTLV.sampleSome
```
Christewart: Same thing here, don't use `val` with scalatest generators.

Christewart: or `lazy val`

docs/oracle/oracle-server.md
```diff
@@ -56,6 +56,12 @@ For more information on how to use our built in `cli` to interact with the serve
   - `label` - Label for this event
   - `maturationtime` - The earliest expected time an outcome will be signed, given in epoch second
   - `outcomes` - Possible outcomes for this event
+- `createrangedevent` `name` `maturationtime` `start` `stop` `step` - Registers an oracle event with a range of outcomes
```
Christewart, line 94: Let's do code samples fo this as well for both creation and signing to be explicit.

Christewart (changes requested):

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
benthecarman: The range stuff is all tested, this is just for if we need a dummy descriptor.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Currently it is not enforced at all. I'm apprehensive of enforcing it anywhere because the oracle is a representation of the real world and things don't always go as expected. 

I can see us requiring that the maturation date be in the future, however, since nothing is enforced after that I don't really see the point. I think we should let the user set whatever time they want.

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
Christewart: You still must add it to the generator so that it will get randomly sampled, currently it is not and never will be in future development.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
Christewart: Can you do DLCs based on past events? 

The only reason I can think of allowing for maturation times in the past is if you are transferring data from one database to another and you want to use this method to do it. I would argue you should use the underlying `DAO`s though.

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
benthecarman: Oh I see now, that was missed in #2149

benthecarman:
Responded to all review minus the db migration. I need to spend some time figuring out a way to do this in a change backwards way.

benthecarman:
Got the db migration to be backwards compatible

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,41 +91,96 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+    *
+    * If the final step would be higher than stop, it will not go past it.
+    */
+  def createNewRangedEvent(
+      eventName: String,
+      maturationTime: Instant,
+      start: Int32,
+      stop: Int32,
+      step: UInt16): Future[EventDb] = {
+    require(stop >= start, s"Start ($start) must be after Stop ($stop)")
```
Christewart: I think the message here is wrong. Stop must be after Start?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,22 +1,38 @@
...
-import org.bitcoins.crypto.StringFactory
+import org.bitcoins.core.protocol.tlv.OracleEventTLV
+import org.bitcoins.crypto.{CryptoUtil, SchnorrNonce, StringFactory}
+import scodec.bits.ByteVector
 
 sealed abstract class SigningVersion {
-  def nonceTag: String
-  def announcementTag: String
-  def outcomeTag: String
+  def calcNonceTweak(nonce: SchnorrNonce, eventName: String): ByteVector
```
Christewart, line 10: nit: scaladocs on all of these would be nice.

Christewart: on all methods in `SigningVersion`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -193,31 +267,60 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+          require(enumV0.outcomes.contains(outcome),
+                  "Outcome given is not a possible outcome for this event")
+          Future.successful((CryptoUtil.serializeForHash(outcome), eventDb))
+        case rangeV0: RangeEventDescriptorV0TLV =>
+          val int32 = Int32(outcome.toLong)
+          require(rangeV0.outcomes.contains(int32),
+                  "Outcome given is not a possible outcome for this event")
+          Future.successful((int32.bytes, eventDb))
+        case external: ExternalEventDescriptorV0TLV =>
+          if (external == oldDbRepDescriptor) {
```
Christewart: So is there a reason we are doing this here rather than just doing something on server startup similar to what we did we `runChainWorkRecalc()`? Then we don't have to worry about keeping this logic in here forever 

https://github.com/bitcoin-s/bitcoin-s/blob/30b6e22662258baac1b802c3e263760a66a8542d/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L238-L237

Christewart: In my mind, all this legacy migration stuff would happen inside of `DLCOracleAppConfig.start()` so a user can just start the module, we will apply migrations and also migrate legacy data if any exists.

Christewart: Looking at this even more, unfortunately it looks like we screwed up our chainwork migration. If a user was using bitcoin-s chain module pre-chainwork, and then updated their bitcoin-s-chain dependency, the `runChainWorkRecalc()`would not get run as it's not affiliated with the chain module, but just run in the `BitcoinSServerMain`class. With breaking changes like this we need to make sure they are affiliated with the specific module rather than the stand alone binary

https://github.com/bitcoin-s/bitcoin-s/blob/30b6e22662258baac1b802c3e263760a66a8542d/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L59-L58

Christewart (changes requested):

Christewart:
Wait to merge this until Ben is done talking with @LLfourn about oracle TLVs

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Waiting on #2173

LLFourn:
I made further comments wrt: https://github.com/discreetlogcontracts/dlcspecs/pull/105

I don't think there is a strong need to reconcile differences in ideas atm (although not putting the public key in the thing signed by that key is something I think is important). IMO you should feel free to go ahead and buildl first and specify later! I will work on doing something similar to what was suggested for my oracle and we can reconcile later :)

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
Christewart: Ok this is merged now :+1:

benthecarman:
This is now also waiting on #2185

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1015,31 +1017,131 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("createrangedevent")
+        .action((_, conf) =>
+          conf.copy(command = CreateRangedEvent("", Instant.MIN, 0, 0, 1)))
```
Christewart: nit: Why not refactor this into a `CreatedRangedEvent.empty` with this value.

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1015,31 +1017,131 @@ object ConsoleCli {
...
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(step = step)
+                case other => other
+              }))
+        ),
+      cmd("createlargerangeevent")
+        .action((_, conf) =>
+          conf.copy(command =
+            CreateLargeRangedEvent("", Instant.MIN, 0, isSigned = false, 0)))
```
Christewart: nit: Same comment here wrt to making `CreateLargeRangedEvent.empty`

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1052,17 +1154,63 @@ object ConsoleCli {
                 case other => other
               }))
         ),
-      cmd("getsignature")
-        .action((_, conf) => conf.copy(command = GetSignature(null)))
-        .text("Get the signature from a signed event")
+      cmd("signnumber")
+        .action((_, conf) => conf.copy(command = SignNumber(null, 0)))
```
Christewart: Don't use `null`. Use a default value if anything. I.e. `RangeEventDescriptorV0TLV.empty`

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1052,17 +1154,63 @@ object ConsoleCli {
...
+            .required()
+            .action((num, conf) =>
+              conf.copy(command = conf.command match {
+                case signNumber: SignNumber =>
+                  signNumber.copy(num = num)
+                case other => other
+              }))
+        ),
+      cmd("signlargenumber")
+        .action((_, conf) => conf.copy(command = SignLargeNumber(null, 0)))
```
Christewart: same with `null` here

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -278,39 +284,134 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
...
     val step = UInt16(iter.takeBits(16))
 
-    RangeEventDescriptorV0TLV(start, stop, step)
+    RangeEventDescriptorV0TLV(nonce, start, count, step)
+  }
+}
+
+case class LargeRangeEventDescriptorV0TLV(
+    base: UInt16,
+    isSigned: Boolean,
```
Christewart: So there seems to be overlap between this and the other PR https://github.com/bitcoin-s/bitcoin-s/pull/2185#discussion_r506661952

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -336,6 +336,16 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
   def posInt: Int = {
     Math.abs(scala.util.Random.nextInt())
   }
+
+  /** From 0th place to numDigits place */
```
Christewart: a more descriptive scaladoc please. Something like "Decomposes the given num into a ....." ? Also linking to the spec here in the future would be great.

db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
```diff
@@ -125,6 +126,16 @@ trait DbManagement extends BitcoinSLogger {
     flyway.info()
   }
 
+  def migrationsApplied(): Int = {
+    val applied = flyway.info().applied()
+    driver match {
+      case SQLite =>
+        applied.size
+      case PostgreSQL =>
+        applied.size - 1
```
Christewart, line 136: nit: Maybe add the comment i have in `DbManagementTest` so that we don't forget why we have a `-1` in the future?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
```diff
@@ -44,17 +45,37 @@ case class DLCOracleAppConfig(
...
+    super.start().flatMap { _ =>
       if (Files.notExists(datadir)) {
         Files.createDirectories(datadir)
       }
       val numMigrations = {
         migrate()
       }
       logger.info(s"Applied $numMigrations to the dlc oracle project")
+
+      if (migrationsApplied() == 2) {
```
Christewart, line 60: nit: Maybe add a comment why this is necessary

Christewart: Isn't this the wrong ordering? 

This seems to do 

1. Migrate tables
2. Read existing data from database
3. Insert data into new tables

but don't we want this ordering:

1. Read existing data from database
2. Apply migration
3. Insert data into new tables

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
```diff
@@ -0,0 +1,29 @@
+package org.bitcoins.dlc.oracle
+
+import org.bitcoins.crypto.CryptoUtil
+import scodec.bits.ByteVector
+
+sealed trait DLCAttestationType {
```
Christewart, line 7: `extends NetworkElement`?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
```diff
@@ -0,0 +1,29 @@
...
+  def bytes: ByteVector
+}
+
+case class EnumAttestation(outcome: String) extends DLCAttestationType {
+  def bytes: ByteVector = CryptoUtil.serializeForHash(outcome)
+}
+
+case class RangeAttestation(outcome: Long) extends DLCAttestationType {
```
Christewart, line 16: Maybe there can be a shared parent trait called `NumberAttestation` between `RangeAttestation` and `LargeRangeAttestation`?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
```diff
@@ -0,0 +1,29 @@
...
+  def bytes: ByteVector = CryptoUtil.serializeForHash(outcome.toString)
+}
+
+case class LargeRangeSignAttestation(positive: Boolean)
```
Christewart: I don't really understand what the purpose of this attestation is? My understanding with all other classes in this file is they are actual attestations themselves. 

For number attestsations, do we imagine there would be a `Vector[RangeAttestation]` that commits to each individual digital and that is what is published? 

So in reality, we would need to publish something like `(LargeRangeSignAttestation, Vector[LargeRangeAttestation])`?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
-      eventDbs <- eventDAO.findAll()
-      outcomes <- eventOutcomeDAO.findAll()
-    } yield {
-      val rValDbsByNonce = rValDbs.groupBy(_.nonce)
-      val outcomesByNonce = outcomes.groupBy(_.nonce)
-      eventDbs.map(db =>
-        Event(rValDbsByNonce(db.nonce).head, db, outcomesByNonce(db.nonce)))
+  def listEvents(): Future[Vector[OracleEvent]] = {
+    eventDAO.findAll().map { eventDbs =>
+      val events = eventDbs.groupBy(_.eventDescriptorTLV)
```
Christewart, line 95: Why do we need the `.groupBy()` here? Is there some sort of implicit sorting required?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      events.values.map(dbs => OracleEvent(dbs)).toVector
     }
   }
 
-  def getEvent(nonce: SchnorrNonce): Future[Option[Event]] = {
+  def getEvent(nonce: SchnorrNonce): Future[Option[OracleEvent]] = {
```
Christewart: nit: I prefer `findEvent()` for things like this as the return type is similar to the scala standard library find. I.e. 

```scala

val x: Option[Int] = Vector(1,2,3).find(_ == 1)

```

`get` is only used to return concrete `T` types, i.e. 

```scala

val x: Int = Vector(1,2,3).find(_ == 1).get

```

I guess at the end of the day I just care about being consistent if we want to use `get`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      maturationTime: Instant,
+      base: UInt16,
+      isSigned: Boolean,
+      numDigits: Int): Future[OracleAnnouncementTLV] = {
+    require(base > UInt16.zero,
+            s"base cannot be less than 1, got ${base.toInt}")
+    require(numDigits > 0, s"numDigits cannot be less than 1, got $numDigits")
+
+    val descriptorTLV = if (isSigned) {
+      UninitializedLargeRangeEventDescriptorV0(base, isSigned, numDigits + 1)
```
Christewart: Why `+ 1` here, is this implicitly required because of number representation? https://github.com/bitcoin-s/bitcoin-s/pull/2185#discussion_r506661952

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+
     for {
       indexOpt <- rValueDAO.maxKeyIndex
-      index = indexOpt match {
+      firstIndex = indexOpt match {
         case Some(value) => value + 1
         case None        => 0
       }
 
       signingVersion = SigningVersion.latest
```
Christewart: Maybe make `signingVersion` a parameter? You can provide a default value if you want (`SigningVersion.latest`). I just think this makes testing much easier.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      epoch = UInt32(maturationTime.getEpochSecond)
+
+      eventTLV = OracleEventV0TLV(publicKey, epoch, descriptorTLV, eventName)
+
+      announcementBytes = signingVersion.calcAnnouncementBytes(eventTLV)
+      announcementSignature = signingKey.schnorrSign(announcementBytes)
+
+      eventOutcomeDbs = descriptorTLV match {
+        case enum: EnumEventDescriptorV0TLV =>
+          require(rValueDbs.size == 1)
```
Christewart: I guess this invariant is obvious to me, can you add a error message so readers can understand what this invariant is required? Plus this gives useful error messages for users of the API.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+          val nonce = rValueDbs.head.nonce
+          enum.outcomes.map { outcome =>
+            val hash = signingVersion.calcOutcomeHash(outcome)
+            EventOutcomeDb(nonce, outcome, hash)
+          }
+        case range: RangeEventDescriptorV0TLV =>
+          require(rValueDbs.size == 1)
```
Christewart: same here.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -91,85 +91,207 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      _ <- eventDAO.createAll(eventDbs)
       _ <- eventOutcomeDAO.createAll(eventOutcomeDbs)
-    } yield eventDb
+    } yield {
+      OracleEvent(eventDbs).announcementTLV
+    }
   }
 
-  def signEvent(nonce: SchnorrNonce, outcome: String): Future[EventDb] = {
+  def signEvent[T <: DLCAttestationType](
```
Christewart: So i think this should be renamed to something like `signSingleNonceEvent()`. 

I propse creating a new marker trait 

```scala
trait SingleNonceEvent { self: EventDescriptorTLV =>
  def nonce: SchnorrNonce
}
```

Then the  parameter should be changed to be `descriptor: SingleNonceEvent`

Christewart: Oh and then the type for 

```scala 
case class EnumEventDescriptorV0TLV(
    nonce: SchnorrNonce,
    outcomes: Vector[String])
    extends EventDescriptorTLV 
    with SingleNonceEvent
    
    
case class RangeEventDescriptorV0TLV(
    nonce: SchnorrNonce,
    start: Int32,
    count: UInt32,
    step: UInt16)
    extends EventDescriptorTLV
    with SingleNonceEvent 
```

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -193,27 +315,82 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
               s"No event saved with nonce ${nonce.hex} $outcome"))
       }
 
-      eventOutcomeOpt <- eventOutcomeDAO.read((nonce, outcome))
-      eventOutcomeDb <- eventOutcomeOpt match {
-        case Some(value) => Future.successful(value)
-        case None =>
-          Future.failed(new RuntimeException(
-            s"No event outcome saved with nonce and message ${nonce.hex} $outcome"))
+      outcomeBytes = eventDb.eventDescriptorTLV match {
```
Christewart: note to self, review this on round 2 of code review. It seems like we can get rid of the `.asInstanceOf[]` hopefully.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -0,0 +1,268 @@
...
+    eventTLV match {
+      case v0TLV: OracleEventV0TLV =>
+        OracleAnnouncementV0TLV(announcementSignature, v0TLV)
+    }
+  }
+}
+
+sealed trait PendingOracleEvent extends OracleEvent
+
+sealed trait CompletedOracleEvent extends OracleEvent {
```
Christewart, line 55: can't we add an invariant here actually checking to see the signatures are valid?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -0,0 +1,268 @@
...
+    maturationTime: Instant,
+    announcementSignature: SchnorrDigitalSignature,
+    eventDescriptorTLV: LargeRangeEventDescriptorV0TLV,
+    attestations: Vector[FieldElement])
+    extends CompletedOracleEvent
+    with LargeRangeV0OracleEvent
+
+object OracleEvent {
+
+  def apply(eventDbs: Vector[EventDb]): OracleEvent = {
```
Christewart: nit: Maybe rename to `fromEventDbs()`? I always prefer using `.from()` rather tahn apply. I guess you can just create it and delegate directly the the `apply()` if you want.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -0,0 +1,268 @@
...
+    val eventDb = eventDbs.head
+    require(eventDbs.forall(_.eventDescriptorTLV == eventDb.eventDescriptorTLV),
```
Christewart, line 163: nice invariant :+1:

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -0,0 +1,268 @@
...
+            "EventDbs must all refer to the same event")
+
+    (eventDb.eventDescriptorTLV, eventDb.attestationOpt) match {
+      case (enum: EnumEventDescriptorV0TLV, Some(sig)) =>
+        require(eventDbs.size == 1, "Enum events may only have have eventDb")
```
Christewart: Maybe with the `SingleNonceEvent` stuff above this can be enforced at the type level instead of needing invariants?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
```diff
@@ -0,0 +1,65 @@
+package org.bitcoins.dlc.oracle
+
+import org.bitcoins.core.number._
+import org.bitcoins.core.protocol.tlv._
+import org.bitcoins.crypto.SchnorrNonce
+
+sealed trait UninitializedEventDescriptor[+T <: EventDescriptorTLV] {
```
Christewart: Can you add a scaladoc on this? I don't really understand what this means. What does it mean if an event descriptor is uninitialized?

Christewart: Is this basically an `UnitializedEventBuilder` similar to a `VectorBuilder` ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
```diff
@@ -0,0 +1,65 @@
...
+case class UninitializedExternalEventDescriptorV0(externalName: String)
+    extends UninitializedEventDescriptor[ExternalEventDescriptorV0TLV] {
+
+  override def addNonces(
+      nonces: Vector[SchnorrNonce]): ExternalEventDescriptorV0TLV = {
+    require(nonces.size == 1, "ExternalEventDescriptorV0 can only have 1 nonce")
+    ExternalEventDescriptorV0TLV(nonces.head, externalName)
+  }
+
+  override val noncesNeeded: Int = 1
```
Christewart: Why is this hard coded?

Christewart: I think it should be left abstract so we don't forget to override this when we add more case classes to our ADT in the future.

Christewart (changes requested):
This is partial review, but thought I should get it to you so you can start addressing. 

I want to mull over the type hierarchy a bit, I know @nkohen is thinking about this as well.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
benthecarman: This is just an internal type, this won't ever be serialized.

Also by doing this it will be annoying for enum because then it will override the `apply(String)` function to be `fromHex`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
benthecarman: This just an internal type for passing around data that we will sign. When we publish the actual attestation it will be a `Vector[SchnorrDigitalSignature]`, this type just holds the preimage to the hash of what we will be signing.

I think the confusion comes from the name of this type. Idk what would be better tho

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: I am using `groupBy` here so we can get all the eventDbs with the same `eventDescriptorTLV`, ones with the same `eventDescriptorTLV` will be for the same event, ie signing a large number.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: If the number is signed then there is an extra nonce for signing if it is positive or negative.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Idk if I like this because this is used for signing multi-nonce events as well. When doing multi-nonce events this function is just called multiple times

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
benthecarman: Since event descriptors have the nonces in them we can't build a descriptor without first initializing it in the oracle. The idea for this type is to hold all the data for an event descriptor minus the nonces.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
benthecarman: Currently in the spec external event descriptors have 1 nonce

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
benthecarman: I don't really see the benefit in adding this as there should never be a case where we don't care if it is a `RangeAttestation` or a `LargeRangeAttestation`, we will always want to know the exact one we have.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
benthecarman: It is correct. We need to the migrations first so we can read the data, otherwise we would need the old DAOs.

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: This is built on top of that PR

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: This is what we do for all cli commands

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: We can't match on our type `T` :( need scala 3

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
benthecarman: We could but it'd be very inefficient. We'd need to go through and check against each outcome

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
benthecarman: What the migrations do is put in a dummy event descriptor. Then here we calculate the correct event descriptor and update it

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Found a better way around this

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
benthecarman: since `eventDbs` is a param we would still need these

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
Christewart: Oh i see why you are doing it now, but this isn't great. Hmm...

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
Christewart: Ok so my comment is right? It's really a `EventDescriptorBuilder`?

Christewart: also scaladoc please!!

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
Christewart: .... that's not a good thing.

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
Christewart: #2200

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
benthecarman: > Ok so my comment is right? It's really a EventDescriptorBuilder?

Not really, you have to add all the nonces at once. Maybe I should rename it to `setNonces`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/UninitializedEventDescriptor.scala
benthecarman: https://github.com/discreetlogcontracts/dlcspecs/pull/55#discussion_r504176719

benthecarman: Looks like this was removed actually

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: Ok, this would be useful information to include in the original comment. If you add documentation your code reviewers wont ask as many stupid questions :-)

benthecarman: Woops, was in the original PRs description. Fixed

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -336,6 +336,16 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
   def posInt: Int = {
     Math.abs(scala.util.Random.nextInt())
   }
+
+  /** Decomposes num from 0th place to numDigits place into base 10 */
```
Christewart: isn't it into the given `base` parameter rather than base 10? Isn't it assumed that `num` is base 10?

Christewart: Also I don't see a unit test for decompose? We probably can have some sort of a symmetric decompostion and "recomposition" (right term?)

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/storage/EventDAOTest.scala
```diff
@@ -28,22 +29,28 @@ class EventDAOTest extends DLCOracleDAOFixture {
...
-  val dummyRValDb: RValueDb = RValueDb(
-    nonce,
-    eventName,
-    HDPurpose(0),
-    HDCoinType.Bitcoin,
-    0,
-    0,
-    0,
-    SchnorrDigitalSignature(nonce, FieldElement.one))
+  val dummyRValDb: RValueDb =
```
Christewart, line 32: I think I was hinting at a empty values earlier in the `Cli` code, could this be an example of `RValueDb.empty` and the same thing applies with `SchnorrDigitalSignature.empty`?

Christewart: Or I think using `RValueDb.dummy` would be ok too since "empty" doesn't really make sense.

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -278,39 +284,134 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
...
-    val stop = Int32(iter.takeBits(32))
+    val count = UInt32(iter.takeBits(32))
     val step = UInt16(iter.takeBits(16))
 
-    RangeEventDescriptorV0TLV(start, stop, step)
+    RangeEventDescriptorV0TLV(nonce, start, count, step)
+  }
+}
+
+case class LargeRangeEventDescriptorV0TLV(
```
Christewart: So it's usually a "code smells" (anti design pattern) when you start seeing a bunch of `Option[]` in your class. Here is how I design this to get rid of excess `Option`

```scala
sealed trait LargeRangeEventDescriptorV0TLV extends EventDescriptorTLV {
  /** The base for the number we are signing */
  def base: UInt16
  /** The nonces corresponding to digits in the range
   * ??? is this right ???
   * */
  def nonces: Vector[SchnorrNonce]
  
  def order: Int = nonces.length

  override val tpe: BigSizeUInt = LargeRangeEventDescriptorV0TLV.tpe

  override val value: ByteVector = {
    val isSignedByte = this match {
      case _: SignedLargeRangeEventDescriptor =>
        ByteVector(0x01)
      case _: UnsignedLargeRangeEventDescriptor => 
        ByteVector(0x00)
    }

    val numNonces = UInt16(nonces.size)
    val noncesBytes = nonces.foldLeft(ByteVector.empty)(_ ++ _.bytes)

    base.bytes ++ isSignedByte ++ numNonces.bytes ++ noncesBytes
  }
}

/** Represents a large range event that uses a signed number scheme 
 * [[ link to doc!! ]]
 * */
case class SignedLargeRangeEventDescriptor(base: UInt16, nonces: Vector[SchnorrNonce], signNonce: SchnorrNonce) extends LargeRangeEventDescriptorV0TLV

/** Represents a large range event of unsigned numbers  scheme 
 * [[ link to doc!! ]]
 * */
case class UnsignedLargeRangeEventDescriptor(base: UInt16, nonces: Vector[SchnorrNonce]) extends LargeRangeEventDescriptorV0TLV
```

Now you don't have to worry about `signNonceOpt` -- it just exists when in the `Signed` subtype. Also now the superclass doesn't need any `digitNonces` since the all nonces in the super type are just digits.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -189,31 +314,60 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
     } yield updated
   }
+
+  def signLargeRange(
+      eventDescriptorTLV: LargeRangeEventDescriptorV0TLV,
+      num: Long): Future[OracleEvent] = {
+
+    // Make this a vec so it is easier to add on
+    val signSigF =
+      eventDescriptorTLV.signedNonceOpt match {
```
Christewart: Now you can pattern match on the `LargeRangeEventDescriptor` instead of `signNonceOpt` if you take my ADT suggestion for `LargeRangeEventDesciptor`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -189,31 +314,60 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+          val signOutcome = LargeRangeSignAttestation(num >= 0)
+          signEvent(nonce, signOutcome).map(db => Vector(db))
+        case None =>
+          FutureUtil.emptyVec[EventDb]
+      }
+
+    val decomposed = NumberUtil.decompose(Math.abs(num),
+                                          eventDescriptorTLV.base.toInt,
+                                          eventDescriptorTLV.digitNonces.size)
+    val digitSigFs = eventDescriptorTLV.digitNonces.reverse.zipWithIndex.map {
```
Christewart: why does this need to be reversed? Does signing order matter here?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -189,31 +314,60 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+
+  def signLargeRange(
+      eventDescriptorTLV: LargeRangeEventDescriptorV0TLV,
+      num: Long): Future[OracleEvent] = {
+
+    // Make this a vec so it is easier to add on
+    val signSigF =
+      eventDescriptorTLV.signedNonceOpt match {
+        case Some(nonce) =>
+          val signOutcome = LargeRangeSignAttestation(num >= 0)
```
Christewart: Hmm... this doesn't seem right. For instance the number `1` can be both `signed` and `unsigned`, just checking if it's `>= 0` doesn't work does it?

It seems if we are signing numbers, the oracle needs to pre-commit if we are doing signed numbers or unsigned numbers, and then we need to track that. We can't just tell that at signing time can we?

Christewart (changes requested):

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: > For instance the number 1 can be both signed and unsigned, just checking if it's >= 0 doesn't work does it?

This is correct, `LargeRangeSignAttestation` takes in a boolean so we check if we want to sign "+" or "-" based on the number. If it is positive or 0 it signs "+", it is negative it signs "-".

> It seems if we are signing numbers, the oracle needs to pre-commit if we are doing signed numbers or unsigned numbers, and then we need to track that. We can't just tell that at signing time can we?

This is correct, we are only signing if it is positive or negative if we precommit to signing it, this is saved in the `eventDescriptorTLV`.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Yes

https://github.com/discreetlogcontracts/dlcspecs/pull/55#discussion_r507920129

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: Done in #2185

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,27 +1,81 @@
...
+import org.bitcoins.crypto.{CryptoUtil, SchnorrNonce, StringFactory}
+import scodec.bits.ByteVector
 
 sealed abstract class SigningVersion {
-  def nonceTag: String
-  def announcementTag: String
-  def outcomeTag: String
+
+  /** Calculates the tweak for the deterministically generated nonce */
+  def calcNonceTweak(nonce: SchnorrNonce, eventName: String): ByteVector
```
nkohen: I have no clue what this tweak is from the scaladoc

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,27 +1,81 @@
...
+
+  /** Calculates the bytes to sign for an OracleAnnouncement */
+  def calcAnnouncementBytes(eventTLV: OracleEventTLV): ByteVector
+
+  /** Calculates the bytes to sign for an event outcome */
+  def calcOutcomeHash(
```
nkohen, line 16: Should probably call this (and below) EnumOutcome because it isn't general

nkohen: e.g. `calcEnumOutcomeHash`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,27 +1,81 @@
...
 sealed abstract class SigningVersion {
-  def nonceTag: String
-  def announcementTag: String
-  def outcomeTag: String
+
+  /** Calculates the tweak for the deterministically generated nonce */
+  def calcNonceTweak(nonce: SchnorrNonce, eventName: String): ByteVector
+
+  /** Calculates the bytes to sign for an OracleAnnouncement */
+  def calcAnnouncementBytes(eventTLV: OracleEventTLV): ByteVector
```
nkohen: Should be called `calcAnnouncementHash`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,27 +1,81 @@
...
 
-    override def outcomeTag: String = "DLCv0/Outcome"
+    override def calcOutcomeHash(
+        descriptor: EventDescriptorTLV,
+        byteVector: ByteVector): ByteVector =
+      CryptoUtil.taggedSha256(byteVector, "DLCv0/Outcome").bytes
+  }
+
+  /** Used before we had an actual signing algorithm in the spec */
+  final case object BasicSHA256SigningVersion extends SigningVersion {
```
nkohen, line 51: I'm confused why we have two pre-spec signing versions

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
```diff
@@ -1,27 +1,81 @@
...
+    override def calcAnnouncementBytes(eventTLV: OracleEventTLV): ByteVector =
+      CryptoUtil.sha256(eventTLV.bytes).bytes
+
+    override def calcOutcomeHash(
+        descriptor: EventDescriptorTLV,
+        byteVector: ByteVector): ByteVector = {
+      descriptor match {
+        case _: EnumEventDescriptorV0TLV | _: RangeEventDescriptorV0TLV |
+            _: DigitDecompositionEventDescriptorV0TLV =>
+          CryptoUtil.sha256(byteVector).bytes
```
nkohen, line 70: What on earth would `byteVector` be for a digit decomp? Should probably throw in that case

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1015,31 +1012,173 @@ object ConsoleCli {
...
+              "base exponent by which to multiply the number represented by " +
+              "the composition of the digits to obtain the actual outcome value.")
+            .action((precision, conf) =>
+              conf.copy(command = conf.command match {
+                case createRangedEvent: CreateRangedEvent =>
+                  createRangedEvent.copy(precision = precision)
+                case other => other
+              }))
+        ),
+      cmd("createlargerangeevent")
```
nkohen: probably needs a rename

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1015,31 +1012,173 @@ object ConsoleCli {
...
+          arg[String]("name")
+            .text("Name for this event")
+            .required()
+            .action((name, conf) =>
+              conf.copy(command = conf.command match {
+                case createLargeRangedEvent: CreateLargeRangedEvent =>
+                  createLargeRangedEvent.copy(eventName = name)
+                case other => other
+              })),
+          arg[Instant]("maturationtime")
```
nkohen, line 1134: nit: consider rename

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1052,17 +1191,63 @@ object ConsoleCli {
                 case other => other
               }))
         ),
-      cmd("getsignature")
-        .action((_, conf) => conf.copy(command = GetSignature(null)))
-        .text("Get the signature from a signed event")
+      cmd("signnumber")
```
nkohen: should rename to make less ambiguous (signforrange or something like that)

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1052,17 +1191,63 @@ object ConsoleCli {
...
+        .action((_, conf) => conf.copy(command = SignNumber(null, 0)))
+        .text("Signs a ranged event")
+        .children(
+          arg[OracleEventV0TLV]("event")
```
nkohen, line 1228: I'm seeing lots of this and we probably want to move to event_id with some of these where that is probably a hash or something

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1052,17 +1191,63 @@ object ConsoleCli {
...
+            .text("Number to sign for this event")
+            .required()
+            .action((num, conf) =>
+              conf.copy(command = conf.command match {
+                case signNumber: SignNumber =>
+                  signNumber.copy(num = num)
+                case other => other
+              }))
+        ),
+      cmd("signlargenumber")
```
nkohen: signdigits

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1558,13 +1782,43 @@ object CliCommand {
...
+      eventName: String,
+      maturationTime: Instant,
+      start: Int,
+      stop: Int,
+      step: Int,
+      unit: String,
+      precision: Int)
+      extends CliCommand
+
+  case class CreateLargeRangedEvent(
```
nkohen: ping

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1558,13 +1782,43 @@ object CliCommand {
...
+      precision: Int)
+      extends CliCommand
+
+  case class SignEvent(oracleEventV0TLV: OracleEventV0TLV, outcome: String)
+      extends CliCommand
+
+  case class SignNumber(oracleEventV0TLV: OracleEventV0TLV, num: Long)
+      extends CliCommand
+
+  case class SignLargeNumber(oracleEventV0TLV: OracleEventV0TLV, num: Long)
```
nkohen: ping

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
```diff
@@ -13,6 +15,14 @@ case class OracleRoutes(oracle: DLCOracle)(implicit system: ActorSystem)
     extends ServerRoute {
   import system.dispatcher
 
+  def getDescriptor(
+      announcementTLV: OracleAnnouncementTLV): EventDescriptorTLV = {
+    announcementTLV match {
+      case v0: OracleAnnouncementV0TLV =>
+        v0.eventTLV.eventDescriptor
+    }
+  }
```
nkohen: descriptor should probably be a member of announcement

nkohen: that or event is and it is a member of event

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
```diff
@@ -43,41 +53,130 @@ case class OracleRoutes(oracle: DLCOracle)(implicit system: ActorSystem)
...
+                    val digits = 0.until(decomp.numDigits.toInt).map { _ =>
+                      0
+                        .until(decomp.base.toInt)
+                        .map(s => Str(s.toString))
+                        .toVector
+                    }
+
+                    val vecs = digits :+ sign
+                    vecs.map(vec => Arr.from(vec))
+                }
```
nkohen, line 155: Is this meant to be a list of all outcomes possible? If so I don't think this is correct, but more importantly shouldn't be allowed

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -519,22 +519,172 @@ object CreateEvent extends ServerJsonModels {
...
+            "Missing label, maturationTime, start, stop, and step arguments"))
+      case other =>
+        Failure(
+          new IllegalArgumentException(
+            s"Bad number of arguments: ${other.length}. Expected: 5"))
+    }
+  }
+}
+
+case class CreateLargeRangedEvent(
```
nkohen: ping

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -519,22 +519,172 @@ object CreateEvent extends ServerJsonModels {
...
+            "Missing oracle event tlv and num arguments"))
+      case other =>
+        Failure(
+          new IllegalArgumentException(
+            s"Bad number of arguments: ${other.length}. Expected: 2"))
+    }
+  }
+}
+
+case class SignLargeRangeEvent(oracleEventTLV: OracleEventV0TLV, num: Long)
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -574,7 +576,9 @@ object OracleEventV0TLV extends TLVFactory[OracleEventV0TLV] {
   }
 }
 
-sealed trait OracleAnnouncementTLV extends TLV
+sealed trait OracleAnnouncementTLV extends TLV {
```
nkohen, line 580: should probably also always contain a Schnorr signature and a pubkey

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -336,6 +336,16 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
...
+  def decompose(num: Long, base: Int, numDigits: Int): Vector[Int] = {
+    var currentNum: Long = num
+    (0 until numDigits).toVector.map { _ =>
+      val digit = currentNum % base
+      currentNum = currentNum / base
+      digit.toInt
+    }
+  }
```
nkohen: Agree with chris on this and also I've reversed this on my branch to make things consistent in Vector[Int] representations everywhere (most significant digit to least)

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCAttestationType.scala
```diff
@@ -0,0 +1,30 @@
+package org.bitcoins.dlc.oracle
+
+import org.bitcoins.crypto.CryptoUtil
+import scodec.bits.ByteVector
+
+sealed trait DLCAttestationType {
```
nkohen, line 7: scaladoc

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -78,11 +80,8 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     require(path.forall(_.hardened),
             s"Cannot use a BIP32Path with unhardened nodes, got $path")
     val priv = extPrivateKey.deriveChildPrivKey(path).key
-    val hash =
-      CryptoUtil.taggedSha256(
-        priv.schnorrNonce.bytes ++ CryptoUtil.serializeForHash(label),
-        signingVersion.nonceTag)
-    val tweak = ECPrivateKey(hash.bytes)
+    val tweakBytes = signingVersion.calcNonceTweak(priv.schnorrNonce, label)
```
nkohen, line 83: what purpose does this tweak serve?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -189,31 +327,74 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+
+      hashBytes = eventOutcomeDb.hashedMessage
+      sig = signingKey.schnorrSignWithNonce(hashBytes, kVal)
 
       updated = eventDb.copy(attestationOpt = Some(sig.sig))
       _ <- eventDAO.update(updated)
     } yield updated
   }
+
+  def signLargeRange(
```
nkohen: ping

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -0,0 +1,231 @@
...
+
+import java.time.Instant
+
+import org.bitcoins.commons.jsonmodels.dlc.SigningVersion
+import org.bitcoins.core.number.UInt32
+import org.bitcoins.core.protocol.tlv._
+import org.bitcoins.crypto._
+import org.bitcoins.dlc.oracle.storage.EventDb
+
+sealed trait OracleEvent {
```
nkohen, line 15: scaladoc

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
benthecarman: Was using tagged hashes before that I randomly chose myself, now in spec it just says the SHA256

Christewart:
needs rebase

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
benthecarman: We get a `ByteVector` from calling `DigitDecompositionAttestation`'s `.bytes` which is just 
```scala
def bytes: ByteVector = CryptoUtil.serializeForHash(outcomeString)
```

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/SigningVersion.scala
benthecarman: This is used for all event types

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
benthecarman: It doesn't list all possible outcomes ie 0-100000, it lists the different outcomes for each digit

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: So if you empty your database but re-use your seed that way you don't reuse your nonces, this ensures you'd need to have the same labels as well

Christewart (approved):

Issue #2161, opened 2020-10-09
Add CI row to actually build and run binaries


Christewart:
This is to prevent things like #2160 in the future. Regresions in actually running standalone servers because of easy to miss things like NullPointerExceptions on startup.

Christewart:
Maybe we can actually already add this to rows that test things like `appServer` and `oracleServer` so we can not waste CI minutes on this? Perhaps we can even cache the binaries some how from the compile task?

benthecarman:
When we add these we should verify that the log file is being created in the correct location

Christewart:
We now have `appServerTest/test` that essentially does this

PR #2163, opened 2020-10-09
Fix NPE and log location on server start up

benthecarman:
#2160 fixed the NPE on the oracle server start up but broke the log location.

This fixes the NPE in the oracle server start up without breaking the setting of the log location

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
@@ -80,7 +80,7 @@ trait BitcoinSRunner extends BitcoinSLogger {
   private lazy val network: BitcoinNetwork =
     BitcoinNetworks.fromString(networkStr)
 
-  lazy val datadir: Path = {
+  val datadir: Path = {
     lazy val lastDirname = network match {
       case MainNet  => "mainnet"
       case TestNet3 => "testnet3"
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
@@ -8,7 +8,7 @@ import org.bitcoins.core.config._
 import org.bitcoins.core.util.BitcoinSLogger
 import org.bitcoins.db.{AppConfig, ConfigOps}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Properties
 
 trait BitcoinSRunner extends BitcoinSLogger {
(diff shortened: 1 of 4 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
```diff
@@ -80,7 +80,7 @@ trait BitcoinSRunner extends BitcoinSLogger {
   private lazy val network: BitcoinNetwork =
     BitcoinNetworks.fromString(networkStr)
 
-  lazy val datadir: Path = {
+  val datadir: Path = {
```
Christewart: I think we should keep this lazy. Since we don't actually have test cases for server startup (#2161 ). We have no way to detect regressions on CI. This is one of the joys of Scala's weird behavior with initialization of traits with `val`s inside of it. Sometimes it's too eager and the thing behind the `val` isn't actually filled -- which gives us a `NullPointerException`. 

Also why is logging set here and not somewhere else? Doesn't this need to be done inside every submodule? If not, why isn't logging working as expected in the server project? It seems we have disconnect from library behavior and standalone binary behavior ?

benthecarman: The log file location needs to be set here. We need to parse the config before so we know our datadir and network.

Christewart: >The log file location needs to be set here.  

Why? It seems odd to me that the logging property is set within `val datadir` to begin with.

benthecarman: That is the soonest time we know it.

Christewart: Here is the scala documentation on this situation: https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html

TLDR: Use `lazy val`

I was actually unaware of this compiler flag: `-Xcheckinit`. Thats pretty cool.

benthecarman: I guess we can move it back to `lazy val`, we just then lose our guarantee that the log file location is set before the application starts

benthecarman: Changed it so it uses a `lazy val` again and will now set the log location automatically before the setting up anything else

Christewart (approved):

Christewart:
Approved, but can you do one last sanity check locally that everything is working with appSEver and oracleServer?

benthecarman:
> Approved, but can you do one last sanity check locally that everything is working with appSEver and oracleServer?

Good call, needed a small fix for the oracle, all good now!

Christewart:
> > Approved, but can you do one last sanity check locally that everything is working with appSEver and oracleServer?
> 
> Good call, needed a small fix for the oracle, all good now!

IT's soooooo easy to do! THat's why we need #2161

PR #2170, opened 2020-10-10
Create fromConfig functions for AppConfigs

### LINKED ISSUE #2114: Allow `bitcoin-s.datadir` configuration key to be used instead of passing in a file
This is how [`AppConfigFactory`](https://github.com/bitcoin-s/bitcoin-s/blob/9dc7b2c0dd373a2cd6c2ff4da1b14bb4af95f294/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala#L9-L8) is defined currently. 

This is weird, because we allow users (specifically when using bitcoin-s as a library!) to specify a configuration in their application.conf for `bitcoin-s.datadir`. You can see this on the website here: https://bitcoin-s.org/docs/next/config/configuration#example-configuration-file

If you look closely at `AppConfigFactory`, we have no option for constructing an `AppConfig` _solely_ from the contents of the configuration file. 

Right now I have to do something like this to construct a `DLCOracleAppConfig()` for instance

```scala
  val datadir: Path =
    Files.createTempDirectory(s"hello-world-${scala.util.Random.nextInt(100)}")
  val appConfig = DLCOracleAppConfig.fromDatadir(datadir)
  val dbConfig = appConfig.slickDbConfig
```

We need to create something like 

`AppConfig.fromConfig()`

benthecarman:
Fixes #2114

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala
@@ -1,12 +1,17 @@
 package org.bitcoins.db
 
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
 
 import com.typesafe.config.Config
 
 import scala.concurrent.ExecutionContext
 
 trait AppConfigFactory[C <: AppConfig] {
+
+  def fromConfig(config: Config)(implicit ec: ExecutionContext): C = {
+    val configDataDir: Path = Paths.get(config.getString("bitcoin-s.datadir"))
+    fromDatadir(configDataDir, Vector(config))
+  }
 
   def fromDefaultDatadir(confs: Vector[Config] = Vector.empty)(implicit
       ec: ExecutionContext): C = {
--- a/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala
@@ -2,7 +2,7 @@ package org.bitcoins.db
 
 import java.nio.file.{Path, Paths}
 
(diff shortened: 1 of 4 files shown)


db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala
```diff
@@ -1,13 +1,18 @@
...
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
 
 import com.typesafe.config.Config
 
 import scala.concurrent.ExecutionContext
 
 trait AppConfigFactory[C <: AppConfig] {
 
+  def fromConfig(config: Config)(implicit ec: ExecutionContext): C = {
```
Christewart, line 11: We should add 

```scala

def fromClassPathConfig()(impliicit ec: ExecutionContext): C = { 
  fromConfig(ConfigFactory.load())
}
```

^ That needs to be unit tested, it can probably be unit tested in the same way we test things here: https://github.com/bitcoin-s/bitcoin-s/blob/475ed19df6634f584ce828ac597ca54704963e8b/db-commons-test/src/test/scala/org/bitcoins/db/DBConfigTest.scala#L24 

However I think the path will need to be adjusted to be inside of 

`src/test/resources/application.conf`

You should assert that the datadir (and maybe a few other random settings in different projects) are set correctly.

Christewart: More typesafe documentation: https://github.com/lightbend/config#standard-behavior

Christewart (changes requested):

Christewart (approved):

PR #2171, opened 2020-10-10
Add test for server startup

benthecarman:


app/server-test/src/test/scala/org/bitcoins/server/ServerRunTest.scala
```diff
@@ -31,4 +40,51 @@ class ServerRunTest extends BitcoinSAsyncTest {
...
+      _ <- binding.terminate(5.seconds)
+      _ = thread.interrupt()
+    } yield {
+      // Cleanup
+      directory.deleteRecursively()
+
+      val expectedDir = datadir.resolve("regtest")
+
+      assert(
+        System.getProperty("bitcoins.log.location") == expectedDir.toString)
```
Christewart, line 93: I think you actually want to use `System.clearPropety()`, probably in the 

```scala
after { 
  System.clearProperty("bitcoins.log.location")
}
```

Christewart (changes requested):

Christewart (approved):

PR #2173, opened 2020-10-12
2020 10 12 issue 2164

### LINKED ISSUE #2164: Move flyway instance out of DbManagement.migrate()
https://github.com/bitcoin-s/bitcoin-s/blob/475ed19df6634f584ce828ac597ca54704963e8b/db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala#L122-L121

We should have it be a `private val` in the class. Then we can use other APIs flyway provides like `flyway.info()`

Christewart:
Fixes #2164 

This makes the flyway config accessible outside of `DbManagement.migrate()`. 

This PR also adds a new method, `DbManagement.info()` which is just a proxy for [`flyway.info()`](https://flywaydb.org/documentation/command/info)

### DIFF
--- a/db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
+++ b/db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
@@ -4,6 +4,7 @@ import com.typesafe.config.Config
 import org.bitcoins.chain.config.ChainAppConfig
 import org.bitcoins.chain.db.ChainDbManagement
 import org.bitcoins.db.DatabaseDriver._
+import org.bitcoins.dlc.oracle.DLCOracleAppConfig
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.db.NodeDbManagement
 import org.bitcoins.testkit.BitcoinSTestAppConfig.ProjectType
@@ -53,6 +54,9 @@ class DbManagementTest extends BitcoinSAsyncTest with EmbeddedPg {
       case PostgreSQL => 4
     }
     assert(result == expected)
+    val flywayInfo = chainDbManagement.info()
+    assert(flywayInfo.applied().length == expected)
+    assert(flywayInfo.pending().length == 0)
   }
 
   it must "run migrations for wallet db" in {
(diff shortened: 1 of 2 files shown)


db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
```diff
@@ -101,26 +118,18 @@ trait DbManagement extends BitcoinSLogger {
         database.run(sql).map(_ => ())
     }
 
+  /** Returns flyway information about the state of migrations
+    * @see https://flywaydb.org/documentation/command/info
+    */
+  def info(): MigrationInfoService = {
```
benthecarman, line 124: We could add more tests using this in `DbManagementTest`.

benthecarman (approved):

benthecarman (approved):

Christewart:
It seems that `flyway.info()` does not return the same "initial" results for a database. When running postgres the very first migration i see is 

><< Flyway Schema Creation >>

whereas if I run on sqlite it is the first migration we have in our repo, for instance with chaindb it is 

>chain db baseline

So I have to add a `+1` for asserting on the size of postgres migrations as show here: 

https://github.com/bitcoin-s/bitcoin-s/pull/2173/commits/4400da47436586ac233b7c33989256aff240fd86#diff-0109b6d92190e19b0c7b505e73720424e816733dfcd8663b8daee91ebc4f2a94R64

db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
```diff
@@ -48,34 +49,69 @@ class DbManagementTest extends BitcoinSAsyncTest with EmbeddedPg {
...
+        assert(result == expected)
+        val flywayInfo = chainDbManagement.info()
+        assert(flywayInfo.applied().length == expected)
+        assert(flywayInfo.pending().length == 0)
+      case PostgreSQL =>
+        val expected = 4
+        assert(result == expected)
+        val flywayInfo = chainDbManagement.info()
+        //+1 for << Flyway Schema Creation >>
+        assert(flywayInfo.applied().length == expected + 1)
```
benthecarman, line 64: Maybe a better solution is to add to `DbManagement` an `appiedMigrations` function and on postgres we just call `_.tail`. Otherwise if we want to add handling for if stuff between migrations we'll have to remember that these differ as well.

Christewart: Let's see if this becomes a bigger issue, for now i think i'm going to leave it as is

benthecarman (approved):

PR #2182, opened 2020-10-13
WIP: Multi-Nonce DLCs

nkohen:
Implemented step-wise polynomial functions, collars, and CET computation.

TODO: Tests, Docs, Integration into DLCs.


dlc/src/main/scala/org/bitcoins/dlc/payouts/CETComputer.scala
```diff
@@ -0,0 +1,220 @@
+package org.bitcoins.dlc.payouts
+
+import org.bitcoins.core.currency.Satoshis
+
+object CETComputer {
```
benthecarman: nit: I like `CETCalculator` better

benthecarman (commented):
Looks like it isn't compiling

benthecarman:
Otherwise looks pretty good

benthecarman:
replaced by #2206

PR #2185, opened 2020-10-13
Update Oracle TLVs

benthecarman:
Updates the oracle tlvs to be compatible with https://github.com/discreetlogcontracts/dlcspecs/pull/55

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
@@ -231,14 +231,16 @@ object PongTLV extends TLVFactory[PongTLV] {
   }
 }
 
-sealed trait EventDescriptorTLV extends TLV
+sealed trait EventDescriptorTLV extends TLV {
+  def noncesNeeded: Int
+}
 
 object EventDescriptorTLV extends TLVParentFactory[EventDescriptorTLV] {
 
   val allFactories: Vector[TLVFactory[EventDescriptorTLV]] =
     Vector(EnumEventDescriptorV0TLV,
            RangeEventDescriptorV0TLV,
-           LargeRangeEventDescriptorV0TLV)
+           DigitDecompositionEventDescriptorV0TLV)
 
   override def typeName: String = "EventDescriptorTLV"
 }
@@ -248,20 +250,20 @@ object EventDescriptorTLV extends TLVParentFactory[EventDescriptorTLV] {
   * @param nonce The oracle's nonce they will be committing to
   * @param outcomes The set of possible outcomes
   */
-case class EnumEventDescriptorV0TLV(
-    nonce: SchnorrNonce,
-    outcomes: Vector[String])
+case class EnumEventDescriptorV0TLV(outcomes: Vector[String])
     extends EventDescriptorTLV {
   override def tpe: BigSizeUInt = EnumEventDescriptorV0TLV.tpe
 
   override val value: ByteVector = {
-    val starting = nonce.bytes ++ UInt16(outcomes.size).bytes
+    val starting = UInt16(outcomes.size).bytes
 
     outcomes.foldLeft(starting) { (accum, outcome) =>
       val outcomeBytes = CryptoUtil.serializeForHash(outcome)
       accum ++ UInt16(outcomeBytes.length).bytes ++ outcomeBytes
     }
   }
+
+  override def noncesNeeded: Int = 1
 }
 
 object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
@@ -271,7 +273,6 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
   override def fromTLVValue(value: ByteVector): EnumEventDescriptorV0TLV = {
     val iter = ValueIterator(value)
 
-    val nonce = SchnorrNonce(iter.take(32))
     val count = UInt16(iter.takeBits(16))
 
     val builder = Vector.newBuilder[String]
@@ -288,19 +289,17 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
     require(count.toInt == result.size,
             "Did not parse the expected number of outcomes")
 
-    EnumEventDescriptorV0TLV(nonce, result)
+    EnumEventDescriptorV0TLV(result)
   }
 }
 
 /**
   * Describes a simple event over a range of numbers
-  * @param nonce The oracle's nonce they will be committing to
   * @param start The first number in the range
   * @param count The number of possible outcomes
   * @param step The increment between each outcome
   */
 case class RangeEventDescriptorV0TLV(
-    nonce: SchnorrNonce,
     start: Int32,
     count: UInt32,
     step: UInt16,
@@ -314,7 +313,7 @@ case class RangeEventDescriptorV0TLV(
     val unitSize = BigSizeUInt(unit.length)
     val unitBytes = CryptoUtil.serializeForHash(unit)
 
-    nonce.bytes ++ start.bytes ++ count.bytes ++ step.bytes ++
+    start.bytes ++ count.bytes ++ step.bytes ++
       unitSize.bytes ++ unitBytes ++ precision.bytes
   }
 
@@ -327,6 +326,8 @@ case class RangeEventDescriptorV0TLV(
 
     range.map(Int32(_)).toVector
   }
+
+  override def noncesNeeded: Int = 1
 }
 
 object RangeEventDescriptorV0TLV extends TLVFactory[RangeEventDescriptorV0TLV] {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -278,16 +284,31 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
...
-    EnumEventDescriptorV0TLV(result)
+    EnumEventDescriptorV0TLV(nonce, result)
   }
 }
 
-case class RangeEventDescriptorV0TLV(start: Int32, stop: Int32, step: UInt16)
+case class RangeEventDescriptorV0TLV(
+    nonce: SchnorrNonce,
+    start: Int32,
+    count: UInt32,
```
Christewart, line 359: So why is there an explicit count required? Can't this be inferred?

Christewart: If there is documentation to these classes it would be great to link to the DLC spec so it's easy for readers of your code to find things

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -278,16 +284,31 @@ object EnumEventDescriptorV0TLV extends TLVFactory[EnumEventDescriptorV0TLV] {
...
+    step: UInt16)
```
Christewart: So what happens when `count` and `step` are conflicting? I.e.

```scala

val count = 1
val step = 5

val start = 1
val stop = 15
```

This is internally inconsistent no?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -298,19 +319,79 @@ object RangeEventDescriptorV0TLV extends TLVFactory[RangeEventDescriptorV0TLV] {
...
-    val stop = Int32(iter.takeBits(32))
+    val count = UInt32(iter.takeBits(32))
     val step = UInt16(iter.takeBits(16))
 
-    RangeEventDescriptorV0TLV(start, stop, step)
+    RangeEventDescriptorV0TLV(nonce, start, count, step)
+  }
+}
+
+case class LargeRangeEventDescriptorV0TLV(
```
Christewart: scaladoc

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -298,19 +319,79 @@ object RangeEventDescriptorV0TLV extends TLVFactory[RangeEventDescriptorV0TLV] {
...
+    base: UInt16,
+    isSigned: Boolean,
+    nonces: Vector[SchnorrNonce])
+    extends EventDescriptorTLV {
+
+  override val tpe: BigSizeUInt = LargeRangeEventDescriptorV0TLV.tpe
+
+  override val value: ByteVector = {
+    val isSignedByte = if (isSigned) ByteVector(0x01) else ByteVector(0x00)
```
Christewart: nit: Maybe factor out `0x01` and `0x00` into a constant somewhere with a nice name?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -298,19 +319,79 @@ object RangeEventDescriptorV0TLV extends TLVFactory[RangeEventDescriptorV0TLV] {
...
     val step = UInt16(iter.takeBits(16))
 
-    RangeEventDescriptorV0TLV(start, stop, step)
+    RangeEventDescriptorV0TLV(nonce, start, count, step)
+  }
+}
+
+case class LargeRangeEventDescriptorV0TLV(
+    base: UInt16,
+    isSigned: Boolean,
```
Christewart: nit: Maybe represent this as an ADT rather than a simple `Boolean` flag? The reason I think this might be useful is: 

What kind of signed number representation are you using? Twos complement? Ones complement? Some other exotic scheme?

https://en.wikipedia.org/wiki/Signed_number_representations

Christewart: I guess there needs to at least be documentation on what sort of signed number representation this actually represents?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -298,19 +319,79 @@ object RangeEventDescriptorV0TLV extends TLVFactory[RangeEventDescriptorV0TLV] {
...
+  override val tpe: BigSizeUInt = BigSizeUInt(55306)
+
+  override def fromTLVValue(
+      value: ByteVector): LargeRangeEventDescriptorV0TLV = {
+    val iter = ValueIterator(value)
+
+    val base = UInt16(iter.takeBits(16))
+    val isSignedByte = iter.take(1).head
+    val isSigned = isSignedByte match {
+      case 0x00 => false
```
Christewart: I guess these constants can be used here too -- or possibly just an ADT

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: There is no more `stop` in the spec

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: I think this would be more of a question for the spec. In the spec it is a boolean so we should do the same here

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: Can do that once it is merged

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: :man_facepalming: , i somehow put "stop" in there magically in my head :P.

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -43,6 +47,9 @@ sealed trait TLVParentFactory[T <: TLV] extends Factory[T] {
 
 object TLV extends TLVParentFactory[TLV] {
 
+  val FALSE_BYTE: Byte = 0x00
```
Christewart, line 50: Isn't this a `IS_UNSIGNED` and `IS_SIGNED`  or something like this?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: Is there a reason you are waiting? Can't we just update links in the future? It's very hard to code review since I need to manually go find links, presumably you are following the spec as currently written?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: Ok, you are more familar with the spec work than I am. Can you file an issue and link to this comment? Since you aren't documenting where you are getting this stuff from it's very hard to cross reference and file issues like this above. Where is the signed stuff?

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: Boolean can be used for other TLVs

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: It is in the description of the PR

https://github.com/discreetlogcontracts/dlcspecs/pull/55/

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: https://github.com/discreetlogcontracts/dlcspecs/pull/55/files#diff-c15525daf7798c0dd4de183850e348ffdd9aea9812ebe5bc95a1c4189122a14dR90

I don't think the different representations apply here. You don't sign the whole number, you sign the base 10 version of each digit

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: What section ? This is the only thing I see in the description of that PR 

>This is a proposal for oracle specification and mainly meant to serve as a basis for discussion. DO NOT MERGE

benthecarman: https://github.com/discreetlogcontracts/dlcspecs/pull/55/files#diff-c15525daf7798c0dd4de183850e348ffdd9aea9812ebe5bc95a1c4189122a14dR44

Christewart (approved):

PR #2191, opened 2020-10-14
Combined OracleInfo and ContractInfo…

nkohen:
… to make construct a useful interface that will work better with multi-nonce DLCs


benthecarman:
replaced by #2206

PR #2192, opened 2020-10-15
Replace DLC outcome hashes with pre-images

nkohen:
built on top of #2191


benthecarman:
You made everything a `String` but shouldn't it be a `ByteVector` bc enum outcomes will be strings but ranges and large ranges won't be

benthecarman:
replaced by #2206

PR #2197, opened 2020-10-16
Make GUI auto-update balance

benthecarman:


Christewart (approved):
So this looks ok to me, but it for other things that need to be updated periodically on the GUI will they be included in this `while` loop? I.e.

```scala
      while (true) {
        updateBalance()
        updateActiveDLCs() //made up
        updateSyncProgress() //made up
        // wait 10 seconds
        Thread.sleep(10000)
      }
```

If so, we should probably consider refactoring this some how into a more central location?

Issue #2179, opened 2020-10-12
Wallet Reorg Handling


benthecarman:
Currently, the wallet has no way of handling a reorg occurs and unconfirms a utxo.

PR #2201, opened 2020-10-19
Properly track DLC transactions in wallet

benthecarman:
Necessary for upcoming GUI work. This will have the wallet use the spk dao to track and look for the funding and closing txs

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -18,7 +18,7 @@ import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.protocol.{Bech32Address, BlockStamp}
 import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerVirtualByte}
-import org.bitcoins.core.wallet.utxo.{InputInfo, ScriptSignatureParams}
+import org.bitcoins.core.wallet.utxo._
 import org.bitcoins.crypto._
 import org.bitcoins.dlc.builder.DLCTxBuilder
 import org.bitcoins.dlc.execution._
@@ -106,6 +106,40 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
     } yield updated
   }
 
+  /** Updates the funding outpoint in the DLCDb corresponding to the given contractId */
+  private def updateFundingOutPoint(
+      contractId: ByteVector,
+      outPoint: TransactionOutPoint): Future[DLCDb] = {
+    for {
+      dlcOpt <- dlcDAO.findByContractId(contractId)
+      dlcDb <- dlcOpt match {
+        case Some(dlc) => Future.successful(dlc)
+        case None =>
+          Future.failed(
+            new IllegalArgumentException(
+              s"No DLCDb found with contractId ${contractId.toHex}"))
+      }
+      updated <- dlcDAO.update(dlcDb.updateFundingOutPoint(outPoint))
+    } yield updated
+  }
+
+  /** Updates the closing txId in the DLCDb corresponding to the given contractId */
+  private def updateClosingTxId(
+      contractId: ByteVector,
+      txId: DoubleSha256DigestBE): Future[DLCDb] = {
+    for {
+      dlcOpt <- dlcDAO.findByContractId(contractId)
+      dlcDb <- dlcOpt match {
+        case Some(dlc) => Future.successful(dlc)
+        case None =>
+          Future.failed(
+            new IllegalArgumentException(
+              s"No DLCDb found with contractId ${contractId.toHex}"))
+      }
+      updated <- dlcDAO.update(dlcDb.copy(closingTxIdOpt = Some(txId)))
+    } yield updated
+  }
+
   private def updateDLCOracleSig(
       contractId: ByteVector,
       sig: SchnorrDigitalSignature): Future[DLCDb] = {
@@ -148,6 +182,83 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
       one <- getAddress(account, HDChainType.External, index + 1)
     } yield {
       Vector(zero, one)
+    }
+  }
+
+  /** Calculates the new state of the DLCDb based on the closing transaction */
+  private def calculateAndSetState(dlcDb: DLCDb): Future[DLCDb] = {
+    (dlcDb.contractIdOpt, dlcDb.closingTxIdOpt) match {
+      case (Some(id), Some(txId)) =>
+        executorAndSetupFromDb(id).map {
+          case (_, setup) =>
+            val state =
+              if (txId == setup.refundTx.txIdBE) {
+                DLCState.Refunded
+              } else if (dlcDb.state == DLCState.Claimed) {
+                DLCState.Claimed
+              } else {
+                DLCState.RemoteClaimed
+              }
+
+            dlcDb.copy(state = state)
+        }
+      case (_, _) => Future.successful(dlcDb)
+    }
+  }
+
+  override def processIncomingUtxos(
+      tx: Transaction,
+      blockHashOpt: Option[DoubleSha256DigestBE],
+      newTags: Vector[AddressTag]): Future[Vector[SpendingInfoDb]] = {
+    super.processIncomingUtxos(tx, blockHashOpt, newTags).flatMap { res =>
+      for {
+        dlcDbs <- dlcDAO.findByFundingTxIds(Vector(tx.txIdBE))
+        _ <-
+          if (dlcDbs.nonEmpty) {
+            val txDb = TransactionDbHelper.fromTransaction(tx)
+            transactionDAO.upsert(txDb)
+          } else FutureUtil.unit
+
+        // Update the state to be confirmed or broadcasted
+        updated = dlcDbs.map { dlcDb =>
+          dlcDb.state match {
+            case DLCState.Offered | DLCState.Accepted | DLCState.Signed |
+                DLCState.Broadcasted =>
+              if (blockHashOpt.isDefined)
+                dlcDb.copy(state = DLCState.Confirmed)
+              else dlcDb.copy(state = DLCState.Broadcasted)
+            case DLCState.Confirmed | DLCState.Claimed |
+                DLCState.RemoteClaimed | DLCState.Refunded =>
+              dlcDb
+          }
+        }
+
+        _ <- dlcDAO.updateAll(updated)
+      } yield res
+    }
+  }
+
+  override def processOutgoingUtxos(
+      transaction: Transaction,
+      blockHashOpt: Option[DoubleSha256DigestBE]): Future[
+    Vector[SpendingInfoDb]] = {
+    super.processOutgoingUtxos(transaction, blockHashOpt).flatMap { res =>
+      val outPoints = transaction.inputs.map(_.previousOutput).toVector
+
+      for {
+        dlcDbs <- dlcDAO.findByFundingOutPoints(outPoints)
+        _ <-
+          if (dlcDbs.nonEmpty) {
+            val txDb = TransactionDbHelper.fromTransaction(transaction)
+            transactionDAO.upsert(txDb)
+          } else FutureUtil.unit
+
+        withTx = dlcDbs.map(_.copy(closingTxIdOpt = Some(transaction.txIdBE)))
+        updatedFs = withTx.map(calculateAndSetState)
+        updated <- Future.sequence(updatedFs)
+
+        _ <- dlcDAO.updateAll(updated)
+      } yield res
     }
   }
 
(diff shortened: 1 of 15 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -106,6 +106,38 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
     } yield updated
   }
 
+  private def updateFundingOutPoint(
```
Christewart, line 110: nit: Can you add a scaladoc? What is it updating? It looks like it updates `DLCDb` adding the outpoint to the database?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -106,6 +106,38 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+      contractId: ByteVector,
+      outPoint: TransactionOutPoint): Future[DLCDb] = {
```
Christewart, line 112: nit: Hmm... I guess I don't have a strong preference here, but do you think it's better to throw an exception rather than just returning `Future[Option[DLCDb]]`? Same applies with the `updateClosingTxId` method.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -521,7 +582,15 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
             case (input, tx) => input.toFundingInput(tx.transaction)
           })
 
-          updatedDLCDb <- updateDLCContractIds(offer, accept)
+          dlcDb <- updateDLCContractIds(offer, accept)
+
+          builder = DLCTxBuilder(offer, accept.withoutSigs)
+          fundingTx <- builder.buildFundingTx
+          outPoint = TransactionOutPoint(fundingTx.txId, UInt32.zero)
```
Christewart, line 655: just curious, why is it safe to use outpiont index `0`?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -890,7 +969,10 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
 
     executorFromDb(dlcDb, dlcOffer, dlcAccept, fundingInputs)
       .flatMap { executor =>
-        val outcomeSigs = outcomeSigDbs.map(_.toTuple).toMap
+        val outcomeSigs = outcomeSigDbs
+          .filter(_.isInitiator == !dlcDb.isInitiator)
```
Christewart, line 1040: This deserves a comment, what is the plain english explaination of what's going on here?

Christewart (changes requested):
It doesn't look like there are any test cases added to make sure we actually update these transactions through the flow of a DLC. It seems to me we need 

1. A test case that makes sure we update the funding txid 
2. A test case that makes sure we update the closing txid where we close the DLC
3. A test that makes sure we update the closing txid in the case where our remote peer closes the DLC
4. A test that makes sure we update the closing txid for the refund case

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: Currently, the funding output is always at that index

benthecarman:
> It doesn't look like there are any test cases added to make sure we actually update these transactions through the flow of a DLC. It seems to me we need
> 
>     1. A test case that makes sure we update the funding txid
> 
>     2. A test case that makes sure we update the closing txid where we close the DLC
> 
>     3. A test that makes sure we update the closing txid in the case where our remote peer closes the DLC
> 
>     4. A test that makes sure we update the closing txid for the refund case

Those are all tested with `verifyProperlySetTxIds`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: I think this is better, we are only going to be calling this function inside of other functions that should have already verified that the `contractId` exists. Making this an option will just make things unnecessarily complicated

Christewart: Ok fair enough, it sees we just return a `T` everywhere else rather than an `Option[T]` when we use the term `update`

Christewart:
> > It doesn't look like there are any test cases added to make sure we actually update these transactions through the flow of a DLC. It seems to me we need
> > ```
> > 1. A test case that makes sure we update the funding txid
> > 
> > 2. A test case that makes sure we update the closing txid where we close the DLC
> > 
> > 3. A test that makes sure we update the closing txid in the case where our remote peer closes the DLC
> > 
> > 4. A test that makes sure we update the closing txid for the refund case
> > ```
> 
> Those are all tested with `verifyProperlySetTxIds`

Ahh ok!

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -151,6 +185,83 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+    super.processOutgoingUtxos(transaction, blockHashOpt).flatMap { res =>
+      val outPoints = transaction.inputs.map(_.previousOutput).toVector
+
+      for {
+        dlcDbs <- dlcDAO.findByFundingOutPoints(outPoints)
+        _ <-
+          if (dlcDbs.nonEmpty) {
+            val txDb = TransactionDbHelper.fromTransaction(transaction)
+            transactionDAO.upsert(txDb)
+          } else FutureUtil.unit
```
nkohen: nit: Probably deserves its own util

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -988,9 +1139,11 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
   override def executeDLCRefund(contractId: ByteVector): Future[Transaction] = {
     for {
       (executor, setup) <- executorAndSetupFromDb(contractId)
+      outcome = executor.executeRefundDLC(setup)
+      _ <- processTransaction(outcome.refundTx, None)
```
nkohen: nit: Name None

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -205,6 +205,7 @@ abstract class Wallet
   override def broadcastTransaction(transaction: Transaction): Future[Unit] =
     for {
       _ <- nodeApi.broadcastTransaction(transaction)
+      _ <- processTransaction(transaction, None)
```
nkohen: nit: name None

nkohen (approved):
Looking good

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -151,6 +185,83 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+                DLCState.Refunded
+              } else if (dlcDb.state == DLCState.Claimed) {
+                DLCState.Claimed
+              } else {
+                DLCState.RemoteClaimed
+              }
+
+            dlcDb.copy(state = state)
+        }
+      case (_, _) => Future.successful(dlcDb)
```
Christewart: Please don't use `case  (_,_) =>`, enumerate the cases so we retain pattern match exhaustiveness 

```scala
case (Some(_),None) | (None, Some(_)) | (None,None) => Future.successful(dlcDb)
```

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -151,6 +185,83 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+    }
+  }
+
+  override def processIncomingUtxos(
```
Christewart: scaladoc on this please, it seems 

```
/** Process an incoming utxo as normal, and then update the DLC state for the incoming utxo */ 
```

Is that right?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -151,6 +185,83 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+            transactionDAO.upsert(txDb)
+          } else FutureUtil.unit
+
+        // Update the state to be confirmed or broadcasted
+        updated = dlcDbs.map { dlcDb =>
+          dlcDb.state match {
+            case DLCState.Offered | DLCState.Accepted | DLCState.Signed |
+                DLCState.Broadcasted =>
+              if (blockHashOpt.isDefined)
+                dlcDb.copy(state = DLCState.Confirmed)
```
Christewart, line 229: Isn't this circumventing our normal confirmation policy settings? Why isn't this the same as a the normal transaction confirmation policy we have?

Christewart: Also, whenever we get around to fixing #2179 we would also need to update this code to handle the reorg properly right?

It seems we should really try and centralize the "confirmation" logic ine one place rather than one offs.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -156,7 +156,7 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
       }
     }
 
-  private def processIncomingUtxos(
+  def processIncomingUtxos(
```
Christewart: can this be `protected`?

Christewart: but as per my suggestion earlier in this PR, i think this should be refactored into a utility method that also takes in a `WalletApi` as a parameter so something like this


```scala
def processIncomingUtxos(
      transaction: Transaction,
      blockHashOpt: Option[DoubleSha256DigestBE],
      newTags: Vector[AddressTag],
      wallet: WalletApi): Future[Vector[SpendingInfoDb]] = ...
```

That way we can call it inside of `DLCWallet` as well? Does that work?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -438,18 +438,26 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
 
         for {
           _ <- insertIncomingTransaction(transaction, totalIncoming)
+
+          addrs <- addressDAO.findAllAddresses()
```
Christewart, line 445: Hmm, why is this needed now?

Christewart (changes requested):

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: This isn't the TxoState, this is the DLCState.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: To differentiate the spks that belong to the wallet and the ones that don't. For example, we don't want to add the DLC funding output to our spending info db as it will effect our balance, and we could not spend it.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: We can't do this, this is an implementation specific function. If we used a `WalletApi` then we wouldn't have access to our DAOs

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
Christewart: They are inherently linked. What happens in the case of reorg? Or is `DLCState.Confirmed` just named badly? It seems you are setting this based off of a blockhash, so it does have some relationship to actually being confirmed in the chain.

benthecarman: Oh yes, in that context it has the same problem. Was more so responding to your first comment

Christewart: Hmm, so does this expand the scope of #2179 then? Or is it the same problem?

Christewart (approved):

PR #2204, opened 2020-10-20
Only Validate PSBT BIP143 vulnerability on signing

benthecarman:
This changes the PSBT to only validate if it is vulnerable to the BIP143 vulnerability when we are actually signing. This has caused some problems where we would constructing a PSBT a adding things in the wrong error can cause the requirement to fail even though it would otherwise be a safe PSBT.

This also makes it so we can parse any valid PSBT which should be nice for other libraries

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
+++ b/core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
@@ -71,9 +71,10 @@ case class PSBT(
   def version: UInt32 = globalMap.version.version
 
   def validateBIP143Vulnerability: PSBT = {
-    require(isFinalized || inputMaps.size == 1 || inputMaps.forall(
-              !_.isBIP143Vulnerable),
-            "One or more of the input maps are susceptible to the BIP 143")
+    require(
+      isFinalized || inputMaps.size == 1 || inputMaps.forall(
+        !_.isBIP143Vulnerable),
+      "One or more of the input maps are susceptible to the BIP 143 vulnerability")
 
     this
   }
@@ -98,6 +99,10 @@ case class PSBT(
       .map { case (output, otherOutput) => output.combine(otherOutput) }
 
     PSBT(global, inputs, outputs)
+  }
+
+  def combinePSBTAndValidate(other: PSBT): PSBT = {
+    combinePSBT(other).validateBIP143Vulnerability
   }
 
   def finalizeInput(index: Int): Try[PSBT] = {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -74,6 +70,14 @@ case class PSBT(
 
   def version: UInt32 = globalMap.version.version
 
+  def validateBIP143Vulnerability: PSBT = {
```
Christewart, line 73: nit: Might as well returning something like a `Try` in this case?

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -74,6 +70,14 @@ case class PSBT(
...
+    require(isFinalized || inputMaps.size == 1 || inputMaps.forall(
```
Christewart: So isn't the whole point of the vuln is that you cannot verify -- without the full utxo set -- that the PSBT is not paying a large fee? Don't we create a new instance of `PSBT` everytime an action is taken so we can't really tell if we are internally handling PSBT that might be safe, or receiving one from an external data source? 

How are retaining any sort of protection from this vulnerability with this PR?

benthecarman: > How are retaining any sort of protection from this vulnerability with this PR?

You are only vulnerable when you are signing so this moves the check to when we sign.  You aren't vulnerable to anything for just having a psbt with a witness utxo.

benthecarman: Otherwise this creates annoyances where you have to add things in a certain order. For example, if I already have signature for an input and want to add them and the witness utxo. If I add the witness utxo first, I will get an error because then the invariant will fail, even though it was 100% safe.

core/src/main/scala/org/bitcoins/core/psbt/PSBT.scala
```diff
@@ -96,6 +101,10 @@ case class PSBT(
     PSBT(global, inputs, outputs)
   }
 
+  def combinePSBTAndValidate(other: PSBT): PSBT = {
```
Christewart, line 104: So we don't seem to actually use this anywhere? I guess that's the tough thing about creating a parallel method, we don't actually check anything anywhere because this is completely unused?

benthecarman: We shouldn't need to use it because we aren't passing PSBTs outside the library for signing. We are safe because our sign function does the check

Christewart: >sign function does the check

Where? BIP143 is vulnerable itself

benthecarman: The PSBT sign function checks that we are not vulnerable.

https://github.com/bitcoin-s/bitcoin-s/pull/2204/files#diff-0f7b76bc99b643f00b04d5214c76cd8bed7fa19db163b91a934f9872b00552b8R167

Christewart (approved):

PR #2213, opened 2020-10-25
Remove Unneeded Escaped param for DLC Cli

benthecarman:
Same as #2178 but for the DLC branch and its exclusive cli commands

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
@@ -178,9 +178,9 @@ object DLCMessage {
         fundingInputs.map { input =>
           val obj = mutable.LinkedHashMap(
             "prevTx" -> Str(input.prevTx.hex),
-            "prevTxVout" -> Num(input.prevTxVout.toInt),
-            "sequence" -> Num(input.sequence.toInt),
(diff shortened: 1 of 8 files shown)


app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -178,9 +178,9 @@ object DLCMessage {
         fundingInputs.map { input =>
           val obj = mutable.LinkedHashMap(
             "prevTx" -> Str(input.prevTx.hex),
-            "prevTxVout" -> Num(input.prevTxVout.toInt),
-            "sequence" -> Num(input.sequence.toInt),
-            "maxWitnessLength" -> Num(input.maxWitnessLen.toInt)
+            "prevTxVout" -> Num(input.prevTxVout.toLong.toDouble),
```
Christewart, line 181: Wait so why is this needed?

Christewart (approved):

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
benthecarman: `toInt` wasn't big enough to cover for all `UInt32` and and `Num()` can't take a long, so had to a double

PR #2214, opened 2020-10-26
Use compact signatures for refund sig in TLVs

benthecarman:


nkohen (approved):

benthecarman:
Was included in #2215

PR #2217, opened 2020-10-27
Make aesPassword option for wallet config

### LINKED ISSUE #2012: bitcoin-s.keymanager.aespassword
This proposes to add a new configuration option for specifying the password for decrypting your mnemonic seed on disk

This is related to #972 

This allows you to use environment variables as well to set the the password, for instance you can do 

```
bitcoin-s { 
  key-manager { 
    mnemonic-password=${MNEMONIC_PASSWORD}
  } 
}
```

Maybe this should be done for the BIP39 password too? :man_shrugging:

benthecarman:
Closes #2012

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerCreateApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerCreateApi.scala
@@ -30,10 +30,10 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * $initialize
     */
   final def initialize(
-      aesPassword: AesPassword,
+      aesPasswordOpt: Option[AesPassword],
       kmParams: KeyManagerParams,
       bip39PasswordOpt: Option[String]): Either[KeyManagerInitializeError, T] =
-    initializeWithEntropy(aesPassword = aesPassword,
+    initializeWithEntropy(aesPasswordOpt = aesPasswordOpt,
                           entropy = MnemonicCode.getEntropy256Bits,
                           bip39PasswordOpt = bip39PasswordOpt,
                           kmParams = kmParams)
@@ -42,7 +42,7 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * $initializeWithEnt
     */
   def initializeWithEntropy(
-      aesPassword: AesPassword,
+      aesPasswordOpt: Option[AesPassword],
       entropy: BitVector,
       bip39PasswordOpt: Option[String],
       kmParams: KeyManagerParams): Either[KeyManagerInitializeError, T]
@@ -55,12 +55,12 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * @return
     */
   final def initializeWithMnemonic(
-      aesPassword: AesPassword,
+      aesPasswordOpt: Option[AesPassword],
       mnemonicCode: MnemonicCode,
       bip39PasswordOpt: Option[String],
       kmParams: KeyManagerParams): Either[KeyManagerInitializeError, T] = {
     val entropy = mnemonicCode.toEntropy
-    initializeWithEntropy(aesPassword = aesPassword,
+    initializeWithEntropy(aesPasswordOpt = aesPasswordOpt,
                           entropy = entropy,
                           bip39PasswordOpt = bip39PasswordOpt,
                           kmParams = kmParams)
--- a/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerCreateApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerCreateApi.scala
@@ -5,6 +5,7 @@ import org.bitcoins.core.wallet.keymanagement.{
   KeyManagerInitializeError,
   KeyManagerParams
 }
+import org.bitcoins.crypto.AesPassword
 import scodec.bits.BitVector
 
 trait KeyManagerCreateApi
@@ -29,9 +30,11 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * $initialize
     */
   final def initialize(
+      aesPassword: AesPassword,
       kmParams: KeyManagerParams,
       bip39PasswordOpt: Option[String]): Either[KeyManagerInitializeError, T] =
-    initializeWithEntropy(entropy = MnemonicCode.getEntropy256Bits,
+    initializeWithEntropy(aesPassword = aesPassword,
+                          entropy = MnemonicCode.getEntropy256Bits,
                           bip39PasswordOpt = bip39PasswordOpt,
                           kmParams = kmParams)
 
@@ -39,23 +42,26 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * $initializeWithEnt
     */
   def initializeWithEntropy(
+      aesPassword: AesPassword,
       entropy: BitVector,
       bip39PasswordOpt: Option[String],
       kmParams: KeyManagerParams): Either[KeyManagerInitializeError, T]
 
   /**
-    * Helper method to initialize a [[KeyManagerCreate$ KeyManager]] with a [[MnemonicCode MnemonicCode]]
+    * Helper method to initialize a [[KeyManagerApi KeyManager]] with a [[MnemonicCode MnemonicCode]]
     *
     * @param mnemonicCode
     * @param kmParams
     * @return
     */
   final def initializeWithMnemonic(
+      aesPassword: AesPassword,
       mnemonicCode: MnemonicCode,
       bip39PasswordOpt: Option[String],
       kmParams: KeyManagerParams): Either[KeyManagerInitializeError, T] = {
     val entropy = mnemonicCode.toEntropy
-    initializeWithEntropy(entropy = entropy,
+    initializeWithEntropy(aesPassword = aesPassword,
+                          entropy = entropy,
                           bip39PasswordOpt = bip39PasswordOpt,
                           kmParams = kmParams)
   }
--- a/core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
@@ -58,4 +58,18 @@ object BIP39Seed extends Factory[BIP39Seed] {
     BIP39Seed.fromBytes(ByteVector(encodedBytes))
   }
 
+  /**
+    * Generates a [[https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki BIP32]]
+    * seed from a mnemonic code. An optional password can be supplied.
+    */
+  def fromMnemonic(
+      mnemonic: MnemonicCode,
+      passwordOpt: Option[String]): BIP39Seed = {
+    passwordOpt match {
+      case Some(pass) =>
+        fromMnemonic(mnemonic, pass)
+      case None =>
+        fromMnemonic(mnemonic)
+    }
+  }
 }
--- a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
+++ b/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
@@ -220,19 +220,22 @@ object DLCOracle {
 
   def apply(
       mnemonicCode: MnemonicCode,
-      password: AesPassword,
+      passwordOpt: Option[AesPassword],
       bip39PasswordOpt: Option[String] = None)(implicit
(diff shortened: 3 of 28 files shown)


nkohen (commented):
LGTM but I think this is missing documentation for what needs to be added to bitcoin-s.conf

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -97,6 +98,12 @@ case class WalletAppConfig(
     confs
   }
 
+  lazy val aesPassword: AesPassword = {
```
Christewart: So I think we should default to not encrypting the data on disk if the password is not specified rather than using `BIP39KeyManager.badPassphrase`. 

We will need some sort of shim for old wallets that do use `BIP39KeyManager.badPassphrase` to encrypt their wallet, but I don't think we should continue this?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -732,10 +732,11 @@ object Wallet extends WalletLogger {
   def initialize(wallet: Wallet, bip39PasswordOpt: Option[String])(implicit
       walletAppConfig: WalletAppConfig,
       ec: ExecutionContext): Future[Wallet] = {
+    val password = walletAppConfig.aesPassword
```
Christewart: So why allow `bip39PasswordOpt` as a parameter, but not `aesPassword`? It seems we should have this precedent

1. The parameter passed in (think a user text box in GUI, or a command line parameter)
2. Read from the config file like you are doing here

I think this would follow the typesafe config resolution precedent as documented [here](https://github.com/lightbend/config#standard-behavior)


We should eventually  `bip39Password` to be in the wallet config file too.

Thinking about this further, we probably need to get issues filed for changing the `aesPassword` (easy, i think) and then `bip39Password` (complex and involves creating a new wallet, i think?)

core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerCreateApi.scala
```diff
@@ -29,33 +30,38 @@ trait BIP39KeyManagerCreateApi[T <: BIP39KeyManagerApi]
     * $initialize
     */
   final def initialize(
+      aesPassword: AesPassword,
```
Christewart: I think we should allow this to be an `aesPasswordOpt: Option[AesPassword]`  ?

The reason I say this is if you have a bip39Password, your seed stored unencrypted on disk is ok as you can't do anything with just the unencryptd seed.

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
```diff
@@ -58,4 +58,18 @@ object BIP39Seed extends Factory[BIP39Seed] {
     BIP39Seed.fromBytes(ByteVector(encodedBytes))
   }
 
+  /**
+    * Generates a [[https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki BIP32]]
+    * seed from a mnemonic code. An optional password can be supplied.
+    */
+  def fromMnemonic(
```
Christewart, line 65: Do we need the other `fromMnemonic()` defined in this object?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -232,7 +232,7 @@ object DLCOracle {
     val key =
       WalletStorage.getPrivateKeyFromDisk(conf.seedPath,
                                           SegWitMainNetPriv,
-                                          password,
+                                          Some(password),
```
Christewart: So we are requiring a password here? It doesn't seem we modified `passwordOpt: Option[AesPassword]` as a parameter to this

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
```diff
@@ -104,7 +104,7 @@ case class DLCOracleAppConfig(
     val key =
       WalletStorage.getPrivateKeyFromDisk(seedPath,
                                           SegWitMainNetPriv,
-                                          password,
+                                          Some(password),
```
Christewart: same here, is there are ason the parameter to `initialize()` is not a `passwordOpt: Option[AesPassword]` ?

key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
```diff
@@ -9,9 +9,12 @@ import scodec.bits.ByteVector
 
 import scala.util.{Failure, Success, Try}
 
-case class DecryptedMnemonic(
-    mnemonicCode: MnemonicCode,
-    creationTime: Instant) {
+sealed trait WritableMnemonic {
```
Christewart: nit: Maybe rename this to `MnemonicState` ? I agree that both `DecryptedMnemonic` and `EncryptedMnemonic` are writable, but I think this class is more powerful -- it represents the state of the Mnemonic too. 

I.e. you can decrypt the `EncryptedMnemonic` to get a `DecryptedMnemonic` too presumably?

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -166,28 +208,94 @@ object WalletStorage {
...
+      }
+    }
+
+    import MnemonicJsonKeys._
+    import ReadMnemonicError._
+
+    val readJsonTupleEither: CompatEither[
+      ReadMnemonicError,
+      (Vector[String], Long)] = jsonE.flatMap { json =>
+      logger.trace(s"Read mnemonic JSON: $json")
```
Christewart: Will this log the unencrypted seed if the `TRACE` log level is set? If so, we should remove it. We don't want the mnemonic seed accidentally getting caught in logs, especially since this is unencrypted.

key-manager-test/src/test/scala/org/bitcoins/keymanager/bip39/BIP39KeyManagerApiTest.scala
```diff
@@ -46,21 +46,23 @@ class BIP39KeyManagerApiTest extends KeyManagerApiUnitTest {
 
   it must "initialize the key manager" in {
     val entropy = MnemonicCode.getEntropy256Bits
-    val keyManager = withInitializedKeyManager(entropy = entropy)
+    val aesPasswordOpt = KeyManagerTestUtil.aesPasswordOpt
```
Christewart, line 49: nit: So I think I would _prefer_ explicit test cases for the combinations of `aesPassword` being defined and `aesPassword == None`. 

1. It makes it easier to debug if we do introduce a regression
2. We now have determinism in these test cases, since `KeyManagerTestUtil.aesPasswordOpt` can be `Some` or `None`, it will only fail somes times if we introduce a regression.
3. This is a sensitive part of the code base, so i think it's best to be explicit as possible with the tests.

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
```diff
@@ -106,47 +155,93 @@ class WalletStorageTest extends BitcoinSWalletTest with BeforeAndAfterEach {
...
 
-  it must "fail to read a mnemonic that has bad JSON in it" in { walletConf =>
-    val badJson =
-      """
-        | {
-        |   "iv":"ba7722683dad8067df8d069ee04530cc",
-        |   "cipherText":,
-        |   "salt":"2b7e7d718139518070a87fbbda03ea33cdcda83b555020e9344774e6e7d08af2"
-        | }
+  it must "fail to read an encrypted mnemonic with bad password" in {
```
Christewart: nit: Can you be more specific with what passphrase is bad? I.e. `AesPassword` or bip39 password

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
```diff
@@ -224,4 +342,33 @@ class WalletStorageTest extends BitcoinSWalletTest with BeforeAndAfterEach {
...
+  it must "write and read an unencrypted ExtPrivateKey from disk" in {
+    walletConf: WalletAppConfig =>
+      assert(!walletConf.seedExists())
+      assert(!walletConf.seedExists())
+      val mnemonicCode = CryptoGenerators.mnemonicCode.sampleSome
+      val writtenMnemonic = DecryptedMnemonic(mnemonicCode, TimeUtil.now)
+      val seedPath = getSeedPath(walletConf)
+      WalletStorage.writeMnemonicToDisk(seedPath, writtenMnemonic)
+
+      val password = getBIP39PasswordOpt().getOrElse(BIP39Seed.EMPTY_PASSWORD)
```
Christewart, line 362: This seems very suspicious. What's going on here? This is also used above. Is there a reason we don't just use `None`? And then inside of the `BIP39KeyManager` convert `None -> BIP39Seed.EMPTY_PASSWORD`?

wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
```diff
@@ -136,16 +136,26 @@ class WalletUnitTest extends BitcoinSWalletTest {
...
         case Right(_)  => fail("Unlocked wallet with bad password!")
         case Left(err) => err
       }
       errorType match {
-        case KeyManagerUnlockError.MnemonicNotFound          => fail(MnemonicNotFound)
-        case KeyManagerUnlockError.BadPassword               => succeed
-        case KeyManagerUnlockError.JsonParsingError(message) => fail(message)
+        case KeyManagerUnlockError.MnemonicNotFound => fail(MnemonicNotFound)
+        case KeyManagerUnlockError.BadPassword      =>
+          // If wallet is unencrypted then we shouldn't get a bad password error
```
Christewart, line 148: Ok this comment is confusing to me, "we shouldn't get a bad password error" but the pattern match this is part of indicates there was a bad password? I.e. `KeyManagerUnlockError.BadPassword`

wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
```diff
@@ -136,16 +136,26 @@ class WalletUnitTest extends BitcoinSWalletTest {
...
+          wallet.walletConfig.aesPasswordOpt match {
+            case Some(_) => succeed
+            case None    => fail()
+          }
+        case KeyManagerUnlockError.JsonParsingError(message) =>
+          // If wallet is encrypted then we shouldn't get a json parsing error
```
Christewart, line 154: same thing applies here, `... shouldn't get a json parsing error` when we are pattern matching for the `JsonParsingError`

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/crypto/BIP39Seed.scala
benthecarman: There already is another

wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
benthecarman: We shouldn't get a `BadPassword` error when the wallet doesn't use an aes password. But if we do have an aes password then we should get the error, the match on the `aesPasswordOpt` handles this

wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
benthecarman: We shouldn't get a `JsonParsingError` error when the wallet uses an aes password. But if we don't have an aes password then we should get the error, the match on the aesPasswordOpt handles this

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
benthecarman: This is fixed in #2234

key-manager-test/src/test/scala/org/bitcoins/keymanager/bip39/BIP39KeyManagerApiTest.scala
benthecarman: I think it is okay as is in this case because the aes password is only used when reading and writing from the disk. All the possible cases are handled in the `WalletStorageTest` and then the other key manager tests shouldn't care if it reading from an encrypted or unencrypted seed

Christewart (approved):

Issue #2216, opened 2020-10-27
Backwards Witness Stacks


nkohen:
Witness stacks are stored backwards!!! This is terrible.

Currently `reverse` is only called on these SPK-style stacks in the `bytes` method meaning anyone who was trying to use the `stack` member to access elements of the witness stack (generally me) was being bamboozled and given the reverse of what was expected.

Christewart:
I'm not necessarily against changing this, but witnesses are serialized backwards because they are "evaluated" stacks

The easiest way to see this is probably by looking at `P2SHScriptSignature.redeemScript` vs `P2WSHWitnessV0.redeemScript`

https://github.com/bitcoin-s/bitcoin-s/blob/4186be51e00fafadcc825899284d84ecacce1676/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala#L163-L162

https://github.com/bitcoin-s/bitcoin-s/blob/8f86c180760bc926c74b943d075fe5c120e96837/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala#L99-L98


Since a witness isn't actually evaluated by the script interpreter (rather just _literally_ dumped onto the stack) this is why things are reversed https://github.com/bitcoin-s/bitcoin-s/blob/4186be51e00fafadcc825899284d84ecacce1676/core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala#L485-L484

https://github.com/bitcoin/bitcoin/blob/3f512f3d563954547061ee743648b57a900cbe04/src/script/interpreter.cpp#L1857


With a script signature you actually push every single element onto the stack inside the main evaluation loop in interpreter.

PR #2218, opened 2020-10-27
Fixed backward stack usage in dlc things

nkohen:
This is due to #2216 and can be reverted when that is fixed.


app/server/src/main/scala/org/bitcoins/server/SerializedTransaction.scala
```diff
@@ -95,7 +95,8 @@ object SerializedTransaction {
                     script = Some(p2wsh.redeemScript.asm.toVector),
                     pubKey = None,
                     signature = None,
-                    stack = Some(p2wsh.stack.toVector.tail)))
+                    stack = Some(p2wsh.stack.toVector.reverse.init)
```
benthecarman, line 98: nit, could have done `tail.reverse`

benthecarman (approved):

PR #2219, opened 2020-10-29
Fee Provider from config

### LINKED ISSUE #1754: Add ability to configure fee providers from bitcoin-s.conf
https://github.com/bitcoin-s/bitcoin-s/pull/1751#issuecomment-668722157

I think at minimum it should be 

```
bitcoin-s { 
  fee-provider { 
     name = mempool-space
     max-rate = 100satvbyte 
     absolute-max-fee = 1BTC //no more than 1BTC of fee
  }
}
```

Any other useful config options that will be available for all fee providers (the weird one to think about is the `ConstantFeeProvider`)

benthecarman:
Closes #1754

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -322,9 +322,8 @@ class BitcoinSServerMain(override val args: Array[String])
           BitGoFeeRateProvider(targetOpt)
         case (Some(MempoolSpace), None) =>
           MempoolSpaceProvider(HourFeeTarget)
-        case (Some(MempoolSpace), Some(blockTarget)) =>
-          val target = MempoolSpaceTarget.fromBlockTarget(blockTarget)
-          MempoolSpaceProvider(target)
+        case (Some(MempoolSpace), Some(target)) =>
+          MempoolSpaceProvider.fromBlockTarget(target)
         case (Some(Constant), Some(num)) =>
           ConstantFeeRateProvider(SatoshisPerVirtualByte.fromLong(num))
         case (Some(Constant), None) =>
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -302,7 +302,10 @@ class BitcoinSServerMain(override val args: Array[String])
     server.start()
   }
 
-  def getFeeProviderOrElse(default: FeeRateApi)(implicit
+  /** Gets a Fee Provider from the given wallet app config
+    * Returns default if there is no config set
+    */
+  def getFeeProviderOrElse(default: => FeeRateApi)(implicit
       system: ActorSystem,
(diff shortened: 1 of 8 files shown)


docs/config/configuration.md
```diff
@@ -158,6 +158,32 @@ bitcoin-s {
...
+    # Bitcoin-S provides manny different fee providers
+    # You can configure your server to use any of them
+    # Below is some examples of different options
+    fee-provider {
+        # name = mempoolspace # Uses mempool.space's api
+        # Mempool.Space has 3 different modes
+        # 0: Hour Fee Target
+        # 1: Half Hour Fee Target
+        # 2: Fastest Fee Target
+        # target = 0
```
Christewart: Hmm, this "target" thing seems like it will be a problem as there is no common target across fee providers? 

What do you think about requiring a "normalization" method in the `FeeRateApi` trait to normalize the number to the next `n` blocks? I.e. `1 block` or `2 blocks` .. etc.  I think this is what bitcoin core does?

fee-provider/src/main/scala/org/bitcoins/feeprovider/MempoolSpaceProvider.scala
```diff
@@ -49,4 +49,14 @@ object MempoolSpaceTarget {
   final case object HalfHourFeeTarget extends MempoolSpaceTarget
 
   final case object HourFeeTarget extends MempoolSpaceTarget
+
+  def fromInt(int: Int): MempoolSpaceTarget = {
```
Christewart: Not necessarily related to this PR, but is this even a good fee provider if we can't target anything over 1 hour?

Especially given the mempool is currently very full?

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -298,6 +301,37 @@ class BitcoinSServerMain(override val args: Array[String])
     }
     server.start()
   }
+
+  def getFeeProviderOrElse(default: FeeRateApi)(implicit
```
Christewart: Can you add a scaladoc?

Also generally when `OrElse()` is used, you can provide in a function to be evaluated when the first case fails. 

Take a look at `Option.getOrElse()` 's signature 

```scala
final def getOrElse[B >: A](default: => B)
```

Christewart (changes requested):

fee-provider/src/main/scala/org/bitcoins/feeprovider/BitcoinerLiveFeeRateProvider.scala
```diff
@@ -34,3 +32,22 @@ case class BitcoinerLiveFeeRateProvider(minutes: Int)(implicit
     }
   }
 }
+
+object BitcoinerLiveFeeRateProvider {
+
+  final val validMinutes =
+    Vector(30, 60, 120, 180, 360, 720, 1440)
+
+  def fromBlockTarget(blocks: Int)(implicit
```
Christewart: Let's make 

```scala

trait FeeRateProviderFactory[T] { 


def fromBlockTarget(blocks: Int)(implicit ...): T

}
```

Christewart (approved):

PR #2225, opened 2020-10-30
DLCMessage Json

nkohen:
Made test for DLCMessage Json serialization symmetry

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/TLVGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/TLVGen.scala
@@ -100,7 +100,7 @@ trait TLVGen {
       outcomes <- Gen.listOfN(numOutcomes, CryptoGenerators.sha256Digest)
       totalInput <-
         Gen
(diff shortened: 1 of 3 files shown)


testkit/src/main/scala/org/bitcoins/testkit/core/gen/TLVGen.scala
```diff
@@ -100,7 +100,7 @@ trait TLVGen {
       outcomes <- Gen.listOfN(numOutcomes, CryptoGenerators.sha256Digest)
       totalInput <-
         Gen
-          .choose(numOutcomes + 1, Long.MaxValue)
+          .choose(numOutcomes + 1, Long.MaxValue / 10000L)
           .map(Satoshis.apply)
```
benthecarman, line 104: Why not use `CurrencyUnitGenerator.positiveRealistic`

nkohen: I suppose I could but this tests even more broadly as we want to make sure DLCs are resilient even to really large things (maybe even larger than the realistic bound)

benthecarman (approved):

PR #2227, opened 2020-10-31
2020 10 31 Fix indeterminism with pairing dlc funding puts with funding transactions

Christewart:
Previously we were assuming that the ordering was correct for inputs and their funding transactions. We would just use `.zip()` on the two vectors pairing inputs with transactions. When it came time to `DLCFundingInput` it would fails this invariant 

https://github.com/bitcoin-s/bitcoin-s/blob/1d1efaf90a553d935e1657a7f9149dda916edf8e/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCFundingInputDb.scala#L23

since the outpoint and the funding transaction were not paired together correctly in some cases. 

Now we just match things up by txids and throw if there is an input that does not have a matching previous transaction.

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -189,9 +189,7 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
             prevTxs <-
               transactionDAO.findByTxIdBEs(fundingInputs.map(_.outPoint.txIdBE))
           } yield {
-            val inputRefs = fundingInputs.zip(prevTxs).map {
-              case (input, tx) => input.toFundingInput(tx.transaction)
-            }
+            val inputRefs = matchPrevTxsWithInputs(fundingInputs, prevTxs)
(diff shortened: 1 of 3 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -776,12 +776,13 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
-        case (input, tx) => input.toFundingInput(tx.transaction)
-      }
-      val remoteFundingInputs = remoteDbFundingInputs.zip(remotePrevTxs).map {
-        case (input, tx) => input.toFundingInput(tx.transaction)
-      }
+      val localFundingInputs = matchPrevTxsWithInputs(inputs =
+                                                        localDbFundingInputs,
+                                                      prevTxs = localPrevTxs)
+
+      val remoteFundingInputs = matchPrevTxsWithInputs(inputs =
```
benthecarman, line 776: Need to do this too for `verifierFromAccept` I believe

benthecarman (changes requested):

benthecarman (approved):

PR #2230, opened 2020-11-01
Move newer functions to common rpc interface

benthecarman:
This is useful because instead of having to make a bitcoind v19 or v20 client, you can make a general bitcoind client and still be able to call the functions, they will just fail on an older version.


Christewart (approved):

Christewart:
needs rebase

PR #2231, opened 2020-11-02
Bitcoind RPC Multi-wallet support 

benthecarman:
build on top of #2230 and #2229 

Allows you to use bitcoind when you have more than just the default wallet

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
@@ -48,6 +48,9 @@ trait Client extends BitcoinSLogger with StartStopAsync[BitcoindRpcClient] {
   def version: BitcoindVersion
   protected val instance: BitcoindInstance
 
+  protected def walletExtension(walletName: String): String =
+    s"/wallet/$walletName"
+
   /**
     * The log file of the Bitcoin Core daemon
     */
@@ -250,9 +253,12 @@ trait Client extends BitcoinSLogger with StartStopAsync[BitcoindRpcClient] {
   protected def bitcoindCall[T](
       command: String,
       parameters: List[JsValue] = List.empty,
-      printError: Boolean = true)(implicit reader: Reads[T]): Future[T] = {
+      printError: Boolean = true,
+      uriExtensionOpt: Option[String] = None
+  )(implicit reader: Reads[T]): Future[T] = {
 
-    val request = buildRequest(instance, command, JsArray(parameters))
+    val request =
+      buildRequest(instance, command, JsArray(parameters), uriExtensionOpt)
     val responseF = sendRequest(request)
 
     val payloadF: Future[JsValue] =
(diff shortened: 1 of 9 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -48,6 +48,9 @@ trait Client extends BitcoinSLogger with StartStopAsync[BitcoindRpcClient] {
   def version: BitcoindVersion
   protected val instance: BitcoindInstance
 
+  protected def walletExtension(walletName: String): String =
+    s"/wallet/$walletName"
```
Christewart, line 52: Should this pattern match on the version? I think it's probably a good idea in case extensions change in the future, OR we can throw an exception here if it's an older version of bitcoind that doesn't support multi wallet?

Christewart: Actually thinking about this more, maybe we want to just incorporate the wallet  name inside of `BitcoindRpcAppConfig`?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
```diff
@@ -264,14 +419,60 @@ trait WalletRpc { self: Client =>
...
+    bitcoindCall[DoubleSha256DigestBE](
+      "sendtoaddress",
+      List(Json.toJson(address),
+           Json.toJson(Bitcoins(amount.satoshis)),
+           JsString(localComment),
+           JsString(toComment),
+           JsBoolean(subractFeeFromAmount)),
+      uriExtensionOpt = walletNameOpt.map(walletExtension)
+    )
+  }
```
Christewart, line 477: So just looking through this file, if we were to move the wallet name into `BitcoindRpcAppConfig` could we avoid all of the duplication of methods? 

Basically all of these would now just look inside of `BitcoindRpcAppConfig` and get the wallet name if it's defined.

Another reason this is probably a good idea is that authentication and having access to a wallet is inherently tied together, so with the `username`/`password` combo in `BitcoindRpcAppConfig` that should authenticate you with a specific wallet which should also be defined in `BitciondRpcAppConfig`?

Christewart (changes requested):

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
benthecarman: I thought about doing it this way but the problem is that we then need a `BitcoindInstance` and thus client for every wallet we want to use which could get really annoying.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/WalletRpc.scala
benthecarman: Bitcoind's rpc auth isn't tied to each wallet so I don't see the benefit there.

Christewart (commented):
So I'm fine with this, but we should add an example to the website of how to actually use this

Christewart (approved):

PR #2233, opened 2020-11-02
2020 11 02 cleanup

Christewart:
Clean up a bunch of things intellij gives compiler warnings for in `core-test` when trying to execute the test suite from intellij


chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -10,22 +10,21 @@ import org.bitcoins.core.number.{Int32, UInt32}
 import org.bitcoins.core.p2p.CompactFilterMessage
 import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.core.protocol.blockchain.BlockHeader
-import org.bitcoins.core.util.{FutureUtil, TimeUtil}
+import org.bitcoins.core.util.{TimeUtil}
```
benthecarman: I think having a single item with `{}` will cause a warning, should remove them, doing ctrl + alt + o, will automatically do this. Could be worth running optimize imports on the whole project if we are doing a cleanup

chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala
```diff
@@ -10,22 +10,21 @@ import org.bitcoins.core.number.{Int32, UInt32}
...
 import org.bitcoins.testkit.chain.{
   BlockHeaderHelper,
   ChainDbUnitTest,
   ChainTestUtil,
   ChainUnitTest
 }
-import org.bitcoins.testkit.util.{FileUtil, ScalaTestUtil}
-import org.scalatest.{Assertion, FutureOutcome}
+import org.bitcoins.testkit.util.FileUtil
+import org.scalatest.{FutureOutcome}
```
benthecarman: ditto

chain-test/src/test/scala/org/bitcoins/chain/models/CompactFilterHeaderDAOTest.scala
```diff
@@ -1,17 +1,12 @@
...
   ChainDbUnitTest,
-  ChainTestUtil,
-  ChainUnitTest
-}
-import org.bitcoins.testkit.core.gen.{
-  BlockchainElementsGenerator,
-  CryptoGenerators
+  ChainTestUtil
 }
+import org.bitcoins.testkit.core.gen.{CryptoGenerators}
```
benthecarman: ditto

project/Deps.scala
```diff
@@ -203,6 +204,9 @@ object Deps {
     val scalameter =
       "com.storm-enroute" %% "scalameter" % V.scalameterV % "test" withSources () withJavadoc ()
 
+    val scalaCollectionCompat =
+      "org.scala-lang.modules" %% "scala-collection-compat" % V.scalaCollectionCompatV
```
benthecarman, line 208: what is this for?

Christewart: This is needed to allow `@nowarn` usage on Scala 2.12.x . The annotation was added in the scala std library in the 2.13.2 release.

benthecarman (approved):

PR #2242, opened 2020-11-06
Use PSBT type in bitcoind calls

benthecarman:
Built on top of #2231, just look at last commit

Previously we were parsing all PSBTs as strings, this fixes it to use the actual PSBT type as well give it access to all the bitcoind clients, not just v17.


Christewart:
rebase needed

Christewart (approved):

PR #2247, opened 2020-11-06
Add listreservedutxos cli command

### LINKED ISSUE #2244: Make `listreservedutxos` rpc command
This is useful for just sanity checks to make sure you don't have reserved utxos you have forgotten about (i have these). 

If I understand correctly, these reserved UTXOs could effect your wallet balance. Also a nice to have would be a timestamp of when the utxo is reserved, but I'm not sure if that is hard to do. 

https://bitcoin-s.org/docs/next/applications/server#wallet

benthecarman:
Closes #2244


Christewart (approved):

Christewart:
Works like a charm!

PR #2252, opened 2020-11-09
Use same config option for key manager projects

benthecarman:


dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
```diff
@@ -73,10 +73,14 @@ case class DLCOracleAppConfig(
   }
 
   lazy val aesPasswordOpt: Option[AesPassword] = {
```
Christewart: Is there a reason we don't just create a `KeyManagerAppConfig` and make `DLCOracleAppConfig` have-a `KeyManagerAppConfig`?

Same thing applies to the `WalletAppConfig`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
```diff
@@ -114,7 +114,8 @@ case class DLCOracleAppConfig(
                                           aesPasswordOpt,
                                           bip39PasswordOpt)
     val oracle = DLCOracle(key)(this)
-    initialize(oracle)
+
+    start().map(_ => oracle)
```
Christewart: Is it possible to to start the `DLCOracleAppConfig` with calling `initialize()`?  

Also more generally, is there a reason we don't just use all our existing code in the key manager project rather than copy pasting some of it over to here?

wallet/src/main/scala/org/bitcoins/wallet/config/WalletAppConfig.scala
```diff
@@ -106,11 +106,11 @@ case class WalletAppConfig(
     config.getIntOpt("bitcoin-s.fee-provider.target")
 
   lazy val bip39PasswordOpt: Option[String] = {
-    config.getStringOrNone("bitcoin-s.wallet.bip39password")
+    config.getStringOrNone("bitcoin-s.key-manager.bip39password")
```
Christewart, line 109: Same comment here wrt to creating a `KeyManagerAppConfig`

Christewart (changes requested):

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
benthecarman: > Is it possible to to start the DLCOracleAppConfig with calling initialize()?

That's what this is doing.

> Also more generally, is there a reason we don't just use all our existing code in the key manager project rather than copy pasting some of it over to here?

Fixed

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracleAppConfig.scala
benthecarman: I guess we could, may be better for a future PR so we could integrate all the key manager stuff to use it as well

Christewart: #2255

Christewart (approved):

PR #2254, opened 2020-11-10
Add ability to change aes password

### LINKED ISSUE #2221: Allow the ability to update passwords
Currently in the code base we have two passwords 

1. the aes password used to encrypt the wallet
2. the bip39 mnemonic password 

We need to allow users to change these passwords

Related to #972

benthecarman:
Fixes #2221

Needed to fix `AesPassword`'s `toStringSensitive` as well as allow the console cli to output null json

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -896,9 +896,9 @@ object ConsoleCli {
                 case other => other
               }))
         ),
-      cmd("walletpassphrasechange")
+      cmd("keymanagerpassphrasechange")
         .action((_, conf) =>
-          conf.copy(command = WalletPassphraseChange(null, null)))
+          conf.copy(command = KeyManagerPassphraseChange(null, null)))
         .text("Changes the wallet passphrase")
         .children(
           arg[AesPassword]("oldpassphrase")
@@ -906,7 +906,7 @@ object ConsoleCli {
             .required()
             .action((oldPass, conf) =>
               conf.copy(command = conf.command match {
-                case wpc: WalletPassphraseChange =>
+                case wpc: KeyManagerPassphraseChange =>
                   wpc.copy(oldPassword = oldPass)
                 case other => other
               })),
@@ -915,13 +915,13 @@ object ConsoleCli {
             .required()
             .action((newPass, conf) =>
               conf.copy(command = conf.command match {
-                case wpc: WalletPassphraseChange =>
+                case wpc: KeyManagerPassphraseChange =>
                   wpc.copy(newPassword = newPass)
                 case other => other
               }))
         ),
-      cmd("walletpassphraseset")
-        .action((_, conf) => conf.copy(command = WalletPassphraseSet(null)))
+      cmd("keymanagerpassphraseset")
+        .action((_, conf) => conf.copy(command = KeyManagerPassphraseSet(null)))
         .text("Encrypts the wallet with the given passphrase")
         .children(
           arg[AesPassword]("passphrase")
@@ -929,7 +929,7 @@ object ConsoleCli {
             .required()
             .action((pass, conf) =>
               conf.copy(command = conf.command match {
-                case wps: WalletPassphraseSet =>
+                case wps: KeyManagerPassphraseSet =>
                   wps.copy(password = pass)
                 case other => other
               }))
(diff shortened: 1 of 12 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -892,6 +896,44 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("walletpassphrasechange")
```
Christewart: This is really `keymanagerpassphrasechange` no?

key-manager-test/src/test/scala/org/bitcoins/keymanager/WalletStorageTest.scala
```diff
@@ -89,6 +89,85 @@ class WalletStorageTest extends BitcoinSWalletTest with BeforeAndAfterEach {
...
+        WalletStorage.decryptMnemonicFromDisk(seedPath, None)
+      read match {
+        case Right(readMnemonic) =>
+          assert(writtenMnemonic.mnemonicCode == readMnemonic.mnemonicCode)
+          // Need to compare using getEpochSecond because when reading an epoch second
+          // it will not include the milliseconds that writtenMnemonic will have
+          assert(
+            writtenMnemonic.creationTime.getEpochSecond == readMnemonic.creationTime.getEpochSecond)
+        case Left(err) => fail(err.toString)
+      }
```
Christewart, line 168: I think we need some "negative" test cases. The obvious one to me is

```scala

it must "failed to change the passphrase if the old passphrase is incorrect" in { ... }

```

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -304,6 +304,33 @@ object WalletStorage {
...
 
+  def changeAesPassword(
+      seedPath: Path,
+      oldPasswordOpt: Option[AesPassword],
+      newPasswordOpt: Option[AesPassword]): MnemonicState = {
+    decryptMnemonicFromDisk(seedPath, oldPasswordOpt) match {
+      case Left(err) => sys.error(err.toString)
+      case Right(decrypted) =>
+        val fileName = seedPath.getFileName.toString
+        val backup = seedPath.getParent.resolve(fileName + ".backup")
```
Christewart, line 317: Is there any way we can test this backup logic?

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -304,6 +304,33 @@ object WalletStorage {
...
+
+        val toWrite = newPasswordOpt match {
+          case Some(pass) =>
+            decrypted.encrypt(pass)
+          case None =>
+            decrypted
+        }
+
+        Try(writeMnemonicToDisk(seedPath, toWrite)) match {
+          case Failure(exception) => throw exception
```
Christewart: We should probably add a `log.warn()` or `log.error()` saying that the backup seed can be found at the backup file path.

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -304,6 +304,33 @@ object WalletStorage {
     }
   }
 
+  def changeAesPassword(
+      seedPath: Path,
+      oldPasswordOpt: Option[AesPassword],
+      newPasswordOpt: Option[AesPassword]): MnemonicState = {
+    decryptMnemonicFromDisk(seedPath, oldPasswordOpt) match {
```
Christewart, line 312: We should probably have a `log.info()` indicating we are attempting to change the password of the seed.

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
Christewart: Do we want to move the backup file back to the original seed in this case? I'm 50/50 on that. It seems dangerous in the sense that it makes it easier to "silently" fail changing the password. 

However you may want to write application logic to handle this case and revert to using the `oldPasswordOpt` without downtime. Thoughts?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
Christewart: Same thing applies to all commands I think

Christewart (changes requested):

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
benthecarman: No :/

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
benthecarman: I think it's fine as is. Likely if we need the backup file something went wrong outside our control, we should probably just leave the file and not touch it

Christewart (approved):

PR #2257, opened 2020-11-11
Pulled down non-dlc project work from the adaptor-dlc branch

nkohen:


nkohen:
@benthecarman correct me if I'm wrong but I think migration diffs are missing here?

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -168,6 +168,8 @@ case class SatoshisPerVirtualByte(currencyUnit: CurrencyUnit)
     SatoshisPerVirtualByte
 
   override def toString: String = s"$toLong sats/vbyte"
+
+  def toSatoshisPerKW: SatoshisPerKW = SatoshisPerKW(currencyUnit * 4000)
```
benthecarman: lazy val

testkit/src/main/scala/org/bitcoins/testkit/core/gen/TLVGen.scala
```diff
@@ -98,14 +111,239 @@ trait TLVGen {
...
+        case wtx: WitnessTransaction =>
+          wtx.copy(outputs = wtx.outputs.updated(prevTxVout, newOutput))
+      }
+    } yield {
+      DLCFundingInputP2WPKHV0(newPrevTx, UInt32(prevTxVout), sequence).toTLV
+    }
+  }
+
+  def fundingInputV0TLV: Gen[FundingInputV0TLV] = {
+    fundingInputP2WPKHTLV // Soon to be Gen.oneOf
```
benthecarman, line 161: do we not have p2wsh?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCFundingInput.scala
```diff
@@ -0,0 +1,153 @@
...
+                                      sequence,
+                                      maxWitnessLen,
+                                      redeemScript)
+          case None =>
+            throw new IllegalArgumentException(
+              "P2SH input requires a redeem script")
+        }
+      case _: P2WPKHWitnessSPKV0 =>
+        require(maxWitnessLen == UInt16(107),
+                s"P2WPKH max witness length must be 107, got $maxWitnessLen")
```
benthecarman: This should be `|| 108` so we can allow non-low r

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -48,6 +45,8 @@ object DLCMessage {
       extends NetworkElement {
 
     override def bytes: ByteVector = pubKey.bytes ++ rValue.bytes
+
+    def toTLV: OracleInfoV0TLV = OracleInfoV0TLV(pubKey, rValue)
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -64,16 +63,30 @@ object DLCMessage {
...
+  case class ContractInfo(outcomeValueMap: Vector[(Sha256Digest, Satoshis)])
       extends NetworkElement
-      with MapWrapper[Sha256DigestBE, Satoshis] {
-    override def wrapped: Map[Sha256DigestBE, Satoshis] = outcomeValueMap
+      with SeqWrapper[(Sha256Digest, Satoshis)] {
+    override def wrapped: Vector[(Sha256Digest, Satoshis)] = outcomeValueMap
+
+    def keys: Vector[Sha256Digest] = outcomeValueMap.map(_._1)
+
+    def values: Vector[Satoshis] = outcomeValueMap.map(_._2)
```
benthecarman, line 73: lazy val for all

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -64,16 +63,30 @@ object DLCMessage {
...
+          throw new IllegalArgumentException(s"No value found for key $key"))
+    }
 
     override def bytes: ByteVector = {
       outcomeValueMap.foldLeft(ByteVector.empty) {
         case (vec, (digest, sats)) => vec ++ digest.bytes ++ sats.bytes
       }
     }
+
+    def toTLV: ContractInfoV0TLV = ContractInfoV0TLV(outcomeValueMap)
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -64,16 +63,30 @@ object DLCMessage {
...
+
+    def apply(key: Sha256Digest): Satoshis = {
+      outcomeValueMap
+        .find(_._1 == key)
+        .map(_._2)
+        .getOrElse(
+          throw new IllegalArgumentException(s"No value found for key $key"))
+    }
 
     override def bytes: ByteVector = {
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -114,29 +132,56 @@ object DLCMessage {
...
 
-    val eventId: Sha256DigestBE =
-      calcEventId(oracleInfo, contractInfo, timeouts)
+    lazy val paramHash: Sha256DigestBE =
+      calcParamHash(oracleInfo, contractInfo, timeouts)
+
+    val tempContractId: Sha256Digest =
+      CryptoUtil.sha256(toMessage.bytes)
+
+    def toTLV: DLCOfferTLV = {
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -114,29 +132,56 @@ object DLCMessage {
...
+        totalCollateralSatoshis = totalCollateral,
+        fundingInputs = fundingInputs.map(_.toTLV),
+        changeSPK = changeAddress.scriptPubKey,
+        feeRate = feeRate,
+        contractMaturityBound = timeouts.contractMaturity,
+        contractTimeout = timeouts.contractTimeout
+      )
+    }
+
+    def toMessage: LnMessage[DLCOfferTLV] = {
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -292,20 +383,66 @@ object DLCMessage {
...
+        fundingInputs = fundingInputs.map(_.toTLV),
+        changeSPK = changeAddress.scriptPubKey,
+        cetSignatures = CETSignaturesV0TLV(cetSigs.adaptorSigs),
+        refundSignature =
+          ECDigitalSignature.fromFrontOfBytes(cetSigs.refundSig.signature.bytes)
+      )
+    }
+
+    def toMessage: LnMessage[DLCAcceptTLV] = {
+      LnMessage(this.toTLV)
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -402,43 +594,61 @@ object DLCMessage {
...
+          cetSigs.refundSig.signature.bytes),
+        fundingSignatures = fundingSigs.toTLV
+      )
+    }
+
+    def toMessage: LnMessage[DLCSignTLV] = {
+      LnMessage(this.toTLV)
+    }
+
     def toJson: Value = {
```
benthecarman: lazy val

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCSignatures.scala
```diff
@@ -1,23 +1,51 @@
...
+    get(outPoint).get
+  }
+
+  def merge(other: FundingSignatures): FundingSignatures = {
+    FundingSignatures(sigs ++ other.sigs)
+  }
+
+  def toTLV: FundingSignaturesV0TLV = {
+    FundingSignaturesV0TLV(sigs.map(_._2))
+  }
```
benthecarman: lazy val

core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -95,6 +101,12 @@ sealed abstract class Satoshis extends CurrencyUnit {
 
   def toLong: Long = underlying.toLong
 
+  def toUInt64: UInt64 = {
+    require(toLong >= 0, "Cannot cast negative value to UInt64")
```
benthecarman: lazy val

benthecarman (changes requested):

benthecarman:
> @benthecarman correct me if I'm wrong but I think migration diffs are missing here?

No, since we have the dlc wallet proj now there shouldn't be any diff :)

testkit/src/main/scala/org/bitcoins/testkit/core/gen/TLVGen.scala
nkohen: I mean we could but I didn't want to add anything on this PR that wasn't on adaptor-dlc

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
nkohen: I think I'll leave this one as is at least until 2206 is merged and that stuff is brought down as this all changes and will cause conflicts

core/src/main/scala/org/bitcoins/core/protocol/tlv/LnMessage.scala
```diff
@@ -0,0 +1,38 @@
+package org.bitcoins.core.protocol.tlv
+
+import org.bitcoins.core.number.UInt16
+import org.bitcoins.core.protocol.BigSizeUInt
+import org.bitcoins.crypto.{Factory, NetworkElement}
+import scodec.bits.ByteVector
+
+case class LnMessage[+T <: TLV](tlv: T) extends NetworkElement {
```
Christewart, line 8: Could we get a scaladoc on this? I know I had to wrap my TLVs in this for exchanging with Nicolas Dorier but I don't understand why really.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCFundingInput.scala
```diff
@@ -0,0 +1,154 @@
...
+      case _: P2WSHWitnessSPKV0 =>
+        DLCFundingInputP2WSHV0(prevTx, prevTxVout, sequence, maxWitnessLen)
+      case spk: UnassignedWitnessScriptPubKey =>
+        throw new IllegalArgumentException(s"Unknown segwit version: $spk")
+      case spk: RawScriptPubKey =>
+        throw new IllegalArgumentException(s"Segwit input required: $spk")
+    }
+  }
+
+  def fromTLV(fundingInput: FundingInputV0TLV): DLCFundingInput = {
```
Christewart, line 91: So we have a factory trait for `fromTLV()` ? I do see `TLVFactory` but it only seems to have [`fromTLVValue()`](https://github.com/bitcoin-s/bitcoin-s/blob/c2fa7fccc3e0edacbe13ac7b03c7808d10eee02d/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala#L111)

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -607,3 +624,362 @@ object OracleAnnouncementV0TLV extends TLVFactory[OracleAnnouncementV0TLV] {
     OracleAnnouncementV0TLV(sig, publicKey, eventTLV)
   }
 }
+
+sealed trait ContractInfoTLV extends TLV
```
Christewart, line 628: nit: This is what we need to link to right?  https://github.com/discreetlogcontracts/dlcspecs/blob/6a89f491e4b5e2bb3ca7384be408f376191f6961/NumericOutcome.md#general-payout-curves

Even though the PR is still open, i'm generally in favor of linking to WIP documents as it's better than nothing in my opinion.

core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
```diff
@@ -73,6 +72,21 @@ object BlockStamp extends StringFactory[BlockStamp] {
...
   }
 
+  def apply(timeLockNumber: UInt32): BlockTimeStamp =
+    fromUInt32(timeLockNumber)
+
+  def apply(timeLockNumber: Int): BlockTimeStamp =
+    fromUInt32(UInt32(timeLockNumber))
+
+  /** @see [[https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki#detailed-specification]] */
+  def fromUInt32(uInt32: UInt32): BlockTimeStamp = {
```
Christewart, line 82: nit: Maybe rename to `fromCLTVUInt32()` and remove the `apply()` methods? Then in the future if we need it we can add a `fromCSVUInt32()` 

https://github.com/bitcoin-s/bitcoin-s/blob/c275a8c8aa68712a8d38b661ce581d1309d55f38/core/src/main/scala/org/bitcoins/core/protocol/transaction/TransactionInput.scala#L18-L17

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -352,6 +352,45 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
 
     backwardsDigits.reverse
   }
+
+  /** Returns a pseudorandom, uniformly distributed long value between 0
+    *  (inclusive) and the specified value (exclusive), drawn from this
+    *  random number generator's sequence.
+    *
+    *  Stolen from scala.util.Random.nextLong (in scala version 2.13)
```
Christewart, line 360: Is there a reason we don't just call this behind the scenes rather than copying the source code?

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,97 @@ object ShufflingNonInteractiveFinalizer
...
+
+    val txT = outputsT.map { outputs =>
+      txBuilderResult.toBaseTransaction.copy(outputs = outputs)
+    }
+
+    Future.fromTry(txT)
+  }
+}
+
+case class SubtractFromOutputFinalizer(spk: ScriptPubKey, subAmt: CurrencyUnit)
```
Christewart, line 412: nit: Scaladoc 

I'm assuming this just subtract the given amount from the output that corresponds to the given `spk` ? It might also be useful to add what contexts that is finalizer is useful for.

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,97 @@ object ShufflingNonInteractiveFinalizer
...
     }
+  }
+}
 
-    val dummyTxF = TxUtil.addDummySigs(txWithPossibleWitness, inputInfos)
+/** Assumes the input transaction has had no fee considerations
+  * and subtracts the estimated fee in equal portions from the
+  * outputs with the specified ScriptPubKeys
+  */
+case class SubtractFeeFromOutputsFinalizer(
```
Christewart, line 434: I don't think this has any unit test cases

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
Christewart: I don't think this has any unit test cases

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,97 @@ object ShufflingNonInteractiveFinalizer
   }
 }
 
-case class SubtractFeeFromOutputsFinalizer(
-    inputInfos: Vector[InputInfo],
-    feeRate: FeeUnit)
+/** Adds a future fee amount to the output with the given
+  * ScriptPubKey and subtracts that amount in equal portions
+  * from the specified change ScriptPubKeys.
```
Christewart, line 366: nit (or maybe a misunderstanding on my part): Add to scaladoc 

>This is complentary to `SubtractFromOutputFinalizer` in the sense that an [[futureFee]] is an _absolute_ fee rather than a fee rate.

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,97 @@ object ShufflingNonInteractiveFinalizer
...
+  */
+case class AddFutureFeeFinalizer(
```
Christewart, line 368: I think this is missing unit test cases

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +484,77 @@ object SubtractFeeFromOutputsFinalizer {
...
+    (newOutputs ++ unchangedOutputs).sortBy(_._2).map(_._1)
+  }
+}
+
+case class DualFundingInput(
+    scriptSignature: ScriptSignature,
+    maxWitnessLen: Int)
+
+/** Finalizes a dual-funded transaction given the DualFundingInputs
+  * from both parties, their change spks and the non-change spk.
```
Christewart, line 496: nit: Maybe `non-change spk` ->  `funding output for the dual funded protocol`

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +484,77 @@ object SubtractFeeFromOutputsFinalizer {
...
+    val addAcceptFutureFee = AddFutureFeeFinalizer(fundingSPK,
+                                                   acceptFutureFee,
+                                                   Vector(acceptChangeSPK))
+    val subtractOfferFundingFee =
+      SubtractFromOutputFinalizer(offerChangeSPK, offerFundingFee)
+    val subtractAcceptFundingFee =
+      SubtractFromOutputFinalizer(acceptChangeSPK, acceptFundingFee)
+
+    addOfferFutureFee
+      .andThen(addAcceptFutureFee)
```
Christewart, line 553: This is nice! I feel like `.andThen()` could be renamed to `.map()`. Nice job!

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +484,77 @@ object SubtractFeeFromOutputsFinalizer {
...
+    offerPayoutSPK: ScriptPubKey,
+    offerChangeSPK: ScriptPubKey,
+    acceptInputs: Vector[DualFundingInput],
+    acceptPayoutSPK: ScriptPubKey,
+    acceptChangeSPK: ScriptPubKey,
+    feeRate: SatoshisPerVirtualByte,
+    fundingSPK: ScriptPubKey)
+    extends RawTxFinalizer {
+
+  private def computeFees(
```
Christewart, line 513: A couple thoughts on this method

1. There is a lot of magic numbers (`249`,`164`, `107` etc) that would be useful to put into vals at least so there is a name associated with it 
2. Do you have any thoughts on how we can adjust `FeeRate` to allow this kind of stuff in one place rather than having to do a one off in this method? Or other utility functions if you think `FeeRate` is an inappropriate place for it.

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -168,6 +168,8 @@ case class SatoshisPerVirtualByte(currencyUnit: CurrencyUnit)
     SatoshisPerVirtualByte
 
   override def toString: String = s"$toLong sats/vbyte"
+
+  lazy val toSatoshisPerKW: SatoshisPerKW = SatoshisPerKW(currencyUnit * 4000)
```
Christewart, line 172: We should add a sanity unit test for this -- basically a couple of simple examples.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
```diff
@@ -88,33 +88,47 @@ case class ScriptSignatureParams[+InputType <: InputInfo](
     this.copy(inputInfo = func(this.inputInfo))
   }
 
+  private var _maxScriptSigAndWitnessWeight: Option[(Long, Long)] = None
```
Christewart, line 91: What's the reason that this needs a `var`? Isnt' the computation relatively inexpensive so we can just compute it everytimg `maxScriptSigAndWitnessWeight()` is called?

crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
```diff
@@ -137,6 +137,22 @@ object ECDigitalSignature extends Factory[ECDigitalSignature] {
     }
   }
 
+  def fromFrontOfBytes(bytes: ByteVector): ECDigitalSignature = {
```
Christewart, line 140: nit: Scaladoc, does it just peal off the first 64 bytes?

Christewart (changes requested):
So as far as I can tell, all the files in `org.bitcoins.commons.jsonmodels.dlc` do not have any external dependencies so they could be moved into `core`? 

I guess we don't have to do it on this PR, but we should file an issue to do it in the future.

core/src/main/scala/org/bitcoins/core/protocol/BlockStamp.scala
nkohen: I don't think I follow, both CLTV and CSV can use this?

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
nkohen: bitcoin-s supports 2.12 which doesn't have this

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: Thanks :) I'm not a fan of map here as this the thing on which the method is being called is not the tx, it is the finalizer (making this a composition of functions rather than a mapping on a wrapper of some kind)

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
nkohen: So the problem is that this code is specific to DLCs and not a general thing to be used in other places. Will give the numbers names though.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputSigningInfo.scala
nkohen: it is not inexpensive is the reason. On mac CI this can take like 10 seconds.

crypto/src/main/scala/org/bitcoins/crypto/ECDigitalSignature.scala
nkohen: will do, and no it parses a variable length ECDSA sig

nkohen:
Replaced by #2339

PR #2268, opened 2020-11-17
Create KeyManagerAppConfig

### LINKED ISSUE #2255: Create KeyManagerAppConfig
https://github.com/bitcoin-s/bitcoin-s/pull/2252#discussion_r520660092

This allows us to consolidate key manager configuration and then use the `KeyManagerAppConfig` inside things like `DLCOracleAppConfig` and `WalletAppConfig`

benthecarman:
Closes #2255

### DIFF
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
@@ -38,12 +38,12 @@ case class KeyManagerAppConfig(
   override def start(): Future[Unit] = FutureUtil.unit
 
   lazy val aesPasswordOpt: Option[AesPassword] = {
-    val passOpt = config.getStringOrNone("bitcoin-s.key-manager.aesPassword")
+    val passOpt = config.getStringOrNone(s"bitcoin-s.$moduleName.aesPassword")
     passOpt.flatMap(AesPassword.fromStringOpt)
   }
 
   lazy val bip39PasswordOpt: Option[String] = {
-    config.getStringOrNone("bitcoin-s.key-manager.bip39password")
+    config.getStringOrNone(s"bitcoin-s.$moduleName.bip39password")
   }
 
   /** Checks if our key manager as a mnemonic seed associated with it */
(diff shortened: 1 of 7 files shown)


key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  override def configOverrides: List[Config] = confs.toList
+
+  override type ConfigType = KeyManagerAppConfig
+
+  override def newConfigOfType(
+      configOverrides: Seq[Config]): KeyManagerAppConfig =
+    KeyManagerAppConfig(directory, configOverrides: _*)
+
+  override def moduleName: String = "key-manager"
```
Christewart: I believe we use this for database names? I think the `-` might cause us issues in the future with various things that don't allow `-` in their namespace. I would recommend just doing `keymanager` for simplicity.

Christewart (commented):
I think everything else looks ok, but the documentation needs to be updated.

key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /** The path to our encrypted mnemonic seed */
+  lazy val seedPath: Path = {
+    baseDatadir.resolve(WalletStorage.ENCRYPTED_SEED_FILE_NAME)
+  }
+
+  override def start(): Future[Unit] = FutureUtil.unit
+
+  lazy val aesPasswordOpt: Option[AesPassword] = {
+    val passOpt = config.getStringOrNone("bitcoin-s.key-manager.aesPassword")
```
Christewart: Shouldn't these be incorrect now? Also a little concerning that unit tests are failing from this...

benthecarman: The `moduleName` is only used for db stuff so it doesn't effect anything for the key manager app config

Christewart: ok but why not be consistent? now I need to remember configuration is `key-manager` and database connections need to be done with `keymanager`

Christewart: We even use `moduleName` for database configuration it seems. While there isn't currently a database component to the key manager, there may be times in the future where we would want one 

https://github.com/bitcoin-s/bitcoin-s/blob/2de17bb4e47961fd3fe4e55d75909b42470e3c59/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala#L165-L164

Christewart (approved):

PR #2271, opened 2020-11-17
Fix datadir config option

### LINKED ISSUE #2261: ./bitcoin-s-server --datadir option not working always
I'm attempting to use an alternative datadir for bitcoin-s-server to test #2260 . This is being run on  ff9119e975706337284d01fd4310c19cf9234995

>./bitcoin-s-server --datadir /tmp/bitcoin-s-issue2258/

Here are the contents of `/tmp/bitcoin-s-issues2258/` 

>tree .
.
└── bitcoin-s.conf

When I run the server with the specific datadir it doesn't read settings from the `bitcoin-s.conf` in the empty data directory.

Here are the contents of the `bitcoin-s.conf` file 
```
bitcoin-s {
    network = testnet3 # regtest, testnet3, mainnet
    node {
        peers = ["neutrino.testnet3.suredbits.com:8333"] # a list of peer addresses in form "hostname:portnumber"
                              # (e.g. "localhost:18444")
                              # (e.g. "neutrino.testnet3.suredbits.com:18333")
                              # (e.g. "neutrino.suredbits.com:8333")
                              # Port number is optional, the default value is 8333 for mainnet,
                              # 18333 for testnet and 18444 for regtest.
        mode = neutrino# neutrino, spv
    }

    chain {
        neutrino {
            filter-header-batch-size.default = 2000
            filter-header-batch-size.regtest = 10
            filter-batch-size = 100
        }
    }

    # settings for wallet module
    wallet {
        defaultAccountType = segwit# legacy, segwit, nested-segwit

        bloomFalsePositiveRate = 0.0001 # percentage
        requiredConfirmations = 1
    }
    server {
   	#rpcport=12345
    }
    keymanager { 

	aesPassword = ""
    }
}

```

benthecarman:
Fixes #2261

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
@@ -6,7 +6,7 @@ import akka.actor.ActorSystem
 import com.typesafe.config.{Config, ConfigFactory}
 import org.bitcoins.core.config._
 import org.bitcoins.core.util.BitcoinSLogger
-import org.bitcoins.db.{AppConfig, ConfigOps}
+import org.bitcoins.db.AppConfig
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Properties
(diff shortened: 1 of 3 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
```diff
@@ -68,26 +70,23 @@ trait BitcoinSRunner extends BitcoinSLogger {
...
+    Paths.get(baseConfig.getString("bitcoin-s.datadir"))
 
   private lazy val networkStr: String =
     baseConfig.getString("bitcoin-s.network")
 
   private lazy val network: BitcoinNetwork =
     BitcoinNetworks.fromString(networkStr)
 
-  lazy val datadir: Path = {
+  private lazy val datadir: Path = {
```
Christewart: Ok, so from looking at this now we have 3 ways to represent `datadir`

1. `datadir`
2. `configDatadir`
3. `withDatadir`

Can we get rid of some of these? If not, we should definitely add scaladocs to all of these values so it's apparent to the reader of the code which each is intended for. This is a bit of a mess

Christewart (commented):
Everything seems to work as of 6ca44c58

app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
```diff
@@ -68,32 +69,30 @@ trait BitcoinSRunner extends BitcoinSLogger {
...
-  lazy val datadir: Path = {
-    lazy val lastDirname = network match {
-      case MainNet  => "mainnet"
-      case TestNet3 => "testnet3"
-      case RegTest  => "regtest"
-      case SigNet   => "signet"
-    }
-    datadirPath.resolve(lastDirname)
-  }
+  /** Base directory for all bitcoin-s data */
```
Christewart:
```scala
/** The datadir specified inside of the bitcoin-s.conf file under the key `bitcoin-s.datadir`. 
  * This will be used as the datadir if a --datadir option was not specified on the command line */ 
```

is more descriptive and useful in my opinion.

app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
```diff
@@ -68,32 +69,30 @@ trait BitcoinSRunner extends BitcoinSLogger {
...
+        baseConfig.getString("bitcoin-s.network")
+      val network: BitcoinNetwork =
+        BitcoinNetworks.fromString(networkStr)
+      lazy val lastDirname = network match {
+        case MainNet  => "mainnet"
+        case TestNet3 => "testnet3"
+        case RegTest  => "regtest"
+        case SigNet   => "signet"
+      }
+      configDataDir.resolve(lastDirname)
```
Christewart: Does this still take into account the `--datadir` command line option? Where does that come into play here?

benthecarman: `configDataDir` is the resulting datadir from the `--datadir` `--conf` and `bitcoin-s.conf`

app/server/src/main/scala/org/bitcoins/server/BitcoinSRunner.scala
benthecarman: This isn't true, it is the final datadir we will be using. Taking into account the `--datadir` option, `--conf` option, and the `bitcoin-s.conf` file

Christewart: ok well clearly I don't understand what you are doing, so let's add docs so it's easier to follow.

Christewart (approved):

PR #2275, opened 2020-11-18
Add directive to configure timeout on the appServer

Christewart:
This adds a timeout for server directives. It keeps it to the current value 10 seconds for as I think that is a sane default for most cases. 

@benthecarman could you test this out with the workload you have by adjusting this setting and making sure it works?

https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/timeout-directives/withRequestTimeout.html#withrequesttimeout

https://doc.akka.io/docs/akka-http/current/common/timeouts.html#request-timeout

### DIFF
--- a/docs/config/configuration.md
+++ b/docs/config/configuration.md
@@ -210,6 +210,7 @@ akka {
         server {
             # The amount of time until a request times out on the server
             # If you have a large payload this may need to be bumped
+            # https://doc.akka.io/docs/akka-http/current/common/timeouts.html#request-timeout
             request-timeout = 10s
         }
     }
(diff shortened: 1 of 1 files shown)


Christewart:
Looking at the documentation a bit closer, there seems to be warnings of race conditions using this directive 

(see `warning` section: https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/timeout-directives/withRequestTimeout.html#description)


it may be better to configure this in the `bitcoin-s.conf` file with this key 

>akka.http.server.request-timeout = 100.seconds #or whatever

https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/timeout-directives/withRequestTimeout.html#example

rather than using this directive.

benthecarman:
The current code works, however, I do agree it would probably be better to just set it in the akka config. Maybe add to the reference.conf file 
```
akka.http.server.request-timeout = 10s
```

benthecarman (approved):

PR #2286, opened 2020-11-21
Experimental: 2020 11 07 scalajs build

Christewart:
This PR is an experiment in adding [scalajs](https://www.scala-js.org/) support to bitcoin-s. 

The goal is to have scalajs modules published for both `bitcoin-s-crypto` and `bitcoin-s-core` which both have minimal dependencies. 

I would recommend reading the [sbt-crossproject](https://github.com/portable-scala/sbt-crossproject#configuration) differnent configuration settings to understand how you can have codebases that are 

1. Shared for both the JVM and JS runtime
2. Specific code for the JS runtime
3. Specific code for the JVM runtime 

### Current state

The current state of this PR is that it adds scalajs support for the [scodec-bits](https://github.com/scodec/scodec-bits) dependency which is compatible with scalajs. 

I just wanted to get something working with a dependency that is core to our entire library. You can see the diff in `Deps.scala` and `build.sbt` to see the things I needed to do to get things working :tm:

As of now, there is NO JAVASCRIPT TESTS. The current state of this PR is just trying to fiddle with the build definition to _theoretically_ support a scalajs dependency, but doesn't actually execute any code that is dependent on scalajs. I want to make sure all of our JVM functionality is working and doesn't regress while adding scalajs.

### Problems I can forsee

One problem that we are going to encounter is that scalajs cannot work with java libraries. Bouncy castle is a java only library so it cannot be compiled to javascript.

As of e8ea413 I have removed the dependency of `crypto.dependsOn(secp256k1jni)`. I've done this for 2 reasons

1. I'm trying simplify the build to just get something working :tm:
2. I'm unsure how to use native code in peoples browsers.

If we can figure out this problem, we may be able to use our [native-lib-loader](https://github.com/scijava/native-lib-loader) with a special javascript native. I'm unsure at the time of this writing. 

### Resources

[scalajs documentation](https://www.scala-js.org/doc/)

[scalajs cross build docs](https://www.scala-js.org/doc/project/cross-build.html)

[sbtcrossproject build tool](https://github.com/portable-scala/sbt-crossproject)

[scalajs gitter](https://gitter.im/scala-js/scala-js)

[cats build.sbt](https://github.com/typelevel/cats/blob/master/build.sbt) -- they support scalajs and are a decent size project 

[Existing scala libraries compatible with scalajs](https://www.scala-js.org/libraries/libs.html)


Christewart:
Some extra information: 

We now will have two separate projects inside of `sbt`, one for the JVM and one for JS. I introduced the JS platform for the crypto project in c3286d1

Now if you type `projects` into the sbt console you will see both `cryptoCrossProjectJVM` and `cryptoCrossProjectJS`. 

You can now type `cryptoCrossProjectJVM/compile` and compile the jvm code, and the same for `cryptoCrossProjectJS/compile` to compile the javascript portion.

Something that _could_ be a problem for us is that our test cases _are not in the same project_ as the source code. In the case of the `crypto` project all of our test cases live in `cryptoTest`. 

The problem with this is that [`cryptoTest.dependsOn(testkit)`](https://github.com/bitcoin-s/bitcoin-s/blob/f22e1b6edb8d37435d803f4e3b332d90341ec7e0/build.sbt#L218) and [`testkit` dependsOn every project](https://github.com/bitcoin-s/bitcoin-s/blob/master/build.sbt#L440-L450). This means we would need to get every subproject working with scalajs to be able to use `testkit` to in turn be able to use our traditional test projects.

Screenshot from 2020-11-21 11-07-38

PR #2287, opened 2020-11-22
Move english word list to be represented by a Vector in memory

Christewart:
When working on #2286 this seems to be the only place in `core`/`crypto` that we read from a file. Reading a classpath file is more complicated that calling `getResource()`. I havent' figured out exactly how it works yet and thought i would just work around it by just moving these words to be an in memory `Vector`.

I move the previous logic for reading from a file to a unit test to make sure the word lists are the same.


core/src/main/scala/org/bitcoins/core/crypto/words/EnglishWords.scala
```diff
@@ -0,0 +1,2055 @@
+package org.bitcoins.core.crypto.words
+
+object EnglishWords extends MnemonicWords {
```
benthecarman: nit: maybe name `EnglishBIP39Words` since they are specific to bip39

benthecarman (approved):

PR #2289, opened 2020-11-23
Request witness blocks from peers

### LINKED ISSUE #2281: Wallet Transaction Table does not have witness data for some txs
@nkohen and I found that some of our db entries do not have their witness data. We haven't found how this could be happening.

benthecarman:
Fixes #2281 

We were requesting blocks without witness data

https://bitcoin.stackexchange.com/questions/92273/how-do-i-ask-for-witness-data-from-a-bitcoin-full-node

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -126,9 +126,7 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
         )
       }
 
-      val countF = node.chainApiFromDb().flatMap(_.getFilterHeaderCount)
       for {
-        count <- countF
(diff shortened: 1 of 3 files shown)


Christewart:
Hmm, so what does an existing wallet need to do to fetch witnesses? Rescan the entire wallet?

benthecarman:
> Hmm, so what does an existing wallet need to do to fetch witnesses? Rescan the entire wallet?

Probably the best solution

nkohen:
Couldn't they just ask for all blocks containing transactions they have and no others?

Christewart:
> Couldn't they just ask for all blocks containing transactions they have and no others?

That is how our rescanning works. We search through our local block filters for matches and then request full blocks that match. 

I guess we could just read the txo block_hash from the database and request those, but that requires custom logic. I think rescanning should be fast enough.

Christewart (approved):

PR #2292, opened 2020-11-25
Make BitcoinSAyncTest more thread safe and make sure we have all the …

### LINKED ISSUE #1536: NPE on PSBT construct and sign test case

Christewart:
…expected test cases aggregated before we starting reducing over the results

I'm hoping this will fix #1536 

Previously it seems that we did not actually aggregate all of our test cases (should be either `10` or `100` of them depenending on [generatorDrivenConfig](https://github.com/Christewart/bitcoin-s-core/blob/a7869fa4395ab99dc46641d007cc0e031997f5ba/testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala#L67))

I've now added two things

1. Using `java.util.concurrent.CopyOnWriteArrayList` rather than `VectorBuilder` to aggregate test cases
2. `forAllHelper()` which checks that we have the expected number of test cases before reducing the results

For (1) i'm not sure what the story is with thread safety and using a `VectorBuilder` but we I decided to do the dumb thing and just use a java data structure. 

For (2) I was able to actually reproduce instances locally of not having all the expected test results which is really concerning! I would run into things instances were we have `99` assertions in the Vector, but we expected `100` (this was the setting in [generatorDrivenConfig](https://github.com/Christewart/bitcoin-s-core/blob/a7869fa4395ab99dc46641d007cc0e031997f5ba/testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala#L67))

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/util/BitcoinSAsyncTest.scala
@@ -6,18 +6,19 @@ import akka.util.Timeout
 import org.bitcoins.core.config.{NetworkParameters, RegTest}
 import org.bitcoins.core.protocol.blockchain.ChainParams
 import org.bitcoins.core.util.BitcoinSLogger
+import org.bitcoins.rpc.util.AsyncUtil
 import org.scalacheck.{Gen, Shrink}
 import org.scalactic.anyvals.PosInt
 import org.scalatest._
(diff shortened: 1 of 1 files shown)


benthecarman (changes requested):
Doesn't compile on 2.12.12

benthecarman (approved):

PR #2312, opened 2020-12-02
Fix TLV parsing for non-standard strings

benthecarman:
Found by @rorp 

Fixes parsing UTF 8 strings with uncommon letters.

Needed to change the comparison between TLVs because it seems scala doesn't correctly compare complex UTF 8 strings, unless this is something to do with us not doing something with NFC normalization after parsing?

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
@@ -179,12 +179,8 @@ case class NormalizedString(private val str: String) {
 
 object NormalizedString {
 
-  def apply(str: String): NormalizedString = {
-    new NormalizedString(CryptoUtil.normalize(str))
-  }
-
   def apply(bytes: ByteVector): NormalizedString = {
-    apply(new String(bytes.toArray, StandardCharsets.UTF_8))
+    NormalizedString(new String(bytes.toArray, StandardCharsets.UTF_8))
   }
 
   import scala.language.implicitConversions
--- a/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
@@ -99,6 +99,13 @@ object TLV extends TLVParentFactory[TLV] {
       case None             => UnknownTLV(tpe, value)
     }
   }
+
+  def getStringBytes(str: String): ByteVector = {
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -379,8 +386,8 @@ case class RangeEventDescriptorV0TLV(
   override val tpe: BigSizeUInt = RangeEventDescriptorV0TLV.tpe
 
   override val value: ByteVector = {
-    val unitSize = BigSizeUInt(unit.length)
     val unitBytes = CryptoUtil.serializeForHash(unit)
+    val unitSize = BigSizeUInt(unitBytes.size)
```
rorp: It would be nice to DRY it up.

rorp (changes requested):

rorp (approved):

benthecarman:
Do not merge, currently having problems with parsing UTF 8 strings in the oracle database

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -155,6 +162,46 @@ sealed trait TLVFactory[+T <: TLV] extends Factory[T] {
...
+      case otherStr: String =>
+        normStr == otherStr
+      case _ => other.equals(str)
+    }
+  }
+
+  override def toString: String = normStr
+}
+
+object NormalizedString {
```
Christewart: nit: Can we use `Factory[NormalizedString]`?

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -155,6 +162,46 @@ sealed trait TLVFactory[+T <: TLV] extends Factory[T] {
   }
 }
 
+case class NormalizedString(private val str: String) {
```
Christewart: `extends NetworkElement`?  It seems it's pretty important that we serialize these correctly for when we are passing outcomes over the wire if I understand correctly

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
benthecarman: Used `StringFactory` instead because `Factory` would then have the `apply(String)` function be the `fromHex` function instead of the contructor

Christewart (approved):

PR #2314, opened 2020-12-03
WIP: Multi-Oracle Support!

nkohen:
Implemented n-of-n oracle CET computation logic, manually tested in console and it works! (After I did some debugging of course, didn't magically get it right the first time this time)

TODO
* real tests
* simplification/refactor & documentation
* numeric version (as opposed to digit manipulation) to test against and for reference
* dual version which minimizes the number of non-required cases for the same number of CETs
* Logic for t-of-n => tCn t-of-ts
* Logic for having different variance parameters between oracles

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
@@ -2,8 +2,10 @@ package org.bitcoins.core.protocol.dlc
 
 import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.protocol.dlc.CETCalculator._
+import org.bitcoins.testkit.core.gen.NumberGenerator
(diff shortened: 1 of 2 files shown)


app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
```diff
@@ -420,4 +421,281 @@ object CETCalculator {
...
+        val (primaryLeft, primaryRight) =
+          computeCETIntervalBinary(primaryCET, numDigits)
+        val (coveringLeft, coveringRight) =
+          computeCETIntervalBinary(coveringCET, numDigits)
+
+        println()
+        println(s"Primary Oracle in [$primaryLeft, $primaryRight]")
+        println(s"Covering Oracle in [$coveringLeft, $coveringRight]")
+    }
+  }
```
benthecarman: Why not change these to `logger.infos` so we can keep the function

nkohen:
Replaced by #2508

PR #2315, opened 2020-12-03
Fix P2PClient parsing unknown messages

benthecarman:
I tried syncing my node with the latest release candidate of Bitcoin Core and found it wouldn't sync.

The error was that it wasn't able to parse the new P2P messages introduced. This fixes that by checking if it's an unknown type to skip the message.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
@@ -355,13 +355,13 @@ object P2PClient extends P2PLogger {
     * @return the parsed [[NetworkMessage]]'s and the unaligned bytes that did not parse to a message
     */
   private[bitcoins] def parseIndividualMessages(
-      bytes: ByteVector): (List[NetworkMessage], ByteVector) = {
+      bytes: ByteVector): (Vector[NetworkMessage], ByteVector) = {
     @tailrec
     def loop(
         remainingBytes: ByteVector,
-        accum: List[NetworkMessage]): (List[NetworkMessage], ByteVector) = {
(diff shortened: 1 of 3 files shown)


node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -363,17 +363,41 @@ object P2PClient extends P2PLogger {
...
+            if (NetworkPayload.commandNames.contains(header.commandName)) {
+              Try(NetworkMessage(header.bytes ++ payloadBytes)) match {
+                case Success(message) =>
+                  logger.trace(
+                    s"Parsed a message=${message.header.commandName} from bytes, continuing with remainingBytes=${newRemainingBytes.length}")
+
+                  loop(newRemainingBytes, message :: accum)
+                case Failure(_) =>
+                  // Can't parse message yet, we need to wait for more bytes
+                  (accum.reverse, remainingBytes)
```
rorp: Why `accum.reverse`?

rorp: Oh I see, never mind

rorp: Although  if we change `accum` type to `Vector` we could simply append new messages, and wouldn't need `reverse`

benthecarman: Changed to use `Vector` and removed the `reverse`s

Christewart (approved):

Christewart:
Tested 03024f2d65b by syncing a few thousand blocks on testnet, everything worked fine.

PR #2321, opened 2020-12-04
Implement BIP 155 addrv2 messages

benthecarman:
https://github.com/bitcoin/bips/blob/master/bip-0155.mediawiki

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
+++ b/core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
@@ -1,11 +1,10 @@
 package org.bitcoins.core.p2p
 
 import java.net.{InetAddress, InetSocketAddress}
-
 import org.bitcoins.core.bloom.{BloomFilter, BloomFlag}
 import org.bitcoins.core.config.NetworkParameters
 import org.bitcoins.core.gcs.{FilterHeader, FilterType, GolombFilter}
-import org.bitcoins.core.number.{Int32, Int64, UInt32, UInt64}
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
```diff
@@ -528,6 +527,212 @@ object AddrMessage extends Factory[AddrMessage] {
...
+  final val CJDNS_ADDR_LENGTH: Int = 16
+
+  val knownNetworkBytes: Vector[Byte] = Vector(IPV4_NETWORK_BYTE,
+                                               IPV6_NETWORK_BYTE,
+                                               TOR_V2_NETWORK_BYTE,
+                                               TOR_V3_NETWORK_BYTE,
+                                               I2P_NETWORK_BYTE,
+                                               CJDNS_NETWORK_BYTE)
+
+  def fromBytes(bytes: ByteVector): AddrV2Message = {
```
Christewart: nit: `overrides`

Christewart (commented):
The only thing I would like to add is a simple hard coded example of each `AddrV2Message`. 

Sometimes with property based tests you can implement the _wrong_ thing, but have tests pass because you implemented the serialization/deserialization to be correct in the sense that they are symmetrical (you can serialize and deserialize the _wrong_ thing, i've done this lol) but they are not functionally correct.

If you add in a simple hard coded unit test -- such as examples from the BIP -- you know you are correct wrt to the spec too, along with being internally consistent :-)

Christewart (approved):

PR #2339, opened 2020-12-07
Pulled down work from adaptor-dlc onto master

nkohen:
Replaces #2257

On next PR: 
* Needs tests:
    - ECDigitalSignature.fromFrontOfBytes[WithSigHash]
    - SatoshisPerVirtualByte.toSatoshisPerKW
    - RawTxFinalizers
    - `DualFundingTxFinalizer.computeFees`
    - sigFromMsgAndSigs

* Move from appCommons to core
* Make OutcomeValuePoint into an ADT (splitting on `isEndpoint`)
* OutcomeValueFunction -> DLCPayoutCurve
* OutcomeValueFunctionComponent -> DLCPayoutCurveComponent
* OutcomeValuePoint -> OutcomePayoutPoint
* Replace tuples in OutcomeValueFunction and RoundingIntervals with types
* Create overloaded SchnorrPublicKey.computeSigPoint(Sha256Digest, SchnorrNonce)
* Create util method to reduce list of digests and nonces to a singe point
* Move local methods in `calculateOutcomeAndSig` out into private helper methods
* Add to website


app/server/src/main/scala/org/bitcoins/server/ChainRoutes.scala
```diff
@@ -35,6 +38,15 @@ case class ChainRoutes(chain: ChainApi)(implicit system: ActorSystem)
           Server.httpSuccess(hash)
         }
       }
+
+    case ServerCommand("getinfo", _) =>
```
nkohen, line 42: @benthecarman does this have a test someplace?

benthecarman: No, can add one to `RoutesSpec`

core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala
```diff
@@ -228,7 +233,7 @@ object BitcoinNetworks extends StringFactory[BitcoinNetwork] {
     }
 
   /** Map of magic network bytes to the corresponding network */
-  val magicToNetwork: Map[ByteVector, NetworkParameters] =
+  def magicToNetwork: Map[ByteVector, NetworkParameters] =
```
benthecarman: why this change?

core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -108,6 +114,12 @@ sealed abstract class Satoshis extends CurrencyUnit {
 
   def toLong: Long = underlying.toLong
 
+  def toUInt64: UInt64 = {
```
benthecarman: `lazy val`

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -168,6 +168,8 @@ case class SatoshisPerVirtualByte(currencyUnit: CurrencyUnit)
     SatoshisPerVirtualByte
 
   override def toString: String = s"$toLong sats/vbyte"
+
+  def toSatoshisPerKW: SatoshisPerKW = SatoshisPerKW(currencyUnit * 4000)
```
benthecarman: lazy val

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/InitEnumContractDialog.scala
```diff
@@ -0,0 +1,101 @@
+package org.bitcoins.gui.dlc.dialog
+
+import org.bitcoins.commons.jsonmodels.dlc.DLCMessage.SingleNonceContractInfo
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.protocol.tlv.EnumOutcome
+import org.bitcoins.gui.GlobalData
+import org.bitcoins.gui.util.GUIUtil.setNumericInput
+import scalafx.Includes._
+import scalafx.geometry.{Insets, Pos}
+import scalafx.scene.control.{Label, _}
```
benthecarman: nit: `._`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
```diff
@@ -0,0 +1,398 @@
...
+object CETCalculator {
+
+  /** Given a range and a payout function with which to build CETs,
+    * the first step is to split the range into sub-ranges which
+    * can be compressed or cannot be compressed, represented here
+    * as CETRanges.
+    *
+    * These ranges are inclusive in both indices.
+    */
+  sealed trait CETRange {
```
Christewart, line 22: nit: Maybe consider wrapping the built in Scala `NumericRange`? https://github.com/scala/scala/blob/a93d87fef29cca6fb40d4d945ecb476beb784bd2/src/library/scala/collection/immutable/NumericRange.scala#L40

I think specifically https://github.com/scala/scala/blob/a93d87fef29cca6fb40d4d945ecb476beb784bd2/src/library/scala/collection/immutable/NumericRange.scala#L402

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
```diff
@@ -0,0 +1,398 @@
...
+  /** This range contains payouts all <= 0 */
+  case class StartZero(indexFrom: Long, indexTo: Long) extends CETRange
+
+  /** This range contains payouts all >= totalCollateral */
+  case class StartTotal(indexFrom: Long, indexTo: Long) extends CETRange
+
+  /** This range contains payouts that all vary at every step and cannot be compressed */
+  case class StartFunc(indexFrom: Long, indexTo: Long) extends CETRange
+
+  /** This range contains some constant payout between 0 and totalCollateral (exclusive) */
```
Christewart: This scaladoc is conflicting with the parent doc (exclusive vs inclusive)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
```diff
@@ -0,0 +1,398 @@
...
+    * as CETRanges.
+    *
+    * These ranges are inclusive in both indices.
+    */
+  sealed trait CETRange {
+    def indexFrom: Long
+    def indexTo: Long
+  }
+
+  /** This range contains payouts all <= 0 */
```
Christewart: Ok dumb question, but how can you have a payout less then zero? You can't have negative satoshis in the bitcoin protocol. 

Same concept applies with `StartTotal` and `>=totalCollateral` (can't have more money than total collateral).

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+/** A point on a DLC payout curve to be used for interpolation
+  *
+  * @param outcome An element of the domain of possible events signed by the oracle
+  * @param value The payout to the local party corresponding to outcome
+  * @param isEndpoint True if this point defines a boundary between pieces in the curve
+  */
+case class OutcomeValuePoint(
+    outcome: BigDecimal,
+    value: Satoshis,
+    isEndpoint: Boolean)
```
Christewart, line 74: nit: In another PR it might be nice to represent `isEndPoint` as a type rather than a flag so we don't forget to 

```scala
if (isEndpoint) { ... } 

``` 

somwhere, but rather get a compiler error.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+
+/** A single piece of a larger piecewise function defined between left and right endpoints */
+sealed trait OutcomeValueFunctionComponent {
+  def leftEndpoint: OutcomeValuePoint
+  def midpoints: Vector[OutcomeValuePoint]
+  def rightEndpoint: OutcomeValuePoint
+
+  require(leftEndpoint.isEndpoint, s"$leftEndpoint not an endpoint")
```
Christewart, line 82: for instance this invariants could be removed

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
+package org.bitcoins.commons.jsonmodels.dlc
+
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.util.NumberUtil
+
+import scala.math.BigDecimal.RoundingMode
+
+/** A DLC payout curve defined by piecewise interpolating points */
+case class OutcomeValueFunction(points: Vector[OutcomeValuePoint]) {
```
Christewart, line 9: nit: Could we maybe rename this to just `PayoutCurve`? As the scaladoc indicates that seems to be what this represents. `OutcomeValueFunction` isn't as intuitive as me :man_shrugging:

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+  * @param value The payout to the local party corresponding to outcome
+  * @param isEndpoint True if this point defines a boundary between pieces in the curve
+  */
+case class OutcomeValuePoint(
+    outcome: BigDecimal,
+    value: Satoshis,
+    isEndpoint: Boolean)
+
+/** A single piece of a larger piecewise function defined between left and right endpoints */
+sealed trait OutcomeValueFunctionComponent {
```
Christewart, line 77: nit: and then maybe rename this to `PayoutCurveInterval`

benthecarman (approved):

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+  lazy val functionComponents: Vector[OutcomeValueFunctionComponent] = {
+    endpoints.init.zip(endpoints.tail).map { // All pairs of adjacent endpoints
+      case ((_, index), (_, nextIndex)) =>
+        OutcomeValueFunctionComponent(points.slice(index, nextIndex + 1))
+    }
+  }
+
+  private lazy val outcomes = endpoints.map(_._1.outcome)
+
+  def componentFor(
```
Christewart, line 35: nit: If I understand this 

```scala
/**
Finds the payout interval that corresponds to the given outcome
*/

```

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+object OutcomeValueFunctionComponent {
+
+  def apply(
+      points: Vector[OutcomeValuePoint]): OutcomeValueFunctionComponent = {
+    points match {
+      case Vector(left, right) =>
+        if (left.value == right.value) {
+          OutcomeValueConstant(left, right)
+        } else {
+          OutcomeValueLine(left, right)
```
Christewart, line 121: does it matter here if `right > left`?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+sealed trait OutcomeValueFunctionComponent {
+  def leftEndpoint: OutcomeValuePoint
+  def midpoints: Vector[OutcomeValuePoint]
+  def rightEndpoint: OutcomeValuePoint
+
+  require(leftEndpoint.isEndpoint, s"$leftEndpoint not an endpoint")
+  require(rightEndpoint.isEndpoint, s"$rightEndpoint not an endpoint")
+  require(midpoints.forall(!_.isEndpoint), s"$midpoints contained an endpoint")
+
+  def apply(outcome: BigDecimal): Satoshis
```
Christewart, line 100: nit: maybe rename to `getPayout()` or provide it as an alternative (if I understand this)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+
+    bigDecimalSats(value)
+  }
+}
+
+/** A polynomial interpolating points and defining a piece of a larger payout curve */
+case class OutcomeValuePolynomial(points: Vector[OutcomeValuePoint])
+    extends OutcomeValueFunctionComponent {
+  override lazy val leftEndpoint: OutcomeValuePoint = points.head
+  override lazy val rightEndpoint: OutcomeValuePoint = points.last
```
Christewart, line 247: probably should have invariants that 

1. left < right
2. midpoints.forall(m => left < m)
3. midpoints.forall(m => right > m)

I guess could also just wrap `Vector[OutcomeValuePoint]` in some sort of class that makes sure these points are ordered? It does seem like `Vector[OutcomeValuePoint]` is used in a decent amount of places and I think this assumption is made implicitly?

hopefully i'm not misunderstanding something

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+/** A point on a DLC payout curve to be used for interpolation
+  *
+  * @param outcome An element of the domain of possible events signed by the oracle
+  * @param value The payout to the local party corresponding to outcome
+  * @param isEndpoint True if this point defines a boundary between pieces in the curve
+  */
+case class OutcomeValuePoint(
```
Christewart, line 71: going with my theme, I have this written down in my notes  as `PayoutPoint` -- (lol close enough to `PaymentPoint`? ;) ) -- with the additional comment 

>The (x,y) corresponds to the payout where `y` is satoshis

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
Christewart: I guess is `outcome` on the x-axis or y-axis? That would probably be useful for the scaladoc

app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala
```diff
@@ -0,0 +1,306 @@
...
+
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
+
+  behavior of "OutcomeValueFunction"
+
+  private val numGen = Gen.choose[Double](0, 1000).map(BigDecimal(_))
+  private val intGen = Gen.choose[Int](0, 1000)
+
+  def nPoints(n: Int): Gen[Vector[OutcomeValuePoint]] = {
```
Christewart, line 20: Ok, so I do have a concern here: 

Does this exhaustively test for realistic numbers

For instance `100,000` seems to be the max value of satoshis we are testing for here, and the max outcome is `1,000`. 

For our domain of BTC/USD price we should probably raise `numGens` max value to be 100,000 and maybe a max payout of like 10BTC or 100BTC? 

A useful tool here might be `Gen.frequency` which allows you to specify the frequency that a number is generated by a generator.

app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala
```diff
@@ -0,0 +1,306 @@
+package org.bitcoins.commons.dlc
+
+import org.bitcoins.commons.jsonmodels.dlc._
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.testkit.util.BitcoinSUnitTest
+import org.scalacheck.Gen
+
+import scala.math.BigDecimal.RoundingMode
+
+class OutcomeValueFunctionTest extends BitcoinSUnitTest {
```
Christewart, line 10: Some compiler warnings for this file

```
[warn] /home/suredbits/dev/bitcoin-s/app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala:36:51: match may not be exhaustive.
[warn] It would fail on the following inputs: (Vector0, _), (Vector1(), _), (Vector2(), _), (Vector3(), _), (Vector4(), _), (Vector5(), _), (Vector6(), _)
[warn]     forAll(nPoints(2), Gen.listOfN(1000, numGen)) {
[warn]                                                   ^
[warn] /home/suredbits/dev/bitcoin-s/app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala:78:51: match may not be exhaustive.
[warn] It would fail on the following inputs: (Vector0, _), (Vector1(), _), (Vector2(), _), (Vector3(), _), (Vector4(), _), (Vector5(), _), (Vector6(), _)
[warn]     forAll(nPoints(3), Gen.listOfN(1000, numGen)) {
[warn]                                                   ^
[warn] /home/suredbits/dev/bitcoin-s/app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala:151:51: match may not be exhaustive.
[warn] It would fail on the following inputs: (Vector0, _), (Vector1(), _), (Vector2(), _), (Vector3(), _), (Vector4(), _), (Vector5(), _), (Vector6(), _)
[warn]     forAll(nPoints(4), Gen.listOfN(1000, numGen)) {
[warn]                                                   ^
[info] done compiling
[warn] three warnings found
[info] done compiling
```

app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala
```diff
@@ -0,0 +1,306 @@
...
+      if (num1 < num2) {
+        (num1, num2)
+      } else {
+        (num2, num1)
+      }
+    }
+
+    forAll(intGen, intGen, Gen.listOfN(1000, numGen), twoNums) {
+      case (slope, yIntercept, outcomes, (x1, x2)) =>
+        def func(outcome: BigDecimal): Satoshis = {
```
Christewart: nit: If I understand, `getExpectedPayout()`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
```diff
@@ -0,0 +1,130 @@
+package org.bitcoins.commons.jsonmodels.dlc
+
+import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.util.NumberUtil
+
+import scala.annotation.tailrec
+
+case class RoundingIntervals(intervalStarts: Vector[(BigDecimal, Long)]) {
```
Christewart, line 14: nit: Scaladoc please

It would be useful to link to the spec section about rounding too.

What is the `(BigDecimal, Long)` input here?

Christewart: Another high level question: 

What do you we call a `OutcomeValueFunction` (PayoutCurve) with rounding applied? Is there a `RoundedPayoutCurve`?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
nkohen: I'm not sure this would add any value as all ranges are inspected symbolically by looking only at endpoints, whereas extending `NumericRange[Long]` would only introduce `Seq` and `Iterable` functionality which I don't currently see any use for

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
nkohen: it isn't conflicting because the word exclusive is modifying the payout range, not the input interval, but I will try to make it more clear

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
nkohen: sure but your interpolated function is allowed to return payouts that are out of bounds and they just get set to `0` and `totalCollateral` respectively.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: I'm not sure I follow, why would changing `isEndpoint` to a type cause a compiler error?

nkohen: oooh, you're talking about making `OutcomeValuePoint` an ADT I think?

nkohen: I'll add it to the list for subsequent PR

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: I'll rename to `DLCPayoutCurve` in subsequent PR

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: Not as descriptive as `DLCPayoutCurveComponent` (also note that it isn't really internal as you can get it from the API), will rename when I rename OutcomeValueFunction

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: payout curves are always a function from outcomes to payouts (which is why I originally called this OutcomeValueFunction) (re-phrased, outcomes are always x-coords and payouts are always y-coords and denominated in Satoshis)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: Not sure what you mean but I added requirements that points are always ascending in outcome so I think that covers your question

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: In this case (and only this case) I think `OutcomeValuePoint` makes much more sense than `PayoutPoint`. I would be willing to use `OutcomePayoutPoint` if you prefer, but a point really is a pair/tuple so I prefer having both in there

app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala
nkohen: My response is that this is unit testing the actual logic for interpolation and going through all of the logical code paths, not testing that DLCs will work in all ranges, that responsibility falls to DLC tests if I'm not mistaken

app-commons-test/src/test/scala/org/bitcoins/commons/dlc/OutcomeValueFunctionTest.scala
nkohen: yeah, I find these quite annoying because scala doesn't have type-level functions and also doesn't provide public access to types like Vector2, Vector3 etc.

These are not actual issues as the input to `nPoints` determines the length of the output (and this is test code anyway, the tests would just fail). Not really sure what to do, if we really cared I could add a `case _ => fail` to every single test :/

nkohen: eh, guess I'll just do that, not as many tests as I thought use `nPoints`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
nkohen: That isn't a thing, the two are separate entities that only come together during processing (a function takes in both an `OutcomeValueFunction` and a `RoundingIntervals`)

Will add a scaladoc though

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
Christewart: That's totally cool with me, but can we get issues filed? Otherwise I know I will forget. 

Some of this stuff is just boiler plate software engineering stuff, so Ben or I can start peeling these off the backlog if we have documentation of it :+1:

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
Christewart: I'm cool with `OutcomePayoutPoint`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,278 @@
...
+  }
+}
+
+/** A quadratic between left and right endpoints defining a piece of a larger payout curve */
+case class OutcomeValueQuadratic(
+    leftEndpoint: OutcomeValuePoint,
+    midpoint: OutcomeValuePoint,
+    rightEndpoint: OutcomeValuePoint)
+    extends OutcomeValueFunctionComponent {
+  override def midpoints: Vector[OutcomeValuePoint] = Vector(midpoint)
```
Christewart: nit: Probably worth making a `val` otherwise a `Vector()` gets allocated everytime this is called.

Christewart: This applies elsewhere in this file too.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
```diff
@@ -0,0 +1,278 @@
...
+
+  override def apply(outcome: BigDecimal): Satoshis = {
+    val value =
+      (outcome - leftEndpoint.outcome) * slope + leftEndpoint.value.toLong
+
+    bigDecimalSats(value)
+  }
+}
+
+/** A quadratic between left and right endpoints defining a piece of a larger payout curve */
```
Christewart: Probably worth adding the equation to the top for people like me to remind use what a quadratic is 

>ax^2 + bx + c

Also a link to wikipedia is probably worthwhile too.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
Christewart: `Interval` rather than `Internal`. I think you misread?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
Christewart: Ok well let's clarify in the scaladoc then

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
Christewart: so one other thing here, I think there should be in invariant here too? Similar to what you added on `OutcomeValueFunction` 

https://github.com/nkohen/bitcoin-s-core/blob/1eda618e18ff071e886bfc25b63636b428454c4d/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala#L10

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
```diff
@@ -0,0 +1,136 @@
...
+
+/** Specifies a list of intervals with corresponding rounding moduli.
+  * In particular, each element (outcome, roundingMod) of intervalStarts
+  * represents the beginning of a new interval at outcome with new modulus roundingMod.
+  *
+  * @see https://github.com/discreetlogcontracts/dlcspecs/blob/8ee4bbe816c9881c832b1ce320b9f14c72e3506f/NumericOutcome.md#rounding-intervals
+  */
+case class RoundingIntervals(intervalStarts: Vector[(BigDecimal, Long)]) {
+
+  def intervalContaining(
```
Christewart, line 23: If I understand correctly, 

```scala
/** 
Given an outcome, this method returns the starting point for the interval, the endpoint for the interval, and the rounding modulos that applies to this interval
*/
```

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
```diff
@@ -0,0 +1,136 @@
...
+      (intervalStart, Long.MaxValue, roundingModulus)
+    } else {
+      val (intervalStart, roundingModulus) = intervalStarts(index)
+      val (intervalEnd, _) = intervalStarts(index + 1)
+
+      (intervalStart, intervalEnd, roundingModulus)
+    }
+  }
+
+  def roundingModulusAt(outcome: BigDecimal): Long = {
```
Christewart, line 45: If I understand correctly

```scala
/**
Given an outcome, we returns the rounding modulos applied to this interval
*/
```

One other question here, why can't we just implement this as 

```scala
intervalContaining(outcome)._3
```

If it is possible, i think we should just do that (simpler, easier to maintain etc)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
Christewart: I would be in favor of minting a type for `(BigDecimal, BigDecimal, Long)`


```scala

case class RoundingInterval(start: BigDecimal, end: BigDecimal, roundingModulos)

```

Now we get nice descriptive names when we use this like

`interval.start` 

or

`interval.roundingModulos`

If you don't like the `RoundingInteval` name that is cool but hopefully we can agree it's worth making a type for this

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
Christewart: Also if you take my suggestion from above it can be 

`intervalContaining(outcome).roundingModulus` :-)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
```diff
@@ -0,0 +1,136 @@
...
+    val roundedPayout = if (mod >= roundingMod / 2.0) {
+      payoutLong + roundingMod - mod
+    } else {
+      payoutLong - mod
+    }
+
+    Satoshis(roundedPayout)
+  }
+
+  def minRoundingWith(other: RoundingIntervals): RoundingIntervals = {
```
Christewart, line 71: 2 things here 

1. Can we add a scaladoc
2. Is this needed? It seems that it is only used in unit test files from me searching on intellij

Christewart: If i was to guess what this is for, is this for the case when us and our peer are trying to negotiate what rounding we are going to use?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala
```diff
@@ -38,74 +27,441 @@ object DLCMessage {
...
+  sealed trait OracleInfo extends TLVSerializable[OracleInfoTLV] {
+    def pubKey: SchnorrPublicKey
+    def nonces: Vector[SchnorrNonce]
+
+    /** The order of the given sigs should correspond to the given outcome. */
+    def verifySigs(
+        outcome: DLCOutcomeType,
+        sigs: Vector[SchnorrDigitalSignature]): Boolean
+
+    def sigPoint(outcome: DLCOutcomeType): ECPublicKey = {
```
Christewart, line 44: So this is one I'm probably going to get wrong, but let's correct my misunderstandings and add a scaladoc! :-).

I also rewrote parts of the method with some type annotations as that made it clearer for me to understand what is going on, but feel free to remove those if you don't like them. But please do take a look at the comment I added wrt to ordering of nonces/outcomes in my scala comment below

```scala
        /** Given an outcome we compute the signature points that correspond to this outcome
     * @return The public key that corresponds with this outcome
      */
    def sigPoint(outcome: DLCOutcomeType): ECPublicKey = {
      //does ordering matter here? For instance with digit decomp events,
      //how do we know that the nonce with zipping with the outcome is for the
      //correct digit?
      val outcomesWithNonce: Vector[(ByteVector, SchnorrNonce)] = {
        outcome.serialized
          .zip(nonces)
      }

      val sigPoints: Vector[ECPublicKey] = outcomesWithNonce.map {
        case (bytes, nonce) =>
          pubKey.computeSigPoint(data = CryptoUtil.sha256(bytes).bytes,
                                 nonce = nonce)
      }

      sigPoints.reduce(_.add(_))
    }
```

core/src/main/scala/org/bitcoins/core/protocol/tlv/DLCOutcomeType.scala
```diff
@@ -0,0 +1,30 @@
...
+  *
+  * If digits.length is less than the the total number of digits to be
+  * signed by the oracle then this outcome represents all outcomes prefixed
+  * by the given digits.
+  */
+case class UnsignedNumericOutcome(digits: Vector[Int]) extends DLCOutcomeType {
+
+  override def serialized: Vector[ByteVector] =
+    digits.map(digit => CryptoUtil.serializeForHash(digit.toString))
+}
```
Christewart, line 30: Shouldn't we have a `SignedNumericOutcome` as well?

core/src/main/scala/org/bitcoins/core/protocol/tlv/DLCOutcomeType.scala
```diff
@@ -0,0 +1,30 @@
...
+    Vector(CryptoUtil.serializeForHash(outcome))
+}
+
+/** An outcome from a multi-nonce unsigned numeric event type.
+  *
+  * If digits.length is less than the the total number of digits to be
+  * signed by the oracle then this outcome represents all outcomes prefixed
+  * by the given digits.
+  */
+case class UnsignedNumericOutcome(digits: Vector[Int]) extends DLCOutcomeType {
```
Christewart, line 26: I think we should have an invariant here of 

```scala

require(digits.forall(_ >= 0), s"Cannot have negative numbers inside of Unsigned numeric outcome, got=${digits.filter(_ < 0)}")

```

core/src/main/scala/org/bitcoins/core/protocol/tlv/DLCOutcomeType.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  override def serialized: Vector[ByteVector] =
```
Christewart: probably worth making this a `val` ? Same with above in `EnumOutcome`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCSignatures.scala
```diff
@@ -1,23 +1,51 @@
...
+    extends SeqWrapper[(TransactionOutPoint, ScriptWitnessV0)]
     with DLCSignatures {
 
-  override protected def wrapped: Map[
-    TransactionOutPoint,
-    Vector[PartialSignature]] = sigs
+  override protected def wrapped: Vector[
+    (TransactionOutPoint, ScriptWitnessV0)] = sigs
+
+  def get(outPoint: TransactionOutPoint): Option[ScriptWitnessV0] = {
```
Christewart, line 20: nit: In general with scala libraries `.get` usually returns the `T` or throws and things like `.find()` return `Option[T]`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCSignatures.scala
```diff
@@ -1,23 +1,51 @@
...
+    sigs.find(_._1 == outPoint).map(_._2)
+  }
+
+  def apply(outPoint: TransactionOutPoint): ScriptWitnessV0 = {
+    get(outPoint).get
+  }
+
+  def merge(other: FundingSignatures): FundingSignatures = {
```
Christewart, line 28: You could maybe add `++` into your `SeqWrapper` type?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
+import org.bitcoins.core.protocol.transaction.{Transaction, WitnessTransaction}
+import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerVirtualByte}
+import org.bitcoins.crypto._
+import scodec.bits.ByteVector
+import ujson._
+
 sealed trait DLCStatus {
-  def eventId: Sha256DigestBE
+  def paramHash: Sha256DigestBE
   def isInitiator: Boolean
```
Christewart, line 29: Not on this PR, but it would be nice to make this into an ADT as well so we can make sure we aren't forgetting a 

```scala
if (isInitiator) { ... } else { ... }
```

somewhere. Also we can then make methods that are _only_ for when you are the initiator, or _only_ for when you are the initatee (right word?) 

Any concept level objections to this or can we add this to the "file an issue for it and deal with it later" bucket?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
-  */
+import org.bitcoins.core.protocol.transaction.{Transaction, WitnessTransaction}
+import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerVirtualByte}
+import org.bitcoins.crypto._
+import scodec.bits.ByteVector
+import ujson._
+
 sealed trait DLCStatus {
-  def eventId: Sha256DigestBE
+  def paramHash: Sha256DigestBE
```
Christewart, line 28: i think all the other names are pretty intuitive but i'm unsure of what this one is for. 

If I am following the code correctly it links here: https://github.com/nkohen/bitcoin-s-core/blob/3a73d650823a51de1d511dbbf8a8fe8522961dd7/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCMessage.scala#L26-L25

So we should add a scaladoc that says 

```scala
  /** The sha256 hash of oracleInfo ++ contractInfo ++ timeoutes */
  def paramHash: Sha256DigestBE
  ```

File: app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala [hunk 35]
```diff
@@ -1,360 +1,676 @@
...
+
+  lazy val statusString: String = state.toString
+
+  def toJson: Value
 }
 
-/** All states other than Offered contain an accept message. */
 sealed trait AcceptedDLCStatus extends DLCStatus {
-  def accept: DLCAccept
+  def contractId: ByteVector
```
Christewart, line 46:
```scala
  override def state: DLCState.Accepted.type
  ```
  
  I think this applies to all subtypes if I understand what is going on here.

File: app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala [hunk 36]
```diff
@@ -1,360 +1,676 @@
...
       offer: DLCOffer,
       accept: DLCAccept,
       sign: DLCSign,
-      fundingTx: Transaction,
-      cet: Transaction,
-      closingTx: Transaction)
-      extends AcceptedDLCStatus {
-    override val statusString: String = "REMOTE CLAIMED"
-  }
+      cet: Transaction): (SchnorrDigitalSignature, DLCOutcomeType) = {
```
Christewart, line 567: Maybe we can just make the `cet: WitnessTransaction` as a parameter?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
-    override val statusString: String = "REMOTE PENALIZED"
-  }
+    def sigFromMsgAndSigs(
+        outcome: DLCOutcomeType,
+        adaptorSig: ECAdaptorSignature,
+        cetSig: ECDigitalSignature): SchnorrDigitalSignature = {
+      val (sigPubKey, numSigs) = outcome match {
+        case EnumOutcome(outcome) =>
+          val sigPoint = oraclePubKey.computeSigPoint(
+            CryptoUtil.sha256(outcome).bytes,
```
Christewart, line 606: Maybe add a helper method on `SchnorrPublicKey.computeSigPoint()` ? It seems we this pattern in a decnet amount of places 


```scala
  def computeSigPoint(data: Sha256Digest, nonce: SchnorrNonce): ECPublicKey = {
    computeSigPoint(data.bytes,nonce)
  }
  ```

File: app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala [hunk 38]
```diff
@@ -1,360 +1,676 @@
...
+        case UnsignedNumericOutcome(digits) =>
+          val sigPoint = digits
+            .zip(rVals.take(digits.length))
+            .map {
+              case (digit, nonce) =>
+                oraclePubKey.computeSigPoint(
+                  CryptoUtil.sha256(digit.toString).bytes,
+                  nonce)
+            }
+            .reduce(_.add(_))
```
Christewart, line 619: going off the theme earlier, it seems that this pattern is used in a lot of places of computing points and then reducing them to one point :-). Maybe we can put this somewhere too.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
+      val (sigPubKey, numSigs) = outcome match {
+        case EnumOutcome(outcome) =>
+          val sigPoint = oraclePubKey.computeSigPoint(
+            CryptoUtil.sha256(outcome).bytes,
+            aggregateR(1))
+
+          (sigPoint, 1)
+        case UnsignedNumericOutcome(digits) =>
+          val sigPoint = digits
+            .zip(rVals.take(digits.length))
```
Christewart, line 612: So this seems a little suspicious to me 

1. Is it ok if we more `rVals` than digits? 
2. I posted earlier with concerns about ordering of `digits` and `rVals` How do we know the ordering is correct here?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
-      offer: DLCOffer,
-      accept: DLCAccept,
-      sign: DLCSign,
-      fundingTx: Transaction,
-      cet: Transaction,
-      penaltyTx: Transaction)
-      extends AcceptedDLCStatus {
-    override val statusString: String = "REMOTE PENALIZED"
-  }
+    def sigFromMsgAndSigs(
```
Christewart, line 599: So here is my best attempt at writing a scaladoc for this

```scala
    /**
     * Completes the given adaptor signature for the given outcome
     * @param outcome the outcome we are signing
     * @param adaptorSig the adaptor signature given from the counterparty that we need to complete
     * @param cetSig our local CET sig that we use
     * @return
     */
    def sigFromMsgAndSigs(
        outcome: DLCOutcomeType,
        adaptorSig: ECAdaptorSignature,
        cetSig: ECDigitalSignature): SchnorrDigitalSignature
```

i'm pretty sure that is wrong though. Could you correct my misconceptions and write an accurate scaladoc? Maybe using the framework i laid out in mine.

Christewart: Could we also move this method `sigFromMsgAndSigs` and `aggregateR` out of `calculateOutcomeAndSig` and into private helper methods?

Christewart (commented):
More code review

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,676 @@
...
+                oraclePubKey.computeSigPoint(
+                  CryptoUtil.sha256(digit.toString).bytes,
+                  nonce)
+            }
+            .reduce(_.add(_))
+
+          (sigPoint, digits.length)
+      }
+
+      val possibleOracleS =
```
Christewart, line 627: So i'm guessing this is called 'possibleOracleS' because there is a chance this does not correspond to a valid schnorr digital signature? 

Is it worth reflecting this in the return type of `sigFromMsgAndSigs` by returning an `Option[]` ?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/OutcomeValueFunction.scala
nkohen: ah, I did, but still prefer component to interval

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
nkohen: Adding a general todo to my list for next PR to replace all the tuples everywhere with types

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
nkohen: yeah, it is called on every DLC

core/src/main/scala/org/bitcoins/core/protocol/tlv/DLCOutcomeType.scala
nkohen: Yeah, just hasn't been done yet

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCSignatures.scala
nkohen: I was trying to mimic `Map`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCSignatures.scala
nkohen: `++` is already inherited from `Iterable`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
Christewart: One other thing, I don't see any unit tests that are testing this method? 

Perhaps they are still up on `adaptor-dlc`, but this seems like an important method so there probably should be stand alone unit tests.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: This already is an ADT? `sealed trait DLCStatus`. `isInitiator` is overriden by every sub-class, I don't see how an `if/else` could be "forgotten"

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: Added a match instead as generally this is only called when you are in a remote-claimed situation and our node setup just gives us a `Transaction` so nice to do it for the caller here

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: Added to todo list for next PR

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: Added to todo list for next PR

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: 1. It is ok and expected, many CETs don't look at all digits
2. added a comment to `nonces` in response to earlier comment saying correct order is required

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: Added moving methods to todo list

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/RoundingIntervals.scala
nkohen: required a small change to `intervalContaining` but can be done this way and changed

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
```diff
@@ -1,360 +1,696 @@
...
+          .map(_._1)
+      case info: DLCMessage.MultiNonceContractInfo =>
+        info.outcomeVec
+          .filter {
+            case (_, amt) =>
+              val amts = Vector(amt, totalCollateral - amt)
+                .filter(_ >= Policy.dustThreshold)
+                .sorted
+
+              Math.abs(
```
Christewart, line 660: Could we break this out into a helper method and add a nice scaladoc to it? 

IIRC on a call you said for this is for handling off by one sat errors. I would imagine this is needed else where in the dlc code ? A scaladoc would be really useful though along with a unit test as this does look suspicious to me.

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -352,6 +353,109 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
 
     backwardsDigits.reverse
   }
+
+  /** Recomposes the input digits into the number they represent.
+    * The input Vector has the most significant digit first and the 1's place last.
+    */
+  def fromDigits(digits: Vector[Int], base: Int, numDigits: Int): Long = {
```
Christewart, line 360: Would be nice to have a unit test

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -352,6 +353,109 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
...
+    digits.indices.foldLeft(0L) { (numSoFar, index) =>
+      numSoFar + digits(index) * math.pow(base, numDigits - 1 - index).toLong
+    }
+  }
+
+  /** Returns a pseudorandom, uniformly distributed long value between 0
+    *  (inclusive) and the specified value (exclusive), drawn from this
+    *  random number generator's sequence.
+    *
```
Christewart, line 382: nit: Could you link to this? https://github.com/scala/scala/blob/4aae0b91cd266f02b9f3d911db49381a300b5103/src/library/scala/util/Random.scala#L131

core/src/main/scala/org/bitcoins/core/util/NumberUtil.scala
```diff
@@ -352,6 +353,109 @@ sealed abstract class NumberUtil extends BitcoinSLogger {
...
+          val res = ord.compare(xe.next(), ye.next())
+          if (res != 0) return res
+        }
+
+        Ordering.Boolean.compare(xe.hasNext, ye.hasNext)
+      }
+    }
+  }
+
+  /** Stolen from Scala 2.13 IndexedSeq::binarySearch */
```
Christewart: nit: Link https://github.com/scala/scala/blob/4aae0b91cd266f02b9f3d911db49381a300b5103/src/library/scala/collection/IndexedSeq.scala#L117

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,105 @@ object ShufflingNonInteractiveFinalizer
...
-    inputInfos: Vector[InputInfo],
-    feeRate: FeeUnit)
+/** Adds a an amount to the output with the given ScriptPubKey
+  * and subtracts that amount in equal proportions from the specified
+  * change ScriptPubKeys.
+  *
+  * This can be useful when you want to account for a future spending
+  * fee in this transaction to get nice output values on the spending tx.
+  */
+case class AddFutureFeeFinalizer(
```
Christewart, line 371: So this LGTM, but it does not have any unit tests. Since it is directly modifying bitcoin transactions it would be wise to have unit tests to define expected behavior. Edge cases to think of 

1. Given spk is not a member of `RawTxBuilderResult` 
2. Given `changeSPKs` are not members of `RawTxBuilderResult`
3. Fee rate is absurly high that spends all funds in changeSPK (is this ok?)
4. Happy path

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,105 @@ object ShufflingNonInteractiveFinalizer
...
+    Future.fromTry(txT)
+  }
+}
+
+/** Subtracts the given fee from the output with the given ScriptPubKey.
+  * This can be useful if you are constructing a transaction without
+  * considering fees and have some after-the-fact external formula for
+  * computing fees and they need the removed.
+  */
+case class SubtractFromOutputFinalizer(spk: ScriptPubKey, subAmt: CurrencyUnit)
```
Christewart, line 420: Same comment wrt to unit tests as this is directly modifying bitcoin transactions

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -361,33 +361,105 @@ object ShufflingNonInteractiveFinalizer
...
+          s"Did not find expected SPK $spk in ${txBuilderResult.outputs.map(_.scriptPubKey)}"))
+    }
+  }
+}
+
+/** Assumes the input transaction has had no fee considerations
+  * and subtracts the estimated fee in equal portions from the
+  * outputs with the specified ScriptPubKeys
+  */
+case class SubtractFeeFromOutputsFinalizer(
```
Christewart, line 442: Same comment wrt to unit tests

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +492,81 @@ object SubtractFeeFromOutputsFinalizer {
...
+    offerChangeSPK: ScriptPubKey,
+    acceptInputs: Vector[DualFundingInput],
+    acceptPayoutSPK: ScriptPubKey,
+    acceptChangeSPK: ScriptPubKey,
+    feeRate: SatoshisPerVirtualByte,
+    fundingSPK: ScriptPubKey)
+    extends RawTxFinalizer {
+
+  /** @see https://github.com/discreetlogcontracts/dlcspecs/blob/8ee4bbe816c9881c832b1ce320b9f14c72e3506f/Transactions.md#fees */
+  private def computeFees(
```
Christewart, line 523: It might be useful to break this out into it's own utility method somehwere so it can easily be unit tested. I doubt there is logic errors in in it currently as we have tested it pretty thorogoughly on the `adaptor-dlc` branch, but if we need to modify it in the future it would be nice to have unit tests to detect regressions. Then we can determine if it was an intended regression (DLC spec changed for whatever reason) or an uninteded one.

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +492,81 @@ object SubtractFeeFromOutputsFinalizer {
...
+
+/** Finalizes a dual-funded transaction given the DualFundingInputs
+  * from both parties, their change spks and the funding scriptpubkey
+  * for the dual funded protocol.
+  *
+  * This includes adding the future fee of spending transactions
+  * to the funding output as well as subtracting relevant fees
+  * from the change outputs. This finalizer filters dust outputs.
+  */
+case class DualFundingTxFinalizer(
```
Christewart, line 511: same comment wrt to unit tests

core/src/main/scala/org/bitcoins/core/wallet/builder/RawTxFinalizer.scala
```diff
@@ -422,7 +492,81 @@ object SubtractFeeFromOutputsFinalizer {
...
+    val subtractOfferFundingFee =
+      SubtractFromOutputFinalizer(offerChangeSPK, offerFundingFee)
+    val subtractAcceptFundingFee =
+      SubtractFromOutputFinalizer(acceptChangeSPK, acceptFundingFee)
+
+    addOfferFutureFee
+      .andThen(addAcceptFutureFee)
+      .andThen(subtractOfferFundingFee)
+      .andThen(subtractAcceptFundingFee)
+      .andThen(FilterDustFinalizer)
```
Christewart, line 568: So just to be clear, `offerPayoutSPK` and `acceptPayoutSPK` will get removed here if they are below dust? 

If this is intended behavior, we should add a unit test making sure we do this. Also in the future we detect regressions if this logic changes.

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -163,6 +172,86 @@ object InputInfo {
...
+
+    asmByteSize + varIntSize.toInt
+  }
+
+  /** Computes the byteSize of witness/scriptSignature for the given info,
+    * and also returns the number of stack elements for the P2WSH case.
+    */
+  private def maxScriptSigLenAndStackHeight(
+      info: InputInfo,
+      forP2WSH: Boolean): (Int, Int) = {
```
Christewart: nit: make a type for the return type too

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -163,6 +172,86 @@ object InputInfo {
     (preImages, conditionalPath)
   }
 
+  /** Returns the maximum byteSize of any resulting ScriptSignature */
+  def maxScriptSigLen(info: InputInfo): Int = {
```
Christewart, line 176: Just a general note with `maxScriptSigLen`, `maxScriptSigLenAndStackHeight` and `maxWitnessLen` would probably be useful to move to BitcoinScriptUtil :-)

https://github.com/bitcoin-s/bitcoin-s/blob/4186be51e00fafadcc825899284d84ecacce1676/core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala#L47

db-commons/src/main/resources/reference.conf
```diff
@@ -111,6 +111,22 @@ bitcoin-s {
         rpcport = 9999
     }
 
+  dlc = ${bitcoin-s.sqlite}
```
Christewart: Hmm, what is this for? Maybe i am missing this, but is other database code for dlc pulled down?

Christewart (commented):
Last batch! :-)

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: so the issue is that this method itself only does extraction and doesn't know whether or not its extracted value corresponds to a signature or not (this is done by the caller method). As such there isn't a great way of making this an `Option`

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
nkohen: Everything there seems to handle scripts, whereas this is purely handling of metadata so it seems to be in the right place in my mind

db-commons/src/main/resources/reference.conf
nkohen: it is not, this likely has no effect. I can get rid of it if you want?

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
Christewart: I guess to me it seems like other places might like to know the `maxScriptSigLen` etc without having an `InputInfo`. Same with the others.

Christewart: Ah -- it's in the companion object at least.

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
nkohen: I think maybe I should just add an in-line comment explaining why this is here. This is not used by any other dlc code and I don't know what a unit test would look like seeing as this is a purely math function (and a simple one at that).

db-commons/src/main/resources/reference.conf
Christewart: Yeah probably best for now as it's confusing for me and it might cause merge conflicts for you if things change on the adaptor-dlc branch

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/DLCStatus.scala
Christewart: Ok that comment makes a lot of sense, maybe just embed this comment in the code on line 603 ?

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/dlc/CETCalculator.scala
Christewart: This extra context makes helps me put the puzzle together, can you add this in the scaladoc so next time I read through this I have that added context?

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
nkohen: And this literally takes an `InputInfo` which is the means for computing this value, if you don't have an `InputInfo` you can't call this method anyway.

Christewart: Oh! I misread that. I assumed it took in a `ScriptWitness` or `ScriptSignature`

PR #2340, opened 2020-12-07
Give raw tx, tx data, and link to broadcast for txs

benthecarman:
image

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
@@ -29,11 +29,16 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea) {
         // if it was print the error
         str
       case Success(tx) =>
-        s"""|Transaction: ${tx.hex}
+        s"""|TxId: ${tx.txIdBE.hex}
             |
-            |TxId: ${tx.txIdBE.hex}
+            |url: ${GlobalData.buildTxUrl(tx.txIdBE)}
+            |
+            |If the tx doesn't show up after a few minutes at this url you may need to manually
+            |broadcast the tx with the full hex below
             |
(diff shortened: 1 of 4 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -15,11 +17,27 @@ import scalafx.collections.ObservableBuffer
...
   var taskRunner: TaskRunner = _
 
+  lazy val txPrintFunc: String => String = str => {
+    // See if it was an error or not
+    Try(Transaction.fromHex(str)) match {
+      case Failure(_) =>
+        // if it was print the error
+        str
+      case Success(tx) =>
+        s"""|Transaction: ${tx.hex}
```
Christewart: so I would actually invert these and have the text say something like 

```scala
       s"""|TxId: ${tx.txIdBE.hex}
            |
            |url: ${GlobalData.buildTxUrl(tx.txIdBE)}
            |
            |If the tx doesn't show up after a few minutes at this url you may need to manually 
            |broadcast the tx with the full hex below
            |
            |Link to broadcast: ${GlobalData.broadcastUrl}
            |
            |Transaction: ${tx.hex}

```

the `buildTxUrl` function 

```scala
  /** Builds a url for the blockstream explorer to view the tx */
  def buildTxUrl(
      txid: DoubleSha256DigestBE,
      network: NetworkParameters): String = {
    network match {
      case MainNet =>
        s"https://blockstream.info/tx/${txid.hex}"
      case TestNet3 =>
        s"https://blockstream.info/testnet/tx/${txid.hex}"
      case net @ (RegTest | SigNet) => 
        s"Broadcast from your own node on $net"
    }
```

That should probably be put in `core` somewhere, as i would imagine that will have uses elsewhere.

Christewart (changes requested):

Christewart (approved):

PR #2345, opened 2020-12-09
Multi Wallet support

benthecarman:
Needed to change how some stuff if configured because now the jdbc url is calculated programmatically. 

Todo: Docs, didn't do these yet as they will likely need to change after review

### DIFF
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -32,14 +32,15 @@ bitcoin-s {
     mode = neutrino # neutrino, spv
     # this config key is read by Slick
     db {
-      name = nodedb.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.node.db.path}${bitcoin-s.node.db.name}
+      name = nodedb
+      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
-    #   url = "jdbc:postgresql://localhost:5432/node"
+    #   name = node
+    #   driverName = postgresql
     #   driver = org.postgresql.Driver
-    #   username = postgres
+    #   user = postgres
     #   password = ""
     # }
   }
@@ -60,14 +61,15 @@ bitcoin-s {
     }
     # this config key is read by Slick
     db {
-      name = chaindb.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.chain.db.path}${bitcoin-s.chain.db.name}
+      name = chaindb
+      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
-    #   url = "jdbc:postgresql://localhost:5432/chain"
+    #   name = chain
+    #   driverName = postgresql
     #   driver = org.postgresql.Driver
-    #   username = postgres
+    #   user = postgres
     #   password = ""
     # }
   }
@@ -94,14 +96,15 @@ bitcoin-s {
 
     # this config key is read by Slick
     db {
-      name = walletdb.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.wallet.db.path}${bitcoin-s.wallet.db.name}
+      name = walletdb
+      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
-    #   url = "jdbc:postgresql://localhost:5432/wallet"
+    #   name = wallet
+    #   driverName = postgresql
     #   driver = org.postgresql.Driver
-    #   username = postgres
+    #   user = postgres
     #   password = ""
     # }
   }
@@ -115,14 +118,15 @@ bitcoin-s {
   oracle {
     # this config key is read by Slick
     db {
-      name = oracle.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.oracle.db.path}${bitcoin-s.oracle.db.name}
+      name = oracle
+      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
-    #   url = "jdbc:postgresql://localhost:5432/oracle"
+    #   name = oracle
+    #   driverName = postgresql
     #   driver = org.postgresql.Driver
-    #   username = postgres
+    #   user = postgres
     #   password = ""
     # }
   }
@@ -131,8 +135,8 @@ bitcoin-s {
   test {
     # this config key is read by Slick
     db {
-      name = testdb.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.test.db.path}${bitcoin-s.test.db.name}
+      name = testdb
+      driverName = sqlite
     }
   }
 }
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -63,6 +63,7 @@ bitcoin-s {
     }
     # this config key is read by Slick
     db {
+        name = chaindb
     }
     # PostgreSQL example:
     # db {
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -2,7 +2,7 @@ bitcoin-s {
 
   network = regtest # regtest, testnet3, mainnet, signet
   datadir = ${HOME}".bitcoin-s/"
-  sqlite = {
+  dbDefault = {
     dataSourceClass = slick.jdbc.DatabaseUrlDataSource
     profile = "slick.jdbc.SQLiteProfile$"
 
@@ -25,7 +25,7 @@ bitcoin-s {
     }
   }
 
-  node = ${bitcoin-s.sqlite}
+  node = ${bitcoin-s.dbDefault}
 
   node {
     peers = ["localhost"] # a list of peer addresses in form "hostname:portnumber"
@@ -48,7 +48,7 @@ bitcoin-s {
   }
 
 
-  chain = ${bitcoin-s.sqlite}
+  chain = ${bitcoin-s.dbDefault}
   chain {
     neutrino {
       filter-header-batch-size.default = 2000
@@ -74,7 +74,7 @@ bitcoin-s {
     # }
   }
 
-  wallet = ${bitcoin-s.sqlite}
+  wallet = ${bitcoin-s.dbDefault}
     # settings for wallet module
   wallet {
       defaultAccountType = legacy # legacy, segwit, nested-segwit
@@ -113,7 +113,7 @@ bitcoin-s {
         rpcport = 9999
     }
 
-  oracle = ${bitcoin-s.sqlite}
+  oracle = ${bitcoin-s.dbDefault}
   oracle {
     # this config key is read by Slick
     db {
@@ -129,7 +129,7 @@ bitcoin-s {
     # }
   }
 
-  test = ${bitcoin-s.sqlite}
+  test = ${bitcoin-s.dbDefault}
   test {
     # this config key is read by Slick
     db {
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -11,7 +11,6 @@ bitcoin-s {
       //https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
       path = ${bitcoin-s.datadir}/${bitcoin-s.network}/
       driver = org.sqlite.JDBC
-      driverName = sqlite
       user = ""
       password = ""
       host = localhost
@@ -40,7 +39,6 @@ bitcoin-s {
     # PostgreSQL example:
     # db {
     #   name = node
-    #   driverName = postgresql
     #   driver = org.postgresql.Driver
     #   user = postgres
     #   password = ""
@@ -68,7 +66,6 @@ bitcoin-s {
     # PostgreSQL example:
     # db {
     #   name = chain
-    #   driverName = postgresql
     #   driver = org.postgresql.Driver
     #   user = postgres
     #   password = ""
@@ -102,7 +99,6 @@ bitcoin-s {
     # PostgreSQL example:
     # db {
     #   name = wallet
-    #   driverName = postgresql
     #   driver = org.postgresql.Driver
     #   user = postgres
     #   password = ""
@@ -123,7 +119,6 @@ bitcoin-s {
     # PostgreSQL example:
     # db {
     #   name = oracle
-    #   driverName = postgresql
     #   driver = org.postgresql.Driver
     #   user = postgres
     #   password = ""
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -11,8 +11,11 @@ bitcoin-s {
       //https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
       path = ${bitcoin-s.datadir}/${bitcoin-s.network}/
       driver = org.sqlite.JDBC
+      driverName = sqlite
       user = ""
       password = ""
+      host = localhost
+      port = 5432
 
       # this needs to be set to 1 for SQLITE as it does not support concurrent database operations
       # see: https://github.com/bitcoin-s/bitcoin-s/pull/1840
@@ -33,7 +36,6 @@ bitcoin-s {
     # this config key is read by Slick
     db {
       name = nodedb
-      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
@@ -61,8 +63,6 @@ bitcoin-s {
     }
     # this config key is read by Slick
     db {
-      name = chaindb
-      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
@@ -97,7 +97,6 @@ bitcoin-s {
     # this config key is read by Slick
     db {
       name = walletdb
-      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
@@ -119,7 +118,6 @@ bitcoin-s {
     # this config key is read by Slick
     db {
       name = oracle
-      driverName = sqlite
     }
     # PostgreSQL example:
     # db {
@@ -136,7 +134,6 @@ bitcoin-s {
     # this config key is read by Slick
     db {
       name = testdb
-      driverName = sqlite
     }
   }
 }
--- a/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
@@ -4,14 +4,11 @@ import com.typesafe.config._
 import org.bitcoins.core.config._
 import org.bitcoins.core.protocol.blockchain.BitcoinChainParams
 import org.bitcoins.core.util.{BitcoinSLogger, FutureUtil, StartStopAsync}
-import org.bitcoins.db.DatabaseDriver.{PostgreSQL, SQLite}
-import slick.basic.DatabaseConfig
-import slick.jdbc.JdbcProfile
 
 import java.nio.file.{Files, Path, Paths}
 import scala.concurrent.Future
+import scala.util.Properties
 import scala.util.matching.Regex
-import scala.util.{Failure, Properties, Success, Try}
 
 /**
   * Everything needed to configure functionality
@@ -33,11 +30,6 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
     */
   override def start(): Future[Unit] = {
     FutureUtil.unit
-  }
-
-  /** Releases the thread pool associated with this AppConfig's DB */
-  override def stop(): Future[Unit] = {
-    Future.successful(slickDbConfig.db.close())
   }
 
   /** Sub members of AppConfig should override this type with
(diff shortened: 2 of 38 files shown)


Christewart:
A CI failure:

https://github.com/bitcoin-s/bitcoin-s/pull/2345/checks?check_run_id=1535372653#step:5:306

and a coverage failure: 

https://github.com/bitcoin-s/bitcoin-s/pull/2345/checks?check_run_id=1535372607#step:5:327

Christewart:
> A CI failure:
> 
> https://github.com/bitcoin-s/bitcoin-s/pull/2345/checks?check_run_id=1535372653#step:5:306

If this ends up being something not obvious, let me know. This seems similar to the bugs i've seen on the sb-api side of things

https://gitlab.com/suredbits/suredbits-api/-/issues/502

db-commons/src/main/resources/reference.conf
```diff
@@ -32,14 +32,15 @@ bitcoin-s {
     mode = neutrino # neutrino, spv
     # this config key is read by Slick
     db {
-      name = nodedb.sqlite
-      url = "jdbc:sqlite:"${bitcoin-s.node.db.path}${bitcoin-s.node.db.name}
+      name = nodedb
+      driverName = sqlite
```
Christewart: Is there a reason this can't be moved up to `bitcoin-s.sqlite` so that we don't have to repeat it in all of the subsequent configurations?

Christewart: another thing that seems like it might be incorrect is the `bitcoin-s.sqlite.db.path` ?

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -159,31 +159,110 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
...
+
+    driver match {
+      case SQLite     => s"$name.sqlite"
+      case PostgreSQL => name
+    }
+  }
+
+  /** The host of our postgresql database */
+  lazy val dbHost: String = {
+    val hostOpt = getConfigStringOpt(s"bitcoin-s.$moduleName.db.host")
```
Christewart: So i don't see any hosts defined in the example reference.conf files? I would definitely be in favor of us explicitly encoding this into our example `reference.conf` files so it's easy to understand how the configuration works.

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -159,31 +159,110 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
...
+
+    hostOpt match {
+      case Some(host) => host
+      case None       => "localhost"
+    }
+  }
+
+  /** The port number of our postgresql database */
+  lazy val dbPort: Int = {
+    val portOpt = config.getIntOpt(s"bitcoin-s.$moduleName.db.port")
```
Christewart: same comment here, we should make our example `reference.conf`s showcase all options. 

Maybe this is what you mean by this though

>Todo: Docs, didn't do these yet as they will likely need to change after review

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -159,31 +159,110 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
...
+    case SQLite =>
+      None
+  }
+
   lazy val slickDbConfig: DatabaseConfig[JdbcProfile] = {
+    // Create overrides if modules want to change their path or db name
+    val overrideConf = ConfigFactory.parseString {
+      s"""
+         |bitcoin-s {
+         |  $moduleName {
```
Christewart: Hmmm. I'm not sure if I understand what this is for. 

Shouldn't this allow just be put into the configuration files?

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -294,7 +294,23 @@ trait BitcoinSWalletTest extends BitcoinSFixture with EmbeddedPg {
 
   def withWalletConfig(test: OneArgAsyncTest): FutureOutcome = {
     val builder: () => Future[WalletAppConfig] = () => {
-      val walletConf = config.walletConf
+      val baseConf = config.walletConf
+      val walletNameOpt = if (NumberGenerator.bool.sampleSome) {
```
Christewart, line 298: So this will add named wallets to test cases that use fixtures right? The probability that we are using one is 1/2 ? 

If that is the case, this is nice! I always love it when our test infra get easily be modified to test a new feature :-)

key-manager-test/src/test/scala/org/bitcoins/keymanager/config/KeyManagerAppConfigTest.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    config.start().map { _ =>
+      assert(Files.exists(seedPath), "We should not delete the old seed!")
+      assert(
+        Files.exists(
+          tempDir
+            .resolve(WalletStorage.SEED_FOLDER_NAME)
+            .resolve(WalletStorage.ENCRYPTED_SEED_FILE_NAME)))
+    }
+  }
```
Christewart, line 109: probably should add a test case just making sure we can have multiple seeds in the `seeds` directory and they get written to the write place?

Christewart (changes requested):
Overall looking really nice! As you noted in your original description it needs docs, but I think the layout is pretty good.

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
benthecarman: Need to do this so when we override `dbPath`, `dbName`, and `jdbcUrl` they are reflected in the config.

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
benthecarman: Yep!

db-commons/src/main/resources/reference.conf
benthecarman: Moved the `driverName` to `bitcoin-s.sqlite.db`.

The `bitcoin-s.sqlite.db.path` is correct, we want our dbs to all go to the same place unless we override it in our AppConfig.

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
Christewart: Interesting, so you are saying that our configuration can be internally inconsistent when we specify `walletName`?

https://bitcoin-s.org/docs/next/config/configuration#database-configuration

benthecarman: Yeah, because we need to add on the wallet name to the `dbPath`

db-commons/src/main/resources/reference.conf
```diff
@@ -11,8 +11,11 @@ bitcoin-s {
       //https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
       path = ${bitcoin-s.datadir}/${bitcoin-s.network}/
       driver = org.sqlite.JDBC
+      driverName = sqlite
       user = ""
       password = ""
+      host = localhost
+      port = 5432
```
Christewart, line 17: I don't think sqlite has the concept of a host/port as it is just a file on your filesystem?

db-commons/src/main/resources/reference.conf
```diff
@@ -32,14 +35,14 @@ bitcoin-s {
...
+      name = nodedb
     }
     # PostgreSQL example:
     # db {
-    #   url = "jdbc:postgresql://localhost:5432/node"
+    #   name = node
+    #   driverName = postgresql
     #   driver = org.postgresql.Driver
-    #   username = postgres
+    #   user = postgres
```
Christewart, line 43: thank you for fixing this :+1:

db-commons/src/main/scala/org/bitcoins/db/AppConfig.scala
```diff
@@ -159,31 +159,110 @@ abstract class AppConfig extends StartStopAsync[Unit] with BitcoinSLogger {
     baseDatadir.resolve(lastDirname)
   }
 
+  lazy val driver: DatabaseDriver = {
```
Christewart: not neccesarily related to this PR, but we probably want to move all this database stuff into it's own trait? Not all things that subclasses of `AppConfig` have database components (like `KeyManagerAppConfig`). 

maybe something like 

```scala

trait DbAppConfig extends AppConfig { 
  lazy val driver: DatabaseDriver = {
      val driverName =
        getConfigString(s"bitcoin-s.$moduleName.db.driverName")
      DatabaseDriver.fromString(driverName)
    }

    //more methods down here...
}

```

key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
```diff
@@ -30,12 +29,40 @@ case class KeyManagerAppConfig(
...
+    }
+    baseDatadir
+      .resolve(WalletStorage.SEED_FOLDER_NAME)
+      .resolve(s"$prefix${WalletStorage.ENCRYPTED_SEED_FILE_NAME}")
   }
 
-  override def start(): Future[Unit] = FutureUtil.unit
+  override def start(): Future[Unit] = {
+    walletNameOpt match {
+      case Some(_) => FutureUtil.unit
```
Christewart: Hmm... so why don't we need to create it -- or at least see if it exists -- in this case too?

wallet-test/src/test/scala/org/bitcoins/wallet/models/AccountDAOTest.scala
```diff
@@ -4,9 +4,9 @@ import org.bitcoins.core.api.wallet.db.AccountDb
 import org.bitcoins.testkit.Implicits._
 import org.bitcoins.testkit.core.gen.CryptoGenerators
 import org.bitcoins.testkit.fixtures.WalletDAOFixture
-import org.bitcoins.testkit.wallet.{BitcoinSWalletTest, WalletTestUtil}
+import org.bitcoins.testkit.wallet.WalletTestUtil
 
-class AccountDAOTest extends BitcoinSWalletTest with WalletDAOFixture {
+class AccountDAOTest extends WalletDAOFixture {
```
Christewart, line 9: Why did we need to remove `extends BitcoinSWalletTest` on all of these?

Christewart (commented):
I think we are starting to get closer on it being worthwhile to start writing up docs

wallet-test/src/test/scala/org/bitcoins/wallet/models/AccountDAOTest.scala
benthecarman: It was putting 2 different `WalletAppConfig`s in context here, 1 from `BitcoinSWalletTest` and 1 from `WalletDAOFixture`.

key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
benthecarman: No, that is handled by the wallet currently, we should probably move that to here in a future PR

benthecarman: #2363

db-commons/src/main/resources/reference.conf
benthecarman: No, I think we should maybe rename this to something like `dbDefault` becuase it has stuff for postgres as well (`host`, `port`, `user`, `password`)

docs/config/configuration.md
```diff
@@ -143,6 +143,11 @@ bitcoin-s {
 
     # settings for wallet module
     wallet {
+        # You can have multiple wallets by setting a different
+        # wallet name for each of them. They will each have
+        # their own seed and database.
```
Christewart: nit: In the case of sqlite each wallet will have it's own database, in the case of postgres backend each wallet will have it's own schema. Both will have seeds placed in `datadir/seeds/`

Christewart: Also are there any known restrictions on naming a wallet ? Can i use `-` `?` `_` etc

Christewart (commented):
We should probably add a `wallet/multi-wallet.md` doc. It should 

1. Explain at a high level how multi wallet bitcoin-s works
2. Show example of how to create the another wallet in Scala code ([similar to how we do this with single wallets](https://bitcoin-s.org/docs/next/wallet/wallet#example))
3. Show how to create a new wallet by adjusting configuration (this is sort of done already)

I'm fine with doing this on a separate PR if you want, we just need to file an issue

Christewart:
EDIT: I believe this can be fixed by just adding the following [here](https://github.com/benthecarman/bitcoin-s/blob/7e1383fcf7d37f472b114064b7c16e4ca723f1e2/db-commons/src/main/resources/reference.conf#L66)

```
    db {
      name = "chaindb"
    }
```

It seems this breaks all existing configurations for `bitcoin-s.conf`. 

The only two things I've modified in my `bitcoin-s.conf` file are the following

1. adjusting node configuration to connect to regtest node
2. adding `bitcoin-s.wallet.walletName = "test-wallet"`


>[info] com.typesafe.config.ConfigException$Missing: reference.conf @ jar:file:/home/suredbits/dev/bitcoin-s/target/bg-jobs/sbt_3178cb11/target/38bf95d6/ce808e69/bitcoin-s-db-commons_2.13-0.4.0+166-126ac1c5-SNAPSHOT.jar!/reference.conf: 9: No configuration setting found for key 'bitcoin-s.chain.db.name'
[info]  at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:157)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:175)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:181)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:181)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:181)
[info]  at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:189)
[info]  at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:194)
[info]  at com.typesafe.config.impl.SimpleConfig.getString(SimpleConfig.java:251)
[info]  at org.bitcoins.db.AppConfig.getConfigString(AppConfig.scala:159)
[info]  at org.bitcoins.db.DbAppConfig.dbName$lzycompute(DbAppConfig.scala:48)
[info]  at org.bitcoins.db.DbAppConfig.dbName(DbAppConfig.scala:47)
[info]  at org.bitcoins.db.DbAppConfig.jdbcUrl$lzycompute(DbAppConfig.scala:28)
[info]  at org.bitcoins.db.DbAppConfig.jdbcUrl(DbAppConfig.scala:25)
[info]  at org.bitcoins.db.DbManagement.org$bitcoins$db$DbManagement$$flyway(DbManagement.scala:39)
[info]  at org.bitcoins.db.DbManagement.org$bitcoins$db$DbManagement$$flyway$(DbManagement.scala:17)
[info]  at org.bitcoins.chain.config.ChainAppConfig.org$bitcoins$db$DbManagement$$flyway$lzycompute(ChainAppConfig.scala:19)
[info]  at org.bitcoins.chain.config.ChainAppConfig.org$bitcoins$db$DbManagement$$flyway(ChainAppConfig.scala:19)
[info]  at org.bitcoins.db.DbManagement.migrate(DbManagement.scala:160)
[info]  at org.bitcoins.db.DbManagement.migrate$(DbManagement.scala:157)
[info]  at org.bitcoins.chain.config.ChainAppConfig.migrate(ChainAppConfig.scala:19)
[info]  at org.bitcoins.chain.config.ChainAppConfig.$anonfun$start$1(ChainAppConfig.scala:72)
[info]  at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
[info]  at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:48)
[info]  at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
[info]  at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
[info]  at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
[info]  at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
[error] com.typesafe.config.ConfigException$Missing: reference.conf @ jar:file:/home/suredbits/dev/bitcoin-s/target/bg-jobs/sbt_3178cb11/target/38bf95d6/ce808e69/bitcoin-s-db-commons_2.13-0.4.0+166-126ac1c5-SNAPSHOT.jar!/reference.conf: 9: No configuration setting found for key 'bitcoin-s.chain.db.name'
[info]  at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
[info]  at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
[error]         at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:157)
[info] 2020-12-15T18:42:11UTC ERROR [BitcoinSServerMain] Error on server startup!
[info] com.typesafe.config.ConfigException$Missing: reference.conf @ jar:file:/home/suredbits/dev/bitcoin-s/target/bg-jobs/sbt_3178cb11/target/38bf95d6/ce808e69/bitcoin-s-db-commons_2.13-0.4.0+166-126ac1c5-SNAPSHOT.jar!/reference.conf: 9: No configuration setting found for key 'bitcoin-s.chain.db.name'
[info]  at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:157)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:175)
[info]  at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:181)

Christewart:
It appears that we do not move the existing seed into the `seeds/` directory. I've thought about this a bit more and here is what I propose we do:

We copy the existing seed in the `seeds/` directory and start referencing that with our code, but we DO NOT delete the existing seed at the root of our datadir. I think this is the safest best thing to do? 

The reason I think we should at least copy the existing seed into the `seeds/` directory is so that we can then tell people that they just need to backup the `seeds/` directory. If we don't copy it into there, all seeds will be there _except_ the default one which is the most important!

This might have implications for restoring a wallet, we should look through the documentation and update accordingly.

benthecarman:
> It appears that we do not move the existing seed into the `seeds/` directory. I've thought about this a bit more and here is what I propose we do:
> 
> We copy the existing seed in the `seeds/` directory and start referencing that with our code, but we DO NOT delete the existing seed at the root of our datadir. I think this is the safest best thing to do?
> 
> The reason I think we should at least copy the existing seed into the `seeds/` directory is so that we can then tell people that they just need to backup the `seeds/` directory. If we don't copy it into there, all seeds will be there _except_ the default one which is the most important!
> 
> This might have implications for restoring a wallet, we should look through the documentation and update accordingly.

This is what this PR does

https://github.com/bitcoin-s/bitcoin-s/pull/2345/files#diff-eb4c4f02029e7d78a948a22d0453d3bd550ceef7eca3f10fda01ffe7a5ba84a2R48

Christewart:
> > It appears that we do not move the existing seed into the `seeds/` directory. I've thought about this a bit more and here is what I propose we do:
> > We copy the existing seed in the `seeds/` directory and start referencing that with our code, but we DO NOT delete the existing seed at the root of our datadir. I think this is the safest best thing to do?
> > The reason I think we should at least copy the existing seed into the `seeds/` directory is so that we can then tell people that they just need to backup the `seeds/` directory. If we don't copy it into there, all seeds will be there _except_ the default one which is the most important!
> > This might have implications for restoring a wallet, we should look through the documentation and update accordingly.
> 
> This is what this PR does

It doesn't work then as of 126ac1c5ab4c468ef700f87fd03d6c709ee909a7 

Screenshot from 2020-12-15 13-33-51

benthecarman:
> > > It appears that we do not move the existing seed into the `seeds/` directory. I've thought about this a bit more and here is what I propose we do:
> > > We copy the existing seed in the `seeds/` directory and start referencing that with our code, but we DO NOT delete the existing seed at the root of our datadir. I think this is the safest best thing to do?
> > > The reason I think we should at least copy the existing seed into the `seeds/` directory is so that we can then tell people that they just need to backup the `seeds/` directory. If we don't copy it into there, all seeds will be there _except_ the default one which is the most important!
> > > This might have implications for restoring a wallet, we should look through the documentation and update accordingly.
> > 
> > 
> > This is what this PR does
> 
> It doesn't work then as of [126ac1c](https://github.com/bitcoin-s/bitcoin-s/commit/126ac1c5ab4c468ef700f87fd03d6c709ee909a7)
> 
> Screenshot from 2020-12-15 13-32-46

It does when you load the default wallet

Christewart:
> The reason I think we should at least copy the existing seed into the `seeds/` directory is so that we can then tell people that they just need to backup the `seeds/` directory. If we don't copy it into there, all seeds will be there _except_ the default one which is the most important!

Christewart:
So I tested sqlite and everything seems to work on 126ac1c5ab4

Postgres is giving me problems though. I've tried two things so far on 126ac1c5ab4

I initially tried to start postgres with the following config 

```
bitcoin-s {
    datadir = /tmp/multi-wallet-test/
    network = regtest# regtest, testnet3, mainnet
    
    common {
        profile = "slick.jdbc.PostgresProfile$"
        db {
            driver = org.postgresql.Driver
            driveName = postgresql

            # these 3 options will result into a jdbc url of
            # "jdbc:postgresql://localhost:5432/bitcoin-s"
            name = database
            host = localhost
            port = 5432

            user = "chris"
            password = ""
            numThreads = 5
        }
    }

    chain.profile = ${bitcoin-s.common.profile}
    chain.db = ${bitcoin-s.common.db}
 
    node.profile = ${bitcoin-s.common.profile}
    node.db = ${bitcoin-s.common.db}
    
    wallet.profile = ${bitcoin-s.common.profile}
    wallet.db = ${bitcoin-s.common.db}

    oracle.profile = ${bitcoin-s.common.profile}
    oracle.db = ${bitcoin-s.common.db}
    # settings for wallet module
    wallet {
	walletName = "test-wallet"
	//more config ...
    }
}
```

I cannot seem to get it to connect to postres, here is what my logs look like 

>2020-12-15T21:48:20UTC INFO [VersionPrinter] Flyway Community Edition 6.4.2 by Redgate
2020-12-15T21:48:20UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/tmp/multi-wallet-test/regtest/test-wallet/database.sqlite (SQLite 3.34)
2020-12-15T21:48:20UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/tmp/multi-wallet-test/regtest/database.sqlite (SQLite 3.34)
2020-12-15T21:48:20UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/tmp/multi-wallet-test/regtest/database.sqlite (SQLite 3.34)
2020-12-15T21:48:20UTC INFO [DbValidate] Successfully validated 5 migrations (execution time 00:00.016s)
2020-12-15T21:48:20UTC INFO [DbValidate] Successfully validated 9 migrations (execution time 00:00.016s)
2020-12-15T21:48:20UTC WARN [NodeAppConfig] Failed to apply first round of migrations, attempting baseline
//...

db-commons/src/main/resources/reference.conf
```diff
@@ -11,8 +11,11 @@ bitcoin-s {
       //https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
       path = ${bitcoin-s.datadir}/${bitcoin-s.network}/
       driver = org.sqlite.JDBC
+      driverName = sqlite
```
rorp: The driver name is a part of JDBC URL, why do we need a separate parameter for it?

Christewart:
Ok tested ACK for postgres on e1fe15361a728f388cd14f0d541df3cbab411b38 with postgres :tada: :tada: 

Very nice job Ben. 

Screenshot from 2020-12-16 10-35-54

Christewart (approved):

rorp (approved):

PR #2346, opened 2020-12-10
Don't allow fee unit gen to be 0

benthecarman:
We had a CI failure because we tried to broadcast a 0 fee transaction. Bitcoind will always reject these

https://github.com/bitcoin-s/bitcoin-s/runs/1531792984#step:5:840

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/core/gen/FeeUnitGen.scala
@@ -79,7 +79,7 @@ abstract class FeeUnitGen {
 
   /** Generates a FeeUnit based on the maxFee allowed for a transaction */
   def feeUnit(maxFee: Long): Gen[FeeUnit] = {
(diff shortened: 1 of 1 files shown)


Christewart (approved):

Christewart:
So this LGTM, but one question: 

Do you think this is the root cause of the spurious CI failures we've had for broadcasting transactions? If so, we should probably reference those issues (if we have them, i can't seem to find them? I could have swore we had some though) so we can put an end to that saga for bugs.

benthecarman:
> So this LGTM, but one question:
> 
> Do you think this is the root cause of the spurious CI failures we've had for broadcasting transactions? If so, we should probably reference those issues (if we have them, i can't seem to find them? I could have swore we had some though) so we can put an end to that saga for bugs.

Idk if this was the cause of all of them, it only had a 0.33% chance of occurring

benthecarman:
Fixed another 0 I missed

Christewart:
Different CI failures now:

https://github.com/bitcoin-s/bitcoin-s/runs/1538230679#step:5:1942

PR #2370, opened 2020-12-16
Add get transaction cli command

benthecarman:


Christewart (commented):
3 things

1. In the case of giving a txid that is not included in the wallet, we may want to give a better error than this 

> ./bitcoin-s-cli gettransaction 89656a6ee8c35d4cdf11681e8f85a231e8c52b192f0497d1b04641c3b448d2e6
Got unexpected response: {"result":null,"error":null}

2.  If you look at the [bitcoin developer reference for their gettransaction](https://developer.bitcoin.org/reference/rpc/gettransaction.html#gettransaction) it returns a deserialized transaction. If you take a look at [`getrawtransaction`](https://developer.bitcoin.org/reference/rpc/getrawtransaction.html#getrawtransaction). One feature I really like about this RPC is that it allows you to specify `1` or `0` after the `txid` to indicate if you want the raw tx or the deseralized version. 


It seems this implementation is a blend of the two rpc calls `gettransaction` and `getrawtransaction` in bitcoin core. it requires that the transaction is in your wallet (this is a fundamental limitation with being a lite client, so we can't do non-wallet txs AFAIK), but returns the raw hex rather than a deserialized version.

In my opinion we should probably return a deserialized data structure, that has a `hex` field in it that contains the full hex tx similar to `gettransaction` in bitcoin core. 

https://developer.bitcoin.org/reference/rpc/gettransaction.html#result

3. We need to document the RPC here: https://bitcoin-s.org/docs/next/applications/server#wallet

benthecarman:
> 1. In the case of giving a txid that is not included in the wallet, we may want to give a better error than this
> 3. We need to document the RPC here: https://bitcoin-s.org/docs/next/applications/server#wallet

Fixed


> 2. .. In my opinion we should probably return a deserialized data structure, that has a hex field in it that contains the full hex tx similar to gettransaction in bitcoin core.

I would like to be able to return everything that core does but we don't have access to all the information. This is what your txDb stores https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/api/wallet/db/TransactionDb.scala

Christewart:
> > 1. In the case of giving a txid that is not included in the wallet, we may want to give a better error than this
> > 2. We need to document the RPC here: https://bitcoin-s.org/docs/next/applications/server#wallet
> 
> Fixed
> 
> > 1. .. In my opinion we should probably return a deserialized data structure, that has a hex field in it that contains the full hex tx similar to gettransaction in bitcoin core.
> 
> I would like to be able to return everything that core does but we don't have access to all the information. This is what your txDb stores https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/api/wallet/db/TransactionDb.scala

I agree that it won't be in `TransactionDb`, but don't we have all of this information in our wallet? 

>{
  "amount" : x.xxx,        (numeric) The transaction amount in BTC
  "fee": x.xxx,            (numeric) The amount of the fee in BTC. This is negative and only available for the
                              'send' category of transactions.
  "confirmations" : n,     (numeric) The number of confirmations
  "blockhash" : "hash",  (string) The block hash
  "blockindex" : xx,       (numeric) The index of the transaction in the block that includes it
  "blocktime" : ttt,       (numeric) The time in seconds since epoch (1 Jan 1970 GMT)
  "txid" : "transactionid",   (string) The transaction id.
  "time" : ttt,            (numeric) The transaction time in seconds since epoch (1 Jan 1970 GMT)
  "timereceived" : ttt,    (numeric) The time received in seconds since epoch (1 Jan 1970 GMT)
  "bip125-replaceable": "yes|no|unknown",  (string) Whether this transaction could be replaced due to BIP125 (replace-by-fee);
                                                   may be unknown for unconfirmed transactions not in the mempool

https://developer.bitcoin.org/reference/rpc/gettransaction.html#result

Christewart (approved):

Issue #1420, opened 2020-05-13
Add documentation of how to import a serialized root xpriv into key manager


Christewart:
If I have a seed that is format of `xpriv...` how do i import this properly into the key-manager?

Christewart:
What should we do with the old xpriv in this case??

1. Delete the old xpriv (bad!)
2. Cache the old xpriv in another file (???)
3. Support multiple concurrent xprvis (??? probably related to #1115)

benthecarman:
This is not currently doable with our current key manager design. Currently, it saves the encrypted seed words, however, with an xprv you can't get it's corresponding mnemonic words. 

To be able to do this we would have to add another way to store the seed and serialize the xprv in the seed file instead. This is doable, however, we would then have 4 different ways to store seeds (encrypted mnemonic, unencrypted mnenomic, encrypted xprv, unencrypted xprv) which can get messy and a lot to maintain.

PR #2372, opened 2020-12-16
Add ability to store ExtPrivateKey instead of Mnemonic

benthecarman:
This is needed for #1420

This allows to read and write `ExtPrivateKey` as a seed.

To be able to use this I needed to change the `BIP39KeyManager` instead hold the root `ExtPrivateKey` instead of storing the mnemonic. This shouldn't change any functionality as we can are able to calculate the root key from the mnemonic

### DIFF
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
@@ -27,25 +27,12 @@ sealed trait DecryptedSeedState extends SeedState {
 
 case class DecryptedMnemonic(mnemonicCode: MnemonicCode, creationTime: Instant)
     extends DecryptedSeedState {
-  override protected def strToEncrypt: String = mnemonicCode.words.mkString(" ")
+  override protected val strToEncrypt: String = mnemonicCode.words.mkString(" ")
 }
 
 case class DecryptedExtPrivKey(xprv: ExtPrivateKey, creationTime: Instant)
     extends DecryptedSeedState {
-
-  override def encrypt(aesPassword: AesPassword): EncryptedSeed = {
-    ByteVector.encodeUtf8(xprv.toStringSensitive) match {
-      case Left(err) => throw err
-      case Right(str) =>
-        val (key, salt) = aesPassword.toKey
-
-        val encrypted = AesCrypt.encrypt(str, key)
-
-        EncryptedSeed(encrypted, salt, creationTime)
-    }
-  }
-
-  override protected def strToEncrypt: String = xprv.toStringSensitive
+  override protected val strToEncrypt: String = xprv.toStringSensitive
 }
 
 case class EncryptedSeed(
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
@@ -1,32 +1,60 @@
 package org.bitcoins.keymanager
 
-import java.time.Instant
-
 import org.bitcoins.core.compat.CompatEither
 import org.bitcoins.core.crypto._
 import org.bitcoins.crypto.{AesCrypt, AesEncryptedData, AesPassword, AesSalt}
 import scodec.bits.ByteVector
 
+import java.time.Instant
 import scala.util.{Failure, Success, Try}
 
-sealed trait MnemonicState {
+sealed trait SeedState {
   def creationTime: Instant
 }
 
-case class DecryptedMnemonic(mnemonicCode: MnemonicCode, creationTime: Instant)
-    extends MnemonicState {
+sealed trait DecryptedSeedState extends SeedState {
+  protected def strToEncrypt: String
 
-  def encrypt(password: AesPassword): EncryptedMnemonic =
-    EncryptedMnemonicHelper.encrypt(this, password)
+  def encrypt(password: AesPassword): EncryptedSeed = {
+    val Right(clearText) = ByteVector.encodeUtf8(strToEncrypt)
+    val (key, salt) = password.toKey
+
+    val encrypted = AesCrypt.encrypt(clearText, key)
+
+    EncryptedSeed(encrypted, salt, creationTime)
+  }
 }
 
-case class EncryptedMnemonic(
+case class DecryptedMnemonic(mnemonicCode: MnemonicCode, creationTime: Instant)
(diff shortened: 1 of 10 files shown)


key-manager/src/main/scala/org/bitcoins/keymanager/bip39/BIP39KeyManager.scala
```diff
@@ -22,42 +19,35 @@ import org.bitcoins.crypto.{AesPassword, Sign}
...
   * BIP39 key manager
   *
-  * @param mnemonic the mnemonic seed used for this wallet
+  * @param rootExtPrivKey the root seed used for this wallet
   * @param kmParams the parameters used to generate the right keychain
   * @see https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
   */
 case class BIP39KeyManager(
-    private val mnemonic: MnemonicCode,
+    private val rootExtPrivKey: ExtPrivateKey,
```
Christewart: so this was a bug on my part, but we should probably do this
```scala
private[this] val rootExtPrivKey: ExtPrivateKey,
```

rather than just `private`. The reason for doing this can be found lower in this file: 

> rootExtPrivKey == bip39Km.rootExtPrivKey

the default behavior in scala `private` is that all classes of the same type can access private members of other classes. We are taking advantage of this in the `equals()` method. 

We probably don't want this behavior. If we have two instances of `BIP39KeyManager` I don't think they should be able to read each others `xprivs`. 

In the `equals()` method below, we should probably just compare xpubs 

https://github.com/bitcoin-s/bitcoin-s/blob/66ec89bcab0bdc1d5aed47e20049737af476861b/core/src/main/scala/org/bitcoins/core/api/keymanager/KeyManagerApi.scala#L15

key-manager/src/main/scala/org/bitcoins/keymanager/bip39/BIP39KeyManager.scala
```diff
@@ -76,24 +66,24 @@ case class BIP39KeyManager(
...
-      s"Masked(bip39password)")},
-       |    creationTime=$creationTime)""".stripMargin
-  }
 }
 
 object BIP39KeyManager
     extends BIP39KeyManagerCreateApi[BIP39KeyManager]
     with BitcoinSLogger {
+
+  def apply(
```
Christewart: nit: I would prefer if we call this `fromMnemonic()`

key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
```diff
@@ -1,32 +1,60 @@
...
+    val Right(clearText) = ByteVector.encodeUtf8(strToEncrypt)
+    val (key, salt) = password.toKey
+
+    val encrypted = AesCrypt.encrypt(clearText, key)
+
+    EncryptedSeed(encrypted, salt, creationTime)
+  }
+}
+
 case class DecryptedMnemonic(mnemonicCode: MnemonicCode, creationTime: Instant)
```
Christewart: Should consider 


`private[this] val mnemonicCode: MnemonicCode, ...)`

Same with other instances of `xprv` in this file.

key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
```diff
@@ -1,32 +1,60 @@
...
 import org.bitcoins.core.compat.CompatEither
 import org.bitcoins.core.crypto._
 import org.bitcoins.crypto.{AesCrypt, AesEncryptedData, AesPassword, AesSalt}
 import scodec.bits.ByteVector
 
+import java.time.Instant
 import scala.util.{Failure, Success, Try}
 
-sealed trait MnemonicState {
+sealed trait SeedState {
```
Christewart, line 11: Do we want to think about using `extends MaskedToString` ? There can be a lot of sensitive information in these `SeedState`.

Christewart (changes requested):
Overall it looks good, i do feel somewhat strongly about trying to make things more `private` or contain them to the `keymanager` package at least. Generally we don't use access modifiers for packages in Scala, but it is a nice feature for things that are security sensitive. 

https://alvinalexander.com/scala/how-to-control-scala-method-scope-object-private-package/

key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
benthecarman: These should be fine since the sensitive information already extend `MaskedToString`

key-manager/src/main/scala/org/bitcoins/keymanager/EncryptedMnemonic.scala
benthecarman: We can't do this because then we can't access it for when we go to actually build the key manager

Christewart: Ok, well at minimum we should be able to do this? 

```scala
private[keymanager] val mnemonicCode: MenonicCode
private[keymanager] val xprv: ExtPrivateKey,
```

With this, only things inside of `org.bitcoins.keymanager._` can access these `val`s

Christewart:
So when I fire up v0.5 for the first time, what will happen with my 0.4 seeds with respect to this PR? From my understanding, it will 

1. Read the existing mnemonic seed
2. Boot up the KeyManager
3. Move the existing mnemonic seed into a backup file (should we have a subfolder in the `seeds/` for this?? Like `seeds/movedseeds/` ?? )
4. Write a new encrypted seed file that has the xprv that corresponds to the mnemonic seed in it

Is that correct? We should probably have a simple explanation like this in the release notes.

benthecarman:
> So when I fire up v0.5 for the first time, what will happen with my 0.4 seeds with respect to this PR? From my understanding, it will
> 
>     1. Read the existing mnemonic seed
> 
>     2. Boot up the KeyManager
> 
>     3. Move the existing mnemonic seed into a backup file (should we have a subfolder in the `seeds/` for this?? Like `seeds/movedseeds/` ?? )
> 
>     4. Write a new encrypted seed file that has the xprv that corresponds to the mnemonic seed in it
> 
> 
> Is that correct? We should probably have a simple explanation like this in the release notes.

This PR shouldn't change anything. This isn't changing all seeds to be Xprv's it is only adding the option for it to be so. The only time we store the xprv is if you import an xprv

Christewart (approved):

PR #2379, opened 2020-12-17
Rescan Improvements

benthecarman:
Previously, we would scan all filters and then afterwards download and process blocks. This would be bad because if we stopped the rescan in the middle it would have 0 effect, as well as could cause a rescan to fail because we do all the major network and computation functions all at once at the end. This makes it so we download a block once we find we need it.

This also fixes rescans so we use what is in our spk table

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala
@@ -159,13 +159,13 @@ class WalletUnitTest extends BitcoinSWalletTest {
 
   it should "match block filters" in { wallet: Wallet =>
     for {
-      matched <- wallet.getMatchingBlocks(
+      height <- wallet.chainQueryApi.getFilterCount()
+      matched <- wallet.fetchFiltersInRange(
         scripts = Vector(
           // this is a random address which is included into the test block
           BitcoinAddress("n1RH2x3b3ah4TGQtgrmNAHfmad9wr8U2QY").scriptPubKey),
-        startOpt = None,
-        endOpt = None
-      )(system.dispatcher)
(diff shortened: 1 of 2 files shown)


Christewart:
> Previously, we would scan all filters and then afterwards download and process blocks. This would be bad because if we stopped the rescan in the middle it would have 0 effect, as well as could cause a rescan to fail because we do all the major network and computation functions all at once at the end. This makes it so we download a block once we find we need it.
> 
> This also fixes rescans so we use what is in our spk table

So one conceptual level question about this implementation vs our previous implementation: 

Did the previous implementation have atomicity with the rescan? What I mean by that is the entire thing succeeded or the entire thing failed.

If I understand your comment above, it seems like we can have partially successful rescans with this implementation that can leave your wallet in a weird state?

benthecarman:
> If I understand your comment above, it seems like we can have partially successful rescans with this implementation that can leave your wallet in a weird state?

Yes, but I still think this is an improvement, because you can then do another rescan from the height you stopped on instead of having to do another full rescan

Christewart (approved):

Christewart:
This unit test runs forever, you can see this behavior on CI and i replicated locally 

https://github.com/benthecarman/bitcoin-s/blob/18755df3d1738a1de968f1e653fbf7b4d23a0582/wallet-test/src/test/scala/org/bitcoins/wallet/WalletUnitTest.scala#L160

Screenshot from 2020-12-17 16-32-34

benthecarman:
Test should be fixed, I had to change it to use `fetchFilters` instead of `matchBlocks`. `matchBlocks` calls `fetchFilters` but now since `matchBlocks` also calls `downloadAndProcessBlocks` it would stall waiting for us to download and process the block.

Christewart (approved):

PR #2380, opened 2020-12-17
QR codes

rorp:
Currently it works only for CLI.


benthecarman (approved):
ACK

project/Deps.scala
```diff
@@ -312,7 +319,9 @@ object Deps {
       //we can remove this dependency when this is fixed
       //https://github.com/oracle/graal/issues/1943
       //see https://github.com/bitcoin-s/bitcoin-s/issues/1100
-      Compile.codehaus
+      Compile.codehaus,
+      Compile.zxingCore,
```
Christewart, line 323: is there a reason these need to be included in the `cli` itself rather than `appServer`? It would be nice to keep the cli as lightweight as possible with dependencies.

Christewart:
Added to the GUI in

PR #2390, opened 2020-12-17
Small DLC GUI touchups 

benthecarman:
Fixes:

- No text appears on the sign and add sigs dialog when you select a file the way it does for accept
- Warning message on confirmation when no announcement is given is not readable as it cuts off and can't be re-sized
- Adds status label at the bottom of the window said something after export result completes
- Display correct payout curve as recipient


nkohen (approved):
ACK on CI (which I think is failing)

PR #2392, opened 2020-12-18
Add wallet function to bump fee with RBF

benthecarman:
Allows you to bump the fee of a transaction that you created

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
@@ -363,7 +363,7 @@ trait WalletApi extends StartStopAsync[WalletApi] {
       amounts: Vector[CurrencyUnit],
       feeRate: FeeUnit)(implicit ec: ExecutionContext): Future[Transaction]
 
-  def bumpFee(
+  def bumpFeeRBF(
       txId: DoubleSha256DigestBE,
       newFeeRate: FeeUnit): Future[Transaction]
 
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
@@ -1,7 +1,5 @@
 package org.bitcoins.core.api.wallet
 
-import java.time.Instant
-
 import org.bitcoins.core.api.chain.ChainQueryApi
 import org.bitcoins.core.api.feeprovider.FeeRateApi
 import org.bitcoins.core.api.keymanager.KeyManagerApi
@@ -22,6 +20,7 @@ import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.core.wallet.utxo.{AddressTag, AddressTagType, TxoState}
 import org.bitcoins.crypto.DoubleSha256DigestBE
 
+import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
@@ -363,6 +362,10 @@ trait WalletApi extends StartStopAsync[WalletApi] {
       addresses: Vector[BitcoinAddress],
       amounts: Vector[CurrencyUnit],
       feeRate: FeeUnit)(implicit ec: ExecutionContext): Future[Transaction]
+
+  def bumpFee(
+      txId: DoubleSha256DigestBE,
+      newFeeRate: FeeUnit): Future[Transaction]
 
   def makeOpReturnCommitment(
       message: String,
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -1,7 +1,5 @@
 package org.bitcoins.wallet
 
-import java.time.Instant
-
 import org.bitcoins.commons.jsonmodels.wallet.SyncHeightDescriptor
 import org.bitcoins.core.api.chain.ChainQueryApi
 import org.bitcoins.core.api.feeprovider.FeeRateApi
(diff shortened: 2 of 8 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -499,6 +494,86 @@ abstract class Wallet
...
+          Future.failed(
+            new RuntimeException(s"Unable to find transaction ${txId.hex}"))
+      }
+
+      outPoints = tx.inputs.map(_.previousOutput).toVector
+      spks = tx.outputs.map(_.scriptPubKey).toVector
+
+      utxos <- spendingInfoDAO.findByOutPoints(outPoints)
+      _ = require(utxos.size == tx.inputs.size,
+                  "Can only bump fee for our own transaction")
```
Christewart: nit: Might be nice to have two different messages for this: 

1. If the transaction has _zero_ inputs relevant to us, this error message seems fine
2. If the transaction has one or more inputs that are ours, but not _all_ of them are ours, it might be nice to say this too with a customized error message.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -499,6 +494,86 @@ abstract class Wallet
...
+          case _: SatoshisPerKiloByte =>
+            SatoshisPerKiloByte.calc(inputAmount, tx)
+          case _: SatoshisPerVirtualByte =>
+            SatoshisPerVirtualByte.calc(inputAmount, tx)
+          case _: SatoshisPerKW =>
+            SatoshisPerKW.calc(inputAmount, tx)
+        }
+
+        require(oldFeeRate.currencyUnit < newFeeRate.currencyUnit,
+                "Cannot bump to a lower fee")
```
Christewart: nit: add the fees to the error message

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -499,6 +494,86 @@ abstract class Wallet
...
+          Random.shuffle(changeSpks).head
+        } else {
+          // If none are explicit change, pick a random one we own
+          Random.shuffle(myAddrs.map(_.scriptPubKey)).head
+        }
+
+      // Mark old outputs as replaced
+      oldUtxos <- spendingInfoDAO.findDbsForTx(txId)
+      _ <- spendingInfoDAO.updateAll(
+        oldUtxos.map(_.copyWithState(TxoState.DoesNotExist)))
```
Christewart, line 563: Why is this going to `TxoState.DoesNotExist` ? Shouldn't this be `PendingConfirmationsReceived`?

core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
```diff
@@ -364,6 +363,10 @@ trait WalletApi extends StartStopAsync[WalletApi] {
       amounts: Vector[CurrencyUnit],
       feeRate: FeeUnit)(implicit ec: ExecutionContext): Future[Transaction]
 
+  def bumpFee(
```
Christewart: Scaladoc on what method this scala method is using to bump the fee? Is it using CPFP or RBF? I would be supportive of changing the method name to 

`bumpFeeCFPF()` or `bumpFeeRBF()`. 

Or we can make it even more complex by adding an extra parameter to the method that with an ADT like this

```scala

sealed trait FeeBumpMethod
object FeeBumpMethod { 
  final case object CPFP extends FeeBumpMethod
  final case object RBF extends FeeBumpMethod
}
```

and then just add that as a parameter. This might be nice if there are future fee bump methods -- but it is more work.

wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
```diff
@@ -280,6 +280,21 @@ case class AddressDAO()(implicit
       })
   }
 
+  def findByScriptPubKeys(
```
Christewart, line 283: This method seems important, let's add a sanity unit test for this?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -277,6 +277,15 @@ case class SpendingInfoDAO()(implicit
 
   }
 
+  /**
+    * Fetches all the incoming TXOs in our DB that are in
+    * the transaction with the given TXID
+    */
+  def findDbsForTx(txid: DoubleSha256DigestBE): Future[Vector[UTXORecord]] = {
```
Christewart, line 284: Same with this method, please add a unit test so we can detect regressions in the future. I think it would be good to also have a negative test case making sure we don't accidentally fetch outgoing txos (the scaladoc indicates this is just for incomign txos).

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: If it is `PendingConfirmationsReceived` then this transaction will double count to our balance.

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
benthecarman: Added test.

> I think it would be good to also have a negative test case making sure we don't accidentally fetch outgoing txos (the scaladoc indicates this is just for incomign txos).

All txos in the spendinginfodao is an incoming txo

core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
benthecarman: Changed the name of the function to `bumpFeeRBF`. In the future we can make `bumpFeeCPFP`

Christewart (approved):

PR #2394, opened 2020-12-18
Remove CompatEither, it was needed for historical purposes to support…

Christewart:
… Scala 2.11.x

We dropped support for Scala 2.11 in #1479 , so we no longer need `CompatEither`. This was originally needed because Scala changed semantics between 2.11.x, 2.12.x & 2.13.x . 

`CompatEither` was primarily used in the key manager module, so let's get rid of it before we release 0.5


benthecarman (approved):
Was thinking of doing this the other day lol

PR #2398, opened 2020-12-18
Windows Secp Update & fix for parsing Windows paths

benthecarman:
Needed to do some preprocessing for when we put a directory in a config because windows uses backslashes and colons.


benthecarman:
Looks like need to fix paths for windows

Edit: Fixed

.github/workflows/Windows.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.ivy2/cache
+            ~/.sbt
+            ~/.bitcoin-s/binaries
+          key: ${{ runner.os }}-cache
+      - name: Windows Crypto, Core, and Database tests
+        run: sbt ++2.13.4 cryptoTest/test coreTest/test dbCommonsTest/test
```
Christewart, line 28: it probably wouldn't be a bad idea to run the `bitcoindRpcTest` and `eclairRpcTest` stuff too? There some file manipulation in those projects too.

benthecarman: Added in a separate workflow. Originally put it in same but the single CI run took over 30 mins

benthecarman: Looks like even on it's own it's extremely slow

Christewart:
Yikes, if I understand this correctly we are going to have to rework how we start bitcoind? 


https://github.com/bitcoin-s/bitcoin-s/pull/2398/checks?check_run_id=1583008381#step:6:1198

Christewart:
Maybe we should just split this work out into a separate PR and leave the scope of this PR to the secp update. I was hoping things would just work :tm:

benthecarman:
> Maybe we should just split this work out into a separate PR and leave the scope of this PR to the secp update. I was hoping things would just work tm

Yeah, oh well

Christewart:
Looks like we have an unused import causing compile to fail 

https://github.com/bitcoin-s/bitcoin-s/pull/2398/checks?check_run_id=1585062139#step:5:423

Christewart (approved):

PR #2399, opened 2020-12-19
Add wallet function to bump fee with CPFP

benthecarman:
Will be useful for implementing stuff in https://github.com/discreetlogcontracts/dlcspecs/blob/master/Non-Interactive-Protocol.md

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -704,6 +704,43 @@ abstract class Wallet
     } yield tx
   }
 
+  /** @inheritdoc */
+  override def bumpFeeCPFP(
+      txId: DoubleSha256DigestBE,
+      feeRate: FeeUnit): Future[Transaction] = {
+    for {
+      txDbOpt <- transactionDAO.findByTxId(txId)
+      tx <- txDbOpt match {
+        case Some(db) => Future.successful(db.transaction)
+        case None =>
+          Future.failed(
+            new RuntimeException(s"Unable to find transaction ${txId.hex}"))
+      }
+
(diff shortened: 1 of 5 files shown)


Christewart:
needs rebase

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -627,6 +627,43 @@ abstract class Wallet
...
+          Future.failed(
+            new RuntimeException(s"Unable to find transaction ${txId.hex}"))
+      }
+
+      spendingInfos <- spendingInfoDAO.findTx(tx)
+      _ = require(spendingInfos.nonEmpty,
+                  s"Transaction ${txId.hex} must have an output we own")
+
+      changeSpendingInfos = spendingInfos.flatMap { db =>
+        if (db.privKeyPath.chain.chainType == HDChainType.Change) {
```
Christewart, line 725: Interesting, do we have no other way to detect if something is a change output ? 

In the future it would probably be nice to decouple detecting change and BIP32. Definitely shouldn't be done in this PR though.

Christewart (commented):
It would be nice to have an integration level test with this against bitcoind? 

I think we could use the bitcoind rpc to determine what bitcoin sees at the fee rate?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
benthecarman: ACK

benthecarman:
Added integration tests but couldn't assert fees from bitcoind. Bitcoind only gives you access to the fees if the transaction is created by its wallet.

Christewart (approved):

PR #2405, opened 2020-12-21
Bump Eclair version

rorp:


benthecarman:
Needs to be rebased

Christewart:
CI failing due to unused import, as ben states i think this will get fixed if you rebase

https://github.com/bitcoin-s/bitcoin-s/pull/2405/checks?check_run_id=1586250740#step:5:622

eclair-rpc-test/src/test/scala/org/bitcoins/eclair/rpc/EclairRpcClientTest.scala
```diff
@@ -65,38 +65,36 @@ class EclairRpcClientTest extends BitcoinSAsyncTest {
...
 
-  lazy val fourthClientF = eclairNodesF.map(_.c4)
+  lazy val fourthClientF: Future[EclairRpcClient] = eclairNodesF.map(_.c4)
 
   /** There is specific cases where we just need two clients,
     * so this is a helper val that pairs two connected
     * clients together with an open channel
     */
-  lazy val clientOtherClientF = {
+//  lazy val clientOtherClientF = {
```
Christewart: is this not needed anymore?

core/src/main/scala/org/bitcoins/core/protocol/ln/routing/Route.scala
```diff
@@ -0,0 +1,13 @@
+package org.bitcoins.core.protocol.ln.routing
+
+import org.bitcoins.core.protocol.ln.channel.ShortChannelId
+import org.bitcoins.core.protocol.ln.node.NodeId
+
+/**
+  * Represent differet types of LN routes. Supports node and channel routes.
+  */
+sealed trait Route
```
Christewart, line 9: Maybe consider this? I think we can get rid of some of the `.asInstanceOf[]` then 

```scala
sealed trait Route[T] {
  def ids: Vector[T]
}

case class NodeRoute(ids: Vector[NodeId]) extends Route[NodeId]

case class ChannelRoute(ids: Vector[ShortChannelId])
    extends Route[ShortChannelId]
```

eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/client/EclairRpcClient.scala
```diff
@@ -904,11 +908,16 @@ class EclairRpcClient(
     val incoming: Sink[Message, Future[Done]] =
       Sink.foreach[Message] {
         case message: TextMessage.Strict =>
-          val parsed: JsValue = Json.parse(message.text)
-          val validated: JsResult[WebSocketEvent] =
-            parsed.validate[WebSocketEvent]
-          val event = parseResult[WebSocketEvent](validated, parsed, "ws")
-          eventHandler(event)
+          try {
```
Christewart, line 912: Hmm.. why did this need to be added? Is there events we don't support?

Christewart (approved):

PR #2413, opened 2020-12-21
Fix server configuration for the app server

### LINKED ISSUE #2410: Server configuration broken
We support changing the `rpcport` of our server. I've tried modifying this inside my `bitcoin-s.conf` file when testing #2409. It did not work.

I then checked out the last master commit  89222eb7669acf5717cf0c125faaa32299185e83  and tried to only modify the `rpcport` setting and it did not work. 

You can see this by the following log messages

>2020-12-21T19:19:00UTC INFO [BitcoinSServerMain] Starting neutrino node sync
2020-12-21T19:19:00UTC INFO [Server] Started Bitcoin-S HTTP server at /127.0.0.1:9999
2020-12-21T19:19:16UTC INFO [NeutrinoNode] Starting sync node, height=1899316 hash=0000000000000003920fb40891d0dadeb8f4b0ab06911829da5151495df8c145

Christewart:
fixes #2410 

This was introduced in #2121 -- i forgot to update the server configuration as it isn't apart of the `AppConfig` hierarchy.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
@@ -67,13 +67,9 @@ case class BitcoinSAppConfig(
     resolved
   }
 
-  def serverConf: Config = {
-    config.getConfig("bitcoin-s.server")
-  }
-
   def rpcPortOpt: Option[Int] = {
-    if (serverConf.hasPath("bitcoin-s.server.rpcport")) {
-      Some(serverConf.getInt("bitcoin-s.server.rpcport"))
+    if (config.hasPath("bitcoin-s.server.rpcport")) {
+      Some(config.getInt("bitcoin-s.server.rpcport"))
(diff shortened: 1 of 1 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
```diff
@@ -67,13 +67,9 @@ case class BitcoinSAppConfig(
...
-  }
-
   def rpcPortOpt: Option[Int] = {
-    if (serverConf.hasPath("bitcoin-s.server.rpcport")) {
-      Some(serverConf.getInt("bitcoin-s.server.rpcport"))
+    if (config.hasPath("bitcoin-s.server.rpcport")) {
+      Some(config.getInt("bitcoin-s.server.rpcport"))
     } else {
       None
     }
```
benthecarman, line 75: nit: can use `config.getIntOpt` instead

benthecarman (approved):

PR #2126, opened 2020-10-05
Update neutrino binary with latest bitcoin core code

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/1629

PR #2414, opened 2020-12-22
Bitcoind v0.21.0 support

### LINKED ISSUE #861: Update BIP 157 core branch to latest version
https://github.com/bitcoin/bitcoin/pull/16442 was rebased today and the update should be reflected here.

### LINKED ISSUE #1629: Update neutrino binary with latest bitcoin core code
Our neutrino binary is out of date, we should update to the latest functionality and test it

https://github.com/bitcoin/bitcoin/pull/18876

benthecarman:
Closes #2126 
Closes #1629
Closes #861

https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.21.0.md

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v21/BitcoindV21RpcClientTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v21/BitcoindV21RpcClientTest.scala
@@ -1,0 +1,222 @@
+package org.bitcoins.rpc.v21
+
+import org.bitcoins.commons.jsonmodels.bitcoind.RpcOpts.WalletFlag
+import org.bitcoins.commons.jsonmodels.bitcoind._
+import org.bitcoins.core.config.RegTest
+import org.bitcoins.core.gcs.{BlockFilter, FilterType}
+import org.bitcoins.core.psbt.PSBT
+import org.bitcoins.crypto.ECPublicKey
+import org.bitcoins.rpc.client.common.BitcoindVersion
+import org.bitcoins.rpc.client.v21.BitcoindV21RpcClient
+import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
+import org.bitcoins.testkit.util.BitcoindRpcTest
+
+import java.io.File
+import java.nio.file.Files
+import scala.concurrent.Future
+
+class BitcoindV21RpcClientTest extends BitcoindRpcTest {
+
+  lazy val clientPairF: Future[(BitcoindV21RpcClient, BitcoindV21RpcClient)] =
+    BitcoindRpcTestUtil.createNodePairV21(clientAccum)
+
+  lazy val clientF: Future[BitcoindV21RpcClient] = clientPairF.map(_._1)
+
+  clientF.foreach(c => clientAccum.+=(c))
+
+  behavior of "BitcoindV21RpcClient"
+
+  it should "be able to start a V21 bitcoind instance" in {
+    clientF.map { client =>
+      assert(client.version == BitcoindVersion.V21)
+    }
+  }
+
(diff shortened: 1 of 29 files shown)


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v21/BitcoindV21RpcClientTest.scala
```diff
@@ -0,0 +1,222 @@
...
+import scala.concurrent.Future
+
+class BitcoindV21RpcClientTest extends BitcoindRpcTest {
+
+  lazy val clientPairF: Future[(BitcoindV21RpcClient, BitcoindV21RpcClient)] =
+    BitcoindRpcTestUtil.createNodePairV21(clientAccum)
+
+  lazy val clientF: Future[BitcoindV21RpcClient] = clientPairF.map(_._1)
+
+  clientF.foreach(c => clientAccum.+=(c))
```
Christewart, line 25: don't you want to add the `clientPairF` to the accumulator? Seems you are only adding one client here

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -103,6 +105,7 @@ trait BitcoindRpcTestUtil extends BitcoinSLogger {
       if (blockFilterIndex)
         conf + """
                  |blockfilterindex=1
+                 |peerblockfilters=1
```
Christewart, line 108: what does this setting do?

Christewart (changes requested):
I vote we don't merge until the official v21 is released

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
benthecarman: This makes it so we can send block filters over the network

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v21/BitcoindV21RpcClientTest.scala
benthecarman: `createNodePairV21` adds  them to it

bitcoind-rpc/bitcoind-rpc.sbt
```diff
@@ -51,7 +51,10 @@ TaskKeys.downloadBitcoind := {
     val location =
       if (version == experimentalVersion)
         s"https://s3-us-west-1.amazonaws.com/suredbits.com/bitcoin-core-$version/bitcoin-$version-$platform.$suffix"
-      else
+      else if (version.init.endsWith("rc")) { // if it is a release candidate
```
Christewart, line 54: i think this can be removed now?

Christewart (approved):
This LGTM overall, i think there is just one thing that needs to be removed in the build, and then we should also remove the `@DoNotDiscover` annotation from this test case

https://github.com/bitcoin-s/bitcoin-s/blob/da54d2e9fbd06f5a9e569bfb89321cc4be97d5ba/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala#L22

bitcoind-rpc/bitcoind-rpc.sbt
benthecarman: imo we should leave this, might want to use a release canidate in the future and will just have to put this back

Christewart:
Looks like we might be having some problems starting the node in some cases? 

https://github.com/bitcoin-s/bitcoin-s/pull/2414/checks?check_run_id=1703821908#step:5:988

PR #2415, opened 2020-12-22
Bump Fee Cli commands

benthecarman:


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -718,6 +718,56 @@ object ConsoleCli {
...
+            .text("Id of transaction to bump fee")
+            .required()
+            .action((txid, conf) =>
+              conf.copy(command = conf.command match {
+                case cpfp: BumpFeeCPFP =>
+                  cpfp.copy(txId = txid)
+                case other => other
+              })),
+          arg[SatoshisPerVirtualByte]("feerate")
+            .text("Fee rate in sats per virtual byte")
```
Christewart: What transaction is this fee rate applied to? Is this the

1. Fee rate for the child transaction
2. The target fee rate for the package (`(parent + child tx fee rates) / 2`)

Christewart (approved):

PR #2416, opened 2020-12-22
Add extra checks for RBF transactions

benthecarman:
Added requirements that the original tx can't be confirmed or was not originally RBF.

Added tests to make sure this is functioning correctly.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -506,6 +506,10 @@ abstract class Wallet
             new RuntimeException(s"Unable to find transaction ${txId.hex}"))
       }
 
+      _ = require(
+        tx.inputs.exists(_.sequence < TransactionConstants.disableRBFSequence),
+        "Transaction is not signaling RBF")
+
       outPoints = tx.inputs.map(_.previousOutput).toVector
       spks = tx.outputs.map(_.scriptPubKey).toVector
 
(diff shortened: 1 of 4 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -506,13 +506,22 @@ abstract class Wallet
             new RuntimeException(s"Unable to find transaction ${txId.hex}"))
       }
 
+      _ = require(
+        tx.inputs.exists(_.sequence < TransactionConstants.disableRBFSequence),
```
Christewart: nit: This might be a useful utility elsewhere. Might be worth moving to `TxUtil.isRBFEnabled()`?

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -506,13 +506,22 @@ abstract class Wallet
...
       outPoints = tx.inputs.map(_.previousOutput).toVector
       spks = tx.outputs.map(_.scriptPubKey).toVector
 
       utxos <- spendingInfoDAO.findByOutPoints(outPoints)
       _ = require(utxos.nonEmpty, "Can only bump fee for our own transaction")
       _ = require(utxos.size == tx.inputs.size,
                   "Can only bump fee for a transaction we own all the inputs")
+
+      oldOutputs <- spendingInfoDAO.findDbsForTx(txId)
+      _ = require(!oldOutputs.exists(_.blockHash.isDefined),
```
Christewart: nit: Put the blockhash it wasn't confirmed in in the error message, that would be useful information to have for debugging purposes.

Christewart: Also don't we need this check for CPFP txs too? If they are confirmed in a block I don't think we want to CFPF do we? 

My guess is we we "fail silently" as the CPFP will just spend the utxo that is in a block? Failing has a weird semantic meaning here, as the child transaction will get included in the mempool, and in a block eventually, but it didn't achieve it's intended purpose and really just wasted fees i think?

benthecarman: Good idea, fixed

Christewart (approved):

PR #2426, opened 2020-12-23
KeyManager Docs

### LINKED ISSUE #1419: Add how to import a mnemonic seed into the key manager
Needs to be added to the website

### LINKED ISSUE #1420: Add documentation of how to import a serialized root xpriv into key manager

### LINKED ISSUE #2367: Add documentation for restoring a wallet from seed(s)
We should add explicit documentation for how to restore a wallet from a set of seeds that were backed up from the `seeds/` directory.

benthecarman:
Closes #2367 
Closes #1419
Closes #1420

### DIFF
--- a/docs/key-manager/server-key-manager.md
+++ b/docs/key-manager/server-key-manager.md
@@ -19,6 +19,16 @@ You can change which wallet and key manager you are using by setting the `bitcoi
 option. If it is not set, it will use the default wallet. Different seeds are saved in the `seeds` folder, and the
 seed's file name will be prefixed with the wallet name.
 
+For example, if you have a default wallet, and 2 wallets named `test-wallet` and `rickRollWallet` your seeds folder will look like:
+
+```
+tree ~/.bitcoin-s/seeds
+/home/suredbits/.bitcoin-s/seeds
+├── encrypted-bitcoin-s-seed.json
+├── test-wallet-encrypted-bitcoin-s-seed.json
+└── rickRollWallet-encrypted-bitcoin-s-seed.json
+```
+
 Currently, Bitcoin-S only supports using a single key manger/wallet at a single instance. When switching wallets, you
 will need to shut down the server and restart it with your modified config.
 
@@ -72,4 +82,7 @@ To do this, you will need the seed file name to be `walletName-encrypted-bitcoin
 if you want it to be the default wallet.
 To load it, just set the `bitcoin-s.wallet.walletName` config option to the appropriate wallet name.
 
-You made need to run a wallet rescan to restore funds, this can be done by using the `rescan` RPC command.
+You may need to run a wallet rescan to restore funds, this can be done by using the `rescan` RPC command.
+
+For example, if you wanted to restore your `rickRollWallet`, you would place your backup at `~/.bitcoin-s/seeds/rickRollWallet-encrypted-bitcoin-s-seed.json`
+and then load the wallet using a config with `bitcoin-s.wallet.walletName = rickRollWallet`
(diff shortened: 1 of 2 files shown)


docs/key-manager/server-key-manager.md
```diff
@@ -0,0 +1,75 @@
...
+
+All seeds are stored in the `seeds` folder inside your bitcoin-s directory, the default location is in `~/.bitcoin-s`.
+The easiest way to back up your seeds is to simply copy the `seeds` folder to your secure back up location.
+
+Do note that if you have a key manager passphrase or a BIP 39 passphrase set for a given seed, you will need these to be
+able to recover the seed.
+
+### Using multiple seeds
+
+You can change which wallet and key manager you are using by setting the `bitcoin-s.wallet.walletName` config
```
Christewart, line 18: maybe make it more concrete by adding an example? For instance here is my seeds directory 

> tree ~/.bitcoin-s/seeds
/home/suredbits/.bitcoin-s/seeds
├── encrypted-bitcoin-s-seed.json
├── test-wallet-encrypted-bitcoin-s-seed.json
└── testwallet-encrypted-bitcoin-s-seed.json

0 directories, 3 files

docs/key-manager/server-key-manager.md
```diff
@@ -0,0 +1,75 @@
...
+### Restoring a Seed
+
+#### Restore from Mnemonic
+
+To restore from a mnemonic seed you can use the `importseed` cli command.
+It takes in the actual mnemonic seed, wallet name, and an optional passphrase to encrypt the seed.
+
+**Example:**
+
+```
```
Christewart, line 63: or if you want to keep this example going, use the rickRollWallet rather than my `tree` above so we are consistent with the docs

docs/key-manager/server-key-manager.md
```diff
@@ -0,0 +1,75 @@
...
+bitcoin-s-cli importxprv "xprv9uHRZZhk6KAJC1avXpDAp4MDc3sQKNxDiPvvkX8Br5ngLNv1TxvUxt4cV1rGL5hj6KCesnDYUhd7oWgT11eZG7XnxHrnYeSvkzY7d2bhkJ7" "xprvWallet" "insecurePassword"
+```
+
+#### Restore from backup
+
+To restore from a backed up seed, simply put the seed file in the `seeds` folder and then load the wallet.
+To do this, you will need the seed file name to be `walletName-encrypted-bitcoin-s-seed.json` or `encrypted-bitcoin-s-seed.json`
```
Christewart, line 81: now you can use the example of restoring `rickRollWallet` here as the concrete example

docs/key-manager/server-key-manager.md
```diff
@@ -0,0 +1,75 @@
...
+if you want it to be the default wallet.
+To load it, just set the `bitcoin-s.wallet.walletName` config option to the appropriate wallet name.
+
+You made need to run a wallet rescan to restore funds, this can be done by using the `rescan` RPC command.
```
Christewart: So what's your opinion on using scala code examples vs cli examples?

Ideally we would have both, but it seems you favor the cli. I like scala code examples personally, and we get the added benefit of compile time check on the docs so we have to make sure we keep them updated. 

These are pretty important commands, so it is worth duplicating in my opinion. You have a pretty good narrative going on in the docs with the example too.

Christewart: Oh, i think I understand what you mean by "Server KeyManager" now -- this is using the key manager as part of the app server binary? 

So in this case, i think we should add the scala code examples above to the normal key manager doc then?

Christewart (changes requested):

Christewart (approved):

PR #2429, opened 2020-12-23
Add doc about no mempool

### LINKED ISSUE #2366: Add documentation saying that bitcoin-s wallet and node only has 'blocks-only' mode currently
no mempool functionality, only blocks only mode.

benthecarman:
Fixes #2366

### DIFF
--- a/docs/node/node.md
+++ b/docs/node/node.md
@@ -11,6 +11,11 @@ Bitcoin-s has experimental support for neutrino which is a new lite client propo
 read more about how neutrino works [here](https://suredbits.com/neutrino-what-is-it-and-why-we-need-it/). At this time, 
 bitcoin-s only supports connecting to one trusted peer.
 
+#### Limitations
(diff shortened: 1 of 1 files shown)


docs/node/node.md
```diff
@@ -11,6 +11,11 @@ Bitcoin-s has experimental support for neutrino which is a new lite client propo
 read more about how neutrino works [here](https://suredbits.com/neutrino-what-is-it-and-why-we-need-it/). At this time, 
 bitcoin-s only supports connecting to one trusted peer.
 
+#### Limitations
```
Christewart, line 14: We should probably note this in the wallet docs too? Something like 

"Due to limitations in the node project, the wallet does not support _inbound_ unconfirmed transactions" 

We should be able to display outbound unconfirmed transactions as we have those locally?

Christewart (approved):

PR #2431, opened 2020-12-23
Remove BlockHeader.getBlockchainsFrom(), rework type signature for Bl…

Christewart:
…ockHeader.getBlockchainFrom() to return Future[Option[Blockchain]]

For some reason we had `getBlockchainsFrom(header)` and `getBlockchainFrom(header)`.

Conceptually these things are the same. If given a header, you can only walk backwards from a header. 

The PR also reworks the return type to be `Future[Option[Blockchain]]` rather than `Future[Vector[Blockchain]]`. As stated above, you can only obtain one blockchain when walking backwards from a given header.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
@@ -115,8 +115,7 @@ class BitcoinPowTest extends ChainDbUnitTest {
               Pow.getNetworkWorkRequired(nextTip.blockHeader, chain)
             assert(nextNBits == nextTip.nBits)
           case None =>
-            fail()
-
+            fail(s"Chain not found best on header at height=$height")
         }
       }
     }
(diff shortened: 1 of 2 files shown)


chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
```diff
@@ -104,14 +104,20 @@ class BitcoinPowTest extends ChainDbUnitTest {
...
+        blockchainOptF.map {
+          case Some(blockchain) =>
+            val chain = Blockchain.fromHeaders(blockchain.tail.toVector)
+            val nextTip = blockchain.tip
+            val nextNBits =
+              Pow.getNetworkWorkRequired(nextTip.blockHeader, chain)
+            assert(nextNBits == nextTip.nBits)
+          case None =>
+            fail()
+
```
benthecarman: nit: spacing

chain-test/src/test/scala/org/bitcoins/chain/pow/BitcoinPowTest.scala
```diff
@@ -104,14 +104,20 @@ class BitcoinPowTest extends ChainDbUnitTest {
...
+        blockchainOptF.map {
+          case Some(blockchain) =>
+            val chain = Blockchain.fromHeaders(blockchain.tail.toVector)
+            val nextTip = blockchain.tip
+            val nextNBits =
+              Pow.getNetworkWorkRequired(nextTip.blockHeader, chain)
+            assert(nextNBits == nextTip.nBits)
+          case None =>
+            fail()
```
benthecarman: Should probably give an error message saying header not found

benthecarman (approved):

Issue #2004, opened 2020-09-11
Break dependency of node project on bitcoindRpc


Christewart:
I don't think this is required? 

https://github.com/bitcoin-s/bitcoin-s/blob/79ca91029c3cc1eb35e9e18edee08c8f8c3347fc/build.sbt#L391

Christewart:
I think there is going to need to be a new module made named `async-utils` or something like that. It seems that we dependened on `bitcoindRpc` in the `node` project because we needed access to `AsyncUtil` which current lives in the `bitcoindRpc` project.

benthecarman:
Can probably just move that to commons?

benthecarman:
Looks like it is also because of this

https://github.com/bitcoin-s/bitcoin-s/blob/master/node/src/main/scala/org/bitcoins/node/models/Peer.scala#L29

benthecarman:
Maybe we can move `BitcoindInstance` to commons. We would need to move the whole `config` package however

PR #2440, opened 2020-12-29
Move RPC server logic into separate project

### LINKED ISSUE #2277: Break oracleServer dependsOn(appServer)
This should be done for various reasons

1. Security (don't include things on classpath that aren't being used)
2. build times will be faster

I'm unsure if there is going to need to be refactoring done to support all the existing features the `oracleServer` has. @benthecarman hinted on slack we use a lot of the routes stuff defined in `appServer` for `oracleServer`. We may want to think about tying this in with #2004 (specificially this comment: https://github.com/bitcoin-s/bitcoin-s/issues/2004#issuecomment-691165386)

benthecarman:
Fixes #2277


Christewart (approved):
This looks great! It's always nice when refactors like this are painless. If you are feeling motivated maybe tackle #2004

PR #2441, opened 2020-12-29
Fix DLCPayoutCurve Tests

### LINKED ISSUE #2389: Re-enable DLCPayoutCurveTest
We disabled them in #2388 until #2369 is resolved. The reason this is done is so we don't have flaky CI until the dlc spec is changed to avoid these off by one satoshi errors.

nkohen:
Fixes #2389 

`OutcomePayoutPoint` now has the correct types and defers rounding due to `extra_precision` in serialized points which was just added to the specification: [`c4fb12d` (#110)](https://github.com/discreetlogcontracts/dlcspecs/pull/110/commits/c4fb12d95a4255eabb873611437d05b740bbeccc)

I tested locally using the following in my coreTest build (.sbt) file
```
commands += Command.command("testUntilFail") { state =>
  "testOnly org.bitcoins.core.protocol.dlc.DLCPayoutCurveTest" :: "testUntilFail" :: state
}
```
left that running for a really long time and no failures now! I also attempted the above on current master after getting rid of the `ignore` statements and it did generate a failure in less than a minute.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
@@ -32,7 +32,7 @@ case class DLCPayoutCurve(points: Vector[OutcomePayoutPoint]) {
   /** Returns the function component on which the given oracle outcome is
     * defined, along with its index
     */
-  def componentFor(outcome: BigDecimal): Indexed[DLCPayoutCurveComponent] = {
+  def componentFor(outcome: Long): Indexed[DLCPayoutCurveComponent] = {
     val endpointIndex = NumberUtil.search(outcomes, outcome)
     val Indexed(endpoint, _) = endpoints(endpointIndex)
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
```diff
@@ -68,13 +68,22 @@ case class DLCPayoutCurve(points: Vector[OutcomePayoutPoint]) {
   * isEndpoint: True if this point defines a boundary between pieces in the curve
   */
 sealed trait OutcomePayoutPoint {
-  def outcome: BigDecimal
-  def payout: Satoshis
+  def outcome: Long
+  def payout: BigDecimal
   def isEndpoint: Boolean
 
+  def roundedPayout: Satoshis = {
```
benthecarman, line 75: nit: `lazy val`

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
```diff
@@ -68,13 +68,22 @@ case class DLCPayoutCurve(points: Vector[OutcomePayoutPoint]) {
...
+    Satoshis(payout.setScale(0, RoundingMode.FLOOR).toLongExact)
+  }
+
+  def extraPrecision: Int = {
```
benthecarman, line 79: nit: `lazy val`

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -862,7 +866,8 @@ case class TLVPoint(outcome: Long, value: Satoshis, isEndpoint: Boolean)
   override def bytes: ByteVector = {
```
benthecarman, line 866: nit: `val`

benthecarman (approved):

PR #2442, opened 2020-12-29
User input passwords on new wallet startup

### LINKED ISSUE #2220: Key Manager setup process in first App Server run
We should prompt the user for a AesPassword and a BIP39 password on the first run of an app server. This way we don't have to have them use the defaults always

benthecarman:
Closes #2220

Lets the user input a wallet password + bip 39 password when creating a new wallet. Works for completely new data dir as well as just setting up a new seed with the `bitcoin-s.wallet.walletName` config option.

### DIFF
--- a/app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
+++ b/app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
@@ -4,8 +4,10 @@ import akka.actor.ActorSystem
 import com.typesafe.config.{Config, ConfigFactory}
 import org.bitcoins.core.config._
 import org.bitcoins.core.util.{BitcoinSLogger, EnvUtil}
+import org.bitcoins.crypto.AesPassword
 import org.bitcoins.db.AppConfig
 import org.bitcoins.db.AppConfig.safePathToString
+import org.bitcoins.keymanager.config.KeyManagerAppConfig
 
 import java.nio.file.{Path, Paths}
 import scala.concurrent.{ExecutionContext, Future}
@@ -119,5 +121,35 @@ trait BitcoinSRunner extends BitcoinSLogger {
       logger.error(s"Failed to startup server!", err)
       sys.exit(1)
     }(scala.concurrent.ExecutionContext.Implicits.global)
+  }
+
+  def getUserPasswords: (Option[AesPassword], Option[String]) = {
+    val kmConf = KeyManagerAppConfig(datadir, baseConfig)
+    kmConf.start()
+
+    if (!kmConf.seedExists()) {
+      logger.info("Creating new wallet!")
+      // Print to user if their log level is too high
+      if (!logger.isInfoEnabled) {
+        println("Creating new wallet!")
+      }
+      print("Enter wallet password (empty for unencrypted wallet): ")
+      val aesPassStr: String = System.console().readPassword().mkString
+      val aesPasswordOpt = AesPassword.fromStringOpt(aesPassStr)
+
+      print("Confirm wallet password: ")
+      val confirmAesStr: String = System.console().readPassword().mkString
+      require(aesPassStr == confirmAesStr, "Wallet passwords did not match")
+
+      print("Enter wallet BIP 39 password (empty for none): ")
+      val bip39PassStr: String = System.console().readPassword().mkString
+      val bip39PassOpt = if (bip39PassStr.nonEmpty) Some(bip39PassStr) else None
+
+      print("Confirm BIP 39 password: ")
+      val confirmBip39Str: String = System.console().readPassword().mkString
+      require(bip39PassStr == confirmBip39Str, "BIP 39 passwords did not match")
+
+      (aesPasswordOpt, bip39PassOpt)
+    } else (None, None)
   }
 }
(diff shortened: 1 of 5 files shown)


app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
```diff
@@ -118,4 +120,34 @@ trait BitcoinSRunner extends BitcoinSLogger {
       sys.exit(1)
     }(scala.concurrent.ExecutionContext.Implicits.global)
   }
+
+  def getUserPasswords: (Option[AesPassword], Option[String]) = {
+    val kmConf = KeyManagerAppConfig(datadir, baseConfig)
+    kmConf.start()
```
Christewart, line 128: `kmConf.start()` returns a `Future`. We need to make this async safe by wrapping the stuff below in a `.map`

Christewart (commented):
So I've tried this flow and get a error (but a strange one)

1. Enter in bip39 password on startup
2. Enter aes password on startup
3. shutdown server
4. enter INCORRECT aes password in `bitcoin-s.conf`
5. start server again

This is the error 

>2020-12-30T23:56:54UTC ERROR [BitcoinSServerMain] Error on server startup!
java.lang.RuntimeException: Error initializing key manager, err=org.bitcoins.core.wallet.keymanagement.KeyManagerUnlockError$JsonParsingError: key not found: xprv

I also may have inadvertently had a aes password configured in both my `bitcoin-s.conf` and one entered from the command line the first time the wallet started up. What is the expected behavior here?

Christewart:
Another question, does this work ok with `bundle/run` ? If I understand correctly this will have to be prompted somehow from the GUI rather than the command line?

app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
```diff
@@ -118,4 +120,34 @@ trait BitcoinSRunner extends BitcoinSLogger {
...
+        println("Creating new wallet!")
+      }
+      print("Enter wallet password (empty for unencrypted wallet): ")
+      val aesPassStr: String = System.console().readPassword().mkString
+      val aesPasswordOpt = AesPassword.fromStringOpt(aesPassStr)
+
+      print("Confirm wallet password: ")
+      val confirmAesStr: String = System.console().readPassword().mkString
+      require(aesPassStr == confirmAesStr, "Wallet passwords did not match")
+
```
Christewart, line 143: So if I understand correctly, the second time the users fires up `appServer` they MUST have set the configs correctly in `bitcoin-s.conf`?

If that is the case, we should definitely link to resources for how to set these options. Basically link to the bitcoin-s.org configuration example IMO.

benthecarman:
> Another question, does this work ok with `bundle/run` ? If I understand correctly this will have to be prompted somehow from the GUI rather than the command line?

Yeah this won't work with the GUI. I am not sure how we would do that

app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
benthecarman: Maybe we should add it to the their conf manually?

Christewart: Modifying a users configuration file is probably not a good idea. It will lead to confusion in my opinion.

Christewart:
> > Another question, does this work ok with `bundle/run` ? If I understand correctly this will have to be prompted somehow from the GUI rather than the command line?
> 
> Yeah this won't work with the GUI. I am not sure how we would do that

Isn't this stuff just cli/server interactions? Can't this just be prompted in the GUI rather than the cli?

benthecarman:
> Isn't this stuff just cli/server interactions? Can't this just be prompted in the GUI rather than the cli?

No, this isn't done in the cli. This is done by using stdin in the server console

Christewart:
> > Isn't this stuff just cli/server interactions? Can't this just be prompted in the GUI rather than the cli?
> 
> No, this isn't done in the cli. This is done by using stdin in the server console

Well i think that would be the problem. It seems this needs to be thought out a bit better.

benthecarman:
> Well i think that would be the problem. It seems this needs to be thought out a bit better.

I am not sure how we would do it through the cli. We can't start the server without loading a wallet

benthecarman:
1. Setup server route
2. Server routes requires having a wallet
3. A wallet requires a key manager
4. Can't have a cli command for key manager because server requires ones

Christewart:
So if I understand correctly, if we don't add this change all user passwords will be required to be configured through `bitcoin-s.conf` ? I think that is probably the simplest solution after talking with you about this and realizing the complexity of this change. In our documentation -- as the 0.5 release stands I think we can just say 

>Passwords are configured through the `bitcoin-s.conf` file. The first time you set passwords you need to set the `aesPassword`/`bip39Password` setting _before_ the first time you start your server. If you want to change your password, you need to use the the [`keymanagerpassphrasechange`](https://bitcoin-s.org/docs/next/applications/server) rpc.

Is this accurate?

Another question, if you do this sequence 

1. start app server
2. stop app server
3. set passwords in `bitcoin-s.conf` 
4. restart app server

Will the key manager now be encrypted?

benthecarman:
> Passwords are configured through the bitcoin-s.conf file. The first time you set passwords you need to set the aesPassword/bip39Password setting before the first time you start your server. If you want to change your password, you need to use the the keymanagerpassphrasechange rpc.

> Is this accurate?

Yes

> Another question, if you do this sequence

> start app server
stop app server
set passwords in bitcoin-s.conf
restart app server
Will the key manager now be encrypted?

No, it reads the conf on startup

PR #2443, opened 2020-12-29
Rework BlockHeaderDAO.chainTips into two methods: BlockHeaderDAO.{get…

### LINKED ISSUE #2437: BlockHeaderDAO.chainTips doesn't actually retrieve all chainTips
If you take a look at our current implementation of [`BlockHeaderDAO.chainTips`](https://github.com/bitcoin-s/bitcoin-s/blob/40f46d7c7a82a300bf2be2f1e3b957789a80df2d/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala#L304)

```scala
  def chainTips: Future[Vector[BlockHeaderDb]] = {
    logger.debug(s"Getting chain tips from database")
    val aggregate = {
      maxWorkQuery.flatMap { work =>
        logger.debug(s"Max block work: $work")
        val atChainWork = getAtChainWorkQuery(work)
        atChainWork.map { headers =>
          logger.debug(s"Headers at $work: $headers")
        }
        atChainWork
      }
    }

    safeDatabase.runVec(aggregate)
  }
```

As you can see, this _only_ fetches the chainTip with the most work (`maxWorkQuery`). What this method _really_ implements is `maxWorkChainTip`. 

The test case that is problematic for this is what I cited in https://github.com/bitcoin-s/bitcoin-s/pull/2436#issuecomment-751716365

https://github.com/bitcoin-s/bitcoin-s/blob/master/chain-test/src/test/scala/org/bitcoins/chain/blockchain/ChainHandlerTest.scala#L170

After applying the fixes in #2436, we now cannot connect a header that _is_ valid, to a chainTip that is not the heaviest one.

Christewart:
…BestChainTips,getForkedChainTips}. getForkedChainTips is needed for reorg situations in the case a block header is received that builds off a stale tip

fixes #2437 

This changes the previous behavior of `BlockHeaderDAO.getBlockchains()`. Previously we would _only_ return blockchains that are based off of our best tips. 

It is possible to have a blockchain that is reorgs a stale blockchain (one that is not currently in the set of best tips). Our code previously did not account for this. Now we can find those stale chains with `BlockHeaderDAO.getForkedChainTips`. 

Now the behavior of `BlockHeaderDAO.getBlockchains()` is to 

1. Retrieve all blockchains associated with `BlockHeaderDAO.getBestChainTips`
2. Retrieve all stale blockchains within `difficultyInterval`

Now if we receive a new header, and it builds off the stale chain, we can add it to our chainstate rather than reject it.

### DIFF
--- a/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala
+++ b/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala
@@ -124,14 +124,14 @@ class BlockHeaderDAOTest extends ChainDbUnitTest {
       }
   }
 
-  it must "retrieve the chain tip saved in the database" in {
+  it must "retrieve the best chain tip saved in the database" in {
     blockHeaderDAO: BlockHeaderDAO =>
       val blockHeader = BlockHeaderHelper.buildNextHeader(genesisHeaderDb)
 
       val createdF = blockHeaderDAO.create(blockHeader)
 
       val chainTip1F = createdF.flatMap { _ =>
-        blockHeaderDAO.chainTips
+        blockHeaderDAO.getBestChainTips
       }
 
       val assert1F = chainTip1F.map { tips =>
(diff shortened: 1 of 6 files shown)


Christewart:
Synced mainnet just fine, syncing testnet now

Christewart:
This synced fine on testnet as well

benthecarman (approved):

PR #2444, opened 2020-12-29
Check if header is valid against in memory headers AND database headers

benthecarman:
I think this might be a simpler alternative to #2443 

This allows us to not rely on us getting every possible tip in memory when processing a header we receive. If a header we receive does not have it's prev header in memory then we will check if it is in our database instead. Because we are checking against our database it should not matter if we do not get the perfect set of headers from the database originally, so we should be able to do the most optimal thing in the normal case.

This should also fix the case where we are unable to process a reorg larger than 2016 blocks


Christewart:
So a couple of initial thoughts

1. I still think we should implement BlockHeaderDAO.getBlockchains() correctly as in #2443 . This work is orthogonal to that IMO.
2. [`ChainHandler.processHeadersWithChains()`](https://github.com/bitcoin-s/bitcoin-s/blob/3883edcedd16de8725ba33c219666aac5a83db48/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala#L113)  is to do as the name suggests -- give me a `Vector[BlockHeader]` and a `Vector[Blockchain]` and I will tell if you if the result is valid. This is sort of breaking that core assumption by also looking into the database for headers we could connect to. It might be useful to use this kind logic where `processHeadersWithChains()` is called? 

Let me think about it a bit more though.

benthecarman:
> 1. I still think we should implement BlockHeaderDAO.getBlockchains() correctly as in #2443 . This work is orthogonal to that IMO.

I see what you mean here, but I think leaving `getBlockchains` as is better with this solution. It allows us to do a much simpler query and should make the chain handler more efficient. 

>  2. `ChainHandler.processHeadersWithChains()` is to do as the name suggests -- give me a `Vector[BlockHeader]` and a `Vector[Blockchain]` and I will > tell if you if the result is valid. This is sort of breaking that core assumption by also looking into > the database for headers we could connect to. It might be useful to use this kind logic where `processHeadersWithChains()` is called?

I think this function still does what is says as advertised. It is able to process headers with the given chains. The chains are now given to make validating the headers more optimal and not rely on expensive database calls for the headers.

Christewart:
> > 1. I still think we should implement BlockHeaderDAO.getBlockchains() correctly as in #2443 . This work is orthogonal to that IMO.
> 
> I see what you mean here, but I think leaving `getBlockchains` as is better with this solution. It allows us to do a much simpler query and should make the chain handler more efficient.

Truthfully i haven't run this with our node yet (planning on doing tonight), but I don't think this change will have a performance impact. [`ChainHandlerCached`](https://github.com/christewart/bitcoin-s-core/blob/2de17bb4e47961fd3fe4e55d75909b42470e3c59/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandlerCached.scala#L45) reads from the database once to initialize it's state and then maintains in memory blockchains. I don't see how your solution makes `BlockHeaderDAO.getBlockchains()` more efficient, it seems to make `processHeadersWithChains()` less efficient, but maybe more correct? 

If I understand correctly, in the case of a cache miss (`blockchains` in `processHeadersWithChains()` not containing the previous header), it will perform an extra read on the database (`BlockHeaderDAO.findPrevHeaders()`) to be certain that the header we have is not part of our internal chainstate.

>
> > 1. `ChainHandler.processHeadersWithChains()` is to do as the name suggests -- give me a `Vector[BlockHeader]` and a `Vector[Blockchain]` and I will > tell if you if the result is valid. This is sort of breaking that core assumption by also looking into > the database for headers we could connect to. It might be useful to use this kind logic where `processHeadersWithChains()` is called?
> 
> I think this function still does what is says as advertised. It is able to process headers with the given chains. The chains are now given to make validating the headers more optimal and not rely on expensive database calls for the headers.

I disagree, the previous behavior would _only_ look at the given blockchains. `processHeadersWithChains()` was indifferent to how the `Vector[Blockchain]` was obtained (they could be cached in memory, they could have been read directly from the database, etc). Your implementation now does database reads for blockchains that the caller of the method may not be aware of. If the caller of the method isn't aware of blockchains stored in our internal database, we probably have bigger problems.

One scenario that could cause this off of the top of my head are really deep reorgs on testnet.

PR #2453, opened 2020-12-31
Use mainnet for default server

### LINKED ISSUE #2452: Set configuration options for mainnet on v0.5 release
We should set the [`reference.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/eaecc1c377be9d9b120a7637dbad848bff099724/db-commons/src/main/resources/reference.conf#L2) to default to mainnet and connect to our existing neutrino node so users can have a nice "out of the box" experience.

benthecarman:
closes #2452

### DIFF
--- a/app/server-test/src/test/resources/reference.conf
+++ b/app/server-test/src/test/resources/reference.conf
@@ -1,0 +1,8 @@
+bitcoin-s {
+    network = regtest
+
+    node {
+        mode = neutrino # neutrino, spv, bitcoind
+        peers = ["localhost"]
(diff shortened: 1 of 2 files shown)


Christewart:
Do you have any thoughts on switching the default fee provider? Is mempool.space still pretty good?

benthecarman:
> Do you have any thoughts on switching the default fee provider? Is mempool.space still pretty good?

I think they are pretty good

Christewart (approved):

PR #2436, opened 2020-12-27
Make ChainApi.processHeaders() return a failed future in the case we …

Christewart:
…have no valid headers

fixes #2365

Previously we would _always_ return a successful future from `ChainApi.processHeaders()`. In places like [`DataMessageHandler`](https://github.com/bitcoin-s/bitcoin-s/blob/7172b4a049b8a18568e5e69d3f947ea6b32d4803/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala#L244) we were assuming that if we received all invalid headers, we would have a failed future returned. Since this was incorrect, we would execute callbacks registered even in the headers were totally invalid.

In this particular case, the wallet project would see that we received a new header, and then would call out to the chain project to retrieve information about that header. Since the chain project  realized that particular header was invalid, it wouldn't store it. However it would return sucessful future like it had.

PR #2458, opened 2021-01-03
Only update sync height if we have block height

### LINKED ISSUE #2457: onBlockReceived Callback failed with error:
This was unveiled in #2456 

https://github.com/bitcoin-s/bitcoin-s/pull/2456/checks?check_run_id=1634596329#step:5:611

It seems we receive a ton of these errors when running `nodeTest/test`. These were uncovered in that PR because previously we were supressing ALL logs in the `nodeTest` project with the old `logback-test.xml` files.

benthecarman:
Fixes #2457

It seems we will sometime process blocks that aren't in the chain api yet when syncing in tests. This would cause error messages to appear.

This fixes it to only update the sync height if we actually can

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -58,8 +58,15 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
       res <- resF
 
       hash = block.blockHeader.hashBE
-      height <- chainQueryApi.getBlockHeight(hash)
-      _ <- stateDescriptorDAO.updateSyncHeight(hash, height.get)
+      heightOpt <- chainQueryApi.getBlockHeight(hash)
+      _ <- heightOpt match {
+        case Some(height) =>
+          stateDescriptorDAO.updateSyncHeight(hash, height)
+        case None =>
+          logger.warn(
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -58,8 +58,15 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
 
       hash = block.blockHeader.hashBE
-      height <- chainQueryApi.getBlockHeight(hash)
-      _ <- stateDescriptorDAO.updateSyncHeight(hash, height.get)
+      heightOpt <- chainQueryApi.getBlockHeight(hash)
+      _ <- heightOpt match {
+        case Some(height) =>
+          stateDescriptorDAO.updateSyncHeight(hash, height)
+        case None =>
+          logger.warn(
```
Christewart, line 66: probably should upgrade this as an `error` log as there really is a big problem if callbacks are being executed on blocks we don't have.

Christewart:
I don't think this actually fixes #2457 , rather just gives it a better error message. I think we need #2436 for a real fix

benthecarman:
> I don't think this actually fixes #2457 , rather just gives it a better error message. I think we need #2436 for a real fix

Looking at it again I think you are right

PR #2460, opened 2021-01-03
Recover errors in DataMessageHandler

benthecarman:
Discovered while reviewing #2436. Previously the node would stop working if any of the handlers would fail to process a message. This could mean that someone could send us data that could potentially stop the node. This fixes this problem by having the `DataMessageHandler` recover on errors.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -12,6 +12,7 @@ import org.bitcoins.node.{NodeCallbacks, NodeType, P2PLogger}
 
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.util.Try
+import scala.util.control.NonFatal
 
 /** This actor is meant to handle a [[org.bitcoins.core.p2p.DataPayload DataPayload]]
   * that a peer to sent to us on the p2p network, for instance, if we a receive a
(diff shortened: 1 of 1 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -297,12 +298,13 @@ case class DataMessageHandler(
...
-    resultF.failed.foreach {
-      case err =>
-        logger.error(s"Failed to handle data payload=${payload}", err)
+    resultF.failed.foreach { err =>
+      logger.error(s"Failed to handle data payload=${payload}", err)
     }
 
-    resultF
+    resultF.recoverWith {
+      case NonFatal(_) => Future.successful(this)
```
Christewart, line 306: LGTM, but let's add a unit test so we don't regress in the future?

benthecarman: I think we need #2436 before we can do that

Christewart (approved):

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
Christewart: I need this for the exact reason you mention on this PR, but I would still like a unit test for this after #2436 is in. I filed an issue #2501

PR #2461, opened 2021-01-03
2021 01 02 issue 2457

Christewart:
This simplifies sync logic in test cases, initial sync is now done in `NodeUnitTest.createNeutrinoNode()` 

This simplifies test cases. I was running into issues with debugging #2436

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -3,7 +3,7 @@ package org.bitcoins.node
 import org.bitcoins.core.currency._
 import org.bitcoins.core.protocol.script.MultiSignatureScriptPubKey
 import org.bitcoins.core.protocol.transaction.TransactionOutput
-import org.bitcoins.core.util.{EnvUtil, FutureUtil}
+import org.bitcoins.core.util.EnvUtil
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.crypto.ECPublicKey
 import org.bitcoins.rpc.client.common.BitcoindVersion
@@ -63,18 +63,9 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
   val TestFees: Satoshis = 2230.sats
 
   def nodeCallbacks: NodeCallbacks = {
-    val onBlockHeadersReceived: OnBlockHeadersReceived = { headers =>
-      logger.error(s"header callback=${headers.length}")
-      FutureUtil.unit
-    }
-
     val onBlock: OnBlockReceived = { block =>
       for {
         wallet <- walletF
-        _ = logger.error(
-          s"block callback for header hash=${block.blockHeader.hashBE.hex} tx.length=${block.transactions.length}")
-        _ = block.transactions.map(t =>
-          logger.error(s"tx.txIdBE=${t.txIdBE.hex}"))
         _ <- wallet.processBlock(block)
       } yield ()
     }
@@ -86,7 +77,6 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
     }
 
     NodeCallbacks(
-      onBlockHeadersReceived = Vector(onBlockHeadersReceived),
       onBlockReceived = Vector(onBlock),
       onCompactFiltersReceived = Vector(onCompactFilters)
     )
@@ -105,17 +95,9 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
           expectedAddresses: Int): Future[Boolean] = {
         for {
           confirmedBalance <- wallet.getConfirmedBalance()
-          _ = logger.error(
-            s"confirmedBalance=$confirmedBalance expectedConfirmedAmount=$expectedConfirmedAmount")
           unconfirmedBalance <- wallet.getUnconfirmedBalance()
-          _ = logger.error(
-            s"unconfirmedBalance=$confirmedBalance expectedUnconfirmedAmount=$expectedUnconfirmedAmount")
           addresses <- wallet.listAddresses()
-          _ = logger.error(
-            s"addresses=${addresses.length} expectedAddresses=$expectedAddresses")
           utxos <- wallet.listDefaultAccountUtxos()
-          _ =
-            logger.error(s"utxos=${utxos.length} expectedUtxos=$expectedUtxos")
         } yield {
           // +- fee rate because signatures could vary in size
           (expectedConfirmedAmount === confirmedBalance +- FeeRate.currencyUnit) &&
@@ -125,20 +107,30 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
         }
       }
 
+      //default wallet utxos are 3BTC, 2BTC, 1BTC
+      //our coin selection algorithm seems to be selecting
+      //the 3BTC utxo to spend, so we should have
+      //confirmed = 2BTC + 1BTC
+      //unconfirmed = 3 BTC - TestAmount - TestFees
       val condition1 = () => {
         condition(
-          expectedConfirmedAmount = 0.sats,
+          expectedConfirmedAmount = 3.bitcoin,
           expectedUnconfirmedAmount =
-            BitcoinSWalletTest.expectedDefaultAmt - TestAmount - TestFees,
+            3.bitcoin - TestAmount - TestFees,
           expectedUtxos = 3,
           expectedAddresses = 7
         )
       }
+
+      //this is just sending TestAmount back to us
+      //so everything should stay the same as above
+      //expected we should have received TestAmount back
+      //and have 1 more address/utxo
       val condition2 = { () =>
         condition(
-          expectedConfirmedAmount = 0.sats,
+          expectedConfirmedAmount = 3.bitcoin,
           expectedUnconfirmedAmount =
-            BitcoinSWalletTest.expectedDefaultAmt - TestFees,
+            (3.bitcoin - TestAmount - TestFees) + TestAmount,
           expectedUtxos = 4,
           expectedAddresses = 8
         )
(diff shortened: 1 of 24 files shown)


node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -184,10 +192,6 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
       val output = TransactionOutput(sats, spk)
 
       for {
-        _ <- node.sync()
-        _ <- NodeTestUtil.awaitSync(node, bitcoind)
-        _ <- NodeTestUtil.awaitCompactFiltersSync(node, bitcoind)
```
benthecarman, line 189: I am assuming there will be CI failures because we don't wait for these to be done syncing

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -107,43 +107,52 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
...
-        _ <- NodeTestUtil.awaitSync(node, bitcoind)
-        _ <- NodeTestUtil.awaitCompactFilterHeadersSync(node, bitcoind)
-        _ <- NodeTestUtil.awaitCompactFiltersSync(node, bitcoind)
-
         // send
         addr <- bitcoind.getNewAddress
         _ <- wallet.sendToAddress(addr, TestAmount, Some(FeeRate))
-
-        _ <-
+        //??? adding this since sendToAddress doesn't seem to broadcast ????
```
benthecarman: Yeah send to address just creates the tx, need to broadcast it with `wallet.broadcastTransaction`

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -107,43 +107,52 @@ class NeutrinoNodeWithWalletTest extends NodeUnitTest {
...
+        _ <- wallet.getUnconfirmedBalance()
+        _ <- wallet.getBalance()
+        hash1 <-
           bitcoind.getNewAddress
             .flatMap(bitcoind.generateToAddress(1, _))
+        _ <- wallet.getConfirmedBalance()
         _ <- NodeTestUtil.awaitSync(node, bitcoind)
         _ <- NodeTestUtil.awaitCompactFiltersSync(node, bitcoind)
-
+        _ <- bitcoind.getBlock(hash1.head)
```
benthecarman: this line doesn't do anything

benthecarman (changes requested):

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
Christewart: Syncing is now done here: https://github.com/bitcoin-s/bitcoin-s/pull/2461/files#diff-7075da7fcab619a015363e10a0f880b3b76b03e6b3cfd8daaf07ccbe53448f3cR584

I am running into some time out issues locally though wrt to `_ <- AsyncUtil.awaitConditionF(condition1)`

benthecarman: It starts syncing, but it will start running the test before it is done because it doesn't have the 
```
        _ <- NodeTestUtil.awaitSync(node, bitcoind)
        _ <- NodeTestUtil.awaitCompactFiltersSync(node, bitcoind)
```

Christewart: :man_facepalming:  [insert I am not a smart man forrest gump meme]

However this is uncovering another problem, our block count and filter count never seem to converge 

>[info] NeutrinoNodeWithWalletTest:
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99
blockCount=101 filterCount=99

Christewart: FWIW, this is the initial bug that blocked me on #2436 and I thought i had solved it, but I guess not. I'll keep digging.

Christewart: So the problem with this seems to be executing callbacks. We are hanging here forever

https://github.com/bitcoin-s/bitcoin-s/blob/0d4af2f83df4d247f9885c09d5c58d11e77c7178/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala#L129

Christewart: So the problem is this: 

We have a race condition between `walletF` and the nodeCallbacks being setup. 

Since we moved sync logic into the fixture creation, we cannot register wallet callbacks anymore from initial sync (we do not have access to a wallet!). 

We were getting around this before with 

```scala
  private var walletP: Promise[Wallet] = Promise()
  private var walletF: Future[Wallet] = walletP.future
```

and then everytime we would start a test case, we would complete the `Promise` with the wallet in the given to us by the fixture. 

```scala
      val NeutrinoNodeFundedWalletBitcoind(node, wallet, bitcoind, _) = param

      walletP.success(wallet)
```

and finally, after that promise is completed, we can actually execute our callbacks as defined

```scala
    val onCompactFilters: OnCompactFiltersReceived = { blockFilters =>
      for {
        wallet <- walletF
        _ <- wallet.processCompactFilters(blockFilters)
      } yield ()
    }
```

However now since we are doing sync logic in the fixture, we can no longer have the callbacks registered when the fixture is being started, else we will hang forever.

testkit/src/main/scala/org/bitcoins/testkit/wallet/FundWalletUtil.scala
```diff
@@ -68,11 +69,7 @@ trait FundWalletUtil {
       hashes <- bitcoind.getNewAddress.flatMap(bitcoind.generateToAddress(6, _))
     } yield (tx, hashes.head)
 
-    val fundedWalletF =
```
Christewart, line 71: This was the cause of a lot of problems in the node project, our wallet would have transactions without having the internal chainstate/wallet state (`wallet.getSyncHeight()`) did not match what the bitcoind has.

node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
```diff
@@ -143,10 +153,13 @@ class DataMessageHandlerTest extends NodeUnitTest {
...
-        _ = nodeConfig.addCallbacks(callbacks)
+        nodeCallbacks = NodeCallbacks.onCompactFilterReceived(callback)
+        _ = spv.nodeAppConfig.addCallbacks(nodeCallbacks)
+        dataMessageHandler =
+          DataMessageHandler(genesisChainApi)(spv.executionContext,
+                                              spv.nodeAppConfig,
+                                              spv.chainConfig)
 
-        dataMessageHandler = DataMessageHandler(genesisChainApi)
         _ <- dataMessageHandler.handleDataPayload(payload, sender)
```
benthecarman, line 163: I don't think the changes on this file are needed, #2476 should have fixed this

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -39,7 +39,7 @@ case class NeutrinoNode(
       _ <- peerMsgSender.sendGetCompactFilterCheckPointMessage(
         stopHash = bestHash.flip)
     } yield {
-      node
+      node.asInstanceOf[NeutrinoNode]
```
benthecarman, line 42: can we just use `this`

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -156,10 +156,9 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
   def stop(): Future[Node] = {
     logger.info(s"Stopping node")
     val disconnectF = for {
-      _ <- nodeAppConfig.stop()
-      _ <- chainAppConfig.stop()
       p <- peerMsgSenderF
       disconnect <- p.disconnect()
+      _ <- nodeAppConfig.stop()
```
benthecarman, line 161: Is there a reason we aren't stopping the chain?

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -73,15 +74,15 @@ case class SpvNode(
...
     for {
       node <- super.start()
       peerMsgSender <- peerMsgSenderF
+      _ <- AsyncUtil.retryUntilSatisfiedF(() => isConnected)
       _ <- peerMsgSender.sendFilterLoadMessage(bloomFilter)
     } yield {
       logger.info(s"Sending bloomfilter=${bloomFilter.hex} to $peer")
-      logger.info(s"Sending bloomfilter=${bloomFilter.hex} to $peer")
-      node
+      node.asInstanceOf[SpvNode]
```
benthecarman, line 85: can we just use `this`

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -58,19 +58,12 @@ trait NodeUnitTest extends BitcoinSFixture with EmbeddedPg {
   }
 
   override def afterAll(): Unit = {
-    Await.result(config.chainConf.stop(), 1.minute)
-    Await.result(config.nodeConf.stop(), 1.minute)
-    Await.result(config.walletConf.stop(), 1.minute)
```
benthecarman, line 63: I think if we remove these the db threads will stay active and slow things down

wallet/src/main/scala/org/bitcoins/wallet/sync/WalletSync.scala
```diff
@@ -0,0 +1,85 @@
...
+        case (wallet, nextBlock) =>
+          wallet.processBlock(nextBlock)
+      }
+    } yield syncedWallet
+
+    syncedWalletF
+  }
+
+  /** Syncs the wallet by walking backwards from the currentTip until we reach our wallet's best blockHash */
+  private def getBlocksToSync(
```
benthecarman, line 44: This function will go forever if the wallet's best block is reorged out. Should probably do some checks with the height.

wallet/src/main/scala/org/bitcoins/wallet/sync/WalletSync.scala
```diff
@@ -0,0 +1,85 @@
...
+            .map(Some(_))
+        }
+      }
+      blocks <- {
+        currentBlockOpt match {
+          case Some(currentBlock) =>
+            //loop again as we need to keep syncing
+            getBlocksToSync(wallet = wallet,
+                            currentTipBlockHashBE =
+                              currentBlock.blockHeader.previousBlockHashBE,
```
benthecarman, line 74: should validate that the prev block isn't genesis, otherwise this function could run forever

benthecarman (changes requested):

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
Christewart: You've got to be careful with things like `this`, if there ever is internal state modified or set by the  `super.start()` method, but you don't return the thing that is returned by `.start()` the state will not returned to the caller of this method. This is how caching bugs are introduced.

In general with programming, your method should return something (`NeutrinoNode` in this case). Even if `this` is the same type as the thing returned from another method, you should always return the thing given to you by the other method as you don't know if it modifies internal state currently, or will modify internal state in the future.

We should probably figure out how to get rid of the `.asInstanceOf[]` though as that is a sign of an anti pattern.

In short, in this instance you are right there isn't a big difference as of now, but we should really strive to return things that are given to us by methods -- like the `Future[Node]` in this case.

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
Christewart: It would appear so, but actually this is is a method not a `val`. I am for renaming the `config` method to `getFreshConfig()` or something like that indicate that this is a new config, not an existing one  associated with the test suite.

https://github.com/christewart/bitcoin-s-core/blob/defdf6b0e0726dd58c0eea21a83241850ae67a78/testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala#L65

benthecarman: Good catch, ACK on that

node/src/main/scala/org/bitcoins/node/Node.scala
Christewart: So IIRC this was causing issues in test cases because shutting down the node doesn't necessarily mean you want to shutdown the chain project. 

The node is _dependent_ upon the chain, but the chain _isn't_ dependent upon on the node. So therefore the node shouldn't be able to stop the chain project from operating.

In the case of `appSever` that may be true, but `wallet` also uses the chain project. I don't think we want to the chain to stop functioning necessarily just because we want to shutdown the node. They are independent projects.

Therefore the caller of the `node.stop()` method should make the conscious choice to shutdown chain as well, rather than the node doing it on it's own.

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
Christewart: https://github.com/bitcoin-s/bitcoin-s/issues/2493 

I'm not going to address on this PR

wallet/src/main/scala/org/bitcoins/wallet/sync/WalletSync.scala
Christewart: So my thoughts is that this should probably be addressed along with #2179

node-test/src/test/scala/org/bitcoins/node/networking/peer/DataMessageHandlerTest.scala
Christewart: ive rebased ontop of this so i guess it is necessary

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -686,4 +695,67 @@ object BitcoinSWalletTest extends WalletLogger {
     } yield ()
   }
 
+  /** Constructs callbacks for the wallet from the node to process blocks and compact filters */
+  def createNeutrinoNodeCallbacksForWallet(wallet: Wallet)(implicit
```
Christewart, line 699: Perhaps in the future we want to use the callbacks defined in 

https://github.com/bitcoin-s/bitcoin-s/blob/4e862b7297668272933101d73b6867dd211ea026/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L197

wallet/src/main/scala/org/bitcoins/wallet/sync/WalletSync.scala
Christewart: Done on line 60

benthecarman (approved):

PR #2464, opened 2021-01-03
Make KeyManager return better error messages

benthecarman:
Previously we would try to read the seed based on if we received a password or not, this would cause us to return a `JsonParsingError` when we would try to read an encrypted seed without a password or read an unencrypted seed with a password.

This flips this around so we see what kind of seed do we have and then parses & decrypts it. This makes it so we can return a `BadPasswordError` when the wallet gives a password when it shouldn't or vice versa.

Overall this should give better error messages when someone uses the wrong password for the key manager.

### DIFF
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
@@ -174,6 +174,12 @@ object WalletStorage extends KeyManagerLogger {
     }
   }
 
+  case class RawEncryptedSeed(
+      rawIv: String,
+      rawCipherText: String,
+      rawSalt: String,
+      rawCreationTime: Long)
+
   /** Reads the raw encrypted mnemonic from json,
     * performing no decryption
     */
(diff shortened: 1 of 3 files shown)


key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -174,20 +174,17 @@ object WalletStorage extends KeyManagerLogger {
...
-
+  private def readEncryptedMnemonicFromJson(
+      json: Value): Either[ReadMnemonicError, EncryptedSeed] = {
     import MnemonicJsonKeys._
     import ReadMnemonicError._
 
     val readJsonTupleEither: Either[
       ReadMnemonicError,
-      (String, String, String, Long)] = jsonE.flatMap { json =>
+      (String, String, String, Long)] = {
```
Christewart: perhaps we should make a type for this?

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -316,40 +300,59 @@ object WalletStorage extends KeyManagerLogger {
...
-
-          encryptedEither.flatMap { encrypted =>
-            encrypted.toMnemonic(passphrase) match {
-              case Failure(_) =>
-                encrypted.toExtPrivKey(passphrase) match {
-                  case Failure(exc) =>
-                    logger.error(s"Error when decrypting $encrypted: $exc")
-                    Left(ReadMnemonicError.DecryptionError)
-                  case Success(xprv) =>
+    import MnemonicJsonKeys._
```
Christewart, line 337: Is there any way we can refactor this to be more readable? This method is really starting to grow into something that is difficult to analyze if it's correct

benthecarman: Moved into a separate function. Should be easier to read now

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
benthecarman: eh, we could but it'd only be used here. I don't think it's a big deal

Christewart: It makes code reviewers and maintainers lives a lot easier to be able to access things like `.rawIv`, `.rawCipherText` etc rather than having to do `._1` `._2` `._4` etc. It's very hard to remember the ordering of tuples, _especially_ when most of them are of the same type (`String` in this case).

Christewart (approved):

PR #2465, opened 2021-01-04
Update DLC Oracle Signing Algo

benthecarman:
Updates the oracle to be compliant with https://github.com/discreetlogcontracts/dlcspecs/pull/113

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/SigningVersion.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/SigningVersion.scala
@@ -72,9 +72,39 @@ object SigningVersion extends StringFactory[SigningVersion] {
     }
   }
 
-  val latest: SigningVersion = BasicSHA256SigningVersion
+  /** V0 DLC Oracle singing algo, specified in https://github.com/discreetlogcontracts/dlcspecs/pull/113 */
+  final case object DLCOracleV0SigningVersion extends SigningVersion {
 
-  val all: Vector[SigningVersion] = Vector(Mock, BasicSHA256SigningVersion)
+    override def calcNonceTweak(
+        nonce: SchnorrNonce,
+        eventName: String): ByteVector = {
+      val bytes = nonce.bytes ++ CryptoUtil.serializeForHash(eventName)
+
+      CryptoUtil.taggedSha256(bytes, "DLC/oracle/nonce/v0").bytes
+    }
+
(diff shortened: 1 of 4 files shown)


Christewart (approved):
So this LGTM, is there going to be any migraiton issues from the `BasicSHA256SigningVersion` -> `DLCOracleV0SigningVersion` ?

benthecarman:
> So this LGTM, is there going to be any migraiton issues from the `BasicSHA256SigningVersion` -> `DLCOracleV0SigningVersion` ?

Nope! That's the beauty of us adding the `SigningVersion` :)

Christewart:
> > So this LGTM, is there going to be any migraiton issues from the `BasicSHA256SigningVersion` -> `DLCOracleV0SigningVersion` ?
> 
> Nope! That's the beauty of us adding the `SigningVersion` :)

Hmm. Maybe I am missing something here, but if I produced an oracle announcement that correlates to the previous signing version, wouldn't I need to use that signing version for the actual announcement? I.e. this change can lead us to produce invalid digital signatures for an announcement that was created before this newest signing version?

benthecarman:
> Hmm. Maybe I am missing something here, but if I produced an oracle announcement that correlates to the previous signing version, wouldn't I need to use that signing version for the actual announcement? I.e. this change can lead us to produce invalid digital signatures for an announcement that was created before this newest signing version?

Not too sure how this would occur. If you are making an announcement you don't have any data in your database.

Christewart:
1. Yesterday (`2021-01-04`) I made an oracle announcement and published it on the oracle explorer. This announcement expires this weekend (`2021-01-08`)
2. I upgraded my version of krystal bull to include this change
3. Fast forward to `2021-01-08`
4. I use my upgraded version of krystal bull to produce signatures that corresponds to my announcement that I created on `2021-01-04` 

Doesn't this use the signing version latest rather than the original signing version (`BasicSHA256SigningVersion`) I published the announcement with?

benthecarman:
No this will work correctly. It is saved in your database which signing version you used to create your announcement, so when you go to sign the actual outcomes it will use the correct signing version `BasicSHA256SigningVersion`

Christewart:
Ok, so this is a implementation concern, and we implement it correctly is what I am understanding.

benthecarman:
> Ok, so this is a implementation concern, and we implement it correctly is what I am understanding.

Yes

PR #2471, opened 2021-01-05
Limit bech32 addresses to segwitv0

benthecarman:
It seems Bitcoin is moving to make bech32 addresses only for segwit v0 and a new address type for segwit v1+ called bech32m. This does the preliminary work for bech32m and limits bech32 addresses to be only for segwit v0.

https://github.com/sipa/bips/blob/bip-bech32m/bip-bech32m.mediawiki

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -1212,7 +1212,7 @@ sealed abstract class WitnessScriptPubKeyV0 extends WitnessScriptPubKey {
   override def witnessProgram: Seq[ScriptToken] = asm.tail.tail
 }
 
-object WitnessScriptPubKeyV0 {
+object WitnessScriptPubKeyV0 extends ScriptFactory[WitnessScriptPubKeyV0] {
 
   /**
     * Mimics the function to determine if a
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -1227,6 +1227,16 @@ object WitnessScriptPubKeyV0 {
     WitnessScriptPubKey.isWitnessScriptPubKey(asm) && asm.headOption.contains(
       OP_0)
   }
+
+  def apply(asm: Seq[ScriptToken]): WitnessScriptPubKeyV0 = fromAsm(asm)
+
+  def fromAsm(asm: Seq[ScriptToken]): WitnessScriptPubKeyV0 =
```
Christewart: nit: `override` needed I think?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -1227,6 +1227,16 @@ object WitnessScriptPubKeyV0 {
...
+    asm match {
+      case _ if P2WPKHWitnessSPKV0.isValid(asm) =>
+        P2WPKHWitnessSPKV0.fromAsm(asm)
+      case _ if P2WSHWitnessSPKV0.isValid(asm) =>
```
Christewart, line 1237: Let's add an explicit case for this

```scala
case _ => sys.error(s"The given asm was not a valid WitnessScriptPubKeyV0, got asm=$asm")
```

Christewart (changes requested):
Good idea, glad we can get this in before 0.5

Christewart (approved):

Issue #1802, opened 2020-08-11
Things to test before release


benthecarman:
A list of things we should test before a release

- [x] Initial Block Download

- [x] Rescanning chain

- [x] Receiving funds

- [x] Sending funds

- [x] All RPC commands

- [x] GUI

- [x] Creating a new wallet

- [x] Config options work as intended

- [ ] Logging

Please list more below

Christewart:
We should probably add "bouncy castle" to the list. Basically starting everything and running with bouncy castle instead of secp. I don't think we necessarily need to duplicate everythign on the list, but basic RPC stuff makes sense to me.

Christewart:
Test all of our different possible logging configs 

1. Logback and ignoring all `bitcoin-s.conf` logging
2. Just set logging in the `bitcoin-s.conf` and not use logback at all

Christewart:
We should also figure out what java versions we work on, I run on java8 and haven't seen any problems. Does anyone run things on java11/java13?

Christewart:
It looks like the "official" java version oracle directs people to is java 14, so perhaps we should test that too? :man_shrugging: 

![Screenshot from 2020-08-26 09-57-00](https://user-images.githubusercontent.com/3514957/91321861-b238fd00-e784-11ea-8d93-4e8dbf766390.png)

Christewart:
We should recommend that people use at least java 11, GUI does not work with java 8. 

>[error] java.lang.UnsupportedClassVersionError: javafx/event/EventTarget has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error]         at java.lang.ClassLoader.defineClass1(Native Method)
[error]         at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
[error]         at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[error]         at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
[error]         at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
[error]         at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
[error]         at java.net.URLClassLoader$1.run(URLClassLoader.java:363)


I think there is an issue for this somewhere, and it _really_ should be noted here on the website: 

https://bitcoin-s.org/docs/next/applications/gui#bitcoin-s-graphical-user-interface

benthecarman:
[...]

PR #2473, opened 2021-01-05
Add download link to navbar & downloads page

### LINKED ISSUE #2310: Make webpage that allows downloading standalone binaries on the website
Basically you should be able to download a jar file from the website for `appServer` and run it

benthecarman:
Fixes #2310 

image

image

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -476,6 +476,7 @@
     },
     "links": {
       "Docs": "Docs",
+      "Download": "Download",
       "API": "API",
       "Help": "Help"
     },
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -344,133 +344,133 @@
       "version-0.3.0/wallet/version-0.3.0-wallet": {
         "title": "Wallet"
       },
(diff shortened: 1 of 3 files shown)


Christewart:
So i don't have any problems with this change per se, but if I just look at our current release pages on github it's very hard to find what you would need to download and run. 

I know v0.4 has long releases notes because it's a large release, but we need to scroll all the way down to the bottom to find binaries. I think that is probably too hard for people to find? Especially for normies. 

It would be very nice if the binaries could be directly downloaded from the site. For example you embed the links to the `.deb` / `.msi` / whatever apple gives us on the website. Those links can be hosted by github at the end of the day, but we really need some sort of page that asks you what your OS is and then links to the appropriate file


Screenshot from 2021-01-06 06-44-45

Christewart (approved):
So this looks good, but a question:

Will these pages have to be manually updated when we do the 0.5 release, or will things just work :tm: if we release a 0.5 website? 

Obviously we will need to make sure we publish all binaries that are on the release page, we should probably note this on our release issue. For 0.4 our release issue was #1802

PR #2476, opened 2021-01-06
Remove callbacks param from DataMessageHandler & PeerMessageReceiver

### LINKED ISSUE #2475: Remove `callbacks` parameter from DataMessageHandler
Currently `DataMessageHandler` parameters look like this: 

```scala
case class DataMessageHandler(
    chainApi: ChainApi,
    callbacks: NodeCallbacks,
    initialSyncDone: Option[Promise[Done]] = None,
    currentFilterBatch: Vector[CompactFilterMessage] = Vector.empty,
    filterHeaderHeightOpt: Option[Int] = None,
    filterHeightOpt: Option[Int] = None,
    syncing: Boolean = false)(implicit
    ec: ExecutionContext,
    appConfig: NodeAppConfig,
    chainConfig: ChainAppConfig)
    extends P2PLogger
```

There is a problem here: We have two `nodeCallbacks`! One that is the parameter in the case class, and the other that is associated with `appConfig: NodeAppConfig`. 

We want to use the latter. We allow users to register callbacks at any point and we maintain these callbacks internally with a [mutable data structure](https://github.com/bitcoin-s/bitcoin-s/blob/18755df3d1738a1de968f1e653fbf7b4d23a0582/node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala#L38).

This specifically is coming up in #2461 wrt to refactoring testkit to actually work with the spv node. The problem is that we are passing in the parameter `callbacks` [here in the DataMessageHandler](https://github.com/bitcoin-s/bitcoin-s/blob/663a9514301a963ff11841cc56fe1d281c33a480/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala#L282). 

There is a subtle bug on this line, we are using the static `callbacks` rather than the `appConfig.nodeCallbacks` which contain callbacks that have been registered since the start of the `DataMessageHandler`.

https://github.com/bitcoin-s/bitcoin-s/blob/663a9514301a963ff11841cc56fe1d281c33a480/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala#L26

benthecarman:
Closes #2475

`PeerMessageReceiver` also had a `NodeCallbacks` param so that was removed as well.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
@@ -8,7 +8,6 @@ import org.bitcoins.core.p2p._
 import org.bitcoins.core.protocol.CompactSizeUInt
 import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.crypto.{CryptoUtil, DoubleSha256Digest}
-import org.bitcoins.node.NodeCallbacks
 import org.bitcoins.node.models.Peer
 import org.bitcoins.node.networking.peer.PeerMessageReceiver
 import org.bitcoins.testkit.async.TestAsyncUtil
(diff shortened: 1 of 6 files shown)


Christewart:
> `PeerMessageReceiver` also had a `NodeCallbacks` param so that was removed as well.

Thanks, i was going to comment this. In general, we should look for all places in the code base that have explicit callback parameters _with_ an `implicit appConfig` associated with them. These are bugs waiting to happen

benthecarman:
> In general, we should look for all places in the code base that have explicit callback parameters _with_ an `implicit appConfig` associated with them. These are bugs waiting to happen

It looks like only places now where we have callbacks as parameters are in the testkit, and we just pass them to the app config to add.

Christewart (approved):

PR #2478, opened 2021-01-06
Rename v0.4 version to 0.4.0 in docs

benthecarman:
The version number needs to match what we used on github. Also it can't start with `v`. I used the [renaming tool](https://docusaurus.io/docs/en/versioning#renaming-existing-versions) to do this.


Christewart (approved):
LGTM, i suppose this will break existing links though if we linked to a specific version on other websites?

I guess there probably is no way around that.

benthecarman:
> LGTM, i suppose this will break existing links though if we linked to a specific version on other websites?
> 
> I guess there probably is no way around that.

Yeah, I'll take a pass through the release notes and see if there are places we need to fix.

Edit: Looks like all of our links in release notes go to the latest docs so it shouldn't break those.

PR #2480, opened 2021-01-06
Implement oracle signing algo for client side

benthecarman:
https://github.com/discreetlogcontracts/dlcspecs/pull/113


app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -156,7 +156,10 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea) {
                                              "dummy oracle")
 
           val announcementSig =
-            privKey.schnorrSign(CryptoUtil.sha256(oracleEvent.bytes).bytes)
+            privKey.schnorrSign(
+              CryptoUtil
+                .taggedSha256(oracleEvent.bytes, "DLC/oracle/announcement/v0")
+                .bytes)
```
nkohen: all invocations of `taggedSha256` should be replaced with a new function in `CryptoUtil` akin to 
https://github.com/bitcoin-s/bitcoin-s/blob/master/crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala#L71
including the per-compiled val for the tag's hash.

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -172,7 +175,9 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea) {
               builder.append("Outcomes and oracle sigs in order of entry:\n")
               contractInfo.keys.foreach { outcome =>
                 val bytes = outcome.serialized.head
-                val hash = CryptoUtil.sha256(bytes).bytes
+                val hash = CryptoUtil
+                  .taggedSha256(bytes, "DLC/oracle/attestation/v0")
```
nkohen: ditto for attestations as well

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestUtil.scala
```diff
@@ -67,7 +67,7 @@ object DLCTestUtil {
     val func = DLCPayoutCurve(
       Vector(
         OutcomePayoutEndpoint(0, leftVal),
-        OutcomePayoutEndpoint(botCollar, leftVal),
+        OutcomePayoutEndpoint(botCollar + 1, leftVal),
```
nkohen, line 80: I assume this fix made tests pass in the `testUntilFailed` command?

nkohen (changes requested):
Concept ACK

nkohen:
Also, so you know, me and Tibo agreed that this is acceptable to get in now, we're both willing to work around this. We just need to make sure to remember to apply the diff to https://github.com/discreetlogcontracts/dlcspecs/pull/125

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTestUtil.scala
benthecarman: Yes

nkohen (commented):
ACK after you add a scaladoc to each of the `_TagBytes` vals in `CryptoUtil.scala` containing the string that was used to generate these (as, I hope, is done for the other vals like this in that file)

nkohen (approved):

Issue #2070, opened 2020-09-28
Remember to publish bitcoin-s-cli in our tarballs for next release


Christewart:
We forgot to do this for v0.4 

This is really nice to have in a release

Christewart:
This was taken care of #2494

PR #2494, opened 2021-01-09
Add new sbt-native-image plugin that helps generate a correct native …

Christewart:
…image for the cli

This seems to work out of the box with our `cli` which is very exciting. This can be used when we ship binaries for the cli as per #2070 

https://github.com/scalameta/sbt-native-image

You can build the cli with 

`cli/nativeImage`

This will even download graalvm native image for you, before building the cli. So you just literally need to run the command. 

I did test this out, and it appears things are working _except_  `--version`

```scala
./bitcoin-s-cli getblockcount
1903161
./bitcoin-s-cli getnewaddress
myJf9nBETPST8jWDbTdpEK73RzwH5Rii8b
./bitcoin-s-cli --version
null
```

### DIFF
--- a/.github/workflows/native.yml
+++ b/.github/workflows/native.yml
@@ -18,13 +18,13 @@ jobs:
         include:
           - os: macOS-latest
             uploaded_filename: bitcoin-s-cli-x86_64-apple-darwin
-            local_path: cli/target/native-image/bitcoin-s-cli
+            local_path: app/cli/target/native-image/bitcoin-s-cli
           - os: ubuntu-latest
             uploaded_filename: bitcoin-s-cli-x86_64-pc-linux
-            local_path: cli/target/native-image/bitcoin-s-cli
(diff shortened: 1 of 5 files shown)


Christewart:
I spent some time fiddling around to try and get the server working as well, but I got stuck. I've filed a comment on the open issue that corresponds to the issue I had 

https://github.com/oracle/graal/issues/2723#issuecomment-757220785

benthecarman (approved):

Christewart:
If we get this working reliably it seems we may be able to automatically generate native binaries for various platforms: https://github.com/scalameta/sbt-native-image#generate-native-image-from-github-actions

Christewart:
So as of ee34df2530505279f794dc2b18a606a3d1479c01 i think i have ~both the `appServer` and `cli`~ just cli projects working.

I'm still doing testing locally to make sure everything is ok.

PR #2499, opened 2021-01-09
add Opt/T fromBytes/fromHex methods to Factory so we have the same basic methods as StringFactory

Christewart:
Basically do what we have on `StringFactory` on our normal `Factory`


benthecarman (approved):
ACK, should eventually do a pass through of all of our from bytes implementations and see which can be used to override the option and try functions instead.

PR #2503, opened 2021-01-11
2021 01 11 issue 2493

### LINKED ISSUE #2493: Rename NodeUnitTest.config -> NodeUnitTest.getFreshConfig()
https://github.com/bitcoin-s/bitcoin-s/pull/2461#discussion_r553561794

Christewart:
fixes #2493 

Previously we had a _method_ named `NodeUnitTest.config`. I think a lot of people thought this was a cached config file in the super trait `NodeUnitTest`. In reality it was a creating a _fresh_ `BitcoinSAppConfig` everytime it was called. This led to incorrect uses through out the `nodeTest` project.


benthecarman:
I believe we do this in the other test suites as well, might be worth doing a pass through

Christewart:
> I believe we do this in the other test suites as well, might be worth doing a pass through

Good call, done in 19a4ab1

benthecarman (approved):
ACK on CI

PR #2516, opened 2021-01-14
Implement OracleAttestmentV0TLV, save outcome to db

benthecarman:


app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
```diff
@@ -195,7 +195,13 @@ case class OracleRoutes(oracle: DLCOracle)(implicit
           complete {
             oracle.signEvent(oracleEventTLV, EnumAttestation(outcome)).map {
               eventDb =>
-                Server.httpSuccess(eventDb.sigOpt.get.hex)
+                val oracleEvent = OracleEvent.fromEventDbs(Vector(eventDb))
+                oracleEvent match {
+                  case _: PendingOracleEvent =>
+                    throw new RuntimeException("Failed to sign event")
```
Christewart, line 203: Why are we throwing here? Shouldn't we be using `Server.httpError()` or `reject()`? 

This applies in other places too in this file.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -208,6 +226,20 @@ object OracleEvent {
 
         val attestations = sortedEventDbs.flatMap(_.attestationOpt)
 
+        val outcomes = decomp match {
+          case _: SignedDigitDecompositionEventDescriptor =>
+            val sign = DigitDecompositionSignAttestation(
+              sortedEventDbs.head.outcomeOpt.get == "+")
+            val digits = sortedEventDbs.tail.map { eventDb =>
+              DigitDecompositionAttestation(eventDb.outcomeOpt.get.toInt)
```
Christewart, line 237: Why is it safe to access `eventDb.outcomeOpt.get`? I don't see where we are guranteed by the type system that this will always be defined

Christewart (changes requested):
Where is this new functionality tested? AFAICT in test cases we have simply added `None` to some `dummy` events. There is this, but it isn't entirely apparent to me this is actually testing this new functionality 

https://github.com/bitcoin-s/bitcoin-s/pull/2516/files#diff-b524842fc6e758032c89b854b5e889d12ba52f998f6cb8ac752497f5aacc12b4R481

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
benthecarman: This is a signed/completed event, so it should have the outcome set

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
benthecarman: throwing will cause a `Server.httpError`

benthecarman: Also, looks like we can't replace it with a `Server.httpError`, it causes an compiler error

benthecarman:
>  There is this, but it isn't entirely apparent to me this is actually testing this new functionality

> #2516 (files)

That is testing we are saving the correct outcome

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
nkohen: I think throwing here is fine, but if you were interested in the marginally more efficient solution it would be using `flatMap` in place of `map` and then using `Future.failed` and `Future.successful` as this throw is being caught by the future and passed up, but catching takes more overhead than flattening

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -817,6 +818,57 @@ object OracleAnnouncementV0TLV extends TLVFactory[OracleAnnouncementV0TLV] {
   }
 }
 
+sealed trait OracleAttestmentTLV extends TLV {
+  def eventId: NormalizedString
+  def publicKey: SchnorrPublicKey
+  def sigs: Vector[SchnorrDigitalSignature]
+  def outcomes: Vector[NormalizedString]
```
nkohen, line 838: super nit (can ignore): But in the case where you make a trait with only one extension such as here where you don't have a future extension in mind, I think it is generally nicer to force people to match to get these `def`s rather than providing them unless you have some reason to believe that future `OracleAttestmentTLV`s are guaranteed to have these fields

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -817,6 +818,57 @@ object OracleAnnouncementV0TLV extends TLVFactory[OracleAnnouncementV0TLV] {
...
+}
+
+object OracleAttestmentV0TLV extends TLVFactory[OracleAttestmentV0TLV] {
+  override val tpe: BigSizeUInt = BigSizeUInt(55400)
+
+  override def fromTLVValue(value: ByteVector): OracleAttestmentV0TLV = {
+    val iter = ValueIterator(value)
+
+    val eventId = iter.takeString()
+    val pubKey = SchnorrPublicKey(iter.take(32))
```
nkohen: super nit: I personally think `iter.take(SchnorrPublicKey, 32)` is more idiomatic

nkohen (approved):
Just some nits

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -817,6 +818,64 @@ object OracleAnnouncementV0TLV extends TLVFactory[OracleAnnouncementV0TLV] {
...
+    publicKey: SchnorrPublicKey,
+    sigs: Vector[SchnorrDigitalSignature],
+    outcomes: Vector[NormalizedString])
+    extends OracleAttestmentTLV {
+  require(sigs.nonEmpty, "Cannot have 0 signatures")
+  require(
+    outcomes.size == sigs.size,
+    s"Number of outcomes must match number of signatures, ${outcomes.size} != ${sigs.size}")
+  override val tpe: BigSizeUInt = OracleAttestmentV0TLV.tpe
+
```
rorp, line 860: `tpe` again?

Christewart: It's in the super trait, we have to rename all of them else we have to continue with naming convention for now 

https://github.com/bitcoin-s/bitcoin-s/blob/abc1fdd23fb3480fb3bde10efe1d1f0389211bd7/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala#L22

rorp (changes requested):
`OracleEvent.verifyAttestations` should be updated

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
rorp: It takes only a few clicks to rename them all in IntelliJ IDEA.

Christewart: We have different branches that are dependent on this code (adaptor-dlc, all of the branches that are dependent on adaptor-dlc)

benthecarman:
> `OracleEvent.verifyAttestations` should be updated

Updated

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
```diff
@@ -239,10 +275,15 @@ object OracleEvent {
...
   def verifyAttestations(
       announcement: OracleAnnouncementTLV,
-      attestations: Vector[SchnorrDigitalSignature],
+      attestationTLV: OracleAttestmentTLV,
       signingVersion: SigningVersion): Boolean = {
+    val attestations = attestationTLV.sigs
     val nonces = announcement.eventTLV.nonces
-    if (nonces.size != attestations.size || nonces != attestations.map(_.rx)) {
+    if (
+      announcement.publicKey != announcement.publicKey ||
```
rorp: I think it's always `false`?

rorp: Should we also use the attestation outcomes for verification somehow? @nkohen

rorp (changes requested):

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/OracleEvent.scala
benthecarman: Fixed to compare against correct key

nkohen: I don't think there's any reason we couldn't check the outcomes since they're there

nkohen: for enum that should just be a contains check and for multi-nonce just use the base and match against the regex for the correct size with the correct number of digits

rorp (approved):

Issue #2355, opened 2020-12-14
TODO after #2339


nkohen:
#2339 pulled down work from `adaptor-dlc` onto `master` and has some unfinished business.

TODO before release:
* Move from appCommons to core
* Make OutcomeValuePoint into an ADT (splitting on `isEndpoint`)
* OutcomeValueFunction -> DLCPayoutCurve
* OutcomeValueFunctionComponent -> DLCPayoutCurveComponent
* OutcomeValuePoint -> OutcomePayoutPoint
* Replace tuples in OutcomeValueFunction and RoundingIntervals with types
* Create overloaded SchnorrPublicKey.computeSigPoint(Sha256Digest, SchnorrNonce)
* Create util method to `reduce` list of digests and nonces to a singe point using computeSigPoint
* Add to website

Other TODOs:
* Needs tests:
    - ECDigitalSignature.fromFrontOfBytes[WithSigHash]
    - SatoshisPerVirtualByte.toSatoshisPerKW
    - RawTxFinalizers
    - `DualFundingTxFinalizer.computeFees`
    - sigFromMsgAndSigs
* Move local methods in `calculateOutcomeAndSig` out into private helper methods

nkohen:
Closed by #2613

PR #2518, opened 2021-01-15
2021 01 15 dlc refactors

Christewart:
Addresses parts of #2355 

Specifically 

1. Kills `OutcomePayoutPoint.isEndpoint` & `DigitDecomp.isSigned` (if there is strong objection for this, i'm ok with re-adding, but it does promote incorrect usage IMO. It's better to use ADTs directly as they are more expressive -- they can represent more than two states unlike `Boolean`. ADTs represent our various states our data structures can be in, not boolean flags). 
2. Creates an ADT for `TLVPoint` with two subtypes `TLVMidpoint` and `TlvEndpoint`. THIS DOES NOT CHANGE BINARY LEVEL SERIALIZATION. This is simply internal types to use in conjuction with `OutcomePayoutEndpoint` & `OutcomePayouMidpoint`

While this was pretty painless to refactor what we have in master, my understanding is this will be more tedious complex on `adaptor-dlc`. I will volunteer to do this refactor up there (i do have concerns about actually doing the rebase though, which I might not have a good enough understanding of the code to actually do this). 

Other musings from this PR, we have a really unfortunate pattern with nonces currently wrt to `UnsignedDigitDecomp` and `SignedDigitDecomp`. Unfortunately we cannot handle these in the `EventDescriptors` themselves as we don't have access to the nonces. 

Perhaps we can try to make a helper method in `OracleEvent` to handle this in the future.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.core.protocol.dlc
 
 import org.bitcoins.core.currency.Satoshis
-import org.bitcoins.core.protocol.tlv.{TLVEndpoint, TLVMidpoint, TLVPoint}
+import org.bitcoins.core.protocol.tlv.TLVPoint
 import org.bitcoins.core.util.{Indexed, NumberUtil}
 
 import scala.math.BigDecimal.RoundingMode
@@ -105,9 +105,15 @@ sealed trait OutcomePayoutPoint {
   def toTlvPoint: TLVPoint = {
     this match {
       case _: OutcomePayoutEndpoint =>
-        TLVEndpoint(outcome, roundedPayout, extraPrecision)
+        TLVPoint(outcome = outcome,
+                 value = roundedPayout,
+                 extraPrecision = extraPrecision,
+                 isEndpoint = true)
       case _: OutcomePayoutMidpoint =>
-        TLVMidpoint(outcome, roundedPayout, extraPrecision)
+        TLVPoint(outcome = outcome,
+                 value = roundedPayout,
+                 extraPrecision = extraPrecision,
+                 isEndpoint = false)
     }
   }
 }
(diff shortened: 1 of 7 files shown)


benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
```diff
@@ -179,9 +198,10 @@ sealed trait DLCPayoutCurveComponent {
 object DLCPayoutCurveComponent {
 
   def apply(points: Vector[OutcomePayoutPoint]): DLCPayoutCurveComponent = {
-    require(points.head.isEndpoint && points.last.isEndpoint,
+    require(points.head.isInstanceOf[OutcomePayoutEndpoint] && points.last
+              .isInstanceOf[OutcomePayoutEndpoint],
```
nkohen: I personally find having a boolean function much nicer (and more efficient for Java) for these kinds of uses. I don't see any reason for getting rid of the `def isEndpoint`

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -860,17 +873,16 @@ object ContractInfoV0TLV extends TLVFactory[ContractInfoV0TLV] {
   }
 }
 
-case class TLVPoint(
-    outcome: Long,
-    value: Satoshis,
-    extraPrecision: Int,
-    isEndpoint: Boolean)
-    extends NetworkElement {
+sealed trait TLVPoint extends NetworkElement {
```
nkohen: I'm relatively indifferent here, but I think this is maybe the wrong approach for a type that is explicitly meant to represent a TLV, we have our internal point type separate from this one for the purpose of allowing internal representation but this external representation type should probably more directly reflect what we see over the wire (which is a bool flag). No operations are done without converting these to internal types, the only logic applied to these is serialization and de-serialization which is made marginally more convoluted by this

nkohen (commented):
Aside from bringing back `def isEndpoint` and probably reverting the TLVPoint diff, I am generally indifferent to most of the changes made here. I'm not sure that there really is any benifit in these cases to directly using case matching on the ADT rather than using some higher level interface-provided property as happens in these particular uses

PR #2535, opened 2021-01-18
Create isChange function for wallet

### LINKED ISSUE #2406: Better abstraction to determine if an output is a change output
This came up in #2399 

We have to look into the keychain to determine if this output is a change output which is not very nice. At minimum we should probably add this to `WalletApi`
```scala

def isChange(output: TransactionOutput): Boolean = {
   ??? // can we just use what we have in 2399 to start?
} 

``` 
https://github.com/bitcoin-s/bitcoin-s/pull/2399#discussion_r546379338

benthecarman:
Fixes #2406

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
@@ -392,6 +392,7 @@ trait WalletApi extends StartStopAsync[WalletApi] {
     } yield tx
   }
 
+  def isChange(output: TransactionOutput): Future[Boolean]
 }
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
```diff
@@ -392,6 +392,7 @@ trait WalletApi extends StartStopAsync[WalletApi] {
     } yield tx
   }
 
+  def isChange(output: TransactionOutput): Future[Boolean]
```
Christewart, line 398: scaladoc

Christewart (approved):
Just add a scaladoc otherwis LGTM

PR #2542, opened 2021-01-20
Fix rpc bind address from config

benthecarman:
Missed on #2409 

The rpc bind was not being read from the config, just if it was set when doing `--rpcbind`


Christewart (commented):
If I understand this correctly we need to apply this to the oracle server as well (smart thing to do here would be just making a common trait that can be shared across projects) 

https://github.com/bitcoin-s/bitcoin-s/blob/166440b34b4d6a0c3df707cb6769ade7fe8111f7/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala#L95

Christewart (approved):

PR #2546, opened 2021-01-21
Add walletinfo rpc

### LINKED ISSUE #2545: add `getrootxpub` rpc command
This is useful for debugging multi wallet situations where you want to confirm that your wallet is using a specific key manager.

benthecarman:
Closes #2545

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
@@ -56,18 +56,11 @@ case class WalletRoutes(wallet: AnyHDWalletApi)(implicit
 
     case ServerCommand("walletinfo", _) =>
       complete {
-        wallet.getDefaultAccount().map { accountDb =>
-          val json = Obj(
-            "keyManager" -> Obj(
-              "rootXpub" -> Str(wallet.keyManager.getRootXPub.toString)
-            ),
-            "xpub" -> Str(accountDb.xpub.toString),
-            "hdPath" -> Str(accountDb.hdAccount.toString)
-          )
+        val jsonF = getInfo
+        jsonF.map { json =>
           Server.httpSuccess(json)
         }
       }
-
     case ServerCommand("getbalance", arr) =>
       GetBalance.fromJsArr(arr) match {
         case Failure(exception) =>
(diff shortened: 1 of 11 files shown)


app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -54,6 +54,18 @@ case class WalletRoutes(wallet: AnyHDWalletApi)(implicit
         }
       }
 
+    case ServerCommand("walletinfo", _) =>
```
Christewart, line 57: so to nit pick a bit. The `rootXPub` is associated with the key manager. Since [wallet does depend on key manager](https://github.com/bitcoin-s/bitcoin-s/blob/b0b56dd5da2f7e3c2428008d9a3c9f7391cf53e1/build.sbt#L538) the structure should look something like this.

```
{
  "wallet" : { 
      "keyManager" : { 
           "rootXPub" : xpub
       }
       "walletXpub" : wallet_xpub
       "hdPath" : hd_path 
   }
}
```

and then we have define `keymanagerinfo` which just returns the json object defined by the `keyManager` key. 

That way we get nice composability of our `info`s if we add them for other modules to -- and the configurations represent the dependency graph for the bitcoin-s proejct (`wallet.dependsOn(keyManager)` in this case). 

If we were to add a new field to `keymanagerinfo` later on -- like `isHardwareWallet : true`, it would magically added to the `wallet` configuration as well so our `walletinfo` would look like! 

```
{
  "wallet" : { 
      "keyManager" : { 
           "rootXPub" : xpub
            "isHardwareWallet" : true
       }
       "walletXpub" : wallet_xpub
       "hdPath" : hd_path 
   }
}
```

Which is really nice IMO and we don't have to  change `walletinfo`

benthecarman: This seems a bit redundant, technically all this info is about the key manager. I think to the user it makes more sense for them to not need to care about the distiction from the key manager and the wallet in our implementation.

Christewart: `hdPath` and `wallet_xpub` are not stored or known by the key manager.

If you look at the code, you see 

```scala
wallet.keyManager.getRootXPub
accountDb.xpub
accountDb.hdAccount
```

the latter two pieces of information are stored by the wallet.

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -54,6 +54,20 @@ case class WalletRoutes(wallet: AnyHDWalletApi)(implicit
...
+      complete {
+        wallet.getDefaultAccount().map { accountDb =>
+          val json = Obj(
+            "keyManager" -> Obj(
```
Christewart: so here is my commit that you can apply to your branch to do some things I would like that help my OCD. If you look carefully here we are using a different configuration key `keyManager` rather than `keymanager` here. 

We should really try to use the same keys for everything other wise it will drive us slowly insane 

https://github.com/Christewart/bitcoin-s-core/commit/7674a3c44ac3958ac4b52f6d574589ab02ca5ffe

Eventually it would be nice to have `getInfo` for every project and then we can nicely compose them for a `getinfo` in the `appServer` project that has information about all of our sub modules.

Christewart (approved):

PR #2384, opened 2020-12-17
v0.5.0 release notes

benthecarman:

PR #2548, opened 2021-01-21
Default to native segwit wallet

benthecarman:


Christewart (changes requested):
Can we keep our example configuration up to date with what our _default_ configuration is ? 

This needs to be updated 

https://bitcoin-s.org/docs/next/config/configuration

also we need to note this on release notes #2384

Christewart (approved):

Issue #2455, opened 2021-01-01
DBConfigTest Unable to obtain connection from database


Christewart:
Seen this on #2436 , I don't think this should be failing as DBConfigTest is unrelated to this PR

https://github.com/bitcoin-s/bitcoin-s/pull/2436/checks?check_run_id=1633902492#step:5:416

PR #2560, opened 2021-01-25
2021 01 25 type maximalist

Christewart:
Add some types for things inside of the CETCalculator like `CETOutcome` `MultiOracleDigits`, `MultiOracleOutcome`.

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPlotUtil.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPlotUtil.scala
@@ -2,6 +2,7 @@ package org.bitcoins.gui.dlc
 
 import breeze.plot.{plot, Figure}
 import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.protocol.dlc.CETCalculator.CETOutcome
 import org.bitcoins.core.protocol.dlc.{
   CETCalculator,
   DLCPayoutCurve,
@@ -96,24 +97,21 @@ object DLCPlotUtil {
       totalCollateral: Satoshis,
       rounding: RoundingIntervals,
       executedCETOpt: Option[Vector[Int]]): Figure = {
-    val cets = CETCalculator.computeCETs(base,
-                                         numDigits,
-                                         function,
-                                         totalCollateral,
-                                         rounding)
+    val cets: Vector[CETOutcome] = CETCalculator.computeCETs(base,
+                                                             numDigits,
+                                                             function,
+                                                             totalCollateral,
+                                                             rounding)
 
     plotCETs(cets, base, numDigits, executedCETOpt)
   }
 
-  def plotCETs(
-      cets: Vector[(Vector[Int], Satoshis)],
-      base: Int,
-      numDigits: Int): Figure = {
+  def plotCETs(cets: Vector[CETOutcome], base: Int, numDigits: Int): Figure = {
     plotCETs(cets, base, numDigits, executedCETOpt = None)
   }
 
   def plotCETs(
-      cets: Vector[(Vector[Int], Satoshis)],
+      cets: Vector[CETOutcome],
       base: Int,
       numDigits: Int,
       executedCET: Vector[Int]): Figure = {
@@ -121,7 +119,7 @@ object DLCPlotUtil {
   }
 
   private def plotCETs(
-      cets: Vector[(Vector[Int], Satoshis)],
+      cets: Vector[CETOutcome], //(Vector[Int], Satoshis)
       base: Int,
       numDigits: Int,
       executedCETOpt: Option[Vector[Int]]): Figure = {
@@ -129,15 +127,15 @@ object DLCPlotUtil {
       NumberUtil.fromDigits(digits, base, numDigits).toInt
     }
 
-    val xs = cets.map(_._1).map(fromDigits)
-    val ys = cets.map(_._2.toLong.toInt)
+    val xs = cets.map(_.digits).map(fromDigits)
+    val ys = cets.map(_.payout.toLong.toInt)
 
     val figure = Figure("DLC Payout Curve")
     val cetPlot = figure.subplot(0)
 
     val canonicalCETOpt = executedCETOpt
       .flatMap { outcome =>
-        CETCalculator.searchForPrefix(outcome, cets.map(_._1))(identity)
+        CETCalculator.searchForPrefix(outcome, cets.map(_.digits))(identity)
       }
       .map(fromDigits)
     val markedCETNumOpt = canonicalCETOpt.map(xs.indexOf)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
```diff
@@ -369,23 +392,34 @@ object CETCalculator {
...
 
     ranges.flatMap { range =>
       range match {
         case StartZero(indexFrom, indexTo) =>
-          groupByIgnoringDigits(indexFrom, indexTo, base, numDigits).map(
-            _ -> Satoshis.zero)
+          groupByIgnoringDigits(indexFrom, indexTo, base, numDigits).map {
+            decomp =>
+              val payout = Satoshis.zero
+              CETOutcome(decomp, payout)
```
nkohen: nit: On all of these I prefer the one liner `CETOutcome(decomp, payout = Satoshis.zero)`

core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
```diff
@@ -13,6 +13,30 @@ import scala.annotation.tailrec
...
+    *
+    * or if we are using base 16 (hex)
+    * Vector(9,8,10,11,2,3,2,4,1,15)
+    */
+  type Digits = Vector[Int]
+
+  /** A Vector of Digits corresponding to each oracle in order */
+  case class MultiOracleDigits(vec: Vector[Digits]) {
+    def toVector: Vector[Digits] = vec
+  }
```
nkohen: I think I prefer this as a type alias after having looked at the code but everything else looks good imo

nkohen (commented):
`MulitOracleDigits` is my only constraint

core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
Christewart: If you are indifferent I would really like to keep this. I find it frustrating to work with type aliases in part due to the scala compiler errors. With type aliases on type aliases (`Digits` inside of `MultiOracleDigits`) you can get error messages that look like this 

```scala
[error]  required: org.bitcoins.core.protocol.dlc.CETCalculator.MultiOracleDigits
[error]     (which expands to)  scala.collection.immutable.Vector[scala.collection.immutable.Vector[Int]]
[error]         builder.++=(Vector(1))
```

as compared to what I have on c15cbde5478c1b14fa932353ea886d8e6bb6cc64

```scala
[error] /home/chris/dev/bitcoin-s/core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala:718:28: type mismatch;
[error]  found   : scala.collection.immutable.Vector[Int]
[error]  required: org.bitcoins.core.protocol.dlc.CETCalculator.MultiOracleDigits
[error]           builder.+=(Vector(1))
```

I find it hard to reason about type aliases inside of type aliases, but maybe this is just a personal preference. If you really want me to switch it back I will.

Christewart:
I am seeing some failures on CI unrelated to this PR (#2455 will get included on `adaptor-dlc` on the next rebase).

core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
nkohen: Hm, between those two error messages I find the first much more informative? It is literally the second one but with even more extra information?

Christewart: Yes, that's the problem. I find it to be a wall of text that includes unnecessary information if I'm just trying to get things to compile (the type I care about is `MultiOracleDigits`, not trying to grok that the implementation of `MultiOracleDigits` is a `Vector[Vector[Int]]`.

This case isn't really that bad, but if you go hog wild with type aliases like the scala slick project or previous iterations of suredbits-api, I find type aliases get expanded into walls of text by the scala compiler that are really hard to understand what is going on and maintain. Maybe scala just has a poor implementation of them, but it's the tool we are using

Any way, this case seems to be confined and we aren't going to agree so I will change the code for the sake of productivity.

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPlotUtil.scala
```diff
@@ -96,48 +97,45 @@ object DLCPlotUtil {
...
+      cets: Vector[CETOutcome],
       base: Int,
       numDigits: Int,
       executedCET: Vector[Int]): Figure = {
     plotCETs(cets, base, numDigits, executedCETOpt = Some(executedCET))
   }
 
   private def plotCETs(
-      cets: Vector[(Vector[Int], Satoshis)],
+      cets: Vector[CETOutcome], //(Vector[Int], Satoshis)
```
nkohen: ping

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
```diff
@@ -334,11 +335,11 @@ class CETCalculatorTest extends BitcoinSUnitTest {
...
 
-    assert(coveringCETsMax.forall(_.length == 2))
-    assert(coveringCETsMin.forall(_.length == 2))
+    assert(coveringCETsMax.forall(_.toVector.length == 2))
+    assert(coveringCETsMin.forall(_.toVector.length == 2))
 
-    (coveringCETsMax.map(ds => (ds.head, ds.last)),
-     coveringCETsMin.map(ds => (ds.head, ds.last)))
+    (coveringCETsMax.map(ds => (ds.toVector.head, ds.toVector.last)),
+     coveringCETsMin.map(ds => (ds.toVector.head, ds.toVector.last)))
```
nkohen: I think these were missed in reverting?

core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
```diff
@@ -669,10 +707,13 @@ object CETCalculator {
           numToVec(start - minFail, numDigits, minFailExp)
         }
 
-        builder.++=(
+        val doubleCovering: Vector[Vector[Digits]] =
```
nkohen: nit: `Vector[MultiOracleDigits]`

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
```diff
@@ -334,8 +335,8 @@ class CETCalculatorTest extends BitcoinSUnitTest {
                                               maximizeCoverage = false,
                                               numOracles = 2)
 
-    assert(coveringCETsMax.forall(_.length == 2))
-    assert(coveringCETsMin.forall(_.length == 2))
+    assert(coveringCETsMax.forall(_.toVector.length == 2))
+    assert(coveringCETsMin.forall(_.toVector.length == 2))
```
nkohen: I don't think these two calls are needed

nkohen (approved):
ACK

Issue #2564, opened 2021-01-26
Add state descriptor to wallet info rpc


Christewart:
This would be extremely useful to have on this endpoint. Basically include the block height and block hash along with all of our other information. I was going to use this rpc as an example for #2486

```
 ./bitcoin-s-cli walletinfo
{
  "wallet": {
    "keymanager": {
      "rootXpub": ...
    },
    "xpub": ...,
    "hdPath": "m/84'/0'/0'",
    "height" : 0,
    "blockHash" : "1234....567890"
  }
}
```

PR #2565, opened 2021-01-26
Add wallet sync documentation

### LINKED ISSUE #2486: Add WalletSync documentation similar to ChainSync
In #2461 i added `WalletSync`. 

We have a nice example of how to use `ChainSync` to sync your chainstate with an arbitrary data source, we should document the same with the wallet. 

https://bitcoin-s.org/docs/next/chain/filter-sync#syncing-block-filters-against-bitcoind

Christewart:
fixes #2486 

This is kind of blocked on #2564 as I need to update the example rpc call to include blockhash/block height information.

### DIFF
--- a/docs/wallet/wallet-sync.md
+++ b/docs/wallet/wallet-sync.md
@@ -1,0 +1,128 @@
+---
+title: Wallet Sync
+id: wallet-sync
+---
+
+## High level wallet state
+
+Our wallet infrastructure has a specific table called `state_descriptors`.
+This tracks chain state for our wallet.
+Here is an example of the contents of this table
+
+>sqlite> select * from state_descriptors;
+SyncHeight|0000000000000000000134aa9e949ea1d053042b8dfa59bdc73b0322a88f009e 665741
+
+If you look carefully in the second column, you will see a string encoding indicating
+what the wallet state is. In this case, the last block hash seen by the wallet is
+
+>0000000000000000000134aa9e949ea1d053042b8dfa59bdc73b0322a88f009e
+
+and height
+
+>665741
+
+## Syncing a wallet
+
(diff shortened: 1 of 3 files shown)


docs/wallet/wallet-sync.md
```diff
@@ -0,0 +1,128 @@
...
+{
+  "wallet": {
+    "keymanager": {
+      "rootXpub": "..."
+    },
+    "xpub": "...",
+    "hdPath": "m/84'/0'/0'"
+  }
+}
+```
```
benthecarman: imo this doesn't really make sense to include here. If you are using bitcoin-s as a library then you won't have access to the cli/server

docs/wallet/wallet-sync.md
```diff
@@ -0,0 +1,128 @@
...
+---
+
+## High level wallet state
+
+Our wallet infrastructure has a specific table called `state_descriptors`.
+This tracks chain state for our wallet.
+Here is an example of the contents of this table
+
+>sqlite> select * from state_descriptors;
+SyncHeight|0000000000000000000134aa9e949ea1d053042b8dfa59bdc73b0322a88f009e 665741
```
benthecarman, line 13: Instead of this, you can use `wallet.getSyncDescriptorOpt`

benthecarman (changes requested):

docs/wallet/wallet-sync.md
Christewart: I really liked this section on the rescan docs, i was trying to replicate it here 

https://bitcoin-s.org/docs/next/wallet/wallet-rescan#rescan-from-cli

maybe I should have an appropriate heading like we do there? 

>Wallet state from the cli

or something of that nature?

benthecarman: Yeah that would make more sense

benthecarman (approved):

PR #2567, opened 2021-01-26
Add sync state to walletinfo

### LINKED ISSUE #2564: Add state descriptor to wallet info rpc

benthecarman:
Fixes #2564

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
@@ -396,6 +396,8 @@ trait WalletApi extends StartStopAsync[WalletApi] {
     * is a change output from this wallet
     */
   def isChange(output: TransactionOutput): Future[Boolean]
+
+  def getSyncState(): Future[BlockSyncState]
 }
 
 /** An HDWallet that uses Neutrino to sync */
--- a/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
@@ -562,7 +562,10 @@ case class WalletRoutes(wallet: AnyHDWalletApi)(implicit
 
   /** Returns information about the state of our wallet */
   def getInfo: Future[Obj] = {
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
```diff
@@ -396,6 +396,8 @@ trait WalletApi extends StartStopAsync[WalletApi] {
     * is a change output from this wallet
     */
   def isChange(output: TransactionOutput): Future[Boolean]
+
+  def getSyncState(): Future[(Int, DoubleSha256DigestBE)]
```
Christewart: So this LGTM, but one thing that might be nice is to rename our current `FilterSyncMarker` to just `SyncMarker` and then use that here too? 

https://github.com/bitcoin-s/bitcoin-s/blob/2de17bb4e47961fd3fe4e55d75909b42470e3c59/core/src/main/scala/org/bitcoins/core/api/chain/FilterSyncMarker.scala#L11

Christewart: It looks like we've had this discussion before, are the contexts the same? https://github.com/bitcoin-s/bitcoin-s/pull/2003#issuecomment-691176472

benthecarman: I don't think it makes sense to use `FilterSyncMarker` or renaming it. `FilterSyncMarker`'s height and hash aren't meant to refer to the same block

Christewart: ok, then possibly make a `BlockSyncMarker` with appropriate field names?

Christewart (approved):

PR #2571, opened 2021-01-27
Bech32 address improvements

benthecarman:
Fixes so we can read uppercase addresses, adds some decoding optimizations, and makes sure we verify the checksum of the address.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/Address.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/Address.scala
@@ -105,6 +105,10 @@ sealed abstract class Bech32Address extends BitcoinAddress {
   def expandHrp: Vector[UInt5] = {
     hrp.expand
   }
+
+  def verifyChecksum: Boolean = {
+    Bech32.verifyChecksum(hrp.expand, data ++ checksum)
+  }
 }
 
 object Bech32Address extends AddressFactory[Bech32Address] {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -113,7 +117,7 @@ object Bech32Address extends AddressFactory[Bech32Address] {
       networkParameters: NetworkParameters,
       data: Vector[UInt5])
       extends Bech32Address {
-    //require(verifyChecksum(hrp, data), "checksum did not pass")
+    require(verifyChecksum, "checksum did not pass")
```
Christewart: nit: Put them in the error messages

Christewart: ooh wow, this isn't applied to _any_ of our address types. We probably want to add this in the super class possibly? 

I'm going to file an issue

https://github.com/bitcoin-s/bitcoin-s/blob/223be8057351017eb0d454b97b36a917f00fe1ac/core/src/main/scala/org/bitcoins/core/protocol/Address.scala#L215

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -305,6 +305,17 @@ object Bech32 extends Bech32 {
   val charset: Vector[Char] = Vector('q', 'p', 'z', 'r', 'y', '9', 'x', '8',
     'g', 'f', '2', 't', 'v', 'd', 'w', '0', 's', '3', 'j', 'n', '5', '4', 'k',
     'h', 'c', 'e', '6', 'm', 'u', 'a', '7', 'l')
+
+  /** The Bech32 character set for decoding. */
```
Christewart: is there a reference to this? Is it reversed or the inverse?

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
benthecarman: Added link from where I got it from

Christewart (approved):

PR #2572, opened 2021-01-27
Implement bech32m

benthecarman:
Implements bech32 in accordance to https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/protocol/Bech32mSpec.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/protocol/Bech32mSpec.scala
@@ -1,42 +1,36 @@
 package org.bitcoins.core.protocol
 
+import org.bitcoins.core.protocol.script.WitnessVersion0
 import org.bitcoins.core.util.{Bech32, Bech32Encoding}
 import org.bitcoins.testkitcore.gen._
-import org.bitcoins.testkitcore.gen.ln.LnInvoiceGen
 import org.scalacheck.{Prop, Properties}
 
 import scala.annotation.tailrec
 import scala.util.{Random, Success}
 
-class Bech32Spec extends Properties("Bech32Spec") {
-  property("split all LN invoices into HRP and data") = {
-    Prop.forAll(LnInvoiceGen.lnInvoice) { invoice =>
-      val splitT =
-        Bech32.splitToHrpAndData(invoice.toString, Bech32Encoding.Bech32)
-      splitT.isSuccess
-    }
-  }
+class Bech32mSpec extends Properties("Bech32mSpec") {
 
-  property("split all Bech32 addresses into HRP and data") = {
-    Prop.forAll(AddressGenerator.bech32Address) { address =>
+  property("split all Bech32m addresses into HRP and data") = {
+    Prop.forAll(AddressGenerator.bech32mAddress) { address =>
       val splitT =
-        Bech32.splitToHrpAndData(address.value, Bech32Encoding.Bech32)
+        Bech32.splitToHrpAndData(address.value, Bech32Encoding.Bech32m)
       splitT.isSuccess
     }
   }
 
   property("serialization symmetry") = {
-    Prop.forAll(ScriptGenerators.witnessScriptPubKeyV0,
+    Prop.forAll(ScriptGenerators.witnessScriptPubKey.suchThat(
+                  _._1.witnessVersion != WitnessVersion0),
                 ChainParamsGenerator.networkParams) {
       case ((witSPK, _), network) =>
-        val addr = Bech32Address(witSPK, network)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/bitcoins/core/protocol/Address.scala
```diff
@@ -210,6 +211,142 @@ object Bech32Address extends AddressFactory[Bech32Address] {
...
+  override def value: String = {
+    val all: Vector[UInt5] = data ++ checksum
+    val encoding = Bech32.encode5bitToString(all)
+
+    hrp.toString + Bech32.separator + encoding
+  }
+
+  def checksum: Vector[UInt5] = Bech32mAddress.createChecksum(hrp, data)
+
+  override def scriptPubKey: WitnessScriptPubKey = {
```
Christewart, line 231: Doesn't this have to be a `WitnessScriptPubKeyV1`? 

I guess we don't have that now, but my understanding is v0 will always use the old bech32 encoding?

Christewart: ah i see below, we should probably add a TODO or file an issue

core/src/main/scala/org/bitcoins/core/util/Bech32.scala
```diff
@@ -7,6 +7,21 @@ import scodec.bits.ByteVector
 import scala.annotation.tailrec
 import scala.util.{Failure, Success, Try}
 
+sealed abstract class Bech32Encoding {
```
Christewart, line 14: add a link to what/where these differences are specified. I'm assuming they are on in the BIP?

core-test/src/test/scala/org/bitcoins/core/protocol/Bech32mSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import org.bitcoins.core.protocol.script.WitnessVersion0
+import org.bitcoins.core.util.{Bech32, Bech32Encoding}
+import org.bitcoins.testkit.core.gen._
+import org.scalacheck.{Prop, Properties}
+
+import scala.annotation.tailrec
+import scala.util.{Random, Success}
+
+class Bech32mSpec extends Properties("Bech32mSpec") {
```
Christewart, line 11: nit: This is the old style of test cases (uses scalacheck, not scalatest). We are trying to slowly move away from these over the long term. Can you write these using `BitcoinSUnitTest` ? Also convert the `Bech32Spec` file if you are feeling ambitious :-)

Christewart: https://github.com/bitcoin-s/bitcoin-s/blob/c5be81d5e21cd22a4e69dc86dcf47fbaa7d623f3/core-test/src/test/scala/org/bitcoins/core/protocol/AddressTest.scala#L12

Christewart: I guess this means this file can be deleted and test cases migrated to `Bech32mTest`

Christewart (changes requested):
So what are your thoughts on merging this? Should we wait until we have `WitnessScriptPubKeyV1` available to us ?

If i understand this, that would mean doing Tapscript/Taproot integration?

Christewart (approved):
So this LGTM, 

but is BIP350 actually finalized? Is it deployed as part of bitcoin core 0.21.0? It seems to me this shouldn't be merged until the spec is finalized ? 

https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki

benthecarman:
> So this LGTM,
> 
> but is BIP350 actually finalized? Is it deployed as part of bitcoin core 0.21.0? It seems to me this shouldn't be merged until the spec is finalized ?
> 
> [bitcoin/bips@`master`/bip-0350.mediawiki](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)

The spec should be finalized, it is not merged into core yet: https://github.com/bitcoin/bitcoin/pull/20861

imo this should be fine to merge though

Christewart:
> > So this LGTM,
> > but is BIP350 actually finalized? Is it deployed as part of bitcoin core 0.21.0? It seems to me this shouldn't be merged until the spec is finalized ?
> > [bitcoin/bips@`master`/bip-0350.mediawiki](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)
> 
> The spec should be finalized, it is not merged into core yet: [bitcoin/bitcoin#20861](https://github.com/bitcoin/bitcoin/pull/20861)
> 
> imo this should be fine to merge though

I vote to wait until it gets merged into core as there is less likely to be changes after that occurs

PR #2584, opened 2021-01-29
Add ContractOraclePair to enforce an invariant that contract descript…

Christewart:
…ors and oracle infos are of the same type

This simply adds a new ADT called `ContractOraclePair` that makes sure that the contract descriptor and oracle info are of the same type.

I then propogated this throughout the rest of the codebase to try and get rid of some of the `Either` usage. 

It seems like this coule be expanded in the future to many subtypes of `EnumPair` and `NumericPair` (is it a single oracle info, multi oracle info?) but I decided to just do the simple thing to start.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorOraclePair.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorOraclePair.scala
@@ -1,0 +1,51 @@
+package org.bitcoins.core.protocol.dlc
+
+sealed trait ContractOraclePair {
+  def contractDescriptor: ContractDescriptor
+  def oracleInfo: OracleInfo
+}
+
+object ContractOraclePair {
+
+  case class EnumPair(
+      contractDescriptor: EnumContractDescriptor,
+      oracleInfo: EnumOracleInfo)
+      extends ContractOraclePair
+
+  case class NumericPair(
+      contractDescriptor: NumericContractDescriptor,
+      oracleInfo: NumericOracleInfo)
+      extends ContractOraclePair
+
+  /** Returns a valid [[ContractOraclePair]] if the
+    * [[ContractDescriptor]] and [[OracleInfo]] are of the same type
+    */
+  def fromDescriptorOracleOpt(
+      descriptor: ContractDescriptor,
+      oracleInfo: OracleInfo): Option[ContractOraclePair] = {
+    (descriptor, oracleInfo) match {
+      case (e: EnumContractDescriptor, o: EnumOracleInfo) =>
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorOraclePair.scala
```diff
@@ -0,0 +1,51 @@
...
+      oracleInfo: OracleInfo): Option[ContractOraclePair] = {
+    (descriptor, oracleInfo) match {
+      case (e: EnumContractDescriptor, o: EnumOracleInfo) =>
+        Some(EnumPair(e, o))
+      case (n: NumericContractDescriptor, o: NumericOracleInfo) =>
+        Some(NumericPair(n, o))
+      case (_: EnumContractDescriptor, _: NumericOracleInfo) =>
+        None
+      case (_: NumericContractDescriptor, _: EnumOracleInfo) =>
+        None
```
benthecarman, line 39: nit: can put these together with a `|`

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTLVGen.scala
```diff
@@ -48,14 +48,29 @@ object DLCTLVGen {
...
       totalInput: CurrencyUnit = defaultAmt * 2): ContractInfo = {
-    ContractInfo(totalInput.satoshis,
-                 genContractDescriptor(outcomes, totalInput),
-                 genOracleInfo(oraclePrivKey, oracleRValue, outcomes))
+    val pair = genEnumContractOraclePair(oraclePrivKey,
+                                         oracleRValue,
+                                         outcomes,
+                                         totalInput)
+
+    //this doesn't ever try numeric contracts?
```
benthecarman, line 72: this is because we are passing in enum outcomes ` outcomes: Vector[String] `

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorOraclePair.scala
```diff
@@ -0,0 +1,51 @@
+package org.bitcoins.core.protocol.dlc
+
+sealed trait ContractOraclePair {
```
nkohen, line 8: scaladoc

dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTLVGen.scala
nkohen: all test vector infra is currently only for single oracle enum pending a re-write

nkohen (approved):

PR #2568, opened 2021-01-26
Optimize filter sync and fetching filter heights

benthecarman:
Closes #2417
Closes #2504 

Removes some unneed db calls in filter syncing and should make fetching the filter height faster. Previously we would get all of the filter's data this changes it to just get the height.

Ran some IBD tests and was able to download all filters in under 15 minutes (from local node)

PR #2585, opened 2021-01-29
Add better Block.toString that doesn't blow up logs

Christewart:
When testing #2568 i ran into errors that ended up outputting a block to our logs. 

The current `Block.toString` logs _EVERY_ bitcoin transaction in the block which can be a ton of data. This adjustment just logs the header and the tx count.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/blockchain/Block.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/blockchain/Block.scala
@@ -51,11 +51,16 @@ sealed abstract class Block extends NetworkElement {
   */
 object Block extends Factory[Block] {
 
-  sealed private case class BlockImpl(
+  private case class BlockImpl(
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/bitcoins/core/protocol/blockchain/Block.scala
```diff
@@ -51,11 +51,16 @@ sealed abstract class Block extends NetworkElement {
...
-  sealed private case class BlockImpl(
+  private case class BlockImpl(
       blockHeader: BlockHeader,
       txCount: CompactSizeUInt,
       transactions: Seq[Transaction])
-      extends Block
+      extends Block {
+
+    override def toString: String = {
+      s"Block(blockHeader=${blockHeader}, txCount=${txCount})"
```
benthecarman: nit: use `txCount.toLong`

benthecarman (approved):

PR #2593, opened 2021-01-29
Add check to see if we are syncing before attempting to process a Blo…

### LINKED ISSUE #2590: BlockMessage during IBD causes errors to be thrown
This is basically the same as #2586, but it does appear that #2587 does fix the problem where IBD stops. This was observed on #2568 on commit  241cf2de06b2aeed5951b7eba726451516a8b9bf while doing IBD on sqlite3

>2021-01-29T15:53:57UTC INFO [DataMessageHandler] Received block message with hash 000000006196b57f795c52fb4fca7be14112fec87a8c131f206108278f392d50
2021-01-29T15:53:57UTC WARN [Blockchain$] No common ancestor found in the chain with tip=000000000495fa1702b9ecd1f323da07383b0e99815029c94fda366a76014cbc to connect to hash=DoubleSha256BDigestBE(000000006196b57f795c52fb4fca7be14112fec87a8c131f206108278f392d50) prevHash=DoubleSha256BDigestBE(000000001f9d38de4129622b2a497e6f61725ff0d970cf0655187d070c34b44e)
2021-01-29T15:53:57UTC ERROR [DataMessageHandler] Failed to handle data payload=BlockMessage(BlockImpl(BlockHeader(hashBE=000000006196b57f795c52fb4fca7be14112fec87a8c131f206108278f392d50,version=Int32Impl(536870912),prevBlockHashBE=000000001f9d38de4129622b2a497e6f61725ff0d970cf0655187d070c34b44e,merkleRootHashBE=79a4aff9395b64695e5ea62e3958ea313f790e5c7499d8c6c795d9da7369745f,time=UInt32Impl(1611935628),nBits=UInt32Impl(486604799),nonce=UInt32Impl(505666722)),CompactSizeUInt(959),Vector(WitnessTransaction(Int32Impl(1),Vector(CoinbaseInputImpl(NonStandardScriptSignature(Vector(BytesToPushOntoStackImpl(3), ScriptConstantImpl(ByteVector(3 bytes, 0xd1171d)), BytesToPushOntoStackImpl(4), ScriptConstantImpl(ByteVector(4 bytes, 0x8c2f1460)), BytesToPushOntoStackImpl(8), ScriptConstantImpl(ByteVector(8 bytes, 0xf80000015b0c0000)), BytesToPushOntoStackImpl(12), ScriptConstantImpl(ByteVector(12 bytes, ...

Christewart:
…ckMessage with the chain api

fixes #2590 

Basically if we are syncing, we should do nothing with a block message. It seems #2587 did this at a lower level. I'm unsure if i should revert that because this covers it.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -254,34 +254,40 @@ case class DataMessageHandler(
         }
       case msg: BlockMessage =>
         val block = msg.block
-        logger.info(
-          s"Received block message with hash ${block.blockHeader.hash.flip.hex}")
 
-        val newApiF = {
-          chainApi
-            .getHeader(block.blockHeader.hashBE)
-            .flatMap { headerOpt =>
-              if (headerOpt.isEmpty) {
-                logger.debug("Processing block's header...")
-                for {
-                  processedApi <- chainApi.processHeader(block.blockHeader)
-                  _ <-
-                    appConfig.nodeCallbacks
-                      .executeOnBlockHeadersReceivedCallbacks(
-                        logger,
-                        Vector(block.blockHeader))
-                } yield processedApi
-              } else Future.successful(chainApi)
-            }
-        }
+        if (syncing) {
+          logger.info(
+            s"Ignoring block message since we are in IBD, hash=${block.blockHeader.hashBE.hex}")
+          Future.successful(this)
+        } else {
+          logger.info(
+            s"Received block message with hash ${block.blockHeader.hash.flip.hex}")
+          val newApiF = {
+            chainApi
+              .getHeader(block.blockHeader.hashBE)
+              .flatMap { headerOpt =>
+                if (headerOpt.isEmpty) {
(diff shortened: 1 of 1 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -254,34 +254,40 @@ case class DataMessageHandler(
...
-                  _ <-
-                    appConfig.nodeCallbacks
-                      .executeOnBlockHeadersReceivedCallbacks(
-                        logger,
-                        Vector(block.blockHeader))
-                } yield processedApi
-              } else Future.successful(chainApi)
-            }
-        }
+        if (syncing) {
```
Christewart, line 258: This might need to be thought about a bit more, what happens in this case? 

1. Block headers are done syncing
2. Filter headers are done syncing
3. We are syncing filters, and one of those filters happens to match in the wallet, and we request the full block. Does this mean we ignore this block?

benthecarman:
If we do this then we'd have to rescan after doing IBD. Otherwise we can't processes blocks while syncing

PR #2599, opened 2021-02-01
Rename PayoutCurveComponent -> PayoutCurveInterval, rename CETRange t…

Christewart:
Renames some data structures in the dlc protocol to be more descriptive, there is still one I would like to find a better name for -- `StartTotal`. I don't understand how this case is possible in the first place though. 
https://github.com/bitcoin-s/bitcoin-s/blob/ed86d8b336d0344bbdffd11aba4f8f73bfcc1a58/core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala#L57

I don't understand how we can have more than our totalCollateral for payout in a DLC.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
@@ -51,18 +51,20 @@ object CETCalculator {
     * (Note that interpolated functions are allowed
     * to be negative, but we set all negative values to 0).
     */
-  case class StartZero(indexFrom: Long, indexTo: Long) extends CETRange
+  case class RemotePayoutRange(indexFrom: Long, indexTo: Long) extends CETRange
 
   /** This range contains payouts all >= totalCollateral */
   case class StartTotal(indexFrom: Long, indexTo: Long) extends CETRange
 
   /** This range contains payouts that all vary at every step and cannot be compressed */
-  case class StartFunc(indexFrom: Long, indexTo: Long) extends CETRange
+  case class VariablePayoutRange(indexFrom: Long, indexTo: Long)
+      extends CETRange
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
```diff
@@ -51,18 +51,20 @@ object CETCalculator {
     * (Note that interpolated functions are allowed
     * to be negative, but we set all negative values to 0).
     */
-  case class StartZero(indexFrom: Long, indexTo: Long) extends CETRange
+  case class RemotePayoutRange(indexFrom: Long, indexTo: Long) extends CETRange
```
nkohen: Local/Remote language should be reserved for values which differ between the two parties where there is a clear "local" actor. `RemotePayoutRange` is very misleading because all payouts from this file are from the local point of view. I'm not attached to `StartZero` and `StartFuncConst` but I find the use of `Remote` and `Local` in these two new names much more confusing than those.

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCPayoutCurve.scala
```diff
@@ -162,7 +162,7 @@ object OutcomePayoutMidpoint {
 }
 
 /** A single piece of a larger piecewise function defined between left and right endpoints */
-sealed trait DLCPayoutCurveComponent {
+sealed trait DLCPayoutCurveInterval {
```
nkohen: upon looking at this again, I don't personally find this name very descriptive. Sure there is an interval which corresponds to each of these but not every interval does, this very specifically corresponds to a single "piece" of the `DLCPayoutCurve` which is a piece-wise polynomial. Thoughts on `DLCPayoutCurvePiece`?

nkohen (approved):

Issue #2529, opened 2021-01-18
Revamp DLC documentation


Christewart:
The page that i'm particularly concerned about is this one 

https://bitcoin-s.org/docs/next/wallet/dlc

We should also take a peak at the oracle server docs, but i'm guessing these are much more up to date 

https://bitcoin-s.org/docs/next/oracle/oracle-server

Christewart:
Taken care of in #2551

PR #2601, opened 2021-02-01
Add enum oracle example

Christewart:
Part of #2529 

Basically just as an explicit example of creating an enum event for an oracle.

### DIFF
--- a/docs/oracle/oracle-election-example.md
+++ b/docs/oracle/oracle-election-example.md
@@ -18,9 +18,9 @@ After building the oracle server, you will need to start it with
 
 ## US 2020 Election
 
-In 2020, the United States held a presidential election. Nicolas Dorier and Chris Stewart
-wanted to do a DLC based on the outcome of this presidential election.
-In this example, we are going to walk through
+In 2020, the United States held a presidential election.
+People want to do a DLC based on the outcome, so you decide to be
+their oracle and attest to the winner of the election.
 
 1. Setting up an oracle that can attest to the outcome of the election
 2. Complete the event by attesting to the winner of the election
--- a/docs/oracle/oracle-election-example.md
+++ b/docs/oracle/oracle-election-example.md
@@ -1,0 +1,111 @@
+---
+id: oracle-election-example
+title: Election Example
+---
+
+## Requirements for example
+
+You need to have a fully built oracle server. You can follow [this guide](build-oracle-server.md) to do this.
+
+You will also need a the `bitcoin-s-cli` command line tool to interact with the server.
+You can find how to build this [here](../applications/cli.md)
+
+After building the oracle server, you will need to start it with
+
+```
+./app/oracle-server/target/universal/stage/bin/bitcoin-s-oracle-server
+```
+
+## US 2020 Election
+
+In 2020, the United States held a presidential election. Nicolas Dorier and Chris Stewart
+wanted to do a DLC based on the outcome of this presidential election.
+In this example, we are going to walk through
+
+1. Setting up an oracle that can attest to the outcome of the election
+2. Complete the event by attesting to the winner of the election
+
+### Setting up the election bet
+
+The election bet can be represented by an enumerated event. What we mean by this is the outcomes
+for this event can be simply enumerated in a list. There were three possible outcomes for the election
+
+1. "Repulican_win"
+2. "Democrat_win"
+3. "other"
+
+The next step is to create an oracle announcement that can be shared with others. This announcement
+contains cryptographic information that is used by people that want to enter into DLCs to setup
+their bitcoin transactions.
+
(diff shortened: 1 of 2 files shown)


docs/oracle/oracle-election-example.md
```diff
@@ -0,0 +1,111 @@
...
+After building the oracle server, you will need to start it with
+
+```
+./app/oracle-server/target/universal/stage/bin/bitcoin-s-oracle-server
+```
+
+## US 2020 Election
+
+In 2020, the United States held a presidential election. Nicolas Dorier and Chris Stewart
+wanted to do a DLC based on the outcome of this presidential election.
```
benthecarman: > Nicolas Dorier and Chris Stewart wanted to do a DLC based on the outcome of this presidential election.

imo this should removed since the oracle does not have knowledge of their users

docs/oracle/oracle-election-example.md
```diff
@@ -0,0 +1,111 @@
...
+fdd824c3988fabec9820690f366271c9ceac00fbec1412075f9b319bb0db1f86460519dd9c61478949f2c00c35aeb8e53a1507616072cb802891e2c189a9fa65a0493de5d3b04a6d7b90c9c43c09ebe5250d583e1c3fc423219b26f6a02ec394a130000afdd8225f0001ae3e30df5a203ad10ee89a909df0c8ccea4836e94e0a5d34c3cdab758fcaee1460189600fdd8062400030e52657075626c6963616e5f77696e0c44656d6f637261745f77696e056f7468657210323032302d75732d656c656374696f6e
+
+Yay! The hex string returned represents an oracle announcement.
+You can submit this on a tool like the [suredbits oracle explorer](https://oracle.suredbits.com)
+so others can find your oracle.
+
+It is important that you store two things
+
+1. The oracle announcement above (`fdd824c...`)
+2. The timestamp that the event matures at (`1612224000`)
```
benthecarman: Neither of these should need to be stored. The oracle will store them itself.

docs/oracle/oracle-election-example.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Signing the outcome for the election bet
+
+Now that we have create an event, we need to wait for the maturation time to pass.
+In the example above, this will be
+
+>Tuesday, February 2, 2021 12:00:00 AM
```
benthecarman: You don't need to wait for it to pass. The timestamp is mostly just used for fraud proofs

docs/oracle/oracle-election-example.md
```diff
@@ -0,0 +1,111 @@
...
+1. The label for the event
+   we will use `2020-us-election`
+2. the maturation time for the event in seconds since the epoch.
+   For our example we will use `1612224000` (Tuesday, February 2, 2021 12:00:00 AM).
+   You will need to adjust this for your local testing. You can use [this](https://www.epochconverter.com/seconds-days-since-y0) tool to calculate one
+3. The outcomes, which we listed above "Republican_win,Democrat_win,other"
+
+With all of this information, we can create the event with `bitcoin-s-cli`!
+
+>./bitcoin-s-cli createevent 2020-us-election 1612224000 "Republican_win,Democrat_win,other"
```
benthecarman: this should be done is a large code block.

docs/oracle/oracle-election-example.md
```diff
@@ -0,0 +1,111 @@
...
+fdd824c3988fabec9820690f366271c9ceac00fbec1412075f9b319bb0db1f86460519dd9c61478949f2c00c35aeb8e53a1507616072cb802891e2c189a9fa65a0493de5d3b04a6d7b90c9c43c09ebe5250d583e1c3fc423219b26f6a02ec394a130000afdd8225f0001ae3e30df5a203ad10ee89a909df0c8ccea4836e94e0a5d34c3cdab758fcaee1460189600fdd8062400030e52657075626c6963616e5f77696e0c44656d6f637261745f77696e056f7468657210323032302d75732d656c656374696f6e
+
+Yay! The hex string returned represents an oracle announcement.
```
benthecarman: nit:

> represents

I would just use `is` because it is the entire announcement

benthecarman (changes requested):

docs/oracle/oracle-election-example.md
Christewart: The purpose is to create a story so users have a frame of reference of what the purpose of the documentation is.

benthecarman: Still, the story shouldn't have to involve explict users. Could do something like

> In 2020, the United States held a presidential election. People want to do a DLC based on the outcome, so you decide to be their oracle and attest to the winner of the election.

benthecarman (approved):

PR #2607, opened 2021-02-02
Update website dependencies to address security warnings

Christewart:
This should fix our dependabot alerts for the website

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -5,32 +5,14 @@
     "previous": "Previous",
     "tagline": "Bitcoin implementation in Scala",
     "docs": {
-      "applications/chain": {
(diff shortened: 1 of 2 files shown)


benthecarman (approved):
I did this earlier and it did not get rid of the warning. Hopefully this does it

Christewart:
> I did this earlier and it did not get rid of the warning. Hopefully this does it

I think you are right this isn't going to fix it. It seems that the vulnerability is within docusaraus itself. I'm trying to figure out how/where we can update this.

Christewart:
So AFAICT we are on the latest version of docusaraus, but that version in vulnerable. I think we are going to have to wait for updates upstream

https://docusaurus.io/docs/en/installation

PR #2612, opened 2021-02-02
Fix bestblockhash rpc call

### LINKED ISSUE #2258: getbestblockhash returning stale block
**Describe the bug**
`getbestblockhash` is returning a block that is 208 blocks old! It appears that might node has fallen out of sync for some reason (sort of). 

This is occurring on the latest `v0.4.0` official release.

> ./bitcoin-s-cli getbestblockhash
00000000000000000002b1202dd118fd7d720493fba2a34b00bcefd8c790584b

This corresponds to the [block at height 656299](https://blockstream.info/block/00000000000000000002b1202dd118fd7d720493fba2a34b00bcefd8c790584b)

If I query my node for blockcount I get 

>./bitcoin-s-cli getblockcount
656508

Which is accurate as of the time of this writing. 

My database also says that I am receiving headers _even though_ my `getbestblockhash` is returning an incorrect value

Screenshot from 2020-11-11 17-00-26

I'm not sure why these are internally inconsistent.

One notable thing that occurred today was a 2 block reorg on bitcoin mainent where my node was running 

https://twitter.com/BitMEXResearch/status/1326627907805065218

However if my `getbestblockhash` is 208 blocks old (as of this writing) it appears that my issue occurred yesterday before this reorg stuff occurred. 

I've attached a log file for the past two days (11-10-2020 to 11-11-2020)

[bitcoin-s-issue.log](https://github.com/bitcoin-s/bitcoin-s/files/5527212/bitcoin-s-issue.log)

benthecarman:
Fixes #2258

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -956,6 +956,14 @@ object ChainHandler {
                      blockFilterCheckpoints)
   }
 
+  def fromChainHandlerCached(cached: ChainHandlerCached)(implicit
+      ec: ExecutionContext): ChainHandler = {
+    new ChainHandler(blockHeaderDAO = cached.blockHeaderDAO,
+                     filterHeaderDAO = cached.filterHeaderDAO,
+                     filterDAO = cached.filterDAO,
+                     blockFilterCheckpoints = Map.empty)(cached.chainConfig, ec)
+  }
+
   /** Constructs a [[ChainHandler chain handler]] from the state in the database
     * This gives us the guaranteed latest state we have in the database
     */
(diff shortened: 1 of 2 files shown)


chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -956,6 +956,14 @@ object ChainHandler {
                      blockFilterCheckpoints)
   }
 
+  def fromChainHandlerCached(cached: ChainHandlerCached)(implicit
```
Christewart, line 959: Why are we creating this from the cached chain handler? 

This seems confusing to me because the purpose of this method is _not_ to use the cached version. We are just using the underlying `DAO`s if I understand.

Wouldn't it just be easier to do `ChainHandler.fromDatabase()(implicit ec: ExecutionContext, chainConfigAppConfig)` with no parameters?

benthecarman: I guess we could, this ensures we use the same config and if we change the chain handler to have more params later on then this api should hopefully ensure we update it correctly

Christewart (approved):

PR #2621, opened 2021-02-03
Refactor split into ranges

nkohen:


benthecarman (approved):

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
```diff
@@ -33,15 +33,13 @@ class CETCalculatorTest extends BitcoinSUnitTest {
...
       VariablePayoutRange(21, 39),
       ConstantPayoutRange(40, 50),
-      VariablePayoutRange(51, 69),
-      ZeroPayoutRange(70, 70),
-      VariablePayoutRange(71, 79),
+      VariablePayoutRange(51, 79),
       ConstantPayoutRange(80, 90),
       VariablePayoutRange(91, 98),
-      MaxPayoutRange(99, 108),
+      ConstantPayoutRange(99, 108), // totalCollateral
```
Christewart, line 42: nit: Maybe we should just add a `payout: Satoshis` field to `ConstantPayoutRange()`?

core/src/main/scala/org/bitcoins/core/protocol/dlc/CETCalculator.scala
```diff
@@ -66,21 +57,10 @@ object CETCalculator {
...
-      if (value <= Satoshis.zero) {
-        ZeroPayoutRange(index, index)
-      } else if (value >= totalCollateral) {
-        MaxPayoutRange(index, index)
-      } else {
-        VariablePayoutRange(index, index)
-      }
-    }
+  /** A CETRange with a single element range */
+  case class SingletonPayoutRange(index: Long) extends CETRange {
```
Christewart, line 61: :+1:

Christewart (commented):
My code review: 

https://github.com/Christewart/bitcoin-s-core/tree/2021-02-04-breakout-processconstant

basically 

1. Move `processConstant()` out of `splitIntoRanges()`
2. Add a type for what is returned from `processConstant()`.

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
nkohen: I don't see why that would be needed? This comment isn't important or anything

nkohen:
Responded @Christewart

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/CETCalculatorTest.scala
Christewart: it's nice to know what the payout was for a specific range. Seems like minimal storage, where the alternative is having to get re-evaluate the function again, but :man_shrugging:

Christewart (approved):

PR #2625, opened 2021-02-04
Remove time restriction for creating oracle events

benthecarman:


nkohen:
Timestamps are meta-data which do not always mean the same thing.

For example if someone wants to sign a message when the bitcoin price breaks a certain threshold, they may put the creation of the event as the timestamp, or something else, but whatever they choose to do as their convention these kinds of events should not be restricted by the gui because only the very boring "price at time T" actually has a timestamp where this enforcing makes sense

Christewart:
> Timestamps are meta-data which do not always mean the same thing.
> 
> For example if someone wants to sign a message when the bitcoin price breaks a certain threshold, they may put the creation of the event as the timestamp, or something else, but whatever they choose to do as their convention these kinds of events should not be restricted by the gui because only the very boring "price at time T" actually has a timestamp where this enforcing makes sense

I agree with nadav on this one, if anything we should just add a `--force` flag or something that allows us to ignore it in cases like demos, but I personally like things that throw explicit exceptions when i'm doing something that may be wrong.

Christewart:
> Timestamps are meta-data which do not always mean the same thing.
> 
> For example if someone wants to sign a message when the bitcoin price breaks a certain threshold, they may put the creation of the event as the timestamp, or something else, but whatever they choose to do as their convention these kinds of events should not be restricted by the gui because only the very boring "price at time T" actually has a timestamp where this enforcing makes sense

So maybe we should add a creation timestamp to the event? The spec is pretty clear at the moment that it is a maturation time, not a generic timestamp.

https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md#version-0-oracle_event

nkohen:
> So maybe we should add a creation timestamp to the event? The spec is pretty clear at the moment that it is a maturation time, not a generic timestamp.

It is out of date (updating is on my todo)

PR #2629, opened 2021-02-05
Create ScriptFactory.isValidAsm() to standardize how check validity o…

Christewart:
…f a script

Now you can just know you can call the `isValidAsm()` method rather than a bunch of names that are specific to the script type.


benthecarman (approved):

benthecarman:
Docs compile failing

PR #2635, opened 2021-02-07
Switch outcomeMap implementation to be parallel

Christewart:
This makes the computation of `ContractInfo.outcomeMap` happen in parallel. 

The current implementation is run on the default scala execution context with an `Await.result()` at the end so I didn't have to start futurifying the return type of `ContractInfo.outcomeMap`. 

Some rough benchmarks show a 2x improvement in test suite time for `dlcTest/test`. It seems the test suite with the old `ContractInfo.outcomeMap` would run in ~3 minutes.
Screenshot from 2021-02-07 15-10-02

The new implementation of `ContractInfo.outcomeMap` runs in ~1 min 45 seconds on my machine.

Screenshot from 2021-02-07 15-05-51

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
@@ -12,6 +12,7 @@ import org.bitcoins.core.protocol.tlv.{
   TLVSerializable,
   UnsignedNumericOutcome
 }
+import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.crypto.ECPublicKey
 
 import scala.concurrent.duration.DurationInt
@@ -160,17 +161,15 @@ case class ContractInfo(
     if (allOutcomesAndPayouts.isEmpty) {
       Future.successful(Map.empty)
     } else {
+
       val batchSize = Math.max(allOutcomesAndPayouts.length / parallelism, 1)
 
-      val results = allOutcomesAndPayouts.grouped(batchSize).map {
-        case batch: Vector[(OracleOutcome, Satoshis)] =>
-          Future {
-            executeBatch(batch)
-          }
-      }
-      Future
-        .sequence(results)
-        .map(_.flatten.toMap)
+      val results = FutureUtil.batchAndParallelExecuteSync(
+        allOutcomesAndPayouts,
+        executeBatch(_),
+        batchSize)
+
+      results.map(_.flatten.toMap)
     }
 
   }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
```diff
@@ -137,17 +138,41 @@ case class ContractInfo(
 
   /** Map OracleOutcomes (which correspond to CETs) to their adpator point and payouts */
   lazy val outcomeMap: Map[OracleOutcome, (ECPublicKey, Satoshis, Satoshis)] = {
-    val builder =
-      HashMap.newBuilder[OracleOutcome, (ECPublicKey, Satoshis, Satoshis)]
+    Await.result(
```
Christewart, line 142: Opinions of changing the return type to `Future[Map[...]]`? 

Also we probably want to give the abillity to passing in a custom execution context rather than running on the scala runtime execution context. This would mean we have to maek a choice between

1. Making this a method, and build in caching so we can provide the execution context
2. Make the execution context a class level parameter.

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala
```diff
@@ -108,4 +108,17 @@ object FutureUtil {
     val doneF = Future.sequence(execute)
     doneF
   }
+
+  /** Takes in a synchronous function f, and then executes that function on a set of elements in parallel */
+  def batchAndParallelExecuteSync[T, U](
+      elements: Vector[T],
+      f: Vector[T] => U,
+      batchSize: Int)(implicit ec: ExecutionContext): Future[Vector[U]] = {
+    val asyncF: Vector[T] => Future[U] = { case vec: Vector[T] =>
```
nkohen, line 117: nit: Don't need `case`

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
```diff
@@ -137,17 +139,39 @@ case class ContractInfo(
...
+
+      val batchSize = Math.max(allOutcomesAndPayouts.length / parallelism, 1)
+
+      val results = FutureUtil.batchAndParallelExecuteSync(
+        allOutcomesAndPayouts,
+        executeBatch(_),
+        batchSize)
 
-      builder.+=((outcome, (adaptorPoint, offerPayout, acceptPayout)))
+      results.map(_.flatten.toMap)
```
nkohen, line 172: should use `HashMap` in place of calling `toMap` because for some reason the default scala `Map` really sucks.

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
nkohen: I'm personally very hesitant towards starting to hyper-optimize and `Future`-ify this code without first being clear about how this library is going to work in the future (e.g. will there be two versions of dlc code where one is really fast using akka and the other is synchronous and simpler?)

nkohen (commented):
This looks fine but I think I might prefer #2642

Christewart:
Superseded by #2642

PR #2638, opened 2021-02-09
Modularize DLC Interface using companion objects

nkohen:
Before this, all DLC functionality required direct access to `DLCMessage`s meaning all functions constructed state from "backups" which is inefficient, hard to follow, and hard to test.

This will make testing modules easier (future PR) and many calls can be made more efficiently in the wallet (future PR).

Note: PSBTs will be further removed in a future PR.


core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCMessage.scala
```diff
@@ -23,6 +23,12 @@ object DLCMessage {
       .flip
   }
 
+  def computeContractId(
+      fundingTx: Transaction,
+      tempContractId: Sha256Digest): ByteVector = {
+    fundingTx.txIdBE.bytes.xor(tempContractId.bytes)
+  }
```
benthecarman: This should probably be in `DLCUtil`

dlc/src/main/scala/org/bitcoins/dlc/sign/DLCTxSigner.scala
```diff
@@ -71,299 +72,125 @@ case class DLCTxSigner(
...
-      pubKey: ECPublicKey): Future[PartialSignature] = {
-    val sigOpt = psbt.inputMaps.head.partialSignatures
-      .find(_.pubKey == pubKey)
-
-    sigOpt match {
-      case None =>
-        Future.failed(new RuntimeException("No signature found after signing"))
-      case Some(partialSig) => Future.successful(partialSig)
-    }
+  def completeFundingTx(remoteSigs: FundingSignatures): Future[Transaction] = {
```
benthecarman, line 80: Would be nice to have a 

```scala
def completeFundingTx(localSigs: FundingSignatures, remoteSigs: FundingSignatures): Transaction
```

benthecarman: same for cet

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCUtil.scala
```diff
@@ -0,0 +1,123 @@
...
+      sigPubKey
+        .extractAdaptorSecret(adaptorSig, ECDigitalSignature(cetSig.bytes.init))
+        .fieldElement
+
+    SchnorrDigitalSignature(outcome.aggregateNonce, possibleOracleS)
+  }
+
+  def computeOutcome(
+      completedSig: ECDigitalSignature,
+      possibleAdaptorSigs: Vector[(OracleOutcome, ECAdaptorSignature)]): (
```
Christewart: nit: Maybe just return `Option[(SchnorrDigitalSignature,OracleOutcome)]`? Seems like it would be useful to signify if it was found or not with an `Option[]` rather than throwing

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCUtil.scala
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  def computeOutcome(
+      isInitiator: Boolean,
+      offerFundingKey: ECPublicKey,
+      acceptFundingKey: ECPublicKey,
+      contractInfo: ContractInfo,
+      localAdaptorSigs: Vector[(OracleOutcome, ECAdaptorSignature)],
+      cet: Transaction): (SchnorrDigitalSignature, OracleOutcome) = {
```
Christewart: any reason we don't just tighten this type up to be `cet: WitnessTransaction`

dlc/src/main/scala/org/bitcoins/dlc/builder/DLCTxBuilder.scala
```diff
@@ -84,88 +84,307 @@ case class DLCTxBuilder(offer: DLCOffer, accept: DLCAcceptWithoutSigs)(implicit
...
+    val offerTotalFunding: CurrencyUnit =
+      offerFundingInputs.map(_.output.value).sum
+
+    // The sum of all funding input amounts from the non-initiator
+    val acceptTotalFunding: CurrencyUnit =
+      acceptFundingInputs.map(_.output.value).sum
+
+    require(
+      offerTotalFunding >= offerInput,
+      "Offer funding inputs must add up to at least offer's total collateral")
```
Christewart: nits: Add the amounts to the error message, always helps when debugging :-)

```scala
 require(
      offerTotalFunding >= offerInput,
      "Offer funding inputs must add up to at least offer's total collateral, offerTotalFunding=$offerTotalFunding offerInput=$offerInput")
```

Christewart: same with below

dlc/src/main/scala/org/bitcoins/dlc/builder/DLCTxBuilder.scala
```diff
@@ -84,88 +84,307 @@ case class DLCTxBuilder(offer: DLCOffer, accept: DLCAcceptWithoutSigs)(implicit
...
+  def buildCETs(
+      outcomes: Vector[OracleOutcome],
+      contractInfo: ContractInfo,
+      offerFundingKey: ECPublicKey,
+      offerFinalSPK: ScriptPubKey,
+      acceptFundingKey: ECPublicKey,
+      acceptFinalSPK: ScriptPubKey,
+      timeouts: DLCTimeouts,
+      fundingOutputRef: OutputReference): Vector[
+    (OracleOutcome, WitnessTransaction)] = {
```
Christewart: Seems like it would be worthwhile to make a

```scala
/** An oracle outcome and it's corresponding CET */
case class OutcomeCETPair(outcome: OracleOutcome, wtx: WitnessTransaction)
```

dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
```diff
@@ -85,33 +85,74 @@ case class DLCExecutor(signer: DLCTxSigner)(implicit ec: ExecutionContext) {
...
+  def executeRefundDLC(dlcSetup: SetupDLC): RefundDLCOutcome = {
+    val SetupDLC(fundingTx, _, refundTx) = dlcSetup
+    RefundDLCOutcome(fundingTx, refundTx)
+  }
+}
+
+object DLCExecutor {
 
-    val threshold = dlcSetup.cets.head._1.oracles.length
+  def executeDLC(
```
Christewart, line 116: Could you add a scaladoc on this method for what the behavior is when we find a signature that does not correspond to an outcome? 

I have a hard time understanding what is happening in these scenarios, as we should be able to always recover the outcome from the signature -- at least that is my understanding?

Specifically i'm talking about https://github.com/nkohen/bitcoin-s-core/blob/ac4aa8edd6063e726e54567a0c72e93a8be74c89/dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala#L132

dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
```diff
@@ -85,33 +85,74 @@ case class DLCExecutor(signer: DLCTxSigner)(implicit ec: ExecutionContext) {
...
-        (msg, cetInfo.remoteSignature)
+
+    val msgAndCETInfoOpt = msgOpt.flatMap { msg =>
+      remoteCETInfos.find(_._1 == msg)
+    }
+
+    val (msg, ucet, remoteAdaptorSig) = msgAndCETInfoOpt match {
+      case Some((msg, CETInfo(ucet, remoteSig))) => (msg, ucet, remoteSig)
       case None =>
         throw new IllegalArgumentException(
```
Christewart, line 140: so is this a "fatal" error? This seems really bad to me, an outcome occurred we didn't know about?

dlc/src/main/scala/org/bitcoins/dlc/sign/DLCTxSigner.scala
Christewart: I'm guessing this is needed because of current `Sign` API, nadav is working on making a synchronous version in #2652

Christewart (changes requested):

dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
nkohen: I mean if all the oracles sign "hehe, screw my users" there isn't really anything you can do about that.

Christewart (approved):

benthecarman (approved):

PR #2642, opened 2021-02-09
Compute `sigPoint`s eagerly but asynchronously

nkohen:
This implements a simpler solution to the problem solved in #2635 because computing the `outcomeMap` in a `ContractInfo` isn't the issue, the bottleneck is solely computing the adaptor points (aka sigPoints) of each `OracleOutcome`.

As such, this solution simply has these `sigPoint`s evaluated eagerly but asynchronously with calls to `sigPoint` delegating to `Await.result` on the eager `Future` where it should be noted that `Await.result` does no blocking or anything else fancy in the case that the `Future` is completed (in which case it synchronously returns as if `fut.value.get.get` was called).

I ran this locally and got the exact same speedup that I got from running #2635 compared to the current `adaptor-dlc` branch.

I'm still open to either approach but figured I should open this for discussion.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/OracleOutcome.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/OracleOutcome.scala
@@ -7,6 +7,9 @@ import org.bitcoins.core.protocol.tlv.{
 }
 import org.bitcoins.crypto.{ECPublicKey, SchnorrNonce}
 
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext, Future}
+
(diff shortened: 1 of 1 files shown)


Christewart (approved):
Reproduced the same results locally, seems good enough to me and simpler! 

However I think we should PR this onto master and then rebase adaptor-dlc?

Christewart:
Hmm.. this looks a little suspicious? 

Failing builds on both 2.13 / 2.12

https://github.com/bitcoin-s/bitcoin-s/pull/2642/checks?check_run_id=1890775987#step:5:829

https://github.com/bitcoin-s/bitcoin-s/pull/2642/checks?check_run_id=1890775731#step:5:968

Christewart (approved):

PR #2652, opened 2021-02-10
Introduced distinct Sign and AsyncSign interfaces

nkohen:
Currently all code which uses signing does so through `ECPrivateKey` which has entirely synchronous signing. As such, the sign interface now also supports synchronous signing and asynchronous signers will instead extend the `AsyncSign` trait. This will mean that eventually, when AsyncSigners what to integrate, every method in `crypto` and `core` which handles a `Sign` will need a duplicate method that handles `AsyncSign`. This will lead to interfaces with different methods for sync and async signing functions so that sync signers won't have to get bogged down in futures and execution contexts. As of right now this PR does not implement these duplicate async functions.

### DIFF
--- a/crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
+++ b/crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
@@ -2,8 +2,8 @@ package org.bitcoins.crypto
 
 import scodec.bits.ByteVector
 
-import scala.concurrent.duration.DurationInt
-import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.annotation.tailrec
+import scala.concurrent.{ExecutionContext, Future}
 
 /** This is meant to be an abstraction for a [[org.bitcoins.crypto.ECPrivateKey]], sometimes we will not
   * have direct access to a private key in memory -- for instance if that key is on a hardware device -- so we need to create an
@@ -17,11 +17,8 @@ import scala.concurrent.{Await, ExecutionContext, Future}
   * If you have a hardware wallet, you will need to implement the protocol to send a message to the hardware device. The
   * type signature of the function you implement must be scodec.bits.ByteVector => Future[ECDigitalSignature]
   */
-trait Sign {
-  def signFunction: ByteVector => Future[ECDigitalSignature]
-
-  def signFuture(bytes: ByteVector): Future[ECDigitalSignature] =
-    signFunction(bytes)
+trait AsyncSign {
+  def asyncSign(bytes: ByteVector): Future[ECDigitalSignature]
 
   /** Note that using this function to generate digital signatures with specific
     * properties (by trying a bunch of entropy values) can reduce privacy as it will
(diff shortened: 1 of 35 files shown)


benthecarman (changes requested):
Docs not compiling

benthecarman (commented):
Otherwise LGTM

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -32,88 +29,164 @@ trait Sign {
...
+    * optional for now. Maybe we rethink the idea of the signFunction in the future.
+    * the public key is still useful here though because it can be used to match against
+    * a specific private key on another server
+    */
+  def dummySign(publicKey: ECPublicKey): AsyncSign = {
+    constant(EmptyDigitalSignature, publicKey)
+  }
+}
+
+trait Sign extends AsyncSign {
```
Christewart, line 110: Is this meant to extend `AsyncSign`? I thought the purpose is to be synchronous

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -32,88 +29,164 @@ trait Sign {
...
-  def signLowR(bytes: ByteVector)(implicit
-      ec: ExecutionContext): ECDigitalSignature = {
-    Await.result(signLowRFuture(bytes), 30.seconds)
+  def constant(sig: ECDigitalSignature, pubKey: ECPublicKey): AsyncSign = {
+    AsyncSignImpl(_ => Future.successful(sig),
+                  (_, _) => Future.successful(sig),
+                  pubKey)
   }
 
+  /** This dummySign function is useful for the case where we do not have the
```
Christewart, line 98: If we go foreword with this refactor, it seems maybe we can eliminate the `signFunction` and just have async signers implement the `AsyncSign` interface?

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -32,88 +29,164 @@ trait Sign {
...
+
+  override def asyncSignWithEntropy(
+      bytes: ByteVector,
+      entropy: ByteVector): Future[ECDigitalSignature] = {
+    Future.successful(signWithEntropy(bytes, entropy))
   }
 
-  def publicKey: ECPublicKey
+  @tailrec
+  private def signLowR(bytes: ByteVector, startAt: Long): ECDigitalSignature = {
```
Christewart, line 137: nit: Any documentation on what is going here?

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
nkohen: If some other function requires an `AsyncSign`, it is always possible to give it a `Sign` which as you'll see below simply wraps results in `Future.successful`.

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
nkohen: I'm just going to delete this scaladoc as it was copied in and no longer seems to make any sense, I've already eliminated `signFunction` stuff from the actual traits

Christewart (approved):

Christewart:
LGTM assuming Ci is fine

benthecarman (approved):

PR #2653, opened 2021-02-11
Give oracle server its own port

### LINKED ISSUE #2646: Different default server ports for oracleServer and appServer
Feedback from Pierre

One paint point in going through the bitcoin-s setup process was the limitation of running the oracle server and appServer together "out of the box". This is because they both default to port `9999`. 

We do offer the ability to specify a different default port in two ways 

1. `--rpcport` arg on startup
2. configuring it in the bitcoin-s.conf file

however for new users that don't understand how bitcoin-s works, this can be confusing. One thing we can do to avoid this problem is just using different default ports, and then have some sort of logic baked into `bitcoin-s-cli` that is aware of default ports for different projects. 

This maybe useful for us in the future too if we start needing the ability to run multiple servers at the same time too (maybe somebody we will want to run an `appServer` and `oracleServer` simultaneously!)

benthecarman:
Closes #2646

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -54,7 +54,7 @@ object ConsoleCli {
         .action((_, conf) => conf.copy(debug = true))
         .text("Print debugging information"),
       opt[Int]("rpcport")
-        .action((port, conf) => conf.copy(rpcPort = port))
+        .action((port, conf) => conf.copy(rpcPortOpt = Some(port)))
         .text(s"The port to send our rpc request to on the server"),
       opt[Unit]("version")
         .action((_, conf) => conf.copy(command = GetVersion))
@@ -1629,14 +1629,21 @@ case class Config(
     command: CliCommand = CliCommand.NoCommand,
     network: Option[NetworkParameters] = None,
     debug: Boolean = false,
-    rpcPort: Int = 9999
-)
+    rpcPortOpt: Option[Int] = None) {
+
+  val rpcPort: Int = rpcPortOpt match {
+    case Some(port) => port
+    case None       => command.defaultPort
+  }
+}
 
 object Config {
   val empty: Config = Config()
 }
 
-sealed abstract class CliCommand
+sealed abstract class CliCommand {
+  def defaultPort: Int = 9999
+}
 
 object CliCommand {
   case object NoCommand extends CliCommand
@@ -1647,9 +1654,15 @@ object CliCommand {
 
   sealed trait ServerlessCliCommand extends CliCommand
 
+  sealed trait AppServerCliCommand extends CliCommand
+
+  sealed trait OracleServerCliCommand extends CliCommand {
+    override def defaultPort: Int = 9998
+  }
+
   case object GetVersion extends ServerlessCliCommand
 
-  case object GetInfo extends CliCommand
+  case object GetInfo extends AppServerCliCommand
 
   // DLC
   case class CreateDLCOffer(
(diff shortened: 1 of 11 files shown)


app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
```diff
@@ -17,6 +17,8 @@ trait BitcoinSRunner extends BitcoinSLogger {
 
   def actorSystemName: String
 
+  def defaultRpcPort: Int
```
Christewart: So in my opinion this should be read from the default `reference.conf` file. 

We have settings for these in these files I believe 

`bitcoin-s.server.rpcport`

`bitcoin-s.oracle.rpcport`

I'm not seeing the latter in the default configuration file, but we should add it. Basically any project that is a server project should have the ability to set the rpc port 

https://bitcoin-s.org/docs/config/configuration#example-configuration-file

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1629,14 +1629,21 @@ case class Config(
...
+  }
+}
 
 object Config {
   val empty: Config = Config()
 }
 
-sealed abstract class CliCommand
+sealed abstract class CliCommand {
+  def defaultPort: Int = 9999
```
Christewart: Nice! I like this design. 

Let's just leave the defaultPort undefined in the superclass and make all the subprojects define it. 

Basically any server project we have is going to require to have a unique rpc port

Christewart: My understanding is this won't be able to be read from a configuration file as the cli doesn't interact with `bitcoin-s.conf`, is that right?

Christewart (changes requested):

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: > My understanding is this won't be able to be read from a configuration file as the cli doesn't interact with bitcoin-s.conf, is that right?

Correct

app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
benthecarman: We still need a default port hard coded in, I think the current way is better.

Christewart: I guess in my mind it would be hardcoded in the reference.conf file and then read as part of the `AppConfig`. In general i think it's worth trying to preserve the property that every configuration 

1. Lives in a `application.conf`/`reference.conf` file
2. Is passed in explicitly with a command line param (`--rpcport`)

With `bitcoin-s-cli` it isn't possible to achieve (1), so I'm more amenably to hard coding there, but I would like to try and preserve this property with the server. That way when we show example confs on the website, users can understand where the configuration is coming from rather than having to dig in source code.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
```diff
@@ -104,13 +104,7 @@ case class DLCOracleAppConfig(
...
   }
 
-  def rpcPortOpt: Option[Int] = {
-    if (serverConf.hasPath("bitcoin-s.server.rpcport")) {
-      Some(serverConf.getInt("bitcoin-s.server.rpcport"))
-    } else {
-      None
-    }
-  }
+  def rpcPort: Int = config.getInt("bitcoin-s.server.rpcport")
```
Christewart: I think this needs to be `bitcoin-s.oracle.rpcport` i think?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
```diff
@@ -104,13 +104,7 @@ case class DLCOracleAppConfig(
     config.getConfig("bitcoin-s.server")
```
Christewart: also this is probably wrong?

Christewart (commented):
Also the example configuration on the website needs to be updated?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
benthecarman: I don't think that makes sense, this isn't configuring the oracle, but the rpc server

Christewart (approved):

PR #2656, opened 2021-02-11
Simplify oracle server RPC api

### LINKED ISSUE #2644: Give ability to use human readable date for createenumevent/createdigitdecompevent
Feedback given by Pierre 

> Instead of passing in a maturation time in seconds since epoch we should be able to pass in a date that just assumes 00:00:00 UTC expiration. The command could look like this 

>./bitcoin-s-cli createenumevent 2020-us-election 2021-01-01 "Republican_win,Democrat_win,other"

Behind the scenes, we can convert this date (`2021-01-01`) to epoch seconds.

### LINKED ISSUE #2645: Add ability to getevent by name
Feedback from Pierre 

When Ben & I watched Pierre go through the setup process and interact with the oracle server, one thing he intuitively thought would be available to him is retrieving an event by name.

This makes sense to me, we should definitely offer this. There is a question around uniqueness of event names (should we enforce uniqueness? If not, do we return all events with that name? Is there a spec level consideration here?) but I think this is a very useful rpc we should add.

### LINKED ISSUE #2647: Create a simpler API for `createdigitdecompevent` in base 10
Feedback from Pierre

We should create a higher level API than what is available currently in `createdigitdecompevent`.

his suggestions

1. I don't care about number base, you guys should choose what's best for me as you know more than I do
2. I should just be able to specify a min/max value in base 10 for what my expected outcomes can be. 

Base on the `min/max`, we can determine if it is unsigned or a signed digit decomp event. 

We didn't introduce the concept of `precision` to Pierre, but we may want to consider that to default to `0` too with the simpler API. 

Personally, I think we should offer both APIs. One for power users, and a simpler one that we tell new users about.

benthecarman:
Closes #2647 
Closes #2645
Closes #2644


dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -182,6 +190,9 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
             s"Event cannot mature in the past, got $maturationTime")
 
     for {
+      dbs <- eventDAO.findByEventName(eventName)
+      _ = require(dbs.isEmpty, s"Event name ($eventName) is already being used")
```
Christewart, line 199: Why don't we just enforce this at the database level?

Christewart (changes requested):
So here is my take on the `createdigitdecompevent`

I think we should keep that API, but advertise the new API (`createnumericevent`) to beginners. The new API should be "easy" to use (base 10 values, human readable dates, allowing a `min` parameter and a `max` parameter from which we infer the appropriate range, etc). 

If someone wants to sign in different bases, we can point them to the "advanced" or "power user" API which should be `createdigitdecompevent`. 

We should probably add an "Advanced usage" section on the website for this API or something similar, and provide the same numeric example using both APIs on the site, with the easy one `createnumericevent`, shown first.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Because we have mulitple event dbs per event (one for each nonce)

Christewart: Can't we have an aggregate unique constraint between the `event_name` and `event_descriptor_tlv` or `announcement_signature`

https://www.sqlitetutorial.net/sqlite-unique-constraint/

>  UNIQUE(column_name1,column_name2,...)

I'm sure there is something similar for postgres

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1823,14 +1899,23 @@ object CliCommand {
...
       label: String,
       maturationTime: Instant,
       outcomes: Seq[String])
       extends CliCommand
 
+  case class CreateNumericEvent(
+      eventName: String,
+      maturationTime: Instant,
+      maxValue: Long,
+      isSigned: Boolean,
```
Christewart: so either we need to get rid of `isSigned` (we don't have `minValue`, so it has to be zero), or add `minValue` so that we can do signed digit decomp events right?

docs/oracle/oracle-price-example.md
```diff
@@ -65,7 +65,7 @@ predetermined interval, you need to sign the min (`0`) or the max (`$131,071`).
 Given the parameters we specified above, we are ready to create our digit decompisition event!
 
 ```
- ./bitcoin-s-cli createdigitdecompevent "bitcoin-s-price-example" 1612396800 2 17 "BTC/USD" 0
+ ./bitcoin-s-cli createnumericevent "bitcoin-s-price-example" 1612396800 2 17 "BTC/USD" 0
```
Christewart: These need to e updated, or even better move this into a "power user"/"advanced" section, and replace the existing section with a `createnumericevent` sample. 

I do think both examples should create teh same announcement, so the reader can compare and contrast the baked in assumptions the `createnumericevent` API makes (base2, inferred num digits,precision=0)

Christewart (commented):
One other thing, i think we should add a `getannouncement [oracle-announcement]` API so that you can query by announcement if you like. I think that could still be useful if you happen to have an announcement, but can't decode it and get the `eventName`

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1086,6 +1086,70 @@ object ConsoleCli {
...
+              })),
+          opt[Unit]("signed")
+            .text("Whether the outcomes can be negative")
+            .action((_, conf) =>
+              conf.copy(command = conf.command match {
+                case createNumericEvent: CreateNumericEvent =>
+                  createNumericEvent.copy(isSigned = true)
+                case other => other
+              })),
+          arg[String]("unit")
```
nkohen, line 1136: This is more of a general comment that might be worth pursuing in a future PR, I think it might be possible to factor out some of these common args into defs so that when they are used multiple places you only need to pass a function putting the value into the case class via copy

app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleRoutes.scala
```diff
@@ -56,6 +56,36 @@ case class OracleRoutes(oracle: DLCOracle)(implicit
...
+              CreateNumericEvent(eventName,
+                                 maturationTime,
+                                 maxValue,
+                                 isSigned,
+                                 unit,
+                                 precision)) =>
+          complete {
+
+            val numDigits =
+              Math.ceil(Math.log10(maxValue.toDouble) / Math.log10(2)).toInt
```
nkohen: super nit: `Math.log10 -> math.log`

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: imo I think minValue will make it more confusing, someone might try to do something like `min=1,000 max=10,000` and we'd have to make an event from 0-10,000. Or someone could do `min=-1,000 max=10,000` and wehave have to do -10,000 - 10,000.

Maybe changing `isSigned` to `includeNegatives` would be a better framing?

Christewart: Doesn't the same principle apply with `maxValue` alone? I.e. if I my application requires `maxValue=2^32+1`, at the protocol level `maxValue=2^33`.

This PR is about providing a simple API, which allowing a user to specify their `minValue` will do in my opinion.

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
Christewart: Worth filing an issue for this?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
nkohen: @benthecarman I agree with your assessment of the consequences of doing it Chris's way but I actually don't think this is a problem, there is no issue with allowing out of bounds things into our domain so long as we cover the entire range from min to max. If we're going for an easy UI we should do this simplification and more sophisticated users can use numdigits is signed etc.

benthecarman: Went ahaead with the minValue param

docs/oracle/oracle-price-example.md
benthecarman: Updated to explain both, lmk what you think

benthecarman:
Responded to review as well as implemented a YYYYMMDD format for #2644 in `createenumevent` and `createnumericevent`, however in `createdigitdecompevent` I left it is as the epoch second since it is the more advanced api. LMK what you guys think.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: After doing some research I don't think it is possible to have a constraint that enforces that only `event_name` can be paired with its one `announcement_signature`

benthecarman:
Updated to use ISO 8601 for time format

Christewart (changes requested):
It's not clear to me if just date is allowed, you should be able to search the input string for the separator `T` (`2021-02-19T19:36:31+00:00`) If it doesn't contain a `T` we can just allow a date and assume the time is just 00:00:00 UTC.

I wouldn't be surprised if our time library already has this baked in, but i don't think there is a test case that checks this?

In general, there seems to be no `oracle-server` level test cases for this API?

.github/workflows/Linux_2.13_App_Chain_Node_Core_Tests.yml
```diff
@@ -23,4 +23,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: run tests
-        run: sbt ++2.12.12 downloadBitcoind coverage chainTest/test chain/coverageReport chain/coverageAggregate chain/coveralls nodeTest/test node/coverageReport node/coverageAggregate node/coveralls cryptoTest/test crypto/coverageReport crypto/coverageAggregate crypto/coveralls coreTest/test core/coverageReport core/coverageAggregate core/coveralls secp256k1jni/test zmq/test zmq/coverageReport zmq/coverageAggregate zmq/coveralls appCommonsTest/test appServerTest/test
+        run: sbt ++2.12.12 downloadBitcoind coverage chainTest/test chain/coverageReport chain/coverageAggregate chain/coveralls nodeTest/test node/coverageReport node/coverageAggregate node/coveralls cryptoTest/test crypto/coverageReport crypto/coverageAggregate crypto/coveralls coreTest/test core/coverageReport core/coverageAggregate core/coveralls secp256k1jni/test zmq/test zmq/coverageReport zmq/coverageAggregate zmq/coveralls appCommonsTest/test appServerTest/test oracleServerTest/test
```
Christewart, line 26: I think  `++2.12.12` -> `++2.13.4`

app/oracle-server-test/src/test/resources/reference.conf
```diff
@@ -0,0 +1,8 @@
+bitcoin-s {
```
Christewart: So why do we need to have this stuff for the oracle app config? 

Is this just a hard requirement for anything that uses a db? 

We should look into refactoring this in the future if that is the case (file an issue).

app/oracle-server-test/src/test/scala/org/bitcoins/oracle/server/OracleRoutesSpec.scala
```diff
@@ -0,0 +1,317 @@
...
+import org.bitcoins.server.routes.ServerCommand
+import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.wordspec.AnyWordSpec
+import ujson._
+
+import java.time.Instant
+import scala.concurrent.Future
+
+class OracleRoutesSpec
```
Christewart, line 32: :tada:

app/oracle-server-test/oracle-server-test.sbt
```diff
@@ -0,0 +1,3 @@
+name := "bitcoin-s-server-test"
```
Christewart: nit: `bitcoin-s-oracle-server-test`

core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
```diff
@@ -0,0 +1,81 @@
...
+import org.bitcoins.core.number._
+import org.bitcoins.core.protocol.Bech32Address
+import org.bitcoins.core.protocol.dlc.SigningVersion
+import org.bitcoins.core.protocol.tlv._
+import org.bitcoins.crypto.{SchnorrNonce, SchnorrPublicKey}
+
+import java.time.Instant
+import scala.concurrent.Future
+
+trait DLCOracleApi {
```
Christewart, line 14: Good idea

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -922,6 +923,18 @@ object OracleAttestmentV0TLV extends TLVFactory[OracleAttestmentV0TLV] {
 
     OracleAttestmentV0TLV(eventId, pubKey, sigs, outcomes)
   }
+
+  val dummy: OracleAttestmentV0TLV = {
```
Christewart: This needs to be lazy i'm guessing, I see native image builds are failing, probably because we are eagerly evaluating `ECPrivateKey.freshPrivateKey` which is not allowed when building a native image 

https://github.com/bitcoin-s/bitcoin-s/pull/2656/checks?check_run_id=1952997296#step:5:322

core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
```diff
@@ -13,4 +17,25 @@ object TimeUtil {
   def currentEpochSecond: Long = {
     now.getEpochSecond
   }
+
+  def iso8601ToDate(str: String): Date = {
```
Christewart, line 21: What was the reason for using `Date` rather than `Instant`? It seems `Instant` is superior in every way. Was there some compatibility issue?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -19,9 +21,10 @@ import org.bitcoins.keymanager.{DecryptedMnemonic, WalletStorage}
 import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
 
-case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
+class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
```
Christewart, line 24: Good idea on `private[this]` , is there a reason for going from `case class ` -> `class` ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -234,7 +250,19 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     }
   }
 
-  def signEvent(
+  override def signEvent(
+      eventName: String,
+      outcome: DLCAttestationType): Future[EventDb] = {
```
Christewart: can't we just tighten this type to 

`outcome: EnumAttestation` so we know it's an enum attestation ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -234,7 +250,19 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
...
+      _ = require(eventDbs.size == 1,
+                  "Use signLargeRange for signing multi nonce outcomes")
+
+      sign <- signEvent(eventDbs.head.nonce, outcome)
+    } yield sign
+  }
+
+  override def signEvent(
       oracleEventTLV: OracleEventTLV,
       outcome: DLCAttestationType): Future[EventDb] = {
```
Christewart: same here, `outcome: EnumAttestation`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
Christewart: also maybe just rename the method to be `signEnumEvent()` as that is all it can do if i understand

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -246,7 +274,7 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     } yield sign
   }
 
-  def signEvent(
+  override def signEvent(
```
Christewart, line 280: a scaladoc would be useful on this method a lot is going on in  here.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -297,7 +325,16 @@ case class DLCOracle(private val extPrivateKey: ExtPrivateKeyHardened)(implicit
     } yield updated
   }
 
-  def signDigits(
+  override def signDigits(eventName: String, num: Long): Future[OracleEvent] = {
+    for {
+      eventOpt <- findEvent(eventName)
+      _ = require(eventOpt.isDefined,
```
Christewart, line 334: It seems like all the return types should just reflect the possibility the event isn't defined with a `Future[Option[...]]` ?

Not just this method, but the other `signEvent()` methods too.

Christewart (changes requested):

.github/workflows/Linux_2.13_App_Chain_Node_Core_Tests.yml
benthecarman: fixed in #2713

app/oracle-server-test/src/test/resources/reference.conf
benthecarman: Oh, this was accidentially copied over from the app server test dir

core/src/main/scala/org/bitcoins/core/util/TimeUtil.scala
benthecarman: We can't use `Instant` because we pass that back and forth using epoch second. To send it using a different format we need to use a different type.

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: Can't have `private[this]` in a `case class`

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: This would be better for a future PR.

However, I disagree with doing this. I think it makes more sense to throw here.

Christewart: I disagree, the "error" is the the `eventName` doesn't exist, which is what `Option` is meant to represent (this thing _may_ exist, or it may not, in which case we give you `None`).

Christewart: Actually, I think you are right. This is signing for something they think exists in the database. Nvm.

Christewart (approved):

Issue #2112, opened 2020-10-02
Refactor HDCoinType to be ADT 


Christewart:
I definitely don't know what `int`s represent what coin types. 

https://github.com/bitcoin-s/bitcoin-s/pull/2097#pullrequestreview-501234705

benthecarman:
Closed by #2657

PR #2657, opened 2021-02-12
Refactor HDCoinType to be ADT

saguywalker:
Hello! this this my first PR here to refactor HDCoinType according to issue https://github.com/bitcoin-s/bitcoin-s/issues/2112, feel free to comment :)

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/hd/HDCoinType.scala
+++ b/core/src/main/scala/org/bitcoins/core/hd/HDCoinType.scala
@@ -19,15 +19,15 @@ sealed trait HDCoinType {
 object HDCoinType {
 
   final case object Bitcoin extends HDCoinType {
-    override def toInt: Int = 0
+    override val toInt: Int = 0
   }
 
   final case object Testnet extends HDCoinType {
-    override def toInt: Int = 1
+    override val toInt: Int = 1
   }
 
   final case class UnknownCoinType(int: Int) extends HDCoinType {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/hd/HDCoinType.scala
```diff
@@ -9,32 +9,46 @@ import org.bitcoins.core.config._
...
 
 /** @see [[https://github.com/satoshilabs/slips/blob/master/slip-0044.md SLIP-0044]]
   *     central registry of coin types
   */
 object HDCoinType {
 
-  final val Bitcoin = HDCoinType(0)
-  final val Testnet = HDCoinType(1)
+  final case object Bitcoin extends HDCoinType {
+    override def toInt: Int = 0
```
Christewart: nit: `override val toInt: Int = 0` (it doesn't need to be re-evaluated everytime)

Christewart (approved):
LGTM assuming no CI errors/warnings. 

We do currently have spurious test case failures on CI, so when the build is done I will let you know these are new errors or legacy issues we know we have. Thanks for the contribution :-)

core/src/main/scala/org/bitcoins/core/hd/HDCoinType.scala
saguywalker: fixed

benthecarman (approved):

PR #2658, opened 2021-02-12
CryptoRuntime abstraction

### LINKED ISSUE #2655: Remove direct usages of CryptoUtil in core, create CryptoContext.cryptoRuntime and use that instead
As discussed with @rorp on a call, we need to avoid direct usage of CryptoUtil through the project. The methods inside of CryptoUtil are platform specific to the JVM. We need to reference the platform specific crypto runtime (this can be javascript, bouncy castle, secp256k1 etc). This is also discussed in #2357. 

Along with `CryptoContext.default`, we should have `CryptoContext.cryptoRuntime`. And then everytime we need to call a method that is related to crypto function, we should do `CryptoContext.cryptoRuntime.sign()` or something along those lines. 

If we didn't want to add the `cryptoRuntime` method, we will need to pattern match through the code base with something like this

```scala
CryptoContext.default match { 
  case BouncyCastle => BouncyCastleUtil.sign() //not exact code, but you get the idea
  case Secp256k1 => NativeSecp256k1.sign() //not exact code, but you get the idea
  case Bcrypto => JsCryptoRuntime.sign() //not exact code, but you get the idea
}
```

This is the style we currently use in places like ECKey

https://github.com/bitcoin-s/bitcoin-s/blob/bf69c90c97e066d40d581a8ae06a0f8a1a38c4b3/crypto/src/main/scala/org/bitcoins/crypto/SchnorrPublicKey.scala#L105

Christewart:
fixes #2655 

This PR introduces a platform agnostic `CryptoRuntime` trait into the `core` modules. This work does not touch any other modules besides `crypto` and `core`. 

Now every class/trait that needs access to cryptographic functions needs to extend `CryptoTrait` which has a reference to the platform specific `cryptoRuntime`.

The diff is so large because every data type in `core` that requires a cryptographic function needs to now extend `CryptoTrait` and NOT use `CryptoUtil` directly. 

We should probably rename `CryptoUtil` -> `JvmCryptoRuntime` or something like that in the future.


benthecarman:
It it not possible to make `CryptoTrait` an object so we don't need to have it extended in everything that needs it?

Christewart:
> It it not possible to make `CryptoTrait` an object so we don't need to have it extended in everything that needs it?

It is possible, but at every call site we need to do `CryptoTrait.cryptoRuntime.method()`. I prefer the way i implemented it because it is shorter to read, but it does pollute the class hierarchy. So :man_shrugging: 

A long winded way of saying, if other feel strongly about it, i can change it, but let's get feedback from Roman/Nadav first and see what they think.

benthecarman:
Could you not have `CryptoUtil` have the `cryptoRuntime` and then move all the implementations in `CryptoUtil` to a new class. Then you also wouldn't need the large refactor

Christewart:
> Could you not have `CryptoUtil` have the `cryptoRuntime` and then move all the implementations in `CryptoUtil` to a new class. Then you also wouldn't need the large refactor

Yes, i think we can and this seems like a much better alternative. I'll code this up later, but that seems superior in pretty much everyway to this big diff i have currently.

Christewart:
> > Could you not have `CryptoUtil` have the `cryptoRuntime` and then move all the implementations in `CryptoUtil` to a new class. Then you also wouldn't need the large refactor
> 
> Yes, i think we can and this seems like a much better alternative. I'll code this up later, but that seems superior in pretty much everyway to this big diff i have currently.

Though this will require me to do this, which makes sense IMO. 

>We should probably rename CryptoUtil -> JvmCryptoRuntime or something like that in the future.

I wouldn't just do a straight rename, but i will have to move the Jvm specific calls into that file.

crypto/src/main/scala/org/bitcoins/crypto/SchnorrNonce.scala
```diff
@@ -14,7 +14,7 @@ case class SchnorrNonce(bytes: ByteVector) extends NetworkElement {
   }
 }
 
-object SchnorrNonce extends Factory[SchnorrNonce] {
+object SchnorrNonce extends Factory[SchnorrNonce] with CryptoTrait {
```
benthecarman: Can these be reverted now?

crypto/src/main/scala/org/bitcoins/crypto/SchnorrPublicKey.scala
```diff
@@ -6,7 +6,9 @@ import scodec.bits.ByteVector
 import scala.annotation.tailrec
 import scala.util.Try
 
-case class SchnorrPublicKey(bytes: ByteVector) extends NetworkElement {
+case class SchnorrPublicKey(bytes: ByteVector)
```
benthecarman: ditto

crypto/src/main/scala/org/bitcoins/crypto/SchnorrNonce.scala
Christewart: oh yeah, i reverted the entire core project (and it worked!) but i suppose it doesn't get the `crypto` module. I'll just delete `CryptoTrait` now as it is unnecessary i believe

benthecarman (approved):

crypto/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
```diff
@@ -1,154 +1,70 @@
...
+  * such as [[JvmCryptoRuntime]].
+  *
+  * This is necessary so that the core module doesn't need to be refactored
+  * to add support for multiple platforms, it can keep referencing CryptoUtil
   */
-trait CryptoUtil {
+trait CryptoUtil extends CryptoRuntime {
+
+  /** The underlying runtime for the specific platform we are running on */
+  private lazy val cryptoRuntime: CryptoRuntime = CryptoContext.cryptoRuntime
```
nkohen, line 18: I'm a little confused at the simultaneous is-a and has-a, why both?

nkohen: Why have a `CryptoUtil` at all if people can just call `CryptoRuntime`?

Christewart: https://github.com/bitcoin-s/bitcoin-s/pull/2658#issuecomment-778315942

If you want to see what a previous implementation looked like before doing this, see f1c9989

nkohen (approved):
Until we think of something better :P

rorp (approved):

PR #2662, opened 2021-02-13
Refactoring `chain`

ariskk:
Hey 👋
Another PPR (Pedantic Pull Request) from me!

This one:
- Fixes the typo in `compObjectfromHeaders`
- Some performance and readability improvements in `getAncestorAtHeight`
- Some readability refactoring in `ChainHandler`
- Replaces `FutureUtils.unit` with `Future.unit`. At the end of the day both are a call to Promise.fromTry

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
@@ -59,7 +59,7 @@ case class BlockHeaderDAO()(implicit
     table.filter(_.hash.inSet(hashes))
   }
 
-  /** Retrives the ancestor for the given block header at the given height
+  /** Retrieves the ancestor for the given block header at the given height
     * @param child
     * @param height
     * @return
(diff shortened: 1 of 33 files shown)


chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
```diff
@@ -71,37 +71,29 @@ case class BlockHeaderDAO()(implicit
...
       headers.foreach { header =>
-        val index = (header.height - height).toInt
+        val index = header.height - height
         headersByHeight(index) = headersByHeight(index).:+(header)
       }
 
-      // Now that the bucket sort is done, we get rid of mutability
-      val groupedByHeightHeaders: List[Vector[BlockHeaderDb]] =
-        headersByHeight.toList
+      val groupedByHeightHeaders: Array[Vector[BlockHeaderDb]] =
```
ariskk, line 90: Since the Array is not leaked via the return type of the method, there is little benefit in converting it here. It takes 5x longer to convert to List and reverse, than to just reverse the Array.
```
@ (1 to 1000).map(_ => takes(() => a.reverse   )).sum / 1000 
res59: Long = 2659141L

@ (1 to 1000).map(_ => takes(() => a.toList.reverse   )).sum / 1000 
res60: Long = 12484410L
```

Christewart (approved):

chain/src/main/scala/org/bitcoins/chain/models/BlockHeaderDAO.scala
Christewart: Wow! Good to know. Thanks!

benthecarman (approved):

PR #2667, opened 2021-02-16
Change oracle db to have its own directory

### LINKED ISSUE #2663: oracle server uses regtest datadir
>021-02-15T14:11:02UTC INFO [OracleServerMain] version=0.5.0+17-89185338+20210215-0730-SNAPSHOT
2021-02-15T14:11:02UTC INFO [OracleServerMain] networkDatadir=/home/demiourgos728/.bitcoin-s/regtest
[DEBUG] [02/15/2021 14:11:02.477] [main] [EventStream] StandardOutLogger started

Is this what we want? if we change this in the future (say, we change the default network setting) will users have their oracle break because it isn't looking in the correct datadir?

benthecarman:
Closes #2663

Removed the `dbFlywaySettings(oracle)` from the build config, so it doesn't make a `oracle.sqlite` file in every network dir. Tested and works fine without it.


db-commons/src/main/resources/reference.conf
```diff
@@ -119,6 +119,7 @@ bitcoin-s {
     # this config key is read by Slick
     db {
       name = oracle
+      path = ${bitcoin-s.datadir}/oracle/
```
Christewart, line 122: Copy to example config on site, let's try and keep the example config on the site consistent with our default config in the code base

Christewart (approved):

Christewart (changes requested):
This doesn't migrate existing databases to the new data directory does it? We should add that code.

Christewart (approved):

Issue #2626, opened 2021-02-04
Publish more user friendly artifacts


Christewart:
The sbt native packager plugin gives us a ton of options that we can use to publish packages. For 0.6 we should publish more friendly artifacts.

We should also look into whether github workflows can auto-publish these on every PR merge

https://sbt-native-packager.readthedocs.io/en/latest/

Christewart:
This is still needed for windows

Christewart:
Taken care of for windows in #3210

Christewart:
I think this done

PR #2668, opened 2021-02-16
Get basic docker image working with oracle server

Christewart:
This is part of #2626 

This PR introduces configuration for building a docker image for the oracle server. 

## Building a docker image

You can build the docker image by doing the following inside of `sbt`. [Here](https://sbt-native-packager.readthedocs.io/en/latest/formats/docker.html#tasks) are the list of tasks you can use to build a docker image, we are going to use `stage

> oracleServer/compile;oracleServer/docker:stage

This builds a docker image and places it inside of the directory `app/oracle-server/target/docker/stage`

If you navigate to that directory, you can build the image by doing 

`docker build app/oracle-server/target/docker/stage/`

This will give you an image which you can see by doing 

Screenshot from 2021-02-16 10-55-54

## Running docker image

From the example above, you can see my docker image hash is `95b3d5989078`. We need this for running the image

Now we can start the image to create a running container with the [custom docker configuration](https://github.com/bitcoin-s/bitcoin-s/blob/7ad7a9fbf53f9d0534b009a50345a481ce2a7926/app/oracle-server/src/universal/docker-application.conf) following command

```
docker run -p 9998:9998 95b3d5989078
```
This should give you a log that looks like the following

Screenshot from 2021-02-16 10-58-53

Once this is done, you should be able to send requests to the oracle server with `bitcoin-s-cli` 

> ./bitcoin-s-cli getpublickey
4a4d9bd3e5f66211024599a3c29a0f6a9a3c2b75a90b5e0860da7d8f13fbd33e

Here is a screenshot of the following interaction

Screenshot from 2021-02-16 11-01-20

## Resources 

[Docker overview](https://docs.docker.com/get-started/overview/)
[Sbt native packager docker config](https://sbt-native-packager.readthedocs.io/en/latest/formats/docker.html#settings)

### DIFF
--- a/app/oracle-server/src/main/resources/reference.conf
+++ b/app/oracle-server/src/main/resources/reference.conf
@@ -1,7 +1,7 @@
 bitcoin-s {
     oracle {
         rpcport = 9998
-        rpcbind = "0.0.0.0"
+        rpcbind = "127.0.0.1"
         hikari-logging = true
         hikari-logging-interval = 5 minute
     }
--- a/app/oracle-server/src/main/resources/reference.conf
+++ b/app/oracle-server/src/main/resources/reference.conf
@@ -1,8 +1,9 @@
 bitcoin-s {
     oracle {
         rpcport = 9998
+        rpcbind = "0.0.0.0"
         hikari-logging = true
-        hikari-logging-interval = 1 minute
+        hikari-logging-interval = 5 minute
     }
 }
 
--- a/build.sbt
+++ b/build.sbt
@@ -203,8 +203,7 @@ lazy val secp256k1jni = project
     coverageEnabled := false
   )
 
-/**
-  * If you want sbt projects to depend on each other in
+/** If you want sbt projects to depend on each other in
   * slighly nonstandard ways, the way to do it is through
   * stringly typed syntax.
   *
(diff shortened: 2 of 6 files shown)


app/oracle-server/src/main/resources/reference.conf
```diff
@@ -1,8 +1,9 @@
 bitcoin-s {
     oracle {
         rpcport = 9998
+        rpcbind = "0.0.0.0"
```
Christewart: I think this is potentially a security problem, we don't want to publish on all interfaces unless this is running inside a docker container. So we probably need some sort of docker specific setting?

benthecarman: Yeah this is dangerous, we shouldn't do this

Christewart: It needs to be done inside the container, however we don't want to default it outside of the container.

build.sbt
```diff
@@ -281,10 +280,14 @@ lazy val appCommonsTest = project
 lazy val oracleServer = project
   .in(file("app/oracle-server"))
   .settings(CommonSettings.prodSettings: _*)
+  .settings(
+    Compile / unmanagedResourceDirectories += baseDirectory.value / "src" / "universal"
```
Christewart, line 284: This is where the custom configuration is stored and added to the resources packaged in the artifact.

benthecarman:
Might be nice to add a docs page explaining how to launch from docker or add to the getting started page

app/oracle-server/src/main/resources/reference.conf
Christewart: Ok this should be resolved in 4357b6f

Christewart:
> Might be nice to add a docs page explaining how to launch from docker or add to the getting started page

Ok added in de9d767

Maybe you could try to set it up from the docs, and report back with what is confusing/misleading etc.

benthecarman (approved):

PR #2673, opened 2021-02-17
2021 02 15 appserver docker

Christewart:
This is built ontop of #2668 

This addresses part of #2626 

This adds instructions for building a docker image for `appServer`. You can see the instructions [here](https://github.com/Christewart/bitcoin-s-core/blob/e54cbffb9662ede4ebfdf4935ffaec071570c960/docs/applications/server.md#docker)

### DIFF
--- a/app/server/server.sbt
+++ b/app/server/server.sbt
@@ -13,4 +13,8 @@ packageSummary := "A Bitcoin neutrino node and wallet"
 packageDescription := "Runs a Bitcoin neutrino node and wallet, has functionality " +
   "for many different modes and configuration options, see more at https://bitcoin-s.org/docs/applications/server"
 
-enablePlugins(JavaAppPackaging)
+dockerExposedPorts ++= Seq(9999)
+
+dockerEntrypoint := Seq("/opt/docker/bin/bitcoin-s-server",
+                        "--conf",
+                        "/opt/docker/docker-application.conf")
--- a/docs/applications/server.md
+++ b/docs/applications/server.md
@@ -16,6 +16,7 @@ The server project provides a away to access information from these three projec
 
 ### Building the server
 
+#### Java binary
 You can build the server with the [sbt native packager](https://github.com/sbt/sbt-native-packager).
 The native packager offers [numerous ways to package the project](https://github.com/sbt/sbt-native-packager#examples).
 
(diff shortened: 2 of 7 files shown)


app/server/server.sbt
```diff
@@ -13,4 +13,8 @@ packageSummary := "A Bitcoin neutrino node and wallet"
 packageDescription := "Runs a Bitcoin neutrino node and wallet, has functionality " +
   "for many different modes and configuration options, see more at https://bitcoin-s.org/docs/applications/server"
 
-enablePlugins(JavaAppPackaging)
```
benthecarman, line 16: Is this supposed to be removed?

docs/applications/server.md
```diff
@@ -31,6 +32,42 @@ This will produce a script to execute bitcoin-s which you can start with
 ./app/server/target/universal/stage/bin/bitcoin-s-server
 ```
 
+#### Docker
+
+The oracle server also has docker support. You can build a docker image with the following commands
+
+```
+sbt "appServer/compile;appServer/docker:stage"
```
benthecarman: does this need the compile first, shouldn't `docker:stage` compile it

project/CommonSettings.scala
```diff
@@ -139,6 +142,20 @@ object CommonSettings {
...
 
+  lazy val appSettings: Seq[Setting[_]] = prodSettings ++ Vector(
+    //gives us the 'universal' directory in build artifacts
+    Compile / unmanagedResourceDirectories += baseDirectory.value / "src" / "universal"
+  )
+
+  lazy val dockerSettings: Seq[Setting[_]] = {
+    Vector(
+      //https://sbt-native-packager.readthedocs.io/en/latest/formats/docker.html
+      dockerBaseImage := "openjdk",
```
benthecarman, line 153: Are we able to / should we define a java version >11 so if in the future we are using something with eclair it will work?

app/server/server.sbt
Christewart: It was moved to here: https://github.com/bitcoin-s/bitcoin-s/pull/2673/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R311

project/CommonSettings.scala
Christewart: It seems the `openjdk` policy is to use the latest release of java, so this is using `openjdk-15` as of now (see layer 4)

https://hub.docker.com/layers/138104531/bitcoinscala/bitcoin-s/latest/images/sha256-35e35532d821abc257399136252a8ff20a679fec2545e31c80b5a0f95f02c8e0?context=explore

docs/applications/server.md
Christewart: It's not strictly necessary, i can remove if you desire

benthecarman: imo it'd be cleaner without, but not a huge deal

benthecarman (approved):

benthecarman (approved):

PR #2695, opened 2021-02-19
Fix normalized string comparison

### LINKED ISSUE #2691: DLC Oracle test for enum signing sometimes fails
https://github.com/bitcoin-s/bitcoin-s/pull/2656/checks?check_run_id=1929863942#step:5:789

After doing some testing it seems that it is just the `outcome == tlvOutcomes.head` check that is failing. Likely we are missing somewhere to normalize a string or something

benthecarman:
Fixes #2691

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
@@ -325,6 +325,8 @@ case class NormalizedString(private val str: String) extends NetworkElement {
     other match {
       case otherStr: String =>
         normStr == otherStr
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -325,6 +325,8 @@ case class NormalizedString(private val str: String) extends NetworkElement {
     other match {
       case otherStr: String =>
         normStr == otherStr
+      case otherNorm: NormalizedString =>
```
Christewart, line 328: test case to make sure we don't regress in the future?

Christewart (approved):

PR #2698, opened 2021-02-20
Decrease false positive rate in BloomFilterTest to avoid spurious CI failures

### LINKED ISSUE #2510: Insert public key into bloom filteres test case failing sporadically
https://github.com/bitcoin-s/bitcoin-s/pull/2509/checks?check_run_id=1695428095#step:5:2012

Christewart:
fixes #2510 

This decreases the false positive rate to so that we avoid more false positives on CI.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/bloom/BloomFilterTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/bloom/BloomFilterTest.scala
@@ -88,7 +88,7 @@ class BloomFilterTest extends BitcoinSUnitTest {
     // for 2 keys
     val numElements = 4
     val bloom = BloomFilter(numElements = numElements,
-                            falsePositiveRate = 0.001,
+                            falsePositiveRate = 0.000001,
                             tweak = UInt32(100),
                             BloomUpdateNone)
 


core-test/src/test/scala/org/bitcoins/core/bloom/BloomFilterTest.scala
```diff
@@ -88,7 +88,7 @@ class BloomFilterTest extends BitcoinSUnitTest {
     // for 2 keys
     val numElements = 4
     val bloom = BloomFilter(numElements = numElements,
-                            falsePositiveRate = 0.001,
+                            falsePositiveRate = 0.000001,
```
benthecarman, line 91: Can we not do 0?

benthecarman (approved):

core-test/src/test/scala/org/bitcoins/core/bloom/BloomFilterTest.scala
Christewart: I think that we would have an infinite size bloom filter. There is a direction relationship between the size of a bloom filter and the rate of false positives allowed. If I have an infinite size bloom filter, and insert 1 element into it, my false positive rate when seeing if another element is contained in the filter would be 0.000000000000000000000000000000000000001

Issue #2596, opened 2021-01-30
Wallet.processBlock() takes longer than 30 seconds to complete on rescans on mainnet


Christewart:
You can see the beginning and the end log here. This was observed on  0ba6e3a2ee02f2b291bccbc9483d442497e3249f (#2568) which doesn't touch wallet logic AFAIK. We should look into this in the future. The surprising part is that this was observed on postgres, which insinuates that the sqlite3 performance might be worse. I'll follow up when i test out sqlite3. 

>2021-01-30T15:20:11UTC INFO [Wallet$WalletImpl] Processing block=DoubleSha256BDigestBE(0000000000000000001cf5043bafb14e0ea11cbbfa0519ce9f5016226983b7d5)
2021-01-30T15:20:39UTC INFO [HikariLogging] "wallet-pool-usage":{"max":12.0,"min":0.0,"median":0.0,"mean":1.6407026234330504,"75thPercentile":4.0,"95thPercentile":6.0,"98thPercentile":8.0,"99thPercentile":9.0,"999thPercentile":12.0}
2021-01-30T15:20:39UTC INFO [HikariLogging] "wallet-activity-update":{"active":1,"idle":4,"waiting":0,"total":5,"maxThreads":5,"activeThreads":1,"maxQueueSize":5000,"queueSize":0}
2021-01-30T15:20:39UTC INFO [HikariLogging] "node-pool-usage":{"max":0.0,"min":0.0,"median":0.0,"mean":0.0,"75thPercentile":0.0,"95thPercentile":0.0,"98thPercentile":0.0,"99thPercentile":0.0,"999thPercentile":0.0}
2021-01-30T15:20:39UTC INFO [HikariLogging] "node-activity-update":{"active":0,"idle":5,"waiting":0,"total":5,"maxThreads":5,"activeThreads":0,"maxQueueSize":5000,"queueSize":0}
2021-01-30T15:20:39UTC INFO [HikariLogging] "chain-pool-usage":{"max":5362.0,"min":0.0,"median":80.0,"mean":74.75416859667669,"75thPercentile":94.0,"95thPercentile":102.0,"98thPercentile":107.0,"99thPercentile":185.0,"999thPercentile":592.0}
2021-01-30T15:20:39UTC INFO [HikariLogging] "chain-activity-update":{"active":0,"idle":5,"waiting":0,"total":5,"maxThreads":5,"activeThreads":0,"maxQueueSize":5000,"queueSize":0}
[...]

Issue #1844, opened 2020-08-17
Database configuration defaults


Christewart:
Our default database configurations live inside of [`db.conf`](https://github.com/bitcoin-s/bitcoin-s/blob/26f47d269c9ecb40dd69c498c04b0b1f47c9c0d3/db-commons/src/main/resources/db.conf#L1). This issue describes how to break out the monolithic settings we currently have based on the environment we are currently operating in

### Prod settings

As per conversation with @rorp , when shipping default database configurations for sqlite3 we should have `numThreads=1` in the default `db.conf` as is stated in #1840 . We can also have `numThreads=5` (or whatever appropriate number) for postgres in the release as it is safe to have multiple database connections in postgres. 

Ultimately we need to re-work the `common` settings in db.conf to have two different common configurations, one for `sqlite3` (with `numThreads=1`) and one for `postgres` that allows `numThreads` >= 1. There may be other specific configurations we want for these database drivers in the future that we can put here as well. 

For a list of slick configurations see [`DataBaseFactoryDef.forConfig`](https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database) method

### Test settings 

However @rorp also brought up a good point that for testing purposes, it might be advantageous for us to NOT set `numThreads=1` but to something > 1. This allows us to see spurious CI failures for `SQLITE_BUSY` where we may have unwanted parallel behavior that is not logically correct. Inside of `db.conf` we can also setup a `sqlite3-test` config and a `postgres` test config. 

### Skeleton 

```
sqlite3 = {
  dataSourceClass = slick.jdbc.DatabaseUrlDataSource
  profile = "slick.jdbc.SQLiteProfile$"

  db {
    //for information on parameters available here see
    //https://scala-slick.org/doc/3.3.1/api/index.html#slick.jdbc.JdbcBackend$DatabaseFactoryDef@forConfig(String,Config,Driver,ClassLoader):Database
    path = ${bitcoin-s.datadir}/${bitcoin-s.network}/
    driver = org.sqlite.JDBC
    username = ""
    password = ""

    # this needs to be set to 1 for SQLITE as it does not support concurrent database operations
    # see: https://github.com/bitcoin-s/bitcoin-s/pull/1840
    numThreads = 1
    #maxConnections = 1
    # as long as we're on SQLite there's no point
    # in doing connection pooling
[...]

PR #2705, opened 2021-02-21
Reintroduce parallelism in transaction processing

Christewart:
This re-introduces parallelism in the wallet to begin addressing #2596 

This was reverted in #1823 erroneously (we did not understand that we can set the sqlite database thread pool to size=1 to get our intended behavior, which we did in https://github.com/bitcoin-s/bitcoin-s/pull/2113/files#diff-d1889eaa42a9c55c8f1783fc367505bd504a12742fe0f54d66ae528cce57ef20R18).

More discussion on database thread pool size and parallelism can be found here: https://github.com/bitcoin-s/bitcoin-s/issues/1844

It takes roughly 30 seconds to process a block now. This PR does "the dumb thing" and re-introduces parallelism for processing transaction inputs and outputs for a single tx. AFAIK there is no reason why we can't process these in parallel as outputs for a transaction can't be spent as inputs on _the same transaction_. 

This will not improve performance significantly in sqlite as there is a single database thread that controls access to a sqlite database. However, on my sophisticated database like postgres we can leverage multiple connections to make parallel requests to the db.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/number/NumberCache.scala
+++ b/core/src/main/scala/org/bitcoins/core/number/NumberCache.scala
@@ -46,6 +46,11 @@ trait NumberCacheBigInt[T] extends NumberCache[T] {
   /** The max number cached (inclusive) */
   def maxCachedBigInt: BigInt = BigInt(maxCached)
 
+  /** [[org.bitcoins.core.protocol.CompactSizeUInt]] uses a UInt64
+    * which means we have larger uint64s used on a regular basis
+    */
+  override def maxCached: Long = 2048
+
   /** Checks if the given number is cached
     * if not, allocates a new object to represent the number
     */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/number/NumberCache.scala
```diff
@@ -46,6 +46,11 @@ trait NumberCacheBigInt[T] extends NumberCache[T] {
   /** The max number cached (inclusive) */
   def maxCachedBigInt: BigInt = BigInt(maxCached)
 
+  /** [[org.bitcoins.core.protocol.CompactSizeUInt]] uses a UInt64
```
Christewart, line 49: Truthfully i'm still guessing with this number. It seems like most things with scriptSigs should be less than this size?

benthecarman (approved):

PR #2707, opened 2021-02-21
2021 02 21 cheap redeemscript check

Christewart:
Introduce even more cheap checks for parsing scripts.

One of the hottest code paths when parsing a block is `P2SHScriptSignature.isRedeemScript()`.  This adds a few more checks around it to see if we can "fail fast" if it isn't a redeem script.

This also does the same thing for `WitnessScriptPubKey.isValidAsm()`. Basically we should check if the first `ScriptToken` is a valid witness version (`OP_0`,`OP_1`, ... , `OP_16`) before we deserialize the asm into a byte vector. This allows us to fail fast using the parsed asm we already have.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
@@ -254,33 +254,41 @@ object P2SHScriptSignature extends ScriptFactory[P2SHScriptSignature] {
 
   /** Detects if the given script token is a redeem script */
   def isRedeemScript(token: ScriptToken): Boolean = {
-    val redeemScript: ScriptPubKey = parseRedeemScript(token)
-
-    def isStandardNonP2SH(
-        spk: ScriptPubKey,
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ScriptSignature.scala
```diff
@@ -254,33 +254,41 @@ object P2SHScriptSignature extends ScriptFactory[P2SHScriptSignature] {
...
+        //if the token is OP_0/OP_1/OP_CHECKSIG etc
+        false
+      case constant: ScriptConstant =>
+        val redeemScript: ScriptPubKey = parseRedeemScript(constant)
+        isStandardNonP2SH(redeemScript, isRecursiveCall = false)
     }
+  }
 
-    isStandardNonP2SH(redeemScript, isRecursiveCall = false)
+  private def isStandardNonP2SH(
```
Christewart, line 268: move only, we should start avoiding nested methods in the future as it makes it harder to refactor things

benthecarman (approved):

Issue #2708, opened 2021-02-22
docker hub versioning not working


Christewart:
https://hub.docker.com/r/bitcoinscala/bitcoin-s-oracle-server/tags?page=1&ordering=last_updated

I thought this would be avoided in #2694 but it looks like I was wrong. 

If you scroll further down, it seems like it _was_ working for awhile, but broke for some reason over the weekend? :man_shrugging: 
![Screenshot from 2021-02-22 06-50-00](https://user-images.githubusercontent.com/3514957/108710848-3244e300-74da-11eb-9b7e-08ef091a5071.png)

Christewart:
It appears the default setting if `fetch-depth: [not-zero]` is used is to _not_ fetch tags

https://github.com/actions/checkout/issues/448

You can see this here on our runner actions, i think the best workaround for now is dealing with slow build times so we can get the tag.

https://github.com/bitcoin-s/bitcoin-s/runs/1947728258#step:2:40

Christewart:
fixed in #2709 , for now

Issue #2675, opened 2021-02-17
Investigate why git clone takes forever


Christewart:
We added a workaround in #2674 , but we should investigate the root cause for this in our repo

Christewart:
superseded by #2830

PR #2709, opened 2021-02-22
Workaround for issue 2708

Christewart:
Workaround until #2708 can be fixed. This will mean _much_ slower build times on our merge github actions as we now have to clone the entire repo. See #2675


benthecarman (approved):
Sucks we have to do this :/

PR #2711, opened 2021-02-22
Wallet Rebroadcast Transactions thread

benthecarman:
This makes it so the wallet will rebroadcast transaction that have yet to be confirmed. This should help with UX for the wallet where the user could need to manually rebroadcast a transaction.

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -14,6 +14,7 @@ import org.bitcoins.core.p2p.CompactFilterMessage
 import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.core.protocol.blockchain.BlockHeader
 import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.crypto.{
   DoubleSha256Digest,
@@ -118,8 +119,9 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
 
   // Node Api
 
-  override def broadcastTransaction(transaction: Transaction): Future[Unit] =
-    sendRawTransaction(transaction).map(_ => ())
+  override def broadcastTransactions(
+      transactions: Vector[Transaction]): Future[Unit] =
+    FutureUtil.sequentially(transactions)(sendRawTransaction(_)).map(_ => ())
 
   override def downloadBlocks(
       blockHashes: Vector[DoubleSha256Digest]): Future[Unit] = Future.unit
--- a/docs/config/configuration.md
+++ b/docs/config/configuration.md
@@ -1,24 +1,21 @@
 ---
-id: configuration
-title: Application Configuration
+id: configuration title: Application Configuration
 ---
 
 Bitcoin-S uses [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md)
-to configure various parts of the application the library offers. HOCON is a
-superset of JSON, that is, all valid JSON is valid HOCON.
+to configure various parts of the application the library offers. HOCON is a superset of JSON, that is, all valid JSON
+is valid HOCON.
 
 All configuration for Bitcoin-S is under the `bitcoin-s` key.
 
-If you have a file `application.conf` anywhere on your classpath when using
-bitcoin-s, the values there take precedence over the ones found in our
-`reference.conf`. We also look for the file `bitcoin-s.conf` in the current
-Bitcoin-S data directory.
+If you have a file `application.conf` anywhere on your classpath when using bitcoin-s, the values there take precedence
+over the ones found in our
+`reference.conf`. We also look for the file `bitcoin-s.conf` in the current Bitcoin-S data directory.
 
 The resolved configuration gets parsed by
 [`AppConfig`](/api/org/bitcoins/db/AppConfig).
-`AppConfig` is an abstract class that's implemented by corresponding case
-classes in the `wallet`, `chain` and `node` projects. Here's some examples of how to
-construct a wallet configuration:
+`AppConfig` is an abstract class that's implemented by corresponding case classes in the `wallet`, `chain` and `node`
+projects. Here's some examples of how to construct a wallet configuration:
 
 ```scala mdoc:compile-only
 import org.bitcoins.wallet.config.WalletAppConfig
@@ -40,8 +37,8 @@ val customOverride = ConfigFactory.parseString("bitcoin-s.network = testnet3")
 val configFromCustomDirAndOverride = WalletAppConfig(customDirectory, customOverride)
 ```
 
-You can pass as many `com.typesafe.config.Config`s as you'd like. If any
-keys appear multiple times the last one encountered takes precedence.
+You can pass as many `com.typesafe.config.Config`s as you'd like. If any keys appear multiple times the last one
+encountered takes precedence.
 
 ## Command Line Options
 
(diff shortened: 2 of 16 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -24,6 +25,41 @@ import scala.util.{Failure, Success, Try}
...
+        // Wrap in try in case of spurious failure
+        Try(Await.result(f, 60.seconds))
+
+        // wait 30 minutes
+        Thread.sleep(30.minutes.toSeconds)
+      }
+    }
+  }
+
+  private lazy val rebroadcastTxsThread = new Thread(
```
Christewart: So instead of allocating a new thread, i think we should just create a wallet scheduler just like we have with hikari

https://github.com/bitcoin-s/bitcoin-s/blob/5b762fb2e13742f057162e14a035b67730053ca0/db-commons/src/main/scala/org/bitcoins/db/HikariLogging.scala#L181

We may even be able to re-use this scheduler for the address queue stuff as well? 

I guess I don't want to keep allocating single threads for various functions in the wallet when realistically all these things could share the same background thread. Thoughts?

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -118,8 +119,9 @@ class BitcoindRpcClient(val instance: BitcoindInstance)(implicit
 
   // Node Api
 
-  override def broadcastTransaction(transaction: Transaction): Future[Unit] =
-    sendRawTransaction(transaction).map(_ => ())
+  override def broadcastTransactions(
+      transactions: Vector[Transaction]): Future[Unit] =
+    FutureUtil.sequentially(transactions)(sendRawTransaction(_)).map(_ => ())
```
Christewart, line 124: nit: I don't see why we can't broadcast these in parallel? Is there some sort of rate limit with bitcoin core?

Christewart (changes requested):
I coded up the patch to use the scheduler for the address queue as well, but this does require the scheduler thread pool size to be 1

https://github.com/Christewart/bitcoin-s-core/commit/bab9b6296fbdffdb3c0af029f0db98d4103fedca

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
benthecarman: The RPC has a max of 16 requests at at time, then it stops dropping them

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -24,6 +26,56 @@ import scala.util.{Failure, Success, Try}
...
   self: Wallet =>
+
+  private case object RebroadcastTransactionsRunnable extends Runnable {
+
+    override def run(): Unit = {
+      val f = for {
+        txs <- getTransactionsToBroadcast
+
+        _ = {
+          if (txs.size > 1)
```
Christewart, line 38: I don't really get why this is here? What does it matter if the sizes are different, aren't we logging the same thing?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -482,4 +534,12 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
         } yield utxos
     }
   }
+
+  private[wallet] def getTransactionsToBroadcast: Future[
+    Vector[Transaction]] = {
+    for {
+      unconfirmedUtxos <- spendingInfoDAO.findAllPendingConfirmation
```
Christewart: So this doesn't really make sense to me, why would we want to re-broadcast transactions that could possibly have 1 confirmation? 

Do we need to add a new `TxoState` that represents a broadcast but not confirmed _at all_ state? 

It seems our TxoState's currently can't represent what is required here?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -24,6 +26,56 @@ import scala.util.{Failure, Success, Try}
...
+    }
+  }
+
+  private[this] var rebroadcastTransactionsCancelOpt: Option[
+    ScheduledFuture[_]] = None
+
+  /** Starts the wallet's rebroadcast transaction scheduler */
+  def startRebroadcastTxsScheduler(): Unit = {
+    val future = scheduler.scheduleAtFixedRate(RebroadcastTransactionsRunnable,
+                                               30,
```
Christewart: nit: Probably make these into configuration options for the wallet. 

Something like 

`bitcoin-s.wallet.rebroadcast-frequency = 30 minutes` 

?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -24,6 +26,56 @@ import scala.util.{Failure, Success, Try}
...
+                                               30,
+                                               TimeUnit.MINUTES)
+    rebroadcastTransactionsCancelOpt = Some(future)
+  }
+
+  /** Kills the wallet's rebroadcast transaction scheduler */
+  def stopRebroadcastTxsScheduler(): Unit = {
```
Christewart: May want to make this private[wallet]? I don't see why we would want to allow somone outside the wallet project the ability to kill the rebroadcaster

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -482,4 +534,12 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
         } yield utxos
     }
   }
+
+  private[wallet] def getTransactionsToBroadcast: Future[
```
Christewart, line 565: Unit tests needed

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -213,39 +213,35 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
 
   /** Broadcasts the given transaction over the P2P network */
-  override def broadcastTransaction(transaction: Transaction): Future[Unit] = {
-    val broadcastTx = BroadcastAbleTransaction(transaction)
+  override def broadcastTransactions(
+      transactions: Vector[Transaction]): Future[Unit] = {
+    val broadcastTxDbs = transactions.map(tx => BroadcastAbleTransaction(tx))
 
-    val addToDbF = txDAO.upsert(broadcastTx)
+    val addToDbF = txDAO.upsertAll(broadcastTxDbs)
```
Christewart, line 220: Hmm... so it seems like something funny is going here: 

Whose responsibility is to re-broadcast transactions? 

It seems that the wallet should tell the node that a transaction needs to be re-broadcast, but the node actually executes it. 

It seems like the node project should have an API that allows the wallet to tell it when 

1. We need to broadcast a transaction
2. We don't need to broadcast this transaction anymore (presumably because it is confirmed)

Now the node project itself should have all details about how often to broadcast, what peer to broadcast to (in the future), etc. 

I think the biggest limitation in the node project right now is we have no ability to do (2). 

Thoughts?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: Just for the plural vs singular

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: Ah yes, seems so

node/src/main/scala/org/bitcoins/node/Node.scala
benthecarman: This is where things get complicated. If we do this, then the node's db needs to be chain aware and the node will need to process incoming blocks and filters. imo it makes more sense for the wallet to handle all of this.

benthecarman: AFIAK this is how core does it as well.

Christewart: So here are my thoughts: 

1. A new transaction originated from the wallet should be broadcast via a wallet callback
2. A wallet transaction confirmed (or pending confirmed, if we introduce a new TxoState) should also have a wallet callback. This callback can then _delist_ the transaction to be broadcast in the node project since it is confirmed.

docs/config/configuration.md
```diff
@@ -202,6 +203,9 @@ bitcoin-s {
         # before we timeout
         addressQueueTimeout = 5 seconds
         
+        # How often the wallet will rebroadcast unconfirmed transactions
+        rebroadcastFrequency = 30 minutes
```
Christewart: This seems like an aggressive default setting. Maybe like 4 hour re-tries? It ends up being 6 times a day? Frankly if it isn't getting confirmed, it's probably too low of a fee, which requires user interaction to bump. Re-trying the broadcast isn't going to get it confirmed.

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
```diff
@@ -450,7 +450,7 @@ private[wallet] trait AddressHandling extends WalletLogger {
     * With this background thread, we poll the [[addressRequestQueue]] seeing if there
     * are any elements in it, if there are, we process them and complete the Promise in the queue.
     */
-  lazy val walletThread = new Thread(AddressQueueRunnable)
+  lazy val addressQueueThread = new Thread(AddressQueueRunnable)
```
Christewart, line 450: Didn't we talk about getting rid of these and just using the scheduler?

benthecarman: woops, added

docs/config/configuration.md
benthecarman: Changed to 4 hours

wallet/src/main/scala/org/bitcoins/wallet/internal/AddressHandling.scala
Christewart: So this still isn't solving the problem (you still allocate a `new Thread(AddressQueueRunnable)`)

I tried fixing this for you, but it appears we have design/architecture decisions that prevent us from doing this cleanly. Every _instance_ of a wallet class has it's own scheduler, which means we need to ensure that `wallet.start()` is called, else the scheduler doesn't have the tasks scheduled on it. 

I tried to make these changes with `AddressQueueRunnable` last night/this morning, but it is tough to do since we have a lot of methods that return a new `Wallet` instance (which is the right thing to do!, we want to maintain referential transparency!) 

When a new wallet instance is returned, we need to make sure that instance has had the `.start()` method called on it. If it doesn't, we won't have the re-broadcast transaction task scheduled on the scheduler, and we don't have the address queue runnable scheduled either. 

If the `AddressQueueRunnable` isn't scheduled, it is super obvious things aren't working as literally you won't be able to fetch a bitcoin address, but the re-broadcast logic is much more subtle since it doesn't occur that often. 


I attempted to hack together something working for the `AddressQueueRunnable`, but am giving up for now as it is getting to be too time consuming and we have higher priorities

https://github.com/Christewart/bitcoin-s-core/tree/2021-03-23-rm-addressthread-queue

Christewart (changes requested):
I still don't see any test cases that end-to-end test this logic with a successful transaction. Where does the tx get removed from the broadcast table when it has been confirmed in the blockchain?

Also this comment: 

https://github.com/bitcoin-s/bitcoin-s/pull/2711/files#r599585752

benthecarman:
Relevant PR from core on the re-broadcast logic they are using

https://github.com/bitcoin/bitcoin/pull/21061

Christewart (approved):

PR #2713, opened 2021-02-23
Upgrade scalac to 2.13.5

Christewart:
https://github.com/scala/scala/releases/tag/v2.13.5

### DIFF
--- a/.github/workflows/Compile.yml
+++ b/.github/workflows/Compile.yml
@@ -23,4 +23,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: Compile and Check Formatting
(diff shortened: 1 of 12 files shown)


.github/workflows/Compile.yml
```diff
@@ -23,4 +23,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: Compile and Check Formatting
-        run: sbt ++2.12.12 test:compile scalafmtCheckAll
+        run: sbt +test:compile scalafmtCheckAll
```
benthecarman, line 26: I think it's best to enforce 2.12 for this since most of us are using 2.13 locally

Christewart: The `+` compiles for both versions

benthecarman (approved):

PR #2715, opened 2021-02-23
Use Sign instead of ECPrivateKey in DLCTxSigner

benthecarman:
This is done so we can once again make the root key in the keymanager private.

The changes to `Sign` and the like should probably be added to #2652

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
@@ -244,6 +244,12 @@ sealed abstract class ExtPrivateKey
     key.signWithEntropyFunction
   }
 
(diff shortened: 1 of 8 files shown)


nkohen (changes requested):
I looked through this briefly and I have two main (structural) comments.

Firstly, I think that rather than adding `adaptorSign` to the `Sign` interface there should be a new `trait AdaptorSign extends Sign` which has this new function as I don't think we can expect all key managing things to support this function in the way we can expect them to support ECDSA.

Secondly, and this is basically a nit, but I don't think you should introduce a `adaptorSignFunction` and instead just have the `def`, this is the uniform way things are now made (once #2652 is merged).

nkohen (approved):

PR #2719, opened 2021-02-24
Refactor crypto module to be compatible with Scala.js part 1

rorp:
This PR moves all crypto library specific code into implementations of `CryptoRuntime` trait.
No changes in implementation, just reshuffling and cleaning up the code.


crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
```diff
@@ -136,4 +139,77 @@ trait CryptoRuntime {
   def sha256SchnorrNonce(bytes: ByteVector): Sha256Digest = {
     sha256(schnorrNonceTagBytes ++ bytes)
   }
+
+  def publicKey(privateKey: ECPrivateKey): ECPublicKey
```
Christewart, line 133: Ok so these are all required to be part of the `CryptoRuntime` interface? 

It seems we just need to start breaking this PR into smaller PRs that make necessary adjustments. 

This seems like it would be a good starting point.

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
```diff
@@ -136,4 +139,77 @@ trait CryptoRuntime {
...
+
+  def add(bytes: ByteVector, pk2: ECPrivateKey): ByteVector
+
+  def add(pk1: ECPublicKey, pk2: ECPublicKey): ECPublicKey
+
+  def pubKeyTweakAdd(pubkey: ECPublicKey, privkey: ECPrivateKey): ECPublicKey
+
+  def isValidPubKey(bytes: ByteVector): Boolean
+
+  def isFullyValidWithBouncyCastle(bytes: ByteVector): Boolean
```
Christewart, line 163: there is going not going to be able to reside here as it is bouncy castle specific

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
```diff
@@ -0,0 +1,290 @@
...
+  KeyParameter
+}
+import org.bouncycastle.math.ec.ECPoint
+import scodec.bits.ByteVector
+
+import java.math.BigInteger
+import java.security.{MessageDigest, SecureRandom}
+import scala.util.{Failure, Success, Try}
+
+trait BouncycastleCryptoRuntime extends CryptoRuntime {
```
Christewart, line 22: We should not have separate runtimes for `BouncyCastle` and `Libsecp256k1`. Libsecp256k1 does _not_ offer all the functionality required to implement `CryptoRuntime`. 

A simple example is libsecp256k1 doesn't offer a `sha256()` hash function, which is why you are using the built in jvm versions here

https://github.com/bitcoin-s/bitcoin-s/pull/2719/files#diff-78da69e02c71190c88599c3535bba36dc6690f6fb6627f2eb60403a11537effeR72

Christewart: We need to use BouncyCastle and Libsecp256k1 on conjunction with each other to get a full formed `JvmCryptoRuntime`.

Bouncy castle _might_ have all the primitives available to implement the full `CryptoRuntime` interface, but even then, we use `SecureRandom` which is jvm specific (not in bouncy castle) for `freshPrivateKey()` 

https://github.com/bitcoin-s/bitcoin-s/pull/2719/files#diff-50317b401c86cfe020be52556b9deb61e20194f4602dbb4c55bf9b0f27b2eba6R27

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
rorp: Yes. Everithing that calls libsecp or bouncycastle must be in `CryptoRuntime`.

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
rorp: It must be implemented for all `CryptoRuntime` anyway.

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
rorp: No, we should have separate runtimes for BouncyCastle and Libsecp256k1. It much cleaner and maintainable, and
all  functions  not implemented in libsecp are delegated to bouncycastle.

Christewart: Yeah, but like you say, it _isn't_ `libsecp256k1`. For instance, if we were ever to enable the "native" platform for sbt cross project we are going to have to have this exact same problem come up (bouncy castle isn't native code, while libsecp256k1 is).

Let's just do the dumb thing for now and call it `JvmCryptoRuntime`

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -243,7 +168,9 @@ object AdaptorStuff {
       adaptorPoint: ECPublicKey,
       dataToSign: ByteVector): ECAdaptorSignature = {
     // Include dataToSign and adaptor in nonce derivation
-    val hash = CryptoUtil.sha256(dataToSign ++ serializePoint(adaptorPoint))
+    val hash =
+      CryptoContext.cryptoRuntime.sha256(
```
nkohen: This and the diffs below can be reverted right?

rorp: Why?

crypto-test/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -49,44 +46,51 @@ class BouncyCastleSecp256k1Test extends BitcoinSUnitTest {
...
-                                    NumberGenerator.bytevector(33))
+    val keyOrGarbageGen =
+      Gen.oneOf(CryptoGenerators.publicKey.map(_.bytes),
+                NumberGenerator.bytevector(33))
     forAll(keyOrGarbageGen) { bytes =>
       assert(
-        ECPublicKey.isFullyValid(bytes, context = BouncyCastle) ==
-          ECPublicKey.isFullyValid(bytes, context = LibSecp256k1)
+//        ECPublicKey.isFullyValid(bytes, context = BouncyCastle) ==
+//          ECPublicKey.isFullyValid(bytes, context = LibSecp256k1)
```
nkohen: ping

crypto/src/main/scala/org/bitcoins/crypto/SchnorrNonce.scala
```diff
@@ -26,10 +26,11 @@ object SchnorrNonce extends Factory[SchnorrNonce] {
       auxRand: ByteVector): ECPrivateKey = {
     val privKeyForUse = privKey.schnorrKey
 
-    val randHash = CryptoUtil.sha256SchnorrAuxRand(auxRand).bytes
+    val randHash =
+      CryptoContext.cryptoRuntime.sha256SchnorrAuxRand(auxRand).bytes
```
nkohen: These and all `CryptoUtil` deletes can be reverted I think

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
nkohen: probably deserves a rename

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: `CryptoUtil.sha256` is fine no? Isn't it shorthand for `CryptoContext.cryptoRuntime.sha256`?

nkohen:
This hash function will also need to be pulled up into the runtime: https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala#L51

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -0,0 +1,232 @@
...
+      adaptorSignature: ECAdaptorSignature,
+      key: ECPublicKey,
+      msg: ByteVector,
+      adaptorPoint: ECPublicKey): Boolean = ???
+}
+
+object BCryptoCryptoRuntime extends BCryptoCryptoRuntime
+
+/*
+class JsCryptoRuntime extends CryptoRuntime {
```
Christewart: So why are all these commented out? There was working code inside of this previously. 

For instance, `ripemd160()`

https://github.com/bitcoin-s/bitcoin-s/pull/2719/files#diff-6952fb2433325cf12e395c8f8ed56815f78b3258b37f0a0f5c567cf79b534cc5R165

I'm ok with keeping `???` that are not actually implemented (for example `extractAdaptorSecret`) but see no purpose if leaving others things that we have working code for stubbed out?

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
Christewart: I'm with nadav on this, why are you using `CryptoContext.cryptoRuntime` directly when you can just use `CryptoUtil`?

crypto/.jvm/src/main/scala/org/bitcoins/crypto/CryptoRuntimeFactory.scala
```diff
@@ -0,0 +1,24 @@
+package org.bitcoins.crypto
+
+import org.bitcoin.Secp256k1Context
+
+sealed protected trait CryptoRuntimeFactory {
```
Christewart, line 5: I don't see why this is needed, why is it in the `.jvm` context? 

If it truly is needed, we should call it `JvmCryptoRuntimeFactory` (you can't have a scalajs crypto runtime here)

crypto/.jvm/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
```diff
@@ -0,0 +1,197 @@
...
+import java.math.BigInteger
+
+/** Utility cryptographic functions
+  * This is a proxy for the underlying implementation of [[CryptoRuntime]]
+  * such as [[LibSecp256k1CryptoRuntime]].
+  *
+  * This is necessary so that the core module doesn't need to be refactored
+  * to add support for multiple platforms, it can keep referencing CryptoUtil
+  */
+trait CryptoUtil extends CryptoRuntime {
```
Christewart: If we are going to want `core` to work, this needs to be moved out of the `.jvm/`

My understanding with this PR is that we aren't actually trying to get scalajs to work, just start introducing changes that need to be made to make scalajs work in the future.

crypto/.jvm/src/main/scala/org/bitcoins/crypto/CryptoUtil.scala
```diff
@@ -0,0 +1,197 @@
...
+    */
+  override def hmac512(key: ByteVector, data: ByteVector): ByteVector = {
+    cryptoRuntime.hmac512(key, data)
+  }
+
+  /** @param x x coordinate
+    * @return a tuple (p1, p2) where p1 and p2 are points on the curve and p1.x = p2.x = x
+    *         p1.y is even, p2.y is odd
+    */
+  def recoverPoint(x: BigInteger): (ECPoint, ECPoint) = {
```
Christewart: does this need to be in the `.jvm/` directory because we are using `ECPoint`? 

If that is the case, I say we omit any method from the `CryptoRuntime` trait that requires bouncy castle, and re-add it in a later PR.

crypto/crypto.sbt
```diff
@@ -1,7 +1,7 @@
 name := "bitcoin-s-crypto"
 
-libraryDependencies ++= Deps.crypto
+//libraryDependencies ++= Deps.crypto
```
Christewart: can be  removed I believe?

Do we need to add CommonSettings.prodSettings here

https://github.com/bitcoin-s/bitcoin-s/pull/2719/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R43

core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -202,15 +201,7 @@ sealed abstract class ExtPrivateKey
...
-    val tweak = CryptoContext.default match {
-      case CryptoContext.LibSecp256k1 =>
-        val tweakByteArr =
-          NativeSecp256k1.privKeyTweakAdd(il.toArray, key.bytes.toArray)
-        ByteVector(tweakByteArr)
-      case CryptoContext.BouncyCastle =>
-        val sum = key.fieldElement.add(FieldElement(il))
-        sum.bytes
-    }
+    val tweak = CryptoUtil.add(il, key)
```
Christewart, line 204: argh, sorry for missing these on #2658 :man_facepalming:

Christewart (changes requested):

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
rorp: `JvmCryptoRuntime` is even more misleading, because both JVM implementation of `CryptoRuntime` are for JVM.
I can rename them into something like `NativeAndBouncyCaslteCryptoRuntime` and `PureBouncyCaslteCryptoRuntime` if it makes sense.

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
rorp: > So why are all these commented out? There was working code inside of this previously.
Because it's out of the scope of this PR.

crypto/.jvm/src/main/scala/org/bitcoins/crypto/CryptoRuntimeFactory.scala
rorp: Should I rename `CryptoRuntimeFactory` in `.js` context as well? How do I pick the right factory class if `src` context, then? It's not going to work this way.

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
Christewart: I would vote for `BouncyCastleCryptoRuntime` & `NativeAndBouncyCastleCryptoRuntime` with a scaladoc on the latter saying it uses libsecp256k1 for certain operations.

build.sbt
```diff
@@ -30,8 +31,39 @@ lazy val benchSettings: Seq[Def.SettingsDefinition] = {
...
+  .jvmSettings(
+    libraryDependencies ++= Seq(
+      "org.scala-js" %% "scalajs-stubs" % "1.0.0" % "provided",
+      "org.bouncycastle" % "bcprov-jdk15on" % V.bouncyCastle withSources () withJavadoc ()
+    )
+  )
+  .jsSettings(commonJsSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "org.scodec" %%% "scodec-bits" % V.scodecV withSources () withJavadoc ()
```
nkohen, line 57: nit: might be cleaner to define these deps in `Deps`

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
nkohen: I'm personally in favor of it staying as it is now but with better docs.
`Libsecp256k1CryptoRuntime` is just a `CryptoRuntime` that defaults to secp when possible.
`BouncyCastleCryptoRuntime` is just a `CryptoRuntime` that defaults to bouncy castle when possible.
When picking a `CryptoRuntime` you are picking a default library, the fact that the library isn't always used is a detail that should be documented. Mixing things in names makes it more confusing because it isn't any more clear from the name what will be used.

nkohen (approved):

Christewart (approved):

PR #2726, opened 2021-02-25
Add testkit-core module

### LINKED ISSUE #2313: Split Generators from Testkit project
Currently, if you want to use the any of the bitcoin-s generators, you must depend on `testkit`. However, this can be annoying for projects as `testkit` depends on almost every module. It looks like all of our generators only depend on the `core` module, so it should be possible to make a generator module that is separate from `testkit`.

Christewart:
This is built on #2725 

fixes #2313

The core concept of this PR is to breakup the `testkit` project into two projects. 

1. `testkit-core`
2. `testkit`

The `testkit-core` is intended to be used for both the `jvm` and `scalajs` platform. [It has fewer dependencies associated with it that -- I believe -- will work with the scalajs platform.](https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2021-02-25-testkit-core?expand=1#diff-62c45d2ad3c70d80d500448f365f1211f1ff9bf69999c50971ccc3b814c86930R388)

Now that we are using this, we _should_ :tm: be able to re-use the test suite for `cryptoTest` and `coreTest` for the scalajs platform that is being introduced in #2719 

The diff is very big because a lot of the `coreTest` project was using `BitcoinSAsyncTest`. 

We can't use `BitcoinSAsyncTest` because it uses akka, which does not at this time provide scalajs support. I had to create a new test trait that i've called [`BitcoinSJvmTest`](https://github.com/christewart/bitcoin-s-core/blob/fcc98bdd0dc00349a9e73a02d8f7eeecaadf4adf/testkit-core/src/main/scala/org/bitcoins/testkitcore/util/BaseAsyncTest.scala#L294) (I don't like the name, but just wanted to get it working). 

`BitcoinSJvmTest` is still an asynchronous test suite, but it uses Scala's default execution context rather than using akka's `dispatcher`. 

As a side note, this seems to speed up compile time of `bitcoin-s` by ~25% . The two biggest bottlenecks for the compiler are the `core` module and the `testkit` module. Now that we have decoupled `testkit-core` and `testkit` we can begin compiling the `cryptoTest` and `coreTest` while `testkit` is still compiling.


app-commons-test/src/test/scala/org/bitcoins/commons/DLCStatusTest.scala
```diff
@@ -1,13 +1,14 @@
-package org.bitcoins.core.protocol.dlc
+package org.bitcoins.commons
```
Christewart, line 1: This has to be moved because of the use of picklers which are defined in `app-commons`

build.sbt
```diff
@@ -542,6 +543,13 @@ lazy val nodeTest =
     )
     .enablePlugins(FlywayPlugin)
 
+lazy val testkitCore = project
+  .in(file("testkit-core"))
+  .settings(CommonSettings.prodSettings: _*)
+  .settings(name := "bitcoin-s-testkit-core",
+            libraryDependencies ++= Deps.testkitCore.value)
+  .dependsOn(asyncUtils, core, cryptoJVM)
```
benthecarman: should this be `core % testAndCompile`

testkit-core/src/main/scala/org/bitcoins/testkitcore/node/NodeMessageTestUtil.scala
```diff
@@ -0,0 +1,62 @@
+package org.bitcoins.testkitcore.node
+
+import org.bitcoins.core.bloom.{BloomFilter, BloomUpdateAll}
+import org.bitcoins.core.p2p.{GetHeadersMessage, NetworkMessage, VersionMessage}
+import org.bitcoins.core.protocol.blockchain.BlockHeader
+import org.bitcoins.core.protocol.transaction.Transaction
+
+abstract class NodeMessageTestUtil {
```
benthecarman: nit: `P2PMessageTestUtil` may be a better name

benthecarman (approved):

benthecarman (approved):

nkohen (approved):

PR #2731, opened 2021-02-28
Resturcutre cryptoTest & coreTest to work with scalajs build

Christewart:
Built ontop of #2726 . This is the reason the diff is so big.

This PR makes the following projects into sbt crossprojects that allows for scalajs 

1. `cryptoTest`
2. `coreTest`

The idea is that we can re-use most of our test suites to test both our js and jvm implementations of the test suites.

No scalajs functionality works on this PR, and it is not intended to. I want to leave that out of scope of this PR. If you run a scalajs test suite you will get errors that look like this

```
sbt:bitcoin-s> cryptoTestJS/test
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Fast optimizing /home/chris/dev/bitcoin-s/crypto-test/.js/target/scala-2.13/bitcoin-s-crypto-test-test-fastopt
[info] Run completed in 435 milliseconds.
[info] Total number of tests run: 0
[info] Suites: completed 0, aborted 0
[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
[info] No tests were executed.
[error] Error during tests:
[error]         org.bitcoins.crypto.SchnorrDigitalSignatureTest
[error]         org.bitcoins.crypto.FieldElementTest
[error]         org.bitcoins.crypto.SchnorrPublicKeyTest
[error]         org.bitcoins.crypto.HashDigestTest
[error]         org.bitcoins.crypto.SchnorrNonceTest
[error]         org.bitcoins.crypto.AesCryptTest
[error]         org.bitcoins.crypto.ECDigitalSignatureTest
[error]         org.bitcoins.crypto.ECPublicKeyTest
[error]         org.bitcoins.crypto.SignTest
[error]         org.bitcoins.crypto.CryptoUtilTest
[error]         org.bitcoins.crypto.DERSignatureUtilTest
[error]         org.bitcoins.crypto.ECPrivateKeyTest
[error] (cryptoTestJS / Test / test) sbt.TestsFailedException:
```

however you should be able to run all JVM specific tests like `cryptoTestJVM/test` or `coreTestJVM/test`


build.sbt
```diff
@@ -282,41 +312,46 @@ lazy val secp256k1jni = project
...
   .settings(CommonSettings.testSettings: _*)
+  .jsSettings(commonJsSettings)
   .settings(
     name := "bitcoin-s-crypto-test",
     libraryDependencies ++= Deps.cryptoTest.value
   )
   .dependsOn(
-    cryptoJVM % testAndCompile,
-    testkit
+    core,
```
rorp: I see here a chicken-or-egg issue here. The test scope should be limited to `crypto` here

build.sbt
```diff
@@ -58,39 +58,66 @@ lazy val cryptoJS = crypto.js
...
+  .crossType(CrossType.Pure)
+  .settings(name := "bitcoin-s-core")
   .settings(libraryDependencies ++= Deps.core.value)
-  .dependsOn(cryptoJVM)
+  //.jvmSettings(libraryDependencies ++= Deps.coreJVM)
+  .jsSettings(commonJsSettings)
+  .in(file("core"))
+  .dependsOn(crypto)
+
+lazy val coreJVM = core.jvm
```
rorp, line 76: `coreJS` is needed too.

rorp (changes requested):

rorp:
`cryptoTestJS/test` doesn't work at all. It fails to run any tests. Try to put this class either in `src` or `.js`

```scala
package org.bitcoins.crypto
  
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.must.Matchers

class BCryptoTest extends AnyFlatSpec with Matchers {
  behavior of "BCrypto"

  it must "work" in {
    println("test")
    assert("1" != "2")
  }

}
```
Note, that it doesn't depend on `bitcoin-s` at all but still`cryptoTestJS/test` fails to run it, whereas `cryptoTestJVM/test` runs it perfectly.

Christewart:
> `cryptoTestJS/test` doesn't work at all. It fails to run any tests. Try to put this class either in `src` or `.js`
> 
> ```scala
> package org.bitcoins.crypto
>   
> import org.scalatest.flatspec.AnyFlatSpec
> import org.scalatest.matchers.must.Matchers
> 
> class BCryptoTest extends AnyFlatSpec with Matchers {
>   behavior of "BCrypto"
> 
>   it must "work" in {
>     println("test")
>     assert("1" != "2")
>   }
> 
> }
> ```
> 
> Note, that it doesn't depend on `bitcoin-s` at all but still`cryptoTestJS/test` fails to run it, whereas `cryptoTestJVM/test` runs it perfectly.

Ok this should be fixed in b15e12e

There is still a _ton_ of linking failures, but we don't have the bizarre behavior before of not getting any output when attempting to run tests

rorp (approved):

PR #2735, opened 2021-03-01
Add Broadcast TxoState

### LINKED ISSUE #2732: New TxoState to represent unconfirmed transactions
[here are our current TxoState](https://github.com/bitcoin-s/bitcoin-s/blob/36b45790cfbe21286ab79a000e67ce4f5eadf5ed/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala#L12)

```scala

  /** Means that no funds have been sent to this utxo EVER */
  final case object DoesNotExist extends TxoState

  /** A coinbase output that has not reached maturity and cannot be spent yet.
    * https://bitcoin.stackexchange.com/questions/1991/what-is-the-block-maturation-time
    */
  final case object ImmatureCoinbase extends TxoState

  /** Means we have received funds to this utxo, but they are not confirmed */
  final case object PendingConfirmationsReceived extends ReceivedState

  /** Means we have received funds and they are fully confirmed for this utxo */
  final case object ConfirmedReceived extends ReceivedState

  /** Means we have not spent this utxo yet, but will be used in a future transaction */
  final case object Reserved extends SpentState

  /** Means we have spent this utxo, but it is not fully confirmed */
  final case object PendingConfirmationsSpent extends SpentState

  /** Means we have spent this utxo, and it is fully confirmed */
  final case object ConfirmedSpent extends SpentState
```

As mentioned in https://github.com/bitcoin-s/bitcoin-s/pull/2711#discussion_r584750782 we do not have a txo state to represent a utxo that has been broadcast, but not confirmed! 

Let's say our wallet created a transaction, and attempted to broadcast it, but are unsure if the broadcast suceeded. The best state we could model this txo with is `DoesNotExist` or `PendingConfirmationsSpent`. 

Well, this utxo could be considered both. It does not exist on the blockchain (so `DoesNotExist` could be argued for), but it is a valid utxo that is pending confirmations received (so you could also argue `PendingConfirmationsSpent`). 

I think we should just add a new state that is `Broadcast` that means that a txo that has been broadcast, but hasn't received a confirmation yet. Once it receives a confirmation, it should be `PendingConfirmationsSpent` state. 

Thoughts?

benthecarman:
Fixes #2732

#2736 should probably be decided on before this can be merged

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
@@ -19,7 +19,10 @@ object TxoState extends StringFactory[TxoState] {
     */
   final case object ImmatureCoinbase extends TxoState
 
-  /** Means we have received funds to this utxo, but they are not confirmed */
+  /** Means we have received funds to this utxo, and they have not been confirmed in a block */
+  final case object BroadcastReceived extends ReceivedState
+
+  /** Means we have received funds to this utxo, and they have not reached our confirmation threshold */
   final case object PendingConfirmationsReceived extends ReceivedState
 
   /** Means we have received funds and they are fully confirmed for this utxo */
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
```diff
@@ -19,7 +19,10 @@ object TxoState extends StringFactory[TxoState] {
     */
   final case object ImmatureCoinbase extends TxoState
 
-  /** Means we have received funds to this utxo, but they are not confirmed */
+  /** Means we have received funds to this utxo, and they have not been confirmed in a block */
+  final case object BroadcastReceived extends ReceivedState
+
+  /** Means we have received funds to this utxo, and they have not reached our confirmation threshold */
```
Christewart: If I understand, we have received some confirmations, but not our confirmation threshold.

I just want to make a clear distinction of when the `BroadcastReceived` state ends and the `PendingConfirmationsReceived` state begins.

core/src/main/scala/org/bitcoins/core/wallet/utxo/TxoState.scala
```diff
@@ -28,33 +31,47 @@ object TxoState extends StringFactory[TxoState] {
   /** Means we have not spent this utxo yet, but will be used in a future transaction */
   final case object Reserved extends SpentState
 
-  /** Means we have spent this utxo, but it is not fully confirmed */
+  /** Means we have spent this utxo, and they have not been confirmed in a block */
+  final case object BroadcastSpent extends SpentState
+
+  /** Means we have spent this utxo, and they have not reached our confirmation threshold */
```
Christewart: same as above

Christewart (approved):
Overall looks good. 

This is the test case that tracks the full life cycle, right? I just want to make sure we aren't missing that.

https://github.com/bitcoin-s/bitcoin-s/pull/2735/files#diff-18b1ff078f6c3b0ec3a3ee865e921754d7e4b7ec485fcb64b2d463704748d8aaR239

benthecarman:
> Overall looks good.
> 
> This is the test case that tracks the full life cycle, right? I just want to make sure we aren't missing that.
> 
> [#2735 (files)](https://github.com/bitcoin-s/bitcoin-s/pull/2735/files#diff-18b1ff078f6c3b0ec3a3ee865e921754d7e4b7ec485fcb64b2d463704748d8aaR239)

Yes

PR #2743, opened 2021-03-02
WIP: Implement bcrypto facades

rorp:


crypto-test/.js/src/test/scala/org/bitcoins/crypto/RandomTest.scala
```diff
@@ -0,0 +1,14 @@
+package org.bitcoins.crypto
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.must.Matchers
+
+class RandomTest extends AnyFlatSpec with Matchers {
+
+  it should "generate random bytes" in {
+    val rnd = 1.to(16).map(_ => BCryptoCryptoRuntime.randomBytes(32))
```
Christewart, line 9: don't we want to make assertions on the size of random bytes? I.e. check that we actually got 32 bytes of entropy? And then aren't all `Vector(0x0,0x0,0x0)` or something like that?

Christewart (approved):

PR #2744, opened 2021-03-02
Move blockhash to tx table from spending info table

### LINKED ISSUE #2736: ConfirmedSpent vs PendingConfirmationsSpent TxoStates have no meaning
I discovered this while wokring on #2735

Currently there is no distinction between `ConfirmedSpent` and `PendingConfirmationsSpent` in the wallet.

We calculate which spent state to based off the the txo/`SpendingInfoDb`'s `blockHashOpt` param. However, this param is for the block that the tx the txo originates from, not the spending transaction so we are calcuating if it is confirmed based on an entirely different transaction and most likely block.

There are 2 solutions for this:

1) add a new column/param to `SpendingInfoDb` for the spending tx and its blockhash
2) Remove the distinction and just have a `Spent` `TxoState` (do we need to care if our spending tx is confirmed?)

benthecarman:
Fixes #2736

Moves the `blockhash` column from the spending info table to the tx table. Adds a `spending_txid` column to the spending info table

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -58,17 +58,17 @@ object BitcoindRpcBackendUtil extends BitcoinSLogger {
       _ <- walletStateOpt match {
         case None =>
           for {
-            utxos <- wallet.listUtxos()
-            lastConfirmedOpt = utxos.filter(_.blockHash.isDefined).lastOption
+            txDbs <- wallet.listTransactions()
+            lastConfirmedOpt = txDbs.filter(_.blockHashOpt.isDefined).lastOption
             _ <- lastConfirmedOpt match {
               case None => Future.unit
-              case Some(utxo) =>
+              case Some(txDb) =>
                 for {
-                  heightOpt <- bitcoind.getBlockHeight(utxo.blockHash.get)
+                  heightOpt <- bitcoind.getBlockHeight(txDb.blockHashOpt.get)
                   _ <- heightOpt match {
                     case Some(height) =>
                       logger.info(
-                        s"Last utxo occurred at block $height, syncing from there")
+                        s"Last tx occurred at block $height, syncing from there")
                       doSync(height, bitcoindHeight)
                     case None => Future.unit
                   }
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOHandlingTest.scala
@@ -1,0 +1,57 @@
+package org.bitcoins.wallet
+
+import org.bitcoins.core.protocol.script.EmptyScriptPubKey
+import org.bitcoins.core.wallet.utxo.TxoState._
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.testkit.wallet.WalletTestUtil._
+import org.scalatest.FutureOutcome
+
+class UTXOHandlingTest extends BitcoinSWalletTest {
+
+  behavior of "UTXOHandling"
+
+  override type FixtureParam = Wallet
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    withNewWallet(test, getBIP39PasswordOpt())
+  }
+
+  it must "correctly update txo state based on confirmations" in { wallet =>
+    val utxo = sampleSegwitUTXO(EmptyScriptPubKey)
+    val requiredConfs = 6
+    assert(wallet.walletConfig.requiredConfirmations == requiredConfs)
+
+    val immatureCoinbase = utxo.copyWithState(ImmatureCoinbase)
+    val pendingConfReceived = utxo.copyWithState(PendingConfirmationsReceived)
+    val pendingConfSpent = utxo.copyWithState(PendingConfirmationsSpent)
+    val confReceived = utxo.copyWithState(ConfirmedReceived)
+    val confSpent = utxo.copyWithState(ConfirmedSpent)
+    val reserved = utxo.copyWithState(Reserved)
+    val dne = utxo.copyWithState(DoesNotExist)
+
+    assert(wallet.updateTxoWithConfs(reserved, 1) == reserved)
+
+    assert(wallet.updateTxoWithConfs(immatureCoinbase, 10) == immatureCoinbase)
+    assert(wallet.updateTxoWithConfs(immatureCoinbase, 101) == confReceived)
+
+    assert(
+      wallet.updateTxoWithConfs(pendingConfReceived, 1) == pendingConfReceived)
+    assert(
+      wallet.updateTxoWithConfs(pendingConfReceived,
+                                requiredConfs) == confReceived)
+
+    assert(wallet.updateTxoWithConfs(pendingConfSpent, 1) == pendingConfSpent)
+    assert(
+      wallet.updateTxoWithConfs(pendingConfSpent, requiredConfs) == confSpent)
+
+    assert(wallet.updateTxoWithConfs(dne, 1) == dne)
+    assert(wallet.updateTxoWithConfs(dne, requiredConfs) == dne)
+
+    assert(wallet.updateTxoWithConfs(confSpent, 1) == confSpent)
+    assert(wallet.updateTxoWithConfs(confSpent, requiredConfs) == confSpent)
+
+    assert(wallet.updateTxoWithConfs(confReceived, 1) == confReceived)
+    assert(
+      wallet.updateTxoWithConfs(confReceived, requiredConfs) == confReceived)
+  }
+}
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
@@ -40,6 +40,7 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
       newTransactions <- wallet.listTransactions()
     } yield {
       assert(updatedCoins.forall(_.state == TxoState.PendingConfirmationsSpent))
+      assert(updatedCoins.forall(_.spendingTxIdOpt.contains(tx.txIdBE)))
       assert(!oldTransactions.map(_.transaction).contains(tx))
       assert(newTransactions.map(_.transaction).contains(tx))
     }
@@ -60,8 +61,11 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
 
       // Give tx a fake hash so it can appear as it's in a block
       hash <- bitcoind.getBestBlockHash
-      _ <- wallet.spendingInfoDAO.upsertAllSpendingInfoDb(
-        updatedCoins.map(_.copyWithBlockHash(hash)).toVector)
+      _ <- wallet.processTransaction(tx, Some(hash))
+
+      pendingCoins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
+      _ <- wallet.updateUtxoPendingStates()
+      _ = assert(pendingCoins.forall(_.state == PendingConfirmationsSpent))
 
       // Put confirmations on top of the tx's block
       _ <- bitcoind.getNewAddress.flatMap(
@@ -70,10 +74,13 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
       // Need to call this to actually update the state, normally a node callback would do this
       _ <- wallet.updateUtxoPendingStates()
       confirmedCoins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
-    } yield assert(confirmedCoins.forall(_.state == ConfirmedSpent))
+    } yield {
+      assert(confirmedCoins.forall(_.state == ConfirmedSpent))
+      assert(confirmedCoins.forall(_.spendingTxIdOpt.contains(tx.txIdBE)))
+    }
   }
 
-  it should "track a utxo state change to pending recieved" in { param =>
+  it should "track a utxo state change to pending received" in { param =>
     val WalletWithBitcoindRpc(wallet, bitcoind) = param
 
     for {
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
@@ -1,9 +1,10 @@
 package org.bitcoins.wallet
 
 import org.bitcoins.core.currency.Satoshis
+import org.bitcoins.core.number._
 import org.bitcoins.core.protocol.BitcoinAddress
-import org.bitcoins.core.protocol.script.{EmptyScriptPubKey, P2PKHScriptPubKey}
-import org.bitcoins.core.protocol.transaction.TransactionOutput
+import org.bitcoins.core.protocol.script._
+import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.core.wallet.utxo.TxoState
 import org.bitcoins.core.wallet.utxo.TxoState._
@@ -78,6 +79,55 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
       assert(confirmedCoins.forall(_.state == ConfirmedSpent))
       assert(confirmedCoins.forall(_.spendingTxIdOpt.contains(tx.txIdBE)))
     }
+  }
+
+  it should "handle processing a new spending tx for a spent utxo" in { param =>
+    val WalletWithBitcoindRpc(wallet, bitcoind) = param
+
+    for {
+      oldTransactions <- wallet.listTransactions()
+      tx <- wallet.sendToAddress(testAddr, Satoshis(3000), None)
+
+      updatedCoins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
+      newTransactions <- wallet.listTransactions()
+      _ = assert(updatedCoins.forall(_.state == PendingConfirmationsSpent))
+      _ = assert(!oldTransactions.map(_.transaction).contains(tx))
+      _ = assert(newTransactions.map(_.transaction).contains(tx))
+
+      // Give tx a fake hash so it can appear as it's in a block
+      hash <- bitcoind.getBestBlockHash
+      _ <- wallet.processTransaction(tx, Some(hash))
+
+      pendingCoins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
+      _ <- wallet.updateUtxoPendingStates()
+      _ = assert(pendingCoins.forall(_.state == PendingConfirmationsSpent))
+
+      // Put confirmations on top of the tx's block
+      _ <- bitcoind.getNewAddress.flatMap(
+        bitcoind.generateToAddress(wallet.walletConfig.requiredConfirmations,
+                                   _))
+      // Need to call this to actually update the state, normally a node callback would do this
+      _ <- wallet.updateUtxoPendingStates()
+      confirmedCoins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
+
+      // Assert tx is confirmed
+      _ = assert(confirmedCoins.forall(_.state == ConfirmedSpent))
+      _ = assert(confirmedCoins.forall(_.spendingTxIdOpt.contains(tx.txIdBE)))
+
+      // Create tx to spend same utxos
+      newTx = {
+        val inputs = updatedCoins.map { db =>
+          TransactionInput(db.outPoint, EmptyScriptSignature, UInt32.zero)
+        }
+        BaseTransaction(Int32.zero, inputs, Vector.empty, UInt32.zero)
+      }
+
+      _ <- wallet.processTransaction(newTx, None)
+
+      coins <- wallet.spendingInfoDAO.findOutputsBeingSpent(tx)
+
+      // coins should not change
+    } yield assert(coins == confirmedCoins)
   }
 
   it should "track a utxo state change to pending received" in { param =>
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
@@ -2,7 +2,6 @@ package org.bitcoins.wallet
 
 import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.number._
-import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.protocol.script._
 import org.bitcoins.core.protocol.transaction._
 import org.bitcoins.core.wallet.fee.SatoshisPerByte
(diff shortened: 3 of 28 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -58,17 +58,17 @@ object BitcoindRpcBackendUtil extends BitcoinSLogger {
       _ <- walletStateOpt match {
         case None =>
           for {
-            utxos <- wallet.listUtxos()
-            lastConfirmedOpt = utxos.filter(_.blockHash.isDefined).lastOption
+            txDbs <- wallet.listTransactions()
+            lastConfirmedOpt = txDbs.filter(_.blockHashOpt.isDefined).lastOption
```
Christewart, line 62: Aren't we assuming the database returns things in the correct order here? I don't see how `.lastOpt` guarantees that it is the "tallest" block

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -75,18 +74,8 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
...
-            .getNumberOfConfirmations(utxo._1.blockHash.get)
-            .map((utxo, _)))
-        confs <- FutureUtil.collect(confFs)
-        immatureCoinbases =
-          confs
-            .filter { case (_, confsOpt) =>
-              confsOpt.isDefined && confsOpt.get < Consensus.coinbaseMaturity
-            }
-            .map(_._1)
+        immatureCoinbases = utxoWithTxs.filter(
```
Christewart, line 77: :thinking_face: why could this be simplified so much? 

I guess i am more confused why we were doing things the way we were before? The new code seems correct

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -96,59 +96,78 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
-                  if (confs > Consensus.coinbaseMaturity) {
+    val txIds =
+      spendingInfoDbs.map(_.txid) ++ spendingInfoDbs.flatMap(_.spendingTxIdOpt)
+
+    val byBlockF = {
+      transactionDAO.findByTxIdBEs(txIds).map { txDbs =>
+        val blockHashMap = txDbs.map(db => db.txIdBE -> db.blockHashOpt).toMap
+        val blockHashAndDb = spendingInfoDbs.map { txo =>
+          val txToUse = txo.state match {
+            case _: ReceivedState | DoesNotExist | ImmatureCoinbase =>
```
Christewart: is this correct in the case that we receive and a spend a transaction the transaction in the same block? I.e. spending an unconfirmed transaction? 

Wouldn't this be an issue with CPFP?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -96,59 +96,78 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
-          case None =>
-            logger.warn(
-              s"Given txos exist in block (${blockHash.hex}) that we do not have or that has been reorged! $txos")
-            Vector.empty
-          case Some(confs) =>
-            txos.flatMap { txo =>
-              txo.state match {
-                case TxoState.ImmatureCoinbase =>
-                  if (confs > Consensus.coinbaseMaturity) {
+    val txIds =
```
Christewart, line 105: So this method is really blowing up in size and complexity. Usually that is a sign that we need to break it up into smaller methods that are testable. 

I don't see any unit tests that directly test this logic, and if I understand our initial bug, it originates from here. 

I think we should change the scope of this method to be `private[wallet]` and add unit tests for various cases like CPFP which i mention below.

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -300,35 +316,19 @@ private[wallet] trait UtxoHandling extends WalletLogger {
     require(unreserved.isEmpty, s"Some utxos are not reserved, got $unreserved")
 
     // unmark all utxos are reserved
-    val groupedUtxos = utxos
+    val updatedUtxos = utxos
       .map(_.copyWithState(TxoState.PendingConfirmationsReceived))
```
Christewart, line 341: Why is the `TxoState.PendingConfirmationsReceived` received here? 

Is this the "safest" state to use in this situation?

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -483,7 +482,8 @@ case class SpendingInfoDAO()(implicit
 
     def scriptWitnessOpt: Rep[Option[ScriptWitness]] = column("script_witness")
 
-    def blockHash: Rep[Option[DoubleSha256DigestBE]] = column("block_hash")
+    def spendingTxIdOpt: Rep[Option[DoubleSha256DigestBE]] = column(
+      "spending_tx_id")
```
Christewart: nit: Let's try to be consistent with how we do `txid` Above we use `txid` but we are using `tx_id`. 

For those of us that use sql queries this with slowly drive us mad.  Consistency gives a much better developer/devops experience :-)

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
benthecarman: The stuff here before was added before we had the immature coinbase txo state

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: Pretty much

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: That would be fine, the parent tx would be in the `ReceivedState` as a `PendingConfirmationReceived`

benthecarman: actually realizing now `Reserved` in handled incorrectly here because it is a `SpentState`

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: We are, I am not sure if this is the best, but probably a topic for a separate PR, also this is only ever used if you have an old wallet.

Christewart:
I think there are some issues with postgres? 

https://github.com/bitcoin-s/bitcoin-s/pull/2744/checks?check_run_id=2024747947#step:5:436

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -253,13 +265,16 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
     * its spending status. Otherwise returns `None`.
     */
-  private def markAsPendingSpent(
-      out: SpendingInfoDb): Future[Option[SpendingInfoDb]] = {
+  private def markAsSpent(
+      out: SpendingInfoDb,
+      spendingTxId: DoubleSha256DigestBE): Future[Option[SpendingInfoDb]] = {
     out.state match {
       case TxoState.ConfirmedReceived | TxoState.PendingConfirmationsReceived |
           TxoState.ImmatureCoinbase =>
```
Christewart: Does immature coinbase work here? 

I don't follow why we would want to transition from `TxoState.ImmatureCoinbase` -> `TxoState.PendingConfirmationsSpent` 

This is for the case where somebody pays us with the coinbase output right? Don't we want to transition the state this way 

1. ImmatureCoinbase
2. PendingConfirmationsReceived
3. ConfirmedReceived

and then whenever we spend the coinbase we go to `PendingConfirmationsSpent`?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -268,7 +283,13 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
         updatedF.map(Some(_))
       case TxoState.Reserved | TxoState.ConfirmedSpent |
```
Christewart: So this isn't making much sense to me. I think you are including too many states here. 

If we ALREADY have a utxo that is `TxoState.ConfirmedSpent` you want to re-write the ALREADY EXISTING `spendingTxId` with a new one? 

This is getting into reorg territory I believe (#2179), but we need to think through these state transitions and test them more. 

I think the valid `TxoState` for this pattern match is just `TxoState.Reserved`? I'm not exactly sure how to think about `TxoState.DoesNotExist`, as that indicates that our wallet was unaware of one of _our own_ utxos to begin with?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -87,68 +88,95 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
   }
 
-  protected def updateUtxoConfirmedState(
+  private[wallet] def updateUtxoConfirmedState(
       txo: SpendingInfoDb): Future[Option[SpendingInfoDb]] = {
     updateUtxoConfirmedStates(Vector(txo)).map(_.headOption)
   }
 
-  protected def updateUtxoConfirmedStates(
+  private[wallet] def getDbsByRelevantBlock(
```
Christewart, line 102: scaladoc

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -87,68 +88,95 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
-                  }
-                case TxoState.Reserved =>
-                  // We should keep the utxo as reserved so it is not used in
-                  // a future transaction that it should not be in
-                  None
-                case TxoState.DoesNotExist | TxoState.ConfirmedReceived |
-                    TxoState.ConfirmedSpent =>
-                  None
-              }
+  private[wallet] def updateTxoWithConfs(
```
Christewart, line 135: scaladoc

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -87,68 +88,95 @@ private[wallet] trait UtxoHandling extends WalletLogger {
...
+        // We should keep the utxo as reserved so it is not used in
+        // a future transaction that it should not be in
+        txo
+      case TxoState.DoesNotExist | TxoState.ConfirmedReceived |
+          TxoState.ConfirmedSpent =>
+        txo
+    }
+  }
+
+  private[wallet] def updateUtxoConfirmedStates(
```
Christewart, line 167: scaladoc

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
Christewart: Also what is the case for the key `None` in the Map? Also i'm unsure where the 'revelevant' part comes into play here? Shouldn't we should call it `groupSpendingInfoByBlock()`?

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOHandlingTest.scala
```diff
@@ -0,0 +1,57 @@
+package org.bitcoins.wallet
+
+import org.bitcoins.core.protocol.script.EmptyScriptPubKey
+import org.bitcoins.core.wallet.utxo.TxoState._
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.testkit.wallet.WalletTestUtil._
+import org.scalatest.FutureOutcome
+
+class UTXOHandlingTest extends BitcoinSWalletTest {
```
Christewart, line 9: other cases that aren't tested AFAICT

1. Coinbase utxos
2. CPFP utxos
3. RBF
4. Reorg scenarios (#2179)
5. marking a utxo as reserved -> unreserved

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
benthecarman: > Also what is the case for the key None in the Map

mempool txs

>  Also i'm unsure where the 'revelevant' part comes into play here? Shouldn't we should call it groupSpendingInfoByBlock()

I used relevant because there are 2 txs "by block" could refer to, the receving, and the spending tx.

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOHandlingTest.scala
benthecarman: These are handled in `UTXOLifeCycleTest`, `WalletSendingTest`, and `WalletIntegrationTest`

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: We don't allow creating a tx that spends an ImmatureCoinbase, however, if we do process one where we spend it, we should handle it correctly and mark it at spent rather than received. Otherwise we would count it as part of our balance once we have enough confirmations.

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: > If we ALREADY have a utxo that is TxoState.ConfirmedSpent you want to re-write the ALREADY EXISTING spendingTxId with a new one?
> This is getting into reorg territory I believe (#2179), but we need to think through these state transitions and test them more.

Yeah these would occur in a RBF reorg scenario.

> I'm not exactly sure how to think about TxoState.DoesNotExist, as that indicates that our wallet was unaware of one of our own utxos to begin with?

Yeah, this is kinda hard to decide because we don't really have a definition of what this state represents. It's only used when we RBF utxos

benthecarman: > I think the valid TxoState for this pattern match is just TxoState.Reserved? 

We need pending conf too because we don't have a the broadcasted state yet

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: Why don't we just return an exception (`Future.failed()`) as this seems to indicate wallet corruption? If we don't allow it, this state shouldn't be possible, and therefore we should return an exception (and it seems we can't make it impossible to represent this state with the type system?)

Is there any unit tests that test this behavior you mention above? If so could you link me?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -253,21 +265,36 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+            .copyWithSpendingTxId(spendingTxId)
         val updatedF =
           spendingInfoDAO.update(updated)
         updatedF.foreach(updated =>
           logger.debug(
             s"Marked utxo=${updated.toHumanReadableString} as state=${updated.state}"))
         updatedF.map(Some(_))
-      case TxoState.Reserved | TxoState.ConfirmedSpent |
-          TxoState.PendingConfirmationsSpent | TxoState.DoesNotExist =>
+      case TxoState.Reserved | TxoState.PendingConfirmationsSpent =>
```
Christewart, line 281: Again, i think this is a problem. If I understand correctly, we saying our wallet behavior is to "only keep track of the _last seen_ spending transaction for an output. Disregard previous spending transactions associated with this utxo" 

At minimum, we should be logging `warn` messages for this. 

If I understand correctly, we do need to "pick one" (first seen spending tx, or last seen spending tx). If this is the decision we want to make until #2179 is done, let's make a unit test for it. 

unfortunately i don't think there is any guarantees we can provide that the last seen transaction wasn't like 10 blocks behind the tip is there?

Do we have unit tests for an old spending transaction being overwritten by a new spending transaction? If so could you link me?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -253,21 +265,36 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+        val updatedF =
+          spendingInfoDAO.update(updated)
+        updatedF.map(Some(_))
+      case TxoState.DoesNotExist | TxoState.ConfirmedSpent =>
+        if (
+          out.spendingTxIdOpt.isDefined && !out.spendingTxIdOpt.contains(
+            spendingTxId)
+        ) {
+          logger.warn(
+            s"Attempted to mark an already spent utxo ${out.outPoint.hex} with a new spending tx ${spendingTxId.hex}")
```
Christewart: Let's make unique log messages for both `TxoState.DoesNotExist` and `TxoState.ConfirmedSpent` since they are pretty different situations. 

it seems we should also be recommending rescans of the wallet at minimum if these cases occur? 

This is getting into wallet corruption territory, maybe we want to return an exception rather than `FutureUtil.none`?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -253,21 +265,36 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
   /** If the given UTXO is marked as unspent, updates
```
Christewart, line 263: is this scaladoc still up to date?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
Christewart: Ok, the scaladoc really helps out. One last nit: Add the mempool case to the scaladoc too.

Christewart (changes requested):

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
benthecarman: yes

Christewart:
Still looks like there might be a problem on postgres: https://github.com/bitcoin-s/bitcoin-s/runs/2043437745#step:5:343

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
```diff
@@ -70,10 +149,49 @@ class UTXOLifeCycleTest extends BitcoinSWalletTest {
...
+
+      // Create tx to spend same utxos
+      newTx = {
+        val inputs = updatedCoins.map { db =>
+          TransactionInput(db.outPoint, EmptyScriptSignature, UInt32.zero)
+        }
+        BaseTransaction(Int32.zero, inputs, Vector.empty, UInt32.zero)
+      }
+
+      res <- recoverToSucceededIf[RuntimeException](
```
Christewart, line 165: If I understand this, we are not asserting our current expected behavior which we talked about here. The new spending txid should be in the database rather than the old one, right?

https://github.com/bitcoin-s/bitcoin-s/pull/2744#discussion_r587752135

Specifically 

>Do we have unit tests for an old spending transaction being overwritten by a new spending transaction? If so could you link me?

https://github.com/bitcoin-s/bitcoin-s/pull/2744/files#diff-fd826a398083437abb5cf5acafa8af283c5420c626aa902469a5b743b8dfa094R285

wallet/src/main/resources/postgresql/wallet/migration/V11__blockhash_to_transaction_table.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE "tx_table"
+    ADD COLUMN "block_hash" TEXT;
+ALTER TABLE "tx_table"
+    Drop COLUMN "id";
```
Christewart, line 4: Hmm.. where was this added? I don't see it in my code base?

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
Christewart: ping

Christewart (commented):
Overall the logs/documentation is an improvement!

Christewart:
More CI failures unrelated to our current problems on CI https://github.com/bitcoin-s/bitcoin-s/pull/2744/checks?check_run_id=2078260350#step:5:616

wallet/src/main/resources/postgresql/wallet/migration/V11__blockhash_to_transaction_table.sql
benthecarman: It just existed on postgres so I dropped it

Christewart:
Still failing as of 3cad724

https://github.com/bitcoin-s/bitcoin-s/pull/2744/checks?check_run_id=2079865278#step:5:489

wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala
benthecarman: > If I understand this, we are not asserting our current expected behavior which we talked about here. The new spending txid should be in the database rather than the old one, right?

That is the case for pending conf utxos, this is asserting it fails when the utxos are confirmed

benthecarman: > Do we have unit tests for an old spending transaction being overwritten by a new spending transaction? If so could you link me?

https://github.com/bitcoin-s/bitcoin-s/pull/2744/files#diff-18b1ff078f6c3b0ec3a3ee865e921754d7e4b7ec485fcb64b2d463704748d8aaR96

Christewart (approved):

benthecarman:
So the postgres statement that is failing is

```
update "wallet"."tx_table" set "transaction"=?,"unsignedTxIdBE"=?,"unsignedTx"=?,"wTxIdBE"=?,"totalOutput"=?,"numInputs"=?,"numOutputs"=?,"locktime"=?,"block_hash"=? where "txIdBE"=?;
insert into "wallet"."tx_table" ("txIdBE","transaction","unsignedTxIdBE","unsignedTx","wTxIdBE","totalOutput","numInputs","numOutputs","locktime","block_hash") select ?,?,?,?,?,?,?,?,?,? where not exists (select 1 from "wallet"."tx_table" where "txIdBE"=?)
```

benthecarman:
From slack by me:

> I figured out the postgres issue on #2744 finally. It happens because we are processing the incoming and outgoing utxos in parallel, and then since postgres allows for multiple db threads, it will attempt to upsert the to the tx table twice at the same time, this causes one to fail sometimes because it will be inserted before it is properly done with the upsert so 2 threads will use an insert statement. A simple way to fix this would to make it we so we don't process these in parallel or we could lower the number of threads to 1

I went ahead and removed the parallel processing of the inputs and outputs

PR #2749, opened 2021-03-02
Fix DLCOracle to be Network Agnostic

### LINKED ISSUE #2748: Oracle attestment tlv has wrong public key
When signing the bitcoin price, the oracle signed something. However, when we went to verify these signatures they did not correspond to the possible outcomes (0 or 1). This is because the public key in the tlv is incorrect.

TLV: ```
fdd868fd048b066e6164617641c7eb8bfdb73abc801ecff1eee6ea266651509a4e6a63153e42919ac3a22dfc3c0011e442d1aed73b4ddeb1c777d8af5090c8f875b34b3b01339771cce9baed8026328dc3f5ca5f7667193ec5b22bc671aa6e19ada8d7a0e85258ac15b519b2b01e6a42c66d5d77854fe552783ccafc3c1ce30efa4aa19c26c792b713dee3780be69a7274ff330504103961d13349de8900f59c025b596c1f1d9360eb1bc9624fbdb5a32477da0ab5f2028fec16150f07c26eaba7837959964821f11dbbb359ba338df31e474798b2952427d31aa28a9b0d3dc9cec2cd5e93a2d6620bf3b47fdabefbcc5c53feec3039f9497957e4fda01dfa846ac646d581b8732ba1e18121f9ae9be962682fb06057bbb8ca8296ad88caf128e059e6369bf1c185815fa12e45f36d66fa89557b00146b2ee71190028674ab13c5dfcfb4d01bf0c362833bd413286847126a8e3b5d0362fcfff81fa6251de228f80492144a5166a49ca2d65e24fbdc6e6409166d6649dbcd86242cbcce96bc4f54176dea68d412a5ebfa98a150a4ba350ba5014717290f281d1d7abb539df866b60f4242d35bbee1c251d7b00fb96449cc4634e3179a72b151f3f2099c060adc67e244d1fd3473bacdc5820c3350c8b5fef757a11d71ddd9dc0a9d221b074cc700743fa2becf71039b57b46925fd5396cc60d19f88f420617432c21ec449df2f83bacca586cd233c0b24a279bfacc43c07ad1fe9b8c125666bf7b6fc1c32e57ce8a73eea895fd685768fb8062dde133eaf76dbd4897ff311d943e3e6444c895dc54a15364fd5124c771638a8c8c2f8ee468c93ea79e19a684e48102a92df69b347f95e902f13ca4a6835415e61d63ec30b744cb98db9cad679ce67a5cf53f88f37f7bca0c51c1eb5fef0b86f369f891a81bdb7b5131045e8f8b836806dbcb652a5f6d6ba061a1c6b0703c5516658b5b1f76ce6b5567428d8a1a57c8cc3610b904f916a4a213e46a53c553d85d7df3a524c7216ccb3e5d5a551e4237543898c82c8cc09aebeaa58a476381813e80d64e3b3a9860aca55942c7e4592ee457c17861e4c3340cdd9d96f539af0296dbf92d3e85402f1989d991be2519b2aef03b386069a71634ad29a1a529a2b9d0ff772ca5563f68c33d977e90b32bb8be7543d903ea639f86378ca2fa6de36c409644f1b0ef33819d5bc583de612777943cc52ff933690994732223194d08623ff65b4367ac4eb8cfdccf4f33e3fc330a3b01da6c4d8f433d8bf5331518e766cbd12a79b6e01202dd9a709116e644ac9c65b8ec46797b460694ff52520f9ae323c7a78a53c48876a3305858aee06ed5d3208a3559e892db0394049dc4a25d1ce7d6f64c5310ebccb390015440d4013d638216e3f06be07e15f9bf4b0d75c942f39c2af3d86e572e43aee554fbc182b9a31a8363d22f130535b110e0df3145947faf1c215ddcfcf998aa70a157bafef201cbbc959c79f0119c7555b209478b203e8a29d358f7c78ba19de8b4a978fa3528dec436f94570fe5ddc3c805215be649785b5aae8e9ec4bee46abc7879a9fb5721e904bf13cfa51b459e2585336d5d4bd7c4e601300131013001310131013101300130013101300130013101300131013001300130```

benthecarman:
Fixes #2748

### DIFF
--- a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
+++ b/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
@@ -31,7 +31,8 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
   private val rValAccount: HDAccount = {
     val purpose = conf.kmParams.purpose
     // We have to use Testnet here because before this was dictated by
-    // by the config, and the default network was regtest
+    // by the network config. The default network config was Regtest,
+    // which uses a Testnet HDCoinType, so that was chosen as the hard coded value.
     val coin = HDCoin(purpose, HDCoinType.Testnet)
     HDAccount(coin, 0)
   }
@@ -41,7 +42,8 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
 
   private def signingKey: ECPrivateKey = {
     // We have to use Testnet here because before this was dictated by
-    // by the config, and the default network was regtest
+    // by the network config. The default network config was Regtest,
+    // which uses a Testnet HDCoinType, so that was chosen as the hard coded value.
     val coin = HDCoin(HDPurposes.SegWit, HDCoinType.Testnet)
     val account = HDAccount(coin, 0)
     val purpose = coin.purpose
--- a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
+++ b/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
@@ -30,8 +30,9 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
 
   private val rValAccount: HDAccount = {
     val purpose = conf.kmParams.purpose
-    val coin =
-      HDCoin(purpose, HDCoinType.fromNetwork(conf.network))
+    // We have to use Testnet here because before this was dictated by
+    // by the config, and the default network was regtest
+    val coin = HDCoin(purpose, HDCoinType.Testnet)
     HDAccount(coin, 0)
   }
 
@@ -39,7 +40,9 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
   private val rValueChainIndex = 0
 
   private def signingKey: ECPrivateKey = {
-    val coin = HDCoin(HDPurposes.SegWit, HDCoinType.fromNetwork(conf.network))
+    // We have to use Testnet here because before this was dictated by
+    // by the config, and the default network was regtest
+    val coin = HDCoin(HDPurposes.SegWit, HDCoinType.Testnet)
     val account = HDAccount(coin, 0)
     val purpose = coin.purpose
     val chain = HDChainType.External
--- a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
+++ b/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
@@ -28,12 +28,18 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
 
   implicit val ec: ExecutionContext = conf.ec
 
+  // We have to use Testnet here because before this was dictated by
+  // by the network config. The default network config was Regtest,
+  // which uses a Testnet HDCoinType, so that was chosen as the hard coded value.
+  // It was a mistake to originally have this dictated by the config
+  // as the oracle should be network agnostic.
+  // see https://github.com/bitcoin-s/bitcoin-s/issues/2748
+  private val coinType = HDCoinType.Testnet
+
   private val rValAccount: HDAccount = {
     val purpose = conf.kmParams.purpose
-    // We have to use Testnet here because before this was dictated by
-    // by the network config. The default network config was Regtest,
-    // which uses a Testnet HDCoinType, so that was chosen as the hard coded value.
-    val coin = HDCoin(purpose, HDCoinType.Testnet)
+
+    val coin = HDCoin(purpose, coinType)
     HDAccount(coin, 0)
   }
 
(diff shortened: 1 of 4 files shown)


dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -30,16 +30,21 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
 
   private val rValAccount: HDAccount = {
     val purpose = conf.kmParams.purpose
-    val coin =
-      HDCoin(purpose, HDCoinType.fromNetwork(conf.network))
+    // We have to use Testnet here because before this was dictated by
```
Christewart: can we add migration code here so we don't have to support this indefinitely? 

Where was this introduced? In #2720 or #2667 ? 

Was this in the latest official release of krystal-bull ? 

This is going to suck to have to maintain forever :/

Christewart: Can we get some unit tests? This seems very concerning that this isn't being caught by tests ?

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
```diff
@@ -165,7 +165,13 @@ case class DLCOracleAppConfig(
                                           bip39PasswordOpt)
     val oracle = new DLCOracle(key)(this)
 
-    start().map(_ => oracle)
+    for {
+      _ <- start()
+      differentKeyDbs <- oracle.eventDAO.findDifferentPublicKey(
+        oracle.publicKey)
+      fixedDbs = differentKeyDbs.map(_.copy(pubkey = oracle.publicKey))
+      _ <- oracle.eventDAO.updateAll(fixedDbs)
```
Christewart, line 173: So we just update them with the new public key? Aren't there possibly announcements published with the old key?

Christewart (changes requested):

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
benthecarman: > Where was this introduced? In #2720 or #2667 ?

This was introduced from the beginning :/

> Was this in the latest official release of krystal-bull ?

Yes but not a problem because you can't change the network config

> This is going to suck to have to maintain forever :/

Shouldn't require any changes after this because it is a hard coded value

benthecarman: > can we add migration code here so we don't have to support this indefinitely?

done in the app config

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
benthecarman: > So we just update them with the new public key? Aren't there possibly announcements published with the old key?

I don't think so, will be adding tests to make sure

benthecarman: Added test, looks like we could have published an announcement on an old key :/. Not sure the best way to fix this

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
Christewart: We are essentially hardcoding a _specific_ network here (Testnet), which isn't really "network-agnostic". 

Perhaps we should just try to allocate a new HDCoinType for the DLC oracle? It seems that might be the best choice we make for the future, while having to live with this bug we currently have to support?

benthecarman: > Perhaps we should just try to allocate a new HDCoinType for the DLC oracle? It seems that might be the best choice we make for the future, while having to live with this bug we currently have to support?

We already did that for the purpose, and if we change it, it will break all current oracles.

> We are essentially hardcoding a specific network here (Testnet), which isn't really "network-agnostic".

Kinda, we have to pick one tho so I am just picking the default so we don't break compatibilty

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala
Christewart: I think we will have to support it then. Here is what I think the logic should be on next startup of a oracle 

1. Check if we have anything in the db
2. If we don't, use the new HDCoinType.DLC (I recommended above)
3. If we do, use `HDCoinType.Testnet` (we have no choice, i believe? Unless you want to create a new oracle)
4. Use the coin type to derive the correct `signingKey`. 

This should be well documented with links to issues that refer to what is going on. I don't think there has been a proper issue filed yet that details the issues here? #2748 just says there is a problem, with no detailed description of how the bug was introduced, why we didn't catch it until now, and long term implications of the bug (we now need to support two HDCoinType, or else have to explain forever why we are using testnet as our HDCoinType)

I do feel somewhat strongly about using a new HDCoinType to reduce amiguity of what is going on. As was our intention, we are trying to make the oracle agnostic, but we use network specific data structures in the code, which is confusing IMO.

Christewart (changes requested):
Thoughts on this, we should probably talk to @nkohen to see what he thinks too

Christewart (approved):

nkohen (approved):

PR #2757, opened 2021-03-03
Add signed outcome to `getevent` rpc, fix other small api bugs

benthecarman:
Add signed outcome to `getevent` rpc
Fixed it so the `listevents` returns all the event names since we now use those as identifiers. 
Fixed a bug where unsigned numeric events would have an extra `[ ]` at the end of its list of outcomes.


core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
```diff
@@ -168,7 +169,35 @@ case class CompletedDigitDecompositionV0OracleEvent(
...
     attestations: Vector[FieldElement])
     extends CompletedOracleEvent
-    with DigitDecompositionV0OracleEvent
+    with DigitDecompositionV0OracleEvent {
+
+  val outcomeBase10: Long = {
+    val (digits, positive) = {
+      eventDescriptorTLV match {
+        case _: SignedDigitDecompositionEventDescriptor =>
+          val positive = outcomes.head
```
Christewart: It seems like we need to make some dedicated types rather than using `outcomes: Vector[DLCAttestationType]` everywhere. 

That way we can at least get some type safety to know the `outcomes: Vector[DLCAttestationType]` doesn't correspond to a enum event for instance. 

it also allows us to get around this nasty casting if I understand the problem. 

If I understand correctly, we could even have 

`outcomes = Vector(RangeAttestation, EnumAttestation, DigitDecompAttestation)` 

basically the vector not being internally consistent with the type since all of these things are subtypes of `DLCAttestationType`.

Christewart (commented):
Idk if the type suggestions need to be done on this PR, but if they aren't, let's get an issue filed so we can address this in the future.

Christewart:
needs rebase, and are we going to make the type that wraps the `Vector[DLCAttestationType]` on this PR or another one? It seems like there was consensus that it was a good idea, but i imagine it will be a decent size diff.

core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
```diff
@@ -130,10 +135,29 @@ case class CompletedDigitDecompositionV0OracleEvent(
     maturationTime: Instant,
     announcementSignature: SchnorrDigitalSignature,
     eventDescriptorTLV: DigitDecompositionEventDescriptorV0TLV,
-    outcomes: Vector[DLCAttestationType],
+    outcomes: Vector[DigitDecompositionAttestationType],
```
Christewart: Why don't we just combine these two fields into one type? It seems that should be used together no? 

What I had in mind what something like 

```scala
case class NumericDLCOutcome(outcomes: Vector[DigitDecompositionAttestationType]) extends DLCOutcome { 
   val attesations: Vector[FieldElement] = outcome.map(_.fieldElement) // or something like this
 }
```

and then we can just have 

```scala
case class CompletedDigitDecompositionV0OracleEvent(
    pubkey: SchnorrPublicKey,
    nonces: Vector[SchnorrNonce],
    eventName: String,
    signingVersion: SigningVersion,
    maturationTime: Instant,
    announcementSignature: SchnorrDigitalSignature,
    eventDescriptorTLV: DigitDecompositionEventDescriptorV0TLV
    outcome: NumericDLCOutcome) { 
       val attestations: Vector[FieldElement] = outcome.attestations
      val outcomes: Vector[DigitDecompositionAttestationType] = outcome.outcomes
    }
```

Christewart (changes requested):

Christewart:
Tests do not pass.... https://github.com/bitcoin-s/bitcoin-s/pull/2757/checks?check_run_id=2043495887#step:5:3435

Christewart (approved):

PR #2775, opened 2021-03-08
Give DLC from file functions destination optional param

benthecarman:
Adds a `destination` parameter to `acceptdlcfromfile` and `signdlcfromfile` that is a file destination and adds it to the GUI. This should allow us to get around the problems with the CLI timing out.

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -1,8 +1,5 @@
 package org.bitcoins.cli
 
-import java.io.File
-import java.nio.file.Path
-import java.time.Instant
 import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.CliReaders._
 import org.bitcoins.commons.jsonmodels.bitcoind.RpcOpts.LockUnspentOutputParameter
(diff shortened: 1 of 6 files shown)


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -242,6 +243,14 @@ object ConsoleCli {
                 case accept: AcceptDLCOfferFromFile =>
                   accept.copy(path = path)
                 case other => other
+              })),
+          arg[Path]("destination")
+            .required()
```
nkohen: since these are both required, is there a reason the field is an `Option`?

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -52,6 +52,18 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
...
   }
 
+  private def handleDestinationOpt(
+      returnedStr: String,
+      destinationOpt: Option[Path]
+  ): String = {
+    destinationOpt match {
+      case Some(path) =>
+        Files.write(path, returnedStr.getBytes)
+        path.toAbsolutePath.toString
```
nkohen: should this maybe say like "Written to ${...}"?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
benthecarman: oops, made optional

nkohen (approved):

PR #2779, opened 2021-03-08
Don't require DLCSetup for refunding DLC

benthecarman:
Fixes #2753

Previously we would need to create a `DLCSetup` to refund a DLC which required fetching all of our outcome signatures. This fixes that to only require to need to fetch the refund signature.

PR #2780, opened 2021-03-08
Don't require DLCSetup for exeuting DLC

benthecarman:
Built on top of #2779

Previously we would need to create a DLCSetup to execute a DLC which required fetching all of our outcome signatures. This fixes that to only require to need to fetch the one outcome signature.


dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
```diff
@@ -116,6 +143,20 @@ case class DLCExecutor(signer: DLCTxSigner)(implicit ec: ExecutionContext) {
...
+      oracleSigs: Vector[OracleSignatures]): Option[
+    (OracleOutcome, Vector[OracleSignatures])] = {
+    val threshold = contractInfo.oracleInfo.threshold
+    val sigCombinations = CETCalculator.combinations(oracleSigs, threshold)
+    sigCombinations.flatMap { sigs =>
+      contractInfo.findOutcome(sigs) match {
+        case Some(outcome) => Some((outcome, sigs))
+        case None          => None
+      }
+    }.headOption // should only be one
```
nkohen: I believe this `headOption` and extra wrapping could be avoided if you used `find` instead of `flatMap`

dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
```diff
@@ -116,6 +143,20 @@ case class DLCExecutor(signer: DLCTxSigner)(implicit ec: ExecutionContext) {
 
 object DLCExecutor {
 
+  def findOutcomeAndSignatures(
```
nkohen, line 149: scaladoc

nkohen (commented):
Generally speaking LGTM, though I think we might want to consider (possibly on a future PR) storing txs/tx templates in the DB so as to remove reliance on creating builders and such (i.e. store funding tx, refund tx, and a CET (where for the CET we can change the payout amounts while keeping the rest the same ... I think)

dlc/src/main/scala/org/bitcoins/dlc/execution/DLCExecutor.scala
benthecarman: then we can't get the oracle outcome

nkohen: in that case this should either be re-written to use an `Option` `var` updated inside of a `find/takeWhile/dropWhile` or else a tail recursive function that goes through `sigCombinations` but returns if it finds something because currently you go through all of `sigCombinations` with `findOutcome` the first of which scales exponentially and the second of which is also a really really costly function call.

Christewart:
Rebase needed, i'm assuming this work is still valuable and relevant though.

PR #2783, opened 2021-03-09
Fix DLCStatus JSON serialzation

### LINKED ISSUE #2746: JSON serialization Incorrect on DLC branch
One of our JSON serializations is broken that is used by the gui for representing DLC Status.

Tests should be added so this doesn't happen again.

benthecarman:
Fixes #2746

`Writer.merge` would call `asInstanceOf[TaggedWriter]` which would cause the error. This fixes it to use the correct writer


nkohen (approved):
ACK assuming you've given this a test on like regtest or something

benthecarman:
> ACK assuming you've given this a test on like regtest or something

Tested it and added tests so we don't regress

PR #2784, opened 2021-03-09
Schnorr sigs for Scala.js

rorp:


crypto-test/.js/src/test/scala/org/bitcoins/crypto/BCryptoECDigitalSignatureTest.scala
```diff
@@ -0,0 +1,19 @@
+package org.bitcoins.crypto
+
+import org.bitcoins.testkitcore.gen.CryptoGenerators
+import org.bitcoins.testkitcore.util.BitcoinSSyncTest
+
+class BCryptoECDigitalSignatureTest extends BitcoinSSyncTest {
+
+  behavior of "BCryptoECDigitalSignatureTest"
+
+  it must "be able to generate valid signatures with bouncy castle" in {
```
nkohen: this isn't using bouncy castle

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -148,49 +151,35 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
...
-  override def add(pk1: ECPublicKey, pk2: ECPublicKey): ECPublicKey = ???
+  override def add(pk1: ECPublicKey, pk2: ECPublicKey): ECPublicKey = {
+    try {
+      val keyBuffer =
+        ecdsa.publicKeyCombine(js.Array(pk1.bytes, pk2.bytes), true)
+      ECPublicKey.fromBytes(keyBuffer)
+    } catch {
+      case _: JavaScriptException =>
+        // if ex.getMessage().contains("Invalid point") =>
+        ECPublicKey.fromHex("00")
```
nkohen: I'm not sure this is correct behavior

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -246,6 +235,25 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
...
+
+      if (decoded.isInfinity())
+        ECPointInfinity
+      else
+        ECPoint(new BigInteger(decoded.getX().toString()),
+                new BigInteger(decoded.getY().toString()))
+    } catch {
+      case _: JavaScriptException =>
+        // if ex.getMessage().contains("Not a point") =>
+        ECPointInfinity
```
nkohen: I don't think this is right to do either

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -26,16 +26,16 @@ object BouncyCastleUtil {
     decodePubKey(point, publicKey.isCompressed)
   }
 
-  def decodePoint(bytes: ByteVector): ECPoint = {
+  def decodePoint(bytes: ByteVector): org.bouncycastle.math.ec.ECPoint = {
```
nkohen, line 29: why is this better than an import?

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
```diff
@@ -42,7 +41,9 @@ trait BouncycastleCryptoRuntime extends CryptoRuntime {
     * @return a tuple (p1, p2) where p1 and p2 are points on the curve and p1.x = p2.x = x
     *         p1.y is even, p2.y is odd
     */
-  def recoverPoint(x: BigInteger): (ECPoint, ECPoint) = {
+  def recoverPoint(x: BigInteger): (
+      org.bouncycastle.math.ec.ECPoint,
+      org.bouncycastle.math.ec.ECPoint) = {
```
nkohen, line 46: same here

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
nkohen: if the problem is collisions with the `crypto.ECPoint` why not just import this as `BouncyCastlePoint` or something like this?

crypto/src/main/scala/org/bitcoins/crypto/ECPoint.scala
```diff
@@ -0,0 +1,29 @@
+package org.bitcoins.crypto
+
+import scodec.bits.ByteVector
+
+import java.math.BigInteger
+
+sealed trait ECPoint
```
nkohen, line 9: scaladocs for this and below

crypto-test/src/test/scala/org/bitcoins/crypto/SignTest.scala
```diff
@@ -16,7 +16,7 @@ class SignTest extends BitcoinSJvmTest {
   behavior of "Sign"
 
   it must "sign arbitrary pieces of data correctly" in {
-    forAllAsync(CryptoGenerators.sha256Digest) { hash =>
+    forAll(CryptoGenerators.sha256Digest) { hash =>
```
nkohen: All three changes made here are unsafe as nothing is being properly tested

crypto-test/.js/src/test/scala/org/bitcoins/crypto/SigningTest.scala
```diff
@@ -0,0 +1,140 @@
...
+  it must "be able to sign and verify Schnorr signatures" in {
+    val privkey = ECPrivateKey.freshPrivateKey
+    val pubkey = privkey.publicKey
+    val msg = BCryptoCryptoRuntime.randomBytes(32)
+    val sig = privkey.schnorrSign(msg)
+    assert(pubkey.schnorrVerify(msg, sig))
+  }
+
+  val lines = Vector(
+    "index,secret key,public key,aux_rand,message,signature,verification result,comment",
```
nkohen: why not put this CSV file in a shared space (and not in the actual test code)?

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
rorp: Which is correct?

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
rorp: Because it works! TM

crypto-test/src/test/scala/org/bitcoins/crypto/SignTest.scala
rorp: Can you elaborate?

crypto-test/.js/src/test/scala/org/bitcoins/crypto/SigningTest.scala
rorp: The code is the only shared space with javascript.

crypto-test/src/test/scala/org/bitcoins/crypto/SignTest.scala
rorp: Oh yes, there are futures inside.

crypto-test/.js/src/test/scala/org/bitcoins/crypto/SigningTest.scala
Christewart: Some more context here, javascript does not give us the ability to read files off the classpath (in `resources/`) -- which we do in the JVM with test files.

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
```diff
@@ -1,86 +1,17 @@
...
-import org.bitcoins.core.crypto.ECPrivateKeyUtil
-import org.bitcoins.testkitcore.gen.{
-  ChainParamsGenerator,
-  CryptoGenerators,
-  NumberGenerator
-}
-import org.bitcoins.testkitcore.util.BitcoinSUnitTest
-
-class ECPrivateKeyTest extends BitcoinSUnitTest {
+class ECPrivateKeyTest extends BitcoinSCryptoTest {
```
nkohen, line 3: Is there a reason the other tests in this file can't be shared?

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -148,49 +151,41 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
...
+    try {
+      val keyBuffer =
+        ecdsa.publicKeyCombine(js.Array(pk1.bytes, pk2.bytes), true)
+      ECPublicKey.fromBytes(keyBuffer)
+    } catch {
+      case ex: JavaScriptException =>
+        // check for infinity
+        if (
+          pk1.isCompressed && pk2.isCompressed &&
+          ((pk1.bytes.head == 0x02 && pk2.bytes.head == 0x03) || (pk1.bytes.head == 0x03 && pk2.bytes.head == 0x02)) &&
```
nkohen: if the pub keys are not compressed they should be compressed prior to this comparison, no reason to fail to add pubkeys together that aren't compressed

nkohen (commented):
With multiple generator files being moved from testkit-core to crypto-test, will this mean that things like testkit will have to depend on crypto-test? That seems quite dirty to me

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/ECPrivateKeyTest.scala
rorp: To avoid circular dependencies between `crypto` and `core`

rorp:
> With multiple generator files being moved from testkit-core to crypto-test, will this mean that things like testkit will have to depend on crypto-test? That seems quite dirty to me

No. They were not moved, I copied the generators that used in `crypto` to remove the dependency to `core` (via `testkit). And `testkit` is independent of `crypto-test`, so again, there's no circular dependencies.

benthecarman (approved):

Issue #2788, opened 2021-03-10
Investigate number of bitcoinds used in test suites


Christewart:
I think some of the problems we experience on CI (particularily mac) are related to the number of bitcoind's that are used in our test suites. We should look to see if there are any unnecessary ones we can remove, or figure out if there are "context free" bitcoinds that can just be pooled and re-used across test suites that don't require specific bitcoind state.

Christewart:
So some more research on this, if we take a look at `FundTransactionHandling`

https://github.com/bitcoin-s/bitcoin-s/blob/0de536de3bdc79615e0599085062f897be0c98c8/wallet-test/src/test/scala/org/bitcoins/wallet/FundTransactionHandlingTest.scala#L18

I believe that a _new_ bitcoind is getting spun up for _every_ test we have in that test suite. If I counted correctly, that means we are using 11 (!) bitcoinds for that suite. 

I would imagine this is done elsewhere as well. This is the problem with using bitcoind in the fixture code.

You can imagine if test suites are run in parallel, how a _ton_ of bitcoinds could be created and the underlying OS might a problem.

Christewart:
Taken care of in #2792

Issue #2701, opened 2021-02-20
OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests


Christewart:
This seems to only occur on our Postgres test suite on CI. I'm _guessing_ that it has something to do with tearing down of fixtures.

> [info] org.bitcoins.node.NeutrinoNodeWithWalletTest *** ABORTED *** (19 seconds, 975 milliseconds)
[info]   akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests
[info]   at akka.http.impl.engine.client.OutgoingConnectionBlueprint$.$anonfun$apply$6(OutgoingConnectionBlueprint.scala:123)
[info]   at akka.http.impl.engine.client.OutgoingConnectionBlueprint$.$anonfun$apply$6$adapted(OutgoingConnectionBlueprint.scala:123)
[info]   at akka.http.impl.util.One2OneBidiFlow$One2OneBidi$$anon$1$$anon$4.onUpstreamFinish(One2OneBidiFlow.scala:97)
[info]   at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:523)
[info]   at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:390)
[info]   at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:625)
[info]   at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:502)
[info]   at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:600)
[info]   at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:773)
[info]   at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:788)
[info]   at akka.actor.Actor.aroundReceive(Actor.scala:537)
[info]   at akka.actor.Actor.aroundReceive$(Actor.scala:535)
[...]

PR #2792, opened 2021-03-11
Implement caching of bitcoind in the walletTest,nodeTest, and partially bitcoindRpcTest project

Christewart:
This is an attempt to reduce the number of bitcoind's used in our test suites. 

From #2788 

>I believe that a new bitcoind is getting spun up for every test we have in that test suite. If I counted correctly, that means we are using 11 (!) bitcoinds for that suite.I would imagine this is done elsewhere as well. This is the problem with using bitcoind in the fixture code. You can imagine if test suites are run in parallel, how a ton of bitcoinds could be created and the underlying OS might a problem.

I'm opening this for now to see how CI reacts, I'm going to try and use this in other test suites as well.

### DIFF
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
@@ -1,10 +1,9 @@
 package org.bitcoins.rpc
 
 import org.bitcoins.core.currency.Bitcoins
-
-import org.bitcoins.rpc.util.{NodeTriple, RpcUtil}
+import org.bitcoins.rpc.util.{NodePair, RpcUtil}
 import org.bitcoins.testkit.rpc.{
-  BitcoindFixturesCachedTriple,
+  BitcoindFixturesCachedPair,
   BitcoindRpcTestUtil
 }
 import org.bitcoins.testkit.util.{BitcoinSAsyncFixtureTest, FileUtil}
@@ -15,14 +14,14 @@ import scala.concurrent.Future
 
 class TestRpcUtilTest
     extends BitcoinSAsyncFixtureTest
-    with BitcoindFixturesCachedTriple {
+    with BitcoindFixturesCachedPair {
 
-  override type FixtureParam = NodeTriple
+  override type FixtureParam = NodePair
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
     val outcomeF: Future[Outcome] = for {
       clients <- clientsF
-      outcome = with3BitcoindsCached(test, clients)
+      outcome = with2BitcoindsCached(test, clients)
       fut <- outcome.toFuture
     } yield fut
 
@@ -32,7 +31,7 @@ class TestRpcUtilTest
   behavior of "BitcoindRpcUtil"
 
   it should "create a temp bitcoin directory when creating a DaemonInstance, and then delete it" in {
-    _: NodeTriple =>
+    _: NodePair =>
       val instance =
         BitcoindRpcTestUtil.instance(RpcUtil.randomPort, RpcUtil.randomPort)
       val dir = instance.datadir
@@ -44,12 +43,12 @@ class TestRpcUtilTest
       assert(!dir.exists)
   }
 
-  it should "be able to generate and sync blocks" in { nodes: NodeTriple =>
-    val NodeTriple(first, second, third) = nodes
+  it should "be able to generate and sync blocks" in { nodes: NodePair =>
+    val NodePair(first, second) = nodes
     for {
       address <- second.getNewAddress
       txid <- first.sendToAddress(address, Bitcoins.one)
-      _ <- BitcoindRpcTestUtil.generateAndSync(Vector(first, second, third))
+      _ <- BitcoindRpcTestUtil.generateAndSync(Vector(first, second))
       tx <- first.getTransaction(txid)
       _ = assert(tx.confirmations > 0)
       rawTx <- second.getRawTransaction(txid)
@@ -60,9 +59,9 @@ class TestRpcUtilTest
   }
 
   it should "ble able to generate blocks with multiple clients and sync inbetween" in {
-    nodes: NodeTriple =>
+    nodes: NodePair =>
       val blocksToGenerate = 10
-      val NodeTriple(first, second, _) = nodes
+      val NodePair(first, second) = nodes
       val allClients = nodes.toVector
       for {
         heightPreGeneration <- first.getBlockCount
@@ -79,8 +78,8 @@ class TestRpcUtilTest
   }
 
   it should "be able to find outputs of previous transactions" in {
-    nodes: NodeTriple =>
-      val NodeTriple(first, second, _) = nodes
+    nodes: NodePair =>
+      val NodePair(first, second) = nodes
       for {
         address <- second.getNewAddress
         txid <- first.sendToAddress(address, Bitcoins.one)
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
@@ -7,15 +7,13 @@ import org.bitcoins.testkit.rpc.{
   BitcoindFixturesCachedPairNewest,
   BitcoindRpcTestUtil
 }
-import org.bitcoins.testkit.util.{BitcoinSAsyncFixtureTest, FileUtil}
+import org.bitcoins.testkit.util.FileUtil
 import org.scalatest.{FutureOutcome, Outcome}
 
 import java.io.File
 import scala.concurrent.Future
 
-class TestRpcUtilTest
-    extends BitcoinSAsyncFixtureTest
-    with BitcoindFixturesCachedPairNewest {
+class TestRpcUtilTest extends BitcoindFixturesCachedPairNewest {
 
   override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
     val outcomeF: Future[Outcome] = for {
--- a/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
+++ b/bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
@@ -1,66 +1,49 @@
 package org.bitcoins.rpc
 
 import org.bitcoins.core.currency.Bitcoins
-import org.bitcoins.rpc.client.common.BitcoindRpcClient
-import org.bitcoins.rpc.util.RpcUtil
-import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
-import org.bitcoins.testkit.util.{BitcoindRpcTest, FileUtil}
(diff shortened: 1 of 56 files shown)


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
```diff
@@ -1,141 +1,54 @@
...
+      assert(dir.getPath().startsWith(scala.util.Properties.tmpDir))
+      assert(
+        dir.listFiles.contains(new File(dir.getAbsolutePath + "/bitcoin.conf")))
+      FileUtil.deleteTmpDir(dir)
+      assert(!dir.exists)
   }
 
-  it should "be able to generate and sync blocks" in {
+  it should "be able to generate and sync blocks" in { nodes: NodePair =>
+    val NodePair(first, second) = nodes
```
nkohen: in general this pattern can be replaced by `case NodePair(first, second) =>`

nkohen: applies below as well

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
```diff
@@ -20,30 +20,44 @@ import org.bitcoins.core.wallet.signer.BitcoinSigner
...
-import scala.async.Async.{async, await}
+//import scala.annotation.nowarn
 import scala.concurrent.duration.DurationInt
-import scala.concurrent.Future
+import scala.concurrent.{Await, Future}
 import scala.reflect.io.Directory
 
-@nowarn
-class WalletRpcTest extends BitcoindRpcTest {
+//@nowarn
```
nkohen: ping

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
```diff
@@ -63,9 +77,9 @@ class WalletRpcTest extends BitcoindRpcTest {
 
   behavior of "WalletRpc"
 
-  it should "be able to dump the wallet" in {
+  it should "be able to dump the wallet" in { nodePair: NodePair =>
+    val client = nodePair.node1
```
nkohen: pattern can be replaced here and elsewhere with
```
case NodePair(client, _) =>
```

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/WalletRpcTest.scala
```diff
@@ -215,44 +230,9 @@ class WalletRpcTest extends BitcoindRpcTest {
...
-
-    val balance = await(otherClient.getBalance)
-    assert(balance == balanceBefore)
-  }
-
-  it should "be able to list address groupings" in {
-
+  it should "be able to list address groupings" in { nodePair: NodePair =>
+    val client = nodePair.node1
+    val otherClient = nodePair.node2
```
nkohen: nit: `case NodePair(client, otherClient) =>`

nkohen: same below

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v17/PsbtRpcTest.scala
```diff
@@ -13,23 +13,39 @@ import org.bitcoins.core.protocol.transaction.{
...
+      fut <- futOutcome.toFuture
+    } yield fut
+
+    new FutureOutcome(f)
+  }
+
   // https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#Test_Vectors
-  it should "decode all the BIP174 example PSBTs" in {
+  it should "decode all the BIP174 example PSBTs" in { nodePair: NodePair =>
+    val client = nodePair.node1
```
nkohen: ditto et al

node-test/src/test/scala/org/bitcoins/node/networking/peer/PeerMessageHandlerTest.scala
```diff
@@ -1,42 +1,54 @@
...
+    val peerHandlerF = NodeUnitTest.buildPeerHandler(peer)
     val peerMsgSenderF = peerHandlerF.map(_.peerMsgSender)
     val p2pClientF = peerHandlerF.map(_.p2pClient)
 
-    val _ =
-      bitcoindPeerF.flatMap(_ => peerHandlerF.map(_.peerMsgSender.connect()))
-
+    //val _ =
+    //  bitcoindPeerF.flatMap(_ => peerHandlerF.map(_.peerMsgSender.connect()))
+    val _ = peerHandlerF.map(_.peerMsgSender.connect())
```
nkohen: ping

testkit/src/main/scala/org/bitcoins/testkit/node/NodeTestWithCachedBitcoind.scala
```diff
@@ -0,0 +1,184 @@
...
+  CachedBitcoindNewest,
+  CachedBitcoindV19
+}
+import org.bitcoins.testkit.wallet.BitcoinSWalletTest
+import org.bitcoins.wallet.WalletCallbacks
+import org.scalatest.FutureOutcome
+
+import scala.concurrent.Future
+
+trait NodeTestWithCachedBitcoind extends BaseNodeTest { _: CachedBitcoind =>
```
nkohen: scaladocs in this file

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
```diff
@@ -0,0 +1,185 @@
...
+import org.bitcoins.rpc.client.v21.BitcoindV21RpcClient
+import org.bitcoins.rpc.util.{NodePair, NodeTriple}
+import org.bitcoins.testkit.EmbeddedPg
+import org.bitcoins.testkit.fixtures.BitcoinSFixture
+import org.bitcoins.testkit.util.BitcoinSAsyncFixtureTest
+import org.scalatest.{FutureOutcome, Outcome}
+
+import scala.concurrent.Future
+
+trait BitcoindFixtures extends BitcoinSFixture with EmbeddedPg {
```
nkohen, line 19: scaladocs

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/TestRpcUtilTest.scala
Christewart: Unfortunately this doesn't work with scala 2.12

Christewart:
So i think i've got to the bottom of #2701 in d91c627


If you take a look at the implemented of [`Client.isStoppedF`](https://github.com/bitcoin-s/bitcoin-s/blob/master/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala#L252) you will see we call `isStartedF` and then just negate the output from that. 

If you look at how [`Client.isStartedF`](https://github.com/bitcoin-s/bitcoin-s/blob/b8140053b916b0da109810b75f336a21b0199fc0/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala#L196) is implemented, you see it attempts to ping bitcoind and checks if we get a response. When we are tearing down test suites, we may be shutting down the actor system while a `ping` request is still pending in the queue. 

Now we just use a `AtomicBoolean` to indicate that the user has requested shutdown, and we shouldn't try to ping bitcoind.

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  override def afterAll(): Unit = {
+    super[BitcoindFixturesCachedPair].afterAll()
+    super[BitcoinSAsyncFixtureTest].afterAll()
+  }
+}
+
+/** Bitcoind fixtures with two cached bitcoind rpc clients that are [[BitcoindVersion.newest]] that are connected via p2p */
+trait BitcoindFixturesCachedPairNewest
```
benthecarman, line 282: nit: Maybe just name this `BitcoindFixturesCachedPair`, the `Newest` was confusing to me and its kinda implied by leaving it out

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSWalletTest.scala
```diff
@@ -26,132 +25,40 @@ import org.bitcoins.rpc.client.common.{BitcoindRpcClient, BitcoindVersion}
...
-
-trait BitcoinSWalletTest extends BitcoinSFixture with EmbeddedPg {
+import scala.util.{Failure, Success}
+
+trait BitcoinSWalletTest
+    extends BitcoinSWalletTestCachedBitcoind
+    with BaseWalletTest
+    with BitcoinSFixture
+    with EmbeddedPg
+    with CachedBitcoind[BitcoindRpcClient] {
```
benthecarman: This is making every wallet test require a bitcoind. We currently have a bunch of tests that don't use a bitcoind, I believe this would slow down those tests

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v17/BitcoindV17RpcClientTest.scala
```diff
@@ -53,149 +59,155 @@ class BitcoindV17RpcClientTest extends BitcoindRpcTest {
...
-    } yield assert(info.address == addr)
+    nodePair: FixtureParam =>
+      val NodePair(client, _) = nodePair
+      for {
+        addr <- client.getNewAddress(addressType = AddressType.Legacy)
+        info <- client.getAddressInfo(addr)
+      } yield assert(info.address == addr)
   }
 
   // needs #360 to be merged
```
benthecarman: nit: this comment can be removed

benthecarman (commented):
couple nits, otherwise LGTM

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
Christewart: Unfortunately that is the name of the parent trait that is extended by all versions. I would have to come up with a different name for that (any suggestions?)

benthecarman: Maybe just change `newest` to `lastest` that sounds better imo

Christewart: The reason I used `newest` is because we have the field named that 

https://github.com/bitcoin-s/bitcoin-s/blob/89185338e04b413d64cfd3175a917e816601246a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala#L280

I can still change if you want, but i thought i would provide the context

benthecarman (approved):

PR #2794, opened 2021-03-13
Adaptor signatures for Scala.js

rorp:


benthecarman (approved):
Looks correct to me, however, I am not very familiar with the spec for this

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -109,11 +109,17 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
       signature: ECDigitalSignature,
       message: ByteVector): (ECPublicKey, ECPublicKey) = {
     val keyBytes =
-      ecdsa.recover(message, signature.bytes, param = 0, compress = true)
+      if (signature.isDEREncoded)
+        ecdsa.recoverDER(message, signature.bytes, param = 0, compress = true)
```
Christewart, line 112: what is `param= 0` for ?

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  // Compute s' = k^-1 * (dataToSign + rx*privateKey)
+  private def adaptorSignHelper(
+      dataToSign: ByteVector,
+      k: FieldElement,
+      r: ECPublicKey,
+      privateKey: ECPrivateKey): FieldElement = {
+    CryptoUtil.decodePoint(r) match {
+      case ECPointInfinity =>
+        throw new IllegalArgumentException("Invalid point")
```
Christewart: nit: 

```scala
throw new IllegalArgumentException(s"Invalid point, got ${ECPointInfinity}")
```

Generally putting the input in the error message is _really_ helpful when debugging.

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
```diff
@@ -134,6 +134,8 @@ trait CryptoRuntime {
 
   def publicKey(privateKey: ECPrivateKey): ECPublicKey
 
+  def publicKeyConvert(key: ECPublicKey, compressed: Boolean): ECPublicKey
```
Christewart, line 140: Scaladoc please, what is it converting to?

If I understand what this method is meant to do, i would recommend removing the `compressed` parameter and calling it 

```scala
def toCompressedPublicKey(key: ECPublicKey): ECPublicKey
``` 

?

crypto/src/main/scala/org/bitcoins/crypto/DERSignatureUtil.scala
```diff
@@ -37,12 +37,16 @@ sealed abstract class DERSignatureUtil {
       val thirdByteIs0x02 = bytes(2) == 0x02
       //this is the size of the r value in the signature
       val rSize = bytes(3)
-
-      //this 0x02 separates the r and s value )in the signature
-      val second0x02Exists = bytes(rSize + 4) == 0x02
-
-      firstByteIs0x30 && signatureLengthIsCorrect && thirdByteIs0x02 &&
-      second0x02Exists
+      val sOffset = rSize + 4
```
Christewart, line 40: What is the bug here? Just an `OutOfBoundsException` if an invalid DER encoded `ByteVector` is passed into this function?

Christewart (commented):
Also looking very good! Mostly nits

crypto/.js/src/main/scala/org/bitcoins/crypto/ECDSA.scala
```diff
@@ -41,11 +41,21 @@ class ECDSA(
...
       msg: Buffer,
       sig: Buffer,
       param: Byte,
       compress: Boolean): Buffer = js.native
 
+  def recoverDER(
+      msg: Buffer,
+      sig: Buffer,
+      param: Byte,
+      compress: Boolean): Buffer = js.native
```
nkohen, line 58: Seeing as DER is the default in Bitcoin, I find it concerning to have `recover` and `verify` and `sign` not be doing DER-related things?

PR #2805, opened 2021-03-17
Schnorr js

rorp:


core-test/src/test/scala/org/bitcoins/core/crypto/ECPrivateKeyTest.scala
```diff
@@ -0,0 +1,79 @@
+package org.bitcoins.core.crypto
```
Christewart: So if I understand this correctly, this means we have _no_ `ECPrivateKeyTest` inside of the `cryptoTest` project? 

It seems like a really important test suite not to have in `cryptoTest`.

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
```diff
@@ -146,13 +148,29 @@ trait CryptoRuntime {
       data: ByteVector,
       signature: ECDigitalSignature): Boolean
 
-  def decompressed(publicKey: ECPublicKey): ECPublicKey
+  def decompressed(publicKey: ECPublicKey): ECPublicKey = {
+    if (publicKey.isCompressed) {
+      decodePoint(publicKey.bytes) match {
+        case ECPointInfinity => ECPublicKey.fromHex("00")
```
Christewart, line 154: Maybe just do 

```scala
case object ECPointInfinity extends ECPoint {
  val publicKey: ECPublicKey = ECPublicKey.fromHex("00")
}
```

We might even just do 

```scala
sealed trait ECPoint {
  def publicKey: ECPublicKey
}
```

so that every point we have can just access the public key by `point.publicKey`?

crypto-test/src/test/scala/org/bitcoins/crypto/BitcoinSCryptoTest.scala
```diff
@@ -0,0 +1,118 @@
...
+  /** Sets the generator driven tests to perform the given amount of execs */
+  def customGenDrivenConfig(executions: Int): PropertyCheckConfiguration = {
+    PropertyCheckConfiguration(
+      minSuccessful = PosInt.from(executions).get,
+      minSize = PosInt.from(executions).get,
+      workers = 1
+    )
+  }
+
+  def forAllAsync[A](gen: Gen[A])(
```
Christewart, line 60: Ok, so just for my own knowledge, was the problem with the previous implemenation was the fact that it used `java.util.concurrent.CopyOnwriteList` ? 

https://github.com/bitcoin-s/bitcoin-s/blob/ac78b60ea0e3e659a0729646074daba0f6f31eb5/testkit-core/src/main/scala/org/bitcoins/testkitcore/util/BaseAsyncTest.scala#L94

If so, i vote that we get rid of the old implementation and replace it with this one. Your implementation seems superior

crypto-test/.js/src/test/scala/org/bitcoins/crypto/KeysTest.scala
```diff
@@ -18,4 +17,62 @@ class KeysTest extends AnyFlatSpec with Matchers {
...
+    assert(BCryptoCryptoRuntime.isValidPubKey(decompressed.bytes))
+    assert(!decompressed.isCompressed)
+
+    assert(pubkey.bytes != converted.bytes)
+    assert(compressed.bytes == converted.bytes)
+    assert(compressed.bytes != decompressed.bytes)
+    assert(pubkey.bytes == decompressed.bytes)
+  }
+
+  private val inf = ECPublicKey.fromHex("00")
```
Christewart, line 50: nit: Can we put this inside of 

```scala 
object ECPublicKey {
  val inf = ECPublicKey.fromHex("00")
} 
```
or if you take my other suggestion 

```scala
object ECPublicKey { 
  val inf = ECPointInfinity.publicKey
}
```

crypto-test/.js/src/test/scala/org/bitcoins/crypto/RandomTest.scala
```diff
@@ -1,9 +1,6 @@
 package org.bitcoins.crypto
 
-import org.scalatest.flatspec.AnyFlatSpec
-import org.scalatest.matchers.must.Matchers
-
-class RandomTest extends AnyFlatSpec with Matchers {
+class RandomTest extends BitcoinSCryptoTest {
 
   it should "generate random bytes" in {
```
Christewart, line 5: Any reason not to do a scalacheck style test with `forAll()` ?

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/JvmECPublicKeyTest.scala
```diff
@@ -1,10 +1,7 @@
 package org.bitcoins.crypto
 
-import org.bitcoins.testkitcore.gen.CryptoGenerators
-import org.bitcoins.testkitcore.util.BitcoinSUnitTest
-
 /** Public key tests specific to the JVM */
-class JvmECPublicKeyTest extends BitcoinSUnitTest {
+class JvmECPublicKeyTest extends BitcoinSCryptoTest {
```
Christewart, line 4: nit: On line 16, we should probably use `fail()` if we have an unexpected crypto runtime rather than `succeed`?

testkit-core/src/main/scala/org/bitcoins/testkitcore/util/BaseAsyncTest.scala
```diff
@@ -30,8 +30,7 @@ trait BaseAsyncTest
     with BeforeAndAfterAll
     with Matchers
     with ScalaCheckPropertyChecks
-    with AsyncTimeLimitedTests
-    with BitcoinSLogger { this: AsyncTestSuite =>
+    with AsyncTimeLimitedTests { this: AsyncTestSuite =>
```
Christewart: Is this still needed to remove `BitcoinSLogger` since the trait is not used in `cryptoTest`?

Christewart (commented):
Looking really good! Nice job.

core-test/src/test/scala/org/bitcoins/core/crypto/ECPrivateKeyTest.scala
nkohen: If I'm not mistaken all this tests is WIF encoding, likely the solution here is a rename

crypto-test/.js/src/test/scala/org/bitcoins/crypto/KeysTest.scala
nkohen: nit: if we have a val like this it shouldn't be called `inf` which is short in math for `infimum` not `infinity`

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -246,6 +257,22 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
             s"Need $len bytes for buffer -> bytevector conversion")
     ByteVector(accum.map(_.toByte))
   }
+
+  override def decodePoint(bytes: ByteVector): ECPoint = {
+    if (bytes.size == 1 && bytes(0) == 0x00) {
```
nkohen, line 267: nit: `if (bytes == ByteVector(0))`

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
```diff
@@ -26,16 +26,18 @@ object BouncyCastleUtil {
...
 
-  def decodePoint(pubKey: ECPublicKey): ECPoint = {
+  private[crypto] def decodePoint(
+      pubKey: ECPublicKey): org.bouncycastle.math.ec.ECPoint = {
     decodePoint(pubKey.bytes)
   }
 
-  def decodePubKey(
-      point: ECPoint,
+  private[crypto] def decodePubKey(
```
nkohen, line 39: Is there a reason these are now private?

crypto/src/main/scala/org/bitcoins/crypto/ECPoint.scala
```diff
@@ -0,0 +1,37 @@
+package org.bitcoins.crypto
+
+import scodec.bits.ByteVector
+
+import java.math.BigInteger
+
+/** Represents a point on an elliptic curve.
```
nkohen: nit: specifically the curve `secp256k1`

nkohen (approved):
Not sure I fully understand all the decisions about what goes where but I didn't notice any functionality diff (which is to say, the linking looks fine to me) and if there's anything wrong placement-wise it can always be changed later and at this point I'd rather get past the functional code freeze on DLC stuff asap.

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
rorp: I believe it's to much of overhead for a one of the most used methods to allocate at least 21 bytes just to check if one byte is set to zero.

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncyCastleUtil.scala
rorp: We don't want to expose Bouncy Castle outside.

PR #2808, opened 2021-03-19
Added invariants to ContractOraclePair classes …

nkohen:
… so that contract descriptor to oracle announcement consistency is now enforced.


Christewart (approved):
Probably want @benthecarman to take a look too

benthecarman (approved):

PR #2810, opened 2021-03-19
Remove logging from core

### LINKED ISSUE #2809: Remove logs from core
We cannot take pre-compiled jars and use them in a scalajs environment. This means we have to use a scalajs aware dependency (`%%%`) to support logging in core. One of the dependency we currently use in core is [`slf4j-api`](https://github.com/bitcoin-s/bitcoin-s/blob/master/project/Deps.scala#L259-L260). This dependency is meant to be a a logging interface that any logging implementation can use to implement their custom features for their logger (we use [`logback`](http://logback.qos.ch/), which implements the slf4j-api).

As of this writing, the current plan is just to get the `core` and `crypto` modules working with scalajs. AFAIK, `crypto` didn't have any logs to begin with. We might just want to remove all logs in core, as I don't think they are being used anywhere? Mostly we just have logs turned off. 

Is there any objections to doing this?

benthecarman:
Closes #2809

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
@@ -1,5 +1,6 @@
 package org.bitcoins.gui.dlc
 
+import grizzled.slf4j.Logging
 import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.{CliCommand, Config, ConsoleCli}
 import org.bitcoins.commons.serializers.Picklers._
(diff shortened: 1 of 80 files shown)


Christewart (changes requested):
I'm realizing i wasn't clear in the issue -- grizzled.sl4fj also does not support scalajs (no `%%%` in `Deps.scala`)

http://software.clapper.org/grizzled-slf4j/

Christewart (approved):

Christewart (changes requested):
`coreTest` needs to be scala-js compatible as well which means logging needs to be removed

benthecarman:
> `coreTest` needs to be scala-js compatible as well which means logging needs to be removed

Already done, there was no logs in core, but did remove the dep on testkit core for slf4j

Christewart:
> > `coreTest` needs to be scala-js compatible as well which means logging needs to be removed
> 
> Already done, there was no logs in core, but did remove the dep on testkit core for slf4j

Oh ok, i see it now. The reason I thought this was because `coreTest` has logback as a dep 

https://github.com/benthecarman/bitcoin-s/blob/ae9d7bd0df9f97e4793f6c3a54dc80610fe00d8d/project/Deps.scala#L285

If we remove that, this PR LGTM

Christewart (approved):

PR #2818, opened 2021-03-23
Move hard coded test vectors from resource files into scala files

### LINKED ISSUE #2812: Figure out how to use coreTest hard coded test vectors in scala-js
### Reading the files

Scalajs explains that resources cannot be loaded from the classpath in javascript because there is no classpath in javascript. 

In `coreTest`, we have quite a few test files

https://github.com/bitcoin-s/bitcoin-s/tree/master/core-test/.jvm/src/test/resources

as of  e9cac3d8b19c5c91e7423dc7528a74bef07b1c95 these live in `core-test/.jvm` which means they just get run on the jvm runtime, not scalajs. Ideally we would like to have these run on both runtimes.

For smaller files, we can just hardcode them in a `.scala` file similar [to what we did with the BIP340 test vectors](https://github.com/bitcoin-s/bitcoin-s/pull/2805/files#diff-8189a916fbe47acdba8c97440c80b2a3def50f1586e755fc7509df12764331f5R5)

### Parsing the test vectors

Most of these test vectors are in json. [We have two separate json parsing dependencies in `core-test`](https://github.com/bitcoin-s/bitcoin-s/blob/b8140053b916b0da109810b75f336a21b0199fc0/project/Deps.scala#L288-L289). 

1. spray-json
2. play json

The spray-json is an ancient depdendency we should try to remove, it was used in the early days of bitcoin-s to get the script test suite passing. 

play-json is not scalajs aware, so we need to drop that too for something like micropickle.We probably need to re-write the parsing code to be scalajs friendly

benthecarman:
Fixes #2812

Just copies the strings and brings them into the actual scala files.

Scala has a limit of 65535 chars per string literal so some are broken up into multiple strs and then combined.

`BlockTest` may need a different solution as that file is now over 4Mb


Christewart (approved):
So this LGTM

>BlockTest may need a different solution as that file is now over 4Mb

I would be in favor of just keeping the `BlockTest` test vectors in the `.jvm` specific folder in there own files. 

I guess they would just get packaged up into test artifactos (`coreTest` jar file) but it still seems unnecessary to run those tests on scalajs specifically.

PR #2509, opened 2021-01-13
Upgrade to scalac 2.12.13

Christewart:
https://github.com/scala/scala/releases/tag/v2.12.13

PR #2822, opened 2021-03-24
Silence scalajs warnings for org.bitcoins.crypto package

Christewart:
### Problem

Currently we have a bunch of warnings that look like this 

>[warn] /home/chris/dev/bitcoin-s/crypto/.js/src/main/scala/org/bitcoins/crypto/Buffer.scala:41:47: dead code following this construct
[warn]   def this(str: String, encoding: String = js.native) = this()
[warn]                                               ^
[warn] /home/chris/dev/bitcoin-s/crypto/.js/src/main/scala/org/bitcoins/crypto/Buffer.scala:115:29: dead code following this construct
[warn]       targetStart: Int = js.native,
[warn]                             ^
....

### Solution

This adjusts the compiler settings for the `cryptoJS` project to ignore all of the warnings that occur in the `facade` package. 

As a by product, this PR moves all scalajs facades into the `facade` package.

Now we use the `-Wconf` setting in the scala compiler with some filters to silence all `unused` and `w-flag-dead-code` warnings in the package `org.bitcoins.crypto.facade` 

### Heavy use of implicits in `BCryptoRuntime` 

part of the reason this diff is so big, is that we used 2 implicit methods to convert from `Buffer` <-> `ByteVector`. These were used _alot_. 

I've removed them and expllicitly called `CryptoJsUtil.toByteVector()` and `CryptoJsUtil.toNodeBuffer()`. The use of these two implicits were causing compiler warnings

https://github.com/bitcoin-s/bitcoin-s/pull/2822/files#diff-6952fb2433325cf12e395c8f8ed56815f78b3258b37f0a0f5c567cf79b534cc5L17

#### Resources 

https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html


nafg:
See https://gist.github.com/lrytz/b0cd674c30e141cb71cc35ce212c43b1#file-wconf-md

You can use src for the source file name. Also, site is a regex against the fully qualified class name so you can be more specific if you want.

Christewart:
Hi @nafg 

> See https://gist.github.com/lrytz/b0cd674c30e141cb71cc35ce212c43b1#file-wconf-md

I think the resource you linked is the same on the scala website here which i have been attempting to follow

https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html#warnings-from-the-scala-compiler

> You can use src for the source file name. Also, site is a regex against the fully qualified class name so you can be more specific if you want.

Maybe I am missing something, but this doesn't work at all for me. This is the specific file i'm trying to exclude, what is the incantation to do so? 

https://github.com/bitcoin-s/bitcoin-s/blob/master/crypto/.js/src/main/scala/org/bitcoins/crypto/Buffer.scala

Christewart:
These errors are occurring because `-Wconf` wasn't backported to the scalac `2.12.x` series until the latest release `2.12.13`. 

We cannot upgrade to 2.12.13 until #2509 is in

Christewart:
> These errors are occurring because `-Wconf` wasn't backported to the scalac `2.12.x` series until the latest release `2.12.13`.
> 
> We cannot upgrade to 2.12.13 until #2509 is in

So for now I added a filter to check that the scala version is `2.13`. If it is apply the compiler flags, if it isn't, don't do anything. 

When we get #2509, we can remove this filter.

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -27,32 +36,38 @@ trait BCryptoCryptoRuntime extends CryptoRuntime {
   private lazy val ecdsa =
     new ECDSA("SECP256K1", sha256, js.constructorOf[SHA256], null)
 
-  private lazy val randomBytesFunc: Int => ByteVector =
+  private lazy val randomBytesFunc: Int => ByteVector = { int =>
```
Christewart, line 39: previously nothing was done with the `int` passed into the function, i assume this is what was intended right @rorp ?

crypto/.js/src/main/scala/org/bitcoins/crypto/BCryptoCryptoRuntime.scala
```diff
@@ -1,11 +1,24 @@
...
+  Hash160,
+  Random,
+  RandomBrowser,
+  RipeMd160,
+  SHA1,
+  SHA256,
+  SHA256Factory,
+  SHA512,
+  SipHash
+}
```
benthecarman, line 16: nit: `._`

benthecarman (approved):

PR #2824, opened 2021-03-25
Disjoint Union DLC

nkohen:
Implemented disjoint union DLC related data structures. Also refactored dlcTest to have a `DLCTest` trait with all of the constructors and parameter case classes and broke up `DLCClientTest` into separate test files for each kind of DLC. Tests in dlcTest now pass for me in under 40 seconds!

`dlcWallet` will need some restructuring work (@benthecarman).


core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
```diff
@@ -127,167 +263,149 @@ case class ContractInfo(
...
       totalCollateral,
       ContractOraclePair.fromDescriptorOracle(contractDescriptor, oracleInfo))
   }
 }
+
+case class DisjointUnionContractInfo(contracts: Vector[SingleContractInfo])
+    extends ContractInfo
+    with TLVSerializable[ContractInfoV1TLV] {
+
+  override val totalCollateral: Satoshis = contracts.head.totalCollateral
```
benthecarman: nit: should add a `require(contracts.nonEmpty)` so a better error message is given

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractInfo.scala
```diff
@@ -127,167 +263,149 @@ case class ContractInfo(
...
+
+  override def toTLV: ContractInfoV1TLV = {
+    ContractInfoV1TLV(
+      totalCollateral,
+      contracts.map(contract =>
+        (contract.contractDescriptor.toTLV, contract.oracleInfo.toTLV)))
+  }
+
+  /** @inheritdoc */
+  override val max: Satoshis = contracts.map(_.max).maxBy(_.toLong)
```
benthecarman: You should be able to use `.max` instead of `.maxBy`

benthecarman (commented):
couple nits

Christewart:
needs rebase

Christewart:
superseded #3839

PR #2826, opened 2021-03-25
Get all of Core working with JS

benthecarman:
fixes #2826


core/.jvm/src/main/scala/org/bitcoins/core/protocol/dlc/SigPointComputer.scala
```diff
@@ -0,0 +1,14 @@
...
+import org.bitcoins.crypto.ECPublicKey
+
+import scala.concurrent._
+import scala.concurrent.duration.DurationInt
+
+case class SigPointComputer(private val computeSigPoint: () => ECPublicKey) {
+  private val sigPointF = Future(computeSigPoint())(ExecutionContext.global)
+
+  lazy val compute: ECPublicKey = {
+    Await.result(sigPointF, 10.seconds)
```
nkohen: This meant to be 20 seconds, my bad as I sent this to you

project/Deps.scala
```diff
@@ -280,6 +284,16 @@ object Deps {
   )
 
   def coreTest = Def.setting {
+    List(
+      Test.junitInterface,
+      Test.scalaTest.value,
+      //      Test.spray,
+      //      Test.playJson,
```
nkohen: If this isn't needed can it be deleted?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
```diff
@@ -21,8 +21,8 @@ sealed abstract class Number[T <: Number[T]]
   /** The underlying scala number used to to hold the number */
   protected def underlying: A
 
-  def toInt: Int = toBigInt.bigInteger.intValueExact()
-  def toLong: Long = toBigInt.bigInteger.longValueExact()
+  def toInt: Int = toBigInt.intValue
```
Christewart: Why is this change required? the purpose of using the `Exact()` suffix is so that an error is explicitly thrown in the case of overflows rather than silently "failing" and going back to `Long.MinValue` or `Int.MinValue`

This is especially concerning in a new environment. IIRC the javascript runtime supports different types of numbers, so i would rather these errors be expllicit rather than silently failing.

core/src/main/scala/org/bitcoins/core/p2p/InetAddress.scala
```diff
@@ -0,0 +1,31 @@
+package org.bitcoins.core.p2p
+
+import scodec.bits._
+
+trait InetAddress {
```
Christewart, line 5: Any test cases for this?

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/LnCurrencyUnit.scala
```diff
@@ -51,9 +51,9 @@ sealed abstract class LnCurrencyUnit
 
   def toBigInt: BigInt
 
-  def toLong: Long = toBigInt.bigInteger.longValueExact()
+  def toLong: Long = toBigInt.toLong
```
Christewart: same comment here as `NumberType`

core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala
```diff
@@ -34,7 +34,7 @@ sealed abstract class MilliSatoshis
 
   def toBigInt: BigInt = underlying
 
-  def toLong: Long = toBigInt.bigInteger.longValueExact
+  def toLong: Long = toBigInt.toLong
```
Christewart: same comment here

Christewart (changes requested):
So there is a lot of test case that are being moved to `core/.jvm/` such as the sighash test cases, base58 test cases. 

I thought we moved these into `.scala` files so that they can run on both platforms?

core/src/main/scala/org/bitcoins/core/number/NumberType.scala
benthecarman: scalajs doesn't have access to these functions for whatever reason, maybe would be best to create our own version in `core/package`

benthecarman:
> So there is a lot of test case that are being moved to `core/.jvm/` such as the sighash test cases, base58 test cases.
> 
> I thought we moved these into `.scala` files so that they can run on both platforms?

Oh, need to move back and them implement with Pickler

benthecarman:
> > So there is a lot of test case that are being moved to `core/.jvm/` such as the sighash test cases, base58 test cases.
> > I thought we moved these into `.scala` files so that they can run on both platforms?
> 
> Oh, need to move back and them implement with Pickler

Looks like we don't have a json reader for scalajs, maybe table for now until we figure out how to do this?

core/src/main/scala/org/bitcoins/core/package.scala
```diff
@@ -31,6 +32,19 @@ package object core {
     }
   }
 
+  implicit class bigIntegerUtil(private val bigInt: BigInteger) extends AnyVal {
```
Christewart, line 35: added but not used?

core/src/main/scala/org/bitcoins/core/package.scala
```diff
@@ -31,6 +32,19 @@ package object core {
...
+
+    implicit def intExact: Int = {
+      if (bigInt.bitLength() <= 31) bigInt.intValue()
```
Christewart, line 38: could we just use a `require(bigInt == bigInt.intValue())` 

Same for below

core-test/.jvm/src/test/scala/org/bitcoins/core/p2p/InetAddressTest.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import java.net.{InetAddress => JvmAddress}
+
+class InetAddressTest extends BitcoinSUnitTest {
+
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    generatorDrivenConfigNewCode
+
+  behavior of "InetAddressTest"
+
```
Christewart: is there any hard coded tests we can just add that is run across both runtimes? 

Even some simple ones would be good, because with these tests they can't be run on js

core/src/main/scala/org/bitcoins/core/package.scala
benthecarman: These are just copied from the actual java implementations, I am guessing this is more effienct

core/src/main/scala/org/bitcoins/core/package.scala
benthecarman: These are used in places like https://github.com/bitcoin-s/bitcoin-s/pull/2826/files#diff-f228c4c619e8243a6789f3b54ec0613dcbe13dec050796d590880a09471b9b03R25

core-test/.jvm/src/test/scala/org/bitcoins/core/p2p/InetAddressTest.scala
benthecarman: Added some

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
```diff
@@ -142,10 +142,15 @@ trait Sign extends AsyncSign {
       signWithEntropy(bytes, startBytes)
     }
 
-    if (sig.bytes.length <= 70) {
-      sig
-    } else {
-      signLowR(bytes, startAt + 1)
+    // If we are using BCrypto then we can't sign with entropy
+    CryptoUtil.cryptoContext match {
```
Christewart, line 146: maybe we should file an issue for this? At least we can track it there and maybe can fix this in the future

Christewart (approved):

Christewart (changes requested):
Need to add `coreTestJS/test` to workflows

crypto/src/main/scala/org/bitcoins/crypto/Sign.scala
Christewart: #2839

Christewart (approved):

PR #2831, opened 2021-03-28
Fix potential unordered nonces in announcement

benthecarman:
We weren't enforcing the `nonceIndex` ordering on the pending decomp events so it was possible that the nonces could be returned in the incorrect order. This fixes that and adds a test to make sure (test fails without the fix).

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
@@ -243,9 +243,11 @@ object OracleEvent {
         require(eventDbs.forall(_.attestationOpt.isEmpty),
                 "Cannot have a partially signed event")
 
+        val sortedEventDbs = eventDbs.sortBy(_.nonceIndex)
+
         PendingDigitDecompositionV0OracleEvent(
           eventDb.pubkey,
-          eventDbs.map(_.nonce),
+          sortedEventDbs.map(_.nonce),
           eventDb.eventName,
           eventDb.signingVersion,
           eventDb.maturationTime,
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
```diff
@@ -243,9 +243,11 @@ object OracleEvent {
         require(eventDbs.forall(_.attestationOpt.isEmpty),
                 "Cannot have a partially signed event")
 
+        val sortedEventDbs = eventDbs.sortBy(_.nonceIndex)
+
```
Christewart, line 247: Is it worth just adding an invariant to `PendingDigitDecompisitionEvent()`  ?  Or in the superclass `OracleEvent`?

benthecarman: We can't really because we don't actually save the event dbs or the nonce index in the oracle event, just use the information from the actual eventdbs

Christewart: So my understanding of the purpose of `OracleEvent.fromEventDbs()` is to take in a `Vector[EventDb]` and spit out an in memory respresentation for that event (`OracleEvent`). 

If we add an invariant, for the lifetime of that `OracleEvent` in memory, we KNOW that the nonces are ordered correctly. This patch does that in a similar way, but if we ever have `OracleEvent.fromEventDbs2()` (because we need some new feature or whatever) we need to make sure we propogate this patch here. 

Same thing applies on the oracle explorer side, we need to remember to patch any custom code we have for reading `EventDb`s out of the database -- rather than just having an exception thrown that will remind us during development.

Christewart: > We can't really because we don't actually save the event dbs or the nonce index in the oracle event, just use the information from the actual eventdbs

I think this qualifies as "saving"? It guarantees their in memory representation is ordered

>sortedEventDbs.map(_.nonce),

Christewart (approved):

PR #2836, opened 2021-03-30
Inital LND rpc with some tests

benthecarman:
Still have lots of functions to give official methods as well as needing to add more extensive tests. However, would like some intial review


.github/workflows/Linux_2.12_RPC_Tests.yml
```diff
@@ -27,4 +27,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: run tests
-        run: sbt ++2.12.12 downloadBitcoind downloadEclair coverage bitcoindRpcTest/test bitcoindRpc/coverageReport bitcoindRpc/coverageAggregate bitcoindRpc/coveralls eclairRpcTest/test eclairRpc/coverageReport eclairRpc/coverageAggregate eclairRpc/coveralls
\ No newline at end of file
+        run: sbt ++2.12.12 downloadBitcoind downloadEclair downloadLnd coverage bitcoindRpcTest/test bitcoindRpc/coverageReport bitcoindRpc/coverageAggregate bitcoindRpc/coveralls eclairRpcTest/test eclairRpc/coverageReport eclairRpc/coverageAggregate eclairRpc/coveralls lndRpcTest/test
```
Christewart, line 30: Can we change the name on this CI matrix above to add lnd

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/config/LndConfig.scala
```diff
@@ -0,0 +1,268 @@
...
+    }.headOption
+  }
+
+  /** Converts a string to an InetSocketAddress */
+  private def toInetSocketAddress(string: String): InetSocketAddress = {
+    val uri = new URI(string)
+    new InetSocketAddress(uri.getHost, uri.getPort)
+  }
+
+  lazy val bitcoindUser: String = getValue("bitcoind.rpcuser").getOrElse("foo")
```
Christewart: Probably should just use `.get` so that we throw in the case it isn't defined.

Let's fail fast rather than use a bad config that probably will fail else where.

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/config/LndConfig.scala
```diff
@@ -0,0 +1,268 @@
...
+
+object LndConfig extends Logging {
+
+  /** The empty `lnd` config */
+  lazy val empty: LndConfig = LndConfig("", DEFAULT_DATADIR)
+
+  /** Constructs a `lnd` config from the given string,
+    * by splitting it on newlines
+    */
+  def apply(config: String, datadir: File): LndConfig =
```
Christewart: Can we get some sort of common interface across all this? I believe this is the same with bitcoind/eclair too 

https://github.com/bitcoin-s/bitcoin-s/blob/f593cf063e53b2c8e6b867d5c975d824694b31cd/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala#L123

Christewart: This is sort of similar, but is meant for `AppConfig` https://github.com/bitcoin-s/bitcoin-s/blob/ee7c96245e99325d1386464c4462d36dce7bcafb/db-commons/src/main/scala/org/bitcoins/db/AppConfigFactory.scala#L9

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/config/LndConfig.scala
```diff
@@ -0,0 +1,268 @@
...
+        s"We will not overwrite the existing lnd.conf in default datadir")
+    } else {
+      Files.write(confFile, confStr.getBytes)
+    }
+
+    confFile
+  }
+}
+
+sealed abstract class LogLevel
```
Christewart: Maybe this can be moved somewhere else?

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/config/LndInstance.scala
```diff
@@ -0,0 +1,87 @@
...
+        val bytes = Files.readAllBytes(path)
+        val hex = ByteVector(bytes).toHex
+
+        macaroonOpt = Some(hex)
+        hex
+    }
+  }
+}
+
+object LndInstance {
```
Christewart: Same thing here, it would be nice to have a common interface. 

I guess we don't have to figure it all out on this PR, at minimum though we should file an issue

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
```diff
@@ -0,0 +1,517 @@
...
+      s"$file --lnddir=${instance.datadir.toAbsolutePath}"
+    case None => ""
+  }
+
+  implicit val executionContext: ExecutionContext = system.dispatcher
+
+  // These need to be lazy so we don't try and fetch
+  // the tls certificate before it is generated
+
+  lazy val certStream = new FileInputStream(instance.certFile)
```
Christewart: Can some of these things be `private`? 

We don't want to expose some of these to users of `LndRpcClient` IMO

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
```diff
@@ -0,0 +1,517 @@
...
+import java.io.{File, FileInputStream}
+import java.net.InetSocketAddress
+import java.util.concurrent.atomic.AtomicInteger
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success, Try}
+
+/** @param binary Path to lnd executable
+  */
+class LndRpcClient(val instance: LndInstance, binary: Option[File] = None)(
```
Christewart, line 35: Back to API discussion, is there some subset of APIs that are common across Eclair and Lnd? 

We could create some sort of `LightningRpcApi` that is extended by both this and `EclairRpcClient`. 

That way, when building lightning applications we can be more abstract rather than coding to a specific implementation. 

Here is `EclairApi` 

https://github.com/bitcoin-s/bitcoin-s/blob/5b762fb2e13742f057162e14a035b67730053ca0/eclair-rpc/src/main/scala/org/bitcoins/eclair/rpc/api/EclairApi.scala#L29

lnd-rpc/lnd-rpc.sbt
```diff
@@ -0,0 +1,59 @@
...
+name := "bitcoin-s-lnd-rpc"
+
+libraryDependencies ++= Deps.lndRpc
+
+CommonSettings.prodSettings
+
+enablePlugins(AkkaGrpcPlugin)
+
+// Disable fatal warning otherwise protobuf deprecation warnings will cause errors
+scalacOptions in Compile ~= (_ filterNot (s => s == "-Xfatal-warnings"))
```
Christewart: You should be able to filter this by using the same tools I used in this PR (specifically ignoring errors in the `protobuf` package)

https://github.com/bitcoin-s/bitcoin-s/pull/2822

testkit/src/main/scala/org/bitcoins/testkit/fixtures/LndFixture.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.bitcoins.lnd.rpc.LndRpcClient
+import org.bitcoins.rpc.client.common.BitcoindRpcClient
+import org.bitcoins.rpc.client.v21.BitcoindV21RpcClient
+import org.bitcoins.testkit.lnd.{LndRpcTestClient, LndRpcTestUtil}
+import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
+import org.bitcoins.testkit.util.BitcoinSAsyncFixtureTest
+import org.scalatest.FutureOutcome
+
+/** A trait that is useful if you need Lnd fixtures for your test suite */
+trait LndFixture extends BitcoinSFixture {
```
Christewart: We should definitely try and incorporate `CachedBitcoind` in here otherwise `withLnd()`/`withDualLnd` will spin up fresh bitcoinds/lnds for every test case. 

see https://github.com/bitcoin-s/bitcoin-s/pull/2792

Happy to help as this is newer code in our testkit project, and it is using some more advanced scala features

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/NativeProcessFactory.scala
```diff
@@ -33,8 +33,10 @@ trait NativeProcessFactory extends Logging {
         logger.debug(s"Binary was already started!")
         ()
       case None =>
-        val started = process.run()
-        processOpt = Some(started)
+        if (cmd.nonEmpty) {
```
Christewart, line 36: maybe put a warning log in the `else` clause?

Christewart (changes requested):

lnd-rpc/lnd-rpc.sbt
benthecarman: Doesn't seem you can disable deprecation warnings for a single package

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
benthecarman: This sounds good in theory but don't think will work out to well. These 2 return and take in a lot of different types for matching functions (ie eclair returns a `FundedChannelId` for openChannel and lnd returns a `Option[TransactionOutpoint]`) so I don't know how useful we could actually make it

benthecarman:
Note: fix one of `App, Chain, Node, and Core Tests` running forever

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
Christewart: I'll file an issue to think about this further. We can make their API conform to our API in some cases, but the example you cite is a good one, if `None` is given to us we can't translate that into a `FundedChannelId` :/

lnd-rpc/lnd-rpc.sbt
Christewart: Here is the incantation you need

```scala
scalacOptions in Compile += {
  //Added to supress all of the deprecation warnings inside of the
  //generated scala code by the grpc plugin. Two packages have deprecations
  //inside of them, 'signrpc' and 'lnrpc'
  //see: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html
  if (scalaVersion.value.startsWith("2.13")) {
    "-Wconf:cat=deprecation:site=lnrpc\\..*:silent,cat=deprecation:site=signrpc\\..*:silent"
  } else {
    //-Wconf doesn't work on Scala 2.12.x until we get 2.12.13
    //which is currently blocked on a scoverage bug
    ""
  }
}
```

lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
Christewart: https://github.com/bitcoin-s/bitcoin-s/issues/2844

testkit/src/main/scala/org/bitcoins/testkit/fixtures/LndFixture.scala
benthecarman: I think I did this correctly

testkit/src/main/scala/org/bitcoins/testkit/fixtures/LndFixture.scala
```diff
@@ -19,30 +18,28 @@ trait LndFixture extends BitcoinSFixture {
...
       }
     )(test)
   }
 }
 
 /** A trait that is useful if you need Lnd fixtures for your test suite */
-trait DualLndFixture extends BitcoinSFixture {
-  _: BitcoinSAsyncFixtureTest =>
+trait DualLndFixture extends BitcoinSFixture with CachedBitcoindV21 {
+  _: BitcoinSAsyncFixtureTest with CachedBitcoindFunded[_] =>
```
Christewart: So in plain english this is what this means
```scala
trait DualLndFixture extends BitcoinSFixture with CachedBitcoindV21 {
  _: BitcoinSAsyncFixtureTest with CachedBitcoindFunded[_] =>
```

>I've created a trait called `DualLndFixture` that extends `BitcoinSFixture with CachedBitcoindV21`. It _requires_ a `BitcoinSAsyncFixtureTest` AND `CachedBitcoindFunded[_]` to be injected into this trait before it can be used.

This doesn't make much sense, because you _already_ have a `CachedBitcoindV21` extended, so why do we need to inject _another_ `CachedBitcoindFunded[_]`. Its a bit redundant. 

you probably want to do something like this

```scala
trait DualLndFixture extends BitcoinSFixture with BitcoindFixturesFundedCachedV21
``` 

and then i don't think you need any `{_: BitcoinSAsyncFixtureTest with CachedBitcoindFunded[_] =>` 

Although I didn't run this locally to test it myself

testkit/src/main/scala/org/bitcoins/testkit/fixtures/LndFixture.scala
```diff
@@ -54,19 +51,18 @@ trait DualLndFixture extends BitcoinSFixture {
     makeDependentFixture[FixtureParam](
       () => {
         for {
-          bitcoind <- LndRpcTestUtil.startedBitcoindRpcClient()
+          bitcoind <- cachedBitcoindWithFundsF
           _ = logger.debug("starting bitcoind")
           _ <- bitcoind.start()
```
Christewart: Shouldn't need to be started if you are using a cached bitcoind

benthecarman:
Re-ran lnd rpc ci a bunch and haven't gotten any failures

app-commons/src/main/scala/org/bitcoins/commons/ConfigFactory.scala
```diff
@@ -0,0 +1,28 @@
+package org.bitcoins.commons
+
+import java.io.File
+import java.nio.file.Path
+
+trait ConfigFactory[T] {
```
Christewart: nit: I don't think there is any reason this can't be moved into `org.bitcoins.core.api.commons`

app-commons/src/main/scala/org/bitcoins/commons/InstanceFactory.scala
```diff
@@ -0,0 +1,13 @@
+package org.bitcoins.commons
+
+import java.io.File
+import java.nio.file.Path
+
+trait InstanceFactory[+T] {
```
Christewart: same comment wrt to moving to `core`

testkit/src/main/scala/org/bitcoins/testkit/lnd/LndRpcTestUtil.scala
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  def lndInstance(datadir: File): LndInstance = {
+    LndInstance.fromDataDir(datadir)
+  }
+
+  /** Returns a `Future` that is completed when both lnd and bitcoind have the same block height
+    * Fails the future if they are not synchronized within the given timeout.
+    */
+  def awaitLndInSync(lnd: LndRpcClient, bitcoind: BitcoindRpcClient)(implicit
```
Christewart, line 146: Not related to this PR, but I wanted to leave a note to re-visit this when we address #2844 

It would be very nice to have abstractions for this so we don't have to write new helper methods to do it

Christewart: Really the rest of the file

Christewart (approved):

PR #2838, opened 2021-03-31
Implement Oracle Explorer Client

### LINKED ISSUE #2827: Implement oracle explorer http client
API docs: 

https://gist.github.com/Christewart/a9e55d9ba582ac9a5ceffa96db9d7e1f

This should probably be a standalone module

Christewart:
fixes #2827 

This implements the oracle explorer http client for bitcoin-s. 

You can find the documentation here: 

https://gist.github.com/Christewart/a9e55d9ba582ac9a5ceffa96db9d7e1f


oracle-explorer-client/src/main/scala/org/bitcoins/explorer/client/SbExplorerClient.scala
```diff
@@ -0,0 +1,123 @@
...
+  import system.dispatcher
+  private val httpClient: HttpExt = Http(system)
+
+  /** Lists all events on oracle explorer
+    * @see https://gist.github.com/Christewart/a9e55d9ba582ac9a5ceffa96db9d7e1f#list-all-events
+    * @return
+    */
+  def listEvents(): Future[Vector[SbAnnouncementEvent]] = {
+    val base = env.baseUri
+    val uri = Uri(base + "events")
```
benthecarman, line 37: All of these require that the `base` ends with a `/`, should put some check that it does

oracle-explorer-client/src/main/scala/org/bitcoins/explorer/env/ExplorerEnv.scala
```diff
@@ -0,0 +1,23 @@
+package org.bitcoins.explorer.env
+
+sealed trait ExplorerEnv {
+  def baseUri: String
+}
+
+object ExplorerEnv {
```
benthecarman: Should make this extend `StringFactory`

oracle-explorer-client/src/main/scala/org/bitcoins/explorer/model/CreateAttestations.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import org.bitcoins.core.protocol.tlv.OracleAttestmentV0TLV
+import org.bitcoins.crypto.Sha256Digest
+
+case class CreateAttestations(
+    announcementHash: Sha256Digest,
+    attestment: OracleAttestmentV0TLV) {
+
+  override def toString: String = {
+    s"attestations=${attestment.hex}"
```
benthecarman, line 11: Does this need the `announcementHash`?

oracle-explorer-client/src/test/scala/org/bitcoins/explorer/model/SbOracleEventExplorerTest.scala
```diff
@@ -0,0 +1,32 @@
...
+    val oracleName = "Chris_Stewart_5"
+    val description = "2021-03-24-sunny-in-chicago"
+    val uriOpt = Some("https://twitter.com/Chris_Stewart_5")
+
+    val event =
+      CreateAnnouncementExplorer(announcement, oracleName, description, uriOpt)
+
+    //badForm={"description":["This field is required"],
+    // "oracleAnnouncementV0":["This field is required"],
+    // "oracleName":["This field is required"]}
```
benthecarman: I'm unsure what these refer to

build.sbt
```diff
@@ -738,6 +740,15 @@ lazy val dlcOracleTest = project
...
   .dependsOn(coreJVM % testAndCompile, dlcOracle, testkit)
 
+lazy val oracleExplorerClient = project
+  .in(file("oracle-explorer-client"))
+  .settings(CommonSettings.settings: _*)
+  .settings(
+    name := "bitcoin-s-oracle-explorer-client",
+    libraryDependencies ++= Deps.oracleExplorerClient
+  )
+  .dependsOn(coreJVM, appCommons, testkit % "test->test")
```
benthecarman, line 753: this should be split into `oracle-explorer-client` & `oracle-explorer-client-test` so any project that wants to use this doesn't need testkit

benthecarman (changes requested):

oracle-explorer-client/src/main/scala/org/bitcoins/explorer/model/CreateAttestations.scala
Christewart: No, it's part the URI that needs to be built though. The endpoint is `/v1/events/:hash/attestation`

build.sbt
Christewart: So this isn't quite right. 

You can inspect classpath's by doing `show fullClasspath` and `show test:fullClasspath` to see the difference between the `src` classpath and the `test` class path. The reason this doesn't show up is the `test->test` qualifier added to the end of `testkit`

You will see that the testkit deps show up on `test:fullClasspath`, but if you just do `fullClasspath` you won't see them. 

The reason we _normally_ don't want to included tests and source files in the same module is because we _want_ to use the sources in the testkit project to create fixtures. 

For example, we create bitcoind fixtures inside of testkit to _test_ our implementations of the exact same `BitcoindRpcClient`. 

https://github.com/bitcoin-s/bitcoin-s/blob/478ab16e28c65a4796f92433a6fef12cd35e351a/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala#L43

To do this, we need the testkit project to act as a middle man between the two. 

We could have just written all the test harness code inside of `bitcoind-rpc/src/test` but then it isn't easily re-usable (and publishable as a standalone dependency -- `testkit`) for other modules.
 
If in the future we want use the explorer http client inside our test fixtures to _test_ the `SbExplorerClient` implementation, we would need to do this.

benthecarman (approved):

Issue #2594, opened 2021-01-29
Contract Templates to make DLC setup easier for first-time testing users


nkohen:
The main complication in setting up a DLC using the bitcoin-s gui on the adapator-dlc branch (aside from downloading/pulling down code and installing/building) is the generation of the `contract_info`.

A `contract_info` has three parts: `total_collateral, contract_descriptor, oracle_info`.

For a first time user, it is probably best to use a single oracle's announcement to generate the oracle info, where this oracle should come from (dummy vs. real vs. ask sb to do one for you) is open for discussion.

The `contract_descriptor` is easier to help users with, in the enum outcome case they can use the enum contract generator tool as it is user-friendly enough, but the numeric tool is not really ready for users. Instead it would be ideal for users to be able to enter parameters into a template (e.g. total collateral, lower and upper strike price for a collared forward contract) and have either the GUI or website generate a contract descriptor from this information.

After all of this, it would be nice if the gui took these pieces of information separately (instead of one big `contract_info` blob) in the offer dialog.

PR #2842, opened 2021-03-31
Hyperbola DLCs

nkohen:
Built on top of #2824 

Implements https://github.com/discreetlogcontracts/dlcspecs/pull/144

PR #2847, opened 2021-04-01
Initial DLC Templates

benthecarman:
Part of #2594 

Implemented CFD templates and Call and Put Options templates

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
@@ -8,6 +8,19 @@ sealed trait ContractDescriptorTemplate {
   def totalCollateral: CurrencyUnit
 
   def toContractDescriptor: ContractDescriptor
+
+  require(
+    individualCollateral >= Satoshis.zero,
+    s"individualCollateral must be greater than or equal to zero, got $individualCollateral")
+
+  require(
+    totalCollateral > Satoshis.zero,
+    s"totalCollateral must be greater than or equal to zero, got $totalCollateral")
+
+  require(
+    individualCollateral <= totalCollateral,
+    s"Cannot have individual collateral ($individualCollateral) be greater than totalCollateral ($totalCollateral)"
+  )
 }
 
 /** A template for creating a Contract For Difference DLC
@@ -20,9 +33,13 @@ sealed trait CFDTemplate extends ContractDescriptorTemplate {
 
   def roundingIntervals: RoundingIntervals
 
-  def isLong: Boolean
-
   def numDigits: Int
+
+  require(numDigits > 0,
+          s"Num digits must be greater than zero, got $numDigits")
+  require(
+    strikePrice >= 0,
+    s"Strike price must be greater than or equal to zero, got $strikePrice")
 
   override val toContractDescriptor: NumericContractDescriptor = {
     val func: Long => Long = { outcome =>
@@ -36,9 +53,10 @@ sealed trait CFDTemplate extends ContractDescriptorTemplate {
     val descriptor =
       NumericContractDescriptor(curve, numDigits = numDigits, roundingIntervals)
 
-    if (isLong) {
-      descriptor.flip(totalCollateral.satoshis)
-    } else descriptor
+    this match {
+      case _: ShortCFD => descriptor
+      case _: LongCFD  => descriptor.flip(totalCollateral.satoshis)
+    }
   }
 }
 
@@ -49,9 +67,7 @@ case class LongCFD(
     numDigits: Int,
     strikePrice: Long,
     roundingIntervals: RoundingIntervals
-) extends CFDTemplate {
-  override val isLong: Boolean = true
-}
+) extends CFDTemplate
 
 /** @inheritdoc */
 case class ShortCFD(
@@ -60,9 +76,7 @@ case class ShortCFD(
     numDigits: Int,
     strikePrice: Long,
     roundingIntervals: RoundingIntervals
-) extends CFDTemplate {
-  override val isLong: Boolean = false
-}
+) extends CFDTemplate
 
 /** A template for doing an options contract DLC
   *
@@ -73,38 +87,47 @@ sealed trait OptionTemplate extends ContractDescriptorTemplate {
 
   def strikePrice: Long
 
-  def isCallOption: Boolean
-
   def numDigits: Int
 
   def roundingIntervals: RoundingIntervals
 
+  require(premium >= Satoshis.zero,
+          s"Premium must be greater than or equal to zero, got $premium")
+
+  require(numDigits > 0,
+          s"Num digits must be greater than zero, got $numDigits")
+  require(
+    strikePrice >= 0,
+    s"Strike price must be greater than or equal to zero, got $strikePrice")
+
   override val toContractDescriptor: NumericContractDescriptor = {
     val maxNum: Long = (BigInt(2).pow(numDigits) - 1).toLong
 
-    val curve = if (isCallOption) {
-      val pointA = OutcomePayoutEndpoint(
-        0L,
-        individualCollateral.satoshis - premium.satoshis)
+    val curve = this match {
+      case _: CallOption =>
+        val pointA = OutcomePayoutEndpoint(
+          0L,
+          individualCollateral.satoshis - premium.satoshis)
 
-      val pointB = OutcomePayoutEndpoint(
-        strikePrice,
-        individualCollateral.satoshis - premium.satoshis)
+        val pointB = OutcomePayoutEndpoint(
+          strikePrice,
+          individualCollateral.satoshis - premium.satoshis)
 
-      val pointC =
-        OutcomePayoutEndpoint(maxNum, totalCollateral)
-      DLCPayoutCurve(Vector(pointA, pointB, pointC))
-    } else {
-      val pointA = OutcomePayoutEndpoint(0L, totalCollateral)
+        val pointC =
+          OutcomePayoutEndpoint(maxNum, totalCollateral)
+        DLCPayoutCurve(Vector(pointA, pointB, pointC))
+      case _: PutOption =>
+        val pointA = OutcomePayoutEndpoint(0L, totalCollateral)
 
-      val pointB = OutcomePayoutEndpoint(
-        strikePrice,
-        individualCollateral.satoshis - premium.satoshis)
+        val pointB = OutcomePayoutEndpoint(
+          strikePrice,
+          individualCollateral.satoshis - premium.satoshis)
 
-      val pointC =
-        OutcomePayoutEndpoint(maxNum,
-                              individualCollateral.satoshis - premium.satoshis)
-      DLCPayoutCurve(Vector(pointA, pointB, pointC))
+        val pointC =
+          OutcomePayoutEndpoint(
+            maxNum,
+            individualCollateral.satoshis - premium.satoshis)
+        DLCPayoutCurve(Vector(pointA, pointB, pointC))
     }
 
     NumericContractDescriptor(curve, numDigits = numDigits, roundingIntervals)
@@ -119,10 +142,7 @@ case class CallOption(
     strikePrice: Long,
     premium: CurrencyUnit,
     roundingIntervals: RoundingIntervals
-) extends OptionTemplate {
-
-  override val isCallOption: Boolean = true
-}
+) extends OptionTemplate
 
 /** @inheritdoc */
 case class PutOption(
@@ -132,7 +152,4 @@ case class PutOption(
     strikePrice: Long,
     premium: CurrencyUnit,
     roundingIntervals: RoundingIntervals
-) extends OptionTemplate {
-
-  override val isCallOption: Boolean = false
-}
+) extends OptionTemplate
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
@@ -1,0 +1,138 @@
+package org.bitcoins.core.protocol.dlc
+
+import org.bitcoins.core.currency._
+
+sealed trait ContractDescriptorTemplate {
+  def individualCollateral: CurrencyUnit
+
+  def totalCollateral: CurrencyUnit
+
+  def toContractDescriptor: ContractDescriptor
+}
+
+/** A template for creating a Contract For Difference DLC
+  *
+  * @see https://www.investopedia.com/terms/c/contractfordifferences.asp
+  */
+sealed trait CFDTemplate extends ContractDescriptorTemplate {
+
+  def strikePrice: Long
+
+  def roundingIntervals: RoundingIntervals
+
+  def isLong: Boolean
+
+  def numDigits: Int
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+
+import org.bitcoins.core.currency._
+import org.bitcoins.core.protocol.dlc.ContractOraclePair._
+import org.bitcoins.core.protocol.dlc.DLCTemplate.validateMatchingOracleDescriptors
+import org.bitcoins.core.protocol.tlv._
+
+sealed trait DLCTemplate {
+  def totalCollateral: CurrencyUnit
+
+  def oracles: Vector[OracleAnnouncementTLV]
```
nkohen, line 10: this can be defined in terms of `oracleInfo`

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+
+  def oracleThreshold: Int
+
+  def contractDescriptor: ContractDescriptor
+
+  def oracleInfo: OracleInfo
+
+  def toContractInfo: ContractInfo
```
nkohen, line 18: I think everything here except for `individualCollateral` can be defined from this right? Either that or the other way around (meaning this can be defined from stuff here)

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+
+  def individualCollateral: CurrencyUnit =
+    totalCollateral.satoshis / Satoshis(2)
```
nkohen: hm, what about skew?

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+sealed trait MultiOracleDLCTemplate extends DLCTemplate {
+  override def oracleInfo: MultiOracleInfo[SingleOracleInfo]
+}
+
+sealed trait CFDTemplate extends DLCTemplate {
+  override def oracleInfo: NumericOracleInfo
+
+  def currentPrice: Long
+
+  def roundingIntervalsOpt: Option[RoundingIntervals]
```
nkohen: No need for the option as `RoundingIntervals.noRounding` can always be used

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+}
+
+sealed trait MultiOracleDLCTemplate extends DLCTemplate {
+  override def oracleInfo: MultiOracleInfo[SingleOracleInfo]
+}
+
+sealed trait CFDTemplate extends DLCTemplate {
+  override def oracleInfo: NumericOracleInfo
+
+  def currentPrice: Long
```
nkohen: for uniformity this might be better named the strike price ... I think? Right @Christewart ?

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+                "Oracles must use same numDigits and base")
+
+        val func: Long => Long = { outcome =>
+          if (outcome == 0) Long.MaxValue
+          else (currentPrice * individualCollateral.satoshis.toLong) / outcome
+        }
+
+        val numDigits = decomp.numDigits.toInt
+
+        val curve = CETCalculator.lineApprox(func, numDigits, 100)
```
nkohen: could you leave a TODO to have this be changed when #2842  gets merged?

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+  override val contractDescriptor: NumericContractDescriptor = {
+    oracles.head.eventTLV.eventDescriptor match {
+      case _: EnumEventDescriptorV0TLV =>
+        throw new IllegalArgumentException(
+          "Cannot do a CFD with a EnumEventDescriptorV0TLV")
+      case decomp: DigitDecompositionEventDescriptorV0TLV =>
+        require(validateMatchingOracleDescriptors(oracles),
+                "Oracles must use same numDigits and base")
+
+        val func: Long => Long = { outcome =>
```
nkohen: this assumes the template holder is taking the fixed other position in the CFD right? Probably needs a bool so the other side can be used too right?

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+sealed trait OptionTemplate extends DLCTemplate {
+  override def oracleInfo: NumericOracleInfo
+
+  def premium: CurrencyUnit
+
+  def strikePrice: Long
+
+  def isCall: Boolean
+
+  def roundingIntervalsOpt: Option[RoundingIntervals]
```
nkohen: ditto

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,296 @@
...
+
+  override val oracleInfo: NumericMultiOracleInfo =
+    NumericMultiOracleInfo(threshold = oracleThreshold,
+                           announcements = oracles,
+                           maxErrorExp = maxErrorExp,
+                           minFailExp = minFailExp,
+                           maximizeCoverage = maximizeCoverage)
+}
+
+sealed trait OptionTemplate extends DLCTemplate {
```
nkohen: scaladoc would be nice (Option is a pretty general term)

nkohen (commented):
I think it might be nice to have links out to investopedia or some other resource on the actual case classes for the contracts :)

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
benthecarman: I think it's better to do it this way so users can enter in each announcement themselves

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
benthecarman: Unsure what you mean here. We want to produce a `ContractInfo` from the template, so we should be entering params to construct one.

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,364 @@
...
+
+import org.bitcoins.core.currency._
+import org.bitcoins.core.protocol.dlc.ContractOraclePair._
+import org.bitcoins.core.protocol.dlc.DLCTemplate._
+import org.bitcoins.core.protocol.tlv._
+
+sealed trait DLCTemplate {
+  def totalCollateral: CurrencyUnit
+
+  def oracles: Vector[OracleAnnouncementTLV]
```
Christewart, line 10: So my thoughts on this are a bit different. 

It seems that a template shouldn't have any knowledge of oracles. It seems that a logical process for building a DLC are 

1. Specifying the contract info
2. Pick the oracles for the contract

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,364 @@
...
+
+  def oracleThreshold: Int
+
+  def contractDescriptor: ContractDescriptor
```
Christewart: I don't understand how we can have a `ContractDescriptor` if we are building a DLC? 

In my mind, the `DLCTemplate` should maybe be renamed to `ContractDescriptorBuilder` as it is taking input that will eventually be used to build a contract descriptor. 

We then use that contract descriptor inside of something like `DLCTemplate` to add in things like the oracles.

So I guess what I am saying is this stuff seems to make sense, but there is a level below this I think we need to add (building the `ContractDescriptor`)

Christewart (changes requested):

benthecarman:
So I am working on splitting up `DLCTemplate` into `ContractInfoTemplate` and `ContractDescriptorTemplate`. However, this might not be doable, for the contract descriptor we need to know the `numDigits` of the oracles, otherwise we aren't able to build a proper contract descriptor

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
```diff
@@ -0,0 +1,138 @@
...
+  def isCallOption: Boolean
+
+  def numDigits: Int
+
+  def roundingIntervals: RoundingIntervals
+
+  override val toContractDescriptor: NumericContractDescriptor = {
+    val maxNum: Long = (BigInt(2).pow(numDigits) - 1).toLong
+
+    val curve = if (isCallOption) {
```
Christewart: I think we can just get rid of this and pattern match ? I think this should work 

```scala
    val curve = this match {
      case _: CallOption =>
        val pointA = OutcomePayoutEndpoint(
          0L,
          individualCollateral.satoshis - premium.satoshis)

        val pointB = OutcomePayoutEndpoint(
          strikePrice,
          individualCollateral.satoshis - premium.satoshis)

        val pointC =
          OutcomePayoutEndpoint(maxNum, totalCollateral)
        DLCPayoutCurve(Vector(pointA, pointB, pointC))
      case _: PutOption =>
        val pointA = OutcomePayoutEndpoint(0L, totalCollateral)

        val pointB = OutcomePayoutEndpoint(
          strikePrice,
          individualCollateral.satoshis - premium.satoshis)

        val pointC =
          OutcomePayoutEndpoint(maxNum,
            individualCollateral.satoshis - premium.satoshis)
        DLCPayoutCurve(Vector(pointA, pointB, pointC))
    }
```

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
```diff
@@ -0,0 +1,138 @@
...
+  def totalCollateral: CurrencyUnit
+
+  def toContractDescriptor: ContractDescriptor
+}
+
+/** A template for creating a Contract For Difference DLC
+  *
+  * @see https://www.investopedia.com/terms/c/contractfordifferences.asp
+  */
+sealed trait CFDTemplate extends ContractDescriptorTemplate {
```
Christewart, line 30: invariants should be things like 

```scala
require(numDigits > 0, s"Num digits must be greater than zero, got=$numDigits")
require(strikePrice > 0,s"Strike price must be greater than zero, got=$stirkePrice")
```

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
```diff
@@ -0,0 +1,138 @@
+package org.bitcoins.core.protocol.dlc
+
+import org.bitcoins.core.currency._
+
+sealed trait ContractDescriptorTemplate {
```
Christewart, line 5:
```scala
require(individualCollateral < totalCollateral, s"Cannot have individual collateral be greater than totalCollateral, individualCollateral=$individualCollateral, totalCollateral=$totalCollateral")
```

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
```diff
@@ -0,0 +1,138 @@
...
+  *
+  * @see https://www.investopedia.com/terms/c/contractfordifferences.asp
+  */
+sealed trait CFDTemplate extends ContractDescriptorTemplate {
+
+  def strikePrice: Long
+
+  def roundingIntervals: RoundingIntervals
+
+  def isLong: Boolean
```
Christewart: scaladoc, `Long` has a lot of meanings in a programming context.

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
```diff
@@ -0,0 +1,138 @@
...
+) extends CFDTemplate {
+  override val isLong: Boolean = false
+}
+
+/** A template for doing an options contract DLC
+  *
+  * @see https://www.investopedia.com/terms/o/optionscontract.asp
+  */
+sealed trait OptionTemplate extends ContractDescriptorTemplate {
+  def premium: CurrencyUnit
```
Christewart, line 86: require(numDigits > 0, s"numDigits must be > 0, got=$numDigits")

Christewart: same with strike price

Christewart: I guess I added that comment on the premium field, which must be > 0 too :joy:. 

Just add appropriate invariants

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,89 @@
...
+    val pair: NumericPair =
+      ContractOraclePair.NumericPair(contractDescriptor, oracleInfo)
+    ContractInfo(totalCollateral.satoshis, pair)
+  }
+}
+
+case class MultiOracleDLCTemplate(
+    oracles: Vector[OracleAnnouncementTLV],
+    oracleThreshold: Int,
+    maxErrorExp: Int,
```
Christewart, line 50: i'm not as familiar with these fields, but i assume there can be invariants placed on these as well (`minFailExp`, `maxErrorExp`)

core/src/main/scala/org/bitcoins/core/protocol/dlc/DLCTemplate.scala
```diff
@@ -0,0 +1,89 @@
+package org.bitcoins.core.protocol.dlc
+
+import org.bitcoins.core.currency.CurrencyUnit
+import org.bitcoins.core.protocol.dlc.ContractOraclePair._
+import org.bitcoins.core.protocol.tlv._
+
+sealed trait DLCTemplate {
```
Christewart, line 7: invariants on 

1. totalCollateral > 0
2. oracleThreshold < oracles.length
3. oracleThreshold > 0

Christewart (changes requested):

core/src/main/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTemplate.scala
benthecarman: clever

Christewart (approved):

PR #2851, opened 2021-04-02
Add ability to delete Oracle signatures

### LINKED ISSUE #2828: Add abillity to delete attestations from the oracle
Something discussed on slack between me and @benthecarman 

### Problem

We've had a few users (myself included) that have accidentally attested to 

1. the wrong event
2. input the wrong attestation

By design with DLCs, if you attest to two different outcomes for the same event, you leak your private key. This means this we have to be _very_ careful when deleting these. Namely, we have to make sure that they aren't publicly available.

### Solution

Add the ability to delete these attestations. In the case where a user immediately notices there error, this is completely safe to do.

If you aren't sure if you have published this yet, a proposed solution with #2827 is to scan the oracle explorer to see if you have publish the attestations on the website. 

If you have _not_ published to the explorer **AND NOT TOLD ANYONE ELSE MANUALLY**, it is safe to delete the attestations and re-attest. 

If you have published to the oracle explorer **OR TOLD ANYONE ELSE MANUALLY**, it _isn't_ safe to delete and re-attest. We should not allow the user to delete the attestations.

Prominent warnings should be displayed when deleting attestations, in either case as it is very risky to do.

benthecarman:
fixes #2828

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
@@ -62,4 +62,8 @@ trait DLCOracleApi {
   def signDigits(eventName: String, num: Long): Future[OracleEvent]
 
   def signDigits(oracleEventTLV: OracleEventTLV, num: Long): Future[OracleEvent]
+
+  def deleteSigs(eventName: String): Future[OracleEvent]
+
+  def deleteSigs(oracleEventTLV: OracleEventTLV): Future[OracleEvent]
 }
--- a/core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
@@ -63,7 +63,17 @@ trait DLCOracleApi {
 
   def signDigits(oracleEventTLV: OracleEventTLV, num: Long): Future[OracleEvent]
 
-  def deleteSigs(eventName: String): Future[OracleEvent]
+  /** Deletes attestations for the given event
+    *
+    * WARNING: if previous signatures have been made public
+    * the oracle private key will be revealed.
+    */
+  def deleteAttestations(eventName: String): Future[OracleEvent]
 
-  def deleteSigs(oracleEventTLV: OracleEventTLV): Future[OracleEvent]
+  /** Deletes attestations for the given event
+    *
+    * WARNING: if previous signatures have been made public
+    * the oracle private key will be revealed.
+    */
+  def deleteAttestations(oracleEventTLV: OracleEventTLV): Future[OracleEvent]
 }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
```diff
@@ -62,4 +62,8 @@ trait DLCOracleApi {
   def signDigits(eventName: String, num: Long): Future[OracleEvent]
 
   def signDigits(oracleEventTLV: OracleEventTLV, num: Long): Future[OracleEvent]
+
+  def deleteSigs(eventName: String): Future[OracleEvent]
```
Christewart: Let's add scaladocs at the interface level saying these are potentially dangerous operations that can cause private key leakage.

Christewart: nit: `deleteAttestations()`

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -727,4 +727,77 @@ class DLCOracleTest extends DLCOracleFixture {
...
+            case _: PendingOracleEvent | _: CompletedOracleEvent =>
+              fail()
+          }
+        }
+
+        _ <- dlcOracle.deleteSigs("test")
+        event <- dlcOracle.findEvent("test").map(_.get)
+      } yield {
+        event match {
+          case _: PendingEnumV0OracleEvent => succeed
```
Christewart, line 762: nice use of ADTs in the test case, exactly what we want :+1:

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -372,6 +372,25 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
       digitSigs <- Future.sequence(digitSigFs)
     } yield OracleEvent.fromEventDbs(signSig ++ digitSigs)
   }
+
+  override def deleteSigs(eventName: String): Future[OracleEvent] = {
```
Christewart: I would add those same scaladocs here just to be SUPER clear that this is a possibly dangerous method

dlc-oracle-test/src/test/scala/org/bitcoins/dlc/oracle/DLCOracleTest.scala
```diff
@@ -727,4 +727,77 @@ class DLCOracleTest extends DLCOracleFixture {
...
+        _ <- dlcOracle.deleteSigs("test")
+        event <- dlcOracle.findEvent("test").map(_.get)
+      } yield {
+        event match {
+          case _: PendingDigitDecompositionV0OracleEvent => succeed
+          case _: PendingOracleEvent | _: CompletedOracleEvent =>
+            fail()
+        }
+      }
+  }
```
Christewart, line 802: since this is such a potentially dangerous method, let's add negative test cases to make sure we get failures when the event does not exist

Christewart (changes requested):

Christewart (approved):

PR #2866, opened 2021-04-06
Rework oracle explorer client to use new api paths

Christewart:
Changes the API paths for the oracle explorer client from using `events/` -> `announcements/`


benthecarman (approved):
ACK, should wait until we deploy it to merge however, otherwise will have CI errors

Issue #2830, opened 2021-03-28
Cloning bitcoin-s repo takes forever


benthecarman:
Doing a `git clone` of the repo takes a very long time. Is there something we can do to reduce the size of the repo besides just squashing commits?

Christewart:
Related to #2674 

So i did some googling on this this morning, and found this stackoverflow post

https://stackoverflow.com/questions/10622179/how-to-find-identify-large-commits-in-git-history

I ran the script the provided and added the filter to check elements that are not in `HEAD` (`grep -vF --file=<(git ls-tree -r HEAD | awk '{print $3}') |`)


```
git rev-list --objects --all |
  git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' |
  sed -n 's/^blob //p' |
  grep -vF --file=<(git ls-tree -r HEAD | awk '{print $3}') |
  sort --numeric-sort --key=2 |
  cut -c 1-12,41- |
  $(command -v gnumfmt || echo numfmt) --field=2 --to=iec-i --suffix=B --padding=7 --round=nearest
```

and I get a huge list of seemingly all of our html files 

>0da968cbbb0a  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
117b43ca63e2  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
251e55c15503  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
273cd58d2d00  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
27b34daf1924  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
2fc75e867421  1.1MiB api/org/bitcoins/testkit/chain/ChainDbUnitTest.html
...

So i'm wondering if we are redundantly committing all these files on every merge? Idk

benthecarman:
These are generated from unidoc, I believe we should ignore those

Christewart:
I'm suspicious of this code, as it seems to be moving the unidoc stuff to another place, which means it isn't in the `.gitignore` any more

https://github.com/bitcoin-s/bitcoin-s/blob/478ab16e28c65a4796f92433a6fef12cd35e351a/build.sbt#L281

Christewart:
[...]

PR #2880, opened 2021-04-08
Add docs for using CachedBitcoind

### LINKED ISSUE #2876: Add testkit documentation for using CachedBitcoind & using the pre-defiend fixtures we have

Christewart:
fixes #2876

### DIFF
--- a/docs/testkit/testkit.md
+++ b/docs/testkit/testkit.md
@@ -31,8 +31,8 @@ This gives the ability to create and destroy `bitcoind` on the underlying operat
 
 Make sure you have run `sbt downloadBitcoind` before running this example, as you need access to the bitcoind binaries.
 
-Our [BitcoindRpcClient](../../bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala) is tested with the functionality provided in the testkit.
-A quick example of a useful utility method is [BitcoindRpcTestUtil.startedBitcoindRpcClient()](../../testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala).
+Our [BitcoindRpcClient](/api/org/bitcoins/rpc/client/common/BitcoindRpcClient) is tested with the functionality provided in the testkit.
+A quick example of a useful utility method is [BitcoindRpcTestUtil.startedBitcoindRpcClient()](/api/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil).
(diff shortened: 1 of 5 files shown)


docs/testkit/testkit.md
```diff
@@ -31,8 +31,8 @@ This gives the ability to create and destroy `bitcoind` on the underlying operat
 
 Make sure you have run `sbt downloadBitcoind` before running this example, as you need access to the bitcoind binaries.
 
-Our [BitcoindRpcClient](/api/org/bitcoins/rpc/client/common/BitcoindRpcClient) is tested with the functionality provided in the testkit.
-A quick example of a useful utility method is [BitcoindRpcTestUtil.startedBitcoindRpcClient()](/api/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil).
+Our [BitcoindRpcClient](../../bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala) is tested with the functionality provided in the testkit.
+A quick example of a useful utility method is [BitcoindRpcTestUtil.startedBitcoindRpcClient()](../../testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala).
```
benthecarman: these links should say with the `/api/` so they link out to the actual api docs, otherwise they will be dead links on bitcoin-s.org

Currently they link to this https://bitcoin-s.org/api/org/bitcoins/rpc/client/common/BitcoindRpcClient

benthecarman (changes requested):

docs/testkit/testkit.md
Christewart: Ok, so this is going to be problematic then with my proposed git solution: 

https://github.com/bitcoin-s/bitcoin-s/issues/2830#issuecomment-812497571

This solution gets rid of everything under `api/` in our git history :/

Christewart: So if I understand this correctly we can just change the path to be 

>website/static/api/path/to/file.scala

and things should work :tm: ? 

https://scalameta.org/mdoc/docs/docusaurus.html#include-scaladoc-in-site

Christewart: Ok, i don't think this is a problem, I will revert the paths. 

For some context of why i don't think it is a problem: 

https://github.com/bitcoin-s/bitcoin-s/issues/2830#issuecomment-816621602

benthecarman (approved):

Issue #2608, opened 2021-02-02
Dependabot security alerts for docusaraus deps immer & highlight


Christewart:
So AFAICT we are on the latest version of docusaraus, but that version in vulnerable. I think we are going to have to wait for updates upstream

The latest version is 1.14.6 

https://docusaurus.io/docs/en/installation 

_Originally posted by @Christewart in https://github.com/bitcoin-s/bitcoin-s/issues/2607#issuecomment-771764374_

Christewart:
`npm update` didn't solve this on 1.9.0. Unfortunately it seems like the dependencies are deep rooted and maybe need changes to docusauraus itself?

Christewart:
Still a problem as of docusaraus 1.14.7

PR #2884, opened 2021-04-08
Bump website dependencies by running yarn update

Christewart:
Related to #2608

Just simply ran `yarn update` in the `website/` directory

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -29,8 +29,8 @@
       "chain/filter-sync": {
         "title": "Syncing Blockfilters"
       },
-      "config/configuration": {
-        "title": "Application Configuration"
+      "config/configuration title: Application Configuration": {
+        "title": "config/configuration title: Application Configuration"
       },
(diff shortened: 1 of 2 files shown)


website/i18n/en.json
```diff
@@ -29,8 +29,8 @@
       "chain/filter-sync": {
         "title": "Syncing Blockfilters"
       },
-      "config/configuration": {
-        "title": "Application Configuration"
+      "config/configuration title: Application Configuration": {
```
Christewart, line 32: I'm not sure why these are getting changed :/

Christewart (commented):
This should fix the following dependency warnings on github

1. is-svg (`3.0.0` -> `4.3.1`)
2. react-dev-utils (`9.1.0` -> `11.0.1`)
3. immer (`1.10.0` -> `8.0.1`)

So after merging this, i believe there will still be one warning which is `highlight.js`

benthecarman (approved):

PR #2885, opened 2021-04-08
Add Schnorr and Adaptor Secp Bindings and Update Adaptor

nkohen:
Removes secp256k1 submodule, adds secp256k1-zkp submodule targeting https://github.com/bitcoin-s/secp256k1-zkp/pull/1

Replaces linux_64 binaries, updates `LibSecp256k1CryptoRuntime`, and adds tests

TODO:

- [x] Update `AdaptorUtil` and `DLEQUtil` to be compliant with https://github.com/discreetlogcontracts/dlcspecs/pull/114
- [x] Add ECDSA Adaptor Signature test vector tests
- [x] Target `bitcoin-s-master` in submodule when java bindings are merged
- [x] Add mac + windows binaries
- [x] Update add-to-jni.md


crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleBIP340Test.scala
```diff
@@ -32,9 +32,10 @@ class BouncyCastleBIP340Test extends BitcoinSCryptoTest {
       sig: SchnorrDigitalSignature,
       expectedResult: Boolean,
       comment: String): Assertion = {
-    val secpResult = pubKey.verify(msg, sig)
+    val secpResult = Try(pubKey.verify(msg, sig)).getOrElse(false)
```
Christewart, line 35: Why is this needed now?

crypto/.jvm/src/main/scala/org/bitcoins/crypto/LibSecp256k1CryptoRuntime.scala
```diff
@@ -276,8 +263,16 @@ trait LibSecp256k1CryptoRuntime extends CryptoRuntime {
   override def sipHash(item: ByteVector, key: SipHashKey): Long =
     BouncycastleCryptoRuntime.sipHash(item, key)
 
-  override def decodePoint(bytes: ByteVector): ECPoint =
-    BouncycastleCryptoRuntime.decodePoint(bytes)
+  override def decodePoint(bytes: ByteVector): ECPoint = {
+    if (bytes == ByteVector.fromByte(0x00)) {
```
Christewart: nit: 

```scala
private val infinity: ByteVector = ByteVector.fromByte(0x00)

override def decodePoint(bytes: ByteVector): ECPoint = { 
if (bytes == infinity) { .. }
...
```

better readability IMO and also caches the byte vector

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -3,7 +3,25 @@ package org.bitcoins.crypto
...
+      algoName: String,
+      auxRand: ByteVector): FieldElement = {
+    val randHash = CryptoUtil.sha256ECDSAAdaptorAux(auxRand).bytes
+    val maskedKey = randHash.xor(privKey.bytes)
+
+    val bytesToHash = maskedKey ++ adaptorPoint.compressed.bytes ++ message
+    val nonceHash = algoName match {
+      case "DLEQ"             => CryptoUtil.sha256DLEQ(bytesToHash)
+      case "ECDSAadaptor/non" => CryptoUtil.sha256ECDSAAdaptorNonce(bytesToHash)
+      case _: String          => CryptoUtil.taggedSha256(bytesToHash, algoName)
```
Christewart, line 42: should we just throw in this case? 

It seems like a tricky "silent" failure (or feature) otherwise.

Seems like we only use it in places where we specify the `algoName` currently

I would prefer this sensitive code to fail loudly if we are doing something that isn't explicitly supported

Screenshot from 2021-04-12 12-33-48

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -3,7 +3,25 @@ package org.bitcoins.crypto
 import scodec.bits.ByteVector
 
 object AdaptorUtil {
-  import ECAdaptorSignature.{deserializePoint, serializePoint}
+
+  def adaptorNonce(
```
Christewart, line 29: Is this the appropriate documentation for this method? 

https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#secure-nonce-generation

Christewart: Either way, scaladoc please

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -3,7 +3,25 @@ package org.bitcoins.crypto
...
+      case "DLEQ"             => CryptoUtil.sha256DLEQ(bytesToHash)
+      case "ECDSAadaptor/non" => CryptoUtil.sha256ECDSAAdaptorNonce(bytesToHash)
+      case _: String          => CryptoUtil.taggedSha256(bytesToHash, algoName)
+    }
+
+    FieldElement(nonceHash.bytes)
+  }
 
   // Compute s' = k^-1 * (dataToSign + rx*privateKey)
   private def adaptorSignHelper(
```
Christewart, line 52: In addition to the existing comment, let's link to 

https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#encrypted-signing

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -28,13 +46,13 @@ object AdaptorUtil {
   def adaptorSign(
```
Christewart, line 72: Scaladoc: https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#encrypted-signing

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -63,40 +80,32 @@ object AdaptorUtil {
     val untweakedPoint =
       m.getPublicKey.add(pubKey.tweakMultiply(rx)).tweakMultiply(s.inverse)
 
-    FieldElement(untweakedPoint.bytes.tail)
+    FieldElement(untweakedPoint.compressed.bytes.tail)
   }
 
   def adaptorVerify(
```
Christewart, line 113: scaladoc: https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#encryption-verification

What are your thoughts on changing parameter names to match what they are in the spec? 

For instance, we have `pubKey` and `adaptor` parameter names. I assume `X=pubKey` and `Y=adaptor`?

crypto/src/main/scala/org/bitcoins/crypto/DLEQUtil.scala
```diff
@@ -81,20 +78,19 @@ object DLEQUtil {
     // if R = y*R' which is what we are trying to prove.
     val s = fe.multiply(e).add(k)
 
-    (s, e)
+    (e, s)
   }
 
   /** Verifies a proof that the DLOG_G of P1 equals the DLOG_adaptor of P2 */
   def dleqVerify(
```
Christewart, line 118: I believe this is the correct reference for a scaladoc? 

https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#verifying

Christewart: I'm having a hard time understanding the parameter mapping from our implementation and what the spec says here

We have paraemter `s`, `e`, `p1`, `adaptor`, `p2` and the spec has `X`,`Y`,`Z` (points) and then a `proof` 

What is the mapping between our parameters and the spec parameters? Can we rename our parameters to be consistent with the spec so this is easier to review?

crypto/src/main/scala/org/bitcoins/crypto/DLEQUtil.scala
```diff
@@ -15,26 +14,32 @@ object DLEQUtil {
   }
 
   def dleqNonceFunc(
```
Christewart: nit: Drop `Func` from the name since it doesn't return a function

Christewart: I believe this is the appropriate reference in a scaladoc? 

https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#secure-nonce-generation

crypto/src/main/scala/org/bitcoins/crypto/DLEQUtil.scala
```diff
@@ -15,26 +14,32 @@ object DLEQUtil {
...
+      .bytes
+
+    AdaptorUtil.adaptorNonce(hash,
+                             fe.toPrivateKey,
+                             adaptorPoint,
+                             "DLEQ",
+                             auxRand)
   }
 
   def dleqChallengeHash(
```
Christewart, line 61: I think this is the best reference for a scaladoc? 

https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#proof-of-discrete-logarithm-equality

crypto/src/main/scala/org/bitcoins/crypto/DLEQUtil.scala
```diff
@@ -44,25 +49,17 @@ object DLEQUtil {
   def dleqProve(
```
Christewart, line 78: scaladoc: https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#proving

Since our parameters aren't the same as the spec, it isn't clear to me where we are enforcing this invariant: 

>x: A non-zero scalar representing the witness for the proof.

It seems if a library is using this independent of the rest of our code base, this is possible?

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
Christewart: So if I understand correctly, we take in `dataToSign` (called `message` in the spec, i would be in favor of renaming our parameter name to this) so we can preserve this invariant right? 

>message_hash: A 32-byte array which must be the hash of the message (in our case a transaction digest).

If so, it might be worth noting the scaladoc why we take the message and not just the hash. 

As a side note, it might be worth adding to the spec itself _why_  this MUST be the hash of the message. I don't understand if it will just fail to verify, or we can do something more sinister like leak a private key.

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -107,12 +116,9 @@ object AdaptorUtil {
 
   def adaptorComplete(
```
Christewart, line 145: scaladoc: https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#decryption

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
```diff
@@ -123,7 +129,15 @@ object AdaptorUtil {
       sig: ECDigitalSignature,
```
Christewart, line 158: scaladoc: https://github.com/discreetlogcontracts/dlcspecs/blob/d01595b70269d4204b05510d19bba6a4f4fcff23/ECDSA-adaptor.md#key-recovery

Christewart (changes requested):
Mostly nits. 

1. Since we wrote code first (rather than spec first) I acknowledge most variable names are not the same for that reason, but it _really_ makes it easier to review when the two are the same. 
2. I have a hard time following where invariants are enforced between the spec and our code base. I have tried to single out specific ones that confuse me. We need to keep in mind that people might use this as a standalone library in the future, which means they might not be using secitons of the codebase that enforce invariants. Related to #2898

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleBIP340Test.scala
nkohen: First one isn't needed (invalid additions and such are handled internally by secp which gives a return code of 0 instead of 1), second one is now needed because pubkey combine is used for addition which can fail for invalid additions.

crypto/src/main/scala/org/bitcoins/crypto/AdaptorUtil.scala
nkohen: If other people want to use us as a library for their own algos then this field will be needed. This is also the reason why this field is exposed on the secp256k1-zkp implementation as well (where internally they also do a `match` like this one.

crypto/src/main/scala/org/bitcoins/crypto/DLEQUtil.scala
nkohen: Added the invariant (it was being enforced on all bitcoin-s calls but you're right that that could be a footgun for an outside user

crypto-test/src/test/scala/org/bitcoins/crypto/ECPublicKeyTest.scala
```diff
@@ -93,20 +94,30 @@ class ECPublicKeyTest extends BitcoinSCryptoTest {
...
+  }
 
-    assert(res2 == ECPublicKey.infinity)
+  it must "correctly compress keys" in {
+    forAll(CryptoGenerators.privateKey) { privKey =>
+      val pubKey = privKey.publicKey
+      val pubKeyCompressed = pubKey.compressed
+      val pubKeyDecompressed = pubKey.decompressed
+
+      assert(pubKey == pubKeyCompressed || pubKey == pubKeyDecompressed)
```
benthecarman: Is there a reason we don't always return a compressed or decompressed key when calling `privKey.publicKey`? This should be deterministic

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -515,6 +564,266 @@ public static boolean isValidPubKey(byte[] pubkey) {
...
+        byte[] sigArray = retByteArray[0];
+        int retVal = new BigInteger(new byte[]{retByteArray[1][0]}).intValue();
+
+        assertEquals(retVal, 1, "Failed return value check.");
+
+        return sigArray;
+    }
+
+    /*
+    public static byte[] schnorrComputeSigPoint(byte[] data, byte[] nonce, byte[] pubkey, boolean compressed) throws AssertFailException {
```
benthecarman, line 641: Can we just remove this?

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -578,8 +886,25 @@ private static void safeRewind(ByteBuffer byteBuff) {
...
+
     private static native byte[][] secp256k1_ec_pubkey_decompress(ByteBuffer byteBuff, long context, int inputLen);
 
     private static native byte[][] secp256k1_ecdh(ByteBuffer byteBuff, long context, int inputLen);
 
+    private static native byte[][] secp256k1_schnorrsig_sign(ByteBuffer byteBuff, long context);
+
+    private static native byte[][] secp256k1_schnorrsig_sign_with_nonce(ByteBuffer byteBuff, long context);
+
+    //private static native byte[][] secp256k1_schnorrsig_compute_sigpoint(ByteBuffer byteBuff, long context, boolean compressed);
```
benthecarman, line 899: ditto

secp256k1jni/src/test/java/org/bitcoin/NativeSecp256k1Test.java
```diff
@@ -258,6 +276,108 @@ public void testCreateECDHSecret() throws AssertFailException{
...
+
+        byte[] sigArr = NativeSecp256k1.schnorrSignWithNonce(data, secKey, nonce);
+        String sigStr = toHex(sigArr);
+        String expectedSig = "5DA618C1936EC728E5CCFF29207F1680DCF4146370BDCFAB0039951B91E3637A958E91D68537D1F6F19687CEC1FD5DB1D83DA56EF3ADE1F3C611BABD7D08AF42";
+        assertEquals(sigStr, expectedSig, "testSchnorrSignWithNonce");
+    }
+
+    /*
+    @Test
+    public void testSchnorrComputeSigPoint() throws AssertFailException{
```
benthecarman, line 305: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -125,28 +125,28 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
-          .schnorrSignWithNonce(bytes, nonceKey, context = LibSecp256k1)
-        assert(sigBC.bytes == sigSecP.bytes)
+           NumberGenerator.bytevector(32)) { case (privKey, nonceKey, bytes) =>
+      assert(
+        LibSecp256k1CryptoRuntime.schnorrSignWithNonce(
+          bytes,
+          privKey,
+          nonceKey) == BouncycastleCryptoRuntime.schnorrSignWithNonce(bytes,
+                                                                      privKey,
+                                                                      nonceKey))
```
benthecarman: nit: give these 2 `val`s

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
         val sig = privKey.schnorrSign(bytes)
         val pubKey = privKey.schnorrPublicKey
+
         assert(
-          pubKey.verify(bytes, sig, context = BouncyCastle) == pubKey
-            .verify(bytes, sig, context = LibSecp256k1))
+          LibSecp256k1CryptoRuntime.schnorrVerify(
+            bytes,
+            pubKey,
+            sig) == BouncycastleCryptoRuntime.schnorrVerify(bytes, pubKey, sig))
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
         assert(
-          pubKey.verify(bytes, badSig, context = BouncyCastle) == pubKey
-            .verify(bytes, badSig, context = LibSecp256k1))
+          LibSecp256k1CryptoRuntime
+            .schnorrVerify(bytes, pubKey, badSig) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
   it must "compute schnorr signature points the same" in {
     forAll(CryptoGenerators.schnorrPublicKey,
            CryptoGenerators.schnorrNonce,
+           NumberGenerator.bytevector(32)) { case (pubKey, nonce, bytes) =>
+      assert(
+        LibSecp256k1CryptoRuntime.schnorrComputeSigPoint(
+          bytes,
+          nonce,
+          pubKey,
+          compressed = true) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
-                                 nonce,
-                                 compressed = true,
-                                 context = BouncyCastle)
+      case (privKey, adaptor, msg, auxRand) =>
+        assert(
+          LibSecp256k1CryptoRuntime.adaptorSign(
+            privKey,
+            adaptor,
+            msg,
+            auxRand) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
+    forAll(CryptoGenerators.privateKey,
+           CryptoGenerators.publicKey,
+           NumberGenerator.bytevector(32),
+           CryptoGenerators.adaptorSignature) {
+      case (privKey, adaptor, msg, badSig) =>
+        val sig = privKey.adaptorSign(adaptor, msg)
+        val pubKey = privKey.publicKey
+
+        assert(LibSecp256k1CryptoRuntime
+          .adaptorVerify(sig, pubKey, msg, adaptor) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
+          .adaptorVerify(sig, pubKey, msg, adaptor))
+        assert(
+          LibSecp256k1CryptoRuntime.adaptorVerify(
+            badSig,
+            pubKey,
+            msg,
+            adaptor) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
+  }
 
-        assert(bouncyCastleSigPoint == secpSigPoint)
+  it must "complete adaptor signatures the same" in {
+    forAll(CryptoGenerators.privateKey, CryptoGenerators.adaptorSignature) {
+      case (adaptorSecret, adaptorSig) =>
+        assert(
+          LibSecp256k1CryptoRuntime.adaptorComplete(
+            adaptorSecret,
+            adaptorSig) == BouncycastleCryptoRuntime
```
benthecarman: ditto

crypto-test/.jvm/src/test/scala/org/bitcoins/crypto/BouncyCastleSecp256k1Test.scala
```diff
@@ -157,33 +157,92 @@ class BouncyCastleSecp256k1Test extends BitcoinSCryptoTest {
...
+    }
+  }
+
+  it must "extract adaptor secrets the same" in {
+    forAll(CryptoGenerators.adaptorSignatureWithDecryptedSignatureAndAdaptor) {
+      case (adaptorSig, sig, adaptor) =>
+        assert(
+          LibSecp256k1CryptoRuntime.extractAdaptorSecret(sig,
+                                                         adaptorSig,
+                                                         adaptor) ==
```
benthecarman: ditto

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
nkohen: I haven't decided if I'm going to re-introduce it in C or if I'm going to do that in scala

crypto-test/src/test/scala/org/bitcoins/crypto/ECPublicKeyTest.scala
nkohen: it is determined by a param of `ECPrivateKey` currently

Christewart:
Still does not compile https://github.com/bitcoin-s/bitcoin-s/pull/2885/checks?check_run_id=2331021263#step:5:1735

Christewart (approved):

PR #2886, opened 2021-04-09
Implement BIP32Path.fromHardenedString().

### LINKED ISSUE #2750: Construct oracle private key with real types rather than a string
https://github.com/bitcoin-s/bitcoin-s/blob/1d5b565abe3a7f9137e8bd2246215c0d32902b8a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala#L54

It seems _very_ easy to make a typo by forgetting a `'`. This seems like a footgun waiting to happen. My understanding is this could have pretty dramatic security implications if this string is refactored / moved in the future and we lose a `'` along the way. 

I guess an alternative could be defining `BIP32.fromStringHardened()` that throws if the entire path isn't hardened.

Christewart:
fixes #2750 

This is a safety feature to make sure we are using only hardened paths with dlc oracle

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
+++ b/core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
@@ -147,40 +147,13 @@ object BIP32Path extends Factory[BIP32Path] with StringFactory[BIP32Path] {
   }
 
   /** Takes in a BIP32 Path and verifies all paths are hardened
-    * @throws IllegalArgumentException is a non hardened path is found
+    * @throws RuntimeException is a non hardened path is found
     */
   def fromHardenedString(string: String): BIP32Path = {
-    val parts: Vector[String] = {
-      val p = string
-        .split("/")
-        .toVector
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bitcoins/core/hd/BIP32Path.scala
```diff
@@ -146,6 +146,43 @@ object BIP32Path extends Factory[BIP32Path] with StringFactory[BIP32Path] {
     BIP32PathImpl(path)
   }
 
+  /** Takes in a BIP32 Path and verifies all paths are hardened
+    * @throws IllegalArgumentException is a non hardened path is found
+    */
+  def fromHardenedString(string: String): BIP32Path = {
```
benthecarman, line 152: instead of redoing the implementation of `BIP32Path.fromString` you could do something much simpler:

```scala
  def fromHardenedString(string: String): BIP32Path = {
    val path = BIP32Path.fromString(string)
    require(path.forall(_.hardened), s"Did not receive all hardened paths from string, got=$string")
    path
}

benthecarman (changes requested):

benthecarman (approved):

PR #2900, opened 2021-04-13
Cache fee rate & remove DLC locktime for GUI

### LINKED ISSUE #2374: Prepopulate fee rate field and possibly locktime field for DLC
Screenshot from 2020-12-16 15-04-51

It would be nice to have the fee field pre-populated with whatever your fee rate provider says for 6 blocks. Then you can adjust the fee rate to be higher or lower dependending on your timeline. This at least gives you a baseline of what fee rates currently are. 

Also as discussed in #2359 the `locktime` is required, but it might be nice to have a default value provided there too? I guess in general, we should provide sane defaults so we don't have as many options for users as it confuses people. 

They can of course adjust those default values in the text boxes, but I think it's good to provide sane defaults for those options.

benthecarman:
Fixes #2374

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -1345,6 +1345,9 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("estimatefee")
+        .action((_, conf) => conf.copy(command = EstimateFee))
+        .text("Returns the recommend fee rate using the fee provider"),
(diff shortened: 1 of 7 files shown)


nkohen (commented):
Looks good but is the intention really to target adaptor-dlc and not some new branch?

benthecarman:
> Looks good but is the intention really to target adaptor-dlc and not some new branch?

I think this is fine for `adaptor-dlc`, we don't have a new branch yet

Christewart (approved):
LGTM given the time crunch but filed #2903 to pull it into master when we have time

PR #2901, opened 2021-04-13
Remove confusing items from DLC Table view

### LINKED ISSUE #2359: Contract maturity not working in the GUI
I am running on a0b60b2ebf954b4db185f09af95419b353794209

It appears we don't have contract maturity showing up. It would be nice if it was displayed in a nice format that makes it easy to determine when the contract matures

Screenshot from 2020-12-15 07-18-31

benthecarman:
Closes #2359

image

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
@@ -9,13 +9,15 @@ import scalafx.scene.control.{ContextMenu, MenuItem, TableColumn, TableView}
 class DLCTableView(model: DLCPaneModel) {
 
   val tableView: TableView[DLCStatus] = {
-    val paramHashCol = new TableColumn[DLCStatus, String] {
-      text = "Temp Contract Id"
+
+    val eventIdCol = new TableColumn[DLCStatus, String] {
+      text = "Event Id"
       prefWidth = 150
       cellValueFactory = { status =>
-        new StringProperty(status,
-                           "Temp Contract Id",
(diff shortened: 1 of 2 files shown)


nkohen (commented):
Should probably at least include the oracle event id no? Seems odd to have no oracle information at all

Christewart (approved):

benthecarman:
> Should probably at least include the oracle event id no? Seems odd to have no oracle information at all

Added

benthecarman:
Added a copy contract id button

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -6,16 +6,21 @@ import scalafx.beans.property.StringProperty
...
+
+    val eventIdCol = new TableColumn[DLCStatus, String] {
+      text = "Event Id"
       prefWidth = 150
       cellValueFactory = { status =>
-        new StringProperty(status,
-                           "Temp Contract Id",
-                           status.value.tempContractId.hex)
+        val eventIdStr =
+          status.value.oracleInfo.singleOracleInfos.head.announcement.eventTLV.eventId
```
Christewart, line 21: is this always safe? I see `.head` which can throw no such element

Christewart (approved):

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
benthecarman: Yeah, we have to have at least one oracle

Issue #2910, opened 2021-04-16
Exeception when processing transaction that doesn't originate from our wallet


benthecarman:
This line is causing the issue https://github.com/bitcoin-s/bitcoin-s/blob/master/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala#L123

I believe this is because we are only updating the utxo a single time to spending info db, and because of this, when we call this function the `spendingTxIdOpt` has not been set in the db yet, even though the software does know it now. Most likely the fix is to make sure we update the db before calling `getDbsByRelevantBlock`

PR #2914, opened 2021-04-18
2021 04 18 wallet received txo state

Christewart:
This PR isn't intended to change any functionality, rather tighten types from `TxoState` -> `ReceivedState` inside of our wallet code for handling a utxo we are receiving. This is intended to be ground work to get to the bottom of #2910

Previously, we would be too general with our types which allows invalid states inside of the code flow for receiving a txo in our wallet.

Finally, this PR adds `ImmatureCoinbase` to the `ReceivedState` ADT. I have mixed feelings about this. When we added `ImmatureCoinbase` in #2029 it seems there wasn't much deliberation (on the PR any way) about how to classify this state. 

It is complicated, as it is technically receiving funds, but unlike all other `ReceivedStates` they aren't spendable immediately. I guess `ImmatureCoinbase` should be treated how we treat funds with immature locktimes, would we consider those received? I think so, but they aren't necessarily spendable. 

IMO, probably need to add more types to the ADT to allow for more expressiveness in what _exact_ states txos are in when we receive them, but that seems out of scope of this PR. 

If this sounds accurate, I will file an issue to address this in the next release.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -11,7 +11,7 @@ import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutput}
 import org.bitcoins.core.util.TimeUtil
 import org.bitcoins.core.wallet.fee.FeeUnit
 import org.bitcoins.core.wallet.utxo.TxoState._
-import org.bitcoins.core.wallet.utxo.{AddressTag, TxoState}
+import org.bitcoins.core.wallet.utxo.{AddressTag, ReceivedState, TxoState}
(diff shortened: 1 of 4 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -504,17 +506,6 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
         Future.successful(Vector.empty)
 
       case outputsWithIndex =>
-        val count = outputsWithIndex.length
```
Christewart, line 507: Noisy log stuff that I just decided to remove, if there are objections LMK

benthecarman (approved):

Issue #2915, opened 2021-04-18
Bitcoind backend is slow to sync


benthecarman:
Currently when we turn on the server with a bitcoind backend, we fetch blocks in batches in batches of 25 until we catch up with the latest block. Currently this is pretty slow. We should figure out a way to optimize this or, change it to get block filters from bitcoind instead of blocks.

benthecarman:
https://github.com/bitcoin-s/bitcoin-s/blob/master/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala#L24

PR #2916, opened 2021-04-18
2021 04 18 Use akka streams in BitcoindRpcBackendUtil.syncWalletToBitcoind

Christewart:
This reworks parts of the block fetching logic to use akka streams to hopefully improve performance for #2915 .

Now rather than fetching things in batches, we will fetch them on demand using akka streams backpressure. This means things will be fetched when there is demand downstream for them ([see backpressure explaination](https://doc.akka.io/docs/akka/current/stream/stream-flows-and-basics.html#core-concepts))

I have not investigated how much this improves performance yet (if any!). I suspect this isn't a huge bottleneck in performance, rather `Wallet.processBlock()` is mostly likely the bottleneck as detailed in #2596

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -7,7 +7,6 @@ import grizzled.slf4j.Logging
 import org.bitcoins.core.api.node.NodeApi
 import org.bitcoins.core.protocol.blockchain.Block
 import org.bitcoins.core.protocol.transaction.Transaction
-import org.bitcoins.core.util.FutureUtil
 import org.bitcoins.crypto.DoubleSha256Digest
 import org.bitcoins.rpc.client.common.BitcoindRpcClient
 import org.bitcoins.rpc.config.ZmqConfig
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -275,21 +275,9 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
-      outputsToUse = blockHashOpt match {
-        case Some(_) =>
-          outputsBeingSpent.map { out =>
-            if (out.state == TxoState.Reserved)
-              out.copyWithState(TxoState.PendingConfirmationsReceived)
-            else out
-          }
-        case None =>
-          outputsBeingSpent
-      }
```
benthecarman: what is the reason for doing this? If we remove this then the utxo won't change to un-reserved

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -211,7 +193,8 @@ object BitcoindRpcBackendUtil extends Logging {
       interval: FiniteDuration = 10.seconds)(implicit
       system: ActorSystem,
       ec: ExecutionContext): Cancellable = {
-    val atomicPrevCount: AtomicReference[Int] = new AtomicReference(startCount)
+    val numParallelism = Runtime.getRuntime.availableProcessors()
```
benthecarman, line 196: This function is probably fine to leave as is, it is only being called every 10 seconds

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: This diff is gone now and moved to #2919

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: Hmm... so you are saying that the default interval we provide is too fast (`10.seconds`). I think you might be right, as if there is a old wallet that hasn't synced in a long time, we could start syncing twice (if the first sync takes > 10 seconds)?

benthecarman: Yeah, this could be a problem.

Christewart: Do you have a recommendation of what to change it to? Like 5 minutes?

benthecarman: I think what might be better is to leave this as it, but only start the polling after the intial sync is done

Christewart: as per conservation with ben, we are going to address this in #2926

benthecarman (approved):

PR #2919, opened 2021-04-18
2021 04 18 Reset txo state when overwriting spendingTxId

Christewart:
This fixes a bug where we weren't resetting the state of a txo when changing the `spendingTxIdOpt`. If we change the `spendingTxId`, we need to change the txo state to be `TxoState.BroadcastSpent`

We can't use the old spending txids state.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -275,21 +275,9 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           insertTransaction(transaction, blockHashOpt)
         else Future.unit
 
-      // unreserved outputs now they are in a block
-      outputsToUse = blockHashOpt match {
-        case Some(_) =>
-          outputsBeingSpent.map { out =>
-            if (out.state == TxoState.Reserved)
-              out.copyWithState(TxoState.PendingConfirmationsReceived)
-            else out
-          }
-        case None =>
-          outputsBeingSpent
-      }
-
       processed <- Future
         .sequence {
-          outputsToUse.map(markAsSpent(_, transaction.txIdBE))
+          outputsBeingSpent.map(markAsSpent(_, transaction.txIdBE))
         }
         .map(_.toVector.flatten)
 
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -272,21 +275,9 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           insertTransaction(transaction, blockHashOpt)
         else Future.unit
 
-      // unreserved outputs now they are in a block
-      outputsToUse = blockHashOpt match {
```
Christewart, line 276: @benthecarman Sorry, this diff was meant to be on this PR

The reason this isn't need is because this gets done in `markAsSpent()` 

https://github.com/bitcoin-s/bitcoin-s/pull/2919/files#diff-fd826a398083437abb5cf5acafa8af283c5420c626aa902469a5b743b8dfa094R328

benthecarman (approved):

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -334,10 +325,18 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
           logger.debug(
             s"Marked utxo=${updated.toHumanReadableString} as state=${updated.state}"))
         updatedF.map(Some(_))
-      case TxoState.Reserved | TxoState.PendingConfirmationsSpent |
-          BroadcastSpent =>
+      case state @ (TxoState.Reserved | TxoState.PendingConfirmationsSpent |
```
Christewart: This state transistion is probably correct for `TxoState.{Reserived,BroadcastSpent}` but more thought needs to be put into `PendingConfirmationsSpent`. 

What this logic is effectively saying is we are reorg the existing `spendingTxId` that HAS confirmations on it (`PendingConfirmationsSpent`) for a new spending txid that has zero confs. 

Is this what we want?

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -334,17 +325,30 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+            //as we are overriding the previous spending tx
+            //therefore we can no longer use the old Txo state
+            .copyWithState(state = BroadcastSpent)
+        val updatedF = spendingInfoDAO.update(updated)
         updatedF.map(Some(_))
       case TxoState.ImmatureCoinbase =>
         Future.failed(new RuntimeException(
           s"Attempting to spend an ImmatureCoinbase ${out.outPoint.hex}, this should not be possible until it is confirmed."))
-      case TxoState.ConfirmedSpent =>
+      case TxoState.ConfirmedSpent | TxoState.PendingConfirmationsSpent =>
```
Christewart, line 347: I move (`TxoState.PendingConfirmationsSpent`) down here, i think we should treat these two as the same thing as it is overwriting a spendingTxId that has confirmations on it

benthecarman (approved):

PR #2921, opened 2021-04-19
2021 04 19 Cleanup after ourselves in postgres tests

### LINKED ISSUE #2920: Postgres wallet threads not being shutdown after test cases
When used `PG_ENABLED=1` and running `walletTest/test` locally all tests pass, but they threads allocated do not get shutdown. I suspect this is why postgres is always failing on CI. 

Here is what i see in visualvm after running `walletTest/test`

Screenshot from 2021-04-19 06-23-55

Christewart:
This fixes #2920 

This makes sure we are shutting down database connection pools when `WalletAppConfig.stop()` is called.

This PR also fixes test cases `BitcoindBackendTest` & `BitcoindBlockPollingTest` which don't use our test fixtures in a conventional way, which means we need to stop the wallet manually inside the test case itself.

### DIFF
--- a/async-utils/src/main/scala/org/bitcoins/asyncutil/AsyncUtil.scala
+++ b/async-utils/src/main/scala/org/bitcoins/asyncutil/AsyncUtil.scala
@@ -3,7 +3,8 @@ package org.bitcoins.asyncutil
 import org.bitcoins.asyncutil.AsyncUtil.scheduler
 import org.bitcoins.core.api.asyncutil.AsyncUtilApi
 
-import java.util.concurrent.{Executors, TimeUnit}
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{Executors, ThreadFactory, TimeUnit}
 import scala.concurrent._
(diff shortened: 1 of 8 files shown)


async-utils/src/main/scala/org/bitcoins/asyncutil/AsyncUtil.scala
```diff
@@ -153,7 +154,19 @@ abstract class AsyncUtil extends AsyncUtilApi {
...
 object AsyncUtil extends AsyncUtil {
 
-  private[bitcoins] val scheduler = Executors.newScheduledThreadPool(2)
+  private[this] val threadFactory = new ThreadFactory {
+    private val atomicInteger = new AtomicInteger(0)
+
+    override def newThread(r: Runnable): Thread = {
+      val t = new Thread(
+        r,
+        s"bitcoin-s-async-util-${atomicInteger.getAndIncrement()}")
```
Christewart: added to make threads show up with a name on them with visualvm

benthecarman (approved):

PR #2926, opened 2021-04-19
Use filters for bitcoind backend syncing if available

benthecarman:
Built on top of #2916 just look at last commit

This makes it so we use filters instead of blocks to sync if it is available.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -156,16 +156,17 @@ class BitcoinSServerMain(override val args: Array[String])
               if err.getMessage.contains("If we have spent a spendinginfodb") =>
             handleMissingSpendingInfoDb(err, wallet)
         }
-        _ <- BitcoindRpcBackendUtil.syncWalletToBitcoind(bitcoind, wallet)
+        _ = BitcoindRpcBackendUtil
+          .syncWalletToBitcoind(bitcoind, wallet)
+          .flatMap { _ =>
+            if (bitcoindRpcConf.zmqConfig == ZmqConfig.empty) {
+              BitcoindRpcBackendUtil.startBitcoindBlockPolling(wallet, bitcoind)
+            } else Future.unit
+          }
 
-        blockCount <- bitcoind.getBlockCount
         // Create callbacks for processing new blocks
         _ =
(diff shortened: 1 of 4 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -345,8 +345,10 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
         Future.failed(new RuntimeException(
           s"Attempting to spend an ImmatureCoinbase ${out.outPoint.hex}, this should not be possible until it is confirmed."))
       case TxoState.ConfirmedSpent =>
-        Future.failed(new RuntimeException(
-          s"Attempted to mark an already spent utxo ${out.outPoint.hex} with a new spending tx ${spendingTxId.hex}"))
+        if (!out.spendingTxIdOpt.contains(spendingTxId)) {
```
Christewart, line 348: Hmm, did this happen to you in normal node operations? Why is this needed to be added here?

benthecarman: This happened when I moved back the sync height without deleting my db. Imo this should probably be kept incase we reprocess a tx that is confirmed.

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -144,16 +144,17 @@ class BitcoinSServerMain(override val args: Array[String])
           tmpWallet)
         _ = logger.info("Starting wallet")
         _ <- wallet.start()
-        _ <- BitcoindRpcBackendUtil.syncWalletToBitcoind(bitcoind, wallet)
+        _ = BitcoindRpcBackendUtil
```
Christewart, line 159: nit: 

Maybe 

```scala
start = System.currentTimeMillis
logger.info(s"Starting to sync wallet from bitcoind")
...
_ = logger.info(s"Done syncing wallet to bitcoind, took ${System.currentTimeMillis - start}ms")

Christewart (approved):
Took me ~10 minutes to sync 500,000 testnet blocks

Christewart:
Does not compile

https://github.com/bitcoin-s/bitcoin-s/pull/2926/checks?check_run_id=2384860726#step:5:1036

Christewart (approved):

PR #2927, opened 2021-04-19
2021 04 19 Zip Bitcoin-s datadir

Christewart:
Adds `FileUtil.zipDirectory()` to provide a utility to zip a file. 

This leaves a script that can be run `ZipDatadir` to zip your `$HOME/.bitcoin-s/` directory. You should have the option to pick it if you do `appServer/run`

This also adds `BitcoinSAppConfig.zipDatadir(target)` that can be used to zip the datadir too.  

It should be noted, that the chaindb.sqlite takes forever to zip. On mainnet, it takes me ~15 minutes on my machine.

>[info] 2021-04-19T20:31:09UTC INFO [FileUtil$] Zipping file=/home/chris/.bitcoin-s/mainnet/chaindb.sqlite to /tmp/zip-datadir.zip
[info] 2021-04-19T20:46:40UTC INFO [FileUtil$] Done zipping file=/home/chris/.bitcoin-s/mainnet/chaindb.sqlite

I've added a filter for `chaindb.sqlite` to avoid storing this. This of course comes with the trade off of having to sync the chainstate again when restoring from the zip.

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/file/FileUtil.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/file/FileUtil.scala
@@ -1,0 +1,41 @@
+package org.bitcoins.commons.file
+
+import grizzled.slf4j.Logging
+
+import java.io.{FileOutputStream, IOException}
+import java.nio.file.Path
+import java.util.zip.ZipOutputStream
+import java.nio.file.FileVisitResult
+import java.nio.file.Files
+import java.nio.file.SimpleFileVisitor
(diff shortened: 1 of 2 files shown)


app/server/src/main/scala/org/bitcoins/server/ZipTest.scala
```diff
@@ -0,0 +1,24 @@
+package org.bitcoins.server
```
benthecarman: this can be removed

benthecarman (approved):

PR #2933, opened 2021-04-20
Move wallet scheduler into WalletAppConfig

Christewart:
We currently have a bug in our wallet code wrt to [`Wallet.scheduler`](https://github.com/bitcoin-s/bitcoin-s/blob/8ce22583a55d926033edf1a14e539ed2a815ecd6/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L64).

As you can see, this is a val on the abstract class. This means everytime we create a new `Wallet` object, a scheduler gets allocated for it. This means we get a TON of scheduler threads in tests.

Even worse, it appears when [`Wallet.stop()`](https://github.com/bitcoin-s/bitcoin-s/blob/8ce22583a55d926033edf1a14e539ed2a815ecd6/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala#L161) is called, the scheduler doesn't get shutdown (!!). 

This PR does two things

1. Moves the scheduler into `WalletAppConfig`. This is because we should be sharing the same `WalletAppConfig` across all instances of the wallet similar to how we just create one `ActorSystem` for the entire application.
2. Changes from `scheduler.shutdown()` -> `scheduler.shutdownNow()`. This seems to actually shutdown the tasks. I'm still not 100% sure if this is what we want, or there is another underlying problem here. 

This is what our visualvm state looks like on current master ( 9e60f187a3d1fe12f74ff585b4d55c23369bd094 ) . Notice how many threads are still running even after the entire `walletTest/test` suite is complete! You can see there is still `190` live threads after the test suite is done!

You can see there is a bunch of unamed threads in this image (`pool-#-thread-1`).
Screenshot from 2021-04-20 13-08-34

This second image is from 9b73443 . This names the threads with a `ThreadFactory` (`bitcoin-s-wallet-scheduler`). After running this we have 86 live threads, almost 100 threads less. 

Screenshot from 2021-04-20 13-13-58

### DIFF
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -20,6 +20,7 @@ bitcoin-s {
       # see: https://github.com/bitcoin-s/bitcoin-s/pull/1840
       numThreads = 1
       queueSize=5000
+      poolName = "bitcoin-s-db-pool"
       connectionPool = "HikariCP"
       registerMbeans = true
     }
--- a/testkit/src/main/scala/org/bitcoins/testkit/BitcoinSTestAppConfig.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/BitcoinSTestAppConfig.scala
@@ -180,6 +180,7 @@ object BitcoinSTestAppConfig {
          |   driver = "org.postgresql.Driver"
          |   user = "postgres"
          |   password = "postgres"
+         |   poolName = "bitcoin-s-db-pool"
          |   port = $port
          |   numThreads = 1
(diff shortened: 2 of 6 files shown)


benthecarman (approved):
Should we do this for all of our DbAppConfigs

Christewart:
> Should we do this for all of our DbAppConfigs

Not all `DbAppConfigs` need schedulers. 

It seems we do need to look to make sure our subclasses of `DbAppConfig` do call the appropriate `super.stop()` as we do here in `WalletAppConfig` to shut down database connection pools.

Issue #1764, opened 2020-08-05
Redesign the Neutrino mode


rorp:
Currently `P2PClient` uses `DataMessageHandler` to process all incoming messages. `DataMessageHandler` is designed to be context free. It works perfectly for SPV mode, because in this case the server (`bitcoind` node) pushes all required data to the client (`bitcoin-s` node). 

However, in Neutrino mode the client requests the data from the server. Since `DataMessageHandler` is context free, every time it requests the data, it recreates the context. It’s not always easy to do, so it leads to performance issues (eg. https://github.com/bitcoin-s/bitcoin-s/issues/1700 or https://github.com/bitcoin-s/bitcoin-s/issues/1714).

Also using the same `DataMessageHandler` in different states of the node makes its logic really complicated.

To resolve these issues we can do following:

Firstly, the client should not download all filters on initial sync. It should download them only if there are `spk`s to watch or by the user’s request. 

Secondly, we can split `DataMessageHandler` in two versions: one for the initial sync that uses checkpoints to sync the filter headers; and another for normal operation, that keeps track of downloaded block headers, filter headers etc.

“Initial sync” `DataMessageHandler` can use checkpoints generated by the server and issue the `getcfheaders` and `getcfilters` commands without performing expensive database queries. In theory it can use checkpoints to download filter headers for all existing forks.

The rationale for a specialized “normal” `DataMessageHandler` is that it works differently than the “initial sync” one. It can do a flow like this:
1. The server pushes a new block header. The client validates it and attaches it to the appropriate chain. At this point the client knows its height and block hash, and it caches them. The client requests the corresponding filter header using the cached height and the hash without accessing the database.
2. The server returns the filter header. The client validates and attaches it to the appropriate chain using the height and block hash cached on step 1. Now the client requests the corresponding block filter and caches the filter hash.
3. The server returns the block filter. It validates the filter, stores it into the database, matches `spk`s, and if there’s a match, the server requests the block using cached data. If there’s no match, the client removes the cached data.
4. The server returns the block, the client processes the block and removes the cached data.

This approach eliminates the need for most of the database queries the client currently performs to process Neutrino data. Also it separates the initial sync mode into a separate state. Which is desirable to have, because during the initial sync the node is effectively unusable, but in the current design the end user can use a syncing node as if it’s already synced.

rorp:
I tried to sync lnd in neutrino mode. It synced testnet filter headers from neutrino.testnet3.suredbits.com in 8 minutes.

PR #2938, opened 2021-04-21
Peer Message Receiver Refactor

benthecarman:
Refactors `PeerMessageReceiver` to have it's parent `Node` instead of a `DataMessageHandler` and then give the `DataMessageHandler` to the `Node` this will allow the `Node` to have different peers that are all contributing and looking at the same view of the blockchain

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/networking/peer/PeerMessageHandlerTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/networking/peer/PeerMessageHandlerTest.scala
@@ -5,6 +5,7 @@ import org.bitcoins.node.models.Peer
 import org.bitcoins.server.BitcoinSAppConfig
 import org.bitcoins.testkit.BitcoinSTestAppConfig
 import org.bitcoins.testkit.async.TestAsyncUtil
+import org.bitcoins.testkit.chain.ChainUnitTest
 import org.bitcoins.testkit.node.{
   CachedBitcoinSAppConfig,
   NodeTestWithCachedBitcoindNewest,
@@ -12,7 +13,7 @@ import org.bitcoins.testkit.node.{
 }
 import org.scalatest.{FutureOutcome, Outcome}
 
-import scala.concurrent.Future
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.DurationInt
 
 /** Created by chris on 7/1/16.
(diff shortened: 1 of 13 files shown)


node-test/src/test/scala/org/bitcoins/node/networking/peer/PeerMessageHandlerTest.scala
```diff
@@ -39,35 +39,44 @@ class PeerMessageHandlerTest
   implicit protected lazy val chainConfig: ChainAppConfig =
     cachedConfig.chainConf
 
-  behavior of "PeerHandler"
+  override def beforeAll(): Unit = {
+    implicit val chainConf: ChainAppConfig = cachedConfig.chainConf
```
Christewart: This is probably going to need to change as per work on #2943 , we probably don't want to block this work on that though.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,12 +45,14 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
 
   val peer: Peer
 
-  protected val initialSyncDone: Option[Promise[Done]]
```
Christewart, line 48: So this is totally gone now? How do we know when IBD is done for a specific node ?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
```diff
@@ -72,7 +72,10 @@ case class PeerMessageSender(client: P2PClient)(implicit conf: NodeAppConfig)
 
   def sendVersionMessage(chainApi: ChainApi)(implicit
       ec: ExecutionContext): Future[Unit] = {
-    chainApi.getBestHashBlockHeight().flatMap { height =>
+    val heightF =
```
Christewart: Hmm... what is this for ? Why is this needed now?

Christewart (commented):
Related to #1764

node/src/main/scala/org/bitcoins/node/Node.scala
benthecarman: We never used it anywhere, we just use the `syncing` value

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
benthecarman: Ran into this with some of the testing stuff, basically, if our db is empty then we should return a `0`

Christewart: Ok here is a patch to fix this so we can remove this: 

https://github.com/Christewart/bitcoin-s-core/commit/69fa407435501bd0f1e4cf8d585c684f37608a0b

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,12 +45,14 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+  def dataMessageHandler: DataMessageHandler
```
Christewart, line 54: /** The current data message handler.
    * It should be noted that the dataMessageHandler contains
    * chainstate. When we update with a new chainstate, we need to
    * maek sure we update the [[DataMessageHandler]] via [[updateDataMessageHandler()]]
    * to make sure we don't corrupt our chainstate cache */

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -206,7 +198,7 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
       // Get all of our cached headers in case of a reorg
       val cachedHeaders =
         blockchains.flatMap(_.headers).map(_.hashBE.flip)
-      peerMsgSenderF.map(_.sendGetHeadersMessage(cachedHeaders))
+      peerMsgSender.sendGetHeadersMessage(cachedHeaders)
```
Christewart: This returns a `Future`, can we move this up into the for comprehension? I.e. 

```scala
// Get all of our cached headers in case of a reorg
cachedHeaders = blockchains.flatMap(_.headers).map(_.hashBE.flip)
_ <- peerMsgSender.sendGetHeadersMessage(cachedHeaders)
 ```

Christewart (changes requested):

Christewart (approved):

nkohen (approved):

PR #2946, opened 2021-04-22
Fix missing teardown code for MultiWalletRpcTest

Christewart:
We weren't shutting down the actor system and de-allocating resources for this test suite in bitcoind-rpc

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
@@ -278,6 +278,28 @@ trait BitcoindFixturesCachedPairV18
   }
 }
 
+trait BitcoindFixturesCachedPairV19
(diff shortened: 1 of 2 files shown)


testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindFixtures.scala
```diff
@@ -278,6 +278,28 @@ trait BitcoindFixturesCachedPairV18
...
+    val futOutcome = for {
+      pair <- clientsF
+      futOutcome = with2BitcoindsCached(test, pair)
+      f <- futOutcome.toFuture
+    } yield f
+    new FutureOutcome(futOutcome)
+  }
+
+  override def afterAll(): Unit = {
+    super[BitcoindFixturesCachedPair].afterAll()
```
Christewart, line 298: This is what was missing

benthecarman (approved):

PR #2949, opened 2021-04-23
Sync from multiple peers

benthecarman:
Built on top of #2938

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -70,10 +70,11 @@ case class NeutrinoNode(
     } else ()
   }
 
-  def removePeer(peer: Peer): Unit = {
+  def removePeer(peer: Peer): Peer = {
     this.peers = peers.filter(_ != peer)
     this.clients = clients.filter(_.peer != peer)
     this.peerMsgSenders = peerMsgSenders.filter(_.client.peer != peer)
+    peer
   }
 
   override def start(): Future[NeutrinoNode] = {
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -13,8 +13,8 @@ import org.bitcoins.node.networking.peer._
 import scala.concurrent.Future
 
 case class NeutrinoNode(
-    nodePeers: Vector[Peer],
-    _dataMessageHandler: DataMessageHandler,
+    private val _nodePeers: Vector[Peer],
+    private val _dataMessageHandler: DataMessageHandler,
     nodeConfig: NodeAppConfig,
     chainConfig: ChainAppConfig,
     actorSystem: ActorSystem)
@@ -29,9 +29,9 @@ case class NeutrinoNode(
 
   implicit override val chainAppConfig: ChainAppConfig = chainConfig
 
-  var peers: Vector[Peer] = nodePeers
+  protected var peers: Vector[Peer] = _nodePeers
 
-  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+  protected var dataMessageHandler: DataMessageHandler = _dataMessageHandler
 
   override def updateDataMessageHandler(
       dataMessageHandler: DataMessageHandler): NeutrinoNode = {
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -4,12 +4,14 @@ import akka.actor.ActorSystem
 import org.bitcoins.chain.config.ChainAppConfig
 import org.bitcoins.chain.models.BlockHeaderDAO
 import org.bitcoins.core.api.chain.ChainQueryApi.FilterResponse
+import org.bitcoins.core.p2p.ServiceIdentifier
 import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.models.Peer
 import org.bitcoins.node.networking.P2PClient
 import org.bitcoins.node.networking.peer._
 
+import scala.collection.mutable
 import scala.concurrent.Future
 
 case class NeutrinoNode(
@@ -29,9 +31,13 @@ case class NeutrinoNode(
 
   implicit override val chainAppConfig: ChainAppConfig = chainConfig
 
-  protected var peers: Vector[Peer] = _nodePeers
+  private var peers: Vector[Peer] = _nodePeers
 
-  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+  override def getPeers: Vector[Peer] = peers
+
+  private[this] var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+
+  override def getDataMessageHandler: DataMessageHandler = dataMessageHandler
 
   override def updateDataMessageHandler(
       dataMessageHandler: DataMessageHandler): NeutrinoNode = {
@@ -39,7 +45,7 @@ case class NeutrinoNode(
     this
   }
 
-  var clients: Vector[P2PClient] = {
+  private[this] var clients: Vector[P2PClient] = {
     peers.map { peer =>
       val peerMsgRecv: PeerMessageReceiver =
         PeerMessageReceiver.newReceiver(node = this, peer = peer)
@@ -49,13 +55,21 @@ case class NeutrinoNode(
     }
   }
 
-  var peerMsgSenders: Vector[PeerMessageSender] = {
-    clients.map { client =>
-      PeerMessageSender(client)
-    }
+  private[this] var peerMsgSenders: Vector[PeerMessageSender] = {
+    clients.map(PeerMessageSender(_))
   }
 
-  def addPeer(peer: Peer): Unit = {
+  private[this] val peerServices: mutable.Map[Peer, ServiceIdentifier] =
+    mutable.Map.empty
+
+  override def getClients: Vector[P2PClient] = clients
+
+  override def getPeerMsgSenders: Vector[PeerMessageSender] = peerMsgSenders
+
+  override def getPeerServices: Map[Peer, ServiceIdentifier] =
+    peerServices.toMap
+
+  override def addPeer(peer: Peer): Unit = {
     if (!peers.contains(peer)) {
       val peerMsgRecv: PeerMessageReceiver =
         PeerMessageReceiver.newReceiver(node = this, peer = peer)
@@ -64,17 +78,28 @@ case class NeutrinoNode(
                              peerMessageReceiver = peerMsgRecv)
       val peerMsgSender = PeerMessageSender(client)
 
+      peerMsgSender.connect()
+
       this.peers = peers :+ peer
       this.clients = clients :+ client
       this.peerMsgSenders = peerMsgSenders :+ peerMsgSender
     } else ()
   }
 
-  def removePeer(peer: Peer): Peer = {
-    this.peers = peers.filter(_ != peer)
-    this.clients = clients.filter(_.peer != peer)
+  override def removePeer(peer: Peer): Peer = {
     this.peerMsgSenders = peerMsgSenders.filter(_.client.peer != peer)
+    this.clients = clients.filter(_.peer != peer)
+    this.peerServices.remove(peer)
+    this.peers = peers.filter(_ != peer)
     peer
+  }
+
+  override def setPeerServices(
+      peer: Peer,
+      serviceIdentifier: ServiceIdentifier): Unit = {
+    logger.info(s"Setting peer $peer with services $serviceIdentifier")
+    peerServices.put(peer, serviceIdentifier)
+    ()
   }
 
   override def start(): Future[NeutrinoNode] = {
@@ -109,24 +134,16 @@ case class NeutrinoNode(
       filterHeaderCount <- chainApi.getFilterHeaderCount()
       filterCount <- chainApi.getFilterCount()
       blockchains <- blockchainsF
-    } yield {
+
       // Get all of our cached headers in case of a reorg
-      val cachedHeaders = blockchains.flatMap(_.headers).map(_.hashBE.flip)
-      randomPeer.sendGetHeadersMessage(cachedHeaders)
-
-      // If we have started syncing filters headers
-      if (filterHeaderCount != 0) {
-        randomPeer.sendNextGetCompactFilterHeadersCommand(
-          chainApi = chainApi,
-          filterHeaderBatchSize = chainConfig.filterHeaderBatchSize,
-          prevStopHash = header.hashBE)
-
-        // If we have started syncing filters
-        if (filterCount != filterHeaderCount && filterCount != 0)
-          randomPeer.sendNextGetCompactFilterCommand(
-            chainApi = chainApi,
-            filterBatchSize = chainConfig.filterBatchSize,
-            startHeight = filterCount)
+      cachedHeaders = blockchains.flatMap(_.headers).map(_.hashBE.flip)
+      _ <- randomPeer.sendGetHeadersMessage(cachedHeaders)
+    } yield {
+      val height = header.height
+      if (
+        header.height == 0 || height != filterHeaderCount || height != filterCount
+      ) {
+        updateDataMessageHandler(getDataMessageHandler.copy(syncing = true))
       }
 
       logger.info(
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -7,13 +7,14 @@ import org.bitcoins.core.api.chain.ChainQueryApi.FilterResponse
 import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.models.Peer
-import org.bitcoins.node.networking.peer.DataMessageHandler
+import org.bitcoins.node.networking.P2PClient
+import org.bitcoins.node.networking.peer._
 
 import scala.concurrent.Future
 
 case class NeutrinoNode(
-    nodePeer: Peer,
-    dataMessageHandler: DataMessageHandler,
+    nodePeers: Vector[Peer],
+    _dataMessageHandler: DataMessageHandler,
     nodeConfig: NodeAppConfig,
     chainConfig: ChainAppConfig,
     actorSystem: ActorSystem)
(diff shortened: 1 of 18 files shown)


node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,87 @@
...
+  implicit override val system: ActorSystem = actorSystem
 
-  implicit override def nodeAppConfig: NodeAppConfig = nodeConfig
+  implicit override val nodeAppConfig: NodeAppConfig = nodeConfig
 
-  override def chainAppConfig: ChainAppConfig = chainConfig
+  implicit override val chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peer: Peer = nodePeer
+  var peers: Vector[Peer] = nodePeers
```
nkohen: `nodePeers` should be `_nodePeers` and both it and `_dataMessageHandler` should be `private val`s

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,87 @@
...
+
+  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+
+  override def updateDataMessageHandler(
+      dataMessageHandler: DataMessageHandler): NeutrinoNode = {
+    this.dataMessageHandler = dataMessageHandler
+    this
+  }
+
+  var clients: Vector[P2PClient] = {
```
nkohen: generally speaking it isn't great practice to publicly expose your vars and probably instead they should be made private with mutator and accessor methods

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,87 @@
...
+    } else ()
+  }
+
+  def removePeer(peer: Peer): Unit = {
+    this.peers = peers.filter(_ != peer)
+    this.clients = clients.filter(_.peer != peer)
+    this.peerMsgSenders = peerMsgSenders.filter(_.client.peer != peer)
+  }
 
   override def start(): Future[NeutrinoNode] = {
```
nkohen, line 105: How do we handle conflicting peers? I.e. if they have different headers or different filter headers? I don't see the compare via binary search and download and ban code.

Also part of the point of having multiple peers should be to speed up processing (though this may be a next PR, an issue would be nice):
```
The header checkpoints allow the client to download filter headers for different intervals from multiple peers in parallel, verifying each range of 1,000 headers against the checkpoints. 
```

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -62,53 +65,40 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
     */
-  def isInitialized: Future[Boolean] = peerMsgSenderF.flatMap(_.isInitialized())
+  def isInitialized: Future[Boolean] = {
+    val initFs = peerMsgSenders.map(_.isInitialized())
+    Future.sequence(initFs).map(_.forall(_ == true))
+  }
 
-  def isDisconnected: Future[Boolean] =
-    peerMsgSenderF.flatMap(_.isDisconnected())
+  def isDisconnected: Future[Boolean] = {
```
nkohen, line 131: scaladoc

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -30,7 +31,7 @@ case class SpvNode(
 
   override def chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peer: Peer = nodePeer
+  var peers: Vector[Peer] = nodePeers
```
nkohen: ditto `var` comments

node/src/main/scala/org/bitcoins/node/util/BitcoinSNodeUtil.scala
```diff
@@ -1,13 +1,16 @@
...
+
 object BitcoinSNodeUtil {
 
   /** Creates a unique actor name for a actor
     * @param className
     * @return
     */
   def createActorName(className: String): String = {
-    s"${className}-${System.currentTimeMillis()}"
+    // add random number in case we gen actors at the same time
```
nkohen, line 12: I think people generally solve this problem using `AtomicInteger`

nkohen (commented):
Doesn't compile.

Some comments but the main thing is that I believe this is missing the handling of the case where peers do not agree.
Also, I don't think this from the BIP is followed:
```
Each time a new valid block header is received, the client SHOULD request the corresponding filter headers from all eligible peers. If two peers send conflicting filter headers, the client should interrogate them as described above and ban any peers that send an invalid header. 
```

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
benthecarman: set them to `protected`

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
benthecarman: this was the plan for a follow up PR

node/src/main/scala/org/bitcoins/node/util/BitcoinSNodeUtil.scala
benthecarman: Like adding one to `BitcoinSNodeUtil` and incrementing it?

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,88 @@
...
-  implicit override def nodeAppConfig: NodeAppConfig = nodeConfig
+  implicit override val nodeAppConfig: NodeAppConfig = nodeConfig
 
-  override def chainAppConfig: ChainAppConfig = chainConfig
+  implicit override val chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peer: Peer = nodePeer
+  protected var peers: Vector[Peer] = _nodePeers
+
+  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
```
Christewart: This seems like a bad idea. Using `var`s in a public scope is likely going to lead to concurrency bugs

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,88 @@
...
+  implicit override val system: ActorSystem = actorSystem
 
-  implicit override def nodeAppConfig: NodeAppConfig = nodeConfig
+  implicit override val nodeAppConfig: NodeAppConfig = nodeConfig
 
-  override def chainAppConfig: ChainAppConfig = chainConfig
+  implicit override val chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peer: Peer = nodePeer
+  protected var peers: Vector[Peer] = _nodePeers
```
Christewart: Probably should be a `Set`? Although we do need to make sure our `Peer` equality checks actually work

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -279,4 +296,28 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
   override def epochSecondToBlockHeight(time: Long): Future[Int] =
     chainApiFromDb().flatMap(_.epochSecondToBlockHeight(time))
 
+  case object ManagePeersRunnable extends Runnable {
```
Christewart: This needs to be in `NodeAppConfig`, else we are going to have resource related issues. See #2957

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -41,6 +42,52 @@ case class SpvNode(
...
 
+  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+
+  override def updateDataMessageHandler(
+      dataMessageHandler: DataMessageHandler): SpvNode = {
+    this.dataMessageHandler = dataMessageHandler
+    this
+  }
+
+  var clients: Vector[P2PClient] = {
```
Christewart: Let's not use `var`, it's a really bad idea as it is not thread safe. Let's keep immutable data structures and update the internals like `updateDataMessageHandler()`

benthecarman:
Responded to review.

Currently has a bug where it will I already have the filter for this blockhash in the db, even though I don't

```
2021-04-27T04:47:13UTC ERROR [SafeDatabase] org.sqlite.SQLiteException: [SQLITE_CONSTRAINT_PRIMARYKEY]  A PRIMARY KEY constraint failed (UNIQUE constraint failed: cfilters.block_hash)
2021-04-27T04:47:13UTC ERROR [DataMessageHandler] Failed to handle data payload=CompactFilterMessage(Basic, 0000000012c720c66528e7309bce4c6a282361d95c83380e296535a891307398, ByteVector(4 bytes, 0x0139e2d0))
org.sqlite.SQLiteException: [SQLITE_CONSTRAINT_PRIMARYKEY]  A PRIMARY KEY constraint failed (UNIQUE constraint failed: cfilters.block_hash)
```

node/src/main/scala/org/bitcoins/node/util/BitcoinSNodeUtil.scala
nkohen: yup, as a `private val`

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
nkohen: I think this one was missed

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -1,42 +1,109 @@
...
+
+  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
+
+  override def updateDataMessageHandler(
+      dataMessageHandler: DataMessageHandler): NeutrinoNode = {
+    this.dataMessageHandler = dataMessageHandler
+    this
+  }
+
+  private[this] var clients: Vector[P2PClient] = {
```
nkohen, line 48: nit: I think just `private` and `private[this]` are equivalent?

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
nkohen: If the concern is that this overrides something then first of all mark with `override` and secondly probably have the `private var` with a `_` prefix and the def delegate to the var

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -43,14 +40,20 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
 
   lazy val txDAO: BroadcastAbleTransactionDAO = BroadcastAbleTransactionDAO()
 
+  def updateDataMessageHandler(dataMessageHandler: DataMessageHandler): Node
+
+  def addPeer(peer: Peer): Unit
+
+  def removePeer(peer: Peer): Peer
+
+  def setPeerServices(peer: Peer, serviceIdentifier: ServiceIdentifier): Unit
```
nkohen: scaladocs

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -62,53 +65,64 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
-                          peerMessageReceiver = peerMsgRecv)
-      p2p
-    }
+  def getClients: Vector[P2PClient]
+
+  def getPeerMsgSenders: Vector[PeerMessageSender]
+
+  def getPeerServices: Map[Peer, ServiceIdentifier]
+
+  def randomPeer: PeerMessageSender = {
```
nkohen, line 89: scaladoc

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -62,53 +65,64 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
   }
 
-  /** Sends the given P2P to our peer.
-    * This method is useful for playing around
-    * with P2P messages, therefore marked as
-    * `private[node]`.
-    */
-  def send(msg: NetworkPayload): Future[Unit] = {
-    peerMsgSenderF.flatMap(_.sendMsg(msg))
+  def getPeerWithFilters: PeerMessageSender = {
```
nkohen: scaladocs

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -229,36 +256,48 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
           s"Wrote tx=${written.map(_.transaction.txIdBE.hex)} to broadcastable table")
     }
 
+    def attemptBroadcast(attemptedPeers: Vector[PeerMessageSender] =
```
nkohen, line 268: scaladoc

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -30,7 +33,7 @@ case class SpvNode(
 
   override def chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peer: Peer = nodePeer
+  private[this] var peers: Vector[Peer] = _nodePeers
```
nkohen, line 36: ditto

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -41,6 +44,72 @@ case class SpvNode(
     this
   }
 
+  var dataMessageHandler: DataMessageHandler = _dataMessageHandler
```
nkohen: ditto

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
```diff
@@ -101,21 +154,55 @@ object NodeAppConfig extends AppConfigFactory[NodeAppConfig] {
...
+        dmhF.map(dmh => NeutrinoNode(peers, dmh, nodeConf, chainConf, system))
       case NodeType.FullNode =>
         Future.failed(new RuntimeException("Not implemented"))
       case NodeType.BitcoindBackend =>
         Future.failed(new RuntimeException("Use a BitcoindRpcClient instead"))
     }
   }
 }
+
+case class ManagePeersRunnable(node: Node)(implicit ec: ExecutionContext)
```
nkohen, line 203: scaladoc

nkohen (commented):
Looks good, mostly just missing some scaladocs

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
benthecarman: `private[this]`  makes it so other instances of `NeutrinoNode` can't access it

nkohen: I'm 99.99% sure that `private` does exactly that (as opposed to `protected`)

benthecarman:
I think this has some resource usage problems, just syncing filters made my computer really slow

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
Christewart: i think there are weird differences here between scala and java. I think you are right in Java but wrong in Scala.

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -22,25 +25,89 @@ case class NeutrinoNode(
...
+    mutable.Map.empty
+
+  override def getClients: Vector[P2PClient] = clients
+
+  override def getPeerMsgSenders: Vector[PeerMessageSender] = peerMsgSenders
+
+  override def getPeerServices: Map[Peer, ServiceIdentifier] =
+    peerServices.toMap
+
+  override def addPeer(peer: Peer): Unit = {
```
Christewart, line 72: I would prefer returning `NeutrinoNode` from these methods rather than `Unit`. It makes it more functional and as a by product more testable. Basically do the same thing you are doing in `updateDataMessageHandler()`

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -22,25 +25,89 @@ case class NeutrinoNode(
...
+    this.peerMsgSenders = peerMsgSenders.filter(_.client.peer != peer)
+    this.clients = clients.filter(_.peer != peer)
+    this.peerServices.remove(peer)
+    this.peers = peers.filter(_ != peer)
+    peer
+  }
+
+  override def setPeerServices(
+      peer: Peer,
+      serviceIdentifier: ServiceIdentifier): Unit = {
```
Christewart, line 99: Same comment here wrt to `Unit` -> `NeutrinoNode`

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -67,24 +134,16 @@ case class NeutrinoNode(
...
-            filterBatchSize = chainConfig.filterBatchSize,
-            startHeight = filterCount)
+      cachedHeaders = blockchains.flatMap(_.headers).map(_.hashBE.flip)
+      _ <- randomPeer.sendGetHeadersMessage(cachedHeaders)
+    } yield {
+      val height = header.height
+      if (
+        header.height == 0 || height != filterHeaderCount || height != filterCount
+      ) {
+        updateDataMessageHandler(getDataMessageHandler.copy(syncing = true))
```
Christewart, line 146: it sucks to drop this on the ground like this. There maybe useful errors returned. Maybe we want to change the return type of sync to `Future[NeutrinoNode]`?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -70,44 +74,64 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+  def getPeerServices: Map[Peer, ServiceIdentifier]
+
+  def randomPeer: PeerMessageSender = {
+    val peerMsgSenders = getPeerMsgSenders
+    peerMsgSenders(Random.nextInt(peerMsgSenders.size))
   }
 
-  lazy val peerMsgSender: PeerMessageSender = {
-    PeerMessageSender(client)
+  def getPeerWithService(f: ServiceIdentifier => Boolean): PeerMessageSender = {
```
Christewart, line 94: nit: `findRandomPeerWithServices` and scaladoc please. This behaves similar to the `find()` method in the scala collections API. Also this should really return an `Option[PeerMessageSender]` rather than throwing an exception.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -70,44 +74,64 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
   }
 
-  /** Sends the given P2P to our peer.
-    * This method is useful for playing around
-    * with P2P messages, therefore marked as
-    * `private[node]`.
-    */
-  def send(msg: NetworkPayload): Future[Unit] = {
-    peerMsgSender.sendMsg(msg)
+  def getPeerWithBlocks: PeerMessageSender = {
```
Christewart, line 102: Can you add a scala doc? I see why you might want to throw in this case if I understand what `getPeerWithBlocks` means. I interpret it to mean you can obtain a full block from this peer.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -70,44 +74,64 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+    val filter = { s: ServiceIdentifier =>
+      s.nodeWitness && s.nodeNetwork
+    }
+    getPeerWithService(filter)
   }
 
-  /** Checks if we have a tcp connection with our peer */
-  def isConnected: Future[Boolean] = peerMsgSender.isConnected()
+  def getPeerWithFilters: PeerMessageSender = {
```
Christewart, line 109: should be `Option[PeerMessageSender]`. There might be a chance you aren't connected to a peer with filters.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -273,4 +327,73 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+
+    override def run(): Unit = {
+      while (!dataPayloadQueueThread.isInterrupted) {
+        val (payload, peerMsgSender) =
+          try {
+            dataPayloadQueue.take()
+          } catch {
+            case _: java.lang.InterruptedException =>
+              return ()
+            case err: Throwable => throw err
```
Christewart, line 370: I think this should be `case scala.util.control.NonFatail(exn) => `

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -273,4 +327,73 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+          }
+        logger.debug(s"Processing $payload in our data payload queue")
+
+        val resultF = for {
+          updated <- getDataMessageHandler.handleDataPayload(payload,
+                                                             peerMsgSender,
+                                                             self)
+        } yield {
+          updateDataMessageHandler(updated)
```
Christewart, line 379: nothing is being done with this.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -118,10 +118,8 @@ class PeerMessageReceiver(
       sender: PeerMessageSender): Future[PeerMessageReceiver] = {
     //else it means we are receiving this data payload from a peer,
     //we need to handle it
-    node.dataMessageHandler.handleDataPayload(payload, sender).map { handler =>
-      val newNode = node.updateDataMessageHandler(handler)
-      new PeerMessageReceiver(newNode, state, peer)
-    }
+    node.dataPayloadQueue.add((payload, sender))
```
Christewart, line 121: ok I undersand why do you had to create this for now, but in the future we probably want to send the `(payload,sender)` to an Actor which can enforce the single threaded behavior we deserve. Akka Actor's mailbox guarantee messages are processed in a FIFO order. You are pseudo doing this with the code in `Node` around `DataPayloadQueueRunnable`

Christewart (changes requested):

PR #2961, opened 2021-04-24
2021 04 24 bitcoin s scripts

Christewart:
This adds a dedicated bitcoin-s scripts project. This is meant to be a place to place useful utilities for running occasionally against bitcoin-s. 

There are two scripts to stat
1. `ScanBitcoind` - scan against bitcoind between a height range
2. `ZipDatadir` - Zip's the bitcoin-s datadir 

As an example, here is the result of `ScanBitcoind` when counting segwit txs between `675,000-tip`

>Count of segwit txs from height675000 to endHeight=680427 is 5707220. It took 335069ms

### DIFF
--- a/app/scripts/src/main/scala/org/bitcoins/scripts/ScanBitcoind.scala
+++ b/app/scripts/src/main/scala/org/bitcoins/scripts/ScanBitcoind.scala
@@ -3,11 +3,7 @@ package org.bitcoins.scripts
 import akka.NotUsed
 import akka.stream.scaladsl.{Keep, Sink, Source}
 import org.bitcoins.core.protocol.blockchain.Block
-import org.bitcoins.core.protocol.transaction.{
-  BaseTransaction,
-  EmptyTransaction,
-  WitnessTransaction
-}
+import org.bitcoins.core.protocol.transaction.WitnessTransaction
 import org.bitcoins.rpc.client.common.BitcoindRpcClient
 import org.bitcoins.server.BitcoindRpcAppConfig
 import org.bitcoins.server.routes.BitcoinSRunner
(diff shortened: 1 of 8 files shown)


app/scripts/src/main/scala/org/bitcoins/scripts/ZipDatadir.scala
```diff
@@ -0,0 +1,33 @@
+package org.bitcoins.scripts
+
+import org.bitcoins.server.BitcoinSAppConfig
+import org.bitcoins.server.routes.BitcoinSRunner
+
+import java.nio.file.Path
+
+//import java.nio.file.Path
```
benthecarman: Ping

app/scripts/src/main/scala/org/bitcoins/scripts/ScanBitcoind.scala
```diff
@@ -0,0 +1,106 @@
...
+    //in this simple example, we are going to count the number of witness transactions
+    val countSegwitTxs: Block => Int = { block: Block =>
+      var counter = 0
+      block.transactions.map {
+        case _: WitnessTransaction =>
+          counter = counter + 1
+        case _: BaseTransaction | EmptyTransaction =>
+        //do nothing
+      }
+      counter
```
benthecarman: I think you can do `block.transactions.count(_.isInstanceOf[WitnessTransaction])`

benthecarman (changes requested):

Christewart:
How long it took to scan testnet with #2963 

>Count of segwit txs from height=1 to endHeight=1972660 is 14248115. It took 2665709ms

benthecarman (approved):

PR #2967, opened 2021-04-25
Add ability to cancel DLC

benthecarman:
Adds the ability to cancel a DLC, deleting all the data and unreserving the utxos.

This is done in the GUI by right clicking in the table view and selecting `Cancel DLC`

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
@@ -343,11 +343,14 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea)
   def cancelDLC(status: DLCStatus): Unit = {
     status.state match {
       case DLCState.Offered | DLCState.Accepted =>
+        val eventId =
+          status.oracleInfo.singleOracleInfos.head.announcement.eventTLV.eventId
         val confirmed = new Alert(AlertType.Confirmation) {
           initOwner(owner)
           headerText = "Confirm Canceling DLC"
-          contentText = "Are you sure you want to cancel this DLC?\n" +
-            "This cannot be undone."
+          contentText =
+            s"Are you sure you want to cancel this DLC for $eventId?\n" +
+              "This cannot be undone."
         }.showAndWait() match {
           case Some(ButtonType.OK) => true
           case None | Some(_)      => false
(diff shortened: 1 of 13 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -339,6 +340,41 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea)
                               this)
   }
 
+  def cancelDLC(status: DLCStatus): Unit = {
+    if (status.state == DLCState.Offered || status.state == DLCState.Accepted) {
```
Christewart: Please pattern match on the states so we have exhaustiveness warnings if we add a new `DLCState` in the future.

Christewart (changes requested):

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -339,6 +340,43 @@ class DLCPaneModel(resultArea: TextArea, oracleInfoArea: TextArea)
                               this)
   }
 
+  def cancelDLC(status: DLCStatus): Unit = {
+    status.state match {
+      case DLCState.Offered | DLCState.Accepted =>
+        val confirmed = new Alert(AlertType.Confirmation) {
+          initOwner(owner)
+          headerText = "Confirm Canceling DLC"
+          contentText = "Are you sure you want to cancel this DLC?\n" +
```
nkohen: should probably, at the very least, include the `event_id` if not more information from the `status` in this confirm alert

benthecarman: added

dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/WalletDLCSetupTest.scala
```diff
@@ -355,6 +356,70 @@ class WalletDLCSetupTest extends BitcoinSDualWalletTest {
...
+        balance <- walletB.getBalance()
+        reserved <- walletB.spendingInfoDAO.findByTxoState(TxoState.Reserved)
+        dlcOpt <- walletB.findDLC(offer.paramHash)
+      } yield {
+        assert(balance == oldBalance)
+        assert(reserved.isEmpty)
+        assert(dlcOpt.isEmpty)
+      }
+  }
+
```
Christewart, line 422: Would be nice to have a negative test case here for failing to cancel a dlc that is not in a cancellable state

Christewart (approved):

PR #2980, opened 2021-04-28
2021 04 27 wallet fixtures config

Christewart:
More cleanup of Wallet test fixtures.

This PR removes the `implicit` modifier from `def getFreshConfig` from `BaseWalletTest`

https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2021-04-27-wallet-fixtures-config?expand=1#diff-153e9b9a23d5616b9a3449a5c020050af40ae9679d399fb8fca9368959441e6bL42

The reason for doing this is because it is dangerous to bind things implicitly that allocate underlying resources on the OS as it isn't entirely clear that it is happening. For instance, in the `WalletUnitTest` we were binding a fresh configuration rather than using a pre-existing one. As a by product of this, we weren't cleaning up the resources of the thing being implicitly allocated because it wasn't clear that it was happening.

https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2021-04-27-wallet-fixtures-config?expand=1#diff-fc0a23e53629ded335e1f4a9b8b9fbe88cd884bd0ae992e63a8cc5283ccf0204L166

The purpose of this PR is to remove the implicit configuration and make it explicitly passed in places. Now that it is explicitly passed, it is easier to see when it needs to be cleaned up.

In general, we should move the configuration allocation into the fixtures rather than having them as class level parameters  in super classes. That way our normal fixture allocation and destruction code can handle the cleaning up of resources.

### DIFF
--- a/.github/workflows/Linux_2.12_App_Chain_Node_Core_Tests.yml
+++ b/.github/workflows/Linux_2.12_App_Chain_Node_Core_Tests.yml
@@ -29,3 +29,4 @@ jobs:
           key: ${{ runner.os }}-cache
       - name: run tests
         run: sbt ++2.12.12 downloadBitcoind coverage chainTest/test chain/coverageReport chain/coverageAggregate chain/coveralls nodeTest/test node/coverageReport node/coverageAggregate node/coveralls cryptoJVM/test cryptoTestJVM/test cryptoJVM/coverageReport cryptoJVM/coverageAggregate cryptoJVM/coveralls coreTestJVM/test coreJVM/coverageReport coreJVM/coverageAggregate coreJVM/coveralls secp256k1jni/test zmq/test zmq/coverageReport zmq/coverageAggregate zmq/coveralls appCommonsTest/test appServerTest/test oracleServerTest/test
+>chainTest/test;nodeTest/test;cryptoJVM/test;cryptoTestJVM/test;coreTestJVM/test;secp256k1jni/test;zmq/test;appCommonsTest/test;appServerTest/test;oracleServerTest/test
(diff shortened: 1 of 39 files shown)


benthecarman (approved):

.github/workflows/Linux_2.12_App_Chain_Node_Core_Tests.yml
```diff
@@ -27,3 +27,4 @@ jobs:
           key: ${{ runner.os }}-cache
       - name: run tests
         run: sbt ++2.12.12 downloadBitcoind coverage chainTest/test chain/coverageReport chain/coverageAggregate chain/coveralls nodeTest/test node/coverageReport node/coverageAggregate node/coveralls cryptoJVM/test cryptoTestJVM/test cryptoJVM/coverageReport cryptoJVM/coverageAggregate cryptoJVM/coveralls coreTestJVM/test coreJVM/coverageReport coreJVM/coverageAggregate coreJVM/coveralls secp256k1jni/test zmq/test zmq/coverageReport zmq/coverageAggregate zmq/coveralls appCommonsTest/test appServerTest/test oracleServerTest/test
+>chainTest/test;nodeTest/test;cryptoJVM/test;cryptoTestJVM/test;coreTestJVM/test;secp256k1jni/test;zmq/test;appCommonsTest/test;appServerTest/test;oracleServerTest/test
```
benthecarman: I think this is an accident?

benthecarman (changes requested):

.github/workflows/Linux_2.12_App_Chain_Node_Core_Tests.yml
Christewart: Yes, sorry using it to try and reproduce more stuff

benthecarman (approved):

PR #2986, opened 2021-04-28
Refactor BitcoinSRunner to use StartStop[Async]

Christewart:
It seems we have some code in `appServerTest` that doesn't clean up after itself. Since `BitcoinSRunner` did not have a uniform way to stop itself, i decided to refactor this code to use the `StartStopAsync` interface we use elsewhere in the codebase. 

Now we have `.stop()` calls in our unit tests that destroy resources allocated by `ServerRunTest`.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -23,8 +23,7 @@ import org.bitcoins.server.routes.{BitcoinSRunner, Server}
 import org.bitcoins.wallet.Wallet
 import org.bitcoins.wallet.config.WalletAppConfig
 
-import scala.concurrent.duration.DurationInt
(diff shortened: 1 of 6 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -33,177 +34,180 @@ class BitcoinSServerMain(override val args: Array[String])
...
+      _ <- system.terminate()
+    } yield {
+      logger.info(s"Actor system terminated")
+      ()
+    }
+  }
 
-      //initialize the config, run migrations
-      val configInitializedF = conf.start()
+  def startBitcoinSBackend(): Future[Unit] = {
```
Christewart, line 77: The reason this diff is so big is because i had to move `startBitcoinSBackend()` and `startBitcoindBackend()` out of the inside of `startup`

benthecarman (approved):

PR #2936, opened 2021-04-21
Pubkey Refactor

nkohen:
Built on top of #2885

(Actual diff here: https://github.com/nkohen/bitcoin-s-core/compare/update-to-secp-zkp..pubkey-refactor-2)

Commits:
1) Removed ExecutionContexts from `ECKey`.
2) Introduced ECKeyBytes case classes to decouple bytes representing keys from the types representing (internally) valid keys with no direct ties to serialization specifics (such as notions of compression, which makes our code faster by using decompressed by default internally and still compressed by default only when serializing).
3) Fixed things
4) Made ECPrivateKey and ECPublicKey into case classes
5) Introduced `isFullyValid` invariant to the `ECPublicKey` case class with tests still passing!!!

Future TODO: https://github.com/bitcoin-s/bitcoin-s/issues/2947

PR #2988, opened 2021-04-28
Call decompression on public keys less

nkohen:
Built on top of #2936

De-duplicates call to decompression on `ECPublicKey::isFullyValid` and removes decompression entirely from `ScriptWitness` parsing.

EDIT: This PR now also introduces `CurveCoordinate` which is for use in places where 32 bytes represent an x or y coordinate of a secp point (as opposed to `FieldElement` which represents a scalar that can be multiplied by a point, e.g. `fe*G`)


Christewart:
This seems work as expected

Screenshot from 2021-04-28 18-33-35

benthecarman (approved):

PR #2990, opened 2021-04-29
Fix race condition with BitcoindChainHandlerViaZmqTest

Christewart:
Fixes race condition inside of `ChainUnitTest.createChainHandlerWithBitcoindZmq`

The problem is we weren't making sure we mapped on `chainHandlerF`. This means we could start generating a block with bitcoind before our chain infrastructure was setup in the fixture.

I also added a few sanity tests on `BlockHeaderDAO.getBlockchains()` to make sure we are getting the correct header, rather than just asserting we got one chain.

In the future, we really need to get rid of `CachedChainAppConfig` as it can lead to unintended resources allocated -- or unintended resources shared between individual unit tests. Similar to #2980

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
@@ -2,6 +2,7 @@ package org.bitcoins.testkit.chain
 
 import akka.actor.ActorSystem
 import com.typesafe.config.ConfigFactory
+import org.bitcoins.asyncutil.AsyncUtil
 import org.bitcoins.chain.ChainVerificationLogger
 import org.bitcoins.chain.blockchain.sync.ChainSync
 import org.bitcoins.chain.blockchain.{ChainHandler, ChainHandlerCached}
@@ -26,7 +27,7 @@ import org.bitcoins.testkit.chain.models.{
 import org.bitcoins.testkit.fixtures.BitcoinSFixture
 import org.bitcoins.testkit.node.CachedChainAppConfig
 import org.bitcoins.testkit.rpc.BitcoindRpcTestUtil
-import org.bitcoins.testkit.util.{AkkaUtil, ScalaTestUtil}
+import org.bitcoins.testkit.util.ScalaTestUtil
 import org.bitcoins.testkit.{chain, BitcoinSTestAppConfig}
 import org.bitcoins.zmq.ZMQSubscriber
 import org.scalatest._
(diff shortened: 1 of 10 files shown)


testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
```diff
@@ -253,12 +261,24 @@ trait ChainUnitTest
...
                         rawTxListener = None,
                         rawBlockListener = Some(handleRawBlock))
-    zmqSubscriber.start()
 
-    for {
-      _ <- AkkaUtil.nonBlockingSleep(1.second)
+    zmqSubscriber.start()
+    //generate a block and make sure we see it so we know the subscription is complete
+    val subscribedF = for {
+      _ <- chainHandlerF
```
Christewart: this is the bug, we didn't have this

testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala
```diff
@@ -253,12 +261,24 @@ trait ChainUnitTest
...
+      addr <- bitcoind.getNewAddress
+      hash +: _ <- bitcoind.generateToAddress(1, addr)
       chainHandler <- chainHandlerF
+      //wait until we see the hash, to make sure the subscription is started
```
Christewart, line 270: This is the key thing to make sure we have zmq fully started and are receiving blocks

benthecarman (approved):

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -455,8 +455,6 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
       newTags: Vector[AddressTag]): Future[Seq[SpendingInfoDb]] = {
     getRelevantOutputs(transaction).flatMap {
       case Nil =>
-        logger.debug(
```
benthecarman, line 458: nit: maybe just leave these as `trace`

wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -193,7 +193,6 @@ private[wallet] trait UtxoHandling extends WalletLogger {
       _ =
         if (toUpdate.nonEmpty)
           logger.info(s"${toUpdate.size} txos are now confirmed!")
-        else logger.debug("No txos to be confirmed")
```
benthecarman, line 196: ditto

PR #2996, opened 2021-04-29
Add Lnd macaroon to GRPC client settings

benthecarman:
Some cleanup on the `LndRpcClient`. This makes it so all of our calls have the macaroon set and we don't manually need to call `addHeader` and then `invoke` with each call.

https://github.com/akka/akka-grpc/issues/1309#issuecomment-827625537

### DIFF
--- a/lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
+++ b/lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
@@ -4,6 +4,7 @@ import akka.actor.ActorSystem
 import akka.grpc.{GrpcClientSettings, SSLContextUtils}
 import com.google.protobuf.ByteString
 import grizzled.slf4j.Logging
+import io.grpc.{CallCredentials, Metadata}
 import lnrpc._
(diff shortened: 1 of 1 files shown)


lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
```diff
@@ -66,6 +68,24 @@ class LndRpcClient(val instance: LndInstance, binary: Option[File] = None)(
     GrpcClientSettings
       .connectToServiceAt(instance.rpcUri.getHost, instance.rpcUri.getPort)
       .withTrustManager(SSLContextUtils.trustManagerFromStream(certStream))
+      .withCallCredentials(new CallCredentials {
```
Christewart: nit: 

```scala
private val callCredentials = new CallCredentials { ... }

...
 GrpcClientSettings
      .connectToServiceAt(instance.rpcUri.getHost, instance.rpcUri.getPort)
      .withTrustManager(SSLContextUtils.trustManagerFromStream(certStream))
      .withCallCredentials(callCredentials)
...
```

Christewart (approved):

PR #2999, opened 2021-04-29
Added fast tooltips to DLC buttons

nkohen:
and removed Export Result button.

(Ben can you run and verify this works before merging?)

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
@@ -20,18 +20,10 @@ class DLCPane(glassPane: VBox) {
     wrapText = true
   }
 
-  private val exportButton = new Button("Export Result") {
-    alignmentInParent = Pos.BottomLeft
-    onAction = _ => model.exportResult(resultTextArea.text.value)
-  }
-
   private val resultArea = new BorderPane() {
     padding = Insets(10)
     center = resultTextArea
-    bottom = exportButton
   }
-
-  BorderPane.setMargin(exportButton, Insets(10))
 
   private val demoOracleArea = new TextArea {
     editable = false
@@ -74,6 +66,9 @@ class DLCPane(glassPane: VBox) {
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onOffer()
     }
+    tooltip = Tooltip(
+      "Initiates a DLC with the given oracle and contract parameters, generating an Offer message.")
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val acceptButton = new Button {
@@ -81,6 +76,8 @@ class DLCPane(glassPane: VBox) {
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onAccept()
     }
+    tooltip = Tooltip("In response to an Offer, generates an Accept message.")
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val signButton = new Button {
@@ -88,6 +85,8 @@ class DLCPane(glassPane: VBox) {
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onSign()
     }
+    tooltip = Tooltip("In response to an Accept, generates a Sign message.")
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val addSigsButton = new Button {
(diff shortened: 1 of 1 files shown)


benthecarman:
Should add tool tips to the various fields in the offer dialog too

benthecarman (approved):

benthecarman:
Tested and they work for me!

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -52,48 +44,66 @@ class DLCPane(glassPane: VBox) {
...
+      "Initiates a DLC with the given oracle and contract parameters, generating an Offer message.")
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val acceptButton = new Button {
     text = "Accept"
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onAccept()
     }
+    tooltip = Tooltip("In response to an Offer, generates an Accept message.")
```
Christewart, line 79: "Allows you to paste in an offer and decode it. If the parameters are desirable, you can generate an Accept Message"

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -52,48 +44,66 @@ class DLCPane(glassPane: VBox) {
...
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val signButton = new Button {
     text = "Sign"
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onSign()
     }
+    tooltip = Tooltip("In response to an Accept, generates a Sign message.")
```
Christewart, line 88: "Allows you to paste in an Accept Message and generate a sign message"

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -52,48 +44,66 @@ class DLCPane(glassPane: VBox) {
...
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val addSigsButton = new Button {
     text = "Add Sigs"
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onAddSigs()
     }
+    tooltip = Tooltip("In response to a Sign, saves signatures.")
```
Christewart, line 75: I think this might broadcast the funding transaction too?

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -52,48 +44,66 @@ class DLCPane(glassPane: VBox) {
...
+    tooltip.value.setShowDelay(new javafx.util.Duration(100))
   }
 
   private val refundButton = new Button {
     text = "Refund"
     onAction = new EventHandler[ActionEvent] {
       override def handle(event: ActionEvent): Unit = model.onRefund()
     }
+    tooltip = Tooltip(
+      "After the refund timeout, broadcasts the refund transaction to the blockchain.")
```
Christewart, line 117: `Paste in contract id to broadcast the refund transaction`

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
nkohen: This is too long, I want Peter to be able to hover over each of them and quickly see that the order is

Offer -> Accept -> Sign -> Add Sigs -> Get Funding

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
nkohen: Same here, I think the current wording more directly clarifies that you click this button when you receive an accept

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
nkohen: Again, I don't think the tooltip is the place to ask for an action like "paste" or to mention a contract id is needed, those are asked for after you click the button. The goal of the tooltip is to let the user know what the button will do and when to press it

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
Christewart: Yeah, I think the purpose of the tool tip is to answer the question "What the heck is an offer?" Whereas the question you seem interested in answering is "What is the next mesage after offer?"

I think we might be trying to solve different problems. I think we need to conceptually agree on what the question we are trying to answer is before writing these messages. We should probably resolve that difference before continuing forward cc @benthecarman , would be curious to hear your opinions

benthecarman: I agree with nadav here, I think his wording is more concise and clear

benthecarman:
needs rebase

benthecarman (approved):

PR #3006, opened 2021-05-02
Use bitcoind v0.21.1

benthecarman:
There were no RPC changes that we need to account for

https://github.com/bitcoin/bitcoin/blob/0.21/doc/release-notes.md

### DIFF
--- a/bitcoind-rpc/bitcoind-rpc.sbt
+++ b/bitcoind-rpc/bitcoind-rpc.sbt
@@ -25,7 +25,7 @@ TaskKeys.downloadBitcoind := {
     "0.18.99" // TODO: change this when new version compiled on suredbits server
 
   val versions =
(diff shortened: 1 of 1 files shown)


Christewart (approved):

Christewart:
FWIW, this doesn't actually use the bitcoind binary it seems for v21 tests. If I run tests i still use 0.21.0

benthecarman:
> FWIW, this doesn't actually use the bitcoind binary it seems for v21 tests. If I run tests i still use 0.21.0

I think you need to delete 0.21.0

Christewart:
> > FWIW, this doesn't actually use the bitcoind binary it seems for v21 tests. If I run tests i still use 0.21.0
> 
> I think you need to delete 0.21.0

Yep, this worked.

PR #3010, opened 2021-05-02
Remove remove request rejection duplication in ServerRoute

ariskk:
Hey 👋
Some unsolicited refactoring:
This is removing request rejection duplication when handling server commands.
Also, you don't seem to be converting routes to directives anywhere, so no need for `StandardRoute` (basic `Route` should suffice).


benthecarman (approved):
This looks great, thank you!

PR #3022, opened 2021-05-03
Add getbalances cli command

benthecarman:
from #3021

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
+++ b/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
@@ -100,7 +100,7 @@ sealed abstract class Satoshis extends CurrencyUnit {
 
   override def toString: String = {
     val num = toLong
-    val postFix = if (num == 1) "sat" else "sats"
+    val postFix = if (num == 1 || num == -1) "sat" else "sats"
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
```diff
@@ -100,7 +100,7 @@ sealed abstract class Satoshis extends CurrencyUnit {
 
   override def toString: String = {
     val num = toLong
-    val postFix = if (num == 1) "sat" else "sats"
+    val postFix = if (num == 1 || num == -1) "sat" else "sats"
```
Christewart, line 103: why this?

Christewart (approved):

core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala
benthecarman: Found from this test, we would print `-1 sats` instead of `-1 sat`
https://github.com/bitcoin-s/bitcoin-s/pull/3022/files#diff-475b69a0d0184d6d4c91c43ce3a2b0580d01b38f8a6e1f8c1b978c23688e19a6R372

PR #3025, opened 2021-05-04
Optimize shift operations in Number

Christewart:
This removes useless checks on our shift operations in Number.scala

This surprisingly seemed like a bottleneck when decoding golomb filters sent from the network, specifically here

https://github.com/bitcoin-s/bitcoin-s/blob/c90f318fd776f366acb7b2c139650a5e05d0df0a/core/src/main/scala/org/bitcoins/core/gcs/GCS.scala#L143


benthecarman:
Not compiling, need to optimize imports

benthecarman (approved):

nkohen (approved):

PR #3028, opened 2021-05-04
Show DLC details when signing DLC, reoder accept details

benthecarman:
image


nkohen (approved):
ACK mod CI

Christewart (commented):
So should I file an issue for removing the oracle public key? The checking against the offer isn't implemented in this PR right?

benthecarman:
> So should I file an issue for removing the oracle public key? The checking against the offer isn't implemented in this PR right?

The oracle public key is fetched from the offer, the only thing you enter is the accept message.

PR #3031, opened 2021-05-04
Added a dust threshold to GUI plots

nkohen:
dust-line
orange version of ^


Christewart:
I would suggest making the orange line thicker and dashed if possible. I think at first glance I would miss the orange line if I was a new user coming to the wallet. A dashed line signifies -- to me any way -- that this is a cut off.

benthecarman (approved):

Issue #3045, opened 2021-05-06
Single threaded behavior with SigPointComputer


Christewart:
Here is how `SigPointComputer` is currently defined

https://github.com/Christewart/bitcoin-s-core/blob/fa80f36d2fb341d56f970954e4fd3a211041b0b7/core/.jvm/src/main/scala/org/bitcoins/core/protocol/dlc/SigPointComputer.scala#L9

```scala
case class SigPointComputer(private val computeSigPoint: () => ECPublicKey) {
  private val sigPointF = Future(computeSigPoint())(ExecutionContext.global)

  lazy val compute: ECPublicKey = {
    Await.result(sigPointF, 20.seconds)
  }
}
```

I'm running some benchmarks this morning that can be found here: 

https://github.com/Christewart/bitcoin-s-core/tree/2021-05-06-sigpoint-bench

and the test I am running is 

> dlcWalletTest/testOnly *DLCMultiOracleNumericExecutionTest* -- -z "execute as the recipient"

Here is the behavior I see when watching this from visualvm 

![Screenshot from 2021-05-06 07-50-13](https://user-images.githubusercontent.com/3514957/117301264-134ebf80-ae40-11eb-9e6b-0787d348eb96.png)

From that image, it appears that only one thread is actually executing the sig point computing code. The rest of the threads are idle. 

If you look carefully at the left hand side of hte image, you can see my crude timing benchmarks of how it takes for accept to complete (`92337ms` in this case)

Christewart:
Consensus between @nkohen and I is to remove `SigPointComputer` all together as it isn't working, and use pre-computed tables for speeding up sig point computation

Christewart:
Removed in #3089

PR #3046, opened 2021-05-06
2021 05 06 cet sigs async

Christewart:
This is PR adds a new method called `createCETSigsAsync()`. This is the bottleneck when creating accept and sign messages. 

This method determines how many cores (`c`) you have available on your machine, and the creates `c` batches of of cets to be computed rather than `createCETSigs()` which just has one batch of signatures. 

This optimization scales relative to the number of cores available on your machine, on a single core machine this optimization is likely slightly slower (thread/execution context overhead) than its synchronous counterpart `createCETSigs()`. Most modern machines have multiple cores. 

I benchmarked it by running 

>dlcWalletTest/testOnly *DLCMultiOracleNumericExecutionTest* -- -z "execute as the recipient"

and observing the accept flow. On my machine, it seems that `createCETSigsAsync` takes 1/6  the time for the test case above

1137ms on 28d2ac5

6419ms on db8af3759b217cbd64d0418fbe391dc0ad6a2f20

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala
@@ -103,9 +103,16 @@ object FutureUtil {
       elements: Vector[T],
       f: Vector[T] => Future[U],
       batchSize: Int)(implicit ec: ExecutionContext): Future[Vector[U]] = {
-    val batches = elements.grouped(batchSize).toVector
-    val execute: Vector[Future[U]] = batches.map(b => f(b))
-    val doneF = Future.sequence(execute)
-    doneF
+    require(
+      batchSize > 0,
+      s"Cannot have batch size less than or equal to zero, got=$batchSize")
+    if (elements.isEmpty) {
+      Future.successful(Vector.empty)
+    } else {
+      val batches = elements.grouped(batchSize).toVector
+      val execute: Vector[Future[U]] = batches.map(b => f(b))
+      val doneF = Future.sequence(execute)
+      doneF
+    }
   }
 }
--- a/dlc/src/main/scala/org/bitcoins/dlc/sign/DLCTxSigner.scala
+++ b/dlc/src/main/scala/org/bitcoins/dlc/sign/DLCTxSigner.scala
@@ -14,6 +14,7 @@ import org.bitcoins.core.protocol.{Bech32Address, BitcoinAddress}
 import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import org.bitcoins.core.psbt.PSBT
 import org.bitcoins.core.script.crypto.HashType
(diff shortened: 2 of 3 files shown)


benthecarman:
LGTM, however we should use this function in tests

https://github.com/bitcoin-s/bitcoin-s/blob/d44893e9ee83090aff4ad0efc98d29324ef5dc84/dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTxGen.scala#L248

benthecarman:
Tests on mac keep failing

Christewart:
> LGTM, however we should use this function in tests
> 
> https://github.com/bitcoin-s/bitcoin-s/blob/d44893e9ee83090aff4ad0efc98d29324ef5dc84/dlc/src/main/scala/org/bitcoins/dlc/testgen/DLCTxGen.scala#L248

It looks like these test are not async. Maybe that is intentional @nkohen ? Is futurifying all that worth the performance improvement in tests?

core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala
```diff
@@ -103,9 +103,16 @@ object FutureUtil {
       elements: Vector[T],
       f: Vector[T] => Future[U],
       batchSize: Int)(implicit ec: ExecutionContext): Future[Vector[U]] = {
-    val batches = elements.grouped(batchSize).toVector
-    val execute: Vector[Future[U]] = batches.map(b => f(b))
-    val doneF = Future.sequence(execute)
-    doneF
+    require(
```
benthecarman, line 106: Should pull down these changes to `master`

benthecarman (approved):

dlc/src/main/scala/org/bitcoins/dlc/sign/DLCTxSigner.scala
```diff
@@ -188,6 +190,36 @@ case class DLCTxSigner(
     CETSignatures(cetSigs, refundSig)
   }
 
+  /** Creates CET signatures async */
+  def createCETSigsAsync()(implicit
```
nkohen, line 194: I'm fine with this for now but this will likely be deleted/rewritten when I've done the sigpoint computation refactor

nkohen (commented):
My take-away from this is that the current sigpoint computation parallelization is essentially not doing very much? Maybe because it uses the global execution context? I think this PR should probably revert the existence of `SigPointComputer` and re-synchronize the computation of individual sigpoints in `OracleOutcome.scala`.

Christewart:
> My take-away from this is that the current sigpoint computation parallelization is essentially not doing very much? Maybe because it uses the global execution context? I think this PR should probably revert the existence of `SigPointComputer` and re-synchronize the computation of individual sigpoints in `OracleOutcome.scala`.

I think this is right, see #3045

PR #3048, opened 2021-05-06
Change outpoint output format for cli commands

### LINKED ISSUE #3017: Change listreservedutxos output format
For some reason we return the raw serialized outpoint in the `listreservedutxos` command. However if you look at what `lockunspent` requires it is the big endian txid and and the vout. It seems very hard to use these in tandem with each other, we should make this easier to do.

To manually do this currently, i have to actually truncate the vout from the raw bytes, and then go and flip the endianness of the txid. It is kind of painful at the moment :/

benthecarman:
Closes #3017

changes to `txIdBE:vout`

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -415,7 +415,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
         assert(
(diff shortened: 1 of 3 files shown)


Christewart (changes requested):
So this seems to be another new format? With this PR we will have three ways (maybe only 2, if we totally killed the old format) 

1. Raw outpoint binary level serialization
2. `txid:idx`
3. The outpoint format needed for lockunspent

Let's just use what `lockunspent` requires as input? I.e. 

```
{
  "outpoint": { 
    "vout" : 0
    "txid" : "..."
  },
  "value": 1000
}
```

If the outpoint is nested, we can then use tools like `jq` to just parse this datastructure and pass that directly into `lockunspent` i believe

Christewart (changes requested):
This needs to be updated and then LGTM 

https://bitcoin-s.org/docs/next/wallet/wallet-rpc#listreservedutxos

Christewart (approved):

PR #3054, opened 2021-05-07
Remove SigPointComputer

Christewart:
After filing #3045 and seeing the performance improvements in #3046 @nkohen and I decided the abstraction isn't useful


benthecarman:
Won't this harm performance

Christewart:
> Won't this harm performance

That is what we thought, but after watching how our current code performs in visualvm it seems we have single threaded behavior (!). See #3045. 

With #3046 we restore the multi threaded behavior. I'm not sure why the current abstraction we have isn't working as expected.

Christewart:
I encourage you to test it out yourself and confirm the behavior. The test cases I was running against are listed in #3045 . It would be good to confirm that the machine i am isn't being weird or whatever.

Christewart:
Replaced by #3089

Issue #2801, opened 2021-03-16
Parallelization of TransactionProcessing causes errors on postgres


Christewart:
This was discovered on #2744 

https://github.com/bitcoin-s/bitcoin-s/blob/b032f4da3bd83a224370a66677a47b23dd4304b0/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala#L254-L255

Both `processIncomingUtxos` and `processOutgoingUtxos` will insert the transaction into the database if it has an `input`/`output` that is relevant. 

This can cause a primary key issue on postgres, as the transaction might already exist in the tx_table if it has _both_ an input and output that is relevant to us.

We should think about restructuring this so we can run this funcitons in parallel in the future, but we are reverting it for now in #2744

PR #3063, opened 2021-05-10
2021 05 09 received utxos

Christewart:
This PR does a few things in the wallet project

1. Renames `Incoming` -> `Received` in wallet methods. This is to be consistent with our `TxoState.Received` type hierarchy
2. Adds parameters `spendingInfoDbs` to `processedReceivedUtxos()` and `processedSpentUtxos()`. This allows us to read from the database in parallel and pass in the appropriate `SpendingInfoDb`s into those methods. Since we can't call these two methods in parallel, hopefully this is a step to re-introducing parallel behavior by determining if the transaction needs to be inserted at all. In the future we should be able to insert the tx if either of those methods returns something, and remove transaction table insertions inside of `processReceivedUtxos()` and `processSpentUtxos()`. See #2801 for more details.
3. Change some `Seq` types to `Vector`. `Seq` can have surprising performance behavior, while Vector is near constant time for every operation on the `Vector`.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -450,7 +450,7 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
   }
 
   private def getRelevantOutputs(
-      transaction: Transaction): Future[Seq[OutputWithIndex]] = {
+      transaction: Transaction): Future[Vector[OutputWithIndex]] = {
     val spks = transaction.outputs.map(_.scriptPubKey)
     scriptPubKeyDAO.findScriptPubKeys(spks.toVector).map { addrs =>
       val withIndex =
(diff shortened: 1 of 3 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -191,40 +191,43 @@ private[wallet] trait TransactionProcessing extends WalletLogger {
...
+      Future.sequence(processedVec)
+    }
+  }
 
   /** Searches for outputs on the given transaction that are
     * being spent from our wallet
     */
   protected def processSpentUtxos(
       transaction: Transaction,
+      outputsBeingSpent: Seq[SpendingInfoDb],
```
benthecarman: can use `Vector` here

benthecarman (commented):
Otherwise LGTM

benthecarman (approved):

PR #3065, opened 2021-05-10
Remove GUI tabs

benthecarman:
Updates to have only a single page

image


Christewart (approved):

Christewart:
Why is this targeting hodl-mccormack? Can we just target normal adaptor-dlc?

benthecarman:
> Why is this targeting hodl-mccormack? Can we just target normal adaptor-dlc?

fixed

Issue #2660, opened 2021-02-12
DLCFeeTestVectorTest timing out


nkohen:
Spuriously, the DLCFeeTestVectorTest which goes (synchronously) through `dlc_fee_test.json` is timing out on CI.

This is not re-producible for me locally (in fact I cannot get this test to take longer than a second, let alone 2 minutes).

This failure happens on both Linux and Mac, both scala 2.12 and scala 2.13.

I have no leads.

PR #3068, opened 2021-05-11
Pulled down dlc and dlcTest projects into core and dlcTest

nkohen:
Don't be too shocked by the lines changed, most of that is a couple JSON tests!


Christewart (approved):
LGTM with the understanding that #2660 needs to be solved on a subsequent PR and will effectively break CI.

PR #3082, opened 2021-05-14
Fix routes spec

nkohen:
Fixes RoutesSpec

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -1051,7 +1051,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
       Post() ~> route ~> check {
         assert(contentType == `application/json`)
         assert(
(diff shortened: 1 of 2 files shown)


benthecarman:
Oh this was done because we changes the return to have the full tx so the user could broadcast in case thebnode failed to broadcast

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -1168,8 +1168,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
       Post() ~> route ~> check {
         assert(contentType == `application/json`)
         assert(
-          responseAs[
-            String] == s"""{"result":"${EmptyTransaction.hex}","error":null}""")
+          responseAs[String] == s"""{"result":"0000000000000000000000000000000000000000000000000000000000000000","error":null}""")
```
benthecarman: Ah, yeah this is a `DoubleSha256Digest.empty.hex`, can replace that here

benthecarman (approved):

PR #3089, opened 2021-05-14
Implement createCETsAndCETSigsAsync() to fix performance issue in test

Christewart:
The purposes of this PR is to fix this CI error by making code async which improves performance.

https://github.com/bitcoin-s/bitcoin-s/runs/2570652014#step:5:2297

The reason this wasn't async in the first place is because of #3045 . This PR replaces #3054 as I had to remove `SigPointComputer` to get this code working (not sure why). 

It seems when i run this locally with `DISABLE_SECP256K1=true` I end up having a 2x performance improvement. When I try to run this locally without the async improvements i get test framework timeouts. 

>[info] DLCClientTest:
[info] AdaptorDLCClient
[info] - should be able to construct and verify with ScriptInterpreter every tx in a DLC for the normal enum case *** FAILED *** (7 minutes, 37 seconds)
[info]   The test did not complete within the specified 300000000000 nanoseconds time limit. (AsyncTimeLimitedTests.scala:99)
[info]   org.scalatest.exceptions.TestFailedDueToTimeoutException:
[info]   at java.base/java.lang.Thread.getStackTrace(Thread.java:1596)
[info]   at org.scalatest.concurrent.TimeLimits$.failAfterImpl(TimeLimits.scala:277)
[info]   at org.scalatest.concurrent.TimeLimits.failAfter(TimeLimits.scala:231)
[info]   at org.scalatest.concurrent.TimeLimits.failAfter$(TimeLimits.scala:230)
[info]   at org.bitcoins.dlc.DLCClientTest.failAfter(DLCClientTest.scala:34)


core/src/main/scala/org/bitcoins/core/protocol/dlc/sign/DLCTxSigner.scala
```diff
@@ -225,10 +225,42 @@ case class DLCTxSigner(
...
+        batchSize = batchSize)
+    }
+
+    val refundSig = signRefundTx
+
+    for {
+      cetsAndSigsNested <- cetsAndSigsF
+      cetsAndSigs = cetsAndSigsNested.flatten
+      (msgs, cets, sigs) = cetsAndSigs.unzip3
+    } yield (CETSignatures(msgs.zip(sigs), refundSig), cets)
```
benthecarman, line 251: Might be worthing putting some `require`s here to make sure things are ordered how we expect them to be

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/OracleOutcome.scala
```diff
@@ -29,7 +29,7 @@ sealed trait OracleOutcome {
 
   /** The adaptor point used to encrypt the signatures for this corresponding CET. */
   def sigPoint: ECPublicKey = {
-    SigPointComputer(() => computeSigPoint).compute
+    computeSigPoint
   }
```
nkohen: This should be stored in a `val` in the end, as it currently stands this appears to re-compute on each call.

core/src/main/scala/org/bitcoins/core/protocol/dlc/sign/DLCTxSigner.scala
Christewart: If you take a look at the implement of `FutureUtil.batchAndParallelExecute()` it preserves ordering: 

https://github.com/bitcoin-s/bitcoin-s/blob/5d6ba1235597a55adfbfcb29126e3a4dcbf7ebb3/core/src/main/scala/org/bitcoins/core/util/FutureUtil.scala#L108

assuming `f` preserves ordering. 

We seemed to preserve ordering in the sync version, but here is the code that gets called as `f` in the `batchAndParallelExecute()`

https://github.com/bitcoin-s/bitcoin-s/blob/5d6ba1235597a55adfbfcb29126e3a4dcbf7ebb3/core/src/main/scala/org/bitcoins/core/protocol/dlc/sign/DLCTxSigner.scala#L133

benthecarman:
Compile issues: 

https://github.com/bitcoin-s/bitcoin-s/pull/3089/checks?check_run_id=2586884143#step:5:76

nkohen (approved):

benthecarman (approved):

PR #3036, opened 2021-05-04
Add offer and accept flow for numeric events

benthecarman:
Needs testing and probably some code cleanup

PR #3092, opened 2021-05-15
Fix Node.sync() bug 

Christewart:
The problem here is we started syncing filters from our chainstates current best block header. This is logically incorrect because we need to start syncing filter headers from the best filter headers block hash. 

Previously, we would miss filters because we wouldn't start sync from the best filter header block rather, just the best block hash we have.

fixes these bugs that are seen often on CI

https://github.com/bitcoin-s/bitcoin-s/pull/3036/checks?check_run_id=2586256057#step:5:1083

Screenshot from 2021-05-14 20-09-55

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -3,7 +3,9 @@ package org.bitcoins.node
 import akka.actor.ActorSystem
 import org.bitcoins.chain.config.ChainAppConfig
 import org.bitcoins.chain.models.BlockHeaderDAO
+import org.bitcoins.core.api.chain.ChainApi
 import org.bitcoins.core.api.chain.ChainQueryApi.FilterResponse
+import org.bitcoins.core.api.chain.db.CompactFilterHeaderDb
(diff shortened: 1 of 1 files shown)


node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -59,39 +61,55 @@ case class NeutrinoNode(
...
       // Get all of our cached headers in case of a reorg
-      val cachedHeaders = blockchains.flatMap(_.headers).map(_.hashBE.flip)
-      peerMsgSender.sendGetHeadersMessage(cachedHeaders)
-
-      // If we have started syncing filters headers
-      if (filterHeaderCount != 0) {
-        peerMsgSender.sendNextGetCompactFilterHeadersCommand(
-          chainApi = chainApi,
-          filterHeaderBatchSize = chainConfig.filterHeaderBatchSize,
-          prevStopHash = header.hashBE)
```
Christewart, line 80: this is the bug, it doesn't make sicne to use our best block header's block hash. We need to use our best filter header's block hash to make sure filter headers are contiguous

benthecarman (approved):

PR #3100, opened 2021-05-18
DLC Wallet database rework

benthecarman:
Refactors to split up the database to store things in individualized tables, there are no more `Vector`s in a column and very minimal TLVs (only event descriptor, contract descriptor, and oracle params).


db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
```diff
@@ -166,7 +166,7 @@ trait DbManagement extends Logging {
           err)
         //maybe we have an existing database, so attempt to baseline the existing
         //database and then apply migrations again
-        flyway.baseline()
+//        flyway.baseline()
```
Christewart: i think we might be able to remove htis and the comments above

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,138 @@
...
 
-CREATE TABLE "wallet_dlc_refund_sigs"
+CREATE TABLE "refund_sigs"
 (
-    "param_hash"   TEXT    NOT NULL,
-    "is_initiator" INTEGER NOT NULL,
-    "refund_sig"   TEXT    NOT NULL,
-    constraint "fk_param_hash" foreign key ("param_hash") references "wallet_dlcs" ("param_hash") on update NO ACTION on delete NO ACTION
+    "dlc_id"        TEXT PRIMARY KEY,
+    "accept_sig"    TEXT NOT NULL,
```
Christewart: why is this `accept_sig` ? Can't you be the acceptor and and then this would need to be `offer_sig`?

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCMessage.scala
```diff
@@ -19,12 +19,9 @@ sealed trait DLCMessage
...
 object DLCMessage {
 
-  def calcParamHash(
-      contractInfo: ContractInfo,
-      timeouts: DLCTimeouts): Sha256DigestBE = {
-    CryptoUtil
-      .sha256(contractInfo.bytes ++ timeouts.bytes)
-      .flip
+  def calcDLCId(outPoints: Vector[TransactionOutPoint]): Sha256Digest = {
+    val bytes = outPoints.sortBy(_.hex).foldLeft(ByteVector.empty)(_ ++ _.bytes)
```
nkohen: nit: More efficient to do `outPoints.map(_.bytes).sorted.foldLeft(ByteVector.empty)(_ ++ _)`

db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
nkohen: intentional?

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,138 @@
...
-    "param_hash"   TEXT    NOT NULL,
-    "is_initiator" INTEGER NOT NULL,
-    "sig_point"    TEXT    NOT NULL,
-    "signature"    TEXT    NOT NULL,
-    constraint "fk_param_hash" foreign key ("param_hash") references "wallet_dlcs" ("param_hash") on update NO ACTION on delete NO ACTION
+    "dlc_id"        TEXT    NOT NULL,
+    "index"         INTEGER NOT NULL,
+    "sig_point"     TEXT    NOT NULL,
+    "accept_sig"    TEXT    NOT NULL,
+    "initiator_sig" TEXT,
```
nkohen, line 120: should probably be called `sign_sig` for consistency?

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,138 @@
...
+    "funding_pub_key"  TEXT    NOT NULL,
+    "payout_address"   TEXT    NOT NULL,
+    "payout_serial_id" INTEGER NOT NULL,
+    "collateral"       INTEGER NOT NULL,
+    "change_address"   TEXT    NOT NULL,
+    "change_serial_id" INTEGER NOT NULL,
+    constraint "fk_dlc_id" foreign key ("dlc_id") references "global_dlc_data" ("dlc_id") on update NO ACTION on delete NO ACTION
+);
+
+CREATE TABLE "accept_dlc_data"
```
nkohen, line 88: shouldn't this have a negotiation fields entry?

dlc-wallet/src/main/resources/sqlite/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,97 +1,129 @@
...
     "payout_serial_id" INTEGER      NOT NULL,
-    "total_collateral" INTEGER      NOT NULL,
+    "collateral"       INTEGER      NOT NULL,
     "change_address"   VARCHAR(254) NOT NULL,
     "change_serial_id" INTEGER      NOT NULL,
-    constraint "fk_param_hash" foreign key ("param_hash") references "wallet_dlcs" ("param_hash") on update NO ACTION on delete NO ACTION
+    constraint "fk_dlc_id" foreign key ("dlc_id") references "global_dlc_data" ("dlc_id") on update NO ACTION on delete NO ACTION
+);
+
+CREATE TABLE "accept_dlc_data"
```
nkohen, line 88: ditto

dlc-wallet/src/main/resources/sqlite/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,97 +1,129 @@
...
-    "param_hash"   VARCHAR(254) NOT NULL,
-    "is_initiator" INTEGER      NOT NULL,
-    "sig_point"    VARCHAR(254) NOT NULL,
-    "signature"    VARCHAR(254) NOT NULL,
-    constraint "fk_param_hash" foreign key ("param_hash") references "wallet_dlcs" ("param_hash") on update NO ACTION on delete NO ACTION
+    "dlc_id"        VARCHAR(254) NOT NULL,
+    "index"         INTEGER      NOT NULL,
+    "sig_point"     VARCHAR(254) NOT NULL,
+    "accept_sig"    VARCHAR(254) NOT NULL,
+    "initiator_sig" VARCHAR(254),
```
nkohen, line 120: ditto

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -0,0 +1,247 @@
...
+import org.bitcoins.core.util.FutureUtil
+import org.bitcoins.core.wallet.utxo.{InputInfo, ScriptSignatureParams}
+import org.bitcoins.crypto.Sha256Digest
+import org.bitcoins.dlc.wallet.DLCWallet
+import org.bitcoins.dlc.wallet.models._
+import scodec.bits._
+
+import scala.concurrent._
+
+trait DLCDataManagement { self: DLCWallet =>
```
nkohen: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  private[wallet] def getUsedOracleAnnouncements(
```
nkohen, line 40: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -0,0 +1,247 @@
...
+      nonceDbs: Vector[OracleNonceDb]): Vector[OracleAnnouncementV0TLV] = {
+    getOracleAnnouncementsWithId(announcementIds, announcementData, nonceDbs)
+      .map(_._1)
+  }
+
+  private[wallet] def getOracleAnnouncementsWithId(
+      announcementIds: Vector[DLCAnnouncementDb],
+      announcementData: Vector[OracleAnnouncementDataDb],
+      nonceDbs: Vector[OracleNonceDb]): Vector[
+    (OracleAnnouncementV0TLV, Long)] = {
```
nkohen, line 87: should probably replace all of these Long tuples with `Indexed` types

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -0,0 +1,247 @@
...
+      announcementData: Vector[OracleAnnouncementDataDb],
+      nonceDbs: Vector[OracleNonceDb]): ContractInfo = {
+
+    val announcementTLVs =
+      getOracleAnnouncements(announcementIds, announcementData, nonceDbs)
+
+    ContractDescriptor.fromTLV(dlcDb.contractDescriptorTLV) match {
+      case enum: EnumContractDescriptor =>
+        val oracleInfo =
+          if (dlcDb.oracleThreshold == 1 && announcementTLVs.size == 1) {
```
nkohen: I think you don't need to be checking threshold here or below

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -0,0 +1,247 @@
...
+    (
+        DLCDb,
+        DLCOfferDb,
+        DLCAcceptDb,
+        Vector[DLCFundingInputDb],
+        ContractInfo)] = {
+    for {
+      dlcDbOpt <- dlcDAO.findByContractId(contractId)
+      dlcDb = dlcDbOpt.get
+      (dlcDb, dlcOffer, dlcAccept, fundingInputs, contractInfo) <-
```
nkohen: kind of dirty to shadow `dlcDb` like this but so long as it is intentional

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -0,0 +1,263 @@
...
+import org.bitcoins.core.wallet.utxo.AddressTag
+import org.bitcoins.crypto.DoubleSha256DigestBE
+import org.bitcoins.dlc.wallet.DLCWallet
+import org.bitcoins.dlc.wallet.models._
+import org.bitcoins.wallet.internal.TransactionProcessing
+
+import scala.concurrent._
+import scala.util.Try
+
+trait DLCTransactionProcessing extends TransactionProcessing {
```
nkohen: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -0,0 +1,263 @@
...
+  self: DLCWallet =>
+
+  /** Calculates the new state of the DLCDb based on the closing transaction */
+  def calculateAndSetState(dlcDb: DLCDb): Future[DLCDb] = {
```
nkohen: probably should mention the deleting side-effect in the scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+          } yield updated
+        }
+      case (None, None) | (None, Some(_)) | (Some(_), None) =>
+        Future.successful(dlcDb)
+    }
+  }
+
+  def calculateAndSetOutcome(dlcDb: DLCDb): Future[DLCDb] = {
```
nkohen, line 75: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -0,0 +1,263 @@
...
+        } yield res
+      }
+  }
+
+  private def getOutcomeDbInfo(oracleOutcome: OracleOutcome): (
+      Vector[DLCOutcomeType],
+      Vector[SingleOracleInfo]) = {
+    oracleOutcome match {
+      case EnumOracleOutcome(oracles, outcome) =>
+        (Vector(outcome), oracles)
```
nkohen: I'm a bit confused as to why this isn't `Vector.fill(oracles.length)(outcome)` for consistency? Or is it always expected that one matches on something to parse the result? Wouldn't it be easier to just call `oraclesAndOutcomes` (a member of `OracleOutcome`)? https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/OracleOutcome.scala#L28

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAcceptDAO.scala
```diff
@@ -58,54 +57,39 @@ case class DLCAcceptDAO()(implicit
...
   }
 
-  def findByParamHash(paramHash: Sha256Digest): Future[Option[DLCAcceptDb]] =
-    findByParamHash(paramHash.flip)
-
   class DLCAcceptTable(tag: Tag)
-      extends Table[DLCAcceptDb](tag, "wallet_dlc_accepts") {
-
-    def paramHash: Rep[Sha256DigestBE] = column("Param_hash", O.PrimaryKey)
+      extends Table[DLCAcceptDb](tag, schemaName, "accept_dlc_data") {
```
nkohen, line 66: missing negotiation fields

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAnnouncementDAO.scala
```diff
@@ -0,0 +1,118 @@
...
+    createAllNoAutoInc(ts, safeDatabase)
+
+  override protected def findByPrimaryKeys(
+      ids: Vector[(Sha256Digest, Long)]): profile.api.Query[
+    DLCAnnouncementTable,
+    DLCAnnouncementDb,
+    Seq] = {
+
+    // is there a better way to do this?
+    val starting = table.filterNot(_.dlcId === Sha256Digest.empty)
```
nkohen, line 44: ping, I'm a bit confused as to how an empty hash would happen? Is this test related?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAnnouncementDb.scala
```diff
@@ -0,0 +1,10 @@
+package org.bitcoins.dlc.wallet.models
+
+import org.bitcoins.crypto.Sha256Digest
+
+case class DLCAnnouncementDb(
```
nkohen, line 9: probably worth documenting here that this is a mapping table

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDAO.scala
```diff
@@ -3,15 +3,15 @@ package org.bitcoins.dlc.wallet.models
 import org.bitcoins.crypto._
 import org.bitcoins.db.{CRUD, SlickUtil}
 import org.bitcoins.dlc.wallet.DLCAppConfig
-import slick.lifted.{ForeignKeyQuery, PrimaryKey, ProvenShape}
+import slick.lifted._
 
 import scala.concurrent.{ExecutionContext, Future}
 
 case class DLCCETSignatureDAO()(implicit
```
nkohen: should be Signatures (plural) now right?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDAO.scala
```diff
@@ -3,15 +3,15 @@ package org.bitcoins.dlc.wallet.models
...
     val ec: ExecutionContext,
     override val appConfig: DLCAppConfig)
-    extends CRUD[DLCCETSignatureDb, (Sha256DigestBE, ECPublicKey)]
-    with SlickUtil[DLCCETSignatureDb, (Sha256DigestBE, ECPublicKey)] {
+    extends CRUD[DLCCETSignatureDb, (Sha256Digest, Long)]
+    with SlickUtil[DLCCETSignatureDb, (Sha256Digest, Long)] {
```
nkohen: out of curiosity, generally speaking is there a reason indices are `Long` instead of `Int`? Wouldn't we run into errors if we have `Vector`s of larger than `Int.MaxValue` size? (I could be wrong)

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDAO.scala
```diff
@@ -27,75 +27,78 @@ case class DLCCETSignatureDAO()(implicit
...
-    table
-      .filter(_.paramHash.inSet(ids.map(_._1)))
-      .filter(_.sigPoint.inSet(ids.map(_._2)))
+  override protected def findByPrimaryKeys(
+      ids: Vector[(Sha256Digest, Long)]): Query[
+    DLCCETSignatureTable,
+    DLCCETSignatureDb,
+    Seq] = {
+    // is there a better way to do this?
+    val starting = table.filter(_.dlcId =!= Sha256Digest.empty)
```
nkohen: ping

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDAO.scala
```diff
@@ -27,75 +27,78 @@ case class DLCCETSignatureDAO()(implicit
...
 
-    def isInitiator: Rep[Boolean] = column("is_initiator")
+    def index: Rep[Long] = column("index")
 
     def sigPoint: Rep[ECPublicKey] = column("sig_point")
 
-    def signature: Rep[ECAdaptorSignature] = column("signature")
+    def acceptSig: Rep[ECAdaptorSignature] = column("accept_sig")
+
+    def initiatorSig: Rep[Option[ECAdaptorSignature]] = column("initiator_sig")
```
nkohen: I think `signSig` but am open to other options

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDb.scala
```diff
@@ -1,11 +1,11 @@
 package org.bitcoins.dlc.wallet.models
 
-import org.bitcoins.crypto.{ECAdaptorSignature, ECPublicKey, Sha256DigestBE}
+import org.bitcoins.crypto.{ECAdaptorSignature, ECPublicKey, Sha256Digest}
 
 case class DLCCETSignatureDb(
```
nkohen: plural

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDb.scala
```diff
@@ -1,11 +1,11 @@
...
-    paramHash: Sha256DigestBE,
-    isInitiator: Boolean,
+    dlcId: Sha256Digest,
+    index: Long,
     sigPoint: ECPublicKey,
-    signature: ECAdaptorSignature) {
-  def toTuple: (ECPublicKey, ECAdaptorSignature) = (sigPoint, signature)
-}
+    acceptSig: ECAdaptorSignature,
+    initiatorSig: Option[ECAdaptorSignature]
```
nkohen: ping

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCDb.scala
```diff
@@ -1,33 +1,43 @@
...
-  DoubleSha256DigestBE,
-  SchnorrDigitalSignature,
-  Sha256Digest,
-  Sha256DigestBE
-}
+import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
+import org.bitcoins.crypto._
 import scodec.bits.ByteVector
 
 case class DLCDb(
```
nkohen, line 14: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCRefundSigDb.scala
```diff
@@ -1,9 +1,9 @@
 package org.bitcoins.dlc.wallet.models
 
 import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
-import org.bitcoins.crypto.Sha256DigestBE
+import org.bitcoins.crypto.Sha256Digest
 
 case class DLCRefundSigDb(
```
nkohen: plural

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCRefundSigDb.scala
```diff
@@ -1,9 +1,9 @@
...
-    paramHash: Sha256DigestBE,
-    isInitiator: Boolean,
-    refundSig: PartialSignature)
+    dlcId: Sha256Digest,
+    acceptSig: PartialSignature,
+    initiatorSig: Option[PartialSignature])
```
nkohen: name

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/OracleNonceDAO.scala
```diff
@@ -0,0 +1,113 @@
...
+    createAllNoAutoInc(ts, safeDatabase)
+
+  override protected def findByPrimaryKeys(
+      ids: Vector[(Long, Long)]): profile.api.Query[
+    profile.api.Table[OracleNonceDb],
+    OracleNonceDb,
+    Seq] = {
+
+    // is there a better way to do this?
+    val starting = table.filterNot(_.announcementId === -1L)
```
nkohen, line 40: ping

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -23,7 +23,7 @@ import scala.util.{Failure, Success, Try}
   * confirmation counts and marking UTXOs as spent when
   * spending from our wallet
   */
-private[wallet] trait TransactionProcessing extends WalletLogger {
+trait TransactionProcessing extends WalletLogger {
```
nkohen: maybe `private[bitcoins]`?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -41,14 +33,21 @@ import scodec.bits.ByteVector
 
 import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Success
 
-abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
+abstract class DLCWallet
```
nkohen, line 37: scaladoc

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -140,48 +139,60 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
-      Vector[DLCOutcomeType],
-      Vector[SingleOracleInfo]) = {
-    oracleOutcome match {
-      case EnumOracleOutcome(oracles, outcome) =>
-        (Vector(outcome), oracles)
-      case numeric: NumericOracleOutcome =>
-        (numeric.outcomes, numeric.oracles)
-    }
-  }
+  private def updateDLCOracleSigs(
```
nkohen, line 165: should this also delegate to update aggregate sig?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -217,234 +228,30 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+      dlcOpt <- findDLC(dlcId)
       isInit = dlcOpt match {
         case Some(db) =>
           require(db.state == DLCState.Offered || db.state == DLCState.Accepted,
                   "Cannot cancel a DLC after it has been signed")
           db.isInitiator
         case None =>
           throw new IllegalArgumentException(
-            s"No DLC Found with param hash ${paramHash.hex}")
+            s"No DLC Found with param hash ${dlcId.hex}")
```
nkohen: Looking at this now, we might want to be super safe and check the funding sigs table as well?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -901,29 +744,49 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+          offerDb <- dlcOfferDAO.findByDLCId(dlc.dlcId).map(_.get)
           offerInputs <-
-            dlcInputsDAO.findByParamHash(dlc.paramHash, isInitiator = true)
+            dlcInputsDAO.findByDLCId(dlc.dlcId, isInitiator = true)
           prevTxs <-
             transactionDAO.findByTxIdBEs(offerInputs.map(_.outPoint.txIdBE))
+
+          announcements <- dlcAnnouncementDAO.findByDLCId(dlcId)
+          announcementIds = announcements.map(_.announcementId)
+          announcementData <- announcementDAO.findByIds(announcementIds)
```
nkohen: This seems like a common enough pattern for a method

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -938,11 +801,10 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
           Future.failed(new RuntimeException(
             s"No DLC found with corresponding tempContractId ${tempId.hex}, this wallet did not create the corresponding offer"))
       }
-      offerDbOpt <- dlcOfferDAO.findByParamHash(dlcDb.paramHash)
-      // .get should be safe now
-      offerDb = offerDbOpt.get
+      (_, _, _, contractInfo) <- getDLCOfferData(dlcDb.dlcId)
```
nkohen: This can probably be slimmed down?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1245,16 +1048,21 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
   private def verifierFromDb(
```
nkohen: not a must but it might be nice to break up all these `_FromDb` methods into a trait like you did with the other stuff.

nkohen (commented):
Looking good, my main higher level concern is the effect of https://github.com/bitcoin-s/bitcoin-s/pull/2824 as we previously discussed, I think it might be worth proofing the data model for this PR now to avoid a nasty migration but up to your judgement

db-commons/src/main/scala/org/bitcoins/db/DbManagement.scala
benthecarman: oops, part of debuging

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCCETSignatureDAO.scala
benthecarman: All of our `TableAutoInc` stuff uses `Long` so I just followed that

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: That's a good idea, so we unreserve utxos

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,138 @@
-CREATE TABLE "wallet_dlcs"
+CREATE TABLE "global_dlc_data"
```
Christewart, line 1: So it seems that at minimum the contract parameters should be moved out of this table. 

Here is the data model @rorp used for the contract explorer, i think you can crib quite a bit from that?

https://gitlab.com/suredbits/suredbits-api/-/commit/a3e15528a0f56ad454e9fc9d6ad4c0ea70a1e499#14ce9ef498223ef8e08f948a62f7e3af8b7c018d_0_1

Christewart: I guess i think this table should be renamed to `contract_info` or something like that. It seems like that is what this table represents? 

https://github.com/bitcoin-s/bitcoin-s/blob/394e9f773d3a73751c02629ee774337e8b70c508/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala#L1221

If that is the case, we should have this table foreign keys on the sub tables that are used to compose this table (tables that represent `oracleInfo` and `contractDescriptor` in `ContractInfo`). I believe things like `oracle_threshold`, `oracle_params`. 

Something just smells about this table where we have three different domains having data directly inserted in here (not just ids for fk)

1. Oracle information
2. Contract information
3. wallet information (`key_index`, `change_index`, `txid`s etc)

dlc-wallet/src/main/resources/sqlite/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,97 +1,129 @@
-CREATE TABLE "wallet_dlcs"
+CREATE TABLE "global_dlc_data"
```
Christewart, line 1: same comments apply to postgres

Christewart (changes requested):

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
benthecarman: Yeah so we currently split up the oracle info into other tables by having the announcements point back, as for the contract descriptor it is not as easy as the enum vs numeric are very different and there isn't much it can be split up by.

We could move these
```
    "oracle_threshold"      INTEGER NOT NULL, 
    "oracle_params"         TEXT,
    "contract_descriptor"   TEXT    NOT NULL,
    "contract_maturity"     INTEGER NOT NULL,
    "contract_timeout"      INTEGER NOT NULL,
    "total_collateral"      INTEGER NOT NULL, 
    "fee_rate"              TEXT    NOT NULL,
    "fund_output_serial_id" INTEGER NOT NULL,
```

to a separate table, however this will just be another database call we need to do.

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
benthecarman: the offerer's sig is the `initiator_sig`

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
benthecarman: imo `initiator_sig` makes more sense as it references who made it, `sign_sig` can be confusing if you don't know that the initiator does a `DLCSign`

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
Christewart: I think it separates the concerns better IMO. From all the information in that table, can we eventually support specific `enum_contract_descriptor` and `numeric_contract_descriptor` tables that fk back into the table you have laid out in your comment?

benthecarman: If wanted to later do it, would require some complicated migrations, I don't think we would need that though, the descriptors are pretty compact and we don't really have a reason to query for individualized pieces.

Christewart: Ok that is fine with me

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -217,234 +228,30 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
       dbs <- spendingInfoDAO.findByOutPoints(inputs.map(_.outPoint))
       _ <- unmarkUTXOsAsReserved(dbs)
 
-      _ <- dlcSigsDAO.deleteByParamHash(paramHash)
-      _ <- dlcRefundSigDAO.deleteByParamHash(paramHash)
-      _ <- dlcInputsDAO.deleteByParamHash(paramHash)
-      _ <- dlcAcceptDAO.deleteByParamHash(paramHash)
-      _ <- dlcOfferDAO.deleteByParamHash(paramHash)
-      _ <- dlcDAO.deleteByParamHash(paramHash)
+      _ <- dlcSigsDAO.deleteByDLCId(dlcId)
```
Christewart, line 254: This is an example of something that could be done better with a database transaction, and we get consistency and a performance increase (1 db call instead of 7)! 

In this example, if you made all of these `dleteByDLCId()` methods return a [`DBIOAction`](https://scala-slick.org/doc/3.3.3/dbio.html#database-io-actions), we could then just submit them to the database for running by using [`DBIO.seq()`](https://scala-slick.org/doc/3.3.3/dbio.html#sequential-execution)

rorp has some good examples of this on the closed source side

https://gitlab.com/suredbits/suredbits-api/-/commit/a3e15528a0f56ad454e9fc9d6ad4c0ea70a1e499#80761a3cdb2a923a3fc46e862f90f2fd928494ea_0_75

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -610,32 +408,65 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
-              outcomesOpt = None,
-              oraclesUsedOpt = None
+              aggregateSignatureOpt = None
             )
           }
           _ <- writeDLCKeysToAddressDb(account, nextIndex)
           writtenDLC <- dlcDAO.create(dlc)
-          _ <- announcementDAO.upsertAll(announcementDbs)
+
+          announcementDataDbs <- announcementDAO.createAll(announcementDataDbs)
```
Christewart, line 474: if i'm not mistaken this same pattern is repeated here: https://github.com/bitcoin-s/bitcoin-s/pull/3100/files#diff-a5db8d33c02ab2b84726b8d0da0ea171ca50097bd2e50982b0f0736be02e3db2R281

Let's just factor the code into a helper method?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1128,12 +991,24 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+              db.copy(initiatorSig = Some(sig))
+            }
+
           _ = logger.info(
             s"CET Signatures are valid for contract ${sign.contractId.toHex}")
 
           _ <- addFundingSigs(sign)
-          _ <- dlcSigsDAO.createAll(sigsDbs)
-          _ <- dlcRefundSigDAO.create(refundSigDb)
+          _ <- dlcSigsDAO.updateAll(updatedSigDbs)
```
Christewart: nit: things like this might be nice to have in a transaction in the future

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
nkohen: In that case I think we should change `accept_sig` to `accepter_sig` just to be consistent

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
benthecarman: The ids won't be equilivant to the actual index

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAnnouncementDAO.scala
benthecarman: This is weird, I agree. I couldn't put just `table` for the init in the `foldLeft` for some reason so I instead just did a filter that shouldn't filter anything.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: I don't think so, the aggregate sig is DLC specific but the oracle signatures are not because an announcement can be used in multiple different DLCs

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: Agreed, this is planned for a follow up

Christewart: Can we file an issue so this doesn't get forgotten about then?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -559,33 +354,50 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
...
+        contractDescriptorTLV = contractInfo.contractDescriptor.toTLV,
+        contractMaturity = timeouts.contractMaturity,
+        contractTimeout = timeouts.contractTimeout,
+        totalCollateral = contractInfo.totalCollateral,
+        feeRate = feeRate,
+        fundOutputSerialId = fundOutputSerialId
+      )
 
-      dlcOfferDb = DLCOfferDbHelper.fromDLCOffer(offer)
+      _ <- dlcDAO.create(dlcDb)
```
Christewart, line 387: for the future db transaction issue, this stuff looks like it should be included there

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -0,0 +1,275 @@
...
+import org.bitcoins.dlc.wallet.models._
+import org.bitcoins.wallet.internal.TransactionProcessing
+
+import scala.concurrent._
+import scala.util.Try
+
+/** Overrides TransactionProcessing from Wallet to add extra logic to
+  * process transactions that could from our own DLC.
+  */
+private[bitcoins] trait DLCTransactionProcessing extends TransactionProcessing {
```
Christewart, line 22: Yikes, in general we need to stop using inheritance in the wallet and start using `has-a` rather than `is-a`. This isn't going to get changed in this PR, but needs to be taken care of eventually.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAnnouncementDAO.scala
Christewart: I think it can just be implemented like this ? 

```scala
    val dlcIds = ids.map(_._1)
    val announcementId = ids.map(_._2)
    table.filter(t =>
      t.dlcId.inSet(dlcIds) && t.announcementId.inSet(announcementId))
```

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCAnnouncementDAO.scala
```diff
@@ -7,11 +7,15 @@ import slick.lifted._
 
 import scala.concurrent.{ExecutionContext, Future}
 
+private case class DLCAnnouncementPrimaryKey(
```
Christewart: I don't think these will be able to be private, i don't think there is a reason to keep them private. I believe this is causing the compiler errors

https://github.com/bitcoin-s/bitcoin-s/pull/3100/checks?check_run_id=2633397282#step:5:133

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -814,7 +679,7 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
       _ <- dlcOfferDAO.create(dlcOfferDb)
       _ <- dlcAcceptDAO.create(dlcAcceptDb)
       _ <- dlcSigsDAO.createAll(sigsDbs)
-      _ <- dlcRefundSigDAO.create(refundSigDb)
+      _ <- dlcRefundSigDAO.create(refundSigsDb)
```
Christewart, line 682: adding comment here for future db transaction issue

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -896,34 +762,53 @@ abstract class DLCWallet extends Wallet with AnyDLCHDWalletApi {
           _ <- remoteTxDAO.createAll(acceptPrevTxs)
```
Christewart, line 762: noting this  for future db transaction issue

Christewart (approved):
I'm fine with merging this under the assumption that issues are filed for the following 

1. Db transactions
2. Breaking up the `DLCWallet` is-a relationship
3. Figure out how to fix all the `.filter` usages in things like `findByPrimaryKeys()`

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,144 @@
...
-    "announcement" TEXT NOT NULL PRIMARY KEY,
-    "pub_key"      TEXT NOT NULL
+    "dlc_id"                TEXT PRIMARY KEY,
+    "oracle_threshold"      INTEGER NOT NULL,
+    "oracle_params"         TEXT,
+    "contract_descriptor"   TEXT    NOT NULL,
+    "contract_maturity"     INTEGER NOT NULL,
+    "contract_timeout"      INTEGER NOT NULL,
+    "total_collateral"      INTEGER NOT NULL,
+    "fee_rate"              TEXT    NOT NULL,
```
nkohen: nit: fee_rate seems like global data and not contract data imo (seeing as we could in the future have an rbf workflow like dual funded channels)

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,108 +1,144 @@
...
+    "fund_output_serial_id" INTEGER NOT NULL,
```
nkohen: `fund_output_serial_id` is also border-line for me on contract vs. global

nkohen (commented):
Looks good, just one thing

nkohen (approved):
ACK assuming CI

PR #3110, opened 2021-05-20
DLC Adaptor Point Computation Memoization

nkohen:
Introduced memoization into the adaptor point computation for non-libsecp CryptoContexts. This results in ~6x speedup in adaptor point computation (but in the secp CryptoContext it actually leads to a slow-down due to JNI overhead as there is less batching in this approach as memoized computation only does one step at a time).

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
@@ -14,7 +14,10 @@ import org.bitcoins.crypto.{
   CryptoUtil,
   ECPublicKey,
   FieldElement,
-  SchnorrPublicKey
+  SchnorrPublicKey,
+  SecpPoint,
+  SecpPointFinite,
+  SecpPointInfinity
 }
 import scodec.bits.ByteVector
 
@@ -50,6 +53,44 @@ object DLCAdaptorPointComputer {
     nonce.add(pubKey.publicKey.tweakMultiply(FieldElement(hash)))
   }
 
+  case class AdditionTrieNode(
+      preComputeTable: Vector[Vector[ECPublicKey]], // Nonce -> Outcome -> Point
+      depth: Int = 0,
+      var children: Vector[AdditionTrieNode] = Vector.empty,
+      var pointOpt: Option[SecpPoint] = None) {
+
+    def initChildren(): Unit = {
+      children = 0
+        .until(base)
+        .toVector
+        .map(_ => AdditionTrieNode(preComputeTable, depth + 1))
+    }
+
+    def computeSum(digits: Vector[Int]): ECPublicKey = {
+      val point = pointOpt.get
+
+      if (digits.isEmpty) {
+        point match {
+          case SecpPointInfinity =>
+            throw new IllegalArgumentException(
+              "Sum cannot be point at infinity.")
+          case point: SecpPointFinite => point.toPublicKey
+        }
+      } else {
+        val digit = digits.head
+        if (children.isEmpty) initChildren()
+        val child = children(digit)
+        child.pointOpt match {
+          case Some(_) => child.computeSum(digits.tail)
+          case None =>
+            val pointToAdd = preComputeTable(depth)(digit).toPoint
+            child.pointOpt = Some(point.add(pointToAdd))
+            child.computeSum(digits.tail)
+        }
+      }
+    }
+  }
+
   /** Efficiently computes all adaptor points, in order, for a given ContractInfo.
     * @see https://medium.com/crypto-garage/optimizing-numeric-outcome-dlc-creation-6d6091ac0e47
     */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
```diff
@@ -11,10 +11,14 @@ import org.bitcoins.core.protocol.tlv.{
...
+  CryptoContext,
   CryptoUtil,
   ECPublicKey,
   FieldElement,
-  SchnorrPublicKey
+  SchnorrPublicKey,
+  SecpPoint,
+  SecpPointFinite,
+  SecpPointInfinity
 }
```
benthecarman: nit: `import org.bitcoins.crypto._`

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
```diff
@@ -50,6 +54,44 @@ object DLCAdaptorPointComputer {
     nonce.add(pubKey.publicKey.tweakMultiply(FieldElement(hash)))
   }
 
+  case class AdditionTrieNode(
```
benthecarman, line 60: scaladoc

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
```diff
@@ -77,33 +119,59 @@ object DLCAdaptorPointComputer {
...
+              contractInfo.oracleInfo.singleOracleInfos.indexOf(info)
+            val outcomeIndices = outcome match {
+              case outcome: EnumOutcome =>
+                Vector(
+                  contractInfo.contractDescriptor
+                    .asInstanceOf[EnumContractDescriptor]
+                    .keys
+                    .indexOf(outcome)
+                )
+              case UnsignedNumericOutcome(digits) => digits
```
benthecarman, line 169: this doesn't seem correct, should we be getting the index here, not the actual digit?

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
```diff
@@ -50,6 +54,44 @@ object DLCAdaptorPointComputer {
...
+      preComputeTable: Vector[Vector[ECPublicKey]], // Nonce -> Outcome -> Point
+      depth: Int = 0,
+      var children: Vector[AdditionTrieNode] = Vector.empty,
+      var pointOpt: Option[SecpPoint] = None) {
+
+    def initChildren(): Unit = {
+      children = 0
+        .until(base)
+        .toVector
+        .map(_ => AdditionTrieNode(preComputeTable, depth + 1))
```
benthecarman, line 75: should this call `initChildren` on the children as well?

benthecarman: oh I see, it is done in `computeSum`, maybe add a comment about this

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
nkohen: The digit is the index of the outcome (0 is 0, 1 is 1, etc.)

benthecarman: ah duh

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
```diff
@@ -50,6 +45,70 @@ object DLCAdaptorPointComputer {
...
+    * This trie stores all already computed sub-sums and new points are computed
+    * by extending this Trie.
+    *
+    * Note that this method should not be used if you have access to LibSecp256k1CryptoRuntime
+    * because calling CryptoUtil.combinePubKeys will outperform memoization in that case.
+    */
+  case class AdditionTrieNode(
+      preComputeTable: Vector[Vector[ECPublicKey]], // Nonce -> Outcome -> Point
+      depth: Int = 0,
+      var children: Vector[AdditionTrieNode] = Vector.empty,
```
Christewart: aren't these global mutable state?

Christewart:
```scala
  private case class AdditionTrieNode(
      preComputeTable: Vector[Vector[ECPublicKey]], // Nonce -> Outcome -> Point
      depth: Int = 0,
      private var children: Vector[AdditionTrieNode] = Vector.empty,
      private var pointOpt: Option[SecpPoint] = None) {

```

Christewart: Is this a binary trie? If so can we add the invariant `require(children.length < 3)`

Christewart (changes requested):
It looks ok to me, but it would be _really_ nice to have unit tests to make sure this is correct and detect regressions in the future

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCAdaptorPointComputer.scala
nkohen: Good point on the `private`s :P Don't know how I missed that (fixed).

Re binary trie: the current implementation of the Trie is base agnostic so that it works with any base. It still has the invariant however that `children.length <= base` but this can't actually be enforced by a `require` since `children` is mutable and always starts out as `Vector.empty` so this `require` would trivially pass.

Christewart (approved):

PR #3114, opened 2021-05-20
Remove need for bitcoind install with remote

benthecarman:
This was missed in pulling down from the dlc branches

https://github.com/bitcoin-s/bitcoin-s/pull/2940


nkohen (approved):
Worked locally for me :+1:

PR #3116, opened 2021-05-20
Removed point multiplication from ECPrivateKey.freshPrivateKey

nkohen:
This cuts `coreTestJVM/test` and `cryptoTestJVM/test` times in half!!!

### DIFF
--- a/crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
+++ b/crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
@@ -18,7 +18,9 @@ trait BouncycastleCryptoRuntime extends CryptoRuntime {
 
   override val cryptoContext: CryptoContext = CryptoContext.BouncyCastle
 
-  /** Cribbed from ECKeyPairGenerator::generateKeyPair */
+  /** Cribbed from ECKeyPairGenerator::generateKeyPair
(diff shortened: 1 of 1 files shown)


rorp (approved):

crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala
```diff
@@ -23,16 +18,23 @@ trait BouncycastleCryptoRuntime extends CryptoRuntime {
...
+    val n = CryptoParams.getN
+    val bitLength = n.bitLength()
+    val minWeight = bitLength >>> 2
+
+    var priv: BigInteger = BigInteger.ZERO
+    var foundNum: Boolean = false
+    while (!foundNum) {
+      priv = BigIntegers.createRandomBigInteger(bitLength, secureRandom)
+
+      if (
```
Christewart: This is easier to read and audit for correctness IMO, it is basically as close to a Scala translation as possible from the java source code IMO

```scala
      priv = BigIntegers.createRandomBigInteger(bitLength, secureRandom)

      if (priv.compareTo(BigInteger.ONE) < 0 || (priv.compareTo(n) >= 0)) {
        //do nothing, keep iterating
      } else if (WNafUtil.getNafWeight(priv) < minWeight) {
        //do nothing, keep iterating
      } else {
        foundNum = true
      }
```

Christewart: Also a link to the implementation we are replicating: https://github.com/bcgit/bc-java/blob/63b18eb973f5731e403f655ee81d6b8456f5b256/core/src/main/java/org/bouncycastle/crypto/generators/ECKeyPairGenerator.java#L39

Christewart (approved):

PR #3122, opened 2021-05-21
Remove calls to Future.apply which apparently runs only on one thread

nkohen:
Created `FutureUtil.makeAsync` function that replaces calls to `Future(_)` and actually executes in a multi-threaded way.

First commit introduces the function and implements it in `DLCTxSigner` leading to a HUGE optimization for DLCs. If people aren't a fan of the larger PR, we should at least merge the first commit.

EDIT: Reverted to just the first commit due to weird interactions on some of the other changes.


nkohen:
@Christewart to test this out for yourself, I've pushed adaptor-dlc cherry-picked on top of this here:
https://github.com/nkohen/bitcoin-s-core/tree/temp2

nkohen:
Before:
Screen Shot 2021-05-21 at 1 59 27 PM

After:
Screen Shot 2021-05-21 at 2 32 56 PM

benthecarman (approved):

Christewart (approved):

Christewart (approved):

PR #3126, opened 2021-05-22
Add FutureUtil tests

Christewart:
This adds sanity test cases for both `FutureUtil.batchAndSyncExecute()` and `FutureUtil.batchAndParallelExecute()`. This is to confirm the expected behavior of the implementation. 

This is partially in response to #3122 . It doesn't seem to be a problem with our impelentation of these two methods, rather we (I in particular) are misunderstanding the expected behavior of `Future.apply` as @nkohen mentions.

To make this stuff work on scalajs, I had to add a scalajs compatible scheduler. The only project I found to do this was [monix](https://monix.io/). Specifically their [monix-execution](https://monix.io/docs/current/#monix-execution) dependency.

Screenshot from 2021-06-28 13-20-27


benthecarman:
test failed on secp disabled CI

https://github.com/bitcoin-s/bitcoin-s/pull/3126/checks?check_run_id=2655900960#step:5:2021

Christewart:
This now introduces a new scheduler that is compatible with scalajs. It uses a library from the monix project, you can find documentation here: https://monix.io/docs/current/execution/scheduler.html

benthecarman (approved):
ACK on CI

PR #3136, opened 2021-05-25
Only sync filters since wallet creation time

### LINKED ISSUE #1731: Only download and store filters starting from wallet creation time
Currently, we download all filters for the chain. This however, is unneeded as we should not care about filters that we know will have no relevance to us. We can instead just download filters after our wallet creation time.

Christewart:
fixes #1731 

The current bottleneck with IBD is compact filter syncing. This PR adds logic to only sync compact filters since the wallet was created on mainnet. Testnet, regtest, and signet do not have this functionality enabled. 

I'm still testing this out locally, but wanted to get it opened up.

### DIFF
--- a/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
+++ b/chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
@@ -7,6 +7,7 @@ import org.bitcoins.chain.pow.Pow
 import org.bitcoins.core.api.chain.ChainQueryApi.FilterResponse
 import org.bitcoins.core.api.chain.db._
 import org.bitcoins.core.api.chain.{ChainApi, FilterSyncMarker}
+import org.bitcoins.core.config.{MainNet, RegTest, SigNet, TestNet3}
 import org.bitcoins.core.gcs.FilterHeader
 import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.p2p.CompactFilterMessage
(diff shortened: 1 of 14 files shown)


nkohen (commented):
Docs need fixing but otherwise ack

benthecarman:
My concern with this is for rescanning, it will not work if trying to rescan older blocks

chain/src/main/scala/org/bitcoins/chain/blockchain/ChainHandler.scala
```diff
@@ -327,6 +329,59 @@ class ChainHandler(
...
+      batchSize: Int,
+      forceSyncFilters: Boolean): Future[Option[FilterSyncMarker]] = {
+    chainConfig.network match {
+      case network @ (TestNet3 | RegTest | SigNet) =>
+        //not supported on other networks, just start from 0
+        logger.warn(
+          s" Ignoring wallet creation timestamp when syncing filters on $network. Syncing from genesis")
+        nextFilterHeaderBatchRange(0, batchSize)
+      case MainNet =>
+        //some crude rounding to optimize syncing filters as they
```
benthecarman, line 344: instead of all this you can use `chainApi.epochSecondToBlockHeight`

Christewart:
> My concern with this is for rescanning, it will not work if trying to rescan older blocks

This is fair. I will be testing rescans this afternoon. I'm hopeful that #3137 will give us enough of a performance boost but :crossed_fingers:

Christewart:
Superseded by #4109

PR #3137, opened 2021-05-25
parallelize matching compact filters as this is a bottleneck during IBD

Christewart:
This actually is a bottleneck rather than I/O. It is computationally expensive to match these filters, and when it is single threaded it takes even longer.

Here is an example on visualvm with filter height ~660,000 on mainnet with sqlite

Screenshot from 2021-05-25 09-20-35

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -219,9 +219,10 @@ abstract class Wallet
   private def searchFilterMatches(spks: Vector[ScriptPubKey])(
       blockFilters: Vector[(DoubleSha256Digest, GolombFilter)]): Future[
     Vector[DoubleSha256Digest]] = FutureUtil.makeAsync { () =>
+    val asmVec = spks.map(_.asmBytes)
     blockFilters.flatMap { case (blockHash, blockFilter) =>
       val matcher = SimpleFilterMatcher(blockFilter)
-      if (matcher.matchesAny(spks.map(_.asmBytes))) {
+      if (matcher.matchesAny(asmVec)) {
         Vector(blockHash)
       } else {
         Vector.empty
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -218,7 +218,7 @@ abstract class Wallet
 
   private def searchFilterMatches(spks: Vector[ScriptPubKey])(
       blockFilters: Vector[(DoubleSha256Digest, GolombFilter)]): Future[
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -216,6 +216,19 @@ abstract class Wallet
     }
   }
 
+  private def searchFilterMatches(spks: Vector[ScriptPubKey])(
+      blockFilters: Vector[(DoubleSha256Digest, GolombFilter)]): Future[
+    Vector[DoubleSha256Digest]] = Future {
```
nkohen: nit: this kind of feels like it should be an Option instead of a Vector. If you really want to use flatten above you can still call `toVector` on the returned Options

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -184,21 +184,21 @@ abstract class Wallet
...
         } else {
-          FutureUtil.sequentially(blockFilters) {
-            case (blockHash, blockFilter) =>
-              val matcher = SimpleFilterMatcher(blockFilter)
-              if (matcher.matchesAny(scriptPubKeys.toVector.map(_.asmBytes))) {
-                nodeApi.downloadBlocks(Vector(blockHash))
-              } else Future.unit
-          }
+          FutureUtil
+            .batchAndParallelExecute(
```
nkohen, line 194: did you test this in visualvm and see an improvement? Just worried about the weird behavior we saw previously and wondering if you might want to use `makeAsync`

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
Christewart: It would have to be an `Option[Vector[DoubleSha256Digest]]` then, i guess i generally prefer just doing a `Vector`. Note the `blockFilters.flatMap` call below.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
Christewart: Yeah I share the same concerns. I'm doing so now. Waiting for a block to come in on mainnet so i can do an IBD with this code :+1:

Christewart: > did you test this in visualvm and see an improvement? Just worried about the weird behavior we saw previously and wondering if you might want to use `makeAsync`

You were absolutely right here. Here is what it looks like with `FutureUtil.makeAsync()` :tada: 

Screenshot from 2021-05-25 11-19-57

nkohen (approved):
ACK assuming CI

benthecarman (approved):

PR #3142, opened 2021-05-26
Add startup config flow for bundle

benthecarman:
Splits up the gui so it can be constructed from the bundle. This then adds a startup screen so you can set bitcoind or neutrino config for starting up.

When setting the config in the gui it will add it to your current bitcoin-s.conf file and write it, this however is not perfect and will not preserve things like comments and ordering, I am unsure on how to do this.

There are other things we may want to add, such as wallet password.

image

image


app/bundle/src/main/scala/org/bitcoins/bundle/gui/BitcoindConfigPane.scala
```diff
@@ -0,0 +1,98 @@
...
+    add(new Label("RPC Port"), 0, nextRow)
+    add(portTF, 1, nextRow)
+    nextRow += 1
+
+    add(new Label("RPC Username"), 0, nextRow)
+    add(rpcUserTF, 1, nextRow)
+    nextRow += 1
+
+    add(new Label("RPC Password"), 0, nextRow)
+    add(rpcPasswordTF, 1, nextRow)
```
benthecarman, line 77: Should this be a password field instead of a text field?

Christewart:
> When setting the config in the gui it will add it to your current bitcoin-s.conf file and write it, this however is not perfect and will not preserve things like comments and ordering, I am unsure on how to do this.

In my opinion these could just be set as system properties. If you look at the typesafe config resolution precedence, system properties take precedence over any configuration.

Maybe the important question is how to restart and persist the config?

Christewart:
So this does seem to work for me :tada: but I really dislike overwriting an existing users `bitcoin-s.conf` file. Does bitcoin core allow config settings to be set from the GUI? If so, do they re-write the entire `bitcoin.conf` file?


IMO, the best course of action would be to make  second config file called `bitcoin-s-bundle.conf` and write the settings to that. 

Then when starting the `bundle`, make sure the `bitcoin-s-bundle.conf` takes higher priority than the `bitcoin-s.conf`

benthecarman:
> So this does seem to work for me  but I really dislike overwriting an existing users `bitcoin-s.conf` file. Does bitcoin core allow config settings to be set from the GUI? If so, do they re-write the entire `bitcoin.conf` file?
> 
> IMO, the best course of action would be to make second config file called `bitcoin-s-bundle.conf` and write the settings to that.
> 
> Then when starting the `bundle`, make sure the `bitcoin-s-bundle.conf` takes higher priority than the `bitcoin-s.conf`

I believe bitcoin core does something similiar to this, I like this idea, will attempt.

benthecarman:
Changed to use a 2nd config file and added explainers at the top of each page

image
image

Christewart (approved):
Seems to work for me, and i think it was a great idea to add more explanation to the top of the screens

nkohen (approved):
Haven't looked at the code but worked like a charm for me when I ran it

PR #3145, opened 2021-05-26
Make sure secp256k1 is published for java8

Christewart:
Previously it would just be published for the jdk it was built on. Now it will be published for java8 so machines that are running java8 can use it.

### DIFF
--- a/project/CommonSettings.scala
+++ b/project/CommonSettings.scala
@@ -3,7 +3,15 @@ import com.typesafe.sbt.SbtNativePackager.Docker
 import com.typesafe.sbt.SbtNativePackager.autoImport.packageName
 
 import java.nio.file.Paths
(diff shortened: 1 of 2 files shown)


project/CommonSettings.scala
```diff
@@ -3,7 +3,15 @@ import com.typesafe.sbt.SbtNativePackager.Docker
 import com.typesafe.sbt.SbtNativePackager.autoImport.packageName
 
 import java.nio.file.Paths
-import com.typesafe.sbt.packager.Keys.{daemonUser, daemonUserUid, dockerAlias, dockerAliases, dockerRepository, dockerUpdateLatest, maintainer}
+import com.typesafe.sbt.packager.Keys.{
```
Christewart, line 6: formatter, not sure why these got changed

PR #3151, opened 2021-05-26
Split up `DLCCETSignaturesDb` by `isInitiator`

### LINKED ISSUE #3149: Add Sigs on DLCWallet taking a long time and using absurd amount of memory
I was setting up a DLC with just over 2K CETs (accept took under 2 seconds) but signing and adding sigs both took long amounts of time, timed out, and in the case of adding sigs I got
```
[error] Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "bitcoind-rpc-client-created-by-bitcoin-s-scheduler-1"
[info] #
[info] # A fatal error has been detected by the Java Runtime Environment:
[info] #
[info] #  SIGSEGV (0xb) at pc=0x0000000105671918, pid=67958, tid=104719
[info] #
[info] # JRE version: OpenJDK Runtime Environment (11.0.10+9) (build 11.0.10+9)
[info] # Java VM: OpenJDK 64-Bit Server VM (11.0.10+9, mixed mode, tiered, compressed oops, g1 gc, bsd-aarch64)
[info] # Problematic frame:
[info] # [thread 86019 also had an error]
[info] V  [libjvm.dylib+0x671918]  _ZN14AccessInternal19PostRuntimeDispatchIN12G1BarrierSet13AccessBarrierILy1097844ES1_EELNS_11BarrierTypeE2ELy1097844EE18oop_access_barrierEPv+0x8
[info] #
[info] # No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
[info] #
[info] # An error report file with more information is saved as:
[info] # /Users/nkohen/dev/bitcoin-s/app/bundle/hs_err_pid67958.log
[info] #
[info] # If you would like to submit a bug report, please visit:
[info] #   https://bugreport.java.com/bugreport/crash.jsp
[info] #
[error] Nonzero exit code returned from runner: 134
```
This crashed my program but on starting it back up I was able to broadcast (meaning add sigs had completed its important operations before crashing).

benthecarman:
Fixes #3149

This changes the `DLCCETSignaturesDb` to instead of having both parties signature in reach row, to instead have a `isInitiator` column to signify who's signature it belongs to. This is done for performance reasons, previously we would need all of both parties signatures to do any task, where here we can just grab the ones we need, this also allows us to use `createAll` instead of `updateAll` when doing `addDLCSigs` which is much more performant.


Christewart:
CI failure: https://github.com/bitcoin-s/bitcoin-s/pull/3151/checks?check_run_id=2678259460#step:5:838

benthecarman:
After discussing with @nkohen, we decided to go with a different fix #3154

PR #3154, opened 2021-05-26
Removed extraneous findAll call from CRUD.updateAll

nkohen:


benthecarman (approved):

db-commons/src/main/scala/org/bitcoins/db/CRUD.scala
```diff
@@ -90,9 +90,15 @@ abstract class CRUD[T, PrimaryKeyType](implicit
     } else {
       val actions = ts.map(t => find(t).update(t))
       for {
-        _ <- safeDatabase.runVec(DBIO.sequence(actions).transactionally)
-        result <- safeDatabase.runVec(findAll(ts).result)
-      } yield result
+        numUpdated <- safeDatabase.runVec(
+          DBIO.sequence(actions).transactionally)
+        tsUpdated <- {
+          if (numUpdated.sum == ts.length) Future.successful(ts)
```
Christewart, line 96: Will this be correct if there is an element in the vector that is up to date in the database? For instance, if we want to update 3 items, but the 2nd item is in sync with the database, will it return 2 items or 3 items?

Christewart (approved):

db-commons/src/main/scala/org/bitcoins/db/CRUD.scala
nkohen: 3!

PR #3155, opened 2021-05-26
2021 05 26 uint64 mapper

Christewart:
Simply adds unit tests for db column mappers for `UInt64`. This does not fix any bugs, but might detect issues on CI in the future.

### DIFF
--- a/db-commons-test/src/test/scala/org/bitcoins/db/UInt64MapperTest.scala
+++ b/db-commons-test/src/test/scala/org/bitcoins/db/UInt64MapperTest.scala
@@ -1,0 +1,36 @@
+package org.bitcoins.db
+
+import org.bitcoins.core.number.UInt64
+import org.bitcoins.testkitcore.gen.NumberGenerator
+import org.bitcoins.testkitcore.util.BitcoinSUnitTest
+import slick.jdbc.{PostgresProfile, SQLiteProfile}
+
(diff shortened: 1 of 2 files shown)


benthecarman (approved):

benthecarman:
I ran the 3rd test with a million generated uint64s and it haven't gotten a failure :/

benthecarman:
Finally got a good error message by having it printed in the mapper and running the DLC tests forever.

```
        Try(UInt64.fromHex(str)) match {
          case Failure(err) =>
            println(str)
            err.printStackTrace()
            throw err
          case Success(value) => value
        }
```

Looks like the Uint64 mapper will sometimes output `Inf` instead of hex?

```
Inf
java.lang.IllegalArgumentException: Invalid hexadecimal character 'I' at index 0
	at scodec.bits.ByteVector$.$anonfun$fromValidHex$1(ByteVector.scala:1816)
	at scala.util.Either.fold(Either.scala:190)
	at scodec.bits.ByteVector$.fromValidHex(ByteVector.scala:1816)
	at org.bitcoins.crypto.CryptoBytesUtil.decodeHex(CryptoBytesUtil.scala:12)
	at org.bitcoins.crypto.CryptoBytesUtil.decodeHex$(CryptoBytesUtil.scala:11)
```

PR #3157, opened 2021-05-26
Use groupByExistingAnnouncements when accepting

benthecarman:
This is so you can do multiple DLCs based on the same announcement

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -440,8 +440,10 @@ abstract class DLCWallet
         val announcements =
           offer.contractInfo.oracleInfo.singleOracleInfos.map(_.announcement)
 
-        val announcementDataDbs =
-          OracleAnnouncementDbHelper.fromAnnouncements(announcements)
+        //filter announcements that we already have in the db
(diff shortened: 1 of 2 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -301,7 +301,7 @@ abstract class DLCWallet
     for {
       feeRate <- feeRateF
       groupedAnnouncements <- groupedAnnouncementsF
-      announcementDataDbs <- announcementDAO.createAll(
+      announcementDataDbs <- announcementDAO.upsertAll(
```
Christewart: can you add a unit test like this to make sure we don't regress in the future? 

https://github.com/bitcoin-s/bitcoin-s/pull/3140/commits/0da17af6f368a678826a4b96841f8e458574d20e#diff-880bebd9e81351c77729c8d406bb5f0fa6b05759c7dccfaec2e3ea845683dbecR360

Christewart (changes requested):

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: done

Christewart (approved):

PR #3159, opened 2021-05-26
Don't fetch all DLC data when canceling DLC

benthecarman:
This is useful in the case where we would have a corrupted DLC and couldn't cancel because `findDLC` would try to build the `DLCStatus`


nkohen (approved):
Worked for me

Christewart (approved):
Worked for me. In the future we should offer an option to double spend the utxo to make it completely safe. Issue filed #3169

PR #3161, opened 2021-05-26
Fix main class in bundle

Christewart:
The mainclass got changed in #3142 but didn't get changed in release.yml


app/bundle/bundle.sbt
```diff
@@ -1,6 +1,6 @@
 name := "bitcoin-s-bundle"
 
-mainClass := Some("org.bitcoins.bundle.BundleGUI")
+mainClass := Some("org.bitcoins.bundle.gui.BundleGUI")
```
benthecarman, line 4: need to fix on line 10 of this file too

benthecarman (approved):

PR #3168, opened 2021-05-27
Make loading icon appear until server is full started

### LINKED ISSUE #3165: it takes 10 second to switch to start app server on new startup gui
When selecting the neutrino configuration on the new startup page, it takes 10-15 seconds for it to actually change. Even worse the "loading" icon isn't shown for most of the period. That makes it seem like something is broken.

To reproduce you just need to select the netreuno config on the startup menu on 6c9d8216

Screen Shot 2021-05-26 at 7 18 20 PM

Screen Shot 2021-05-26 at 7 18 39 PM

benthecarman:
Fixes #3165

To make it any faster we need to optimize `startBitcoinSBackend`

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
@@ -10,6 +10,8 @@ import scalafx.stage.Window
 
 import java.nio.file.Files
 import scala.concurrent.ExecutionContext.global
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, Promise}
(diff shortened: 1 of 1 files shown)


Christewart (changes requested):
So this is less of a problem on the bitcoind backend as the daemon is already running, but there still is a noticeable "wtf, loading screen done but nothing happened" on that startup as well. 

Can you apply the same technique there too?

benthecarman:
> So this is less of a problem on the bitcoind backend as the daemon is already running, but there still is a noticeable "wtf, loading screen done but nothing happened" on that startup as well. 
> 
> 
> 
> Can you apply the same technique there too?

This change applies to both

Christewart (approved):

PR #3138, opened 2021-05-25
DLC Wallet pulldown

benthecarman:

PR #3170, opened 2021-05-27
Fix serialId columns to use a String instead of Integer

benthecarman:
Fixes that annoying error where we get `java.lang.IllegalArgumentException: Invalid hexadecimal character 'I' at index 0` from the `UInt64` Mapper.

This was occurring because the uint64 mapper uses a string whereas the db columns were using `INTEGER` so sometimes sqlite would store `Inf` instead of the hex.

### DIFF
--- a/dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
+++ b/dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
@@ -1,33 +1,33 @@
 CREATE TABLE "global_dlc_data"
 (
-    "dlc_id"              TEXT PRIMARY KEY,
-    "temp_contract_id"    TEXT    NOT NULL UNIQUE,
-    "contract_id"         TEXT UNIQUE,
-    "protocol_version"    INTEGER NOT NULL,
-    "state"               TEXT    NOT NULL,
-    "is_initiator"        INTEGER NOT NULL,
-    "account"             TEXT    NOT NULL,
-    "change_index"        INTEGER NOT NULL,
-    "key_index"           INTEGER NOT NULL,
+    "dlc_id"                TEXT PRIMARY KEY,
+    "temp_contract_id"      TEXT    NOT NULL UNIQUE,
+    "contract_id"           TEXT UNIQUE,
+    "protocol_version"      INTEGER NOT NULL,
+    "state"                 TEXT    NOT NULL,
+    "is_initiator"          INTEGER NOT NULL,
+    "account"               TEXT    NOT NULL,
+    "change_index"          INTEGER NOT NULL,
+    "key_index"             INTEGER NOT NULL,
 
-    "fee_rate"              VARCHAR(254) NOT NULL,
-    "fund_output_serial_id" INTEGER      NOT NULL,
+    "fee_rate"              TEXT    NOT NULL,
+    "fund_output_serial_id" TEXT    NOT NULL,
 
-    "funding_outpoint"    TEXT,
-    "funding_tx_id"       TEXT,
-    "closing_tx_id"       TEXT,
-    "aggregate_signature" TEXT
+    "funding_outpoint"      TEXT,
+    "funding_tx_id"         TEXT,
+    "closing_tx_id"         TEXT,
+    "aggregate_signature"   TEXT
 );
 
 CREATE TABLE "contract_data"
 (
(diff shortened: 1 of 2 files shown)


Christewart:
> Fixes that annoying error where we get `java.lang.IllegalArgumentException: Invalid hexadecimal character 'I' at index 0` from the `UInt64` Mapper.
> 
> This was occurring because the uint64 mapper uses a string whereas the db columns were using `INTEGER` so sometimes sqlite would store `Inf` instead of the hex.

So isn't the real problem here that serial id can be `Inf` ? That seems like a bug to me

benthecarman:
> So isn't the real problem here that serial id can be `Inf` ? That seems like a bug to me

The serial Id was never `Inf` there just some encodings of `UInt64` that sqlite interprets as `Inf`

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -1,33 +1,33 @@
...
+    "state"                 TEXT    NOT NULL,
+    "is_initiator"          INTEGER NOT NULL,
+    "account"               TEXT    NOT NULL,
+    "change_index"          INTEGER NOT NULL,
+    "key_index"             INTEGER NOT NULL,
 
-    "fee_rate"              VARCHAR(254) NOT NULL,
-    "fund_output_serial_id" INTEGER      NOT NULL,
+    "fee_rate"              TEXT    NOT NULL,
+    "fund_output_serial_id" TEXT    NOT NULL,
```
Christewart, line 14: Can we apply these as a new migration rather than modifying an existing one?

dlc-wallet/src/main/resources/sqlite/dlc/migration/V1__dlc_db_baseline.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE "global_dlc_data"
     "key_index"             INTEGER      NOT NULL,
 
     "fee_rate"              VARCHAR(254) NOT NULL,
```
Christewart, line 13: fee rate not changed here?

Christewart (changes requested):
Yeah, in general this will break migrations i believe with existing dlcdb. We need to make a V2 migration

dlc-wallet/src/main/resources/sqlite/dlc/migration/V1__dlc_db_baseline.sql
benthecarman: only changed in the other because postgres should use `TEXT`

dlc-wallet/src/main/resources/postgresql/dlc/migration/V1__dlc_db_baseline.sql
benthecarman: I don't think we are able to write a valid v2 migration, idk if we can get the proper uint64 hex encoding from sql's integer, and even if we can, will require decontruscting and then rebuilding the entire db because the `global_dlc_data` has a serial id in it

Christewart:
> Fixes that annoying error where we get `java.lang.IllegalArgumentException: Invalid hexadecimal character 'I' at index 0` from the `UInt64` Mapper.
> 
> This was occurring because the uint64 mapper uses a string whereas the db columns were using `INTEGER` so sometimes sqlite would store `Inf` instead of the hex.

Some more context from discussions @benthecarman and I had on slack

1. [sqlite can only store values from `-2^63` - `2^63-1`](https://stackoverflow.com/questions/4448284/sqlite3-integer-max-value)
2. serial ids are `UInt64` which gives them the possibility to represent up to `2^64-1`
3. sqlite will just place `Inf` in the column if the value is out of range for their `Integer` column type
4. We cannot safely migrate the serial ids in existing databases as we cannot do math on `Inf`. We have no idea what the original serialid was [because it is randomly generated](https://github.com/bitcoin-s/bitcoin-s/blob/be8e965367d8adad2b1bf478de215d88f835e84a/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCMessage.scala#L29). The only thing we know is it is between `2^63` - `2^64-1`.

Given all this, and that this change was introduced into master a few days ago (#3138) I think it is reasonable to make this change.

nkohen (approved):

nkohen:
Keep in mind if there's anyone who has been testing for us beforehand, they will need to delete their DBs (once making sure they have no active DLCs)

PR #3174, opened 2021-05-27
Replace GUI balance thread with a akka scheduler

### LINKED ISSUE #3156: Change isn't reflected on GUI
I had 10,000,000 sats in the wallet, entered into a DLC with nadav for 50,000 sats. Now I don't have any balance in my wallet under `unconfirmed` or `confirmed`, but oddly enough my total balance seems correct

Screenshot from 2021-05-26 15-34-31

benthecarman:
Fixes #3156

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
@@ -1,18 +1,20 @@
 package org.bitcoins.gui
 
+import akka.actor.{ActorSystem, Cancellable}
 import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.ConsoleCli
 import org.bitcoins.core.currency.{Bitcoins, Satoshis}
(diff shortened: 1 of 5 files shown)


app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
```diff
@@ -23,23 +25,22 @@ class WalletGUIModel() {
   lazy val parentWindow: ObjectProperty[Window] =
     ObjectProperty[Window](null.asInstanceOf[Window])
 
-  private case object UpdateBalanceRunnable extends Runnable {
+  val system: ActorSystem = ActorSystem(
```
Christewart: Can we take in the system as a parameter? So we can share the system across gui + backend

i.e. 

```scala
class WalletGUIModel()(implicit system: ActorSystem) { ... } 
```

That way we don't have two redundant thread pools, and in the future we might use akka's event bus to propagate updates to the GUI

Christewart (approved):

PR #3179, opened 2021-05-27
Make add sigs broadcast the funding transaction

benthecarman:
Makes it so when we add sigs with the GUI we also broadcast the funding transaction, this should simplify the flow a bit.

Curious on thoughts on if we should rename the "Add Sigs" button to something else.


Christewart:
> Curious on thoughts on if we should rename the "Add Sigs" button to something else.

:100: agree with this. It is very vague. Maybe `Complete DLC` or `Finalize DLC` ?

benthecarman:
>  agree with this. It is very vague. Maybe `Complete DLC` or `Finalize DLC` ?

Complete and finalized are too close to execution, maybe `Open DLC`

Christewart (approved):

PR #3185, opened 2021-05-27
Add unit test we can handle spending funds and receiving funds in same tx

Christewart:
I thought this would a bug @nkohen was running into, apparently it is something else. I still think this unit test is useful though.


benthecarman (approved):

benthecarman:
I think the error will occur if instead of calling `processTransaction` twice, you put both txs in a block and have the wallet process that block.

PR #3194, opened 2021-05-28
Fix verify funding sigs

benthecarman:
We were filtering for the accepter's inputs instead of the initiator's inputs for funding input verifcation. This wasn't caught because all of our tests use small enough amounts that we only need 1 input.


Christewart (changes requested):
Any explanation why this is wrong or a unit test tot prevent regression?

benthecarman:
> Any explanation why this is wrong or a unit test tot prevent regression?

Updated description

Christewart (approved):

Christewart:
Compiler error: https://github.com/bitcoin-s/bitcoin-s/pull/3194/checks?check_run_id=2698586340#step:5:230

PR #3198, opened 2021-05-30
2021 05 29 dlc pnl

Christewart:
This adds a DLC PNL & rate of return columns to our table. Now you can see how much money you made on your DLC

Screenshot from 2021-05-30 09-18-38

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
@@ -260,6 +260,8 @@ object Picklers {
       )
     }
 
+  private val pnlKey: String = "pnl"
+
   implicit val claimedW: Writer[Claimed] = writer[Obj].comap { claimed =>
     import claimed._
     val (oraclesJs, outcomesJs) = oracleOutcome match {
@@ -290,7 +292,8 @@ object Picklers {
       "closingTxId" -> Str(closingTxId.hex),
       "oracleSigs" -> oracleSigs.map(sig => Str(sig.hex)),
       "outcomes" -> outcomesJs,
-      "oracles" -> oraclesJs
+      "oracles" -> oraclesJs,
+      pnlKey -> Num(claimed.pnl.satoshis.toLong.toDouble)
     )
   }
 
@@ -325,7 +328,8 @@ object Picklers {
         "closingTxId" -> Str(closingTxId.hex),
         "oracleSigs" -> oracleSigs.map(sig => Str(sig.hex)),
         "outcomes" -> outcomesJs,
-        "oracles" -> oraclesJs
+        "oracles" -> oraclesJs,
+        pnlKey -> Num(remoteClaimed.pnl.satoshis.toLong.toDouble)
       )
     }
 
@@ -347,7 +351,8 @@ object Picklers {
       "localCollateral" -> Num(localCollateral.satoshis.toLong.toDouble),
       "remoteCollateral" -> Num(remoteCollateral.satoshis.toLong.toDouble),
       "fundingTxId" -> Str(fundingTxId.hex),
-      "closingTxId" -> Str(closingTxId.hex)
+      "closingTxId" -> Str(closingTxId.hex),
+      pnlKey -> Num(refunded.pnl.satoshis.toLong.toDouble)
     )
   }
 
@@ -420,6 +425,9 @@ object Picklers {
       case signed: SignedNumericOutcome =>
         throw new IllegalArgumentException(s"Unexpected outcome $signed")
     }
+
+    lazy val pnlJs = obj("pnl")
+    lazy val pnlOpt = pnlJs.numOpt.map(sats => Satoshis(sats.toLong))
 
     state match {
       case DLCState.Offered =>
@@ -497,7 +505,8 @@ object Picklers {
           fundingTxId,
           closingTxId,
           oracleSigs,
-          oracleOutcome
+          oracleOutcome,
+          pnlOpt.get
         )
       case DLCState.RemoteClaimed =>
         require(oracleSigs.size == 1,
@@ -515,7 +524,8 @@ object Picklers {
           fundingTxId,
           closingTxId,
           oracleSigs.head,
-          oracleOutcome
+          oracleOutcome,
+          pnlOpt.get
         )
       case DLCState.Refunded =>
         Refunded(
@@ -529,7 +539,8 @@ object Picklers {
           totalCollateral,
           localCollateral,
           fundingTxId,
-          closingTxId
+          closingTxId,
+          pnlOpt.get
         )
     }
   }
--- a/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
@@ -261,6 +261,7 @@ object Picklers {
     }
 
   private val pnlKey: String = "pnl"
+  private val rateOfReturnKey: String = "rateOfReturn"
 
   implicit val claimedW: Writer[Claimed] = writer[Obj].comap { claimed =>
     import claimed._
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/bitcoins/core/dlc/accounting/DlcAccounting.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.bitcoins.core.currency.CurrencyUnit
+import org.bitcoins.core.protocol.BitcoinAddress
+import org.bitcoins.crypto.Sha256Digest
+
+case class DlcAccounting(
+    dlcId: Sha256Digest,
+    myCollateral: CurrencyUnit,
+    theirCollateral: CurrencyUnit,
+    myPayoutAddress: BitcoinAddress,
+    theirPayoutAddress: BitcoinAddress,
```
benthecarman: I don't see the need for these 2

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCStatus.scala
```diff
@@ -40,6 +40,8 @@ sealed trait BroadcastedDLCStatus extends AcceptedDLCStatus {
 
 sealed trait ClosedDLCStatus extends BroadcastedDLCStatus {
   def closingTxId: DoubleSha256DigestBE
+  def pnl: CurrencyUnit
+  def rateOfReturn: BigDecimal
```
benthecarman: Instead of this we could put `myPayout` and calculate the rest

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1223,13 +1226,26 @@ abstract class DLCWallet
...
+
   override def findDLC(dlcId: Sha256Digest): Future[Option[DLCStatus]] = {
     for {
       dlcDbOpt <- dlcDAO.read(dlcId)
       contractDataOpt <- contractDataDAO.read(dlcId)
       offerDbOpt <- dlcOfferDAO.read(dlcId)
+      acceptDbOpt <- dlcAcceptDAO.read(dlcId)
       (announcements, announcementData, nonceDbs) <- getDLCAnnouncementDbs(
         dlcId)
+      closingTxFOpt <- dlcDbOpt.map(dlcDb => getClosingTxOpt(dlcDb)) match {
```
benthecarman, line 1245: imo instead of doing this, we should add `myPayoutOpt` as a column to `global_dlc_data` and just do all this calculation once.

benthecarman (commented):
Should also add the ViewDLCDialog

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
Christewart: I think we should just add a `dlc_accounting` table so we can just keep all this stuff separate. That way we can extend it for more interesting accounting metrics in the future.

Christewart: #3201

Christewart:
Added to `ViewDLCDialog` 

Screenshot from 2021-05-31 08-15-21

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
```diff
@@ -99,6 +99,47 @@ object ViewDLCDialog {
...
+          add(new Label("My payout:"), 0, row)
+          add(new TextField() {
+                text = s"${closed.myPayout}"
+                editable = false
+              },
+              columnIndex = 1,
+              rowIndex = row)
+
+          row += 1
+          add(new Label("Their payout:"), 0, row)
```
benthecarman: Should use `Counter Party Payout` as that is what we call the colateral

core/src/main/scala/org/bitcoins/core/dlc/accounting/DlcAccounting.scala
```diff
@@ -0,0 +1,26 @@
+package org.bitcoins.core.dlc.accounting
+
+import org.bitcoins.core.currency.CurrencyUnit
+import org.bitcoins.crypto.Sha256Digest
+
+case class DlcAccounting(
```
benthecarman: nit: `DLCAccounting`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1372,6 +1411,45 @@ abstract class DLCWallet
     }
   }
 
+  private def calculatePnl(
+      dlcDb: DLCDb,
+      offerDb: DLCOfferDb,
+      acceptDb: DLCAcceptDb,
+      closingTxDb: TransactionDb): DlcAccounting = {
```
benthecarman: should make this just `Transaction`, then we can call in the processing code

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
```diff
@@ -99,6 +99,47 @@ object ViewDLCDialog {
...
+          add(new TextField() {
+                text = s"${closed.theirPayout}"
+                editable = false
+              },
+              columnIndex = 1,
+              rowIndex = row)
+
+          row += 1
+          add(new Label("PNL:"), 0, row)
```
benthecarman, line 123: Would be nice to have a tool tip explaining PNL, same with rate of return

app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
```diff
@@ -260,6 +260,11 @@ object Picklers {
       )
     }
 
+  private val myPayoutKey: String = "myPayout"
+  private val theirPayoutKey: String = "theirPayout"
```
benthecarman: same thing with `counter party payout`

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -90,10 +96,41 @@ class DLCTableView(model: DLCPaneModel) {
       }
     }
 
+    val pnlCol = new TableColumn[DLCStatus, String] {
```
benthecarman, line 86: these new columns make it so the table is larger than the default screen, maybe we should drop the my and counter party col columns in place

Christewart: Here is what i got it to on 872a9fe 

Screenshot from 2021-05-31 08-57-16

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -47,7 +47,7 @@ class DLCTableView(model: DLCPaneModel) {
 
     val statusCol = new TableColumn[DLCStatus, String] {
       text = "Status"
-      prefWidth = 150
+      prefWidth = 75
```
benthecarman: I'd rather not reduce the size of this one, for `RemoteClaimed` it will be cut off

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
benthecarman: ping

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
benthecarman: for me total col is a bit cut off and RoR is cut off a bit on both, imo it'd be better to drop the 2 collateral columns and give the others some room, we don't want it too cluttered

image

Christewart:
What the tableview looks like as of 6bd877e8

Screenshot from 2021-05-31 09-17-19

Christewart:
Table view as of b1e4b56

Screenshot from 2021-05-31 09-30-23

benthecarman (approved):

PR #3199, opened 2021-05-30
Add dlc wallet test to postgres matrix

Christewart:
This was missed on #3138 and likely is the reason we are constantly failing in our postgres matrix


benthecarman (approved):

benthecarman:
Was error because we had the primary key defined twice and we used an `INTEGER` col instead of `BOOLEAN`, I didn't know postgres has a boolean type

PR #3200, opened 2021-05-30
Get logging working in bundle 

Christewart:
I suspect that adding the `actorSystem` in #3174 broke this comment that we had left previously about using `System.setProperty()` to set the log location

https://github.com/bitcoin-s/bitcoin-s/blob/41468763695a11e44658f6e7c4d8797d58841057/app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala#L98


benthecarman (changes requested):
Should do the same for `org.bitcoins.gui.Gui`

benthecarman (approved):

PR #3203, opened 2021-05-31
Sort points in numeric gui

benthecarman:


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -573,7 +575,8 @@ object CreateDLCOfferDialog {
           }
 
           val sorted = outcomesValuePoints.sortBy(_.outcome)
-          require(sorted == outcomesValuePoints, "Must be sorted by outcome")
+          require(sorted == outcomesValuePoints,
+                  s"Must be sorted by outcome, got $sorted")
```
Christewart: you should print `outcomesValuePoints` as that is the pre-sorted data structure. `sorted` isn't useful IMO unless you are comparing it against `outcomesValuePoints`

Christewart (approved):

PR #3205, opened 2021-05-31
Update balance on cancel DLC

benthecarman:


project/Deps.scala
```diff
@@ -24,7 +24,7 @@ object Deps {
     val nativeLoaderV = "2.3.5"
     val typesafeConfigV = "1.4.1"
 
-    val scalaFxV = "16.0.0-R22"
+    val scalaFxV = "15.0.1-R21"
```
Christewart: why downgrade here?

benthecarman: oh accident, was testing something

PR #3210, opened 2021-05-31
Windows packaging

benthecarman:
redo of #3208 so we get proper CI to run


.github/workflows/release.yml
```diff
@@ -1,7 +1,7 @@
 name: Release
 on:
   push:
-    branches: [master, main, adaptor-dlc]
+    branches: [master, main, windows-packaging]
```
Christewart: Remove when we are certain this is working

app/bundle/bundle.sbt
```diff
@@ -11,8 +16,50 @@ assembly / mainClass := Some("org.bitcoins.bundle.gui.BundleGUI")
...
+  id = "shortcuts",
+  title = "Shortcuts in start menu",
+  desc = "Add shortcuts for execution and uninstall in start menu",
+  components = Seq(
+    AddShortCuts(Seq("bin/bitcoin-s-bundle.bat"))
+  )
+)
+
+// for windows class paths being too long
+scriptClasspath := Seq("*")
```
Christewart, line 52: So is all this stuff needed? My understanding is this all this is extra configuration that wasn't needed because we weren't uninstalling the program correctly.

app/bundle/src/main/resources/application.conf
```diff
@@ -1,8 +1,209 @@
 bitcoin-s {
-    network = mainnet
+  network = mainnet # regtest, testnet3, mainnet, signet
```
Christewart: Same here, my understanding is our problem was the uinstall process, that corrupted the configuration of a new installation.

app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
```diff
@@ -43,6 +44,10 @@ object BundleGUI extends WalletGUI with JFXApp {
     val usedDir = DatadirUtil.getFinalDatadir(datadir, baseConfig, None)
 
     System.setProperty("bitcoins.log.location", usedDir.toAbsolutePath.toString)
+
+    if (Properties.isWin) {
+      System.setProperty("HOME", datadir.getParent.toAbsolutePath.toString)
```
Christewart: What is this? Things like this scare me. Why do we need a special exception on windows? At least add a comment

app/server-routes/src/main/scala/org/bitcoins/server/routes/BitcoinSRunner.scala
```diff
@@ -90,6 +90,10 @@ trait BitcoinSRunner extends StartStopAsync[Unit] with Logging {
     val usedDir: Path =
       DatadirUtil.getFinalDatadir(datadir, baseConfig, customFinalDirOpt)
 
+    if (Properties.isWin) {
```
Christewart: Same here, ideally we wouldn't need to do this twice rather do it in one spot.

Christewart (changes requested):

app/bundle/src/main/resources/application.conf
benthecarman: No, the uninstall process made it so I wasn't correctly installing any of the changes I made so I wasn't sure what changes I did worked.

ATM we need to se the config here for the windows packaging to work

app/bundle/bundle.sbt
benthecarman: No, the uninstall process made it so I wasn't correctly installing any of the changes I made so I wasn't sure what changes I did worked.

Christewart (approved):

PR #3211, opened 2021-05-31
Only update DLC we are viewing

benthecarman:
should speed up the gui


Christewart (approved):
I can't see how this hurts, but it doesn't really fix the latency on `View DLC`

Issue #3213, opened 2021-06-01
DLCWallet OracleOutcome computation for completed events too slow


nkohen:
This line is the culprit:
```
            contractInfo.sigPointMap(aggSig.sig.toPrivateKey.publicKey)
```
which takes a fresh (non-cached) `ContractInfo` and computes all adaptor points and all outcomes and then finds the outcome at the index of the detected adaptor point which was used.

Some options:
1. Store the outcome in `DLCDb` (the global table) as an Option so that it still needs to be computed, but only once so that all future calls will just do a database read instead of an expensive computation. I personally think this is the biggest bang for our buck as it will effectively solve the problem and is a minimal addition of a nullable column.
2. Store the outcome for every CETSig/Adaptor point. This requires a significant increase in data to be stored in the DB, but removes the problem here entirely.

Curious to hear @benthecarman's thoughts?

Generally I will note however that short of storing a linked structure (a binary trie in this case) of adaptor points, there is no particularly nice way to handle this problem which is why I'm in favor of just storing the executed outcome in the DLCDb at the moment.

benthecarman:
Yeah we should just store it in the global table

PR #3214, opened 2021-06-01
Rework findDLC()

Christewart:
This PR was an attempt to address #3213 but ended up just being a refactor. This PR is *NOT* intended to change functionality

There is way too much stuff going on in `DLCWallet`, this breaks things out into helper classes 

1. `DLCStatusBuilder`
2. `AccountingUtil`

As a by product, I added a few new states to `DLCState` 

1. `InProgressState` - this means that the DLC is being setup or exists onchain waiting for settlement
2. `ClosedState` - this means the DLC has been settled (not necessarily onchain, but someone has attempted to close it)
3. `ClosedViaOracleOutcomeState` - this means the DLC has been closed with an oracle attestment. This is different than the `Refund` state where there isn't an oracle attestment used.

I don't think this significantly improves performance, but it definitely helps readability, maintainability and correctness

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -7,14 +7,12 @@ import org.bitcoins.core.api.wallet.db._
 import org.bitcoins.core.config.BitcoinNetwork
 import org.bitcoins.core.crypto.ExtPublicKey
 import org.bitcoins.core.currency._
-import org.bitcoins.core.dlc.accounting.DLCAccounting
 import org.bitcoins.core.hd._
 import org.bitcoins.core.number._
 import org.bitcoins.core.protocol._
 import org.bitcoins.core.protocol.dlc.build.DLCTxBuilder
 import org.bitcoins.core.protocol.dlc.models.DLCMessage.DLCAccept._
 import org.bitcoins.core.protocol.dlc.models.DLCMessage._
-import org.bitcoins.core.protocol.dlc.models.DLCStatus._
 import org.bitcoins.core.protocol.dlc.models._
 import org.bitcoins.core.protocol.dlc.sign._
 import org.bitcoins.core.protocol.script._
@@ -26,6 +24,7 @@ import org.bitcoins.core.wallet.utxo._
 import org.bitcoins.crypto._
 import org.bitcoins.dlc.wallet.internal._
 import org.bitcoins.dlc.wallet.models._
+import org.bitcoins.dlc.wallet.util.DLCStatusBuilder
 import org.bitcoins.keymanager.bip39.BIP39KeyManager
 import org.bitcoins.wallet.config.WalletAppConfig
 import org.bitcoins.wallet.models.TransactionDAO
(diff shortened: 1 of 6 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1235,228 +1234,102 @@ abstract class DLCWallet
...
-        // Only called when safe
-        lazy val (oracleOutcome, sigs) = {
-          val aggSig = dlcDb.aggregateSignatureOpt.get
-          val outcome =
-            contractInfo.sigPointMap(aggSig.sig.toPrivateKey.publicKey)
-
-          val sigs = nonceDbs.flatMap(_.signatureOpt)
-
-          (outcome, sigs)
+    val dlcDbOptF = dlcDAO.read(dlcId)
```
Christewart, line 1238: parallelize these to try and improve performance, doesn't make a significant difference due to #3213

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/util/DLCStatusBuilder.scala
```diff
@@ -0,0 +1,245 @@
+package org.bitcoins.dlc.wallet.util
+
+import org.bitcoins.core.dlc.accounting.DLCAccounting
+import org.bitcoins.core.protocol.dlc.models.{
```
benthecarman, line 4: nit: could use an `Optimize imports`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/util/DLCStatusBuilder.scala
```diff
@@ -0,0 +1,245 @@
...
+  OracleNonceDb
+}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object DLCStatusBuilder {
+
+  /** Helper method to convert a bunch of indepdendent datastructures into a in progress dlc status */
+  def buildInProgressDLCStatus(
+      state: DLCState.InProgressState,
```
benthecarman: why even take `state` as a param here and just use `dlcDb.state` everywhere

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/util/DLCStatusBuilder.scala
```diff
@@ -0,0 +1,245 @@
...
+          localCollateral,
+          dlcDb.fundingTxIdOpt.get
+        )
+    }
+
+    status
+  }
+
+  def buildClosedDLCStatus(
+      state: DLCState.ClosedState,
```
benthecarman: ditto

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/util/DLCStatusBuilder.scala
Christewart: it is useful because it is a more precise type. The problem with `DlcDb` is that it represents too many states. We need to eventually make sub types to represent the precise state of the DLC

benthecarman: Maybe the best solution would be to make a new type (maybe `DLCGlobalData`?) and have it by typed based on the state. Then we can just make a simple `toDLCGlobalData` function on `DLCDb`.

nkohen (approved):
I think this is an acceptable workaround until the DB column is added 👍

PR #3215, opened 2021-06-01
Remove all flyway plugin sbt config

Christewart:
This removes the sbt plugin for flyway. We keep flyway around, but just as a normal scala dependency. 

The motivation for doing this is because

1. We aren't using it
2. We are experiencing reports from users about the `~/.bitcoin-s` directory not existing. I'm worried us developer are missing bugs because -- as @benthecarman pointed out -- this flyway logic would create the directories

Screenshot from 2021-06-01 14-49-20

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,11 +4,6 @@ import scala.util.Properties
 
 Global / cancelable := true
 
-//don't allow us to wipe all of our prod databases
-flywayClean / aggregate := false
-//allow us to wipe our test databases
-Test / flywayClean / aggregate := true
-
 lazy val Benchmark = config("bench") extend Test
 
 lazy val benchSettings: Seq[Def.SettingsDefinition] = {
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -760,60 +737,3 @@ lazy val dlcTest = project
...
-  lazy val testnetDir = s"${System.getenv("HOME")}/.bitcoin-s/testnet3/"
-  lazy val regtestDir = s"${System.getenv("HOME")}/.bitcoin-s/regtest/"
-  lazy val signetDir = s"${System.getenv("HOME")}/.bitcoin-s/signet/"
-  lazy val unittestDir = s"${System.getenv("HOME")}/.bitcoin-s/unittest/"
-
-  lazy val dirs =
-    List(mainnetDir, testnetDir, regtestDir, signetDir, unittestDir)
-
-  //create directies if they DNE
-  dirs.foreach { d =>
```
Christewart, line 782: This is where the build config creates the directories automatically for us that leads us not to see errors that users my encounter

benthecarman (approved):

PR #3220, opened 2021-06-03
Implement sharing of actor systems across the GUI and backend app server

Christewart:
fixes #3174 

Previously we would have two actor systems (and two associated thread pools) when using `bundle`. This PR makes it so that the actor systems are shared. 

It also introduces a new trait called `BitcoinSApp` that  `extends App`. This should be extended by all objects that we want to have runnable by the jvm on startup.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -375,7 +375,9 @@ class BitcoinSServerMain(override val args: Array[String])(implicit
 }
 
 object BitcoinSServerMain extends BitcoinSApp {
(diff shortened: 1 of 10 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -374,7 +374,9 @@ class BitcoinSServerMain(override val args: Array[String])
   }
 }
 
-object BitcoinSServerMain extends App {
+object BitcoinSServerMain extends BitcoinSApp {
+  override val actorSystemName = "bitcoin-s-server"
```
benthecarman: maybe should concat the system time incase we have multiple bitcoin-s running

PR #3221, opened 2021-06-03
2021 06 02 adddlcsigsfromfile

Christewart:
This is the same as #3219 with some bonus features

1. Makes `adddlcsigsandbroadcastfromfile` so that we can broadcast a tx immediately after adding sigs
2. Restores the abillity for `signdlcfromdialog` to be from a file too

This ends up being a bottleneck that i've encountered with users executing numeric DLCs.

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -316,6 +316,21 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("adddlcsigsandbroadcastfromfile")
+        .action((_, conf) =>
+          conf.copy(command =
+            AddDLCSigsAndBroadcastFromFile(new File("").toPath)))
+        .text("Adds DLC Signatures into the database")
+        .children(
+          arg[Path]("path")
+            .required()
+            .action((path, conf) =>
+              conf.copy(command = conf.command match {
+                case addDLCSigs: AddDLCSigsAndBroadcastFromFile =>
+                  addDLCSigs.copy(path = path)
+                case other => other
+              }))
+        ),
       cmd("adddlcsigsandbroadcast")
         .action((_, conf) => conf.copy(command = AddDLCSigsAndBroadcast(null)))
         .text("Adds DLC Signatures into the database and broadcasts the funding transaction")
@@ -1506,6 +1521,8 @@ object ConsoleCli {
         RequestParam("adddlcsigsfromfile", Seq(up.writeJs(path)))
       case AddDLCSigsAndBroadcast(sigs) =>
         RequestParam("adddlcsigsandbroadcast", Seq(up.writeJs(sigs)))
+      case AddDLCSigsAndBroadcastFromFile(path) =>
+        RequestParam("adddlcsigsandbroadcastfromfile", Seq(up.writeJs(path)))
       case ExecuteDLC(contractId, oracleSigs, noBroadcast) =>
         RequestParam("executedlc",
                      Seq(up.writeJs(contractId),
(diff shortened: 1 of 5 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -1,190 +1,220 @@
...
-import scalafx.scene.control._
-import scalafx.scene.layout._
-import scalafx.stage.Window
 
-import scala.collection._
-import scala.util.{Failure, Success, Try}
+import org.bitcoins.gui.dlc.GlobalDLCData
+
+import scalafx.scene.Node
+
```
benthecarman, line 9: this removes the data appearing for the sign message, is that what we want?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -316,6 +316,21 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("adddlcsigsandbroadcastfromfile")
```
nkohen, line 319: nit: I prefer `adddlcsigsfromfileandbroadcast`

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -1,190 +1,220 @@
...
+                                                   Some(file)
+                                                 DLCDialog.acceptFileChosenLabel.text =
+                                                   file.toString
+                                               }),
+                                           DLCDialog.fileChosenStr -> DLCDialog.acceptFileChosenLabel
+                                         ),
+                                         Vector(DLCDialog.dlcAcceptStr,
+                                                DLCDialog.dlcAcceptFileStr)) {
+
+//  override def showAndWait(parentWindow: Window): Option[SignDLCCliCommand] = {
```
nkohen, line 30: ping

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -1,190 +1,220 @@
...
+//    }
+//  }
+
+  override def constructFromInput(
+      inputs: Predef.Map[String, Node]): SignDLCCliCommand = {
+    DLCDialog.acceptDLCFile match {
+      case Some(file) =>
+        DLCDialog.acceptDLCFile = None // reset
+        DLCDialog.acceptFileChosenLabel.text = "" // reset
+        SignDLCFromFile(file.toPath, destination = None) //is none right here?
```
nkohen, line 207: destination is None if you aren't writing the result to a file, otherwise I believe it is supposed to contain the path

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -1,190 +1,220 @@
...
-    dialog.resultConverter = dialogButton =>
-      if (dialogButton == ButtonType.OK) {
-
-        val acceptHex = acceptTLVTF.text.value
         val accept = LnMessageFactory(DLCAcceptTLV).fromHex(acceptHex)
 
-        Some(SignDLC(accept))
-      } else None
+        //is this right???
+        GlobalDLCData.lastContractId = accept.tlv.tempContractId.hex
```
nkohen, line 214: I believe so

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
nkohen: I'm also a bit confused about what is happening here, this is for the sign not the add sigs and broadcast flow?

nkohen: oh I see that it is to add file stuff to sign but it also gets rid of existing deserialization logic instead of being an alternative, @benthecarman could you maybe make a quick fix for this?

Christewart:
I believe this is superseded by #3243

PR #3229, opened 2021-06-07
2021 06 07 dlc wallet pnl

Christewart:
Adds profit and loss and rate of return for all _closed_ DLCs in your wallet. The current calculation does _not_ account for blockchain fees

Screenshot from 2021-06-07 15-31-02

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
@@ -1,6 +1,7 @@
 package org.bitcoins.gui
 
 import akka.actor.{ActorSystem, Cancellable}
+import grizzled.slf4j.Logging
 import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.ConsoleCli
 import org.bitcoins.commons.serializers.PicklerKeys
@@ -21,7 +22,8 @@ import scala.concurrent.duration.DurationInt
 import scala.concurrent.{Await, Promise}
 import scala.util.{Failure, Success, Try}
 
-class WalletGUIModel(dlcModel: DLCPaneModel)(implicit system: ActorSystem) {
+class WalletGUIModel(dlcModel: DLCPaneModel)(implicit system: ActorSystem)
+    extends Logging {
   var taskRunner: TaskRunner = _
   import system.dispatcher
 
@@ -158,13 +160,7 @@ class WalletGUIModel(dlcModel: DLCPaneModel)(implicit system: ActorSystem) {
   private def updateWalletAccounting(): Unit = {
     ConsoleCli.exec(GetDLCWalletAccounting, GlobalData.consoleCliConfig) match {
       case Failure(err) =>
-        err.printStackTrace()
-        val _ = new Alert(AlertType.Error) {
-          initOwner(owner)
-          title = "Could not retrieve dlc wallet accounting"
-          headerText = s"Operation failed. Exception: ${err.getClass}"
-          contentText = err.getMessage
-        }.showAndWait()
+        logger.error(s"Error fetching accounting", err)
       case Success(commandReturn) =>
         val json = ujson.read(commandReturn).obj
         val pnl = json(PicklerKeys.pnl).num.toLong.toString
--- a/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
@@ -3,7 +3,9 @@ package org.bitcoins.gui
 import akka.actor.{ActorSystem, Cancellable}
 import org.bitcoins.cli.CliCommand._
 import org.bitcoins.cli.ConsoleCli
+import org.bitcoins.commons.serializers.PicklerKeys
 import org.bitcoins.core.currency.{Bitcoins, Satoshis}
+import org.bitcoins.core.dlc.accounting.RateOfReturnUtil
 import org.bitcoins.core.protocol.BitcoinAddress
 import org.bitcoins.core.wallet.fee.FeeUnit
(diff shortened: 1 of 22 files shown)


app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
```diff
@@ -151,4 +154,25 @@ class WalletGUIModel(dlcModel: DLCPaneModel)(implicit system: ActorSystem) {
...
+  private def updateWalletAccounting(): Unit = {
+    ConsoleCli.exec(GetDLCWalletAccounting, GlobalData.consoleCliConfig) match {
+      case Failure(err) =>
+        err.printStackTrace()
+        val _ = new Alert(AlertType.Error) {
+          initOwner(owner)
+          title = "Could not retrieve dlc wallet accounting"
+          headerText = s"Operation failed. Exception: ${err.getClass}"
+          contentText = err.getMessage
+        }.showAndWait()
```
benthecarman: We shouldn't create an alert here, this could get really annoying where every 10 seconds an alert pops up with an error.

We should probably just log the error and continue.

Maybe eventually put up a warning `!` next to the text with a tooltip saying there's an error with the wallet accounting.

core/src/main/scala/org/bitcoins/core/dlc/accounting/RateOfReturnUtil.scala
```diff
@@ -2,7 +2,9 @@ package org.bitcoins.core.dlc.accounting
 
 object RateOfReturnUtil {
 
+  /** @see https://alvinalexander.com/scala/how-to-format-numbers-commas-international-currency-in-scala/ */
   def prettyPrint(ror: BigDecimal): String = {
-    (ror * 100).toString() + "%"
+    val percent = ror * 100
+    f"${percent}%1.2f" + "%"
```
benthecarman, line 8: is it possible to have it only display the decimal place when it's not zero, that way it is `100%` instead of `100.00%`?

app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
```diff
@@ -33,6 +35,7 @@ class WalletGUIModel(dlcModel: DLCPaneModel)(implicit system: ActorSystem) {
     override def run(): Unit = {
       Platform.runLater {
         updateBalance()
+        updateWalletAccounting()
         updateWalletInfo()
         dlcModel.updateDLCs()
```
benthecarman, line 42: Might be worth filing an issue to batch some of these things into a single command. `updateWalletAccounting` and 
`updateDLCs`are both calling `listDLCs`, it would be nice if we could put these together

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -794,6 +794,13 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
           Server.httpSuccess(fee.toString)
         }
       }
+
+    case ServerCommand("getdlcwalletaccounting", _) =>
+      complete {
+        wallet.getWalletAccounting.map { accounting =>
+          Server.httpSuccess(writeJs(accounting))
```
benthecarman, line 801: nit: you shouldn't need to put the `writeJs` it should be implicitly called

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -794,6 +794,13 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
           Server.httpSuccess(fee.toString)
         }
       }
+
+    case ServerCommand("getdlcwalletaccounting", _) =>
```
benthecarman, line 798: need a test for this in `RoutesSpec`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/accounting/Financials.scala
```diff
@@ -0,0 +1,17 @@
+package org.bitcoins.dlc.wallet.accounting
+
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.dlc.wallet.models.{DLCAcceptDb, DLCDb, DLCOfferDb}
+
+case class Financials(
```
benthecarman: nit: imo this name doesn't really reflect what it contains.

Maybe `DLCAccountingDbs`

benthecarman (changes requested):

benthecarman:
This doesn't account for Refunds correct? imo that shouldn't count as the bet never truly happened

app/gui/src/main/scala/org/bitcoins/gui/WalletGUIModel.scala
Christewart: #3234

Christewart:
> This doesn't account for Refunds correct? imo that shouldn't count as the bet never truly happened

This does account for refunds. If refunds are correct (i.e. we don't footgun like we did in #3230) they should be accounted IMO. 

In the current implementation, refunds do not affect PNL (you receive your contributed collateral back). However when we implement

>  The current calculation does not account for blockchain fees

Refunds will have a slightly negative PNL as blockchain fees do cost money.

core/src/main/scala/org/bitcoins/core/dlc/accounting/RateOfReturnUtil.scala
Christewart: I can't find how to do this easily. It is probably possible with java libraries, but I'm a little worried about introducing java specific libraries into `core` as they may not be supported in scalajs land.

benthecarman (approved):

PR #3236, opened 2021-06-08
Skip adding DLC sigs if we already have them

benthecarman:
This should help make `adddlcsigsandbroadcast` better if you are using very large messages, if the first attempt times out, when you redo it, you don't need to attempt to add all the sigs again and can instead just broadcast the ready dlc funding tx

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
@@ -1,7 +1,6 @@
 package org.bitcoins.server
 
 import akka.actor.ActorSystem
-import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server._
 import akka.stream.Materializer
--- a/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
@@ -1,6 +1,7 @@
 package org.bitcoins.server
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.HttpEntity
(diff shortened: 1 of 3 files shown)


Christewart:
compile failure: https://github.com/bitcoin-s/bitcoin-s/pull/3236/checks?check_run_id=2777059196#step:5:157

Christewart:
CI failure 

https://github.com/bitcoin-s/bitcoin-s/pull/3236/checks?check_run_id=2777380505#step:5:3475

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -796,6 +791,23 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
       }
   }
 
+  def addSigsAndBroadcastDLC(sigs: DLCSignTLV): Future[HttpEntity.Strict] = {
+    // attempt to broadcast first if we have already added the dlc sigs
+    // then we don't need to attempt to do it again
+    val attemptBroadcastF = wallet.broadcastDLCFundingTx(sigs.contractId)
```
Christewart: I don't understand how this is safe. Can't we just look up in the database to see if we have signatures already? If we do, just broadcast? It seems we are blindly attempting to broadcast, and if that fails, we try to actually add the sigs.

Christewart: I guess it seems to be we just need to make a database call to see if signatures already exist, if they do just broadcast right away, if not, add them

Christewart (changes requested):

Christewart:
Compile failing

Christewart:
CI failing.. https://github.com/bitcoin-s/bitcoin-s/pull/3236/checks?check_run_id=2785660654#step:5:3480

Christewart (approved):

PR #3237, opened 2021-06-08
Fix for showing txid on send

benthecarman:
We were writting to the `GlobalData.log` which is no longer used. This fixes it to be in the result area


Christewart (approved):

Christewart:
needs rebase

PR #3243, opened 2021-06-09
Populate Sign GUI when choosing a file

benthecarman:
image


Christewart:
Neither sign or accept seems to work for me on ae65189

Screen Shot 2021-06-09 at 6 37 18 PM
Screen Shot 2021-06-09 at 6 38 40 PM

Christewart (approved):
Tested and works for me

PR #3244, opened 2021-06-09
2021 06 07 markasspent optimization

Christewart:
Related to #2596 

Refactors database writes to be outside of `markAsSpent`. Now we determine all the state changes for the utxos, and then write them all together in a batch.


benthecarman (approved):

benthecarman:
I believe this is a bugfix too where we could mark some utxos as spent and throw an error on others in the same tx

PR #3245, opened 2021-06-10
2021 06 10 cache spendinginfodbs for block

Christewart:
Built on #3244 

related to #2596 

This moves database reads for utxos outside of `processReceivedUtxos()` and `processSpentUtxos()` into `Wallet.processBlock()`. 

This means we no longer have to reads from the database for every transaction in a block.

Some performance comparisons on rescans 

latest master (4b2bc379e35d302d7f0580bf59545b831794e907)

> 2021-06-10T15:35:27UTC INFO [DLCWallet$DLCWalletImpl] Finished rescanning the wallet. It took=3064954ms

on 96709e8ef387bc5bb3473e34174d175ecf88ddfa
> 2021-06-10T14:35:22UTC INFO [DLCWallet$DLCWalletImpl] Finished rescanning the wallet. It took 2176183ms

So it looks like this PR saves `888771 ms` or 15 minutes on rescanning blocks from height `550,000 - 687068`

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -36,7 +36,10 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
       blockHashOpt: Option[DoubleSha256DigestBE]
   ): Future[Wallet] = {
     for {
-      result <- processTransactionImpl(transaction, blockHashOpt, Vector.empty)
+      result <- processTransactionImpl(transaction,
+                                       blockHashOpt,
+                                       Vector.empty,
+                                       None)
     } yield {
       logger.debug(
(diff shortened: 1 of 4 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -107,6 +89,65 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
...
+                processTxResult.updatedIncoming.filter { spendingInfoDb =>
+                  block.transactions.exists(
+                    _.inputs.exists(
+                      _.previousOutput == spendingInfoDb.outPoint))
+                }
+              }
+
+              //add it to the cache
+              cachedSpentF =
+                Future.successful(spentSpendingInfo ++ spentInSameBlock)
```
benthecarman, line 141: shouldn't this be `Future.successful(cachedSpentF ++ spentInSameBlock)`

Christewart: `cachedSpentF` is a `Future`. We need to have the underlying vector which is `<-` on above.

benthecarman (approved):

PR #3249, opened 2021-06-11
Fix DLC Accounting errors

### LINKED ISSUE #3242: Error fetching account due to division by 0?
```
2021-06-09T18:51:29UTC ERROR [WalletGUIModel] Error fetching accounting
java.lang.RuntimeException: Request failed: Division undefined
	at org.bitcoins.cli.ConsoleCli$.error$1(ConsoleCli.scala:1464)
	at org.bitcoins.cli.ConsoleCli$.$anonfun$exec$1(ConsoleCli.scala:1799)
	at scala.util.Try$.apply(Try.scala:210)
	at org.bitcoins.cli.ConsoleCli$.exec(ConsoleCli.scala:1741)
	at org.bitcoins.gui.WalletGUIModel.org$bitcoins$gui$WalletGUIModel$$updateWalletAccounting(WalletGUIModel.scala:161)
	at org.bitcoins.gui.WalletGUIModel$UpdateWalletInfoRunnable$.$anonfun$run$1(WalletGUIModel.scala:41)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scalafx.application.Platform$$anon$1.run(Platform.scala:148)
```

benthecarman:
Fixes #3242

Also DLC Accounting was taking into account refunds as `Refunded` is a `ClosedDLCStatus`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/dlc/accounting/PayoutAccounting.scala
+++ b/core/src/main/scala/org/bitcoins/core/dlc/accounting/PayoutAccounting.scala
@@ -1,6 +1,6 @@
 package org.bitcoins.core.dlc.accounting
 
-import org.bitcoins.core.currency.CurrencyUnit
+import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
 
 /** Utility trait for metrics we need to do accounting */
 trait PayoutAccounting {
@@ -20,7 +20,9 @@ trait PayoutAccounting {
   /** Rate of return for the DLC
     * @see https://www.investopedia.com/terms/r/rateofreturn.asp
     */
-  def rateOfReturn: BigDecimal = pnl.toBigDecimal / myCollateral.toBigDecimal
+  def rateOfReturn: BigDecimal = if (myCollateral != Satoshis.zero)
+    pnl.toBigDecimal / myCollateral.toBigDecimal
+  else BigDecimal(0)
 
   def rorPrettyPrint: String = {
     RateOfReturnUtil.prettyPrint(rateOfReturn)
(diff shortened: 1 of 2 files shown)


Christewart:
> Also DLC Accounting was taking into account refunds as `Refunded` is a `ClosedDLCStatus`

Divide by zero bug fix makes sense to me, but i thought we agreed that `Refunded` is worth including as it will incur blockchain fees which will negatively affect your PNL

benthecarman:
> Divide by zero bug fix makes sense to me, but i thought we agreed that `Refunded` is worth including as it will incur blockchain fees which will negatively affect your PNL

Only problem is that it acts as you went break even on a bet. Example: I do one bet where I lose all my money, then do a refunded, my actual RoR should be I lost everything, but now it's not because I got money back from the refund.

Christewart:
> > Divide by zero bug fix makes sense to me, but i thought we agreed that `Refunded` is worth including as it will incur blockchain fees which will negatively affect your PNL
> 
> Only problem is that it acts as you went break even on a bet. Example: I do one bet where I lose all my money, then do a refunded, my actual RoR should be I lost everything, but now it's not because I got money back from the refund.

I'm confused by this hypothetical, how can you simultaneously get a refund and lose a bet? It seems to me you either have to publish a CET to the chain, or the refund transaction, but the hypoethetical you are proposing means you can double spend the funding utxo?

benthecarman:
> I'm confused by this hypothetical, how can you simultaneously get a refund and lose a bet? It seems to me you either have to publish a CET to the chain, or the refund transaction, but the hypoethetical you are proposing means you can double spend the funding utxo?

I am talking about the scenario where you have done 2 separate DLCs

Christewart:
> > I'm confused by this hypothetical, how can you simultaneously get a refund and lose a bet? It seems to me you either have to publish a CET to the chain, or the refund transaction, but the hypoethetical you are proposing means you can double spend the funding utxo?
> 
> I am talking about the scenario where you have done 2 separate DLCs

Let's keep discussing this on #3261

Issue #3162, opened 2021-05-26
Create Safe and Unsafe cancel DLC


benthecarman:
Currently we only have an unsafe cancel DLC where we do not double spend the funding inputs of the DLC. This should be done by the user if they want to safely cancel the DLC.

This only matters for the offerer after they have done a `DLCSign`

PR #3250, opened 2021-06-11
Only unreserve our utxos when canceling a DLC

benthecarman:
This would cause issues when doing a DLC with the same wallet, very niche but may as well fix

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -251,21 +251,22 @@ abstract class DLCWallet
     */
   override def cancelDLC(dlcId: Sha256Digest): Future[Unit] = {
     for {
-      inputs <- dlcInputsDAO.findByDLCId(dlcId)
-      dbs <- spendingInfoDAO.findByOutPoints(inputs.map(_.outPoint))
-      // allow this to fail in the case they have already been unreserved
-      _ <- unmarkUTXOsAsReserved(dbs).recover { case _: Throwable => () }
-
       dlcOpt <- dlcDAO.read(dlcId)
-      _ = dlcOpt match {
+      dlcDb = dlcOpt match {
         case Some(db) =>
           require(
             db.state == DLCState.Offered || db.state == DLCState.Accepted || db.state == DLCState.Signed,
             "Cannot cancel a DLC after it has been signed")
+          db
         case None =>
           throw new IllegalArgumentException(
             s"No DLC Found with dlc id ${dlcId.hex}")
       }
+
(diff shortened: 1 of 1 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -251,22 +251,23 @@ abstract class DLCWallet
...
+          db
         case None =>
           throw new IllegalArgumentException(
             s"No DLC Found with dlc id ${dlcId.hex}")
       }
 
+      inputs <- dlcInputsDAO.findByDLCId(dlcId, dlcDb.isInitiator)
+      dbs <- spendingInfoDAO.findByOutPoints(inputs.map(_.outPoint))
+      // allow this to fail in the case they have already been unreserved
+      _ <- unmarkUTXOsAsReserved(dbs).recover { case _: Throwable => () }
```
Christewart, line 269: it would be nice if we had a specific error here. `Throwable` can represent all sorts of things

Christewart: Ugh, i guess this is from before. We should probably offer an `Option` here on the `unmarkUTXOsAsReserved` method or something like that.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -251,22 +251,23 @@ abstract class DLCWallet
...
-      dbs <- spendingInfoDAO.findByOutPoints(inputs.map(_.outPoint))
-      // allow this to fail in the case they have already been unreserved
-      _ <- unmarkUTXOsAsReserved(dbs).recover { case _: Throwable => () }
-
       dlcOpt <- dlcDAO.read(dlcId)
-      _ = dlcOpt match {
+      dlcDb = dlcOpt match {
         case Some(db) =>
           require(
             db.state == DLCState.Offered || db.state == DLCState.Accepted || db.state == DLCState.Signed,
```
Christewart, line 258: Is `Signed` a safe state here? Doesn't this imply you have sent your signatures to your counterparty?

benthecarman: We have #3162 for this

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: Yeah, imo we should always recover from this, we want to cancel the DLC still, we can always unreserve later

PR #3259, opened 2021-06-12
Add default rounding interval

benthecarman:
Should help with performance and I think this is an acceptable default

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
@@ -412,7 +412,7 @@ class CreateDLCOfferDialog extends Logging {
                     }
                   } else {
                     // add empty rounding intervals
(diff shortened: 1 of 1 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -412,7 +412,7 @@ class CreateDLCOfferDialog extends Logging {
                     }
                   } else {
                     // add empty rounding intervals
-                    addRoundingRow(None, None)
+                    addRoundingRow(Some(0), Some(Satoshis(1000)))
```
Christewart, line 415: so i don't know much about this to be honest, but when we do add this we definitely want to make this configurable. Maybe a new configuration key `bitcoin-s.dlc`

benthecarman: The GUI doesn't have access to the config

PR #3274, opened 2021-06-16
Add ability to sweep wallet

### LINKED ISSUE #3265: Sweep all funds from GUI
Allow for an easy way to send all funds out of the wallet from GUI

benthecarman:
Closes #3265 

Adds a MAX button that will change it to sweep the wallet

image

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
@@ -278,6 +278,20 @@ trait WalletApi extends StartStopAsync[WalletApi] {
     } yield tx
   }
 
+  def sweepWallet(address: BitcoinAddress)(implicit
+      ec: ExecutionContext): Future[Transaction] = sweepWallet(address, None)
+
+  def sweepWallet(address: BitcoinAddress, feeRateOpt: Option[FeeUnit])(implicit
+      ec: ExecutionContext): Future[Transaction] = {
+    for {
+      feeRate <- determineFeeRate(feeRateOpt)
+      tx <- sweepWallet(address, feeRate)
+    } yield tx
+  }
+
+  def sweepWallet(address: BitcoinAddress, feeRate: FeeUnit)(implicit
+      ec: ExecutionContext): Future[Transaction]
+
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
```diff
@@ -278,6 +278,20 @@ trait WalletApi extends StartStopAsync[WalletApi] {
...
+
+  def sweepWallet(address: BitcoinAddress, feeRateOpt: Option[FeeUnit])(implicit
+      ec: ExecutionContext): Future[Transaction] = {
+    for {
+      feeRate <- determineFeeRate(feeRateOpt)
+      tx <- sweepWallet(address, feeRate)
+    } yield tx
+  }
+
+  def sweepWallet(address: BitcoinAddress, feeRate: FeeUnit)(implicit
```
Christewart, line 295: So.... what is the difference between `emptyWallet()` and `sweepWallet()` ? Neither of them have scaladocs on them

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -545,6 +545,14 @@ abstract class Wallet
     } yield tx
   }
 
+  override def sweepWallet(address: BitcoinAddress, feeRate: FeeUnit)(implicit
+      ec: ExecutionContext): Future[Transaction] = {
+    for {
+      outpoints <- listUtxos().map(_.map(_.outPoint))
+      tx <- sendFromOutPoints(outpoints, address, feeRate)
+    } yield tx
```
Christewart, line 557: nit: Maybe worth adding a `logger.info(s"Sweeping wallet balance=${balance} to address=${address}")`

Christewart (changes requested):
A unit test would be good too.

core/src/main/scala/org/bitcoins/core/api/wallet/WalletApi.scala
benthecarman: Oh was the same thing, removed `emptyWallet` as we already have `isEmpty` so could be confusing

Christewart (approved):

PR #3275, opened 2021-06-16
Fix rows in ViewDLCDialog

benthecarman:
Had overlapping rows, this fixes it


Christewart:
I think there was an error in the build unrelated to the PR, can you push an empty commit to this when you wake up to see if we just get a :heavy_check_mark:

benthecarman:
rebased

PR #3276, opened 2021-06-16
Update utxo states correctly for bitcoind backend

### LINKED ISSUE #3247: unconfirmed balance not getting switched to confirmed after approprate confs
Reported by a user, the utxo has 10 confs on it on the mainchain but still shows up as unconfirmed

Screenshot from 2021-06-10 12-13-46

benthecarman:
I remember @Christewart posting that Ian found this issue, cannot find the issue however

Fixes #3247

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -179,6 +179,7 @@ class BitcoinSServerMain(override val args: Array[String])(implicit
       }
       _ = BitcoindRpcBackendUtil
         .syncWalletToBitcoind(bitcoind, wallet)
+        .flatMap(_ => wallet.updateUtxoPendingStates())
         .flatMap { _ =>
           if (bitcoindRpcConf.zmqConfig == ZmqConfig.empty) {
             BitcoindRpcBackendUtil.startBitcoindBlockPolling(wallet, bitcoind)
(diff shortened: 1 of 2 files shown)


Christewart (changes requested):
So i think the better question to ask here is: 

Is there any time we want to call `wallet.processBlock()` and NOT call `updateUtxoPendingStates()`? 

If not, we should just move this method call into `wallet.processBlock()` and remove other places we have to call it manually. Thoughts?

Christewart:
Also i believe this is #3247

benthecarman:
> So i think the better question to ask here is:
> 
> Is there any time we want to call `wallet.processBlock()` and NOT call `updateUtxoPendingStates()`?
> 
> If not, we should just move this method call into `wallet.processBlock()` and remove other places we have to call it manually. Thoughts?

We only call process block on blocks that we have a filter match, this means that we wouldn't update the utxo state until we spent or received another tx. Doing it this way means we just need to actually wait for the confirmations

Christewart:
> > So i think the better question to ask here is:
> > Is there any time we want to call `wallet.processBlock()` and NOT call `updateUtxoPendingStates()`?
> > If not, we should just move this method call into `wallet.processBlock()` and remove other places we have to call it manually. Thoughts?
> 
> We only call process block on blocks that we have a filter match, this means that we wouldn't update the utxo state until we spent or received another tx. Doing it this way means we just need to actually wait for the confirmations

Hmm ok. 

So the reason love this doesn't get executed is because we never have `NodeCallbacks` added to bitcoind? 

https://github.com/bitcoin-s/bitcoin-s/blob/63083bf6425f0854053a779ce407702247b32435/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L220

We may want to think of a way to re-use the callbacks stuff between `BitcoindRpcAppConfig` and `NodeAppConfig` in the future so that we don't end up having this divergence between two bitcoin-s backends that retrieve information from the network. I don't think we necessarily want to do this on this PR, but it seems like something we should work towards in the future.

benthecarman:
> So the reason love this doesn't get executed is because we never have NodeCallbacks added to bitcoind?

Yes

Christewart (approved):

PR #3285, opened 2021-06-16
Use same network as bitcoind backend

### LINKED ISSUE #3263: Set bitcoin-s-bundle.conf to the network that bitcoind is running on if bitcoind is chosen as a backend
Bug report from a user trying to use the DLC wallet on testnet. Currently we require users to manually modify `bitcoin-s.conf` to adjust the network from mainnet to testnet. 

If it is the first time the wallet is booting up, we should just write `bitcoin-s.network = testnet3` to `bitcoin-s-bundle.conf` if the bitcoind instance is running on testnet (or regtest etc)

Slack convo: https://suredbits.slack.com/archives/CB1CB4AUX/p1623359810115300

benthecarman:
Fixes #3263

Was able to get around having to set it in the config file and instead pass it as a argument so we don't get into weird states with 2 configs having different networks

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
@@ -13,7 +13,6 @@ import scalafx.beans.property.ObjectProperty
 import scalafx.stage.Window
 
 import java.nio.file.Files
-import scala.concurrent.ExecutionContext.global
 import scala.concurrent.duration.DurationInt
 import scala.concurrent._
 import scala.jdk.CollectionConverters._
@@ -32,7 +31,7 @@ class LandingPaneModel()(implicit system: ActorSystem) extends Logging {
     taskRunner.run(
       "Launching Wallet",
       op = {
-        implicit val ec: ExecutionContextExecutor = global
+        import system.dispatcher
         val file = appConfig.baseDatadir.resolve("bitcoin-s-bundle.conf")
 
         // if the user made changes in the gui, write to file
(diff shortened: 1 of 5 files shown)


app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
```diff
@@ -30,6 +32,7 @@ class LandingPaneModel()(implicit system: ActorSystem) extends Logging {
     taskRunner.run(
       "Launching Wallet",
       op = {
+        implicit val ec: ExecutionContextExecutor = global
```
Christewart: `import system.dispatcher`

app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
```diff
@@ -31,6 +31,11 @@ case class BitcoinSAppConfig(
   lazy val chainConf: ChainAppConfig = ChainAppConfig(directory, confs: _*)
   lazy val dlcConf: DLCAppConfig = DLCAppConfig(directory, confs: _*)
 
+  def copyWithConfig(newConfs: Config*): BitcoinSAppConfig = {
```
Christewart: nit: Let's use `Vector[]` rather than `*` for new methods. I understand it is used elsewhere but it has more predictable performance and easier to reason about.

app/bundle/src/main/scala/org/bitcoins/bundle/gui/LandingPaneModel.scala
```diff
@@ -44,6 +47,19 @@ class LandingPaneModel()(implicit system: ActorSystem) extends Logging {
 
         Files.write(file, confStr.getBytes)
 
+        val extraArgsF: Future[Vector[String]] = {
+          val usedConf = appConfig.copyWithConfig(config)
+          usedConf.nodeType match {
+            case _: InternalImplementationNodeType =>
+              Future.successful(Vector.empty)
```
Christewart, line 55: Hmm... not sure if I understand this. I guess if we are connecting to an arbitrary neutrino node, we can't glean network information right away? 

If that is the case, let's at least add a comment saying that.

Christewart: I guess it does seem like we have access to app configs here, but i think the original comment still stands that we have no idea what network the remote neutrino node is on AFAICT

benthecarman: Yeah, the `node` will handle this as the different networks have different network bytes

benthecarman: Added the comment

Christewart (approved):

PR #3289, opened 2021-06-17
Multi wallet support for DLC database

### LINKED ISSUE #3241: dlcdb is wallet specific
Maybe this was clear to everyone but me, but if i use the `bitcoin-s.wallet.walletName` setting and set my wallet to something, dlcdb doesn't find txs and throws errors on startup.

benthecarman:
Fixes #3241

Needed to mess with some testkit stuff so we can have a funded `DLCWallet` so we are able to create an offer

### DIFF
--- a/dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/MultiWalletDLCTest.scala
+++ b/dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/MultiWalletDLCTest.scala
@@ -1,0 +1,64 @@
+package org.bitcoins.dlc.wallet
+
+import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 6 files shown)


Christewart:
postgres fails: https://github.com/bitcoin-s/bitcoin-s/pull/3289/checks?check_run_id=2852521013#step:5:1342

benthecarman:
> postgres fails: https://github.com/bitcoin-s/bitcoin-s/pull/3289/checks?check_run_id=2852521013#step:5:1342

Fixed, looks like we weren't fully testing postgres and had a bad migration :/

Christewart:
Ran through this with @user411 today and it seems to work for me on testnet!

Christewart (approved):

PR #3293, opened 2021-06-18
Fixed bitcoin-s-cli dir location under bin

Giamme:
It seems like the bitcoin-s-cli build is under ```./app/cli/target/universal/stage/bin/bitcoin-s-cli``` rather than ```./app/cli/target/universal/stage/bitcoin-s-cli```


Christewart:
Ty! :-)

Issue #3188, opened 2021-05-28
Add `SortedNonces` type to make sure we preserve nonce ordering


Christewart:
We keep making errors where the nonces aren't sorted. We should make a type that has the nonces sorted and has an invariant to make sure we don't lose ordering.

Christewart:
Related to #2831

Christewart:
Taken care of in #3310

PR #3310, opened 2021-06-22
Sorted and Ordered Vector Type Support

nkohen:
Added an `SortedVector[T]` and more general `SortedVec[T, B >: T]` types which wrap a `Vector[T]` with the invariant that it is sorted under an `Ordering[T]`/`Ordering[B]` (respectively).

In particular, for ordered vectors (such as a list of oracle nonce keys) one can use `SortedVec.forOrdered(nonces)` to get a `SortedVector[SchnorrNonce]` (where the order is defined by the input index).


Christewart (commented):
I believe we discussed that the problem with this approach is the implicit that now needs to be carried everywhere. In #3188 the example has a well defined ordering in the DLC spec, we don't necessarily want a general purpose data structure unless you are planning on extending these data structures for concrete implementations, which i think i'm ok with

nkohen:
> I believe we discussed that the problem with this approach is the implicit that now needs to be carried everywhere. In #3188 the example has a well defined ordering in the DLC spec, we don't necessarily want a general purpose data structure unless you are planning on extending these data structures for concrete implementations, which i think i'm ok with

I could modify this to be a trait or class instead of a case class so that it can be extended with special case classes like you say?

Christewart:
> > I believe we discussed that the problem with this approach is the implicit that now needs to be carried everywhere. In #3188 the example has a well defined ordering in the DLC spec, we don't necessarily want a general purpose data structure unless you are planning on extending these data structures for concrete implementations, which i think i'm ok with
> 
> I could modify this to be a trait or class instead of a case class so that it can be extended with special case classes like you say?

I think that would be best. `SortedNonce` can be the first implementation of it.

Christewart (approved):

PR #3321, opened 2021-06-23
Make it so table doesn't reorder on single update

benthecarman:


user411:
Testing this I still see the reordering behavior on right-click view dlc in the table, but it does open to the correct dlc details in the detail window.

I right-clicked and did View DLC on the bd6 Contract Id row, which was first in the list before taking this screen shot
Screen Shot 2021-06-23 at 10 33 43 AM

user411:
This holds ordering for me on right-click and gets the row clicked open in details. Oddly, the row that I don't right-click becomes the highlighted/focused row.

I would throw out there that I'm not sure if it makes sense to sort on event id or status as the first order sort by default when users are actually trying to manage a number of these, but the user can override the default sort order.

benthecarman:
> Oddly, the row that I don't right-click becomes the highlighted/focused row.

I think this happens because calling `table.sort()` focuses the first element

user411:
In my testing, even if I'm right-clicking the first row, the second row ends up highlighted. And and the reverse happens right clicking the second row.

user411:
That change fixed the focus issue after right-click for me.

user411 (approved):
Tests well for me

PR #3328, opened 2021-06-24
Add view on oracle explorer button

benthecarman:
image

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
@@ -1,8 +1,10 @@
 package org.bitcoins.gui.dlc
 
+import org.bitcoins.commons.jsonmodels.ExplorerEnv
 import org.bitcoins.core.dlc.accounting.RateOfReturnUtil
 import org.bitcoins.core.protocol.dlc.models.DLCStatus._
 import org.bitcoins.core.protocol.dlc.models._
+import org.bitcoins.gui.GUI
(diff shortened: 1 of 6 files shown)


user411 (approved):

Christewart:
Docs still failing: https://github.com/bitcoin-s/bitcoin-s/pull/3328/checks?check_run_id=2907855260#step:5:211

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -133,6 +135,18 @@ class DLCTableView(model: DLCPaneModel) {
         }
       }
 
+      val viewOnExplorer: MenuItem = new MenuItem("View on Oracle Explorer") {
+        onAction = _ => {
+          val status = selectionModel.value.getSelectedItem
+          val primaryOracle =
+            status.oracleInfo.singleOracleInfos.head.announcement
+          val baseUrl = ExplorerEnv.Production.siteUrl
```
Christewart, line 143: Is there any way to pattern match on the network setting? 

Oracles are network agnostic, but it seems we should probably point to the test explorer when users are on testnet3?

Christewart: This might be a nice feature for @user411 to implement. It is pretty small in scope, should be easy to view on his local wallet too. 

If we think this is a good idea, we should merge this PR, and file an issue for him to implement the testnet3 feature. Thoughts?

PR #3336, opened 2021-06-28
Silence all scaladoc warnings

### LINKED ISSUE #3232: Silence scaladoc warnings when using universal:packageBin or universal:stage
A user was asking me where these artifacts are, and because of all the scaladoc warnings, it isn't apparent where the location is.

Christewart:
fixes #3232

This silences all scaladoc warnings that are output from invalid syntax in the scaladoc. 

I hope this doesn't break other parts of the build, but we won't know for sure until we merge :crossed_fingers: 

Here is what `universal:stage` looks like when running on acbf68c

Screenshot from 2021-06-28 06-48-50

### DIFF
--- a/app/bundle/bundle.sbt
+++ b/app/bundle/bundle.sbt
@@ -26,7 +26,6 @@ assembly / assemblyMergeStrategy := {
   case _                                 => MergeStrategy.first
 }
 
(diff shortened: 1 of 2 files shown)


benthecarman (approved):
Works for me

PR #3342, opened 2021-06-28
Fix DLC not storing nSequence for funding inputs

benthecarman:
Started out trying to make it default to signaling for RBF for DLCs, ended up finding out that we assumed a sequence for every input. Fixed that and added the new default so we get better privacy benefits.

This way we will by default share the anon set with dual funded lightning txs, as they appear to be using an nSequence of `0xfffffffd`

https://mempool.space/tx/91538cbc4aca767cb77aa0690c2a6e710e095c8eb6d8f73d53a3a29682cb7581


nkohen (approved):

Christewart:
CI is being weird here, not sure why it is failing. 

Is there any implications here with the spec? I.e. if CG or AF didn't upgrade, we can still use them right?

benthecarman:
> CI is being weird here, not sure why it is failing.

Just updated description, was a bug in our code :(

> Is there any implications here with the spec? I.e. if CG or AF didn't upgrade, we can still use them right?

Spec allows for any sequence number, other clients should be compliant unless they have the same bug we did

dlc-test/src/test/scala/org/bitcoins/dlc/testgen/TestVectorGen.scala
```diff
@@ -36,6 +36,12 @@ trait TestVectorGen[T <: TestVector, Input] {
   }
 
   def readFromDefaultTestFile(): JsResult[Vector[T]] = {
+    // If we are missing the test file, generate it
+    if (!defaultTestFile.canRead) {
+      val gen = generateAndWriteTestVectors()
+      Await.result(gen, 60.seconds)
```
Christewart, line 42: Any reason to not propagate the Future?

Christewart (approved):

dlc-test/src/test/scala/org/bitcoins/dlc/testgen/TestVectorGen.scala
benthecarman: Would have been required a much larger refactor, probably better for a separate PR

PR #3348, opened 2021-06-29
Call trim on DLC message text boxes

benthecarman:
Was having issues when copy-pasting messages


Christewart:
Is there an obvious way to manually test this?

benthecarman:
> Is there an obvious way to manually test this?

When pasting in a contract info or announcement, add spaces to the beginning and end of what you copy

Christewart:
> > Is there an obvious way to manually test this?
> 
> When pasting in a contract info or announcement, add spaces to the beginning and end of what you copy

@user411 Can you run through this and confirm this patch works?

user411 (approved):
Text box didn't seem to want to take Tabs from the keyboard, but everything went fine with extra spaces around Offer hex data. Worked fine forcing tabs into string and using that

PR #3352, opened 2021-06-29
Use OrderedNonces type in DLC data types

### LINKED ISSUE #3330: Actually use `OrderedNonce` in places that it is needed
Now that we have merged #3310 merged, we need to actually use it! I think we should search the codebase for `Vector[SchnorrNonce]` and replace it where approprpiate with `OrderedNonce`?

benthecarman:
Closes #3330 

Uses `OrderedNonces` in the DLC data types like the announcement, then implements it in the `DLCWallet` and `DLCOracle`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
@@ -5,6 +5,7 @@ import org.bitcoins.core.number.UInt32
 import org.bitcoins.core.protocol.dlc.compute.SigningVersion
 import org.bitcoins.core.protocol.tlv._
 import org.bitcoins.core.util.NumberUtil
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/bitcoins/core/api/dlcoracle/OracleEvent.scala
```diff
@@ -60,7 +61,7 @@ sealed trait CompletedOracleEvent extends OracleEvent {
           "Must have a signature for every nonce")
 
   def signatures: Vector[SchnorrDigitalSignature] =
-    nonces
+    nonces.vec
```
Christewart, line 64: nit: I would add a `toVector` method on OrderedNocnes rather than using `vec`. But it isn't a big deal.

Christewart (approved):
LGTM, but would like @nkohen 's eyes on this too to make sure we are getting everything.

PR #3353, opened 2021-06-30
Fix showing error popups in GUI

benthecarman:
The popup was defined as `lazy val _ = ..` so it would never been shown


Christewart:
Any way to easily test?

benthecarman:
> Any way to easily test?

Just add a `throw new RuntimeExecption` somewhere, to test I put one in the `GetNetAddressDialog`

user411 (approved):

PR #3356, opened 2021-06-30
Create and implement OrderedAnnouncements type

benthecarman:


core/src/main/scala/org/bitcoins/core/util/sorted/OrderedAnnouncements.scala
```diff
@@ -0,0 +1,15 @@
+package org.bitcoins.core.util.sorted
+
+import org.bitcoins.core.protocol.tlv._
+
+/** Represents an ordered set of OracleAnnouncementTLVs */
```
nkohen: probably worth mentioning that the order represents preference

core/src/main/scala/org/bitcoins/core/util/sorted/OrderedAnnouncements.scala
```diff
@@ -0,0 +1,15 @@
...
+case class OrderedAnnouncements(vec: Vector[OracleAnnouncementTLV])
+    extends SortedVec[OracleAnnouncementTLV, OracleAnnouncementTLV](
+      vec,
+      SortedVec.forOrdered(vec))
+
+/** Represents an ordered set of OracleAnnouncementV0TLV */
```
nkohen: ditto

Christewart (approved):

PR #3365, opened 2021-07-03
Implement BIP 86, HDPath for Taproot

benthecarman:
https://github.com/bitcoin/bips/blob/master/bip-0086.mediawiki

Need #3366 until we can add the output key and need to implement basic taproot spks until we can implement the spk and address checks


Christewart:
Superseded by #5768

Issue #3363, opened 2021-07-01
Windows libsecp link errors


benthecarman:
From a user:

```
I ran into issue running bitcoin-s-bundle.bat

12:25:35,266 |-INFO in ch.qos.logback.classic.joran.JoranConfigurator@121ea7a - Registering current configuration as safe fallback point

UnsatisfiedLinkError: http://java.io.IOException: Couldn't load library library libsecp256k1-0
Exception in Application start method
```

user411:
I was seeing the same

[info] UnsatisfiedLinkError: java.io.IOException: Couldn't load library library secp256k1

 running local on the work Mac with adopt:11 OpenJDK installed via 

cs java --jvm adopt:11 --setup

Changing to zulu:11 removed the warning.

I also get the same warning doing a build like

bundle/universal:packageBin

on my work Mac and trying to move it to another Mac when running the output (with either JVM used at build time).

benthecarman:
The user that reported this to me was able to resolve by updating their java from 8 to 16

Christewart:
Pulled https://github.com/bitcoin-s/bitcoin-s/issues/3074 down to 0.7, hopefully we can get to the bottom of all these issues

Issue #3302, opened 2021-06-21
secp256k1 library not building / available by default


user411:
**Describe the bug**
secp256k1 library does not appear to be present, but appears to be required

**To Reproduce**
Checkout bitcoin-s repo
sbt bundle/run
In either Neutrino Config or Bitcoin RPC Config tabs, click Launch Wallet
Notice log line [info] UnsatisfiedLinkError: java.io.IOException: Couldn't load library library secp256k1
Connection fails and ultimately times out

**Expected behavior**
secp256k1 library would be available through build process

**Logs**
[info] UnsatisfiedLinkError: java.io.IOException: Couldn't load library library secp256k1
...
[info] 2021-06-21T14:54:58UTC INFO [PeerMessageReceiver] Connection established with peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333)
[info] 2021-06-21T14:54:58UTC INFO [P2PClientActor] We've been disconnected by Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) command=PeerClosed
[info] 2021-06-21T14:54:58UTC ERROR [OneForOneStrategy] Cannot disconnect from peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) when in state=Initializing
[info] java.lang.RuntimeException: Cannot disconnect from peer=Peer(neutrino.testnet3.suredbits.com/34.210.154.40:18333) when in state=Initializing

**Platform (please complete the following information):**
 - OS: macOS
 - Version: top of tree

**Additional context**

benthecarman:
How was this resolved?

Christewart:
> How was this resolved?

So neither of us seem to be able to reproduce. @user411 was able to do a few DLCs with me this morning before with no issues and i didn't have any problems on my mac either. 

Did a user you are helping end up having a problem?

benthecarman:
> Did a user you are helping end up having a problem

No I just saw you guys did a DLC so I figured you got it working

user411:
Not sure how anything resolved, but the message doesn't seem to be in the logs today. II have built and run several more packages inside the project by now, but didn't consciously do anything to solve the issue.

Christewart:
I'm going to close this as we can't seem to reproduce on my mac or Ivans.

user411:
[...]

PR #3367, opened 2021-07-03
Add secp256k1jni tests to Mac & Windows CI

benthecarman:


Christewart (approved):
LGTM, assuming CI. I'm hoping this catches the behavior that @user411, #3363 and #3302 documents

PR #3368, opened 2021-07-05
Set sync height on new wallet

benthecarman:
This would cause an error where if the user received their first funds while offline they would not see it because it would sync above the first tx.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -38,7 +38,12 @@ object BitcoindRpcBackendUtil extends Logging {
             txDbs <- wallet.listTransactions()
             lastConfirmedOpt = txDbs.filter(_.blockHashOpt.isDefined).lastOption
             _ <- lastConfirmedOpt match {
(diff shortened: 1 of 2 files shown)


Christewart (commented):
I'm not sure if I understand the bug here. Is there also a unit test we could write to make sure we don't regress in the future?

benthecarman:
> I'm not sure if I understand the bug here.

We would only set the sync height after processing our first block. If a user just generated an address, sent to it, and then closed bitcoin-s. The first block their processed would be after their funds got confirmed so they would never see the funds they sent and need to do a rescan.

>  Is there also a unit test we could write to make sure we don't regress in the future?

We can't really make a unit test because this only effects wallets that are offline.

Christewart:
> > I'm not sure if I understand the bug here.
> 
> We would only set the sync height after processing our first block. If a user just generated an address, sent to it, and then closed bitcoin-s. The first block their processed would be after their funds got confirmed so they would never see the funds they sent and need to do a rescan.
> 
> > Is there also a unit test we could write to make sure we don't regress in the future?
> 
> We can't really make a unit test because this only effects wallets that are offline.

Hmm ok. I think this is likely something Ian ran into as a bug when he was testing out a DLC.

PR #3370, opened 2021-07-06
Export Result and Save Backup handle no file selection

user411:
Basic Bugfix PR from working with things today

Handle null (no file) selection from FileChooser save dialogs for Save Backup menu item and Export Result DLCPane button.

Screen Shot 2021-07-06 at 1 52 36 PM

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
@@ -35,8 +35,9 @@ private class FileMenu() {
         initialFileName = "bitcoin-s-backup.zip"
       }
       val chosenFile = fileChooser.showSaveDialog(null)
-      ConsoleCli.exec(ZipDataDir(chosenFile.toPath),
-                      GlobalData.consoleCliConfig)
+      if (chosenFile != null)
+        ConsoleCli.exec(ZipDataDir(chosenFile.toPath),
+                        GlobalData.consoleCliConfig)
       ()
     }
   }
--- a/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
@@ -37,7 +37,9 @@ private class FileMenu() {
       val chosenFileOpt = Option(fileChooser.showSaveDialog(null))
       chosenFileOpt match {
         case Some(chosenFile) =>
-          ConsoleCli.exec(ZipDataDir(chosenFile.toPath),GlobalData.consoleCliConfig)
+          ConsoleCli.exec(ZipDataDir(chosenFile.toPath),
+                          GlobalData.consoleCliConfig)
+          ()
         case None => // User canceled in dialog
       }
     }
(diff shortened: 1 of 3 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -369,7 +369,7 @@ class CreateDLCOfferDialog extends Logging {
           decompOpt = Some(digitDecomp)
 
           val addPointButton: Button = new Button("+") {
-            onAction = _ => addPointRow()
+            onAction = _ => addPointRow(isEndPoint = false)
```
benthecarman: These can be endpoints

Endpoint doesn't refer to at then end of the list but rather, if it is an endpoint in the payout curve

benthecarman:
otherwise LGTM

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
user411: Can a payout curve have more than 2 endpoints, or is 2 the limit?

If I wanted to do a payout for values [10, 100] with only 100 paying me out, would I set up the points like:

Screen Shot 2021-07-06 at 2 08 45 PM
?

user411: Ok, seeing more of what 'Endpoint' impacts. I'll remove the value change.
Screen Shot 2021-07-06 at 3 10 58 PM

benthecarman (approved):

app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
```diff
@@ -35,8 +35,9 @@ private class FileMenu() {
         initialFileName = "bitcoin-s-backup.zip"
       }
       val chosenFile = fileChooser.showSaveDialog(null)
-      ConsoleCli.exec(ZipDataDir(chosenFile.toPath),
-                      GlobalData.consoleCliConfig)
+      if (chosenFile != null)
```
Christewart: So the best way to handle this in Scala is wrapping the potentially null thing in `Option`. This gives us types rather than `null` which circumvents the entire type system. Unfortunately other programming languages like Java has decided to use `null` rather than types to represent something that is not available.

```scala
val chosenFileOpt = Option(fileChoose.showSaveDialog(null))
chosenFileOpt match {
  case Some(choseFilen) => 
      ConsoleCli.exec(ZipDataDir(chosenFile.toPath),GlobalData.consoleCliConfig)
  case None => 
     //should we warn or something in this case?
}
```

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPane.scala
```diff
@@ -125,7 +125,8 @@ class DLCPane(glassPane: VBox)(implicit ec: ExecutionContext) {
         initialDirectory = new File(Properties.userHome)
       }
       val chosenFile = fileChooser.showSaveDialog(null)
-      Files.write(chosenFile.toPath, resultTextArea.text.value.getBytes)
+      if (chosenFile != null)
```
Christewart: same here, let's use the `Option` pattern.

Christewart (changes requested):

app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
benthecarman: there's no need to log, this will happen if the user hit cancel instead of selecting a file

Christewart (approved):

PR #3371, opened 2021-07-07
Try using adoptopen jdk to reproduce secp256k1jni errors

Christewart:
See if we can reproduce #3363 on CI with adoptopenjdk


Christewart:
Can't seem to reproduce this with openjdk/adoptjdk on 11 and 8

PR #3373, opened 2021-07-07
1.7.0 version of website

Christewart:
Followed instructions here to generate the website 

https://bitcoin-s.org/docs/contributing-website


user411:
When I try to build the website locally to run I get stuck on

[info] started sbt server
sbt:bitcoin-s> unidoc
[info] Main Scala API documentation to /Users/ivan/code/bitcoin-s/target/scala-2.13/unidoc...
[error] /Users/ivan/code/bitcoin-s/async-utils/src/main/scala/org/bitcoins/asyncutil/AsyncUtil.scala:10:16: AsyncUtil is already defined as class AsyncUtil
[error] abstract class AsyncUtil extends AsyncUtilApi {
[error]                ^
[error] /Users/ivan/code/bitcoin-s/async-utils/src/main/scala/org/bitcoins/asyncutil/AsyncUtil.scala:157:8: AsyncUtil is already defined as object AsyncUtil
[error] object AsyncUtil extends AsyncUtil {
[error]        ^
[error] /Users/ivan/code/bitcoin-s/core/src/main/scala-2.13/org/bitcoins/core/compat/CompactPackage.scala:10:16: package is already defined as package object compat
[error] package object compat {
// ...

Christewart:
Try skipping the `unidoc` step. That is for generating scaladocs. I believe that `docs/mdoc` command turns `.md` files -> `.html`. You should be able to `yarn install` and `yarn start` after that.

PR #3377, opened 2021-07-07
Use Wallet's transactionDAO instead of creating a copy

benthecarman:
`Wallet` has its own `transactionDAO` so we don't need `DLCWallet` to create it own instaniation.


Christewart (approved):

Christewart:
Compile: https://github.com/bitcoin-s/bitcoin-s/pull/3377/checks?check_run_id=3012807614#step:5:368

PR #3381, opened 2021-07-08
Add connection indicator to GUI

benthecarman:
Can test by using the server and gui separately, and turning off the server after you have the gui started

image


user411 (approved):
Runs well disconnecting and reconnecting locally to server down/up. This was a good thread to see run through the backend -> UI Model flow and state handling

Christewart (approved):

PR #3383, opened 2021-07-08
Add ability to label DLCs

### LINKED ISSUE #3374: Add labeling to DLCs
Would be nice to add an field to the dlc table for a label for the DLC, so a user can put something like "Price bet with Chris" or "Superbowl bet with John", etc

benthecarman:
Closes #3374

Adds the ability to add a label to a DLC so the user can better identify what the contract is.


benthecarman:
todo: fix postgres migrations

PR #3384, opened 2021-07-08
Update user agent

benthecarman:
Looks like we missed this last time


Christewart (approved):
Can you update this too? Ideally we could have them use the same field, but not sure if that is possible :/

https://github.com/bitcoin-s/bitcoin-s/blob/cb5ec20eacff655d7ec3416b674edb4f8c461adf/project/CommonSettings.scala#L24

PR #3394, opened 2021-07-12
Allow accepting multi oracle announcements

benthecarman:
Technially we allowed for creating multi oracle offers if you input a contract info so we should do the same for the accept flow

https://github.com/bitcoin-s/bitcoin-s/blob/253f7444c6e61131bd840b48d6e901646ce1e2eb/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala#L324

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
@@ -53,8 +53,10 @@ class AcceptOfferDialog {
       val (oracleKey, eventId) = offer.contractInfo.oracleInfo match {
         case OracleInfoV0TLV(announcement) =>
           (announcement.publicKey.hex, announcement.eventTLV.eventId)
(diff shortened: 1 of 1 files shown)


user411 (approved):

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
```diff
@@ -53,8 +53,10 @@ class AcceptOfferDialog {
       val (oracleKey, eventId) = offer.contractInfo.oracleInfo match {
         case OracleInfoV0TLV(announcement) =>
           (announcement.publicKey.hex, announcement.eventTLV.eventId)
-        case _: MultiOracleInfoTLV =>
-          throw new RuntimeException("This is impossible.")
+        case multi: MultiOracleInfoTLV =>
```
Christewart, line 56: is there really much of an improvement to display 1 rather than throwing an exception? I think this is misleading and will likely be harder for us to debug/trackdown later. I prefer the loud exception until we have end to end support

PR #3396, opened 2021-07-12
SOCKS5 client transport

rorp:


tor-test/src/test/scala/org/bitcoins/tor/Socks5ClientTransportSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+class Socks5ClientTransportSpec
+    extends TestKit(ActorSystem("test"))
+    with TestSuite
+    with BeforeAndAfterAll
+    with AnyFunSuiteLike
+    with ImplicitSender {
+
+  implicit val ec = system.dispatcher
+
+  ignore("connect to real daemon") {
```
Christewart: This works locally I presume?

tor-test/src/test/scala/org/bitcoins/tor/Socks5ClientTransportSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+import akka.util.ByteString
+import org.scalatest.funsuite.AnyFunSuiteLike
+import org.scalatest.{BeforeAndAfterAll, TestSuite}
+
+import java.net.InetSocketAddress
+import scala.concurrent.Await
+import scala.concurrent.duration.DurationInt
+
+class Socks5ClientTransportSpec
+    extends TestKit(ActorSystem("test"))
```
Christewart: It would be nice if we could figure out a way to get this all to hook into our existing test harnesses (`BitcoinSAsyncTest` or `BitcoinSUnitTest`). Likely we do need a new trait, but if we start doing more actor/stream programming, we will want a common test trait

tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala
```diff
@@ -36,73 +36,33 @@ class Socks5Connection(
...
     }
   }
 
   def authenticate: Receive = { case Tcp.Received(data) =>
-    if (data(0) != 0x01) {
-      throw Socks5Error("Invalid SOCKS5 proxy response")
-    } else if (data(1) != 0) {
-      throw Socks5Error("SOCKS5 authentication failed")
-    }
+    parseAuth(data)
```
Christewart: Is this just being dropped on the floor? It would be nice to do something with the return type even if it is just throwing an error. 

For instance, if we change the behavior of `parseAuth()` in the future to not throw exceptions, we will never get a compiler error. I also don't think we have a unit test for this behavior?

tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala
```diff
@@ -213,6 +173,59 @@ object Socks5Connection {
...
+          val end = start + len
+          val domain = data.slice(start, end).utf8String
+          val port = data(end).toInt << 8 | data(end + 1)
+          new InetSocketAddress(domain, port)
+        case 0x04 =>
+          val ip = Array.ofDim[Byte](16)
+          data.copyToArray(ip, 4, 4 + ip.length)
+          val port = data(4 + ip.length).toInt << 8 | data(4 + ip.length + 1)
+          new InetSocketAddress(InetAddress.getByAddress(ip), port)
+        case _ => throw Socks5Error(s"Unrecognized address type")
```
Christewart: nit: Add the byte to the error message for easier debugging.

```scala
case b => throw Socks5Error(s"Unrecognized address type=$b")
```

tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala
```diff
@@ -213,6 +173,59 @@ object Socks5Connection {
 
   def portToByteString(port: Int): ByteString =
     ByteString((port & 0x0000ff00) >> 8, port & 0x000000ff)
+
+  def parseGreetings(data: ByteString, passwordAuth: Boolean): Byte = {
```
Christewart, line 183: Do you have any references to the SOCKS5 specification you could link a scaladoc for future users of the code? I'm not sure where these magic bytes are coming from.

tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala
Christewart: Even a class level scaladoc on `class Socks5Connection` would be very helpful.

tor/src/main/scala/org/bitcoins/tor/Socks5ClientTransport.scala
```diff
@@ -0,0 +1,167 @@
...
+
+      def sendConnect(): Unit = {
+        state = Connecting
+        send(connectMessage)
+      }
+
+      def parseResponse(data: ByteString): Unit = {
+        state match {
+          case Greeting =>
+            Try(parseGreetings(data, credentialsOpt.nonEmpty)) match {
```
Christewart: maybe just worth returning a `Try[]` from `parseGreetings`?

Christewart (changes requested):

Christewart (approved):

Christewart:
@benthecarman does this look to you?

tor-test/src/test/scala/org/bitcoins/tor/Socks5ClientTransportSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+  val socks5ClientTransport = new Socks5ClientTransport(proxyParams)
+
+  val clientConnectionSettings =
+    ClientConnectionSettings(system).withTransport(socks5ClientTransport)
+
+  val settings = ConnectionPoolSettings(system).withConnectionSettings(
+    clientConnectionSettings)
+
+  it should "handle clear net addresses" in {
+    assume(torEnabled, "TOR environment variable is not set")
```
benthecarman, line 37: Why `assume` and not `assert`?

tor-test/src/test/scala/org/bitcoins/tor/Socks5ClientTransportSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+          .contains("https://blockstream.info/"))
+    }
+  }
+
+  it should "handle onion addresses" in {
+    assume(torEnabled, "TOR environment variable is not set")
+    for {
+      response <- Http().singleRequest(
+        HttpRequest(uri =
+          "http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/"),
```
benthecarman, line 58: maybe worth eventually switching this to our own sites

benthecarman (approved):

PR #3402, opened 2021-07-13
DLC P2P Client

benthecarman:
Still working on tests and would like to move to a separate module, but putting this out for now so I can get initial feedback.


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/networking/P2PClient.scala
```diff
@@ -0,0 +1,430 @@
...
+            val msg = LnMessageReceived(m, P2PClient(self, peer))
+            peerMsgRecv.handleLnMessageReceived(msg)
+        }
+
+        logger.trace(s"About to process ${messages.length} messages")
+        val newMsgReceiverF =
+          FutureUtil.foldLeftAsync(currentPeerMsgHandlerRecv, messages)(f)(
+            context.dispatcher)
+
+        val newMsgReceiver = Await.result(newMsgReceiverF, timeout)
```
rorp: That's the kind of legacy I'd like to get rid of.
```scala

def asynchronoslyProcessData(data: ByteString): Future[DataProcessed] = ???

def receive: Receive = {
  case Tcp.Received(data) => asynchronoslyProcessData(data).pipeTo(self)
  case DataProcessed(newMsgReceiver) => currentPeerMsgHandlerRecv = newMsgReceiver
}
```

rorp: Also it makes sense to make all those receive methods more granular. 
Each step could have it's own `Receive` handler like:

```scala
def sendOffer: Receive = {
  case Send(offer) => 
     sendOffer(offer)
     context become awaitOfferAcceptance
}

def awaitOfferAcceptance: Receive = {
  case OfferAccepted(acc) => 
    processAcc(acc)
    context become nextStep
}

def nextStep: Receive = ???
```

rorp: That could make the whole cleaner code easier to understand.

rorp (dismissed):

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/networking/P2PClient.scala
rorp: Also, it look like the the protocol handler is not linear step-by-step one, so it would more appropriate to use Akka's own FSM capabilities: https://doc.akka.io/docs/akka/current/typed/fsm.html

Christewart: Concept ACK on @rorp 's suggestions

Christewart:
Looks like we have a bug here: https://github.com/bitcoin-s/bitcoin-s/pull/3402/checks?check_run_id=3165469042#step:5:3770

db-commons/src/main/resources/reference.conf
```diff
@@ -140,6 +154,10 @@ bitcoin-s {
     # }
   }
 
+  dlcnode {
+    listen = "0.0.0.0:2862"
```
Christewart, line 159: does this need to be `0.0.0.0` ? Could you add a little comment on top of this config setting to indicate what it is for

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCConnectionHandlerTest.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.bitcoins.testkitcore.gen.LnMessageGen
+
+class DLCConnectionHandlerTest
+    extends BitcoinSAsyncTest
+    with CachedBitcoinSAppConfig {
+
+  behavior of "parseIndividualMessages"
+
+  it must "DLC Accept message that is not aligned with a tcp frame" in {
+    val accept: LnMessage[DLCAcceptTLV] =
```
Christewart: could you just use a `forAll()` here?

Christewart: same with the next 2 messages

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTest.scala
```diff
@@ -0,0 +1,231 @@
...
+import scodec.bits.ByteVector
+
+import java.io.File
+import java.net.{InetSocketAddress, Socket}
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, Promise}
+import scala.util.Try
+
+class DLCServerTest
+    extends TestKit(ActorSystem("test"))
```
Christewart: you probably want to use different  names for the actor system (`"test"`) as I'm fairly certain they are required to be unique. If suites get run in parallel you may have unintended consequences.

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTest.scala
```diff
@@ -0,0 +1,231 @@
...
+      InetSocketAddress.createUnresolved("localhost", port)
+
+    var serverConnectionHandlerOpt = Option.empty[ActorRef]
+    val serverProbe = TestProbe()
+
+    val boundAddressPromise = Promise[InetSocketAddress]()
+
+    TestActorRef(
+      DLCServer.props(
+        null, // todo remove nulls
```
Christewart: ping

Christewart (commented):
So I took a crack at integrating the new actor based tests into our current test harness. I created `BitcoinSActorTest` and `BitcoinSActorFixtureWithDLCWallet`.

For the latter test trait, i removed the `null` you previously had in `DLCServerTest` to use a real `dlcWallet`.

Unfortunately i can't seem to figure out how to get this line to pass 

>    assume(torControlEnabled,
           "Tor daemon is not running or listening port 9051")

on the test traits

https://github.com/Christewart/bitcoin-s-core/tree/2021-07-26-dlc-node-code-review

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTest.scala
Christewart: ignore this, fixed on my branch: https://github.com/Christewart/bitcoin-s-core/commit/e7becfa287004f12f54a1396738f4674c470b5c9#diff-3986e60b1bff3f11f307f8b0547f111c7121b911643bb0629ab51be44355be61R49

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTest.scala
Christewart: Ignore this, I created a new test trait to take care of this: https://github.com/Christewart/bitcoin-s-core/commit/e7becfa287004f12f54a1396738f4674c470b5c9#diff-bb3469450cb4d6b3233db5de3d61b90d2863b43d7b521a8eea6fa6a4fd4bfbd2R10

db-commons/src/main/resources/reference.conf
Christewart: also i think this needs to go out to our example configuration file too

rorp:
> Unfortunately i can't seem to figure out how to get this line to pass
> 
> > assume(torControlEnabled,
> > "Tor daemon is not running or listening port 9051")

Uncomment
```
ControlPort 9051
``` 
and
```
CookieAuthentication 1
```
in `torrc` file.

db-commons/src/main/resources/reference.conf
```diff
@@ -57,6 +57,20 @@ bitcoin-s {
     port = 9050
   }
 
+  tor {
+    # You can enable Tor for incoming connections
+    enabled = false
+    control = "127.0.0.1:9051"
```
rorp, line 62: I think we might want to have the same `host:port` format for proxy as well

dlc-node/src/main/scala/org/bitcoins/dlc/node/config/DLCNodeAppConfig.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  lazy val torParams: Option[TorParams] = {
+    if (config.getBoolean("bitcoin-s.tor.enabled")) {
+      val controlURI = new URI(config.getString("bitcoin-s.tor.control"))
+      val control = InetSocketAddress.createUnresolved(controlURI.getHost,
+                                                       controlURI.getPort)
+
+      val auth = config.getStringOrNone("bitcoin-s.tor.password") match {
+        case Some(pass) => Password(pass)
+        case None       => SafeCookie() // todo allow configuring the cookie?
```
Christewart, line 69: probably worth an issue with a good first issue tag?

dlc-node/src/main/scala/org/bitcoins/dlc/node/config/TorParams.scala
```diff
@@ -0,0 +1,12 @@
+package org.bitcoins.dlc.node.config
+
+import org.bitcoins.tor.TorProtocolHandler.Authentication
+
+import java.net.InetSocketAddress
+import java.nio.file.Path
+
+case class TorParams(
```
Christewart: nit: Maybe move to tor module? Would this be useful in other contexts?

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCConnectionHandler.scala
```diff
@@ -0,0 +1,154 @@
...
+  private[bitcoins] def parseIndividualMessages(
+      bytes: ByteVector): (Vector[LnMessage[TLV]], ByteVector) = {
+    @tailrec
+    def loop(
+        remainingBytes: ByteVector,
+        accum: Vector[LnMessage[TLV]]): (Vector[LnMessage[TLV]], ByteVector) = {
+      if (remainingBytes.length <= 0) {
+        (accum, remainingBytes)
+      } else {
+        // todo figure out how to properly handle unknown messages
```
Christewart, line 137: worth an issue?

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCDataHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  override def receive: Receive = LoggingReceive {
+    case lnMessage: LnMessage[TLV] =>
+      log.info(s"Received LnMessage ${lnMessage.typeName}")
+      lnMessage.tlv match {
+        case msg @ (_: UnknownTLV | _: DLCOracleTLV | _: DLCSetupPieceTLV) =>
+          log.error(s"Received unhandled message $msg")
+        case _: InitTLV =>
+          () // todo init logic
```
Christewart: worth an issue?

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCDataHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+        case error: ErrorTLV =>
+          log.error(error.toString)
+        case ping: PingTLV =>
+          val pong = PongTLV.forIgnored(ping.ignored)
+          connectionHandler ! LnMessage(pong)
+        case pong: PongTLV =>
+          log.debug(s"Received pong message $pong")
+        case dlcOffer: DLCOfferTLV =>
+          val _ = for {
```
Christewart: so it isn't ideal that we are dropping all these `Future` on the floor. We have no way of logging an error that occurs from these. 

i've written a patch to improve this and add a log

https://github.com/Christewart/bitcoin-s-core/commit/281e6c2f1e2ac5c53db92552805f4eba1a074323

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
```diff
@@ -0,0 +1,60 @@
...
+      config.torParams
+    )
+  }
+
+  override def start(): Future[Unit] = {
+    serverBindF.map(_ => ())
+  }
+
+  override def stop(): Future[Unit] = {
+    Future.unit
```
Christewart: Probably should have some logic to kill the binding? If we don't want to address on this PR, let's file an issue so we remember to do it in the future

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCServer.scala
```diff
@@ -0,0 +1,98 @@
...
+      log.info(s"Received a connection from $remoteAddress")
+      val _ = context.actorOf(
+        Props(
+          new DLCConnectionHandler(dlcWalletApi,
+                                   connection,
+                                   None,
+                                   dataHandlerFactory)))
+  }
+
+  override def aroundReceive(receive: Receive, msg: Any): Unit = try {
```
Christewart, line 57: This is an internal API to akka. Is there a need for overriding this? 

https://github.com/akka/akka/blob/f3fb2a577d804fb69e6b1ff2fd8f3774627ad81f/akka-actor/src/main/scala/akka/actor/Actor.scala#L535

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTorTest.scala
```diff
@@ -0,0 +1,163 @@
...
+            succeed
+          }
+        }
+        resultF.flatten
+    }
+
+    resultF
+  }
+
+  private def portIsBound(address: InetSocketAddress): Boolean =
```
Christewart: nit: This seems like a really useful utility, i'm surprised we don't have it anywhere else yet. Any thoughts on where to put it so we don't have to re-implement everytime?

db-commons/src/main/resources/reference.conf
Christewart: :+1:

Christewart (changes requested):

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCDataHandler.scala
Christewart: I've also got a patch for thread pool leaks: https://github.com/Christewart/bitcoin-s-core/tree/2021-07-30-dlc-threadpool-leaks

Send me a DM on slack if you want me to open these as PRs on your repo, otherwise they should be easy enough to cherry pick

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
benthecarman: I am unsure how to do this, maybe @rorp does

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCServer.scala
benthecarman: this was from @rorp's code, maybe he can answer

rorp: If we want `try/catch` in all `Receive` methods then there's no need for overriding it.

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
rorp: You can just stop the actor.

rorp: Or better, send `Tcp.Unbind`

benthecarman: Done

db-commons/src/main/resources/reference.conf
```diff
@@ -53,8 +53,21 @@ bitcoin-s {
   # You can configure SOCKS5 proxy to use Tor for outgoing connections
   proxy {
     enabled = false
-    host = "127.0.0.1"
-    port = 9050
+    sock5 = "127.0.0.1:9050"
```
rorp: It should be kind of plural SOCKS5.

benthecarman: fixed

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
```diff
@@ -35,7 +35,9 @@ case class DLCNode(wallet: DLCWalletApi)(implicit
   }
 
   override def stop(): Future[Unit] = {
-    Future.unit
+    serverBindF.map { case (_, actorRef) =>
+      system.stop(actorRef)
```
Christewart, line 39: So i'm not sure if this actually unbinds the underlying port? Just because you stop the actor doesn't mean all resources allocated to that actor are freed. 

Here is how to handle shutdowns correctly inside of the actor: https://doc.akka.io/docs/akka/current/typed/actor-lifecycle.html#stopping-actors

I believe what @rorp said is what you need to send to the underlying `IO(TCP)` actor 

https://github.com/bitcoin-s/bitcoin-s/pull/3402#discussion_r681166523

https://github.com/akka/akka/blob/e127d1916f772de70bd6c144a9e04e1821e3b834/akka-actor/src/main/scala/akka/io/Tcp.scala#L181

rorp:
```scala
var socket = _
def receive = {
  case Tcp.Bound(localAddress) =>
    socket = sender()
  case DLCServer.Disconnect =>
    socket ! Tcp.Unbind
  case Tcp.Unbound =>
    context.stop(self)
}
```

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCServer.scala
```diff
@@ -49,6 +49,11 @@ class DLCServer(
                                    dataHandlerFactory)))
   }
 
+  override def postStop(): Unit = {
+    super.postStop()
+    socket ! DLCServer.Disconnect
```
Christewart: So the `socket` will not understand `DLCServer.Disconnect`. You need to send `Tcp.Unbind` here I believe as this is the underlying IO actor that akka manages itself. Only `DLCServer` knows what to do with the `DLCServer.Disconnect` method

Christewart (approved):

PR #3409, opened 2021-07-15
Add network selection combobox for Neutrino

### LINKED ISSUE #3406: Allow user to pick the network from the startup GUI
I'm running into some issues with a non technical user that is trying to start his node on testnet3. He is trying to follow these instructions: 

https://bitcoin-s.org/docs/getting-setup#step-5-optional-moving-to-testnet

There is a couple of problems with this section, namely it doesn't use fully qualified path names for config (`network` rather than `bitcoin-s.network`) 

As an aside, we should probably allow the user to pick the network they want to connect on when they startup the GUI. I.e. 

- Mainnet 
- testnet3
- regtest
- signet

With the default being mainnet.

### LINKED ISSUE #3410: If given neutrino peer in GUI is a onion address the tor proxy should be enabled
Right now if you enter an onion address into the neutrino config GUI it will only work if you have `bitcoin-s.proxy.enabled = true` but this will need to be manually set, we should set this in the config automatically if the given address contains `.onion`

Should be noted that if the address _doesn't_ contain `.onion` that we shouldn't set the proxy to disabled, we should only enable it for the user.

user411:
Screen Shot 2021-07-15 at 1 38 22 PM

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
@@ -4,7 +4,6 @@ import com.typesafe.config.{Config, ConfigFactory}
 import org.bitcoins.core.config._
 import org.bitcoins.server.BitcoinSAppConfig
 import org.bitcoins.server.BitcoinSAppConfig.toNodeConf
-import scalafx.collections.ObservableBuffer
 import scalafx.geometry._
 import scalafx.scene.Node
 import scalafx.scene.control._
@@ -49,23 +48,21 @@ class NeutrinoConfigPane(
     }
   }
 
-  private val networkList = new ObservableBuffer[String]()
-
-  private val networkComboBox: ComboBox[String] = new ComboBox[String]() {
-    networkList ++= BitcoinNetworks.knownNetworks.map(_.toString)
-    items = networkList
-
-    val networkName =
-      BitcoinNetworks.fromString(appConfig.chainConf.network.name).toString
-    selectionModel.value.select(networkName)
-    onAction = _ => {
-      if (peerAddressTF.text.value.contains(".suredbits.com")) { // Or should this use the peers.headOption logic?
-        val selection = selectionModel().getSelectedItem
-        val network = BitcoinNetworks.fromString(selection)
-        peerAddressTF.text.value = defaultPeerForNetwork(network)
+  private val networkComboBox: ComboBox[NetworkParameters] =
+    new ComboBox[NetworkParameters](BitcoinNetworks.knownNetworks) {
+      value = BitcoinNetworks.fromString(appConfig.chainConf.network.name)
+      onAction = _ => {
+        val peer = peerAddressTF.text.value
+        if (
+          peer.contains(".suredbits.com") || peer.contains("localhost") || peer
+            .contains("127.0.0.1")
+        ) {
+          val network = selectionModel().getSelectedItem
+          peerAddressTF.text.value = defaultPeerForNetwork(
+            network.asInstanceOf[BitcoinNetwork])
+        }
       }
     }
-  }
 
   private val peerAddressTF: TextField = new TextField() {
     text = startingPeerAddress
@@ -100,13 +97,18 @@ class NeutrinoConfigPane(
   }
 
   def getConfig: Config = {
-    val configStr =
+    var configStr =
       s"""
          |bitcoin-s.network = ${networkComboBox.value.value}
          |bitcoin-s.node.mode = neutrino
          |bitcoin-s.node.peers = ["${peerAddressTF.text.value}"]
          |""".stripMargin
-
+    // Auto-enable proxy for .onion peers
+    if (peerAddressTF.text.value.contains(".onion")) {
+      configStr = configStr + s"""
+                                 |bitcoin-s.proxy.enabled = true
+                                 |""".stripMargin
+    }
     ConfigFactory.parseString(configStr)
   }
 }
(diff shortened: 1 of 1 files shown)


user411:
fixes #3406

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -46,6 +49,24 @@ class NeutrinoConfigPane(
     }
   }
 
+  private val networkList = new ObservableBuffer[String]()
+
+  private val networkComboBox: ComboBox[String] = new ComboBox[String]() {
+    networkList ++= BitcoinNetworks.knownNetworks.map(_.toString)
+    items = networkList
```
benthecarman: to avoid making the `ObservableBuffer` you shoud be able to do this instead
```
new ComboBox[String](BitcoinNetworks.knownNetworks.map(_.toString))
```

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -46,6 +49,24 @@ class NeutrinoConfigPane(
...
+
+    val networkName =
+      BitcoinNetworks.fromString(appConfig.chainConf.network.name).toString
+    selectionModel.value.select(networkName)
+    onAction = _ => {
+      if (peerAddressTF.text.value.contains(".suredbits.com")) { // Or should this use the peers.headOption logic?
```
benthecarman: > // Or should this use the peers.headOption logic?

I don't think so, this looks correct to me

benthecarman: However it may be worth adding this so when switching from regtest or signet you can get the suredbits nodes
```
|| peerAddressTF.text.value.contains("localhost") || peerAddressTF.text.value.contains("127.0.0.1")
```

Would probably be worth making a `peerAddressTF.text.value` `val` too

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
user411: I get

[info] compiling 1 Scala source to /Users/ivan/code/bitcoin-s/app/bundle/target/scala-2.13/classes ...
[error] /Users/ivan/code/bitcoin-s/app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala:53:67: type mismatch;
[error]  found   : Seq[String]
[error]  required: javafx.collections.ObservableList[String]
[error]     new ObservableBuffer[String](BitcoinNetworks.knownNetworks.map(_.toString))

any hints?

user411:
fixes #3410

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -46,6 +48,22 @@ class NeutrinoConfigPane(
     }
   }
 
+  private val networkComboBox: ComboBox[NetworkParameters] =
+    new ComboBox[NetworkParameters](BitcoinNetworks.knownNetworks) {
```
benthecarman: You should be able to make these ` ComboBox[BitcoinNetwork]`

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -76,12 +97,18 @@ class NeutrinoConfigPane(
   }
 
   def getConfig: Config = {
-    val configStr =
+    var configStr =
       s"""
+         |bitcoin-s.network = ${networkComboBox.value.value}
```
benthecarman: you are gonna wanna call `DatadirUtil.networkStrToDirName` on this value so we use the correct folder.

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -76,12 +97,18 @@ class NeutrinoConfigPane(
...
          |bitcoin-s.node.mode = neutrino
          |bitcoin-s.node.peers = ["${peerAddressTF.text.value}"]
          |""".stripMargin
-
+    // Auto-enable proxy for .onion peers
+    if (peerAddressTF.text.value.contains(".onion")) {
```
benthecarman: instead of doing a `var`  (we tend to want to avoid `var`), it would be cleaner to do something like

```
val proxyConfStr = if (peerAddressTF.text.value.contains(".onion"))  {
...
} else ""

val configStr = proxyConfStr + ...

File: app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala [hunk 6]
```diff
@@ -46,6 +48,22 @@ class NeutrinoConfigPane(
...
+      value = BitcoinNetworks.fromString(appConfig.chainConf.network.name)
+      onAction = _ => {
+        val peer = peerAddressTF.text.value
+        if (
+          peer.contains(".suredbits.com") || peer.contains("localhost") || peer
+            .contains("127.0.0.1")
+        ) {
+          val network = selectionModel().getSelectedItem
+          peerAddressTF.text.value = defaultPeerForNetwork(
+            network.asInstanceOf[BitcoinNetwork])
```
benthecarman: if you do the `ComboBox[BitcoinNetwork]` you won't need to cast here

benthecarman (changes requested):

benthecarman (approved):

PR #3411, opened 2021-07-15
Disallow calling addDLCSigs as Initiator

benthecarman:


Christewart (approved):
LGTM pending CI. Will this result in some sort of a comprehensible error message to end users though?

benthecarman:
> LGTM pending CI. Will this result in some sort of a comprehensible error message to end users though?

it should

PR #3416, opened 2021-07-17
Made relay flag in version message configurable

shruthii625:
'relay' is set to false by default.
If we want to change it, we can go to the bitcoin-s.conf file and set it to true. When set to true, we see that the transaction messages get logged.

### DIFF
--- a/app/server/src/main/resources/logback.xml
+++ b/app/server/src/main/resources/logback.xml
@@ -58,16 +58,16 @@
     <!-- ╚═════════════════╝ -->
 
     <!-- See incoming message names and the peer it's sent from -->
-    <logger name="org.bitcoins.node.networking.peer.PeerMessageReceiver" level="INFO"/>
+    <logger name="org.bitcoins.node.networking.peer.PeerMessageReceiver" level="DEBUG"/>
 
     <!-- See outgoing message names and the peer it's sent to -->
-    <logger name="org.bitcoins.node.networking.peer.PeerMessageSender" level="INFO"/>
+    <logger name="org.bitcoins.node.networking.peer.PeerMessageSender" level="DEBUG"/>
 
     <!-- Inspect handling of headers and inventory messages  -->
-    <logger name="org.bitcoins.node.networking.peer.DataMessageHandler" level="INFO"/>
+    <logger name="org.bitcoins.node.networking.peer.DataMessageHandler" level="DEBUG"/>
 
     <!-- inspect TCP details -->
-    <logger name="org.bitcoins.node.networking.P2PClientActor" level="INFO"/>
+    <logger name="org.bitcoins.node.networking.P2PClientActor" level="DEBUG"/>
 
     <!-- ╔════════════════════╗ -->
     <!-- ║   Chain module     ║ -->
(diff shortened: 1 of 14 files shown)


Christewart:
Looks like there is compile issues: https://github.com/bitcoin-s/bitcoin-s/pull/3416/checks?check_run_id=3092924515#step:5:190

shruthii625:
> Looks like there is compile issues: https://github.com/bitcoin-s/bitcoin-s/pull/3416/checks?check_run_id=3092924515#step:5:190

oops, I'll check it

Christewart:
> > Looks like there is compile issues: https://github.com/bitcoin-s/bitcoin-s/pull/3416/checks?check_run_id=3092924515#step:5:190
> 
> oops, I'll check it

I'm hoping on your next commit CI automatically gets run, I need to figure out how to adjust the settings so that you and @shreyanshyad don't need me to manually approve CI runs :/

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -287,11 +287,11 @@ case class DataMessageHandler(
         MerkleBuffers.putTx(tx, appConfig.nodeCallbacks).flatMap {
           belongsToMerkle =>
             if (belongsToMerkle) {
-              logger.trace(
+              logger.debug(
```
Christewart: lets revert these too

app/server/src/main/resources/logback.xml
```diff
@@ -58,16 +58,16 @@
     <!-- ╚═════════════════╝ -->
 
     <!-- See incoming message names and the peer it's sent from -->
-    <logger name="org.bitcoins.node.networking.peer.PeerMessageReceiver" level="INFO"/>
+    <logger name="org.bitcoins.node.networking.peer.PeerMessageReceiver" level="DEBUG"/>
```
Christewart: Please revert these for actually merging into master

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
```diff
@@ -106,6 +106,13 @@ case class NodeAppConfig(
     }
   }
 
+  lazy val relay: Boolean = {
+    if (config.hasPath("bitcoin-s.node.relay"))(config.getBoolean(
```
Christewart: let's clean up the format here. Scala gives you a lot of optionality with syntax, but it is 99% of the time best to use braces. 

```scala
if (config.hasPath("bitcoin-s.node.relay")) {
  config.getBoolean("bitcoin-s.node.relay")
} else {
  false
}

**Review by Christewart — CHANGES_REQUESTED** [review]
Works for me! 🎉 We just have to add a few more things

You need to modify the [example configuration file that we publish on our website](https://bitcoin-s.org/docs/config/configuration#example-configuration-file). 

The idea for this page is to show all the available options to configure bitcoin-s in one place, with their defaults. you need to add the `bitcoin-s.node.relay=false` configuration to this file 

https://github.com/bitcoin-s/bitcoin-s/blob/8b663d91b647e8e22e4ee1a4f6cd976b00c6a766/docs/config/configuration.md#L156

Specifically you need to add 

```
bitcoin-s { 
   node { 
        # whether to have p2p peers relay us unconfirmed txs
        relay = false
    }
}
```

Once we get these final changes in, we can merge your first PR 🎉

Christewart (approved):

benthecarman (approved):

PR #3429, opened 2021-07-20
Implemented the ability to fetch unconfirmed transactions

shruthii625:
Previously, only the SPV nodes could receive transactions as a callback and not the Neutrino nodes.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -219,7 +219,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       nodeConf: NodeAppConfig,
       ec: ExecutionContext): Future[NodeCallbacks] = {
     lazy val onTx: OnTxReceived = { tx =>
-      logger.info(s"Receiving transaction txid=${tx.txIdBE} as a callback")
+      logger.info(s"Receiving transaction txid=${tx.txIdBE.hex} as a callback")
       wallet.processTransaction(tx, blockHashOpt = None).map(_ => ())
     }
     lazy val onCompactFilters: OnCompactFiltersReceived = { blockFilters =>
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -217,6 +217,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       nodeConf: NodeAppConfig,
       ec: ExecutionContext): Future[NodeCallbacks] = {
     lazy val onTx: OnTxReceived = { tx =>
+      logger.info("Receiving transaction as a callback")
       wallet.processTransaction(tx, blockHashOpt = None).map(_ => ())
     }
     lazy val onCompactFilters: OnCompactFiltersReceived = { blockFilters =>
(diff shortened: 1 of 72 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -217,6 +217,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       nodeConf: NodeAppConfig,
       ec: ExecutionContext): Future[NodeCallbacks] = {
     lazy val onTx: OnTxReceived = { tx =>
+      logger.info("Receiving transaction as a callback")
```
Christewart: nit: add the txid to the log message

Christewart (approved):

Christewart:
Actually is there any unit test we could write to make sure we don't regress on this in the future?

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
shruthii625: > nit: add the txid to the log message

okay

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -217,7 +217,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       nodeConf: NodeAppConfig,
       ec: ExecutionContext): Future[NodeCallbacks] = {
     lazy val onTx: OnTxReceived = { tx =>
-      logger.info("Receiving transaction as a callback")
+      logger.info(s"Receiving transaction txid=${tx.txId} as a callback")
```
Christewart: So another nit, but a lesson worth learning: 

we have two methods for `txId`. 

There is `txId` (which you are using) and `txIdBE`. The BE stands for big endian. Satoshi Nakamoto made the bitcoin protocol use little endian byte encoding rather than the common big endian standard. 

To confuse you even more, all transaction ids that are used on block explorer or returned from rpc calls are big endian. Thus, when logging you shoudl always use `txIdBE` so you can easily search for the tx on a block explorer.

https://bitcoin.stackexchange.com/a/41059/20835

shruthii625: Oh, that was pretty informative! I'll keep that in mind the next time I'm logging. I'll change txId to txIdBE here also

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -217,6 +217,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       nodeConf: NodeAppConfig,
       ec: ExecutionContext): Future[NodeCallbacks] = {
     lazy val onTx: OnTxReceived = { tx =>
+      logger.info(s"Receiving transaction txid=${tx.txIdBE} as a callback")
```
benthecarman: nit: should call `.hex` on `txIdBE`, otherwise will log `DoubleSHA256DigestBE(...)` instead of just the hash

shruthii625: yikes, I'll change that

shruthii625:
> Actually is there any unit test we could write to make sure we don't regress on this in the future?

I've added the unit test. Also, I've set relay = true for Neutrino nodes in the BitcoinSTestAppConfig. Hope that's okay

Christewart (approved):

Christewart:
> > Actually is there any unit test we could write to make sure we don't regress on this in the future?
> 
> I've added the unit test. Also, I've set relay = true for Neutrino nodes in the BitcoinSTestAppConfig. Hope that's okay

Yep, I think you had to do that :-)

Issue #3431, opened 2021-07-21
 requirement failed: The nonce from derived seed did not match database from DLCOracle


Christewart:
This occurred when running the `suredbits-oracle-bot` as a daemon to sign the deribit index price at 8AM UTC. 

Weirdly enough, I re-ran the bot when I woke up this morning and signing seemed to work just fine. Perhaps there is a race condition somewhere. This was run with a postgres backend on 1c6d7281

>2021-07-21 03:03:01,008 INFO  DeribitOracleClient.scala:136 - Attempting to sign announcement for today with price=30816
2021-07-21 03:03:01,164 ERROR DeribitOracleClient.scala:275 - Encountered error while create attestation for deribit
java.lang.IllegalArgumentException: requirement failed: The nonce from derived seed did not match database
        at scala.Predef$.require(Predef.scala:337)
        at org.bitcoins.dlc.oracle.DLCOracle.$anonfun$signEvent$8(DLCOracle.scala:299)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)

Christewart:
Related to #3451

Christewart:
Done in #3719

PR #3432, opened 2021-07-21
Add large digits test in DLCOracle

Christewart:
Related to #3431 

This adds a unit test for a large `numDigits` to try and eventually replicate the error detailed in #3431. I can't seem to replicate it locally :/


benthecarman (approved):
ACK on CI

PR #3433, opened 2021-07-21
Use AtomicInteger for keyIndex in DLCOracle

### LINKED ISSUE #3428: Investigate `DLCOracle.createNewEvent()` for reusing key indexes that could result in nonce leakage
https://github.com/bitcoin-s/bitcoin-s/blob/78e2fceb908a5cff1d4cdee7b759c78bd9aca88d/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala#L180

I'm wondering if there are certain situations where this could result in reusage of a nonce. Maybe to be more defensive we shoudl throw and error if we can't retrieve an index from `RValueDAO`? 

In general, I think we should add methods to `RValueDAO` that works like `incrementAndGet()` inside of `java.util.concurrent.atomic.AtomicInteger`

It seems like us automatically incrementing indices ourselves in various places throughout the application code seems likely to lead to nonce reusage. Thankfully we have a unique property at the sql level to prevent this from happening in #3415

benthecarman:
Fixes #3428

This should give us safety for concurrency when using the DLC Oracle

### DIFF
--- a/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
+++ b/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
@@ -22,6 +22,7 @@ import org.bitcoins.keymanager.{DecryptedMnemonic, WalletStorage}
 import scodec.bits.ByteVector
 
 import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
 import scala.concurrent.{ExecutionContext, Future}
 
 class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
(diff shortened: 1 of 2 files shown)


dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -71,6 +72,11 @@ class DLCOracle(private[this] val extPrivateKey: ExtPrivateKeyHardened)(implicit
   protected[bitcoins] val eventDAO: EventDAO = EventDAO()
   protected[bitcoins] val eventOutcomeDAO: EventOutcomeDAO = EventOutcomeDAO()
 
+  lazy val nextKeyIndexF: Future[AtomicInteger] = rValueDAO.maxKeyIndex.map {
```
Christewart: `private`

Christewart (approved):
So i think this make sense as a good first step to solving this problem. Eventually we will probably need database level incrementing of `keyIndex` so it is safe across all instances of `DLCOracle`. 

This solution won't work if you just create two instances of `DLCOracle` AFAICT, but I don't think this is a huge concern for now.

benthecarman:
> Eventually we will probably need database level incrementing of `keyIndex` so it is safe across all instances of `DLCOracle`.

I don't know if this is possible. Currently we need the `keyIndex` to be able to calculate the `nonce` which is needed first to actually create a something in the database

Christewart:
> > Eventually we will probably need database level incrementing of `keyIndex` so it is safe across all instances of `DLCOracle`.
> 
> I don't know if this is possible. Currently we need the `keyIndex` to be able to calculate the `nonce` which is needed first to actually create a something in the database

It seems we need to break apart the table in the future

PR #3446, opened 2021-07-23
Multi peer: initialization with tests

solo1g:
Continuing from last pr, initialized all peers and added a test for checking the same.


Christewart (approved):
This LGTM, but we might want to move away from `idx` base methods in the future, it is confusing and dependent that we never change `peers` from a immutable to a mutable data structure. Most things like `isIntialized()` should take in a `node` parameter instead.

benthecarman (approved):

Issue #3447, opened 2021-07-23
Redesign `DLCOracle` so that it can't be internally inconsistent wrt to extPrivateKey on postgres


Christewart:
This is possible since we just take a `extPrivateKey` as a parameter

https://github.com/bitcoin-s/bitcoin-s/blob/36c4da7c9520dec0c705fa1a9ead4b72d962bce3/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala#L28

I'm not opposed to having us take in a private key as a parameter, but we need to make it more clear that the `seeds` folder in the `DLCOracleAppConfig` MAY NOT have been used to generate the extPrivateKey.

My bug was caused by sharing the same postgres database (which do NOT need to be in the same `datadir` as postgres is a server) with different datadir (more specifically, different `seeds` in those datadirs). We do not have any sanity checks on startup to make sure we aren't using a different seed with the same postgres database.

Everytime I started up the server with a different datadir, I would overwrite the previous public keys used in `events` table with new ones from the different seed due to this logic 

https://github.com/bitcoin-s/bitcoin-s/blob/2a6da6a4eaea01283fbc56ef5d7da12cf3302e36/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala#L170

This ultimately is a bug on the application developer's part (mine), but we can improve this experience by throwing exceptions if we detect a wrong master `xprv` being used with our oracle to prevent them from going down the rabbit hole. 

We may want to consider similar startup logic for the wallet with postgres as I believe the esame thing can happen.

Christewart:
Done in #3449

PR #3451, opened 2021-07-26
Implement MasterXPubDAO

Christewart:
This is the first step in addressing #3447

This PR also implements validating the master xpub stored in the database against the actual key manager on startup for `wallet`, `oracle`. This means you will get an exception on startup if they aren't the same.

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
@@ -5,11 +5,11 @@ import org.bitcoins.commons.jsonmodels.wallet.{
   WalletStateDescriptorType
 }
 import org.bitcoins.core.config.{BitcoinNetwork, BitcoinNetworks}
-import org.bitcoins.core.crypto._
+import org.bitcoins.core.crypto.{ExtKeyPubVersion, _}
 import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
 import org.bitcoins.core.gcs.FilterType
 import org.bitcoins.core.hd._
-import org.bitcoins.core.number.{Int32, UInt32, UInt64}
+import org.bitcoins.core.number.{Int32, UInt32, UInt64, UInt8}
 import org.bitcoins.core.protocol.dlc.compute.SigningVersion
 import org.bitcoins.core.protocol.dlc.models.{
   ContractInfo,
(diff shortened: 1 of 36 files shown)


db-commons/src/main/scala/org/bitcoins/db/models/MasterXPubDAO.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import org.bitcoins.core.crypto.{ChainCode, ExtKeyPubVersion, ExtPublicKey}
+import org.bitcoins.core.number.{UInt32, UInt8}
+import org.bitcoins.crypto.ECPublicKey
+import org.bitcoins.db.{CRUD, DbAppConfig, SlickUtil}
+import scodec.bits.ByteVector
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** The primary key type is the fingerprint [[ExtPublicKey.fingerprint]] */
```
Christewart: old, needs to change

db-commons/src/main/scala/org/bitcoins/db/DbCommonsColumnMappers.scala
```diff
@@ -5,11 +5,11 @@ import org.bitcoins.commons.jsonmodels.wallet.{
   WalletStateDescriptorType
 }
 import org.bitcoins.core.config.{BitcoinNetwork, BitcoinNetworks}
-import org.bitcoins.core.crypto._
+import org.bitcoins.core.crypto.{ExtKeyPubVersion, _}
```
benthecarman, line 8: nit: `_`

db-commons/src/main/scala/org/bitcoins/db/models/MasterXPubDAO.scala
```diff
@@ -0,0 +1,82 @@
...
+      if (xpub != xpubs.head) {
+        sys.error(
+          s"Keymanager xpub and stored xpub are different, stored=${xpubs.head}, keymanager=${xpub}")
+      }
+    }
+  }
+
+  def existsOneXpub(): Future[Boolean] = {
+    safeDatabase.run(table.exists.result)
+  }
```
benthecarman: this only checks if the table exists, not if there is only 1 xpub

benthecarman (commented):
otherwise lgtm

Christewart:
oops, didn't mean to re-request review yet @benthecarman . Apologies

PR #3454, opened 2021-07-27
Multi peer: Sync with random peer

### LINKED ISSUE #3450: Implement ability to connect to a bitcoin node and determine if they support blockfilter services
https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#node-operation

If we connect to a bitcoin node, but they don't support blockfilters over p2p, we shouldn't use that node for initial block download as they can't serve us block filters.

In the future, we may way want to connect to some nodes that don't serve block filters as they can still be used to retrieve full blocks.

solo1g:
Closes #3450.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -42,7 +42,7 @@ case class NeutrinoNode(
     copy(dataMessageHandler = dataMessageHandler)
   }
 
-  def getRandomPeerMsgSenderWithCompactFilters: PeerMessageSender = {
+  lazy val randomPeerMsgSenderWithCompactFilters: PeerMessageSender = {
     val filteredPeers =
       peerServices.filter(_._2.nodeCompactFilters).keys.toVector
     if (filteredPeers.isEmpty)
@@ -58,8 +58,8 @@ case class NeutrinoNode(
       node <- super.start()
       chainApi <- chainApiFromDb()
       bestHash <- chainApi.getBestBlockHash()
-      _ <- peerMsgSenders(0).sendGetCompactFilterCheckPointMessage(
-        stopHash = bestHash.flip)
+      _ <- randomPeerMsgSenderWithCompactFilters
+        .sendGetCompactFilterCheckPointMessage(stopHash = bestHash.flip)
     } yield {
       node.asInstanceOf[NeutrinoNode]
     }
@@ -77,7 +77,7 @@ case class NeutrinoNode(
     * @return
     */
   override def sync(): Future[Unit] = {
-    val randomPeerMsgSender = getRandomPeerMsgSenderWithCompactFilters
+    val peerMsgSender = randomPeerMsgSenderWithCompactFilters
     val blockchainsF =
       BlockHeaderDAO()(executionContext, chainConfig).getBlockchains()
     for {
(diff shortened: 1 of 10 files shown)


node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -41,13 +42,24 @@ case class NeutrinoNode(
     copy(dataMessageHandler = dataMessageHandler)
   }
 
+  lazy val randomPeerMsgSenderWithCompactFilters: PeerMessageSender = {
```
benthecarman: this should be a `def` so we get a different peer every time

benthecarman (commented):
This will also need to be done for selecting a peer in the `DataMessageHandler`. As this currently stands it will select a random peer to sync with to start and only use it for the duration. This can be done in a separate PR however

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
solo1g: Ok, changing that. I was under the impression that we we would need to have same peer for getcfheaders and for syncing. Also, looking into data message handler, missed that one.

Christewart:
Looks like there was a compile failure: https://github.com/bitcoin-s/bitcoin-s/pull/3454/checks?check_run_id=3189403551#step:5:142

solo1g:
Yeah, this issue I have resolved and am running the tests locally now before pushing again. Would request review again once all is done.

solo1g:
1 test fails on 2.12 which passes on 2.13. The failing one is unrelated to the changes here and I have confirmed that it passes locally, so it would probably pass if I run CI again but leaving that for now.

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -19,7 +19,7 @@ import scala.concurrent.Future
 
 case class NeutrinoNode(
     nodePeer: Vector[Peer],
-    dataMessageHandler: DataMessageHandler,
+    var dataMessageHandler: DataMessageHandler,
```
Christewart: What was the reason for changing this? This makes `dataMessageHandler` a publicly available mutable variable which is generally a bad idea.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,6 +46,36 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
 
   val peers: Vector[Peer]
 
+  val peerServices: mutable.Map[Peer, ServiceIdentifier] =
```
Christewart: This is a publicly available mutable variable. Can this be made `private`?

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
solo1g: The reason for changing this here is for changing `updateDataMessageHandler` to what it is now. Earlier used `copy` to update and but did not copy the vals defined in `Node` trait that `NeutrinoNode` extends so this led to the `peerServices` map we had in `Node` trait to be initialized to empty. Also, there was one other place where states diverged in `DataMessageHandler` and led to errors in `handleDataPayload` of  `PeerMessageReceiver`. 

I get that this should not be public so made it private and added a getter for it.

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: Done

Christewart (approved):

PR #3456, opened 2021-07-27
Dialogs to CliCommandProducers with buildView(), common fileChooser

user411:
Starting point for pulling fileChooser creation to common spot to persist folder location and common buildView() / getCliCommand() for dialogs so their views can be re-used for flat views

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
@@ -15,10 +15,20 @@ import scalafx.stage.Window
 import scala.collection._
 import scala.util.{Failure, Success, Try}
 
-class AcceptOfferDialog {
+class AcceptOfferDialog extends CliCommandProducer {
 
-  def showAndWait(parentWindow: Window): Option[AcceptDLCCliCommand] = {
-    val dialog = new Dialog[Option[AcceptDLCCliCommand]]() {
+  def getCliCommand() = {
+    val offerHex = offerTLVTF.text.value
+    val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerHex)
+    Some(AcceptDLCOffer(offer))
+  }
+
+  private var dialog: Dialog[Option[AcceptDLCCliCommand]] = null
+
+  def showAndWait(
+      parentWindow: Window,
+      hex: String = ""): Option[AcceptDLCCliCommand] = {
+    dialog = new Dialog[Option[AcceptDLCCliCommand]]() {
       initOwner(parentWindow)
       title = "Accept DLC Offer"
       headerText = "Enter DLC Offer to accept"
@@ -28,15 +38,36 @@ class AcceptOfferDialog {
     dialog.dialogPane().stylesheets = GlobalData.currentStyleSheets
     dialog.resizable = true
 
-    var dlcDetailsShown = false
-    val offerTLVTF = new TextField() {
-      minWidth = 300
+    dialog.dialogPane().content = new ScrollPane {
+      content = buildView(hex)
     }
+
+    // When the OK button is clicked, convert the result to a CreateDLCOffer.
+    dialog.resultConverter = dialogButton =>
+      if (dialogButton == ButtonType.OK) {
+        getCliCommand()
+      } else None
+
+    val result = dialog.showAndWait()
+
+    result match {
+      case Some(Some(cmd: AcceptDLCCliCommand)) =>
+        Some(cmd)
+      case Some(_) | None => None
+    }
+  }
+
+  lazy val offerTLVTF = new TextField {
+    minWidth = 300
+  }
+
+  def buildView(initialOffer: String = "") = {
+    var dlcDetailsShown = false
 
     var nextRow: Int = 2
     val gridPane = new GridPane {
       alignment = Pos.Center
-      padding = Insets(top = 10, right = 10, bottom = 10, left = 10)
+      padding = Insets(10)
       hgap = 5
       vgap = 5
 
@@ -199,7 +230,9 @@ class AcceptOfferDialog {
       nextRow += 1
     }
 
-    offerTLVTF.onKeyTyped = _ => {
+    offerTLVTF.onKeyTyped = _ => onOfferKeyTyped()
+
+    def onOfferKeyTyped() = {
       if (!dlcDetailsShown) {
         Try(
           LnMessageFactory(DLCOfferTLV).fromHex(
@@ -208,31 +241,19 @@ class AcceptOfferDialog {
           case Success(lnMessage) =>
             dlcDetailsShown = true
             showOfferTerms(lnMessage.tlv)
-            dialog.dialogPane().getScene.getWindow.sizeToScene()
+            offerTLVTF.editable = false
+            if (dialog != null)
+              dialog.dialogPane().getScene.getWindow.sizeToScene()
         }
       }
     }
 
-    dialog.dialogPane().content = new ScrollPane {
-      content = new VBox(gridPane)
+    // Set initial state
+    if (initialOffer.nonEmpty) {
+      offerTLVTF.text = initialOffer
+      onOfferKeyTyped()
     }
 
-    // When the OK button is clicked, convert the result to a CreateDLCOffer.
-    dialog.resultConverter = dialogButton =>
-      if (dialogButton == ButtonType.OK) {
-
-        val offerHex = offerTLVTF.text.value
-        val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerHex)
-
-        Some(AcceptDLCOffer(offer))
-      } else None
-
-    val result = dialog.showAndWait()
-
-    result match {
-      case Some(Some(cmd: AcceptDLCCliCommand)) =>
-        Some(cmd)
-      case Some(_) | None => None
-    }
+    new VBox(gridPane)
   }
 }
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
@@ -23,12 +23,12 @@ class AcceptOfferDialog extends CliCommandProducer[AcceptDLCOffer] {
     AcceptDLCOffer(offer)
   }
 
-  private var dialog: Dialog[Option[AcceptDLCCliCommand]] = null
+  private var dialogOpt: Option[Dialog[Option[AcceptDLCCliCommand]]] = None
 
   def showAndWait(
       parentWindow: Window,
       hex: String = ""): Option[AcceptDLCCliCommand] = {
-    dialog = new Dialog[Option[AcceptDLCCliCommand]]() {
+    val dialog = new Dialog[Option[AcceptDLCCliCommand]]() {
       initOwner(parentWindow)
       title = "Accept DLC Offer"
       headerText = "Enter DLC Offer to accept"
@@ -48,7 +48,8 @@ class AcceptOfferDialog extends CliCommandProducer[AcceptDLCOffer] {
         Some(getCliCommand())
       } else None
 
-    val result = dialog.showAndWait()
+    dialogOpt = Some(dialog)
+    val result = dialogOpt.map(_.showAndWait())
 
     result match {
       case Some(Some(cmd: AcceptDLCCliCommand)) =>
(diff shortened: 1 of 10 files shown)


Christewart (commented):
One thing to note from looking at this is how you are using `Some`. If you use `Some` as the return type there isn't a point in using `Some` as the return type as it _has_ to be defined, thus you should return the underlying `A` type.

```scala
sealed abstract class Option[+A]
final case class Some[+A](value: A) extends Option[A]
case object None extends Option[Nothing]
```

If you want to have optionality, you need to return the parent type which is `Option[]` That means it can either be `None` or `Some` as they share the same parent type `Option`


Here are my changes to `CliCommandProducer`

1. Parameterize it with a `T <: CliCommand` so that we can get the specific `CliCommand` we are interested in
2. Adjust the return type `getCliCommand()` from `Some[T]` -> `T`. This means we need to wrap in `Some()` [here](https://github.com/christewart/bitcoin-s-core/blob/f058e7baaa01f11cc07f36367c3dbd4ebe9617df/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala#L57). This is fine, but if we ever need to reuse `getCliCommand()` for something else, we don't have to deal with the `Some` wrapper where it isn't necessary.

Here is the commits: https://github.com/Christewart/bitcoin-s-core/commits/2021-07-28-dialog-interface-code-review

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -22,10 +22,18 @@ import java.time.ZoneOffset
...
-class CreateDLCOfferDialog extends Logging {
+class CreateDLCOfferDialog extends Logging with CliCommandProducer {
 
-  def showAndWait(parentWindow: Window): Option[CreateDLCOffer] = {
-    val dialog = new Dialog[Option[CreateDLCOffer]]() {
+  def getCliCommand() = {
+    createDLCOffer()
+  }
+
+  private var dialog: Dialog[Option[CreateDLCOffer]] = null
```
Christewart: Don't use `null` ever. It's the billion dollar mistake with java. 

With Scala, we can use `Option` (specifically `None`) to represent `null`. Here is the diff for fixing this file. It needs to be fixed in the other files.

https://github.com/Christewart/bitcoin-s-core/commit/9790618ee6acabc63f6b76a1b688a9d29e9914c1

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -35,12 +43,56 @@ class CreateDLCOfferDialog extends Logging {
...
+
+  lazy val collateralTF = new TextField() {
+    promptText = "Satoshis"
+  }
+  setNumericInput(collateralTF)
+
+  lazy val refundDatePicker = new DatePicker()
+
+  def buildView(
+      initialContract: String = "",
```
Christewart: the convention is to make the parameter that has a default argument last, i.e. 

```scala
def buildView(
  announcement: OracleAnnouncementV0TLV,
  contractInfoOpt: Option[ContractInfoV0TLV],
  initialContract: String = ""
)
```


also with methods please make sure you annotate the return type.

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/CreateDLCOfferDialog.scala
```diff
@@ -434,99 +469,100 @@ class CreateDLCOfferDialog extends Logging {
...
-        val collateralLong =
-          numberFormatter.parse(collateralTF.text.value).longValue()
-        val collateral = Satoshis(collateralLong)
+    vbox
+  }
 
-        val feeRateStr = feeRateTF.text.value
+  def getOracleInfo: Option[OracleInfo] = {
+    Try(
+      OracleAnnouncementV0TLV.fromHex(
```
Christewart: Just so you know, we provide methods that return a `Try[]`. In this case, you should be using `OracleAnnouncementV0TLV.fromHexT`. The `T` indicates that this method returns a `Try`. The suffix `Opt` means we are returning `Option[]`. The suffix `F` means we are returning a `Future`.

Christewart: https://github.com/bitcoin-s/bitcoin-s/blob/8f958bf5d9957438d703819a087d87813f3edd89/crypto/src/main/scala/org/bitcoins/crypto/Factory.scala#L25

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -31,6 +38,78 @@ object GUIUtil {
...
+    selectedExtensionFilter = txtExtensionFilter
+    initialDirectory = new File(Properties.userHome)
+  }
+
+  private lazy val txtExtensionFilter =
+    new ExtensionFilter("Text Files", "*.txt")
+  private lazy val allExtensionFilter = new ExtensionFilter("All Files", "*")
+
+  def showSaveDialog(
+      bytes: String = null,
```
Christewart: get rid of `null`s

Christewart: It's used pretty pervasively in this file, we need to remove all usages and replace with `Option`. 


It seems like we are using default arguments quite a bit too. Generally this should be used sparingly.

Christewart (changes requested):

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
```diff
@@ -242,8 +243,8 @@ class AcceptOfferDialog extends CliCommandProducer[AcceptDLCOffer] {
             dlcDetailsShown = true
             showOfferTerms(lnMessage.tlv)
             offerTLVTF.editable = false
-            if (dialog != null)
-              dialog.dialogPane().getScene.getWindow.sizeToScene()
+            if (dialogOpt.isDefined)
+              dialogOpt.get.dialogPane().getScene.getWindow.sizeToScene()
```
benthecarman, line 247: rather than doing this it is better to do a `match` on `dialogOpt` and handle both cases

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/BroadcastDLCDialog.scala
```diff
@@ -281,7 +282,8 @@ object BroadcastDLCDialog
           errorLabel.text = errMsg
           vbox.children.add(errorLabel)
       }
-      if (dialog != null) dialog.dialogPane().getScene.getWindow.sizeToScene()
+      if (dialogOpt.isDefined)
+        dialogOpt.get.dialogPane().getScene.getWindow.sizeToScene()
```
benthecarman, line 287: same here

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/BroadcastDLCDialog.scala
```diff
@@ -299,8 +301,8 @@ object BroadcastDLCDialog
           logger.error(errMsg)
           errorLabel.text = errMsg
           vbox.children.add(errorLabel)
-          if (dialog != null)
-            dialog.dialogPane().getScene.getWindow.sizeToScene()
+          if (dialogOpt.isDefined)
+            dialogOpt.get.dialogPane().getScene.getWindow.sizeToScene()
```
benthecarman, line 306: same here

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -301,7 +302,8 @@ object SignDLCDialog
           errorLabel.text = errMsg
           vbox.children.add(errorLabel)
       }
-      if (dialog != null) dialog.dialogPane().getScene.getWindow.sizeToScene()
+      if (dialogOpt.isDefined)
+        dialogOpt.get.dialogPane().getScene.getWindow.sizeToScene()
```
benthecarman, line 306: same here

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -319,8 +321,8 @@ object SignDLCDialog
           logger.error(errMsg)
           errorLabel.text = errMsg
           vbox.children.add(errorLabel)
-          if (dialog != null)
-            dialog.dialogPane().getScene.getWindow.sizeToScene()
+          if (dialogOpt.isDefined)
+            dialogOpt.get.dialogPane().getScene.getWindow.sizeToScene()
```
benthecarman, line 325: same here

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -56,42 +59,40 @@ object GUIUtil {
...
+        if (bytes.isDefined) {
+          val _ = Files.write(chosenFile.toPath, bytes.get.getBytes)
         }
         // Remember last-used directory
         fileChooser.initialDirectory = chosenFile.getParentFile
 
-        if (handleFile != null) {
-          handleFile(chosenFile)
+        if (handleFile.isDefined) {
+          handleFile.get(chosenFile)
```
benthecarman: same here

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -56,42 +59,40 @@ object GUIUtil {
...
+      bytes: Option[String],
+      handleFile: Option[File => Unit]): Unit = {
+    fileChooser.initialFileName = filename
     val chosenFileOpt = Option(fileChooser.showSaveDialog(null))
     chosenFileOpt match {
       case Some(chosenFile) =>
-        if (bytes != null) {
-          val _ = Files.write(chosenFile.toPath, bytes.getBytes)
+        if (bytes.isDefined) {
+          val _ = Files.write(chosenFile.toPath, bytes.get.getBytes)
```
benthecarman: same here

benthecarman (changes requested):

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
user411: Changed this one, added comment

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
user411: Changed this one, added comment

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
user411: I left all the `dialogOpt.isDefined`s in place. These only apply to the case where a dialog is getting used and there is no extra value in the 'else' case which automatically sizes

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -66,14 +66,17 @@ object GUIUtil {
...
-        if (handleFile.isDefined) {
-          handleFile.get(chosenFile)
+        bytes match {
+          case Some(_) => Files.write(chosenFile.toPath, bytes.get.getBytes)
+          case None    => // There was nothing sent in to write out
+        }
+
+        handleFile match {
+          case Some(_) => handleFile.get(chosenFile)
+          case None    => // No callback defined
```
benthecarman: this is kinda  a weird way to do it.

It's generally more standard to do something like
```scala

bytesOpt match {
   Some(bytes) => Files.write(chosenFile.toPath, bytes.getBytes)
   case None => ...
}
```

Christewart:
Looks like `scalafmt` failed, if you just re-run `test:compile` locally, scalafmt should do it's thing and fix this 

https://github.com/bitcoin-s/bitcoin-s/pull/3456/checks?check_run_id=3203031364

Christewart (changes requested):
This should be the last of the nitting. We already have concurrency issues on the GUI and making `var`s public adds more complexity. It doesnt seem like there was a reason these needed to be public so I made them private: 

https://github.com/Christewart/bitcoin-s-core/commit/357fd3f5d10a6819f02fcddf176910c756f49fa0

Christewart (approved):

PR #3462, opened 2021-07-28
DLC P2P in AppServer and GUI

benthecarman:
This hooks up the `DLCNode` from #3402 to the app server, cli, and gui.

Adds tor docs

@Christewart and I were able to execute 2 DLCs using this (numeric and enum)

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
@@ -1,8 +1,9 @@
 package org.bitcoins.gui.dlc.dialog
 
-import org.bitcoins.cli.CliCommand.{AcceptDLCCliCommand, AcceptDLCOffer}
+import org.bitcoins.cli.CliCommand._
 import org.bitcoins.core.protocol.dlc.models._
 import org.bitcoins.core.protocol.tlv._
+import org.bitcoins.core.util.NetworkUtil
 import org.bitcoins.gui.GlobalData
 import org.bitcoins.gui.dlc.DLCPlotUtil
 import org.bitcoins.gui.util.GUIUtil
@@ -12,15 +13,27 @@ import scalafx.scene.control._
 import scalafx.scene.layout._
 import scalafx.stage.Window
 
+import java.net.{InetSocketAddress, URI}
 import scala.collection._
 import scala.util.{Failure, Success, Try}
 
-class AcceptOfferDialog extends CliCommandProducer[AcceptDLCOffer] {
+class AcceptOfferDialog extends CliCommandProducer[AcceptDLCCliCommand] {
 
-  override def getCliCommand(): AcceptDLCOffer = {
+  override def getCliCommand(): AcceptDLCCliCommand = {
     val offerHex = offerTLVTF.text.value
     val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerHex)
-    AcceptDLCOffer(offer)
+    if (peerAddressTF.text.value.nonEmpty) {
+      val peer = new URI(
+        "tcp://" + NetworkUtil.parseInetSocketAddress(peerAddressTF.text.value,
+                                                      2862))
+
+      val peerAddr =
+        InetSocketAddress.createUnresolved(peer.getHost, peer.getPort)
+
+      AcceptDLC(offer, peerAddr)
+    } else {
+      AcceptDLCOffer(offer)
+    }
   }
 
   private var dialogOpt: Option[Dialog[Option[AcceptDLCCliCommand]]] = None
@@ -62,23 +75,40 @@ class AcceptOfferDialog extends CliCommandProducer[AcceptDLCOffer] {
     minWidth = 300
   }
 
+  private lazy val peerAddressTF = new TextField() {
+    minWidth = 300
+    promptText = "(optional)"
+  }
+
   def buildView(initialOffer: String = "") = {
     var dlcDetailsShown = false
 
-    var nextRow: Int = 2
+    var nextRow: Int = 0
     val gridPane = new GridPane {
       alignment = Pos.Center
       padding = Insets(10)
       hgap = 5
       vgap = 5
 
-      add(new Label("DLC Offer") {
-            tooltip = Tooltip("Offer message given from your counter party.")
+      add(
+        new Label("DLC Offer") {
+          tooltip = Tooltip("Offer message given from your counter party.")
+          tooltip.value.setShowDelay(new javafx.util.Duration(100))
+        },
+        0,
+        nextRow
+      )
+      add(offerTLVTF, 1, nextRow)
+      nextRow += 1
+
+      add(new Label("Peer Address") {
+            tooltip = Tooltip("Peer's IP or onion address")
             tooltip.value.setShowDelay(new javafx.util.Duration(100))
           },
           0,
-          0)
-      add(offerTLVTF, 1, 0)
+          nextRow)
+      add(peerAddressTF, 1, nextRow)
+      nextRow += 1
     }
 
     def showOfferTerms(offer: DLCOfferTLV): Unit = {
--- a/docs/tor/tor.md
+++ b/docs/tor/tor.md
@@ -1,0 +1,145 @@
+---
+id: tor
+title: Tor Setup
+---
+
+It is possible to run Bitcoin-S through tor.
+[Tor](https://www.torproject.org/) is an onion routed private network that allows us to send and receive messages in an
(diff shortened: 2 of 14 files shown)


docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+You will also need permissions for the auth cookie file, this can be done doing
+
+```bash
+sudo usermod -a -G debian-tor $USER
+```
+
+After changing these settings, you will need to restart your computer.
+
```
Christewart, line 105: Does this tor configuration make our users be exit nodes? I doubt we want that by default if so.

benthecarman: No, that is a different config

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+You can install tor using `brew install tor` if on a debian system.
+
+### Other
+
+Otherwise, you can install the Tor Browser from [here](https://www.torproject.org/download/).
+
+## Starting Tor
+
+To connect to onion addresses you need th enable the tor proxy.
+To do so you need to have tor currently running, this can be checked by using the command `sudo systemctl status tor`.
```
Christewart: does not work on mac

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+        enabled = true
+        host = "127.0.0.1"
+        port = 9050
+    }
+}
+```
+
+## Creating our own Tor hidden service
+
+Enabling the tor hidden services allows you to create inbound connections over tor.
```
Christewart: nit: `allows you to create` -> `allows for`

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+This is needed if you want to create DLCs over tor.
+
+To enable the tor hidden services you need to set a couple config options after you have tor running in your bitcoin-s config, as well as
+have tor configured for it.
+
+### Configuring Tor
+
+You may need to set up the Tor Control Port. On Linux distributions there may be
+some or all of the following settings in `/etc/tor/torrc`, generally commented
```
Christewart: path is not there on mac

Christewart: If installed via brew the path should be `/opt/homebrew/etc/tor/torcc` 

I asked how to find it more generally on the tor irc chanel, and they said if your just start `tor` from the command ine, it will output a log that looks like this: 

`Configuration file "/opt/homebrew/etc/tor/torrc" not present, using reasonable defaults`

docs/tor/tor.md
Christewart: Ok, finally found it. It looks like if you install via brew it is `/opt/homebrew/etc/tor/torcc`

The tor irc channel recommended that I just start the `tor` binary in the terminal to determine what path tor is using

Screen Shot 2021-07-29 at 12 03 29 PM

Christewart: One other problem with this is the file is named `torrc.sample`. If I understand correctly, you need to copy the contents of `torrc.sample` to a new file called `torrc` in the same directory and then make the modifications

Screen Shot 2021-07-29 at 12 09 49 PM

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+line:
+
+```
+DataDirectoryGroupReadable 1
+```
+
+You will also need permissions for the auth cookie file, this can be done doing
+
+```bash
+sudo usermod -a -G debian-tor $USER
```
Christewart, line 101: does not work on mac 

Screen Shot 2021-07-29 at 12 19 21 PM

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+### Configuring Tor
+
+You may need to set up the Tor Control Port. On Linux distributions there may be
+some or all of the following settings in `/etc/tor/torrc`, generally commented
+out by default (if not, add them):
+
+```
+ControlPort 9051
+CookieAuthentication 1
+CookieAuthFileGroupReadable 1
```
Christewart, line 85: This produces a warning message for me when i start `tor`

>Jul 29 12:17:57.230 [warn] CookieAuthFileGroupReadable is set, but will have no effect: you must specify an explicit CookieAuthFile to have it group-readable.

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+```bash
+sudo usermod -a -G debian-tor $USER
+```
+
+After changing these settings, you will need to restart your computer.
+
+### Configuring Bitcoin-S
+
+You need to enable tor and set the control option, `127.0.0.1:9051` is the default.
+If you are using the default settings you should only need to set `bitcoin-s.enabled.enabled = true`.
```
Christewart: typo `enabled.enabled=true`

docs/tor/tor.md
```diff
@@ -0,0 +1,143 @@
...
+
+```$xslt
+bitcoin-s {
+    tor {
+        # You can enable Tor for incoming connections
+        enabled = true
+        control = "127.0.0.1:9051"
```
Christewart, line 116: nit: The port is separate above but embedded here. Not sure what is best, but probably should try to be consistent.

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
```diff
@@ -224,7 +242,16 @@ class AcceptOfferDialog {
         val offerHex = offerTLVTF.text.value
         val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerHex)
 
-        Some(AcceptDLCOffer(offer))
+        if (peerAddressTF.text.value.nonEmpty) {
+          val peer = new URI("tcp://" + peerAddressTF.text.value)
```
rorp: It'd be great to make port optional for this field, so the end user could type just an onion address without port.

```scala
val peer = new URI("tcp://" + NetworkUtil.parseInetSocketAddress(peerAddressTF.text.value, DefaultDLCPort))```

File: docs/tor/tor.md (hunk 1 above)
Comment (rorp) [reply docs/tor/tor.md:105]: Exit relay is enabled by default. One needs to set `ExitRelay 0` to disable it

File: docs/tor/tor.md (hunk 5 above)
Comment (Christewart) [reply docs/tor/tor.md:101]: not needed on mac according to Ivan, Roman and me.

File: docs/tor/tor.md [hunk 10]
```diff
@@ -0,0 +1,145 @@
...
+or `sudo systemctl restart tor` on most systemd-based systems, including recent Debian and Ubuntu, or just restart the
+computer).
+
+On some systems (such as Arch Linux), you may also need to add the following line:
+
+```
+DataDirectoryGroupReadable 1
+```
+
+You may also need permissions for the auth cookie file, this can be done doing
```
Christewart, line 98: >You may also need permissions for the auth cookie file on debian based systems, this can be done doing

I think is what we concluded

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
```diff
@@ -12,15 +13,27 @@ import scalafx.scene.control._
...
 
-  override def getCliCommand(): AcceptDLCOffer = {
+  override def getCliCommand(): AcceptDLCCliCommand = {
     val offerHex = offerTLVTF.text.value
     val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerHex)
-    AcceptDLCOffer(offer)
+    if (peerAddressTF.text.value.nonEmpty) {
+      val peer = new URI(
+        "tcp://" + NetworkUtil.parseInetSocketAddress(peerAddressTF.text.value,
+                                                      2862))
```
Christewart, line 28: Do we want to check if the user entered a custom port? It seems to me that we should default to `2862` (maybe put this in a constant somewhere?) if they didn't add a port into the text box?

Christewart (changes requested):

docs/tor/tor.md
Christewart: Can we get some confirmation on this from tor docs? @benthecarman told me it is disabled by default IIRC

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
benthecarman: It does, 2862 is the default if they don't enter one

benthecarman: I am not sure where we would put the constant, the gui module doesn't have access to the DLC node stuff

PR #3470, opened 2021-08-02
Tor support for BTC RPC

rorp:
It requires manual setup of hidden service for `bitcoind`. 

For mainnet you need to add these lines into your `torrc` file:

```
HiddenServiceDir /usr/local/var/tor/hidden_service/
HiddenServicePort 8332 127.0.0.1:8332
```

The onion address can be found in `/usr/local/var/tor/hidden_service/hostname`.


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala
```diff
@@ -90,6 +91,23 @@ case class BitcoindRpcAppConfig(
   lazy val rpcPassword: String =
     config.getString("bitcoin-s.bitcoind-rpc.rpcpassword")
 
+  lazy val socks5ProxyParams: Option[Socks5ProxyParams] = {
+    if (config.getBoolean("bitcoin-s.proxy.enabled")) {
```
benthecarman, line 96: in #3402 we switched it to be like this, should probably do the same here

```scala
      val uri = new URI("tcp://" + config.getString("bitcoin-s.proxy.sock5"))
      val sock5 = InetSocketAddress.createUnresolved(uri.getHost, uri.getPort)
      Some(
        Socks5ProxyParams(
          address = sock5,
          credentialsOpt = None,
          randomizeCredentials = true
        )
      )
```

benthecarman (approved):

benthecarman:
needs rebase

Christewart:
> For mainnet you need to add these lines into your `torrc` file:
> 
> ```
> HiddenServiceDir /usr/local/var/tor/hidden_service/
> HiddenServicePort 8332 127.0.0.1:8332
> ```
> 
> The onion address can be found in `/usr/local/var/tor/hidden_service/hostname`.

So i added the lines above here to my `torrc` file and then restarted the service with `sudo service tor restart` and then tried to check for the onion but it seems like it doesn't exist

Screenshot from 2021-08-03 09-47-08

Christewart:
> > For mainnet you need to add these lines into your `torrc` file:
> > ```
> > HiddenServiceDir /usr/local/var/tor/hidden_service/
> > HiddenServicePort 8332 127.0.0.1:8332
> > ```   
> >   
> > The onion address can be found in `/usr/local/var/tor/hidden_service/hostname`.
> 
> So i added the lines above here to my `torrc` file and then restarted the service with `sudo service tor restart` and then tried to check for the onion but it seems like it doesn't exist
> 
> Screenshot from 2021-08-03 09-47-08

Ok this works for me, but my path was different than the one @rorp suggests 

My datadir is ` /var/lib/tor ` rather than `/usr/local/var/tor`

Also if I understand correclty, can pick a name better than `hidden_service`, perhaps `bitcoind_rpc_hidden_service` ? 

These were useful resources in helping me figure this out 

https://en.bitcoin.it/wiki/Setting_up_a_Tor_hidden_service

https://www.linuxjournal.com/content/tor-hidden-services

Christewart (approved):

PR #3472, opened 2021-08-03
Default Use Tor to config value, space login pages the same

user411:
GUIUtil updates, icons and styling updates that will be used with other checkins

Network ComboBox to full width to match Bitcoin version selector
'Use Tor' button defaults to bitcoin-s.config value.
"Launch" button lays out the same on both pages after change.

Screen Shot 2021-08-03 at 12 15 15 PM
Screen Shot 2021-08-03 at 12 15 28 PM


app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -114,10 +115,40 @@ object GUIUtil {
     }
   }
 
+  def getCopyToClipboardButton(property: StringProperty) = new Button {
+    styleClass += "icon-button" += "copy-button"
```
benthecarman: nit instead of this you can also do
```scala
styleClass ++= Vector("icon-button", "copy-button")
```

app/bundle/src/main/scala/org/bitcoins/bundle/gui/BitcoindConfigPane.scala
```diff
@@ -53,7 +53,9 @@ class BitcoindConfigPane(
       minWidth = 300
     }
 
-  private val torCheckBox: CheckBox = new CheckBox()
+  private val torCheckBox: CheckBox = new CheckBox() {
+    selected = appConfig.config.getBoolean("bitcoin-s.proxy.enabled")
```
benthecarman: would be better to do `appConfig.nodeConf.socks5ProxyParams.isDefined` in case we change how this is defined, then we don't need to remember to change it here as well

app/bundle/src/main/scala/org/bitcoins/bundle/gui/NeutrinoConfigPane.scala
```diff
@@ -63,21 +64,24 @@ class NeutrinoConfigPane(
...
     }
 
   private val peerAddressTF: TextField = new TextField() {
     text = startingPeerAddress
     minWidth = 300
   }
 
-  private val torCheckBox: CheckBox = new CheckBox()
+  private val torCheckBox: CheckBox = new CheckBox() {
+    selected = appConfig.config.getBoolean("bitcoin-s.proxy.enabled")
```
benthecarman: same here

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
```diff
@@ -114,10 +115,40 @@ object GUIUtil {
...
+    imageView.fitHeight = 16
+    imageView.fitWidth = 16
+    imageView
+  }
+
   def getHSpacer() = {
     new Region { hgrow = Priority.Always }
   }
 
+  def getVSpacer() = {
```
benthecarman: Should add return types to all of these functions in this file

ie `def getHSpacer(): Region`

benthecarman (changes requested):

app/bundle/src/main/scala/org/bitcoins/bundle/gui/BitcoindConfigPane.scala
user411: thanks, changed these

app/gui/src/main/scala/org/bitcoins/gui/util/GUIUtil.scala
user411: Marked up all the defs with return types

Christewart (approved):

benthecarman (approved):

PR #3473, opened 2021-08-03
Add preferences for wallet window x,y,w,h persistence

user411:
Add preference keys and loading for main stage x, y, width, and height.

Set window minimum width, height to try to avoid any kind of issues with tiny windows getting saved out.

Checks screens at boot to make sure window will layout on one or more or fail out to default screen, centered.

Tested with second monitor to left and right of primary monitor including failing over after repositioning monitor in between app runs. Worked to fail over to default monitor.


benthecarman (approved):
works for me, nice job

Christewart (approved):

PR #3474, opened 2021-08-03
Modifer keys per OS, Mac native menubar

user411:
Updates modifier keys used in application menus, moves the menubar to the top of screen on macOS, fix issue with application quit from menu that doesn't hit BundleGUI.stopApp() which is required to catch window position at application close

Screen Shot 2021-08-03 at 12 36 08 PM

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
@@ -3,6 +3,7 @@ package org.bitcoins.gui
 import org.bitcoins.cli.CliCommand.ZipDataDir
 import org.bitcoins.cli.ConsoleCli
 import org.bitcoins.gui.settings.Themes
+import scalafx.application.Platform
 import scalafx.scene.control._
 import scalafx.scene.input.{KeyCode, KeyCodeCombination, KeyCombination}
 import scalafx.stage.FileChooser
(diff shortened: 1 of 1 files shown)


app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
```diff
@@ -48,8 +65,8 @@ private class FileMenu() {
   private val quit: MenuItem = new MenuItem("_Quit") {
     mnemonicParsing = true
     accelerator =
-      new KeyCodeCombination(KeyCode.Q, KeyCombination.ControlDown) // CTRL + Q
-    onAction = _ => sys.exit()
+      new KeyCodeCombination(KeyCode.Q, menuUtil.modifierKey) // Ctrl/Cmd + Q
+    onAction = _ => Platform.exit()
```
user411, line 60: Gets BundleGUI.stopApp() called when quitting application from keyboard

benthecarman (approved):
Can't test but looks good to me

app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
```diff
@@ -11,14 +12,30 @@ import scalafx.stage.FileChooser.ExtensionFilter
 import java.io.File
 import scala.util.Properties
 
+// Utility functions for menuing. May want in a util file somewhere
+object menuUtil {
```
Christewart: nit: Upper case `MenuUtil`

Christewart (approved):

app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
user411: I removed menuUtil in favor of using the builtin KeyCombination.ShortcutDown which covers the use case natively

benthecarman (approved):

PR #3477, opened 2021-08-04
Set disable to context menu items based on selection, add Refund/Execute

user411:
Adds Refund and Execute context menu items in DLCTableView and sets disable state on Cancel/Refund/Execute actions based on DLCStatus type

Screen Shot 2021-08-04 at 11 33 11 AM

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
@@ -191,15 +191,21 @@ class DLCTableView(model: DLCPaneModel) {
               }
             }
           }
-          // Enable/disable menu items per selection
+          // Enable/disable menu items per selection state
           row.item.onChange { (_, _, newContent) =>
             if (newContent != null) {
-              cancelDLCItem.disable =
-                row.item.value.isInstanceOf[AcceptedDLCStatus]
-              val enableRefundExecute =
-                row.item.value.isInstanceOf[BroadcastedDLCStatus]
-              refundDLCItem.disable = !enableRefundExecute
-              executeDLCItem.disable = !enableRefundExecute
+              cancelDLCItem.disable = row.item.value.state match {
+                case DLCState.Offered | DLCState.Accepted | DLCState.Signed =>
+                  false
+                case _ => true
+              }
+              val disableRefundExecute = row.item.value.state match {
+                case DLCState.Broadcasted | DLCState.Confirmed =>
+                  false
+                case _ => true
+              }
+              refundDLCItem.disable = disableRefundExecute
+              executeDLCItem.disable = disableRefundExecute
             }
           }
 
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
@@ -197,12 +197,18 @@ class DLCTableView(model: DLCPaneModel) {
               cancelDLCItem.disable = row.item.value.state match {
                 case DLCState.Offered | DLCState.Accepted | DLCState.Signed =>
                   false
-                case _ => true
+                case DLCState.Confirmed | DLCState.Broadcasted |
+                    DLCState.Claimed | DLCState.Refunded |
+                    DLCState.RemoteClaimed =>
+                  true
               }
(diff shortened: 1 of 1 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -121,47 +125,94 @@ class DLCTableView(model: DLCPaneModel) {
...
+                GlobalDLCData.lastContractId = id.toHex
+                model.onExecute()
+              }
+            }
+          }
+          // Enable/disable menu items per selection
+          row.item.onChange { (_, _, newContent) =>
+            if (newContent != null) {
+              cancelDLCItem.disable =
+                row.item.value.isInstanceOf[AcceptedDLCStatus]
```
benthecarman: You should be able to cancel if it is, offered, accepted, or signed

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -121,47 +125,94 @@ class DLCTableView(model: DLCPaneModel) {
...
+              val enableRefundExecute =
+                row.item.value.isInstanceOf[BroadcastedDLCStatus]
```
benthecarman: Should be able to if it is broadcasted or confirmed

user411:
Ben, I put in matching against row.item.value.state, the IDE was giving me a hard time type matching against row.item.value for some reason. Please let me know if you think it should match differently and I'll change it.

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -191,15 +191,21 @@ class DLCTableView(model: DLCPaneModel) {
...
-              cancelDLCItem.disable =
-                row.item.value.isInstanceOf[AcceptedDLCStatus]
-              val enableRefundExecute =
-                row.item.value.isInstanceOf[BroadcastedDLCStatus]
-              refundDLCItem.disable = !enableRefundExecute
-              executeDLCItem.disable = !enableRefundExecute
+              cancelDLCItem.disable = row.item.value.state match {
+                case DLCState.Offered | DLCState.Accepted | DLCState.Signed =>
+                  false
+                case _ => true
```
benthecarman: we tend to want to avoid doing `case _ =>` as if we add a new state we won't have the compiler yell at us that we need to handle it. Should explicity list all of the possible states

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
```diff
@@ -191,15 +191,21 @@ class DLCTableView(model: DLCPaneModel) {
...
+              }
+              val disableRefundExecute = row.item.value.state match {
+                case DLCState.Broadcasted | DLCState.Confirmed =>
+                  false
+                case _ => true
```
benthecarman: same here

benthecarman (changes requested):

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
user411: This is a known subset we want to select for, so I have a bit of a visceral reaction to naming out every single thing we know we don't want when _ is setup to be that and the in-group assumed to be a known into the future and guaranteeing that this code would require updated now for underlying type updates.

I understand why there's reason to be explicit to every case in underlying wallet cases, but at the UI, this obfuscates intent without much to gain.

I checked in a full explicit class callout.

user411: I guess another option would be to use double equals notation on an explicit subset if the goal is to not use match with a _. I can change the code to that notation if it's more desirable.

benthecarman (approved):

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCTableView.scala
benthecarman: to simplify something you could do is add a `canClose` and a `canCancel` function to the states. Then can just do 
```scala
cancelDLCItem.disable = row.item.value.state.canCancel
```

PR #3479, opened 2021-08-04
Fix dialog results optionality nesting to match dialogOpt

user411:
Add an additional level of option nest result handling to go with dialog -> dialogOpt in recent refactor


Christewart (approved):
Works for me locally

benthecarman (approved):

PR #3483, opened 2021-08-04
Package Tor with Bitcoin-s

benthecarman:
Adds the tor binaries to `tor` module and will save them to your bitcoin-s datadir if you don't have tor installed. Then will run tor on startup if it is enabled.

Will add to testkit in follow up PR


tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
```diff
@@ -0,0 +1,230 @@
...
+      sys.env
+        .getOrElse("PATH", "")
+        .split(File.pathSeparator)
+        .map(directory => new File(directory, name))
+        .find(file => file.isFile && file.canExecute)
+
+    if (EnvUtil.isWindows) {
+      findExecutableOnPath("tor.exe")
+    } else {
+      findExecutableOnPath("tor")
```
benthecarman: when testing, if you want to get the binaries from bitcoin-s instead of running your local version you should change this value to something like `tor1` so it doesn't find it on the class path

Christewart:
Some preliminary thoughts before diving in for code review: 

How do we handle the case where the user has tor installed, but _doesn't_ have configuration settings we need? Do we manually modify their `torrc` file for them?

benthecarman:
> Some preliminary thoughts before diving in for code review:
> 
> How do we handle the case where the user has tor installed, but _doesn't_ have configuration settings we need? Do we manually modify their `torrc` file for them?

We won't have permissions to do that, and that'd be pretty dangerous.

If they have tor installed and not running, this will handle it correctly. The only time it will error is if they have it running and their config is different.

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -330,18 +330,22 @@ trait Client
   /** Cached http client to send requests to bitcoind with */
   private lazy val httpClient: HttpExt = Http(system)
 
-  private lazy val httpConnectionPoolSettings: ConnectionPoolSettings =
+  private lazy val httpConnectionPoolSettings: ConnectionPoolSettings = {
     instance.proxyParams match {
       case Some(proxyParams) =>
-        val socks5ClientTransport = new Socks5ClientTransport(proxyParams)
+        if (instance.rpcUri.getHost.contains(".onion")) {
```
rorp: I think we should connect over tor for all kind of addresses if `proxy.enabled=true`
Otherwise it will leak the user's data when running over clear net.

As far as I understand, this check is needed to avoid connection errors when BTC RPC backend is enabled and running on `localhost`, but tor cannot proxy connections to `localhost` and fails. 

So it looks like it's more appropriate just to check if  rpcUri is pointing to `localhost` (`127.0.0.1` or `::1`) and don't set up a proxy in this case.

tor/src/main/resources/osx_64/tor
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
Christewart, line 1: is this copied from somewhere? If so, can you link the original source in a comment?

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
```diff
@@ -0,0 +1,231 @@
...
+
+import java.io.{File, FileNotFoundException}
+import java.net.InetSocketAddress
+import java.nio.file.{Files, Path, StandardCopyOption}
+import scala.concurrent.{ExecutionContext, Future}
+import scala.sys.process.{Process, ProcessBuilder}
+import scala.util.{Failure, Success, Try}
+
+/** A trait that helps start bitcoind/eclair when it is started via bitcoin-s */
+class TorClient(implicit ec: ExecutionContext, conf: TorAppConfig)
```
Christewart: nit: `class TorClient()(implicit ec: ExecutionContext, conf: TorAppConfig)`

Christewart: Could we use `NativeProcessFactory` to have some continuity? 

I believe `NativeProcessFactory` can be moved into `core` with no issues 

https://github.com/bitcoin-s/bitcoin-s/blob/4055de7690f6a96d63be225c019243749102f687/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/util/NativeProcessFactory.scala#L9

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
```diff
@@ -0,0 +1,231 @@
...
+      val auth = SafeCookie()
+      val privKeyPath = conf.datadir.resolve("tor_priv_key")
+
+      TorParams(control, auth, privKeyPath)
+  }
+
+  private lazy val authenticationArg = torParams.authentication match {
+    case Password(_) =>
+      "error"
+//      s"--HashedControlPassword $password" // todo: need to hash the password correctly
```
Christewart, line 51: is this still TODO? I'm not sure if I understand what is happening here. Perhaps a log message?

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
```diff
@@ -0,0 +1,231 @@
...
+      findExecutableOnPath("tor")
+    }
+  }
+
+  /** Copies the tor executable and needed files to the given datadir
+    * Returns the tor executable file
+    * @param datadir Directory where to write files
+    * @return Tor executable file
+    */
+  def torBinaryFromResource(datadir: Path): File = {
```
Christewart: nit: `private`, This seems like something that we want to hide if possible

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
```diff
@@ -0,0 +1,231 @@
...
+
+    val executableFileName = datadir.resolve(torFileName).toFile
+
+    logger.info(
+      s"Using prepackaged Tor from bitcoin-s resources, $executableFileName")
+
+    if (executableFileName.exists() && executableFileName.canExecute) {
+      executableFileName
+    } else {
+      logger.info("Tor executable is not written to datadir, creating...")
```
Christewart: nit: Add `datadir` we are writing to in log

tor/src/main/scala/org/bitcoins/tor/config/TorAppConfig.scala
```diff
@@ -71,6 +72,15 @@ case class TorAppConfig(private val directory: Path, private val confs: Config*)
       None
     }
   }
+
+  lazy val enabled: Boolean = socks5ProxyParams.isDefined || torParams.isDefined
```
Christewart, line 76: This is a `||` not `&&` ? My understanding (which probably is wrong) is we need _both_ these settings.

tor/src/main/resources/osx_64/tor
benthecarman: this is from the tor browser download

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
benthecarman: Oh yeah, copy pasted it initially, meant to move it

tor/src/main/scala/org/bitcoins/tor/config/TorAppConfig.scala
benthecarman: No this should be `||` we are using tor if either of these are enabled

tor/src/main/resources/osx_64/tor
benthecarman: don't want to edit this file so the shasums don't change, added more context to the readme

tor/src/main/scala/org/bitcoins/tor/client/TorClient.scala
benthecarman: Yeah, I can't seem to find how to correctly hash the password, made it throw an error now

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -54,6 +56,13 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
 
     for {
       _ <- startedConfigF
+
+      _ <-
+        if (torConf.enabled) {
+          val tor = torConf.createClient
+          tor.startBinary()
```
rorp, line 63: What if tor daemon is already up and running (by `systemd`, for example)? Do we want to start it anyway?

Probably, first we should check if the ports defined in `torConf` are already bound, and if not, then start the binary.

benthecarman: currently it will attempt to start up but fail if it is already running

```
Aug 09 12:29:22.656 [notice] Tor 0.4.2.7 running on Linux with Libevent 2.1.11-stable, OpenSSL 1.1.1f, Zlib 1.2.11, Liblzma 5.2.4, and Libzstd 1.4.4.
Aug 09 12:29:22.656 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://www.torproject.org/download/download#warning
Aug 09 12:29:22.657 [notice] Read configuration file "/etc/tor/torrc".
Aug 09 12:29:22.660 [warn] CookieAuthFileGroupReadable is set, but will have no effect: you must specify an explicit CookieAuthFile to have it group-readable.
Aug 09 12:29:22.660 [notice] Opening Socks listener on 127.0.0.1:9050
Aug 09 12:29:22.660 [warn] Could not bind to 127.0.0.1:9050: Address already in use. Is Tor already running?
Aug 09 12:29:22.660 [notice] Opening Control listener on 127.0.0.1:9051
Aug 09 12:29:22.660 [warn] Could not bind to 127.0.0.1:9051: Address already in use. Is Tor already running?
Aug 09 12:29:22.660 [warn] Failed to parse/validate config: Failed to bind one of the listener ports.
Aug 09 12:29:22.660 [err] Reading config failed--see warnings above.
```

Christewart:
needs rebase

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
Christewart: Since we start the binary here, don't we need to stop it as well? Otherwise a users machine when the JVM gets killed will still have tor running

PR #3484, opened 2021-08-04
Tor in Testkit

benthecarman:
Adds a `CachedTor` to the testkit so we can startup tor for tests automatically

### DIFF
--- a/dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTorTest.scala
+++ b/dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTorTest.scala
@@ -5,11 +5,12 @@ import akka.testkit.{TestActorRef, TestProbe}
 import org.bitcoins.asyncutil.AsyncUtil
 import org.bitcoins.core.number.UInt16
 import org.bitcoins.core.protocol.tlv.{LnMessage, PingTLV, PongTLV}
-import org.bitcoins.core.util.EnvUtil
-import org.bitcoins.core.util.NetworkUtil._
 import org.bitcoins.dlc.node.peer.Peer
 import org.bitcoins.rpc.util.RpcUtil
+import org.bitcoins.server.BitcoinSAppConfig
+import org.bitcoins.testkit.tor.CachedTor
 import org.bitcoins.testkit.util.BitcoinSActorFixtureWithDLCWallet
+import org.bitcoins.testkit.util.TorUtil._
 import org.bitcoins.testkit.wallet.FundWalletUtil.FundedDLCWallet
 import org.bitcoins.tor.{Socks5ProxyParams, TorController, TorProtocolHandler}
 import org.scalatest.{Assertion, FutureOutcome}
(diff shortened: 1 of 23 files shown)


testkit/src/main/scala/org/bitcoins/testkit/tor/CachedTor.scala
```diff
@@ -0,0 +1,57 @@
...
+  protected val isTorStarted: AtomicBoolean = new AtomicBoolean(false)
+
+  protected lazy val torF: Future[TorClient] = {
+    val _ = isTorStarted.set(true)
+
+    val torClient = torConfig.createClient
+
+    for {
+      _ <- torClient.startBinary()
+      // wait for tor to startup
```
Christewart: Is there any way we can actually test this with some sort of ping like we do with bitcoind rpc? Frankly there is probably a race condition between starting the tor process and tor actually being ready to do stuff that is causing things like #3547

testkit/src/main/scala/org/bitcoins/testkit/tor/CachedTor.scala
```diff
@@ -0,0 +1,57 @@
...
+      _ <- TestAsyncUtil.nonBlockingSleep(5.seconds)
```
Christewart: If we can't actually test this, we need to move all this logic into `TorAppConfig.start()` including this non blocking sleep that you had to add. Obviously this ended up being a problem in test cases, which means it is a problem for our users (i'm guessing #3547)

testkit/src/main/scala/org/bitcoins/testkit/util/TorUtil.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.net.InetSocketAddress
+import scala.util.Properties
+
+object TorUtil extends Logging {
+
+  val torEnabled: Boolean = Properties
+    .envOrNone("TOR")
+    .isDefined
+
+  def torProxyAddress = new InetSocketAddress("localhost", 9050)
```
Christewart: nit: Can the ports be factored out into this file and then referenced elsewhere in the code base with `TorUtil.PROXY_PORT` and `TorUtil.CONTROL_PORT`

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTorTest.scala
```diff
@@ -20,27 +21,21 @@ import java.net.InetSocketAddress
...
+  implicit val conf: BitcoinSAppConfig = getFreshConfig
+
   override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
-    // Skip on CI as we don't have access to tor control port
-    if (EnvUtil.isCI) {
-      FutureOutcome.succeeded
-    } else withFundedDLCWallet(test, getBIP39PasswordOpt())(getFreshConfig)
+    if (TorUtil.torEnabled) {
+      withFundedDLCWallet(test, getBIP39PasswordOpt())
+    } else FutureOutcome.succeeded
```
Christewart, line 35: Hmm, why succeed? Isn't it implied that this test case _needs_ tor to be able to run? It seems like we should fail loudly if tor is not enabled? Otherwise you have trivial silent failures?

node-test/src/test/scala/org/bitcoins/node/NeutrinoUnsupportedPeerTest.scala
```diff
@@ -16,13 +17,21 @@ class NeutrinoUnsupportedPeerTest extends NodeTestWithCachedBitcoindV19 {
...
-      outcome = withNeutrinoNodeUnstarted(test, bitcoind)(system,
-                                                          getFreshConfig)
-      f <- outcome.toFuture
-    } yield f
-    new FutureOutcome(outcomeF)
+    if (TorUtil.torEnabled) {
+      // We must skip this test for tor enabled
+      // because bitcoind only supported tor v2 at the time
+      // which is now deprecated and no longer supported by
+      // the tor network
```
Christewart, line 24: my suggestion here so we don't have this trivially succeed in the future if we change versions 

```
      // We must skip this test for tor enabled
      // because bitcoind only supported tor v2 at the time
      // which is now deprecated and no longer supported by
      // the tor network

      val f: Future[Outcome] = for {
        bitcoind <- cachedBitcoindWithFundsF
        _ = require(
          bitcoind.version == BitcoindVersion.V19,
          s"Expected bitcoind version 19, got version=${bitcoind.version}." +
            s"If this version is >= 21, we need to remove this if clause as bitcoind 21+ supports tor v2"
        )
        trivialSucceed <- FutureOutcome.succeeded.toFuture
      } yield trivialSucceed

      new FutureOutcome(f)
```

Although now i think about it, all these new versions will have neutrino supported, so feel free to omit

Christewart (changes requested):

dlc-node-test/src/test/scala/org/bitcoins/dlc/node/DLCServerTorTest.scala
benthecarman: We just want to skip this test if tor isn't enabled. I think this is the one way to do it, maybe can add a log

Christewart:
Compile failure: https://github.com/bitcoin-s/bitcoin-s/pull/3484/checks?check_run_id=3325001451#step:5:133

Christewart:
needs rebase and this tor test case seems to be failing consistently: https://github.com/bitcoin-s/bitcoin-s/pull/3484/checks?check_run_id=3334207653#step:5:711

Christewart:
needs rebase

PR #3485, opened 2021-08-04
Evenly space footer status items

user411:
Setup footer status bar items to evenly space over any window size

Doesn't functionally do much right now, but will allow for wide range of window sizing with evenly spaced labels

Screen Shot 2021-08-04 at 3 14 03 PM
[Other UI change not relevant]


benthecarman (approved):
I tried so long trying to figure this out originally, thank you

Christewart (approved):

PR #3492, opened 2021-08-05
Move to GridPane wallet, dynamic sizing

user411:
Wallet layout changes and some dynamic sizing to support more window sizes

Screen Shot 2021-08-05 at 2 18 22 PM

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
@@ -34,7 +34,6 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
 
   val DEFAULT_WIDTH = 1400.0
   val DEFAULT_HEIGHT = 800.0
-  val DEFAULT_MAXIMIZED = false
   val NOT_SET = Double.MinValue
   val MINIMUM_WIDTH = 800.0
   val MINIMUM_HEIGHT = 500.0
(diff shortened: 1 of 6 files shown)


app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
```diff
@@ -34,7 +34,6 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
 
   val DEFAULT_WIDTH = 1400.0
   val DEFAULT_HEIGHT = 800.0
-  val DEFAULT_MAXIMIZED = false
```
user411, line 37: Code cleanup, this preference didn't end up getting persisted

core/src/main/scala/org/bitcoins/core/dlc/accounting/RateOfReturnUtil.scala
```diff
@@ -7,4 +7,10 @@ object RateOfReturnUtil {
     val percent = ror * 100
     f"${percent}%1.2f" + "%"
   }
+
```
benthecarman, line 9: can make the above function now:

```scala
def prettyPrint(ror: BigDecimal): String = {
    toPercentage(ror, 2)
  }
```

benthecarman (commented):
otherwise lgtm

core/src/main/scala/org/bitcoins/core/dlc/accounting/RateOfReturnUtil.scala
user411: Added commit to reuse toPercentage()

benthecarman (approved):

PR #3493, opened 2021-08-05
Add where tor.md modifications need to be made explicitly

Christewart:
A couple of users so far have asked me where they need to put these changes. Since the `$HOME/.bitcoin-s/bitcoin-s.conf` doesn't get created if you don't have any custom settings, it is confusing where to put the configuration.


benthecarman (commented):
I think it'd be better to say 'in your bitcoin-s config file' and then link to our configuration docs.

PR #3498, opened 2021-08-06
2021 08 06 app config refacotr

Christewart:
This makes it so projects like `keyManager` and `tor` projects don't get all these database drivers included in their classpath. 

This does not change any functionality, strictly a move only PR.


benthecarman (approved):

benthecarman (changes requested):
compile failing

benthecarman (approved):

benthecarman (approved):

PR #3503, opened 2021-08-09
Update configuration.md

shruthii625:


docs/config/configuration.md
```diff
@@ -128,7 +128,9 @@ bitcoin-s {
     bitcoind-rpc {
         # bitcoind rpc username
         rpcuser = user
-        # bitcoind rpc password
+        # bitcoind rpc password.
+        # If you're generating the rpc password using rpcauth.py, then there are chances that the password generated might contain an '='.
+        # Please refrain from using such passwords.
```
Christewart: I don't think we want to encourage users to refrain from using them, i think we just need to tell them to wrap them in `"` no?

shruthii625: Oh, wrapping them in "" was why my authentication failed

shruthii625: Like initially I hadn't wrapped the password in "" and I got an error while the config file was being parsed. So I added the double quotes and then authentication failed

Christewart: Hmm. Ok that is interesting. Did you try escaping the characters with a `\=` or something like that? It would be a good idea to try and understand the limitations. 

This is the config library we use, it is a derivative of JSON, so i think you should be able to scape with a `\` in the string IIRC. 

https://github.com/lightbend/config/blob/main/HOCON.md

Since you rightly used `rpcauth.py` to generate credentials, we should add a comment specifying how to properly use that in conjunction with our software. Every output from `rpcauth.py` has a `=` at the end I believe


Screenshot from 2021-08-10 06-39-57

shruthii625: Sorry, but actually, just enclosing it in double quotes seem to work. I have no idea what I was doing wrong earlier.

Christewart (approved):

PR #3504, opened 2021-08-09
GUI code cleanup

user411:
There are no functional changes here, just layout cleanup, add/remove typings, remove unused fileChooserButton()

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/DLCDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/DLCDialog.scala
@@ -10,10 +10,9 @@ import scalafx.geometry.Insets
 import scalafx.scene.Node
 import scalafx.scene.control._
 import scalafx.scene.layout.GridPane
-import scalafx.stage.{FileChooser, Window}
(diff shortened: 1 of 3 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/DLCDialog.scala
```diff
@@ -10,10 +10,9 @@ import scalafx.geometry.Insets
 import scalafx.scene.Node
 import scalafx.scene.control._
 import scalafx.scene.layout.GridPane
-import scalafx.stage.{FileChooser, Window}
+import scalafx.stage.{Window}
```
benthecarman: nit: can remove the braces here

benthecarman (approved):

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/DLCDialog.scala
user411: done

benthecarman (approved):

Christewart (approved):

PR #3505, opened 2021-08-09
SplitPane flat Dialog UI

user411:
Moves UI to a SplitPane based main view with Dialog content in the main view.

Screen Shot 2021-08-09 at 11 22 54 AM

A new event can be pasted into the 'New Event Hex' textfield to load a New Offer pane. A contract being exchanged to create a DLC can be pasted into the 'Contract Hex' textfield or selected from the 'Browse...' button in the Contracts TitledPane.

DLC specific operations are accessible in an additional window:

Screen Shot 2021-08-09 at 11 30 38 AM
Screen Shot 2021-08-09 at 11 23 07 AM

The DLC window is closable with OS standard key commands.

Here's what creating a new Offer from Event Hex looks like after Executing the operation:

Screen Shot 2021-08-09 at 11 24 40 AM

### DIFF
--- a/app/gui/src/main/resources/themes/dark-theme.css
+++ b/app/gui/src/main/resources/themes/dark-theme.css
@@ -66,10 +66,22 @@
     -fx-border-width: 1px;
 }
 
-/* Accordion Group heading content */
+/* Accordion TitledPane heading content */
 .small {
     -fx-font-size: 0.8em;
 }
+
+/* Accordion TitledPane heading textfield */
+.title-textfield {
+    -fx-pref-width: 90;
+}
+
+/* Adjust divider visibility and padding */
+/*
+.split-pane > .split-pane-divider {
+    -fx-padding: 0;
+}
+*/
 
 /* Right content area */
 .scroll-pane {
--- a/app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
@@ -1,6 +1,5 @@
 package org.bitcoins.gui
 
-//import akka.actor.ActorSystem
 import org.bitcoins.cli.CliCommand.{
   AcceptDLCCliCommand,
   AddDLCSigsAndBroadcastCliCommand,
@@ -47,10 +46,6 @@ import scala.util.{Failure, Success}
 
 class ContractGUI(glassPane: VBox) {
 
-  def fetchStartingData(): Unit = {
-    // Nothing to fetch for state yet...
-  }
-
   private[gui] lazy val model = new ContractGUIModel()
 
   private lazy val addEventTF = new TextField {
@@ -84,8 +79,8 @@ class ContractGUI(glassPane: VBox) {
     styleClass += "title-textfield"
     promptText = "Contract Hex"
     onKeyTyped = _ => {
-      val bool = onContractAdded(text.value.trim, None)
-      if (bool) clearContractTF() // Clear on valid data
+      val validAddition = onContractAdded(text.value.trim, None)
+      if (validAddition) clearContractTF() // Clear on valid data
       ()
     }
   }
@@ -96,8 +91,8 @@ class ContractGUI(glassPane: VBox) {
 
   private lazy val fileChooserButton = GUIUtil.getFileChooserButton(file => {
     val hex = Files.readAllLines(file.toPath).get(0)
-    val bool = onContractAdded(hex, Some(file))
-    if (bool) addContractTF.clear()
+    val validAddition = onContractAdded(hex, Some(file))
+    if (validAddition) clearContractTF() // Clear on valid data
   })
 
   lazy val addContractHBox = new HBox {
@@ -177,24 +172,25 @@ class ContractGUI(glassPane: VBox) {
 
   // Text paste / file browse handler
   private def onContractAdded(hex: String, file: Option[File]): Boolean = {
-    var success = true
     // Accept / Sign / Broadcast identifier
     LnMessageFactory(DLCOfferTLV).fromHexT(hex) match { // Accept
       case Success(_) =>
         showAcceptOfferPane(hex)
+        true
       case Failure(_) =>
         LnMessageFactory(DLCAcceptTLV).fromHexT(hex) match { // Sign
           case Success(_) =>
             showSignDLCPane(hex, file)
+            true
           case Failure(_) =>
             LnMessageFactory(DLCSignTLV).fromHexT(hex) match { // Broadcast
               case Success(_) =>
                 showBroadcastDLCPane(hex, file)
-              case Failure(_) => success = false // Nothing else to check for
+                true
+              case Failure(_) => false // Nothing else to check for
             }
         }
     }
-    success
   }
 
   private lazy val contractStepPane: VBox = new VBox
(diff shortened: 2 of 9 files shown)


Christewart:
Compile is failing because Scala `2.12.x` and Scala `2.13.x` are not completely compatible with each other. You can switch to the `2.12.x` compiler by typing `++2.12.14` in your sbt shell, or do `+test:compile` which will compile the project with both versions of Scala we support.

benthecarman:
Overall really good!! This looks amazing

Some design feedback from testing:

1) Before you click on an event the right side is empty and it's unclear what is going on. Should add text saying something like "Click an event to view it" or just auto select an event on startup

2) In the view dlc sidebar `Counter Party Payout` is the longest label but isn't present for DLCs that havent been closed. This results in the text shifting back and forth we selecting between closed and unclosed DLCs. Should add a `minWidth` to the labels so this doesn't happen

3) The DLC operations are hard to find if you don't already know where they are. I think it'd be worth adding the buttons to the `Wallet` accordion (probably just offer and accept since we will be able to get rid of sign and broadcast with the p2p sutff now)

    3a) Since we have the execute button in the DLC view, we should probably but the refund button as well.

    3b) If we do this we can also remove the `contract id` field in the dialog and just pass it in the background

4) For the left side I think it'd make more sense to have the wallet accordion on top, that way your balance and sending and receving are more in your face.

5) For the events table, should make the `Event Id` col larger, and should add the oracle public key. Also instead of calling `contractInfo.toString` should call `contractInfo.hex`

6) When creating an offer on the right side from the events table, I think the `Execute` button is confusing, would make more sense imo to have it just say `Create Offer` or even just `Ok`, we already use `Execute` to mean closing the DLC.

7) I don't understand what the `Browse` and text field do at the top left of the Contract tabs. I tried inputting a `ContractInfo` like the Events tab but did not work

8) Looks like the events table doesn't persist between runs

app/gui/src/main/resources/themes/dark-theme.css
```diff
@@ -66,11 +66,23 @@
...
+/* Accordion TitledPane heading textfield */
+.title-textfield {
+    -fx-pref-width: 90;
+}
+
+/* Adjust divider visibility and padding */
+/*
+.split-pane > .split-pane-divider {
+    -fx-padding: 0;
+}
```
benthecarman, line 83: deletable?

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+import java.nio.file.Files
+import java.text.SimpleDateFormat
+import java.util.Date
+import scala.util.{Failure, Success}
+
+class ContractGUI(glassPane: VBox) {
+
+  def fetchStartingData(): Unit = {
+    // Nothing to fetch for state yet...
+  }
```
benthecarman: can delete?

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
+package org.bitcoins.gui
+
+//import akka.actor.ActorSystem
```
benthecarman: ping

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+  lazy val addEventHBox = new HBox {
+    styleClass += "small"
+    children = Seq(addEventTF)
+  }
+
+  private lazy val addContractTF = new TextField {
+    styleClass += "title-textfield"
+    promptText = "Contract Hex"
+    onKeyTyped = _ => {
+      val bool = onContractAdded(text.value.trim, None)
```
benthecarman: nit: should use a more descriptive name than `bool`

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  private def clearContractTF(): Unit = {
+    addContractTF.clear()
+  }
+
+  private lazy val fileChooserButton = GUIUtil.getFileChooserButton(file => {
+    val hex = Files.readAllLines(file.toPath).get(0)
+    val bool = onContractAdded(hex, Some(file))
+    if (bool) addContractTF.clear()
```
benthecarman: same here

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+    String] {
+    text = "Event Id"
+    prefWidth = 160
+    cellValueFactory = { status =>
+      val eventIdStr = status.value._1.eventTLV.eventId
+      new StringProperty(status, "Event Id", eventIdStr)
+    }
+  }
+
+  private val ISO_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'")
```
benthecarman, line 114: we use a different format in the `ViewDLCDialog` should stick to one universally. Imo ISO is good for passing around data but not the most human readable

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+  private lazy val contractCol = new TableColumn[
+    (OracleAnnouncementV0TLV, Option[ContractInfoV0TLV]),
+    String] {
+    text = "Contract"
+    prefWidth = 160
+    cellValueFactory = { status =>
+      val contractStr =
+        status.value._2 match {
+          case None        => ""
+          case Some(value) => value.contractDescriptor.toString()
```
benthecarman, line 137: should use `_.hex` here instead of `toString`

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
```diff
@@ -0,0 +1,274 @@
...
+          case Failure(_) =>
+            LnMessageFactory(DLCSignTLV).fromHexT(hex) match { // Broadcast
+              case Success(_) =>
+                showBroadcastDLCPane(hex, file)
+              case Failure(_) => success = false // Nothing else to check for
+            }
+        }
+    }
+    success
+  }
```
benthecarman: instead of using a `var` here you can just make every case return a `true` and the final failure return a `false`

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
Christewart: `isContractAdded`?

Christewart (commented):
Looking really good! Once ben's review is address let's get this thing merged!

app/gui/src/main/resources/themes/dark-theme.css
user411: I would leave this one in case we want to play with it. It took some googling to find.

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
user411: Yeah, I had this in to match pattern assuming it would probably see some use later. I can remove it.

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
user411: I used ISO because they visually order well versus weekday format.

app/gui/src/main/scala/org/bitcoins/gui/ContractGUI.scala
user411: I used toString because I wanted to see the JSON structure. I don't know that either are really worth presenting to the user in string form.

PR #3507, opened 2021-08-09
Update tor.md

sambradbury:
Typo correction and further clarification for mac users

### DIFF
--- a/docs/tor/tor.md
+++ b/docs/tor/tor.md
@@ -95,7 +95,6 @@ them):
 ControlPort 9051
 CookieAuthentication 1
 CookieAuthFileGroupReadable 1
-CookieAuthFile /usr/local/var/tor/control_auth_cookie
 ```
 
 Add or uncomment those, save, and restart Tor (usually `systemctl restart tor`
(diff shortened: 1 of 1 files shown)


docs/tor/tor.md
```diff
@@ -88,13 +88,14 @@ config, as well as have tor configured for it.
...
 You may need to set up the Tor Control Port. On Linux distributions there may be some or all of the following settings
-in `/etc/tor/torrc` for linux or `/opt/homebrew/etc/tor/torcc` for mac, generally commented out by default (if not, add
+in `/etc/tor/torrc` for linux or `/opt/homebrew/etc/tor/torrc` for mac, generally commented out by default (if not, add
 them):
 
 ```
 ControlPort 9051
 CookieAuthentication 1
 CookieAuthFileGroupReadable 1
+CookieAuthFile /usr/local/var/tor/control_auth_cookie
```
benthecarman: we should probably move these 2 setting to an optional setting section as they aren't always needed. Also the path should be different for each os

PR #3512, opened 2021-08-10
WIP: Peer discovery

solo1g:
This PR adds support for finding peers through dns seeds and addr messages, storing them in our database and using them for subsequent node runs. A new key `maxConnectedPeers` at path `bitcoin-s.node.maxConnectedPeers` in `bitcoin-s.config` needs to be added with a value>1 to use multiple connections, absence of which keeps it to the current default of 1 connection.  Without this key, none of original behavior is intended to change other than storing peers in db.

### Details

I will roughly explain all that I did here. So, ultimately our aim is to quickly get peers that support compact filters from addr messages and dns seeds. 

What we do for this is to first start our node with reliable peers (taking from config and then from db uptil `maxConnectedPeers`) and on initialization a request for addr messages is sent to those. A stack `peersToCheckStack` is maintained and on receiving any peers through addr messages, they are added to that stack. Now while we do get services supported with the addr message, they may be outdated, as in reporting support for compact filters in addr but not in the version msg that we get while initializing it. And I found this to be an issue for quite some cases so only way to go was to check if they show support in addr messages and then initialize it ourselves to verify it, which we do for peers from dns seeds since that's the only way to get `ServiceIdentifier` for them.

Now, we also have a scheduler that runs after every fixed interval right from start of node. On each run, it takes a certain number of peers from the stack and initializes them, if they are found to support compact filters then firstly we add them to our database and then if the number of peers that we are connected to (determined by `keepConnection`, I will come to this later) are less than `maxConnectedPeers` then we keep that connection and mark `keepConnection` as true else we disconnect from it.

Now use for `keepConnection` is for identifying which peers are the ones that we have to use in our node from the other temporary connections that we establish for peer discovery (since both get added to same `peerData` map). The peers which we select from start, the "reliable" ones from conf and then db, are added to `peerData` as `keepConnection=true` and for the peer discovery part, they are added with `keepConnection` as false, meaning we can disconnect from them. It is used for getting a random peer for use in node's operations as we filter what peers have `keepConnection` as true and also to determine reconnection attemps, as we would want to reconnect peers from conf but not the other temporary ones since with many of them connection are offline and reconnecting would be useless. This is also why `maxConnectionTries` was made an argument.

A major refactoring change done is addition of `ControlMessageHandler` (similar to `DataMessageHandler`) that moves the logic related to handling of `ControlPayload` from `PeerMessageReceiver` to `ControlMessageHandler` to keep it more in line with how things are done for `DataPayload`.

One last thing as for using a stack to store peers to check, is because I found addr messages to be more reliable and faster way of getting required peers (dns seeds are only used if stack is empty), and addr messages are received continuously while node operation so the top would usually be a peer from addr message.
Also about the need for a scheduler and stack. This was added so that we are only trying a limited number of connections at a time. Trying all 50+ peers at the same instant led to many errors due to race conditions so that's why scheduler was needed.


Christewart:
So one of the frustrating things about Scala is the incompatabilities between versions `2.12.x` and `2.13.x`. In the `2.13.x` they redesigned the collections in the Scala standard library. This means that some methods that were added (such as `appendAll()`) and are not available in the `2.12.x` release of Scala. 

https://github.com/bitcoin-s/bitcoin-s/pull/3512/checks?check_run_id=3290321734#step:5:161

You can cross compile both versions locally by doing `+test:compile` or just switch your sbt session to use the 2.12 compiler by typing `++2.12.14` in your sbt shell.

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -88,7 +90,42 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
     }
 
     val peerSockets = {
-      nodeConf.peers.map(
+      val allPeers = {
+        if (nodeConf.network == MainNet) {
+          val dnsSeeds = Vector(
+            "seed.bitcoin.sipa.be",
```
Christewart: You should be able to use the seeds encoded in `NetworkParameters`. As a side note, that file hasn't been updated in a long time, so some of the seeds may need to be updated/replaced in `NetworkParameters.dnsSeeds` 

https://github.com/bitcoin-s/bitcoin-s/blob/8f958bf5d9957438d703819a087d87813f3edd89/core/src/main/scala/org/bitcoins/core/config/NetworkParameters.scala#L34

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -44,24 +45,62 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+
+    lazy val client: P2PClient = {
+      val peerMessageReceiver =
+        PeerMessageReceiver.newReceiver(node = Node.this, peer = peer)
+      P2PClient(context = system,
+                peer = peer,
+                peerMessageReceiver = peerMessageReceiver)
+    }
+
+    private var _serviceIdentifier: ServiceIdentifier = _
```
Christewart: Don't do `_` when initializing a variable, it's equivalent to using `null` which leads to `NullPointerException`. Just do 

```scala
private var _serviceIdentifierOpt: Option[ServiceIdentifier] = None
```

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -159,30 +187,37 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+              case NodeType.NeutrinoNode =>
+                if (peerData(peer).serviceIdentifier.nodeCompactFilters) {
+                  logger.info(s"Our peer=$peer has been initialized")
+                } else {
+                  logger.info(
+                    s"Our peer=$peer does not support compact filters. Disconnecting.")
+                  peerData(peer).peerMessageSender.disconnect()
+                }
+              case NodeType.SpvNode         =>
+              case NodeType.BitcoindBackend =>
```
Christewart: Seems like an error should be thrown in this case, it doesn't really make sense `BitcoindBackend` to be a node

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -159,30 +187,37 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+                maxTries = 50,
+                interval = 250.millis)
+              .recover { case NonFatal(_) =>
+                logger.info(s"Failed to initialize $peer")
+                peerData(peer).peerMessageSender.disconnect()
+                removePeer(peer)
               }
+          } yield ()
+        isInitializedF.map { _ =>
+          if (peerData.contains(peer)) {
```
Christewart: is is possible that `peerData` _doesn't_ contain the peer?

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: Yes, there were peers we could connect to but not initialize, in which case they were removed, so it might happen that this map would be called corresponding to a removed peer.

solo1g:
Added basic db features. No tests yet. The new table is very simple at the moment with only two fields, the address and last connected time, only peers having compact filters are added and on next run they too will be used to connect. As dns seeds for mainnet only so currently both peer discovery and db related operations are only done for mainnet.

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.bitcoins.db.{CRUD, SlickUtil}
+import org.bitcoins.node.config.NodeAppConfig
+import slick.lifted.ProvenShape
+
+import java.time.Instant
+import scala.concurrent.{ExecutionContext, Future}
+
+case class PeerDB(
+    address: String,
+    lastConnected: Instant
```
Christewart: maybe add `firstSeen`?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -159,30 +183,42 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+            nodeAppConfig.nodeType match {
+              case NodeType.NeutrinoNode =>
+                if (peerData(peer).serviceIdentifier.nodeCompactFilters) {
+                  logger.info(s"Our peer=$peer has been initialized")
+                  if (nodeAppConfig.network == MainNet) {
```
Christewart: why only on mainnet?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -159,30 +183,42 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
-            case NodeType.NeutrinoNode => {
-              if (peerServices(peers(idx)).nodeCompactFilters) {
-                logger.info(s"Our peer=${peers(idx)} has been initialized")
-              } else {
-                logger.info(
-                  s"Our peer=${peers(idx)} does not support compact filters. Disconnecting.")
-                peerMsgSenders(idx).disconnect()
+      logger.info(peers)
+      val isInitializedFs = peers.map { peer =>
+        peerData(peer).peerMessageSender.connect()
```
Christewart: Can we move this lambda implementation into a helper method with a nice name and possibly scaladoc? I.e. 

```scala
private def initializePeer(peer): Future[Unit] = { ... }

private def createInDbIfBlockFilterPeer: Future[Option[PeerDb]] = { ... } 

```

it always makes code so much more readable with nicely named methods :-)

Christewart (commented):
So this is looking really good, the only thing that is missing is a test with bitcoind nodes. In my mind, you should 

1. Spin up two connected bitcoind nodes
2. Connect the neutrino node to one of the bitcoind nodes
3. Have that bitcoind node gossip the other bitcoind nodes address
4. Store the bitcoind node's address that you _aren't_ directly connected to
5. Assert that address is stored in the database

This is pushing up against my understanding of the bitcoin p2p network, I may be misunderstanding step `3` as I'm unsure of how normal nodes gossip addresses. All of this may not be possible if they _dont_ gossip addresses and all you can do is receive them from a seed.

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
benthecarman: may want to add `networkId` for differentiating tor vs clearnet vs i2p, etc

https://github.com/bitcoin-s/bitcoin-s/blob/1c6d728197e7276e319b912a184311ae493b02fe/core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala#L506

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: ~Dns seeds only give addresses for mainnet. I could only connect to 8333 port not 18333.~
This was wrong, changed that.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -177,9 +177,22 @@ class PeerMessageReceiver(
...
-      case _: AddrMessage =>
+      case addr: AddrMessage =>
+        /* For this, storing is fine, it gives ipv4 addresses the normal ones (88.232.34.12 type) same as the ones
+        given by dns seeds. No problem in this so far so good.
+         */
+        addr.addresses.foreach(node.createInDbIfBlockFilterPeer)
         Future.successful(this)
-      case _: AddrV2Message =>
+      case addr: AddrV2Message =>
+        /* In the case of V2 messages, messages are of a wider variety, ipv6, torv2, torv3 etc usually stored as bytes
```
solo1g: For the test you mentioned, for gossiping addresses among nodes, storing peers from addr messages would be required but I'm having trouble with the different types. @Christewart

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -147,6 +170,90 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+    Future.unit
+  }
+
+  def createInDbIfBlockFilterPeer(networkAddress: NetworkIpAddress): Unit = {
+    val ipv4bytes: Option[ByteVector] =
+      try {
+        Some(networkAddress.address.ipv4Bytes)
+      } catch {
+        case _: Throwable => {
+          logger.info("Ignoring ipv6 address from addr message")
```
benthecarman: why do we ignore ipv6 addresses?

node-test/src/test/scala/org/bitcoins/node/models/PeerTableTest.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  behavior of "PeerDAO"
+
+  it must "write a peer and read it back" in { daos =>
+    val peerDAO = daos.peerDAO
+    //using a hardcoded peer for now
+    val peer = PeerDB(address = "127.0.0.1",
+                      lastSeen = Instant.now,
+                      firstSeen = Instant.now,
+                      networkId = AddrV2Message.IPV6_NETWORK_BYTE)
```
benthecarman: nit: `"127.0.0.1"` is a IPV4 addr

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -90,25 +93,68 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
...
-    if (nodeConf.peers.isEmpty) {
-      throw new IllegalArgumentException(
-        "No peers specified, unable to start node")
-    }
 
-    val peerSockets = {
-      nodeConf.peers.map(
-        NetworkUtil.parseInetSocketAddress(_, nodeConf.network.port)
-      )
+    val peerSocketsF = {
```
benthecarman: this should probably be done inside the `Node`, we can probably change it to not take `peers` as a param now

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.bitcoins.db.{CRUD, SlickUtil}
+import org.bitcoins.node.config.NodeAppConfig
+import slick.lifted.ProvenShape
+
+import java.time.Instant
+import scala.concurrent.{ExecutionContext, Future}
+
+case class PeerDB(
+    address: String,
+    lastSeen: Instant,
```
benthecarman, line 14: should add `port` as well

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  override protected def findAll(
+      ts: Vector[PeerDB]): Query[Table[_], PeerDB, Seq] = findByPrimaryKeys(
+    ts.map(_.address))
+
+  def upsertPeer(
+      address: String,
+      lastSeen: Instant = Instant.now,
+      networkId: Byte = AddrV2Message.IPV4_NETWORK_BYTE): Unit = {
+    logger.info(s"Adding peer to db $address")
```
benthecarman: nit: should probably be debug or trace

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -202,9 +202,14 @@ class PeerMessageReceiver(
...
         Future.successful(this)
-      case _: AddrMessage =>
+      case addr: AddrMessage =>
+        addr.addresses.foreach(node.createInDbIfBlockFilterPeer)
         Future.successful(this)
-      case _: AddrV2Message =>
+      case addr: AddrV2Message =>
+        addr match {
+          case ipv4: IPv4AddrV2Message => node.createInDbIfBlockFilterPeer(ipv4)
+          case _                       => logger.warn("Unsupported address type")
```
benthecarman: we should be able to support ipv4, ipv6, and torv3

node/src/main/scala/org/bitcoins/node/Node.scala
benthecarman: also nit: don't need braces around this

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -147,6 +170,90 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+      }
+    val stringAddress = ipv4bytes.get.toArray
+      .map(x => {
+        val short = x.toShort
+        if (short < 0) x + 256
+        else short
+      })
+      .mkString(".") + s":${networkAddress.port}"
+    if (networkAddress.services.nodeCompactFilters) {
+      logger.info(s"Peer from addr: $stringAddress supports compact filters.")
```
benthecarman: this should be `debug` or `trace`

benthecarman (changes requested):
I think it'd be a lot cleaner if we had a separate file called `PeerHandling` or something of that sort and had all the `createInDbIfBlockFilterPeer`. We could then make a function just called `handlePeer` and this would delegate to the correct handing function based off of `nodeType`, so in neutrino's case it would call `createInDbIfBlockFilterPeer`

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
benthecarman: Could be done in a follow up PR as this might be a pretty large change

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
benthecarman: it also might be easier to store the address as a `ByteVector`, that way we are able to support anything that is in an AddrV2Message

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -202,9 +202,14 @@ class PeerMessageReceiver(
         //we don't want to have to request them manually
         sender.sendHeadersMessage()
         Future.successful(this)
-      case _: AddrMessage =>
+      case addr: AddrMessage =>
+        addr.addresses.foreach(node.createInDbIfBlockFilterPeer)
```
solo1g: `AddrMessage` has addresses along with their service bytes. I noticed that for quite some cases, service bytes of an addr message differ from the service bytes we get from the version message when we try to connect it. So this leads to the problem of storing addresses in db, on the basis of addr messages, (only nodes with compact filters are stored) which end up not supporting compact filters when we try to connect to them later on. Should this behavior be possible?

benthecarman: Yeah, most likely the node has changed config since the last time that peer saw them. I think in that case we should just disconnect from them and find another one

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -90,25 +93,68 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
...
+        val peersFromDbF = PeerDAO().findAll().map(_.map(_.address))
+
+        val peersFromConf = nodeConf.peers.distinct
+
+        val all = for {
+          peersFromDB <- peersFromDbF
+        } yield {
+          logger.info(s"db peers $peersFromDB")
+          val maxPeers = 20
+          var ret: Vector[String] = Vector()
```
Christewart:
```scala
val builder = Vector.newBuilder
```

then when you are done 

`builder.result()`

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
Christewart: I agree with Ben's sentiment here. This doesn't like it should be the responsibility of the `appServer` module to select the peers.

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -90,25 +93,68 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
...
+          //choosing "maxPeers" no of elements from lists randomly in the order of peersFromConf, peersFromDB, peersFromSeed
+          ret = ret ++ Random.shuffle(peersFromConf).take(maxPeers)
+          ret = ret ++ Random
+            .shuffle(peersFromDB.diff(ret))
+            .take(maxPeers - ret.length)
+          //dns seeds won't be used if they are not required as startup time greatly increases because of them
```
Christewart: Yeah, i think dns seeds have a policy to disconnect you fairly quickly. Could you look through the seeder github to see if they state a policy. 

I believe this is the active github repo: https://github.com/sipa/bitcoin-seeder

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
Christewart: agree, `debug` seems best.

node/src/main/scala/org/bitcoins/node/models/PeerDAO.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def deleteByKey(address: String):Future[Int]={
+    val q=table.filter(_.address===address)
+    safeDatabase.run(q.delete)
+  }
+
+  def upsertPeer(
+      address: String,
+      lastSeen: Instant = Instant.now,
+      networkId: Byte = AddrV2Message.IPV4_NETWORK_BYTE): Unit = {
```
Christewart: is it a good idea to specify a default value for the argument? It seems like we shoudl require the caller of this method to specify this? 

I'm open to hearing arguments to the contrary

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -76,7 +76,7 @@ case class P2PClientActor(
 
   private var reconnectHandlerOpt: Option[() => Future[Unit]] = None
 
-  private val maxReconnectionTries = 16
+  private val maxReconnectionTries = 1
```
Christewart: Why was this changed?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -44,24 +70,33 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
-
-  private val _peerServices: mutable.Map[Peer, ServiceIdentifier] =
-    mutable.Map.empty
+  def peers: Vector[Peer] = peerData.keys.toVector
 
-  def peerServices: Map[Peer, ServiceIdentifier] = _peerServices.toMap
+  def addPeer(peer: Peer): Unit = {
+    if (!_peerData.contains(peer)) {
+      _peerData.put(peer, PeerData(peer, this))
+    }
```
Christewart: nit: maybe add `debug` or `warn` logs in the case that we already have the peer added? It seems like this might be interesting information

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -44,24 +70,33 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+  }
 
-  def setPeerServices(
-      peer: Peer,
-      serviceIdentifier: ServiceIdentifier): Unit = {
-    _peerServices.put(peer, serviceIdentifier)
+  def removePeer(peer: Peer): Unit = {
+    if (_peerData.contains(peer)) {
+      _peerData.remove(peer)
+    }
```
Christewart: same here with adding a `debug` or `warn` log in the `else` clause

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -107,22 +142,10 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
-    val zipped = peers.zip(peerMsgRecvs)
-    val p2p = zipped.map { case (peer, peerMsgRecv) =>
-      P2PClient(context = system,
-                peer = peer,
-                peerMessageReceiver = peerMsgRecv,
-                onReconnect = sync)
-    }
-    p2p
-  }
+  def clients: Vector[P2PClient] = peerData.values.map(_.client).toVector
```
Christewart: So this needed to be changed to a `def` as we can dynamically add or remove `P2PClient`s from the underlying `peerData` data structure right?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -147,6 +170,90 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
   def isDisconnected(idx: Int): Future[Boolean] =
     peerMsgSenders(idx).isDisconnected()
 
+  private def createInDbIfBlockFilterPeer(peer: Peer): Future[Unit] = {
```
Christewart: So this seems like it should belong in the `NeutrinoNode` class rather than the `Node` superclass. 

Ideally, we would like to keep generic bitcoin p2p stuff in the `Node` trait, while keeping neutrino specific stuff in the `NeutrinoNode` trait.

node/src/main/scala/org/bitcoins/node/Node.scala
Christewart: For some context of conversations I had with Shreyansh, i think we decided to limit scope and try to get something working to start, and then add other address types later on. 

Do you have differing opinions on this? Perhaps you and Shreyansh could colloborate how to best build the data model for this.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -147,6 +170,90 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+      })
+      .mkString(".") + s":${networkAddress.port}"
+    if (networkAddress.services.nodeCompactFilters) {
+      logger.info(s"Peer from addr: $stringAddress supports compact filters.")
+      PeerDAO().upsertPeer(stringAddress)
+    }
+    ()
+  }
+
+  def createInDbIfBlockFilterPeer(addr: IPv4AddrV2Message): Unit = {
```
Christewart: the same comment above applies to this, this is `Neutrino` specific, so we should move to `NeutrinoNode`

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
solo1g: Oh, this change is not complete. I suppose we would want to try to reconnect to the peers that are loaded from our config file and that we have connected to in the past and have compact filters, but not the ones that we are connecting to for the first time (from addr messages or dns seeds) to see if they support compact filters or complete handshake at all. P2P client does not differentiate this and so I'll add connection tries as an argument to it.

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: Yes

node/src/main/scala/org/bitcoins/node/Node.scala
benthecarman: Let me take a stab at it and see if my idea will work

node-test/src/test/scala/org/bitcoins/node/models/PeerTableTest.scala
```diff
@@ -0,0 +1,36 @@
+package org.bitcoins.node.models
+
+import org.bitcoins.core.p2p.AddrV2Message
+import org.bitcoins.testkit.fixtures.NodeDAOFixture
+import scodec.bits.ByteVector
+
+import java.time.Instant
+
+class PeerDAOTest extends NodeDAOFixture {
```
benthecarman: Should rename this file to `PeerDAOTest`

solo1g:
Haven't tested everything fully yet but this works fine for default config on clearnet. Changing tests related to `NeutrinoNode` as per the new changes is remaining.

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -25,7 +25,7 @@ import org.bitcoins.feeprovider.MempoolSpaceTarget.HourFeeTarget
 import org.bitcoins.feeprovider._
 import org.bitcoins.node._
 import org.bitcoins.node.config.NodeAppConfig
-import org.bitcoins.node.models.Peer
+import org.bitcoins.node.models.{Peer}
```
benthecarman: nit: no need for brackets here

core/src/main/scala/org/bitcoins/core/util/NetworkUtil.scala
```diff
@@ -17,6 +19,19 @@ abstract class NetworkUtil {
...
   }
 
+  def parseInetSocketAddress(
+      address: ByteVector,
+      port: Int): InetSocketAddress = {
+    val hostAddress = InetAddress.getByAddress(address.toArray).getHostAddress
+    val uri: URI = {
+      address.size match {
+        case 4  => new URI("tcp://" + hostAddress)
+        case 16 => new URI(s"tcp://[$hostAddress]")
```
benthecarman: Should use `AddrV2Message.IPV4_ADDR_LENGTH` here and for the others

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -45,6 +56,96 @@ case class NeutrinoNode(
...
+    val networkByte =
+      if (addrBytes.length == AddrV2Message.IPV4_ADDR_LENGTH) {
+        AddrV2Message.IPV4_NETWORK_BYTE
+      } else if (addrBytes.length == AddrV2Message.IPV6_ADDR_LENGTH) {
+        AddrV2Message.IPV6_NETWORK_BYTE
+      } else if (addrBytes.length == AddrV2Message.TOR_V3_ADDR_LENGTH) {
+        AddrV2Message.TOR_V3_NETWORK_BYTE
+      } else {
+        throw new IllegalArgumentException("Unknown peer network")
+      }
```
benthecarman: would be better to use a `match` here

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -10,28 +10,62 @@ import org.bitcoins.chain.models.{
...
+  private var _serviceIdentifier: Option[ServiceIdentifier] = None
+
+  def serviceIdentifier: ServiceIdentifier = {
+    _serviceIdentifier.getOrElse(
+      throw new RuntimeException("Service identifier not initialized"))
+  }
+
+  def setServiceIdentifier(serviceIdentifier: ServiceIdentifier): Unit = {
+    _serviceIdentifier = Some(serviceIdentifier)
+  }
```
benthecarman: `PeerData` should probably get its own file

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -44,24 +78,122 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+                                            Duration(5, TimeUnit.SECONDS)) {
+    new Runnable() {
+      override def run(): Unit = {
+        if (peersToCheckStack.size < 10)
+          peersToCheckStack.pushAll(getPeersFromDnsSeeds)
+        val peers = for { _ <- 0 to 9 } yield peersToCheckStack.pop()
+        peers.foreach(peer => {
+          addPeer(peer, keepConnection = false)
+          initializePeer(peer)
+        })
```
benthecarman: should make this runnable it's own object

benthecarman: cc @Christewart should this be added to the node app config?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -103,26 +235,22 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
                                     CompactFilterDAO())
   }
 
+  //todo think of a better name
+  def handlePeerGossipMessage(message: Any): Unit
```
benthecarman: We don't want to be using `Any`, should make a `sealed trait` that encapsulates `AddrMessage` and `AddrV2Message`

benthecarman: I think the function name is good though

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -202,9 +202,11 @@ class PeerMessageReceiver(
...
         sender.sendHeadersMessage()
         Future.successful(this)
-      case _: AddrMessage =>
+      case addr: AddrMessage =>
+        node.handlePeerGossipMessage(addr)
         Future.successful(this)
-      case _: AddrV2Message =>
+      case addr: AddrV2Message =>
+        node.handlePeerGossipMessage(addr)
         sender.sendSendAddrV2Message()
```
benthecarman: I think this `sender.sendSendAddrV2Message()` is a mistake by me from awhile back, should be under the 
`case SendAddrV2Message => `

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -202,9 +202,11 @@ class PeerMessageReceiver(
         //we don't want to have to request them manually
         sender.sendHeadersMessage()
         Future.successful(this)
-      case _: AddrMessage =>
+      case addr: AddrMessage =>
+        node.handlePeerGossipMessage(addr)
         Future.successful(this)
-      case _: AddrV2Message =>
+      case addr: AddrV2Message =>
+        node.handlePeerGossipMessage(addr)
```
benthecarman: When you make the super type for these 2, it might be better to do

```scala

case msg: GossipAddrMessage =>
	node.handlePeerGossipMessage(msg)
	
	msg match {
		case addr: AddrMessage =>
			...
		case addr: AddrV2Message =>
			...
    }

```

this way we always handle the gossip message and then the compile will yell at us here to handle it correctly

benthecarman (commented):
Looking really good! Some final review :)

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
benthecarman: This will allow us to receive addrv2 messages so we can test it too

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: I suppose you mean making it an object as in `object PeerScheduler extends Runnable` and then I would not be able to access members from node as objects don't take any arguments. Making a class would work but then it won't ensure it to be a singleton which is why we are making it an object in the first place (?). How would I go about it?

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -44,24 +41,132 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+            if (cnt > 10) peerConnectionScheduler.cancel())
+          if (peersToCheckStack.size < 8)
+            peersToCheckStack.pushAll(getPeersFromDnsSeeds)
+          val peers = for { _ <- 0 to 7 } yield peersToCheckStack.pop()
+          peers.foreach(peer => {
+            addPeer(peer, keepConnection = false)
+            initializePeer(peer)
+          })
+        }
+      }
```
solo1g: Numbers here are what I thought best. Limit of 10 to stop finding peers is arbitrary and we might want to change that. I put up a limit because otherwise it would go on indefinitely.

solo1g:
This should be ready now. It works fine mostly, though since we connect to a lot of peers occasionally #3619 shows up.

benthecarman:
> It works fine mostly, though since we connect to a lot of peers occasionally #3619 shows up.

Maybe a race condition from reconnecting?

core/src/main/scala/org/bitcoins/core/util/NetworkUtil.scala
```diff
@@ -17,6 +20,54 @@ abstract class NetworkUtil {
...
+            InetAddress.getByAddress(address.toArray).getHostAddress
+          new URI("tcp://" + hostAddress)
+        case AddrV2Message.IPV6_ADDR_LENGTH =>
+          val hostAddress =
+            InetAddress.getByAddress(address.toArray).getHostAddress
+          new URI(s"tcp://[$hostAddress]")
+        case AddrV2Message.TOR_V3_ADDR_LENGTH =>
+          val hostAddress = bytesToTorV3Address(address)
+          new URI("tcp://" + hostAddress)
+      }
```
Christewart, line 42: you need to handle the case where it isn't one of above lengths

```scala
case AddrV2Message.IPV6_ADDR_LENGTH => ...
case AddrV2Message.TOR_V3_ADDR_LENGTH=> ...
case unknownSize => 
    sys.error(s"Attempted to parse InetSocketAddress with unknown size, got=${unknownSize}")
```

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -112,7 +111,7 @@ case class P2PClientActor(
           handleEvent(message, peerConnection, unalignedBytes)
         context.become(awaitNetworkRequest(peerConnection, newUnalignedBytes))
       case P2PClient.CloseCommand =>
-        logger.info(s"disconnecting from peer $peer")
+        logger.debug(s"disconnecting from peer $peer")
```
Christewart: Why did all these logs get moved from `info` -> `debug`? It seems like these are worthy logs to have at the `info` level. Same with logs in `def reconnecting`

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -65,7 +65,8 @@ import scala.util._
 case class P2PClientActor(
     peer: Peer,
     initPeerMsgHandlerReceiver: PeerMessageReceiver,
-    onReconnect: () => Future[Unit]
+    onReconnect: () => Future[Unit],
+    private val maxReconnectionTries: Int
```
Christewart, line 69: Why did `maxReconnectionTries` get moved to be a parameter?

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -11,20 +11,34 @@ import org.bitcoins.core.api.chain.db.{
...
 
 case class NeutrinoNode(
-    nodePeer: Vector[Peer],
     private var dataMessageHandler: DataMessageHandler,
     nodeConfig: NodeAppConfig,
     chainConfig: ChainAppConfig,
-    actorSystem: ActorSystem)
-    extends Node {
+    actorSystem: ActorSystem,
+    confPeersOverride: Vector[Peer] = Vector()
```
Christewart: nit: `Vector.empty` rather than `Vector()`

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
```diff
@@ -111,6 +111,12 @@ case class NodeAppConfig(
     strs.map(_.replace("localhost", "127.0.0.1"))
   }
 
+  lazy val maxConnectedPeers: Int = {
```
Christewart, line 114: This needs to be documented in configuration.md (our example config file)

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -41,7 +41,7 @@ class PeerMessageReceiver(
           _: Disconnected) =>
         throw new RuntimeException(s"Cannot call connect when in state=${bad}")
       case Preconnection =>
-        logger.info(s"Connection established with peer=${peer}")
+        logger.debug(s"Connection established with peer=${peer}")
```
Christewart, line 44: Maybe switch this back to `info` ?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -202,12 +202,11 @@ class PeerMessageReceiver(
         //we don't want to have to request them manually
         sender.sendHeadersMessage()
         Future.successful(this)
-      case _: AddrMessage =>
-        Future.successful(this)
-      case _: AddrV2Message =>
-        sender.sendSendAddrV2Message()
+      case msg: GossipAddrMessage =>
+        node.handlePeerGossipMessage(msg)
```
Christewart: Small race condition here: 

```scala

node.handlePeerGosshipMessage(msg)
  .map(_ => this)
```

This is the same below too on 209

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala
```diff
@@ -46,13 +46,13 @@ case class PeerMessageSender(client: P2PClient)(implicit conf: NodeAppConfig)
   def disconnect()(implicit ec: ExecutionContext): Future[Unit] = {
     isConnected().flatMap {
       case true =>
-        logger.info(s"Disconnecting peer at socket=${socket}")
+        logger.debug(s"Disconnecting peer at socket=${socket}")
```
Christewart, line 49: Back to `info` maybe?

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -45,6 +57,98 @@ case class NeutrinoNode(
     this
   }
 
+  override def getPeersFromConf: Vector[Peer] = {
+    if (confPeersOverride.isEmpty) super.getPeersFromConf
+    else confPeersOverride
+  }
+
+  override def handlePeerGossipMessage(message: GossipAddrMessage): Unit = {
```
Christewart: Is there a reason this isn't in `DataMesageHandler`? It seems out place here. Or perhaps we need something called `ControlMessageHandler`? And then we can replicate this with a `ControlPayload` parameter rather than `DataPayload` parameter

https://github.com/shreyanshyad/bitcoin-s/blob/0746b1433189ea23894c74475a67750e98f6297f/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala#L49

That should make this more testable too

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -45,6 +57,98 @@ case class NeutrinoNode(
...
+              peersToCheckStack.push(peer)
+          case TorV3AddrV2Message(_, _, _, _) =>
+            if (nodeConfig.torConf.enabled && services.nodeCompactFilters)
+              peersToCheckStack.push(peer)
+          case _ => logger.debug("Unsupported network. Skipping.")
+        }
+    }
+  }
+
+  override def onPeerInitialization(peer: Peer): Future[Unit] = {
```
Christewart: I think this should be moved into the `ControlMessageHandler` i talked about above. 

Also, more generally, there are some race conditions in this code. You need to be very careful about not dropping futures on the floor like you are doing with

```peerData(peer).peerMessageSender.sendGetAddrMessage()```

In your currently implementation, that `Future[Unit]` returned by this method is dropped on the floor, this means if an exception is caused by the method `sendGetAddrMessage()` , the caller of this method (`onPeerInitialization()`) will not have the exception propogated up the stack. It will silently fail which makes it _really_ hard to debug.

I think this could be re-written as
```scala
    if (
      peerData.contains(peer) && peerData(
        peer).serviceIdentifier.nodeCompactFilters
    ) {
  
      val createdInDbF = createInDb(peer)
      val handlePeerF = if (
        peerData(
          peer).keepConnection || connectedPeersCount < maxConnectedPeers
      ) {
        val sendMsgF = peerData(peer).peerMessageSender.sendGetAddrMessage()
        //only the nodes that have keepConnection as true would be actually used by us
        peerData(peer).keepConnection = true
        logger.info(
          s"Connected to peer $peer with compact filters. Connected peer count $connectedPeersCount")
        sendMsgF
      } else {
        removePeer(peer)
        Future.unit
      }
  for {
        _ <- createdInDbF
       _ <- handlePeerF
      } yield {
    ()
      }
    } else {
      removePeer(peer)
      Future.unit
    }

```

Sorry for the poor formatting, hopefully the idea gets across.

node/src/main/scala/org/bitcoins/node/PeerData.scala
```diff
@@ -0,0 +1,45 @@
...
+  PeerMessageReceiver,
+  PeerMessageSender
+}
+
+import scala.concurrent.Future
+
+case class PeerData(
+    peer: Peer,
+    node: Node,
+    var keepConnection: Boolean
```
Christewart: public mutability is a bad idea. Can you make this private with a setter if it is needed? Also what does `keepConnection` mean in this context? A Scaladoc on the `PeerData` class would be useful to document `keepConnection`

Christewart (changes requested):

solo1g:
Ok, I will roughly explain all that I did here. So, ultimately our aim is to quickly get peers that support compact filters from addr messages and dns seeds. What we do for this is to first start our node with reliable peers (taking from config and then from db uptil `maxConnectedPeers`) and on initialization a request for addr messages is sent to those. A stack `peersToCheckStack` is maintained and on receiving any peers through addr messages, they are added to that stack. Now while we do get services supported with the addr message, they may be outdated, as in reporting support for compact filters in addr but not in the version msg that we get while initializing it. and I found this to be an issue for quite some cases so only way to go was to check if they show support in addr messages and then initialize it ourselves to verify it, which we do for peers from dns seeds since that's the only way to get `ServiceIdentifier` for them. Now, we have a scheduler that runs after every fixed interval right from start of node, and what it does is that it takes a certain number of peers from the stack and initializes them, if they are found to support compact filters then firstly we add them to our database and then if the number of peers that we are connected to (determined by `keepConnection`, I will come to this later) are less than `maxConnectedPeers` then we keep that connection and mark `keepConnection` as true else we disconnect from it. Now use for `keepConnection` is for identifying which peers are the ones that we have to use in our node from the other temporary connections that we establish for peer discovery (since both get added to same `peerData` map). The peers which we select from start, the "reliable" ones from conf and then db, are added to `peerData` as `keepConnection=true` and for the peer discovery part, they are added with `keepConnection` as false, meaning we can disconnect from them. It is used for getting a random peer for use in node's operations as we filter what peers have `keepConnection` as true and also to determine reconnection attemps, as we would want to reconnect peers from conf but not the other temporary ones since with many of them connection are offline and reconnecting would be useless. This is why `maxConnectionTries` was made an argument. One last thing as for using a stack to store peers to check, is because I found addr messages to be more reliable and faster way of getting required peers (dns seeds are only used if stack is empty), and addr messages are received continuously while node operation so the top would usually be a peer from addr message.

Edit: Also about the need for a scheduler and stack. This was added so that we are only trying a limited number of connections at a time. Trying all 50+ peers at the same instant led to many errors due to race conditions so that's why scheduler was needed.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
solo1g: They were fine earlier when we connect to just 1 peer but with many connections, most of which are temporary just to determine service, that would lead to way too many log messages.

core/src/main/scala/org/bitcoins/core/util/NetworkUtil.scala
Christewart: ping on this, also maybe consier renaming the method to `parseUnresolvedInetSocketAddress()`

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
Christewart: ping on this

benthecarman:
It could be worth adding some hardcoded seeds from here as well

https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt

solo1g:
> It could be worth adding some hardcoded seeds from here as well
> 
> https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt

These seem useful. For adding these, one way that I can think of is to create all these as entries in our database itself as part of a migration. Or do we want to keep this separate, intending the database to only consist of peers we have actually connected to? What's your opinion?

benthecarman:
> These seem useful. For adding these, one way that I can think of is to create all these as entries in our database itself as part of a migration. Or do we want to keep this separate, intending the database to only consist of peers we have actually connected to? What's your opinion?

If you do them in an insert statement in the migration we won't be able to edit them later which could be a problem.

Maybe add the list as a resource file and we can add them to the db in the `NodeAppConfig.start()` function or something

Christewart:
Disregard this, i forgot to startup my local bitcoind node before testing :-) :man_facepalming:

Christewart:
So after running for 1 hour on mainnet connected to 1 peer (my local bitcoind node), it appears that i have 1 peer stored in my table.  Does this seem right? My understanding is other addresses should be gossiped to me and then stored if they are block filter peers?

Screenshot from 2021-09-14 08-27-55

Christewart:
Perhaps best for another PR, but another task to tackle on another PR that is independent of this one? I was attempting to look at the peers i'm connected to to test `bitcoin-s.node.maxConnected = 2`. I believe you could work on this before this PR is actually merged. In my mind, `getpeers` should return the connected peers. Does this seem correct @benthecarman ?

Screenshot from 2021-09-14 08-32-27

Christewart:
> So after running for 1 hour on mainnet connected to 1 peer (my local bitcoind node), it appears that i have 1 peer stored in my table. Does this seem right? My understanding is other addresses should be gossiped to me and then stored if they are block filter peers?
> 
> Screenshot from 2021-09-14 08-27-55

So i adjusted `bitcoin-s.node.maxConnectedPeers=2` and then restarted my node, now it seems more peers are getting stored in my database. This is after running for 15-30 minutes compared to `bitcoin-s.node.maxConnectedPeers=1` running for 1.5 hours and storing only 1 peer. 

Just to be clear, do we store peers in the database when `bitcoin-s.node.maxConnectedPeers=1`? We shouldn't connect to them, but we should store them IMO. 

Screenshot from 2021-09-14 08-41-34

Christewart:
Ok so i've managed to get two peers connected at the same time, and unsurprisingly one of these peers dies relatively quickly because of exceptions being thrown in the `chain` project.

With log messages like this 

>2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 headers

it would be useful to add `peer=$peer` to the log message so that we know which peer we are receiving headers from. It appears when I am doing initial block download with 2 connected peers, we have multiple peers receiving the same set of headers causing errors in the chain project. Here is a stack trace, but i'm attaching the full logs to this message to for help debugging 

[bitcoin-s.log](https://github.com/bitcoin-s/bitcoin-s/files/7162463/bitcoin-s.log)


It seems we are getting the  same set of block ehaders from our two nodes causing exceptions in the chain project. My understanding is we should only be doing initial block download from one of our peers, not both of them redundantly

```
2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 header

>2021-09-14T13:43:55UTC INFO [ControlMessageHandler] Connected to peer Peer(136.24.9.2:8333) with compact filters. Connected peer count 2
2021-09-14T13:43:56UTC INFO [NeutrinoNode] Starting sync node, height=610000 hash=0000000000000000000a6f607f74db48dae0a94022c10354536394c17672b7f7
2021-09-14T13:43:56UTC INFO [BitcoinSServerMain] Starting neutrino node sync, it took=3353ms
2021-09-14T13:43:56UTC INFO [Server] Started Bitcoin-S HTTP server at /127.0.0.1:9999
2021-09-14T13:43:56UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:43:56UTC INFO [NeutrinoNode] Starting sync node, height=610000 hash=0000000000000000000a6f607f74db48dae0a94022c10354536394c17672b7f7
2021-09-14T13:43:56UTC INFO [BitcoinSServerMain] Done starting Main! It took 3557ms
2021-09-14T13:43:59UTC INFO [DataMessageHandler] Received maximum amount of headers in one header message. This means we are not synced, requesting more
2021-09-14T13:43:59UTC WARN [P2PClientActor] unhandled message=NetworkMessage(NetworkHeader(MainNet,version,UInt32Impl(103),ByteVector(4 bytes, 0xe51e83e4)), VersionMessage(ProtocolVersion70013, ServiceIdentifier(none), epoch=1631627039, receiverServices=00000000000000000000ffff7f000101, receiverAddress=00000000000000000000ffff7f000101, receiverPort=8333), userAgent=/Bitcoin-S:1.7.0/, startHeight=612000, relay=true))
2021-09-14T13:43:59UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:44:00UTC ERROR [DataMessageHandler] Failed to handle data payload=HeadersMessage(CompactSizeUInt(2000),0000000000000000000b9cfe321443cb2ced7a4182b8594f077ee4e23d6c2ae2..000000000000000000103c2cffc3704b18ec98229b19d85f069bdbbfb13d0689
java.lang.RuntimeException: Failed to connect any headers to our internal chain state, failures=Vector(Successful(BaseBlockchain(tip=BlockHeaderDb(height=612000,hashBE=000000000000000000103c2cffc3704b18ec98229b19d85f069bdbbfb13d0689,version=Int32Impl(1073725440),prevBlockHashBE=000000000000000000110988dd87532a7d0a75c8dea6991bdbd12cbdca4f2290merkleRootHashBE=c73e856b4cfd7e95cc4eafddf5f825fce3f73753a4b1f9d4fa28ef712cf2b7e0,time=UInt32Impl(1578549787),nBits=UInt32Impl(387212786),nonce=UInt32Impl(3788433512),chainWork=3579856433888292367987533353),last=BlockHeaderDb(height=609984,hashBE=0000000000000000001465502b31eca279827150c13e449616a708272ed27e7b,version=Int32Impl(536870912),prevBlockHashBE=00000000000000000006f83f1f5052793cfa3fce1b7b3ba222cf37daaad99ebemerkleRootHashBE=7d479a058e44911e359ab9ff157c04d1414783cc8e17df75ab5b5b33277418a7,time=UInt32Impl(1577439816),nBits=UInt32Impl(387300560),nonce=UInt32Impl(2041549584),chainWork=3463527042123651387623202066),length=2017),Vector()))
	at org.bitcoins.chain.blockchain.ChainHandler.processHeadersWithChains(ChainHandler.scala:142)
	at org.bitcoins.chain.blockchain.ChainHandler.$anonfun$processHeaders$1(ChainHandler.scala:186)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:434)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:100)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:49)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:295)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1016)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
2021-09-14T13:44:00UTC ERROR [DataMessageHandler] Error when processing headers message
java.lang.RuntimeException: Failed to connect any headers to our internal chain state, failures=Vector(Successful(BaseBlockchain(tip=BlockHeaderDb(height=612000,hashBE=000000000000000000103c2cffc3704b18ec98229b19d85f069bdbbfb13d0689,version=Int32Impl(1073725440),prevBlockHashBE=000000000000000000110988dd87532a7d0a75c8dea6991bdbd12cbdca4f2290merkleRootHashBE=c73e856b4cfd7e95cc4eafddf5f825fce3f73753a4b1f9d4fa28ef712cf2b7e0,time=UInt32Impl(1578549787),nBits=UInt32Impl(387212786),nonce=UInt32Impl(3788433512),chainWork=3579856433888292367987533353),last=BlockHeaderDb(height=609984,hashBE=0000000000000000001465502b31eca279827150c13e449616a708272ed27e7b,version=Int32Impl(536870912),prevBlockHashBE=00000000000000000006f83f1f5052793cfa3fce1b7b3ba222cf37daaad99ebemerkleRootHashBE=7d479a058e44911e359ab9ff157c04d1414783cc8e17df75ab5b5b33277418a7,time=UInt32Impl(1577439816),nBits=UInt32Impl(387300560),nonce=UInt32Impl(2041549584),chainWork=3463527042123651387623202066),length=2017),Vector()))
	at org.bitcoins.chain.blockchain.ChainHandler.processHeadersWithChains(ChainHandler.scala:142)
	at org.bitcoins.chain.blockchain.ChainHandler.$anonfun$processHeaders$1(ChainHandler.scala:186)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:434)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:100)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:49)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:295)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1016)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
2021-09-14T13:44:00UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:44:00UTC INFO [ControlMessageHandler] Connected to peer Peer(136.24.9.2:8333) with compact filters. Connected peer count 2
2021-09-14T13:44:00UTC INFO [NeutrinoNode] Starting sync node, height=612000 hash=000000000000000000103c2cffc3704b18ec98229b19d85f069bdbbfb13d0689
2021-09-14T13:44:02UTC INFO [DataMessageHandler] Received maximum amount of headers in one header message. This means we are not synced, requesting more
2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 headers
2021-09-14T13:44:03UTC INFO [DataMessageHandler] Received headers message with 2000 headers
 ```

**Comment by solo1g** [comment]
>  So after running for 1 hour on mainnet connected to 1 peer (my local bitcoind node), it appears that i have 1 peer stored in my table. Does this seem right? My understanding is other addresses should be gossiped to me and then stored if they are block filter peers?

I have been able to replicate this and it seems that you do not get addr messages from peers while you are getting headers from them. This causes us to rely only on dns seeds, which as I mentioned earlier, were quite ineffective in finding peers with compact filters. It might be that headers were synced in that time or on running the second time you got addr messages before the first header response (this is a race condition that I observed while testing today that allows us to get supported peers immediately in some cases, though I confirmed that there were no subsequent addr messages while getting headers).

This behavior can be improved using hardcoded nodes rather than dns seeds as Ben mentioned in https://github.com/bitcoin-s/bitcoin-s/pull/3512#issuecomment-918594759. I picked 3 randomly from that list and all of them supported compact filters so it seems more promising to try from that list first and then go about trying dns seeds.

**Comment by solo1g** [comment]
> It appears when I am doing initial block download with 2 connected peers, we have multiple peers receiving the same set of headers causing errors in the chain project.

I verified that this was indeed the case. This is a bug from my previous PR :sweat_smile: .

**Comment by Christewart** [comment]
> > So after running for 1 hour on mainnet connected to 1 peer (my local bitcoind node), it appears that i have 1 peer stored in my table. Does this seem right? My understanding is other addresses should be gossiped to me and then stored if they are block filter peers?
> 
> I have been able to replicate this and it seems that you do not get addr messages from peers while you are getting headers from them.

Hmmmm ok. That seems like a setting bitcoin core likely has. I bet bitcoin core doesn't gossip addreses while an initial block download is going on. Ben might know for sure cc @benthecarman

**Comment by Christewart** [comment]
When i run unit tests locally (`nodeTest/test`) i see the same failures that are being reported on CI.

For what it is worth, when the unit test says `SUITE ABORTED` it usually means that there is a bug in the fixture setup or tear down logic. 

Screenshot from 2021-09-28 08-00-50

**Comment by Christewart** [comment]
So here are my thoughts on this PR:

I think we should break this out into smaller PRs that are easier to review and get merged. Here are small PRs we could break out of this 

1. `ControlPayloadHandler` (just refactoring existing code into this)
2. `PeerDAO` (simply adding the ability store a peer in the database)
3. Logic to actually store peers in the database using `PeerDAO` (I personally think this should be a separate PR, but might be able to bundled with `2` if it is trivial)
4. Managing multiple connections

I'm guessing `4` is where the complexity is introduced and where testing/review needs to be applied. Since this current diff is so big, it is hard for me to give proper guidance/review.

What do you think of that?

**Comment by Christewart** [comment]
Here is my `bitcoin-s.log` file from running with two peers. 

```
bitcoin-s.node.peers =["neutrino.suredbits.com:8333", "neutrino1.suredbits.com:8333"]
```

I did NOT set `maxConnectedPeers`, perhaps this is why i'm getting these errors? 

Here is the log

[bitcoin-s.log](https://github.com/bitcoin-s/bitcoin-s/files/7244723/bitcoin-s.log)

Issue #3466, opened 2021-08-02
Rethink `DEFAULT_BITCOIND_LOCATION` exception safety


Christewart:
I've been working through a problem with @shruthii625 about implementing cookie based authentication in bitcoin-s (#2891). 

One issue she has had is downloading bitcoind and configuring bitcoin-s to where the binary lives with the configuration key 

`bitcoin-s.bitcoind-rpc.binary`

This leads us to a default helper `val` we provide called `DEFAULT_BITCOIND_LOCATION` 

This is causing us problems on multiple fronts: 

1. [It assumes bitcoind is on your path](https://github.com/bitcoin-s/bitcoin-s/blob/1e87cb0fdecea0e67eb2679ee6c7b47ad644d72c/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala#L107)
2. [It throws exceptions eagerly if the place is not defined!](https://github.com/bitcoin-s/bitcoin-s/blob/1e87cb0fdecea0e67eb2679ee6c7b47ad644d72c/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala#L120)

## What we need to do 

1. We need to modify the return type from `File` to `Option[File]` so we don't have exceptions thrown eagerly
2. Everywhere we there are compiler errors, we need to handle the `None` case by referring to the binary configured in `bitcoin-s.bitcoind-rpc.binary`

benthecarman:
ACK, currently we have an `isRemote` option which will skip the file checks but is annoying because it requires us still to put a `File` for the `binary` and the `datadir`.

It might be better to split up `BitcoindInstance` and/or `BitcoindRpcClient` into 2 separate types, one that is for a locally running bitcoind and will start the binary, and one that is for a "remote" bitcoind where it won't start the binary and do the file checks.

Christewart:
Done in #3600

PR #3522, opened 2021-08-11
solves issue 3464 and half of issue 3466

### LINKED ISSUE #3464: `rpcuser` and `rpcpassword` MUST be set when using `BitcoindRpcAppConfig`
https://github.com/bitcoin-s/bitcoin-s/blob/1e87cb0fdecea0e67eb2679ee6c7b47ad644d72c/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala#L88

This is not necessary. For instance if you want to authenticate against a locally running bitcoind instance via cookie authentication, these fields do _not_ need to be defined inside of the `bitcoin-s.conf` file. When using cookie based authentication (#2891) i should not need to specify `rpcuser` and `rpcpassword`. 

We need to modify the return types of `rpcuser` and `rpcpassword` to be `Option[String]` rather than `String` and then propagate the changes.

shruthii625:
fixes #3464 and part of #3466 

If rpcuser and rpcpassword are not set, then cookie based authentication will be used. 
If bitcoind is not found on PATH, bitcoin-s.conf will be read to obtain the binary.


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala
```diff
@@ -137,16 +147,25 @@ case class BitcoindRpcAppConfig(
...
       uri = uri,
       rpcUri = rpcUri,
       authCredentials = authCredentials,
       zmqConfig = zmqConfig,
-      binary = binaryOpt.getOrElse(fallbackBinary),
+      binary = binaryOpt match {
+        case Some(file) => file
+        case None => {
+          val homeVar = sys.env("HOME");
+          val config = ConfigFactory
```
Christewart: Is there a reason this config reference doesn't work? 

https://github.com/bitcoin-s/bitcoin-s/blob/4f7b6422ea3390f276733ee3c0fb21d426bde751/app-commons/src/main/scala/org/bitcoins/commons/config/AppConfig.scala#L120

shruthii625: AppConfig already has a 'network' and a 'datadir'. Plus, if I extend AppConfig, I'd have to implement all of it's methods here. So, I wasn't too sure if I should use it

Christewart: It already extends `AppConfig`

https://github.com/shruthii625/bitcoin-s/blob/2fa4dddbcc673fd21c71ec55c09c00ab88f17da6/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcAppConfig.scala#L25

shruthii625: I hadn't properly read your message and assumed you were talking about BitcoindInstance yikes
I've changed it to use the config of AppConfig and it seems to be working fine

PR #3530, opened 2021-08-11
Add proxy and tor sections to reference conf

user411:


Christewart:
For context, this happened when I walked a user through using our embedded tor binaries. The tor binary seems to have started correctly, but configuration on our side is bad. We shouldn't require the user to manually configure

>2021-08-11T21:38:49UTC INFO [DLCNode] Binding server to /0.0.0.0:2862, with tor hidden service: false
2021-08-11T21:38:49UTC INFO [DLCServer] Bound at /[0:0:0:0:0:0:0:0]:2862

Christewart (approved):

benthecarman (approved):

PR #3534, opened 2021-08-12
Add TorBundle, make all mac files executable that are required

### LINKED ISSUE #3533: osx permission denied cannot execute tor binary
I'm trying to do a DLC between my two laptops this morning on 560edf66d8cedffe1b4a731ab7038fcb1ef8324a 

It appears that my tor binary is not getting started on my mac with the following error

Screen Shot 2021-08-12 at 8 08 39 AM

Christewart:
fixes #3533 

There are multiple files that need to bet set to executable on mac osx, so make a helper class called `TorBundle` to deal with multiple files that need to be set to executable.


Christewart:
This worked for me on my macbook, but would like to confirm with @rorp / @user411 that this also works for them

benthecarman (approved):

user411 (approved):
Worked in local testing

PR #3536, opened 2021-08-12
Add debug window to view log fixes #3516

### LINKED ISSUE #3516: Add ability to export logs from GUI
for non technical issues

user411:
fixes #3516

Adds a debug window to host any debugging operation we want to expose to users. First available: open the log file in default OS editor

Screen Shot 2021-08-12 at 11 11 45 AM
Screen Shot 2021-08-12 at 11 07 52 AM

I don't know if this is the preferred way to grab the path to the current data folder, copied strategy from BundleGUI:94

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/AppMenuBar.scala
@@ -18,7 +18,7 @@ object AppMenuBar {
   def menuBar(model: WalletGUIModel, dlcPane: DLCPane): MenuBar = {
     val menuBar = new MenuBar {
       menus = List(new FileMenu().fileMenu,
-                   new ViewMenu(dlcPane).viewMenu,
+                   new ViewMenu(model, dlcPane).viewMenu,
                    new HelpMenu(model).helpMenu)
     }
     // Use MacOS native menuing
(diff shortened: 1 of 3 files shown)


user411:
Fixes #3516

Christewart:
On my linux machine off of 9ad49bd

Screenshot from 2021-08-12 14-24-50

Christewart:
Looks like the path is getting duplicated somehow?

Screenshot from 2021-08-12 16-00-38

user411:
@Christewart Please test with the updated code

Christewart:
Some compile issues: https://github.com/bitcoin-s/bitcoin-s/pull/3536/checks?check_run_id=3316254379#step:5:141

user411:
Amazing that ran locally, do not understand that at all

user411:
The thread freezes during Desktop.open() on Ubuntu/linux in JavaFX and it should be run in another thread, and Platform.runLater() wasn't handling it (which seemed to be the "right-way" to handle it). The current code uses `xdg-open` to pop open the log as a text file on linux and Desktop.open() for everything else that supports it. I added logging to at least see any case hit that falls outside the expected.

I have tested on Mac and Ubuntu 20

Christewart (approved):

PR #3540, opened 2021-08-13
Wallet on top UI, minWidth on DLC View fields

user411:
Puts the wallet back on top of the accordion. Sets minWidths on each side of the SplitPane and adds minWidths to the DLC View fields

Left minWidth:
Screen Shot 2021-08-12 at 5 59 36 PM

Right minWidth:
Screen Shot 2021-08-12 at 5 59 51 PM

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
@@ -38,6 +38,29 @@ object ViewDLCDialog {
     val _ = dialog.showAndWait()
   }
 
+  private def getLabel(label: String): Label = {
+    new Label {
+      styleClass += "view-dlc-label"
+      text = label
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
```diff
@@ -229,6 +193,8 @@ object ViewDLCDialog {
       }
       add(node, columnIndex = 1, rowIndex = row)
 
+      // TODO : Refund button and discriminator
```
Christewart, line 196: Worth an issue?

Christewart (approved):
LGTM, waiting for @benthecarman for review though

benthecarman (approved):
tested, LGTM

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
user411: Made #3544

PR #3542, opened 2021-08-13
Style wallet startup screens

user411:
Update styling for wallet startup screen and backend selection tabs.

Screen Shot 2021-08-12 at 6 49 12 PM
Screen Shot 2021-08-12 at 6 49 20 PM


Christewart (approved):
Looks great!

PR #3546, opened 2021-08-13
Add Tor running text to UI

user411:
Add a indicator style label for Tor Proxy running

"Tor Address" -> "Tor DLC Host Address"

Fix bug with network change at startup not making it into UI network label.

Screen Shot 2021-08-19 at 2 10 00 PM

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
@@ -107,7 +107,8 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
         serverArgParser)(system.dispatcher)
 
     GlobalData.network = toNodeConf(appConfig).network
-    GlobalData.torEnabled.value = appConfig.torConf.enabled
+    GlobalData.proxyEnabled.value =
+      appConfig.nodeConf.socks5ProxyParams.isDefined
 
     validatePreferenceValues()
 
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
@@ -5,7 +5,6 @@ import org.bitcoins.commons.util.{DatadirParser, ServerArgParser}
 import org.bitcoins.gui._
 import org.bitcoins.gui.util.GUIUtil
 import org.bitcoins.server.BitcoinSAppConfig
-import org.bitcoins.server.BitcoinSAppConfig.toNodeConf
 import scalafx.application.{JFXApp3, Platform}
 import scalafx.geometry.Pos
 import scalafx.scene.Scene
@@ -105,10 +104,6 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
       BitcoinSAppConfig.fromDatadirWithBundleConfWithServerArgs(
         datadirParser.datadir,
         serverArgParser)(system.dispatcher)
-
-    GlobalData.network = toNodeConf(appConfig).network
-    GlobalData.proxyEnabled.value =
-      appConfig.nodeConf.socks5ProxyParams.isDefined
 
     validatePreferenceValues()
 
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
@@ -107,6 +107,7 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
         serverArgParser)(system.dispatcher)
 
     GlobalData.network = toNodeConf(appConfig).network
+    GlobalData.torEnabled.value = appConfig.torConf.enabled
 
     validatePreferenceValues()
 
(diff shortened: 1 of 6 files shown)


Christewart:
So there seems to be a bug AFAICT, but i guess it depends on what you mean by "Running over tor". For our DLC negotitation, we require TOR. However, for our blockchain data source, we can do clearnet or tor. 

In the case we are using clearnet for our neutrino node, it seems we still say it is receiving chainstate over tor

Screenshot from 2021-08-13 10-34-02
Screenshot from 2021-08-13 10-33-40

user411:
Right now, when a user deselects Use Tor when starting the wallet, it doesn't turn off Tor, but if they select Use Tor, they can turn it on. This is around the enable logic that auto-flips it on in the config if an onion address is entered for a peer or if the box is checked. I think we probably want to just go with the state of the checkbox impacting tor enabling by itself, allowing overriding the config positively or negativel

user411:
I think the fix is in NeutrinoConfigPane

```scala
def getConfig: Config = {
    val proxyEnabled =
      torCheckBox.selected.value || peerAddressTF.text.value.contains(".onion")
    val configStr = s"""
                       |bitcoin-s.proxy.enabled = ${proxyEnabled}
                       |bitcoin-s.network = ${DatadirUtil.networkStrToDirName(
      networkComboBox.value.value.toString)}
                       |bitcoin-s.node.mode = neutrino
                       |bitcoin-s.node.peers = ["${peerAddressTF.text.value}"]
                       |""".stripMargin
    ConfigFactory.parseString(configStr)
  }
```

However, this does not turn off Tor either. Is it possible this is the bug @benthecarman is working on with lapping configs?

benthecarman:
> However, this does not turn off Tor either. Is it possible this is the bug @benthecarman is working on with lapping configs?

There is the `proxy` and `tor`, `proxy` routes traffic through the tor proxy and `tor` is for setting up the hidden service.

user411:
Setting `bitcoin-s.proxy.enabled = false` in the config here in the codebase today does not disable routing wallet traffic over the proxy. That is the value that is getting displayed as 'over Tor' next to the network in the UI

benthecarman:
> Setting `bitcoin-s.proxy.enabled = false` in the config here in the codebase today does not disable routing wallet traffic over the proxy.

hmm, that is weird

user411:
I don't think I put that well. I meant to say that dynamically setting `bitcoin-s.proxy.enabled = false` over a loaded config with a true value does not seem to disable the proxy. I thought that was in the logic for dynamically overwriting not handling the checkbox is unchecked case, but when I fixed that issue with the code block above, it still does not seem to ultimately cause `appConfig.torConf.enabled` to go false when BundleGUI is start()ing.

user411:
It took me some time to catch up on the tor/proxy point of what you were saying, but I think I got there.

The current behavior with the last commit is: 
1. Use Tor serializes to the config as both true and false.
2. The value of `appConfig.nodeConf.socks5ProxyParams.isDefined` in BundleGUI is always what it was initially in the filesystem config file. However, the application is running true to the dynamically set values from the wallet launch UI.

user411:
I think 2 above is a symptom of some timing initialization issues. The current code relies on the config state at the time of window load, which is only the filesystem config. Working on where to set the new hybrid filesystem/dynamic config.

user411:
New commit sets Network and proxy in UI from finalConfig used for BitcoinSServer instance

app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
```diff
@@ -106,8 +105,6 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
         datadirParser.datadir,
         serverArgParser)(system.dispatcher)
 
-    GlobalData.network = toNodeConf(appConfig).network
```
user411, line 109: This is the loaded from filesystem or default value. We want to track the final server value

app/gui/src/main/scala/org/bitcoins/gui/GlobalData.scala
```diff
@@ -19,6 +19,17 @@ object GlobalData {
...
+  val proxyEnabled = BooleanProperty(false)
+  val networkString: StringProperty = new StringProperty("")
+
+  def setBitcoinNetwork(
+      network: BitcoinNetwork,
+      proxyEnabled: Boolean): Unit = {
+    this.network = network
+    this.proxyEnabled.value = proxyEnabled
+    networkString.value = "Network: " + network
+    if (proxyEnabled) networkString.value += " over Tor"
```
benthecarman: if we are using the bitcoind backend should we include `over Tor`?

benthecarman (commented):
I think it'd make more sense to have the `tor running` similiar to the backend connected indicator. 

The GUI shouldn't really care if your bitcoind or neutrino are routed through tor, just if we have the processing running imo

app/gui/src/main/scala/org/bitcoins/gui/GlobalData.scala
user411: I was thinking from the point of view of informing the user that tor is running on the node connection. Whether "over Tor" or a pseudo indicator is the right way to do it, I'm not sure. I started thinking they were correlated, but I understand that's just the connection to the underlying node.

I would always show the state of tor just to acknowledge that it is running as the user set  with bitcoind backend as well as neutrino.

From my point of view, the user needs to know whether we are running a tor instance on their behalf. We could in the debug window, but I see it as an implementation detail they usually don't need to worry about.

benthecarman: > From my point of view, the user needs to know whether we are running a tor instance on their behalf. 

I agree and this PR doesn't do that imo, it says their node is running over tor even when it might not be.

benthecarman (approved):

benthecarman:
Needs rebase

Issue #3465, opened 2021-08-02
Rework how `Client.start()` works


Christewart:
https://github.com/bitcoin-s/bitcoin-s/blob/e2c2798e60ac26728c6f42d9c49a59d4745dd5e3/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L162

This will not work with cookie based authentication (#2891) since we try to call `startBinary()` everytime bitcoind is started

https://github.com/bitcoin-s/bitcoin-s/blob/78f4dfb8c66eedc9c5b758ae5880274bf74571d2/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala#L134

This leads to an error that looks like this:

>[error] Error: Cannot obtain a lock on data directory /home/chris/.bitcoin. Bitcoin Core is probably already running.
[info] 2021-08-02T12:51:49UTC INFO [BitcoindV21RpcClient] Could not start bitcoind instance! Message: No content to map due to end-of-input
[info]  at [Source: (String)""; line: 1, column: 0]
[info] 2021-08-02T12:51:49UTC ERROR [BitcoinSServerMain] Failed to initialize configuration for BitcoinServerMain
[info] com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
[info]  at [Source: (String)""; line: 1, column: 0]
[info]  at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
[info]  at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4360)
[info]  at com.fasterxml.jackson.databind.ObjectMapper._readValue(ObjectMapper.java:4176)
[info]  at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2476)
[info]  at play.api.libs.json.jackson.JacksonJson$.parseJsValue(JacksonJson.scala:282)
[info]  at play.api.libs.json.StaticBinding$.parseJsValue(StaticBinding.scala:17)
[info]  at play.api.libs.json.Json$.parse(Json.scala:173)
[info]  at org.bitcoins.rpc.client.common.Client.$anonfun$getPayload$1(Client.scala:340)
[info]  at scala.util.Success.map(Try.scala:262)

Since we error there, we never get to run this code which actually does the cookie authentication

[...]

PR #3551, opened 2021-08-16
Changes in bitcoind-rpc module

### LINKED ISSUE #3464: `rpcuser` and `rpcpassword` MUST be set when using `BitcoindRpcAppConfig`

shruthii625:
fixes  #3464, #3465 and a part of #3466

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
@@ -125,15 +125,6 @@ trait Client
     *         cannot be started
     */
   override def start(): Future[BitcoindRpcClient] = {
-    if (version != BitcoindVersion.Unknown) {
-      val foundVersion = instance.getVersion
-      if (foundVersion != version) {
-        throw new RuntimeException(
-          s"Wrong version for bitcoind RPC client! Expected $version, got $foundVersion")
-      }
-    }
-
-    val startedF = startBinary()
 
     // if we're doing cookie based authentication, we might attempt
     // to read the cookie file before it's written. this ensures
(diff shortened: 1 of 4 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -151,16 +142,34 @@ trait Client
       case _: PasswordBased => Future.successful(())
 
     }
+    val isAlreadyStarted: Future[Boolean] = isStartedF
+
+    val started: Future[Future[BitcoindRpcClient]] = isAlreadyStarted map {
```
benthecarman: should use `isAlreadyStarted.map{`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/Client.scala
```diff
@@ -151,16 +142,34 @@ trait Client
...
+      case false =>
+        if (version != BitcoindVersion.Unknown) {
+          val foundVersion = instance.getVersion
```
benthecarman, line 150: I don't think we can get the version when it is `isRemote` (if bitcoind isn't installed)

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -146,7 +164,19 @@ object BitcoindInstance extends InstanceFactory[BitcoindInstance] {
...
+    fromDatadir(
+      dir,
+      DEFAULT_BITCOIND_LOCATION match {
+        case Some(file) => file
+        case None => {
+          val homeVar = sys.env("HOME");
+          val config = ConfigFactory
+            .parseFile(new File(homeVar + "/.bitcoin-s/bitcoin-s.conf"))
+            .resolve()
+          new File(config.getString("bitcoin-s.bitcoind-rpc.binary"))
```
benthecarman: this is done in multiple places, should be moved to a val or function

benthecarman (changes requested):

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
```diff
@@ -107,16 +117,23 @@ object BitcoindInstance extends InstanceFactory[BitcoindInstance] {
                          proxyParams = proxyParams)
   }
 
-  lazy val DEFAULT_BITCOIND_LOCATION: File = {
+  val bitcoindLocationFromConfigFile: File = {
```
benthecarman, line 120: this should be a `lazy val` or `def` otherwise the program could fail there is no config file or `bitcoin-s.bitcoind-rpc.binary` option in the config file

PR #3549, opened 2021-08-14
Implement TorAppConfig.start(), use tor logs to detect when the tor is fully started

Christewart:
The purpose of this is to try and fix #3547 

This tries to fix the race condition where we could have started the tor binary, but it isn't fully initialized and ready to receive messages. 

The way I do this is by looking inside of the `torLogFile` to detect this log line 

>Bootstrapped 100% (done): Done

If that log is seen, we complete the Future and indicate that tor binary start is done

Here are log lines from running the tor binary

Screenshot from 2021-08-14 07-45-36

Maybe we don't need to wait for that line and can use a earlier one, but I decided to make the conservative choice to wait for that bootstrapping log. On my machine this takes ~4 seconds.

This probably conflicts with #3484 in some regards.

PR #3558, opened 2021-08-16
Add guard for the case where tor binary hasn't created log file yet

### LINKED ISSUE #3556: NoSuchFileException: .bitcoin-s/tor/TorLogs.txt
This was introduced in #3549 and doesn't work on mac according to @user411

My guesses 

1. The log file is getting created after we call `.start()` and we check if it exists too soon causing a race condition
2. the log file is different on mac?

Christewart:
fixes #3556 

There is a race condition between our application code and the tor binary to create tor's log file in #3549 . Previously we did not have a guard to check if the log file even exists, which would result in Scala throwing a `NoSuchFileException`. 

It is safe to assume the tor binary is not fully started if it hasn't even written the log file yet, so just return false in the case the log file doesn't exist.

### DIFF
--- a/tor/src/main/scala/org/bitcoins/tor/config/TorAppConfig.scala
+++ b/tor/src/main/scala/org/bitcoins/tor/config/TorAppConfig.scala
@@ -78,12 +78,19 @@ case class TorAppConfig(
 
   /** Checks it the [[isBootstrappedLogLine]] exists in the tor log file */
   private def checkIfLogExists: Boolean = {
-    val stream = Files.lines(torLogFile)
-    try {
-      stream
-        .filter((line: String) => line.contains(isBootstrappedLogLine))
-        .count() > 0
-    } finally if (stream != null) stream.close()
+    if (Files.exists(torLogFile)) {
+      val stream = Files.lines(torLogFile)
+      try {
+        stream
+          .filter((line: String) => line.contains(isBootstrappedLogLine))
+          .count() > 0
+      } finally if (stream != null) stream.close()
+    } else {
+      //log file may not exist quite yet as we just started the tor binary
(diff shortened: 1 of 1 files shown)


user411:
I turned off local tor service, killed running tor binary, got the following:


[info] Aug 16 14:40:20.412 [notice] Opening Socks listener on 127.0.0.1:9050
[info] Aug 16 14:40:20.412 [notice] Opened Socks listener connection (ready) on 127.0.0.1:9050
[info] Aug 16 14:40:20.412 [notice] Opening Control listener on 127.0.0.1:9051
[info] Aug 16 14:40:20.412 [notice] Opened Control listener connection (ready) on 127.0.0.1:9051
[info] Aug 16 14:40:20.412 [warn] Fixing permissions on directory /Users/ivan/.bitcoin-s/tor
[info] 2021-08-16T20:40:25UTC ERROR [BitcoinSServerMain] Failed to initialize configuration for BitcoinServerMain
[info] org.bitcoins.asyncutil.AsyncUtil$RpcRetryException: Condition timed out after 50 attempts with interval=100 milliseconds waiting periods
[info] 	at org.bitcoins.tor.config.TorAppConfig.isBinaryFullyStarted(TorAppConfig.scala:76)

Christewart:
When testing this, i ran into a case where it took ~25 seconds for our log line to be reached to indicate we are 100% bootstrapped on the network. 

The first log

>Aug 16 16:00:36.000 [notice] Tor 0.4.5.7 (git-83f895c015de5520) opening new log file.
...
>Aug 16 16:01:01.000 [notice] Bootstrapped 100% (done): Done

So it seems this run took ~25 seconds to bootstrap the node.

user411 (approved):

PR #3555, opened 2021-08-16
Add rebroadcast funding tx button

user411:
Adds a Rebroadcast button next to the Funding Tx in the DLC View

Issue #1821, opened 2020-08-13
Robust re-connection in the case of laptop going to sleep


Christewart:
We should be able to handle a laptop "waking up" from sleeping ok. Currently I don't think this works.

Christewart:
Taken care of in #3572

PR #3561, opened 2021-08-17
2021 08 17 rebroadcast funding tx

Christewart:
Same as #3555 but rebased ontop of #3560 . I also removed the `println` and implemented a nice error message for closing rebroadcast


Christewart:
Tested this this morning between my macbook and desktop, the macbook was on the accept side. I disconnected my macbook from it's locally running bitcoind and then accepted a DLC offered from my desktop. It failed to broadcast (as expected), I then rebooted the bitcoind instance and restarted the wallet (no re-connection logic #1821 ) and then used @user411 's rebroadcast button and it worked like a charm.

PR #3566, opened 2021-08-17
Improve filter header verification

rorp:


Christewart (commented):
Unit tests to prevent this bug in the future?

Christewart (approved):

PR #3567, opened 2021-08-17
Move DLC/Event loading operations to Pane

user411:
Moves the hex/file loading operations in the section accordion headers above the accordion.

Screen Shot 2021-08-19 at 9 51 49 AM

### DIFF
--- a/app/gui/src/main/resources/themes/dark-theme.css
+++ b/app/gui/src/main/resources/themes/dark-theme.css
@@ -66,14 +66,13 @@
     -fx-border-width: 1px;
 }
 
-/* Accordion TitledPane heading content */
-.small {
-    -fx-font-size: 0.8em;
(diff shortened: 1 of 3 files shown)


benthecarman:
Looks good to me but we should really standardize the language. Currently it uses `Contract` to refer to an `Offer` and `Event` to refer to an `Announcement`

user411:
I agree all the views and sites should use the same nomenclature for things.

In this update, DLC / Contract refers to anything the UI can load from a file or hex perspective (so any state of a DLC being created) and Event refers to an Announcement or Contract Template.

What do you think is the best way to refer to any state of a DLC that can be loaded? Do you think 'Announcement' is the best label for Announcement/Contract Template or should we call out both?

benthecarman (approved):

Christewart (approved):

PR #3575, opened 2021-08-19
Fix app server docker docs

benthecarman:


Christewart (approved):
:man_facepalming: :man_facepalming:

PR #3576, opened 2021-08-19
Implement versioning for packaged tor

benthecarman:
Adds a `version.txt` file to the tor dir that we use to check if we have the lastest version of tor


Christewart (approved):
This looks fine to me, but what was the reason needed to do this? did you run into bugs related to this?

benthecarman:
> This looks fine to me, but what was the reason needed to do this? did you run into bugs related to this?

This is needed for when there is an update to tor. Then we can update the binaries we package and then people can update bitcoin-s and it will use the updated version of tor, otherwise it would keep the first version of tor they got from bitcoin-s

PR #3577, opened 2021-08-19
Add welcome pane to right view on startup

user411:
Adds welcome view to wallet on startup

Screen Shot 2021-08-19 at 1 49 21 PM


Christewart (approved):
:heavy_check_mark:

Issue #3525, opened 2021-08-11
Consider shipping 1.8 with `bitcoin-s.node.relay=true`, think about divergence in behavior between block filters & bitcoind


Christewart:
This is great because when using neutrino you will get your funding/closing txs gossiped over the network fixing #3510 

However, one concern I do have is divergence between behavior of block filters as a backend and bitcoind rpc as a backend. IIUC, bitcoind will _not_ retrieve the funding tx/closing tx?

The trade off here is of course increased bandwidth usage, but since we are shipping a desktop wallet currently this shouldn't be an issue IMO

PR #3580, opened 2021-08-19
Turn on bitcoin-s.node.relay=true by default on bundle/appServer

Christewart:
This allows for the offerer and the non closer of a DLC to learn about their unconfirmed txs being relayed around the network. This does increase bandwidth requirements on our nodes, but since we expect users to be running on a desktop environment, this should be fine. Partially addresses #3525


benthecarman:
wen erlay :P

benthecarman:
oops accidental close

benthecarman (approved):

PR #3597, opened 2021-08-24
CI fixes

rorp:


.github/workflows/Linux_2.12_Node_Tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Linux 2.12 App, Chain, Node, and Core Tests
```
benthecarman: should rename

.github/workflows/Linux_2.13_Node_Tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Linux 2.13 App, Chain, Node, and Core Tests
```
benthecarman: Same here

benthecarman (changes requested):

benthecarman (approved):

benthecarman (changes requested):
Compile Failing

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -164,13 +164,12 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
         peerMsgSenders(idx).connect()
         val isInitializedF = for {
           _ <- AsyncUtil.retryUntilSatisfiedF(() => isInitialized(idx),
-                                              maxTries = 1024,
+                                              maxTries = 240,
```
benthecarman: This will make the tor tests much more flaky, maybe bump it up but only when tor is enabled

Christewart (approved):

PR #3605, opened 2021-08-25
More bitcoind backend wallet tests

benthecarman:
Can't reproduce any errors users have reported


Christewart (approved):
Good work, but disappointing result :-( . I was hoping we would find :bug: :bug:

Christewart (changes requested):
Looks like there might be a bug related to postgres? 

https://github.com/bitcoin-s/bitcoin-s/pull/3605/checks?check_run_id=3426169478#step:5:389


This does seem like work that is related to #3451

Christewart (approved):

PR #3591, opened 2021-08-23
Add BitcoinSServerMain test to connect to bitcoind

Christewart:

PR #3609, opened 2021-08-25
Run appServerTest on both windows and mac

### LINKED ISSUE #3684: `-daemon` flag on bitcoind does not work on windows operating systems
In #3609 i was adding support to run `appServerTest` in windows. 

I see this error on CI

https://github.com/bitcoin-s/bitcoin-s/pull/3609/checks?check_run_id=3676540475#step:6:2693

Screenshot from 2021-09-22 10-26-02

Christewart:
We should be testing `appServerTest` on all of our supported operating systems especially when we get #3591 in.

This also fixes #3684

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
@@ -94,15 +94,20 @@ trait BitcoindRpcTestUtil extends Logging {
 
     /* pruning and txindex are not compatible */
     val txindex = if (pruneMode) 0 else 1
+    //windows environments don't allow the -daemon flag
+    //see: https://github.com/bitcoin-s/bitcoin-s/issues/3684
+    val isDaemon = if (EnvUtil.isWindows) 0 else 1
+    //if bitcoind is not a daemon, we get a ton of logs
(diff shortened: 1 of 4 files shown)


testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -94,15 +94,20 @@ trait BitcoindRpcTestUtil extends Logging {
...
     val txindex = if (pruneMode) 0 else 1
+    //windows environments don't allow the -daemon flag
+    //see: https://github.com/bitcoin-s/bitcoin-s/issues/3684
+    val isDaemon = if (EnvUtil.isWindows) 0 else 1
+    //if bitcoind is not a daemon, we get a ton of logs
+    //from bitcoind written to stdout, so turn off debug if we are not a daemon
+    val isDebug = if (isDaemon == 1) 1 else 0
     val conf = s"""
                   |regtest=1
-                  |daemon=1
```
benthecarman, line 99: this should be `daemon = $isDaemon`, no?

benthecarman (approved):
This is a better way to cause an error than deleting the directory like I did lol

Christewart:
> This is a better way to cause an error than deleting the directory like I did lol

yeah `ServerRunTest` would run like 5 minutes locally for me :sweat_smile:

PR #3612, opened 2021-08-26
Add CORS handling for running from a local dev server

user411:
This adds CORS headers so a local browser can make localhost requests for oracleServer data


user411:
As discussed via Slack, we may not want this here. Probably want to bubble up to a NodeJS 'proxy server' that can handle calls from the frontend and act as a dev server on localhost.

But this works for testing oracle-server-ui now.

PR #3640, opened 2021-08-31
Give funding txid in DLC Signed state

benthecarman:
We know this value by the `Signed` state, but don't display it. This is nice so we can check on a block explorer after our peer has broadcasted


Christewart (commented):
I have mixed feelings on displaying it. Just because we have the txid doesn't mean the acceptor will be actually contribute their funding signatures and complete the tx. 


All that said, that is probably my engineering bias coming out and this would be useful to normal users for sharing information for debugging, especially at this early stage in the project. Would like to hear others thoughts on this though cc @nkohen

benthecarman:
I think it's nice because if you have `relay = false` or are using the bitcoind backend without zmq (default) then you have to wait for the tx to be confirmed until you can get the txid, this is nice so you can lookup on the block explorer in the mean time. This will only result in something weird if sending of the `DLCSign` message fails

Christewart:
> I think it's nice because if you have `relay = false` or are using the bitcoind backend without zmq (default) then you have to wait for the tx to be confirmed until you can get the txid, this is nice so you can lookup on the block explorer in the mean time. This will only result in something weird if sending of the `DLCSign` message fails

or your coutnerparty is malicious. But we cannot do anything about that regardless, i guess.

Christewart (approved):

PR #3646, opened 2021-09-05
Decide which coin selection solution to use based on waste metric

benthecarman:
Did for fun, take your time to review

Supposed to be a copy of https://github.com/bitcoin/bitcoin/pull/22009

Adds a waste metric to coin selection, so now when selecting coins it tries all of our algos and picks the best one based on waste.

This requires a `longTermFeeRate` which is what the wallet assumes will be the average fee rate for the wallet over the long term. I added this to the wallet config and gave it a default of 10 sats/vbyte which is the same default bitcoin core chose

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -8,6 +8,7 @@ import org.bitcoins.core.wallet.fee.SatoshisPerByte
 import org.bitcoins.crypto.ECPublicKey
 import org.bitcoins.server.BitcoinSAppConfig
 import org.bitcoins.testkit.BitcoinSTestAppConfig
+import org.bitcoins.testkit.async.TestAsyncUtil
 import org.bitcoins.testkit.node.{
   NeutrinoNodeFundedWalletBitcoind,
   NodeTestUtil,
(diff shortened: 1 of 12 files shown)


benthecarman:
Getting concerning CI failures, will need to look into

benthecarman:
Rebased on top of #3647 so CI passes

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
```diff
@@ -73,8 +71,7 @@ class NeutrinoNodeWithWalletTest extends NodeTestWithCachedBitcoindNewest {
     //unconfirmed = 3 BTC - TestAmount - TestFees
     val condition1 = () => {
       condition(
-        expectedConfirmedAmount = 3.bitcoin,
-        expectedUnconfirmedAmount = 3.bitcoin - TestAmount - TestFees,
+        expectedBalance = 6.bitcoin - TestAmount - TestFees,
```
benthecarman, line 74: we are able to change this because we never change the `expectedConfirmedAmount`. This needed to change because sometimes we would select different size utxos and it would change how much sats become confirmed vs unconfirmed

Christewart:
Rebase please

benthecarman:
Rebased

Christewart (approved):

PR #3647, opened 2021-09-07
Add ordering of funding inputs to DLC Wallet

benthecarman:
Found from #3646

We did not preserve the order of the funding inputs in the dlc wallet

This is needed because for the DLCSign the signatures should be in the same order as they were in the original offer

https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#requirements-2

### DIFF
--- a/db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
+++ b/db-commons-test/src/test/scala/org/bitcoins/db/DbManagementTest.scala
@@ -82,13 +82,13 @@ class DbManagementTest extends BitcoinSAsyncTest with EmbeddedPg {
     val result = dlcDbManagement.migrate()
     dlcAppConfig.driver match {
       case SQLite =>
-        val expected = 3
+        val expected = 4
         assert(result == expected)
         val flywayInfo = dlcAppConfig.info()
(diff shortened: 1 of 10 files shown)


Christewart:
Looks like there is failures wrt to number of migrations: https://github.com/bitcoin-s/bitcoin-s/pull/3647/checks?check_run_id=3531152291#step:5:221

Christewart:
It looks like a different exception is being thrown here now causing this unit test to fail: https://github.com/bitcoin-s/bitcoin-s/pull/3647/checks?check_run_id=3531152345#step:5:818

Christewart (commented):
So this shouldn't affect DLCs that are in progress/closed because the negotiation process is done? 

I.e. we won't need to write any application migration code to restore the inputs for old DLCs, right?

benthecarman:
> So this shouldn't affect DLCs that are in progress/closed because the negotiation process is done?
> 
> I.e. we won't need to write any application migration code to restore the inputs for old DLCs, right?

It shouldn't. We should be extra safe because most people are doing small bets so it only requires a single input too

Christewart (approved):

PR #3654, opened 2021-09-09
Adjust the permissions of the statup script for the oracle server

Christewart:
This is needed so that any user can run the startup script for the oracle server. This is needed for the umbrel platform.

Screenshot from 2021-09-09 16-08-35


rorp (approved):
👍

Issue #3661, opened 2021-09-13
Socks5Connection unreliable, fails with `General Failure`


Christewart:
I've been working on #3629 to try and improve CI reliability. It seems that the root problem here is our socks5 proxy. For reference on the socks5 protocol, please see: https://en.wikipedia.org/wiki/SOCKS#SOCKS5 

Here is an example stack trace of us _constantly_ trying to reconnect to the socks5 proxy from the [`P2PClientActor`](https://github.com/Christewart/bitcoin-s-core/blob/27edabc9b159dc957e587a3ef90abd684fcb0d7d/node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala#L155)

It seems that we are

1. [Parsing the greeting correctly](https://github.com/Christewart/bitcoin-s-core/blob/5c625314c1fc9ef44ee4e0d814510c1aa42bf44a/tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala#L193)
2. [Parsing authentication correctly](https://github.com/bitcoin-s/bitcoin-s/blob/5c625314c1fc9ef44ee4e0d814510c1aa42bf44a/tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala#L207)
3. [We FAIL to handle the connectionRequest](https://github.com/Christewart/bitcoin-s-core/blob/5c625314c1fc9ef44ee4e0d814510c1aa42bf44a/tor/src/main/scala/org/bitcoins/tor/Socks5Connection.scala#L69)

Looking at the low level bytes, it appears that when we fail, we receive these bytes from the socks5 proxy

>2021-09-13T13:35:52UTC INFO [Socks5Connection] ConnectionRequest data=ByteString(5, 1, 0, 1, 0, 0, 0, 0, 0, 0)

A VALID connection requests bytes look like this: 

>2021-09-13T13:32:03UTC INFO [Socks5Connection] ConnectionRequest data=ByteString(5, 0, 0, 1, 0, 0, 0, 0, 0, 0)

So if you look carefully, the only different byte is the second byte. On the INVALID connect request, the second byte is `1`, whereas on a VALID connection request, the second byte is `0`.

I'm unsure at this moment what this means, or why it is happening. I've attached full logs for this below.

[...]

PR #3662, opened 2021-09-13
Fix case where we forget to correctly assign currentPeerMsgHandlerRecv

Christewart:
Fix missing update to `currentPeerMsgHandlerRecv` when we are disconnecting the node.


benthecarman (approved):
Is there somewhere you can add a test for this?

Christewart:
> Is there somewhere you can add a test for this?

Possibly, I'm wondering if it passes CI. It passes for me locally and i'm even seeing behavior that this _could_ fix the `nodeTest` test suite with tor locally (#3661)

I guess regardless if this fixes all the tor stuff, it is a bug and should be tested as you correctly point out.

PR #3669, opened 2021-09-15
Fix approximateUtxoSize calculation for p2wpkh

benthecarman:
In the comment for this fuction it says it was cribbed from https://github.com/bitcoinjs/coinselect/blob/master/utils.js

looks like `TX_OUTPUT_PUBKEYHASH` was copied instead of `TX_INPUT_PUBKEYHASH`

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelector.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelector.scala
@@ -104,7 +104,7 @@ object CoinSelector extends CoinSelector {
       case None =>
         utxo.scriptWitnessOpt match {
           case Some(script) => script.bytes.length
-          case None         => 25 // PUBKEYHASH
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/CoinSelector.scala
```diff
@@ -104,7 +104,7 @@ object CoinSelector extends CoinSelector {
       case None =>
         utxo.scriptWitnessOpt match {
           case Some(script) => script.bytes.length
-          case None         => 25 // PUBKEYHASH
+          case None         => 107 // PUBKEYHASH
```
Christewart, line 107: So they are wrong? https://github.com/bitcoinjs/coinselect/blob/becfd26096fd9fc719f5b4928935be61d87bbb98/utils.js#L6 

We should probably update they comment if they are

Christewart (approved):

PR #3672, opened 2021-09-17
Implement ability to provide external entropy to bitcoin-s

### LINKED ISSUE #3670: Add ability to import entropy from environment variable into the keymanager
Platforms such as umbrel offer entropy as an environment variable to running applications on the platform. This is very nice because a user can just backup their master seed for the entire umbrel platform and have their bitcoin-s wallet/oracle seeds backed up along with that. 

This likely means we need to offer a configuration that allows users to import entropy from an environment variable such as `BITCOIN_S_KEYMANGER_ENTROPY`. 

This can then be used in our bitcoin-s.conf files like this

```
bitcoin-s.keymanager.entropy=${?BITCOIN_S_KEYMANAGER_ENTROPY}
```

If the configuration is not set, we should default to our normal methods of generating entropy with [`BouncycastleCryptoRuntime.freshPrivateKey`](https://github.com/bitcoin-s/bitcoin-s/blob/f2a2874177c0c50b023bff3207371e13b5f08bdf/crypto/.jvm/src/main/scala/org/bitcoins/crypto/BouncycastleCryptoRuntime.scala#L24)

Christewart:
fixes #3670

Now you provide external entropy to bitcoin-s by using the configuration key `bitcoin-s.keymanager.entropy` in your `bitcoin-s.conf` file. 

Projects that are dependent on the keymanager -- such as the wallet and dlc oracle -- will now use this entropy.

As a by product of adding this feature, I consolidated a lot of key manager code. Previously we would pass around the keymanger as a parameter to wallet methods. This made little sense as we should be able to re-create the keymanager from `WalletAppConfig.kmConf`. 

This PR adds a new method [`KeyManagerAppConfig.toBip39KeyManager`](https://github.com/bitcoin-s/bitcoin-s/blob/3ec35d3cdd3f3c58b58057a1e64538f668d36f84/key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala#L120) that allows anyone with a reference to `KeyManagerAppConfig` to get the bip39 key manager. 

Now if you have a reference to a `WalletAppConfig`, you can just do `walletConf.kmConf.toBip39KeyManager` if you need direct access to a key manager.

### DIFF
--- a/.github/workflows/Linux_2.12_KeyManager_Wallet_DLC_Tests.yml
+++ b/.github/workflows/Linux_2.12_KeyManager_Wallet_DLC_Tests.yml
@@ -28,4 +28,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: run tests
-        run: sbt ++2.12.14 downloadBitcoind coverage keyManagerTest/test keyManager/coverageReport keyManager/coverageAggregate keyManager/coveralls feeProviderTest/test walletTest/test dlcWalletTest/test wallet/coverageReport wallet/coverageAggregate wallet/coveralls dlcOracleTest/test asyncUtilsTestJVM/test asyncUtilsTestJS/test oracleExplorerClient/test dlcOracle/coverageReport dlcOracle/coverageAggregate dlcOracle/coveralls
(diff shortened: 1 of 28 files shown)


Christewart:
Coverge needs to be improved: https://github.com/bitcoin-s/bitcoin-s/pull/3672/checks?check_run_id=3633273322#step:5:379

rorp (approved):

PR #3689, opened 2021-09-24
Add 'getversion' endpoint to fetch the version from the backend

Christewart:
Adds a `getversion` endpoint that can be queried in Docker environments, this should allow us to show the backend commit too on https://github.com/bitcoin-s/bitcoin-s-ts/pull/19

### DIFF
--- a/app/server-routes/src/main/scala/org/bitcoins/server/routes/CommonRoutes.scala
+++ b/app/server-routes/src/main/scala/org/bitcoins/server/routes/CommonRoutes.scala
@@ -1,0 +1,17 @@
+package org.bitcoins.server.routes
+
+import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.directives.RouteDirectives
+import org.bitcoins.core.util.EnvUtil
+
+case class CommonRoutes() extends ServerRoute {
+
+  override def handleCommand: PartialFunction[ServerCommand, Route] = {
+    case ServerCommand("getversion", _) =>
+      RouteDirectives.complete {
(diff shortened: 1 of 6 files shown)


app/server-routes/src/main/scala/org/bitcoins/server/routes/CommonRoutes.scala
```diff
@@ -0,0 +1,17 @@
...
+
+case class CommonRoutes() extends ServerRoute {
+
+  override def handleCommand: PartialFunction[ServerCommand, Route] = {
+    case ServerCommand("getversion", _) =>
+      RouteDirectives.complete {
+        val vec = Vector(("version", ujson.Str(EnvUtil.getVersion)))
+        val obj = ujson.Obj.from(vec)
+        Server.httpSuccess(obj)
+      }
```
benthecarman, line 15: Is there a reason this can't this be in core routes?

Christewart: is it ok to extend `CoreRoutes` into the `oracleServer`?

user411 (approved):
Looks good. I will add this as a separate call down the stack in https://github.com/bitcoin-s/bitcoin-s-ts/pull/19

app/server-routes/src/main/scala/org/bitcoins/server/routes/CommonRoutes.scala
benthecarman: Oh I see now, I guess it could, but I think common routes makes more sense.

benthecarman (approved):

Issue #3593, opened 2021-08-23
Windows user reports invalid zip produced by wallet backup


Christewart:
![Screenshot from 2021-08-23 13-51-41](https://user-images.githubusercontent.com/3514957/130501685-a71f2557-e938-4ca3-aa69-5aa2821ec48a.png)

![BadZip](https://user-images.githubusercontent.com/3514957/130501711-a50f2c0d-12b8-4fb2-865e-d29d5ba7fa6f.png)

Christewart:
I tried this on 5475a994cf60d513d878c0ab3f12a4cb64301324 and we now produce a valid zip.

PR #3694, opened 2021-09-27
Add backupwallet and backuporacle endpoints

rorp:


core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala
```diff
@@ -85,4 +85,9 @@ trait DLCOracleApi {
 
   /** Signs the SHA256 hash of the given bytes using the oracle's signing key */
   def signMessage(message: ByteVector): SchnorrDigitalSignature
+
+  /** Backup oracle database
+    * @param location baclup file location
+    */
+  def backup(location: String): Future[Unit]
```
Christewart: Any reason to to change location to a `Path` rather than `String`? Also it is probably worth noting in the scaladoc this is effictively a noop for postgres backends and only works for sqlite

Christewart (changes requested):
So we already sort of have this funcitonality with our `zipdatadir` (#3183) command: 

https://bitcoin-s.org/docs/next/wallet/backups#backing-up-from-the-gui

Also there seems to be issues with the `zipdatadir` command on windows (#3593). Do we want to remove `zipdatadir` in favor this? The nice thing about `zipdatadir` is it includes the `seeds/` directory so that the root private keys are backed up as well.

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -733,6 +733,20 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("backupwallet")
+        .action((_, conf) => conf.copy(command = BackupWallet(null)))
+        .text("Backs the wallet database up")
+        .children(
+          arg[String]("dest")
+            .text("")
```
benthecarman: Should add a description for this arg

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -733,6 +733,20 @@ object ConsoleCli {
                 case other => other
               }))
         ),
+      cmd("backupwallet")
+        .action((_, conf) => conf.copy(command = BackupWallet(null)))
+        .text("Backs the wallet database up")
+        .children(
+          arg[String]("dest")
```
benthecarman, line 740: Would make more sense to have this as a `Path` or `File` imo

dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala
```diff
@@ -433,6 +434,22 @@ case class DLCOracle()(implicit val conf: DLCOracleAppConfig)
       _ <- eventDAO.updateAll(updated)
     } yield OracleEvent.fromEventDbs(eventDbs)
   }
+
+  /** Backup oracle database
+    *
+    * @param location baclup file location
+    */
+  override def backup(location: String): Future[Unit] = {
+    if (conf.driver == DatabaseDriver.SQLite) {
```
benthecarman: better to do a `match` here for if we add a different backend in the future

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -933,6 +934,17 @@ abstract class Wallet
       this
     }
   }
+
+  override def backup(location: String): Future[Unit] = {
+    if (walletConfig.driver == DatabaseDriver.SQLite) {
```
benthecarman: same here

db-commons/src/main/scala/org/bitcoins/db/SQLiteUtil.scala
```diff
@@ -0,0 +1,12 @@
+package org.bitcoins.db
+
+object SQLiteUtil {
+
+  def backup(jdbcUrl: String, backupFileName: String): Unit = {
+    val conn = java.sql.DriverManager.getConnection(jdbcUrl)
+    try {
+      val _ = conn.createStatement().execute(s"BACKUP TO $backupFileName")
```
benthecarman: might be worth doing some checks beforehand that the file doesn't exist already or that we have permissions to write there

benthecarman (approved):

Christewart (approved):

Issue #3695, opened 2021-09-27
bitcoind backend does not update confirmations properly on mac


Christewart:
This is running on 2191eb20 on my macbook with a sqlite database. 

I did a DLC with funding txid `5d81ad28079abc2713f8c98518cbeb423284e22024d196e9fb1bc3adfc67cbb3` which has one confirmation on it. 

My desktop wallet says that the status is `CONFIRMED` whereas my macbook still has the status of the transaction as `BROADCAST`. 

My macbook: 

<img width="1274" alt="Screen Shot 2021-09-27 at 9 28 35 AM" src="https://user-images.githubusercontent.com/3514957/134928553-d1122c7c-1f81-4bab-9063-f9833c851f97.png">

also my bitcoind on my macbook HAS seen the transaction (we broadcast via this bitcoind) and it has one confirmation on it

<img width="1299" alt="Screen Shot 2021-09-27 at 9 30 03 AM" src="https://user-images.githubusercontent.com/3514957/134928858-983a7078-c738-43cc-be8e-c5df10f91fa5.png">

here are the logs i see on my macbook: 

>2021-09-27T14:00:36UTC INFO [DLCWallet$DLCWalletImpl] DLC 705478661ec9e57733cf9247860a9ed2c8f7926a92a098226bd354dcadfde296 sigs are verified and stored, ready to broadcast
2021-09-27T14:00:36UTC INFO [DLCWallet$DLCWalletImpl] Created funding transaction 5d81ad28079abc2713f8c98518cbeb423284e22024d196e9fb1bc3adfc67cbb3 for contract 705478661ec9e57733cf9247860a9ed2c8f7926a92a098226bd354dcadfde296
2021-09-27T14:00:36UTC INFO [DLCWallet$DLCWalletImpl] Broadcasting funding transaction 5d81ad28079abc2713f8c98518cbeb423284e22024d196e9fb1bc3adfc67cbb3 for contract 705478661ec9e57733cf9247860a9ed2c8f7926a92a098226bd354dcadfde296
2021-09-27T14:00:36UTC INFO [DLCWallet$DLCWalletImpl] Adding UTXO to wallet: b3cb67fcadc31bfbe996d12420e2843242ebcb1885c9f81327bc9a0728ad815d:1 amt=14262870 sats
[...]

PR #3697, opened 2021-09-28
Make it so bitcoind backend exceptions don't get swallowed

Christewart:
This was found when looking into #3695 . 

We have a problem when using bitcoind as a backend for our bitcoin-s wallet. The problem is we don't want to wait for the entire wallet to sync before we actually bind the http server to allow for incoming http requests. We do this by not mapping on the future returned from `BitcoindRpcBackendUtil.syncWalletToBitcoind()`. 

A by product of this though is that if an exception is thrown by `syncWalletToBitcoind()`, we completely ignore it.

This PR adds a callback to make sure we log the error if something is thrown by `syncWalletToBitcoind()`.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -215,16 +215,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
       //intentionally doesn't map on this otherwise we
       //wait until we are done syncing the entire wallet
       //which could take 1 hour
-      _ = syncWalletWithBitcoindAndStartRpcPolling(bitcoind, wallet)
-
-      // Create callbacks for processing new blocks
-      _ =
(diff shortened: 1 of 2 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -430,6 +428,35 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
...
+    val f = BitcoindRpcBackendUtil
+      .syncWalletToBitcoind(bitcoind, wallet)
+      .flatMap(_ => wallet.updateUtxoPendingStates())
+      .flatMap { _ =>
+        if (bitcoindRpcConf.zmqConfig == ZmqConfig.empty) {
+          BitcoindRpcBackendUtil
+            .startBitcoindBlockPolling(wallet, bitcoind)
+            .map(_ => ())
+        } else {
+          Future.unit
```
benthecarman: Could probably put the zmq callback starting in here too

PR #3701, opened 2021-09-30
2021 09 29 delete announcement

Christewart:
@user411 requested these endpoints for the bitcoin-s-ts project in https://github.com/bitcoin-s/bitcoin-s-ts/pull/14

I'm not certain why these are needed rather than spinning up ephemeral docker containers to test against. 

This introduces risk into the public endpoints we expose: 

1. This now allows us to delete announcements that we previously may have published. This means DLCs built off of these announcements will never be settled if this endpoint is used incorrectly.
2. This now allows an oracle to delete attestations for announcement. This was originally implemented in the backend api in #2851. This can lead to a user revealing their private key if they delete the attestations, and resign the announcement with a different outcome. The reason for introducing this on the backend was so that we could use it in krystal bull for cases where a user puts a typo in the attestation. See: https://github.com/bitcoin-s/krystal-bull/pull/45

I would like a more detailed writeup of why this is needed in our bitcoin-s-ts repo so we have some written documentation, previously this has just been discussed on calls. 

This PR still needs documentation on the website, with appropriate warnings around the endpoints.

### DIFF
--- a/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
+++ b/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
@@ -1900,6 +1900,10 @@ object ConsoleCli {
       case SignMessage(message) =>
         RequestParam("signmessage", Seq(up.writeJs(message)))
 
+      case DeleteAnnouncement(eventName) =>
+        RequestParam("deleteannouncement", Seq(up.writeJs(eventName)))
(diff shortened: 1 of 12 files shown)


benthecarman:
needs rebase

benthecarman:
If this is just for testing you can just call `drop()` in the database

user411 (approved):

PR #3702, opened 2021-09-30
Add endpoint for publishing announcements/attestations to the oracle server

rorp:


app/oracle-server/src/main/scala/org/bitcoins/oracle/server/OracleServerMain.scala
```diff
@@ -24,7 +27,12 @@ class OracleServerMain(override val serverArgParser: ServerArgParser)(implicit
     for {
       _ <- conf.start()
       oracle = new DLCOracle()
-      routes = Seq(OracleRoutes(oracle))
+      _ = println(conf.network)
```
Christewart, line 30: remove `println`

Christewart: Also i believe we have the network in the DLC oracle hard coded. We have to do this for backward compatability reasons: 

https://github.com/bitcoin-s/bitcoin-s/blob/48213d9b81d52ddfaa98577640fe527f4fa4678e/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/config/DLCOracleAppConfig.scala#L48

https://github.com/bitcoin-s/bitcoin-s/blob/5a23d324668f82ebb842ae3292b0ef7ddf4ed070/dlc-oracle/src/main/scala/org/bitcoins/dlc/oracle/DLCOracle.scala#L44

https://github.com/bitcoin-s/bitcoin-s/issues/2748

Christewart:
needs rebase

PR #3705, opened 2021-10-01
Update ConsoleCli endpoints, update docs

Christewart:
This fixes things missed in #3703 

This updates the `ConsoleCli` endpoints to use `announcement` rather than `event`. 

This also updates documentation examples to use the new rpc endpoints.


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -1433,35 +1435,35 @@ object ConsoleCli {
...
+                case getEvent: GetAnnouncement =>
                   getEvent.copy(eventName = eventName)
                 case other => other
               }))
         ),
-      cmd("signevent")
-        .action((_, conf) => conf.copy(command = SignEvent("", "")))
-        .text("Signs an event")
+      cmd("signannouncement")
+        .action((_, conf) => conf.copy(command = SignAnnouncement("", "")))
```
benthecarman: sign announcement sounds like we are doing a signature over the annoncement, something like `createattestation` would be better imo

Christewart: Would you recommend changing all of these then ? 

https://github.com/bitcoin-s/bitcoin-s/blob/26d7f99173cf740424c2afadc6e90722626ff562/core/src/main/scala/org/bitcoins/core/api/dlcoracle/DLCOracleApi.scala#L117

Christewart: I do think your rename suggestion is probably right, and I guess if we are changing things we might as well do it now. I can do it on this PR, but would like a concept ACK from you

benthecarman: yeah change those too

benthecarman (approved):

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
user411: What about 'signenum'? The other being 'signdigit' it seems like each is specific to one type of announcement.

Christewart: > What about 'signenum'? The other being 'signdigit' it seems like each is specific to one type of announcement.

I don't have a strong preference, I can add that too as it does make sense IMO. Are you ok with having both `createattestation` and the `signenum` endpoints available?

user411: I'm not sure why we'd need both. I like how signenum / signdigits would align with their underlying types and be absolutely clear which to call based on the type of announcement you are trying to sign.

Christewart: because otherwise i need to go quibble with ben about the exact same naming scheme, i'd rather just get this done

user411: Doesn't both me, but no reason to me to add more entry points for the same thing. I guess I'd just want to name entry points with what they are for for customer ease-of-use.

PR #3707, opened 2021-10-01
Fixed splitIntoRanges bug for zig-zag payout curve

nkohen:
Fixed edge case of numeric range computation where payout(0) > 0 and each point is a new function component on a numeric event. (After and before graphs attached)
Screen Shot 2021-10-01 at 1 11 06 PM
Screen Shot 2021-10-01 at 12 32 53 PM


benthecarman:
Does the spec need to be updated for this?

Christewart:
Appears correct on my end, for reference this is the contract that caused this to be found: https://test.oracle.suredbits.com/contract/numeric/75af51341862ab8d8fb52f4146b8cafb3a5067d877e98d971e57fe3c76e26d19

nkohen:
> Does the spec need to be updated for this?

Nope, @benthecarman this is purely an implementation off-by-one

Christewart (approved):

PR #3721, opened 2021-10-05
Fix logical error on getunconfirmed balance for an account

### LINKED ISSUE #3585: User creditted with 6x the amount of satoshis they deposited
I walked through a user with a DLC today on 20575bcd6898c48ed5e567e38a56e2952eed6a51

The user deposited 100k sats into the wallet, but was credited with 600k sats. This was using a bitcoind backend rather than a neutrino backend.

This was his tx, his output the first 

https://mempool.space/tx/44ce8ff1b272dc33cfb3fa03637ade3c26e3a08b41a21b344009539182c1a536

I proceeded to do a DLC with this user, he was able to fund his side of the DLC using the wallet this corrupted state where he was credited with too much money. After he funded the DLC, the balance went back to correct value which as ~49,000 sats. 

Here is an image the user sent me 

2021-08-20 at 5 13 PM

### LINKED ISSUE #3601: Users funds are not being switched from `unconfirmed` -> `confirmed` when using bitcoind as a backend
This is occuring on e62615dc

I've attached the logs

[bitcoin-s-j.log](https://github.com/bitcoin-s/bitcoin-s/files/7041019/bitcoin-s-j.log)

Screenshot from 2021-08-24 12-28-11

Christewart:
This probably fixes #3585 and possibly fixes #3601

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -316,10 +316,7 @@ abstract class Wallet
         HDAccount.isSameAccount(utxo.privKeyPath.path, account) &&
         utxo.state == ConfirmedReceived
       }
-      confirmedUtxos.foreach { c =>
-        println(s"c=$c")
-      }
-      println(s"^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^")
+      
       confirmedUtxos.foldLeft(CurrencyUnits.zero)(_ + _.output.value)
     }
   }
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -342,11 +342,11 @@ abstract class Wallet
     for {
       allUnspent <- spendingInfoDAO.findAllUnspent()
     } yield {
-      val confirmedUtxos = allUnspent.filter { utxo =>
+      val unconfirmed = allUnspent.filter { utxo =>
         HDAccount.isSameAccount(utxo.privKeyPath.path, account) &&
-        utxo.state == PendingConfirmationsReceived || utxo.state == BroadcastReceived
```
Christewart, line 347: Very subtle logical error. This filter will evaluate to true if 

1. It's the same account and the utxo state is `PendingConfirmationsReceived`
2. if `utxo.state == BroadcastReceived`.

That means regardless of account, the balance will get credited if the utxo state is `BroadcastReceived`. 

What we meant to check is if it (`utxo.state` is EITHER `PendingConfirmationsReceived` OR `BroadcastReceived`) AND the account is the same

benthecarman (approved):
Good catch

Issue #3715, opened 2021-10-03
`executedlc` doesn't work when funding tx is in mempool with bitcoind backend


Christewart:
I'm running on  961f113c28dd8029706b78da274611bcdc1b8d24  trying to do #3631 

I'm at the final step of executing my DLC. I'm using bitcoind as a backend. My funding transaction is still in the mempool as of this writing. The txid of the funding tx is [`0d74896958cee0b142a189211cac215f9f0455462290966d65ebf9fb1865de91`](https://mempool.space/tx/0d74896958cee0b142a189211cac215f9f0455462290966d65ebf9fb1865de91). 

I attempted to close the DLC with this command: 

```
./bitcoin-s-cli executedlc a995aeca167f1c28c2b936e5bb4d87cba061dd9147217501c3523e377fd23439  fdd8688010323032302d75732d656c656374696f6ed3b04a6d7b90c9c43c09ebe5250d583e1c3fc423219b26f6a02ec394a130000a0001ae3e30df5a203ad10ee89a909df0c8ccea4836e94e0a5d34c3cdab758fcaee1447a59ba58797e55b967aa79c89ffec67023578009c4dc1e3dee2fd75277993590c44656d6f637261745f77696e
Request failed: Error -26: non-final
```

This appears to be coming from my bitcoind (`0.21.1`), here are the logs

```
[info] 2021-10-03T22:03:04UTC INFO [DLCWallet$DLCWalletImpl] Executing dlc with contractId=a995aeca167f1c28c2b936e5bb4d87cba061dd9147217501c3523e377fd23439 sigs=Vector(fdd8688010323032302d75732d656c656374696f6ed3b04a6d7b90c9c43c09ebe5250d583e1c3fc423219b26f6a02ec394a130000a0001ae3e30df5a203ad10ee89a909df0c8ccea4836e94e0a5d34c3cdab758fcaee1447a59ba58797e55b967aa79c89ffec67023578009c4dc1e3dee2fd75277993590c44656d6f637261745f77696e)
[info] 2021-10-03T22:03:05UTC INFO [DLCWallet$DLCWalletImpl] Created DLC execution transaction 3cad29be34216e279d13f19153f296c4e3f8240d4a1aa04eb96a3c275478c4d1 for contract a995aeca167f1c28c2b936e5bb4d87cba061dd9147217501c3523e377fd23439
[...]

PR #3724, opened 2021-10-05
2021 10 04 issue 3715

Christewart:
Test case to try and reproduce #3715

Unfortunately I am unable to reproduce, would like eyes from @benthecarman to make sure i'm not missing something.

### DIFF
--- a/testkit-core/.jvm/src/main/resources/logback-test.xml
+++ b/testkit-core/.jvm/src/main/resources/logback-test.xml
@@ -20,7 +20,7 @@
         </encoder>
     </appender>
 
-    <root level="OFF">
+    <root level="ERROR">
         <appender-ref ref="FILE"/>
         <appender-ref ref="STDOUT"/>
     </root>
--- a/testkit-core/.jvm/src/main/resources/logback-test.xml
+++ b/testkit-core/.jvm/src/main/resources/logback-test.xml
@@ -20,7 +20,7 @@
         </encoder>
     </appender>
 
-    <root level="ERROR">
+    <root level="OFF">
         <appender-ref ref="FILE"/>
         <appender-ref ref="STDOUT"/>
     </root>
--- a/testkit-core/.jvm/src/main/resources/logback-test.xml
+++ b/testkit-core/.jvm/src/main/resources/logback-test.xml
@@ -20,7 +20,7 @@
         </encoder>
     </appender>
 
-    <root level="ERROR">
+    <root level="INFO">
         <appender-ref ref="FILE"/>
         <appender-ref ref="STDOUT"/>
     </root>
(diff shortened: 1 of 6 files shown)


benthecarman:
In `createFundedDLCWalletWithBitcoind` you should call `startBitcoindBlockPolling` so you can simulate the exact thing the appServer is doing

Christewart:
> In `createFundedDLCWalletWithBitcoind` you should call `startBitcoindBlockPolling` so you can simulate the exact thing the appServer is doing

I don't think this is relevant. We are failing to broadcast from **our** wallet to bitcoind. Not syncing it from external chain data.

testkit-core/.jvm/src/main/resources/logback-test.xml
```diff
@@ -20,7 +20,7 @@
         </encoder>
     </appender>
 
-    <root level="OFF">
+    <root level="ERROR">
```
benthecarman: shuold revert

testkit/src/main/scala/org/bitcoins/testkit/wallet/BitcoinSDualWalletTest.scala
```diff
@@ -61,41 +66,127 @@ trait BitcoinSDualWalletTest extends BitcoinSWalletTest {
...
+        extraConfig = Some(segwitWalletConf))
+      walletB <- FundWalletUtil.createFundedDLCWallet(
+        nodeApi = nodeApi,
+        chainQueryApi = chainQueryApi,
+        bip39PasswordOpt = getBIP39PasswordOpt(),
+        extraConfig = Some(segwitWalletConf))(config2, system)
+      _ = println(s"---------------------------------")
+      amt = expectedDefaultAmt / Satoshis(2)
+      contractInfo = ContractInfo(amt.satoshis, contractOraclePair)
+      _ = println(s"@@@@@@@@@@@@@@@@@@@@@@@@@@")
```
benthecarman: printlns

testkit/src/main/scala/org/bitcoins/testkit/wallet/FundWalletUtil.scala
```diff
@@ -78,6 +76,17 @@ trait FundWalletUtil extends Logging {
...
 
+  /** Funds a bitcoin-s wallet with 3 utxos, 1,2,3 bitcoin in the utxos
+    * that are funded by REAL utxos from the given bitcoind.
+    * Other test methods such as [[fundWallet()]] create fictitious utxos
+    * that doesn't correspond to a live network.
+    */
+  def fundWalletWithBitcoind(
+      wallet: Wallet,
+      bitcoind: BitcoindRpcClient): Future[FundedTestWallet] = {
+    ???
```
benthecarman: should implement or delete

benthecarman (approved):

PR #3732, opened 2021-10-07
Add ControlMessageHandler

solo1g:
Refactored existing code to move logic related to handling of `ControlPayload` from `PeerMessageReceiver` to a new `ControlMessageHandler` making it similar to how we deal with `DataPayload` in `DataMessageHandler`.

This would be part 1 of breaking down my previous PR (#3512) into smaller units.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
+++ b/node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
@@ -14,7 +14,10 @@ import org.bitcoins.core.api.node.NodeType
 import org.bitcoins.core.protocol.BlockStamp
 import org.bitcoins.node.config.NodeAppConfig
 import org.bitcoins.node.models.Peer
-import org.bitcoins.node.networking.peer.DataMessageHandler
+import org.bitcoins.node.networking.peer.{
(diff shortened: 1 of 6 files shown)


Christewart:
Part 1 of  #3512

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -37,6 +40,8 @@ case class SpvNode(
 
   def bloomFilter: BloomFilter = _bloomFilter.atomicGet
 
+  val controlMessageHandler = ControlMessageHandler(this)
```
Christewart, line 43: nit: `override`

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -37,6 +40,8 @@ case class NeutrinoNode(
 
   override val peers: Vector[Peer] = nodePeer
 
+  val controlMessageHandler: ControlMessageHandler = ControlMessageHandler(this)
```
Christewart, line 43: nit: `override`

Christewart (approved):

PR #3737, opened 2021-10-08
Added WitnessScriptPubKeyV1 for sending to Taproot addresses

nkohen:
Not usable for spending from Taproot addresses, just sending.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
@@ -1159,7 +1159,8 @@ object WitnessScriptPubKey extends ScriptFactory[WitnessScriptPubKey] {
                                                          OP_15,
                                                          OP_16)
 
-  val unassignedWitVersions: Seq[ScriptNumberOperation] = validWitVersions.tail
+  val unassignedWitVersions: Seq[ScriptNumberOperation] =
+    validWitVersions.tail.tail
 
   def apply(asm: Seq[ScriptToken]): WitnessScriptPubKey = fromAsm(asm)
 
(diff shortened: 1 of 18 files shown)


benthecarman (approved):
LGTM maybe worth adding a test

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -1337,6 +1340,27 @@ object P2WSHWitnessSPKV0 extends ScriptFactory[P2WSHWitnessSPKV0] {
...
+
+object WitnessScriptPubKeyV1 extends ScriptFactory[WitnessScriptPubKeyV1] {
+
+  override def fromAsm(asm: Seq[ScriptToken]): WitnessScriptPubKeyV1 = {
+    buildScript(asm.toVector,
+                WitnessScriptPubKeyV1.apply,
+                s"Given asm was not a P2WSHWitnessSPKV0, got $asm")
+  }
+
+  def isValidAsm(asm: Seq[ScriptToken]): Boolean = {
```
Christewart, line 1369: nit: `override`

core/src/main/scala/org/bitcoins/core/wallet/utxo/InputInfo.scala
```diff
@@ -319,11 +322,12 @@ object InputInfo {
                                 witness,
                                 conditionalPath,
                                 hashPreImages)
-      case wspk: UnassignedWitnessScriptPubKey =>
+      case wspk @ (_: UnassignedWitnessScriptPubKey |
+          _: WitnessScriptPubKeyV1) =>
```
Christewart, line 326: any reason to not just do a separate case here and throw the `UnsupportedException`?

Christewart (commented):
LGTM, but like ben said a test case would be nice. I guess this would mean generating a taproot address on bitcoind and then seeing if we can construct transaction that spends to that taproot address? 

I guess this would mean manually constructing the tx? I'm not sure what the best test case could be. There was some chatter on the bitcoin dev mailing list about how to best test taproot wallets, and it appears this isn't straight forward with bitcoind is my understanding: 
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-October/019533.html

@benthecarman what are your thoughts?

benthecarman:
> LGTM, but like ben said a test case would be nice. I guess this would mean generating a taproot address on bitcoind and then seeing if we can construct transaction that spends to that taproot address?
> 
> I guess this would mean manually constructing the tx? I'm not sure what the best test case could be. There was some chatter on the bitcoin dev mailing list about how to best test taproot wallets, and it appears this isn't straight forward with bitcoind is my understanding: [lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-October/019533.html](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-October/019533.html)
> 
> @benthecarman what are your thoughts?

There are taproot addresses in the bech32m tests, could just try to send to one of those (just convert it to regtest)

PR #3739, opened 2021-10-10
Remove unknown fields from BitGoResult

solo1g:
Seems like the BitGo api [https://www.bitgo.com/api/v2/btc/tx/fee](https://www.bitgo.com/api/v2/btc/tx/fee) no longer gives `confidence` and `multiplier` in result and this was causing `FeeRateProviderTest` to fail.


Christewart:
Thanks for fixing this

Christewart (approved):

Issue #3744, opened 2021-10-12
Rework `getbalance` rpcs


Christewart:
From @user411 

- GetBalance()
- GetConfirmedBalance()
- GetUnconfirmedBalance()

These rpcs return strings rather than numbers. I believe the string is of the format `12345 sats` if you use the `--sats` flag. 

https://github.com/bitcoin-s/bitcoin-s/blob/698fe9e800143b79c99c6891ff3536be99bf6e44/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala#L93

Christewart:
Done in #3746

PR #3746, opened 2021-10-12
Update getbalance rpcs to return a number rather than a string

Christewart:
Addresses #3744 .

This is what responses look like now

Screenshot from 2021-10-12 13-28-09

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTest.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTest.scala
@@ -30,7 +30,7 @@ class BitcoinSServerMainBitcoindTest
       } yield {
         assert(info.isSuccess)
         assert(balance.isSuccess)
-        assert(balance.get == "0 sats")
+        assert(balance.get == "0")
         assert(addr.isSuccess)
         assert(blockHash.isSuccess)
       }
--- a/app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTorTest.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTorTest.scala
@@ -33,7 +33,7 @@ class BitcoinSServerMainBitcoindTorTest
       } yield {
         assert(info.isSuccess)
         assert(balance.isSuccess)
-        assert(balance.get == "0 sats")
+        assert(balance.get == "0")
         assert(addr.isSuccess)
         assert(blockHash.isSuccess)
       }
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -280,7 +280,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(responseAs[String] == """{"result":"50","error":null}""")
+        assert(responseAs[String] == """{"result":50,"error":null}""")
       }
     }
 
@@ -295,7 +295,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(responseAs[String] == """{"result":"5000000000","error":null}""")
+        assert(responseAs[String] == """{"result":5000000000,"error":null}""")
       }
     }
 
@@ -310,7 +310,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(responseAs[String] == """{"result":"50","error":null}""")
+        assert(responseAs[String] == """{"result":50,"error":null}""")
       }
     }
 
@@ -325,7 +325,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(responseAs[String] == """{"result":"5000000000","error":null}""")
+        assert(responseAs[String] == """{"result":5000000000,"error":null}""")
       }
     }
 
@@ -340,7 +340,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(responseAs[String] == """{"result":"50","error":null}""")
+        assert(responseAs[String] == """{"result":50,"error":null}""")
       }
     }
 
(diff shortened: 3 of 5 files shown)


app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTest.scala
```diff
@@ -30,7 +30,7 @@ class BitcoinSServerMainBitcoindTest
       } yield {
         assert(info.isSuccess)
         assert(balance.isSuccess)
-        assert(balance.get == "0 sats")
+        assert(balance.get == "0")
```
user411, line 33: I would expect this to be a number now

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -280,8 +280,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"50.00000000 BTC","error":null}""")
+        assert(responseAs[String] == """{"result":"50","error":null}""")
```
user411: I would expect "result" to be a number now

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -296,8 +295,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"5000000000 sats","error":null}""")
+        assert(responseAs[String] == """{"result":"5000000000","error":null}""")
```
user411: Same

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -312,8 +310,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"50.00000000 BTC","error":null}""")
+        assert(responseAs[String] == """{"result":"50","error":null}""")
```
user411: Same

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -328,8 +325,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"5000000000 sats","error":null}""")
+        assert(responseAs[String] == """{"result":"5000000000","error":null}""")
```
user411: Same

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -344,8 +340,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"50.00000000 BTC","error":null}""")
+        assert(responseAs[String] == """{"result":"50","error":null}""")
```
user411: Same

app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -396,8 +391,7 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
 
       Get() ~> route ~> check {
         assert(contentType == `application/json`)
-        assert(
-          responseAs[String] == """{"result":"5000000000 sats","error":null}""")
+        assert(responseAs[String] == """{"result":"5000000000","error":null}""")
```
user411: Same

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -853,4 +849,14 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
       )
     }
   }
+
+  private def formatCurrencyUnit(
+      currencyUnit: CurrencyUnit,
+      isSats: Boolean): BigDecimal = {
```
user411: Does BigDecimal always serialize to string? I think we want to hand back a float vs a string

user411 (commented):
I was expecting the value type to end up as a Javascript number (float) vs Javascript string.

Maybe that doesn't have good enough decimal resolution for our application? It may not at 8 digits and I'm barking up the wrong tree.

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
Christewart: Since `BigDecimal` is arbitrary precision, I believe it has to be serialized to a `String`. I've changed this to be a `Double`. 

https://github.com/com-lihaoyi/upickle/blob/ad9023b3762bac253ea8a5e0472574e19d31cbc0/implicits/src/upickle/implicits/Writers.scala#L52

app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTorTest.scala
```diff
@@ -33,7 +33,7 @@ class BitcoinSServerMainBitcoindTorTest
       } yield {
         assert(info.isSuccess)
         assert(balance.isSuccess)
-        assert(balance.get == "0 sats")
+        assert(balance.get == "0")
```
benthecarman, line 36: This should he a number

user411 (approved):

app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTorTest.scala
benthecarman: Wait nvm console cli always returns a string

benthecarman (approved):

app/server-test/src/test/scala/org/bitcoins/server/BitcoinSServerMainBitcoindTorTest.scala
Christewart, line 36: Ok, so this has to stay a string because the `ConsoleCli` (used in `bitcoin-s-cli` command line utility` serializes every response to a string. 

We may want to review this in the future, but i don't think it needs to change now. All server endpoints now return a json number

https://github.com/bitcoin-s/bitcoin-s/blob/6cbbd8825d24507c31df38defbb9b94cb849056e/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala#L2051

PR #3748, opened 2021-10-13
Store oracle name in the database

rorp:


core/src/main/scala/org/bitcoins/core/crypto/ExtKey.scala
```diff
@@ -46,6 +46,9 @@ sealed abstract class ExtKey extends NetworkElement {
   /** The key at this path */
   def key: BaseECKey
 
+  /** The key name or description */
+  def name: Option[String]
```
Christewart: The `ExtKey` is meant to follow the bip32 spec, there is no `name` in the bip32 spec 

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

Perhaps we can just wrap a key and then change the type in `MasterXPubDAO`? 

I.e. 

```scala
case class MasterXPubKey(extKey: ExtKey, name: Option[String])
```

and then use that in `MasterXPubDAO`?

Christewart (changes requested):

Christewart (approved):

user411 (approved):

Issue #3749, opened 2021-10-13
Issues with wallet routes found while building wallet-ts


user411:
Nomenclature: Server endpoints like `getnewaddress` are "Romanized" to `GetNewAddress()` here.

GetNewAddress() currently requires a label when calling via the JSON interface, but it is optional under the covers. I don't know if we want to always force a label on the user.

Trying to call LabelAddress() on something that already has a label produces an error like:

```
GetNewAddress() test
GetNewAddress() { result: 'tb1qucv84ux5k4y4wvtcpn5g0d05r7x0c48fnlsjy2', error: null }
LabelAddress() tb1qucv84ux5k4y4wvtcpn5g0d05r7x0c48fnlsjy2 test label
SendOracleMessage() error Request failed: [SQLITE_CONSTRAINT_PRIMARYKEY]  A PRIMARY KEY constraint failed (UNIQUE constraint failed: wallet_address_tags.address, wallet_address_tags.tag_type)
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Request failed: [SQLITE_CONSTRAINT_PRIMARYKEY]  A PRIMARY KEY constraint failed (UNIQUE constraint failed: wallet_address_tags.address, wallet_address_tags.tag_type)".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

Calling DropAddressLabel() ahead of LabelAddress() avoids the above error, but address labels are returned in an array, implying support for multiple labels simultaneously.

We should probably deprecate GetAddressTags() on the appServer side as it is identical to GetAddressLabels() and all the other nomenclature is using 'label'.

The new sats/BTC formatter function or pulling `isSats` parameter from the parameters array may have an issue. I am seeing behavior like:

```
GetBalance() true
GetBalance() in sats { result: 12833894, error: null }
GetBalance() false
GetBalance() in BTC { result: 12833894, error: null }
```

Christewart:
The get new address label was solved in #3818 

Unsure about other parts of the issue, need to look into this

PR #3750, opened 2021-10-13
Fix bug where getbalances was always retruning sats

Christewart:
fixes part of #3749 

This bug was introduced in #3746 

This is what output looks like now on a400c04

Screenshot from 2021-10-13 13-33-27

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -355,6 +355,31 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
       spendingTxIdOpt = Some(DoubleSha256DigestBE.empty)
     )
 
+    "return the wallet's balances in bitcoin" in {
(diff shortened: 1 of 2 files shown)


app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
```diff
@@ -355,6 +355,31 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
...
+        .returning(Future.successful(Vector(spendingInfoDb)))
+
+      val route =
+        walletRoutes.handleCommand(
+          ServerCommand("getbalances", Arr(Bool(false))))
+
+      Get() ~> route ~> check {
+        assert(contentType == `application/json`)
+        assert(
+          responseAs[String] == """{"result":{"confirmed":50,"unconfirmed":50,"reserved":-1.0E-8,"total":99.99999999},"error":null}""")
```
Christewart, line 379: is the `1.0E-8` going to be a problem?

user411: No, that's the standard

user411 (approved):

PR #3755, opened 2021-10-15
Initial c-lightning rpc support

benthecarman:
Adds initial support for making c-lightning rpc requests

### DIFF
--- a/clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/CLightningUnixSocketHandler.scala
+++ b/clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/CLightningUnixSocketHandler.scala
@@ -1,0 +1,144 @@
+package com.bitcoins.clightning.rpc
+
+import akka.http.scaladsl.model._
+import akka.util._
+import org.newsclub.net.unix.AFUNIXSocketAddress
+import play.api.libs.json._
+
+import java.io._
+import java.util.UUID
+import scala.concurrent._
+import scala.util._
+
+abstract class CLightningUnixSocketHandler { self: CLightningRpcClient =>
+
+  def clightningCall[T](command: String, parameters: JsValue = JsArray.empty)(
+      implicit reader: Reads[T]): Future[T] = {
+    logger.trace(s"clightning rpc call $command")
+    val responseF = sendRequest(command, parameters)
+
+    responseF.map { payload =>
+      parseResult(payload, command)
+    }
+  }
+
+  private def parseResult[T](payload: RpcResult, commandName: String)(implicit
+      reader: Reads[T]): T = {
+    payload.result match {
+      case Some(result) =>
+        result.validate[T] match {
+          case JsSuccess(value, _) => value
+          case JsError(_) =>
+            val errMsg =
+              s"Could not parse JsResult for command: $commandName json: $result"
+            logger.error(errMsg)
+            throw new IllegalArgumentException(errMsg)
+        }
+      case None =>
+        payload.error match {
+          case Some(errJs) =>
+            import RpcResult._
+            errJs.validate[RpcError] match {
+              case JsSuccess(err, _) =>
+                val errMsg =
+                  s"Error for command=$commandName, error code: ${err.code}, msg: ${err.message}"
+                logger.error(errMsg)
+                throw new RuntimeException(err.message)
+              case JsError(_) =>
+                throw new RuntimeException(
+                  s"Could not parse error for command: $commandName, json: $errJs")
+            }
+          case None =>
+            throw new RuntimeException(
+              s"Did not get result or error for command: $commandName")
+        }
+    }
+  }
+
+  private def parseJson(response: String): RpcResult = {
+    val json = Json.parse(response)
+
+    json
+      .validate[RpcResult]
+      .getOrElse {
+        val errMsg = s"Could not parse $json"
+        logger.error(errMsg)
+        throw new RuntimeException(errMsg)
+      }
+  }
+
+  private def sendRequest(
+      methodName: String,
+      params: JsValue): Future[RpcResult] = {
+    val uuid = UUID.randomUUID().toString
+    val map: Map[String, JsValue] = Map("jsonrpc" -> JsString("2.0"),
+                                        "id" -> JsString(uuid),
+                                        "method" -> JsString(methodName),
+                                        "params" -> params)
+
+    val request: ByteString =
+      HttpEntity(ContentTypes.`application/json`, JsObject(map).toString()).data
+
+    Future.fromTry(doRawCall(request, uuid))
+  }
+
+  // -- below cribbed from https://github.com/clightning4j/JRPClightning/blob/7e1708d7ed67511ab5e67e9f4bc103d6b2c5be32/src/main/java/jrpc/service/socket/UnixDomainSocketRpc.java
+  // todo make this more scala-y
+  import org.newsclub.net.unix.AFUNIXSocket
+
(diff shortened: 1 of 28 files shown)


Christewart:
needs rebase

benthecarman:
CI passing now, should be good for review :)

clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/CLightningUnixSocketHandler.scala
```diff
@@ -0,0 +1,144 @@
...
+                                        "method" -> JsString(methodName),
+                                        "params" -> params)
+
+    val request: ByteString =
+      HttpEntity(ContentTypes.`application/json`, JsObject(map).toString()).data
+
+    Future.fromTry(doRawCall(request, uuid))
+  }
+
+  // -- below cribbed from https://github.com/clightning4j/JRPClightning/blob/7e1708d7ed67511ab5e67e9f4bc103d6b2c5be32/src/main/java/jrpc/service/socket/UnixDomainSocketRpc.java
```
Christewart, line 85: Is there not a dependency that can handle this?

clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/config/CLightningInstance.scala
```diff
@@ -0,0 +1,63 @@
...
+case class CLightningInstance(
+    datadir: Path,
+    network: BitcoinNetwork,
+    rpcFile: File,
+    listenBinding: URI,
+    logFileOpt: Option[File],
+    bitcoindAuthCredentials: PasswordBased,
+    bitcoindRpcUri: URI)
+
+object CLightningInstance
```
Christewart: `CLightningInstanceLocal` ?

clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/config/CLightningInstance.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.bitcoins.core.api.commons.InstanceFactoryLocal
+import org.bitcoins.core.config._
+import org.bitcoins.rpc.config.BitcoindAuthCredentials.PasswordBased
+
+import java.io.File
+import java.net.URI
+import java.nio.file.{Path, Paths}
+import scala.util.Properties
+
+case class CLightningInstance(
```
Christewart: Can we call that `CLightningInstanceLocal` if I understand correctly. We can add the remote later if it is needed?

core/src/main/scala/org/bitcoins/core/protocol/ln/channel/ShortChannelId.scala
```diff
@@ -43,20 +44,28 @@ object ShortChannelId extends Factory[ShortChannelId] {
...
+        apply(blockHeight, txIndex, outputIndex)
       case _: Array[String] => fromHex(str)
     }
+
+  def apply(
+      blockHeight: BigInt,
+      txIndex: Int,
+      outputIndex: Int): ShortChannelId = {
+    require(blockHeight >= 0 && blockHeight <= 0xffffff,
+            "ShortChannelId: invalid block height")
```
Christewart: at the invalid things in the error message so it is easier to debug if something goes wrong

project/Deps.scala
```diff
@@ -103,6 +104,9 @@ object Deps {
     val akkaTestkit =
       "com.typesafe.akka" %% "akka-testkit" % V.akkaActorV withSources () withJavadoc ()
 
+    val jUnixSocket =
+      "com.kohlschutter.junixsocket" % "junixsocket-core" % V.jUnixSocketV
```
Christewart, line 108: interesting, are we sure this isn't handled by akka or netty or something like that?

testkit/src/main/scala/org/bitcoins/testkit/clightning/CLightningRpcTestUtil.scala
```diff
@@ -0,0 +1,319 @@
...
+import org.bitcoins.testkit.util.{FileUtil, TestkitBinaries}
+
+import java.io.{File, PrintWriter}
+import java.net.InetSocketAddress
+import java.nio.file.Path
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Properties
+
+trait CLightningRpcTestUtil extends Logging {
```
Christewart, line 25: Everything looks ok to me in this file, but in the future it would be nice to have a thin interface like `LnApi` or something that all our lightning rpc clients extend, so we don't have to re-write this for every implementation.

testkit/src/main/scala/org/bitcoins/testkit/fixtures/CLightningFixture.scala
```diff
@@ -0,0 +1,127 @@
...
+    with CachedBitcoindV21 {
+
+  override type FixtureParam =
+    (BitcoindRpcClient, CLightningRpcClient, CLightningRpcClient)
+
+  override def withFixture(test: OneArgAsyncTest): FutureOutcome = {
+    makeDependentFixture[FixtureParam](
+      () => {
+        cachedBitcoindWithFundsF.flatMap { bitcoind =>
+          val actorSystemA =
```
Christewart: Why new `ActorSystem`'s here? Also i don't see them getting shutdown in the tear down

Christewart (changes requested):

project/Deps.scala
benthecarman: akka has something that I tried using but couldn't get it to work

benthecarman: @rorp recommened to use this one to me

clightning-rpc/src/main/scala/com/bitcoins/clightning/rpc/CLightningUnixSocketHandler.scala
benthecarman: AFUNIXSocketAddress is doing most of it

testkit/src/main/scala/org/bitcoins/testkit/fixtures/CLightningFixture.scala
benthecarman: Was able to get rid of `ActorSystem`s inside of `CLightningRpcClient`

project/Deps.scala
benthecarman: CLightning is weird and doesn't have a standard json rpc server, you pipe in commands into a file using a unix socket

Christewart:
It seems to take 10 minutes to run the test suite locally. Each test case that uses the `lightningd` takes ~1 minute on my machine. Not sure why

Screenshot from 2021-10-19 10-28-15

benthecarman:
> It seems to take 10 minutes to run the test suite locally. Each test case that uses the lightningd takes ~1 minute on my machine. Not sure why

I get the same thing. In testing it takes awhile for it to see blocks from bitcoind. I think it polls for blocks from bitcoind instead of using zmq so it is slower.

Christewart:
> > It seems to take 10 minutes to run the test suite locally. Each test case that uses the lightningd takes ~1 minute on my machine. Not sure why
> 
> I get the same thing. In testing it takes awhile for it to see blocks from bitcoind. I think it polls for blocks from bitcoind instead of using zmq so it is slower.

I'll file an issue to look into this later,  i don't think it necessarily needs to be solved now though

Christewart (approved):

Issue #3767, opened 2021-10-19
C lightning test cases extremely slow


Christewart:
It seems to take 10 minutes to run the test suite locally. Each test case that uses the `lightningd` takes ~1 minute on my machine. Not sure why

![Screenshot from 2021-10-19 10-28-15](https://user-images.githubusercontent.com/3514957/137942781-b5cb61be-9ed3-4562-a427-6c23ba1860aa.png)

_Originally posted by @Christewart in https://github.com/bitcoin-s/bitcoin-s/issues/3755#issuecomment-946840910_

PR #3768, opened 2021-10-19
Make clightning tests execute async

Christewart:
This runs tests asynchronously for clightning. This doesn't address the root cause documented in #3767 , but it does speed things up ~2 minutes locally for me.


benthecarman (approved):
ACK if CI works, should probably run multiple times

Christewart:
Ran multiple times and no failures on CI

PR #3773, opened 2021-10-20
Storing peers in database

solo1g:
This PR implements logic to use `PeerDao` to store and read peers in database and also adds basic ability to use peers from DNS seeds, peers from [Bitcoin Core's nodes_main.txt](https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt).

This does not include managing multiple connections but has a few minor refactors that would be needed for it, like addition of `PeerData`, to keep the diff for those changes clean. Any existing behavior should not be affected.


node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -38,12 +38,16 @@ case class NeutrinoNode(
...
-  override val peers: Vector[Peer] = nodePeer
-
   val controlMessageHandler: ControlMessageHandler = ControlMessageHandler(this)
 
   override def getDataMessageHandler: DataMessageHandler = dataMessageHandler
 
+  // if a non-empty configPeersOverride is not one of the arguments, then use peers from 'bitcoin-s.conf' file.
+  override def getPeersFromConfig: Vector[Peer] = {
+    if (configPeersOverride.isEmpty) super.getPeersFromConfig
+    else configPeersOverride
```
benthecarman: it might make more sense to do `configPeersOverride ++ super.getPeersFromConfig`

node/src/main/scala/org/bitcoins/node/SpvNode.scala
```diff
@@ -44,6 +42,11 @@ case class SpvNode(
 
   override def getDataMessageHandler: DataMessageHandler = dataMessageHandler
 
+  override def getPeersFromConfig: Vector[Peer] = {
+    if (configPeersOverride.isEmpty) super.getPeersFromConfig
+    else configPeersOverride
```
benthecarman: same here

node/src/main/scala/org/bitcoins/node/networking/peer/ControlMessageHandler.scala
```diff
@@ -75,4 +77,11 @@ case class ControlMessageHandler(node: Node)(implicit ec: ExecutionContext)
         Future.successful(peerMessageReceiver)
     }
   }
+
+  def onPeerInitialization(peer: Peer): Future[Unit] = {
+    node.nodeAppConfig.nodeType match {
+      case _ @(NeutrinoNode | SpvNode) => node.createInDb(peer).map(_ => ())
+      case _                           => Future.unit
```
benthecarman: should avoid doing `case _ =>` and spell out every node type, this way if we add a new type in the future we are forced to handle it here

benthecarman (changes requested):

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
solo1g: Hm, the reason for having `configPeersOverride` is because for tests it is handy to have it as an argument and this way no changes would be required in the existing tests too. Problem is, in tests the config file has a `peers=["localhost"]` and instead of reading from it we specify peers as an argument, hence adding would also add that non existent peer. Any suggestions?

core/src/main/scala/org/bitcoins/core/util/NetworkUtil.scala
```diff
@@ -17,6 +20,63 @@ abstract class NetworkUtil {
...
+    }
+    InetSocketAddress.createUnresolved(uri.getHost, port)
+  }
+
+  // AddrV2 messages give pubkey bytes for TorV3 addresses and in config files addresses are as strings like
+  // dfghj...vbnm.onion hence such conversions to and from bytes to string is needed
+
+  /** Parses TorV3 address bytes (pubkey) to string address */
+  def parseUnresolvedInetSocketAddress(bytes: ByteVector): String = {
+    def sha3(bytes: Array[Byte]): ByteVector = {
```
Christewart: This is likely going to cause problems on the scalajs runtime. I believe scalajs doesn't have access to `java.security`.

See what we do in `CryptoUtil` for providing access to hash functions across different runtimes.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,30 +45,106 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+    val addresses = nodeAppConfig.peers.filter(
+      nodeAppConfig.torConf.enabled || !_.contains(".onion"))
+    val inetSockets = addresses.map(
+      NetworkUtil.parseInetSocketAddress(_, nodeAppConfig.network.port))
+    val peers =
+      inetSockets.map(Peer.fromSocket(_, nodeAppConfig.socks5ProxyParams))
+    peers
+  }
+
+  def getPeers: Future[Vector[Peer]] = {
```
Christewart: Scaladoc of what the intention of this method is, my understanding

```scala
/** 
Returns a set of peers we can connect to on the bitcoin network. Note that this does NOT return the set of current of connected peers, just potential peers to connect to. This set is likely to be different everytime you call this method
*/
```

Is that accurate? If so please add to the method

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,30 +45,106 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+    val peersFromConfig = getPeersFromConfig
+    lazy val peersFromDbF = getPeersFromDb
+    val maxConnectedPeers = 1
```
Christewart: I think we had some point talked about propogating this out to a configuration in `bitcoin-s.conf`  and then make it accessible via `NodeAppConfig.maxConnectedPeers`

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -45,30 +45,106 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
+
+    val allF = for {
+      peersFromDb <- peersFromDbF
+    } yield {
+      val ret = Vector.newBuilder[Peer]
+      //choosing maxConnectedPeers no of elements from lists randomly in the order of peersFromConf, peersFromDB
+      ret ++= Random.shuffle(peersFromConfig).take(maxConnectedPeers)
+      if (maxConnectedPeers - ret.result().size > 0)
```
Christewart: How do we end up with extra peers in some cases? My undersatnding of this logic is that it is needed because we exceed `maxConnectedPeers` in some cases?

node/src/main/resources/hardcoded-peers.txt
```diff
@@ -0,0 +1,676 @@
+2.39.173.126:8333
```
Christewart, line 1: Where are these coming from? It might be nice to add a comment to the top of the file indicating where to get these or how to update these in the future

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -38,12 +38,18 @@ case class NeutrinoNode(
...
   override def chainAppConfig: ChainAppConfig = chainConfig
 
-  override val peers: Vector[Peer] = nodePeer
+  override val peers: Vector[Peer] = peerData.keys.toVector
 
   val controlMessageHandler: ControlMessageHandler = ControlMessageHandler(this)
 
   override def getDataMessageHandler: DataMessageHandler = dataMessageHandler
 
+  // if a non-empty configPeersOverride is not one of the arguments, then use peers from 'bitcoin-s.conf' file.
```
Christewart: nit: Use proper scala doc 

```scala
/**  if a non-empty configPeersOverride is not one of the arguments, then use peers from 'bitcoin-s.conf' file. */
```

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -110,22 +231,12 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
     * object. Internally in [[org.bitcoins.node.networking.P2PClient p2p client]] you will see that
```
Christewart, line 72: I believe this scaladoc needs to change as `clients` is no longer cached.

Christewart: nvm, they are

core/src/main/scala/org/bitcoins/core/util/NetworkUtil.scala
Christewart: I believe the reason this is passing on CI is there isn't any independent test cases for this inside of `coreTest`. We should write a few sanity unit tests. I expect this to break CI

Christewart:
IBD works for me on 8603b5f0d324ccaee905eaa704772fdd6e3bb2d5

Christewart (approved):

PR #3793, opened 2021-11-03
2021 11 03 protocol version

Christewart:
This PR introduces the protocl_version field introduced in this PR on this DLC spec: 

https://github.com/discreetlogcontracts/dlcspecs/pull/163

https://github.com/discreetlogcontracts/dlcspecs/pull/163/files#diff-e0f5b925f91a1c09c6daf26f9a7d28816cb9ff9f08863faca719b7ee0a1cc065R69

I decided to break this out of the branch that I have going to implement the test vectors contained in https://github.com/discreetlogcontracts/dlcspecs/pull/163

This is not intended to change any functionality, but to start incorporating new data structures in to the core module for spec breaking changes.

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
@@ -15,6 +15,7 @@ import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutPoint}
 import org.bitcoins.core.protocol.{BitcoinAddress, BlockStamp}
 import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import org.bitcoins.core.psbt.PSBT
+import org.bitcoins.core.serializers.PicklerKeys
 import org.bitcoins.core.util.TimeUtil
 import org.bitcoins.core.util.TimeUtil._
 import org.bitcoins.core.wallet.fee.SatoshisPerVirtualByte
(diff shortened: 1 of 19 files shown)


app-commons/src/main/scala/org/bitcoins/commons/serializers/Picklers.scala
```diff
@@ -15,6 +15,7 @@ import org.bitcoins.core.protocol.transaction.{Transaction, TransactionOutPoint}
 import org.bitcoins.core.protocol.{BitcoinAddress, BlockStamp}
 import org.bitcoins.core.psbt.InputPSBTRecord.PartialSignature
 import org.bitcoins.core.psbt.PSBT
+import org.bitcoins.core.serializers.PicklerKeys
```
Christewart, line 18: I moved `PicklerKeys` into `core` which is why the diff is so large

testkit-core/src/main/scala/org/bitcoins/testkitcore/gen/TLVGen.scala
```diff
@@ -403,6 +403,7 @@ trait TLVGen {
       }
 
       DLCOfferTLV(
+        protocolVersionOpt = None, //TODO: Comeback and change this
```
benthecarman, line 406: Do you need to do something in this to do

benthecarman (commented):
It is probably worth adding in this pr handling for unknown versions as well

testkit-core/src/main/scala/org/bitcoins/testkitcore/gen/TLVGen.scala
Christewart: Since there isn't any actual serialization changes in this PR, no. I think the `None` type should indicate the current protocol version (pre 163 on dlc spec)

Christewart:
> It is probably worth adding in this pr handling for unknown versions as well

I disagree, this PR doesn't handle any serialization stuff. The next PR will include code to handle serialization and thus the unknown types will have to be handled in that PR.

PR #3794, opened 2021-11-03
Move ValueIterator out of TLVFactory trait

Christewart:
This moves `ValueIterator` out of the trait `TLVFactory`. I don't think there was a need for this an it can lead to very confusing behavior as each `ValueIterator` instance is _technically_ not of the same type as the class is scoped to the `TLVFactory` trait that it is contained in. I.e. each `ValueIterator` is a unique type.


benthecarman (approved):
ACK, I had to do this in a side project

Issue #3813, opened 2021-11-11
Can't remove Offered DLCs, can't reliably send funds from wallet state


user411:
**Describe the bug**
This is a set of behaviors I've seen on my primary wallet. I am having issues using 'Send' to reliably send funds and cannot 'Cancel DLC' on two DLCs in Offered state in the wallet. It feels like some preceding transaction never made it to the network or a utxo being "not reserved" is keeping additional processes from happening.

I tried to send a number of transactions from this wallet to my secondary wallet and one of many tries broadcast and transmitted funds.

wallet-ts GetFundedAddresses() returns exactly one utxo at this point.
```
{
  address: 'tb1qhls7qarxudmktzkdsky9jc84nffyn6gyhmef5e',
  value: 11960234
}
```

Trying to delete a Offered DLC:
```
[info] 2021-11-11T15:43:52UTC INFO [Server] Unhandled error in server:
[info] java.lang.IllegalArgumentException: requirement failed: Some utxos are not reserved, got Vector(SegwitV0SpendingInfo(TransactionOutPoint(b671d0b5ac844a33934909da909b7fd8c6f39abdc9ee14920bd2fe39c5ffa4f9:1),TransactionOutput(124792 sats,wpkh(556cce341a0ef7120a9a84ee6ba617b083ad33a8)),m/84'/1'/0'/1/102,P2WPKHWitnessV0(List(0209eee74b5527f046e2cf09c318c814e78717a4aa40669ff8f800b5c35919ba4a, )),DoubleSha256BDigestBE(b671d0b5ac844a33934909da909b7fd8c6f39abdc9ee14920bd2fe39c5ffa4f9),BroadcastSpent,Some(DoubleSha256BDigestBE(669f0012ab800caf4212800ca4ffc00f8a189c45513f1a59b2684a1214612ec8)),Some(26)))
[info] 	at scala.Predef$.require(Predef.scala:337)
[info] 	at org.bitcoins.wallet.internal.UtxoHandling.unmarkUTXOsAsReserved(UtxoHandling.scala:338)
[info] 	at org.bitcoins.wallet.internal.UtxoHandling.unmarkUTXOsAsReserved$(UtxoHandling.scala:335)
[info] 	at org.bitcoins.wallet.Wallet.unmarkUTXOsAsReserved(Wallet.scala:47)
[...]

PR #3816, opened 2021-11-11
Make sure exception is caught by Future inside of unmarkUTXOsAsReserved

Christewart:
fixes parts of #3813 

The requirement was outside of a Future which meant the exception didn't get handled properly by the .recover() inside of `cancelDLC()`


benthecarman (approved):
We could also change it so it doesn't throw an exception and just unreserves the reserved ones

PR #3817, opened 2021-11-12
Add deleting address tags to to clearAllUtxosAndAddresses()

Christewart:
fixes the second half of #3813 

We weren't clearing tags which have database level foreign keys on the address table.

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
@@ -283,6 +283,9 @@ class AddressHandlingTest extends BitcoinSWalletTest {
       for {
         _ <- addressF
         _ <- wallet.clearAllUtxosAndAddresses()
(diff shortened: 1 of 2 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/AddressHandlingTest.scala
```diff
@@ -274,4 +274,15 @@ class AddressHandlingTest extends BitcoinSWalletTest {
...
   }
+
+  it must "generate an address with a label, and then clear all addresses in the wallet" in {
+    fundedWallet: FundedWallet =>
+      val wallet = fundedWallet.wallet
+      val tag = AddressLabelTag("test")
+      val addressF = wallet.getNewAddress(Vector(tag))
+      for {
+        _ <- addressF
+        _ <- wallet.clearAllUtxosAndAddresses()
```
benthecarman, line 285: should verify that the tag is gone

benthecarman (approved):

PR #3796, opened 2021-11-04
Improve app server validation error handling

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/3795

PR #3819, opened 2021-11-12
revert logging directives on request/response back to DEBUG

Christewart:
Reverts #3734 

We have a consistent exception handler introduced in #3796 so hopefully this isn't a problem anymore.


user411 (approved):
Big fan

PR #3821, opened 2021-11-12
Bitcoind Rpc V22 Support

shruthii625:
- [x] The getpeerinfo RPC returns two new boolean fields, bip152_hb_to and bip152_hb_from, that respectively indicate whether we selected a peer to be in compact blocks high-bandwidth mode or whether a peer selected us as a compact blocks high-bandwidth peer. High-bandwidth peers send new block announcements via a cmpctblock message rather than the usual inv/headers announcements. See BIP 152 for more details. (#19776)

- [x]  getpeerinfo no longer returns the following fields: addnode, banscore, and whitelisted, which were previously deprecated in 0.21. Instead of addnode, the connection_type field returns manual. Instead of whitelisted, the permissions field indicates if the peer has special privileges. The banscore field has simply been removed. (#20755)
    
- [x] The following RPCs: gettxout, getrawtransaction, decoderawtransaction, decodescript, gettransaction, and REST endpoints: /rest/tx, /rest/getutxos, /rest/block deprecated the following fields (which are no longer returned in the responses by default): addresses, reqSigs. The -deprecatedrpc=addresses flag must be passed for these fields to be included in the RPC response. This flag/option will be available only for this major release, after which the deprecation will be removed entirely. Note that these fields are attributes of the scriptPubKey object returned in the RPC response. However, in the response of decodescript these fields are top-level attributes, and included again as attributes of the scriptPubKey object. (#20286)

- [ ] When creating a hex-encoded bitcoin transaction using the bitcoin-tx utility with the -json option set, the following fields: addresses, reqSigs are no longer returned in the tx output of the response. (#20286)

- [x] The listbanned RPC now returns two new numeric fields: ban_duration and time_remaining. Respectively, these new fields indicate the duration of a ban and the time remaining until a ban expires, both in seconds. Additionally, the ban_created field is repositioned to come before banned_until. (#21602)

- [ ]  The setban RPC can ban onion addresses again. This fixes a regression introduced in version 0.21.0. (#20852)

- [ ]  The getnodeaddresses RPC now returns a "network" field indicating the network type (ipv4, ipv6, onion, or i2p) for each address. (#21594)

- [ ]  getnodeaddresses now also accepts a "network" argument (ipv4, ipv6, onion, or i2p) to return only addresses of the specified network. (#21843)

- [ ]  The testmempoolaccept RPC now accepts multiple transactions (still experimental at the moment, API may be unstable). This is intended for testing transaction packages with dependency relationships; it is not recommended for batch-validating independent transactions. In addition to mempool policy, package policies apply: the list cannot contain more than 25 transactions or have a total size exceeding 101K virtual bytes, and cannot conflict with (spend the same inputs as) each other or the mempool, even if it would be a valid BIP125 replace-by-fee. There are some known limitations to the accuracy of the test accept: it's possible for testmempoolaccept to return "allowed"=True for a group of transactions, but "too-long-mempool-chain" if they are actually submitted. (#20833)

- [ ]   addmultisigaddress and createmultisig now support up to 20 keys for Segwit addresses. (#20867)

- [ ]     External signers such as hardware wallets can now be used through the new RPC methods enumeratesigners and displayaddress. Support is also added to the send RPC call. This feature is experimental. See external-signer.md for details. (#16546)

- [ ]    A new listdescriptors RPC is available to inspect the contents of descriptor-enabled wallets. The RPC returns public versions of all imported descriptors, including their timestamp and flags. For ranged descriptors, it also returns the range boundaries and the next index to generate addresses from. (#20226)

- [ ]   The bumpfee RPC is not available with wallets that have private keys disabled. psbtbumpfee can be used instead. (#20891)

- [ ]   The fundrawtransaction, send and walletcreatefundedpsbt RPCs now support an include_unsafe option that when true allows using unsafe inputs to fund the transaction. Note that the resulting transaction may become invalid if one of the unsafe inputs disappears. If that happens, the transaction must be funded with different inputs and republished. (#21359)

- [ ]   We now support up to 20 keys in multi() and sortedmulti() descriptors under wsh(). (#20867)

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/BlockchainResult.scala
@@ -47,7 +47,28 @@ case class GetBlockResult(
     nextblockhash: Option[DoubleSha256DigestBE])
     extends BlockchainResult
 
-case class GetBlockWithTransactionsResult(
+abstract trait GetBlockWithTransactionsResult extends BlockchainResult {
+  def hash: DoubleSha256DigestBE
+  def confirmations: Int
+  def strippedsize: Int
+  def size: Int
+  def weight: Int
+  def height: Int
+  def version: Int
+  def versionHex: Int32
+  def merkleroot: DoubleSha256DigestBE
+  def tx: Vector[RpcTransaction]
+  def time: UInt32
+  def mediantime: UInt32
+  def nonce: UInt32
+  def bits: UInt32
+  def difficulty: BigDecimal
+  def chainwork: String
+  def previousblockhash: Option[DoubleSha256DigestBE]
+  def nextblockhash: Option[DoubleSha256DigestBE]
+}
+
+case class GetBlockWithTransactionsResultPreV22(
     hash: DoubleSha256DigestBE,
     confirmations: Int,
     strippedsize: Int,
@@ -57,7 +78,7 @@ case class GetBlockWithTransactionsResult(
     version: Int,
     versionHex: Int32,
     merkleroot: DoubleSha256DigestBE,
-    tx: Vector[RpcTransaction],
+    tx: Vector[RpcTransactionPreV22],
     time: UInt32,
     mediantime: UInt32,
     nonce: UInt32,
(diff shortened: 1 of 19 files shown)


benthecarman:
Can this be closed? We have v22 and v23 support

Christewart:
Superseded by #4424

PR #3822, opened 2021-11-12
2021 11 12 fix multiple tag issue

Christewart:
fixes the second half of #3749 and adds a test case for it

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressLabelTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressLabelTest.scala
@@ -1,5 +1,6 @@
 package org.bitcoins.wallet
 
+import org.bitcoins.core.wallet.utxo.UnknownAddressTag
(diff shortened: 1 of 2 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/AddressLabelTest.scala
```diff
@@ -0,0 +1,30 @@
...
+  it must "add two labels to the database" in { fundedWallet =>
+    val wallet = fundedWallet.wallet
+    val tag1 = UnknownAddressTag("test_tag_name_1", "test_tag_type_1")
+    val tag2 = UnknownAddressTag("test_tag_name_2", "test_tag_type_2")
+    val addressF = for {
+      address <- wallet.getNewAddress(Vector(tag1))
+      //add another tag to address
+      _ <- wallet.tagAddress(address, tag2)
+      _ <- wallet.tagAddress(address, tag1)
+    } yield succeed
```
benthecarman: should check the tags are correct

benthecarman (approved):

PR #3827, opened 2021-11-15
Add more checks on remote DLC closing

rorp:


benthecarman:
I am confused what this is fixing, is it processing unsigned txs?

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -91,10 +95,30 @@ private[bitcoins] trait DLCTransactionProcessing extends TransactionProcessing {
...
-        cet <-
-          transactionDAO
+        cet <- for {
+          txFromDb <- transactionDAO
             .read(dlcDb.closingTxIdOpt.get)
-            .map(_.get.transaction.asInstanceOf[WitnessTransaction])
+            .map(_.get.transaction)
+        } yield {
+          require(
+            txFromDb.txIdBE == closingTx.txIdBE,
```
benthecarman, line 104: if the concern is witness tx vs non wit tx should compare wtxid

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -91,10 +95,30 @@ private[bitcoins] trait DLCTransactionProcessing extends TransactionProcessing {
         (announcements, announcementData, nonceDbs) <- getDLCAnnouncementDbs(
           dlcDb.dlcId)
 
-        cet <-
-          transactionDAO
+        cet <- for {
+          txFromDb <- transactionDAO
             .read(dlcDb.closingTxIdOpt.get)
-            .map(_.get.transaction.asInstanceOf[WitnessTransaction])
```
benthecarman, line 97: if we aren't getting the witness version of the tx there might be a bug in the node where we aren't requesting the witness version of the tx

Issue #3686, opened 2021-09-22
java.lang.RuntimeException: Request failed: requirement failed: Error, no oracles used


Christewart:
**Describe the bug**
After starting via `bundle/run` on 911fd342 I get an error on startup that is logged everytime we query the backend dlc wallet to update dlc status state. I also have no dlcs in my table on startup.

It seems we have a bug around casting a transactions inside of the `DLCTransactionProcessing`. I'm not sure how can get a base transaction in this flow though, the txid was [`3f929128e70d4186a194a35c5062263938526015e1ead85a453bd5cf71b7b9b9`](https://mempool.space/tx/3f929128e70d4186a194a35c5062263938526015e1ead85a453bd5cf71b7b9b9)

>2021-09-11T22:48:45UTC INFO [PeerMessageSender] Requesting compact filters from FilterSyncMarker(startHeight=700105, stopBlockHash=00000000000000000008b05b8ed352c360524f74cd2c70653427142c79331966)
2021-09-11T22:48:45UTC INFO [DataMessageHandler] We are synced
2021-09-11T22:48:45UTC INFO [DataMessageHandler] Processing 1 filters
2021-09-11T22:49:04UTC INFO [DLCWallet$DLCWalletImpl] Processing tx 3f929128e70d4186a194a35c5062263938526015e1ead85a453bd5cf71b7b9b9 for 1 DLC(s)
2021-09-11T22:49:05UTC ERROR [DataMessageHandler] onTxReceived Callback failed with error: 
java.lang.ClassCastException: class org.bitcoins.core.protocol.transaction.BaseTransaction cannot be cast to class org.bitcoins.core.protocol.transaction.WitnessTransaction (org.bitcoins.core.protocol.transaction.BaseTransaction and org.bitcoins.core.protocol.transaction.WitnessTransaction are in unnamed module of loader 'app')
	at org.bitcoins.dlc.wallet.internal.DLCTransactionProcessing.$anonfun$calculateAndSetOutcome$16(DLCTransactionProcessing.scala:97)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
[...]

PR #3829, opened 2021-11-16
Request witness versions of transactions from nodes

benthecarman:
When a peer is announcing a tx to us and we want to see it we need to request it with a `GetData`. We were defaulting to use the `MsgTx` version which would give us the non witness version of the tx, meaning the transactions that do have witnesses would be  appear to be unsigned to us. I think this is what #3827 attemptingn to catch. This will fix by requesting the witness version of the transaction.


Christewart (commented):
This sounds right to me, but would like confirmation from @rorp

Christewart:
fixes #3686

PR #3834, opened 2021-11-17
Support for Bitcoin Core v22

rorp:


testkit/src/main/scala/org/bitcoins/testkit/rpc/CachedBitcoind.scala
```diff
@@ -189,15 +202,15 @@ trait CachedBitcoindPair[T <: BitcoindRpcClient]
   }
 }
 
-trait CachedBitcoindPairV21
```
Christewart: Why delete the `CachedBitcoindPairV21`? Just create a new trait

testkit/src/main/scala/org/bitcoins/testkit/node/fixture/NodeConnectedWithBitcoind.scala
```diff
@@ -15,9 +16,14 @@ case class SpvNodeConnectedWithBitcoind(
     bitcoind: BitcoindRpcClient)
     extends NodeConnectedWithBitcoind
 
-case class SpvNodeConnectedWithBitcoindV21(
```
Christewart, line 18: Why delete the `SpvNodeConnectedWithBitcoindV21` ? Just create a new case class `SpvNodeConnectedWithBitcoindV22`

testkit/src/main/scala/org/bitcoins/testkit/node/NodeUnitTest.scala
```diff
@@ -93,24 +93,24 @@ trait NodeUnitTest extends BaseNodeTest {
     )(test)
   }
 
-  def withSpvNodeConnectedToBitcoindV21(test: OneArgAsyncTest)(implicit
```
Christewart: Why delete the `V21` version? I use this on other projects and don't need to upgrade to `V22` yet

Christewart (changes requested):

Christewart:
This seems like a duplicate of #3821

Christewart (approved):

PR #3838, opened 2021-11-19
Make DLC node's external IP configurable

rorp:


Christewart (commented):
I think we need to add this to our example config on the website but otherwise LGTM

dlc-node/src/main/scala/org/bitcoins/dlc/node/config/DLCNodeAppConfig.scala
```diff
@@ -55,6 +55,16 @@ case class DLCNodeAppConfig(
     new InetSocketAddress(uri.getHost, uri.getPort)
   }
 
+  lazy val externalIP: Option[InetSocketAddress] = {
+    val path = s"bitcoin-s.$moduleName.external-ip"
+    if (config.hasPath(path)) {
+      val str = config.getString(path)
+      Some(NetworkUtil.parseInetSocketAddress(str, listenAddress.getPort))
```
benthecarman, line 62: the port should probably be configurable in case they are doing some reverse proxy or something

rorp: It is configurable `listenAddress.getPort` provides a default value.

PR #3839, opened 2021-11-20
Disjoint union dlc

Christewart:
rebases and replaces #2824 

This is needed along with #2842 to pass test vectors provided by Thibaut that I am working on in https://github.com/christewart/bitcoin-s-core/tree/2021-10-30-dlc-spec-163


Christewart:
c419932 works on both wallets on my desktop machine. However when trying to boot up on my mac i get this error when doing `getdlcs`

>[info] 2021-11-22T13:48:44UTC INFO [Server] Unhandled error in server:
[info] java.lang.IllegalArgumentException: requirement failed: Error, no oracles used
[info]  at scala.Predef$.require(Predef.scala:337)
[info]  at org.bitcoins.dlc.wallet.util.DLCStatusBuilder$.getOracleOutcomeAndSigs(DLCStatusBuilder.scala:221)
[info]  at org.bitcoins.dlc.wallet.util.DLCStatusBuilder$.buildClosedDLCStatus(DLCStatusBuilder.scala:159)
[info]  at org.bitcoins.dlc.wallet.DLCWallet.$anonfun$buildDLCStatus$5(DLCWallet.scala:1443)
[info]  at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:434)
[info]  at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
[info]  at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
[info]  at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[info]  at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)

I'm still investigating this, i wonder if it is somehow related to #3686 ? 

This seems to only happen with DLCs that is closed based on this announcement: https://oracle.suredbits.com/announcement/1f61de428e0ba44d23afb16c94c37798919e98e04c7c9d8dae943e11dccf4209

I applied the manual database queries listed here on #3686 and things seem to be working again

https://github.com/bitcoin-s/bitcoin-s/issues/3686#issuecomment-950183537

Screen Shot 2021-11-22 at 7 54 09 AM

Christewart:
This is unrelated to this PR, see #3844 for the original comment

Christewart:
It seems that this might be the funding transaction that has a ton of dust outputs associated with it 

https://blockstream.info/tx/d16951b1b91ee315310928be07c52fa92e7c98f73bb94e7b705a6e8ca972b939

nkohen:
In case anyone else plans to do review, just gonna drop this resource here:
https://github.com/Christewart/bitcoin-s-core/compare/disjoint-union-dlc...nkohen:disjoint-union-dlc
this is the diff between the old and the new version of this PR which is nice to look at on top of the actual changes to master

app/gui/src/main/scala/org/bitcoins/gui/dlc/DLCPaneModel.scala
```diff
@@ -304,15 +304,25 @@ class DLCPaneModel(pane: DLCPane)(implicit ec: ExecutionContext)
...
+                        txId,
+                        GUIUtil.epochToDateString(
+                          status.timeouts.contractTimeout),
+                        GlobalData.buildAnnouncementUrl(announcementHash),
+                        true))
+                  case disjointUnionContractInfo: DisjointUnionContractInfo =>
+                    logger.error(
+                      s"Don't know how to show correcit funding transaction dialog for" +
+                        s"disjoint untion contracts, contracts=${disjointUnionContractInfo.contracts}")
+                    ()
```
nkohen: You should alter `FundingTransactionDialog::show` to take a `Vector` of `oracleLinks` and add all of the hyperlinks to the outer VBox's children, everything else should be the same (the only difference is that there should be multiple oracle links)

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/AcceptOfferDialog.scala
```diff
@@ -170,7 +174,9 @@ class AcceptOfferDialog extends CliCommandProducer[AcceptDLCCliCommand] {
                    nextRow)
       nextRow += 1
 
-      offer.contractInfo.contractDescriptor match {
+      offer.contractInfo
+        .asInstanceOf[ContractInfoV0TLV]
```
nkohen: should probably match and handle `ContractInfoV1`, at the very least throwing a better exception

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/BroadcastDLCDialog.scala
```diff
@@ -182,7 +183,7 @@ object BroadcastDLCDialog
                    nextRow)
       nextRow += 1
 
-      status.contractInfo.contractDescriptor.toTLV match {
+      status.contractInfo.contractDescriptors.head.toTLV match {
```
nkohen: @benthecarman how do you think this should be handled?

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/SignDLCDialog.scala
```diff
@@ -197,7 +198,7 @@ object SignDLCDialog
                    nextRow)
       nextRow += 1
 
-      status.contractInfo.contractDescriptor.toTLV match {
+      status.contractInfo.contractDescriptors.head.toTLV match {
```
nkohen: ditto

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
```diff
@@ -70,7 +70,7 @@ object ViewDLCDialog {
       add(getLabel("Event Id"), 0, row)
       add(
         getTextField(
-          status.oracleInfo.singleOracleInfos.head.announcement.eventTLV.eventId),
+          status.oracleInfos.head.singleOracleInfos.head.announcement.eventTLV.eventId),
```
nkohen: I guess in general it seems like the GUI will have at best partial support for disjoint union stuff on this PR so perhaps it would be best to just have that be and open an issue so that this can get merged without blocking other things

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCStatus.scala
```diff
@@ -24,13 +25,22 @@ sealed trait DLCStatus {
...
   def remoteCollateral: CurrencyUnit = totalCollateral - localCollateral
 
+  def announcements: Vector[OracleAnnouncementTLV] = {
+    oracleInfos.flatMap(_.singleOracleInfos.map(_.announcement))
+  }
+
+  def eventId: String = {
+    //is it right to pick this one?
+    announcements.head.eventTLV.eventId
+  }
```
nkohen: probably best to have `def eventIds: Vector[String] = announcements.map(_.eventTLV.eventId)` and then explicitly call head elsewhere because once disjoint union support is added to the gui and other places like that they will probably need the whole list

dlc-test/src/test/scala/org/bitcoins/dlc/NumericDLCTest.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  it should "be able to construct and verify with ScriptInterpreter every tx in a DLC for the large numeric case" in {
+    val numDigits = 17
+    val contractParams =
+      NumericContractParams(numDigits, oracleThreshold = 1, numOracles = 1)
+    val nums = tenRandomNums(numDigits)
+
+    executeForCases(nums, contractParams)
+  }
```
nkohen: This seems to be timing out on CI in the non-secp-binary setting so perhaps it is worth it to ratchet down `numDigits` when secp is not enabled?

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
```diff
@@ -21,6 +28,139 @@ import org.bitcoins.crypto.ECPublicKey
...
+  }
+
+  /** Searches all possible outcomes for one which corresponds to the given signatures.
+    *
+    * Warning: This will return false if too many OracleSignatures are given.
+    */
+  def findOutcome(sigs: Vector[OracleSignatures]): Option[OracleOutcome] = {
+    // TODO: Optimize by looking at nonces
+    // TODO: Optimize using NumericOracleSignatures.computeOutcome
+    allOutcomes.find(verifySigs(_, sigs))
```
nkohen, line 114: An issue would be nice for this so it doesn't get lost (this might be a good early-stage (math) intern project)

nkohen (commented):
I recall now that this PR only implements Disjoint Union DLC support on the backend (which does get tested) but not on the wallet or GUI, so probably a good idea to open an issue for implementing disjoint union support in the wallet and gui in the future (and feel free to disregard some of my comments on the review from before I realized this). Otherwise, looks good once CI is up to snuff :)

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/BroadcastDLCDialog.scala
nkohen: oops, ignore this (see top review comment above)

app/gui/src/main/scala/org/bitcoins/gui/dlc/dialog/ViewDLCDialog.scala
benthecarman: should probably throw `// fixme`s everywhere so we remember where to fix

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
Christewart: #3850

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCStatus.scala
```diff
@@ -24,13 +25,22 @@ sealed trait DLCStatus {
...
   def localCollateral: CurrencyUnit
   def remoteCollateral: CurrencyUnit = totalCollateral - localCollateral
 
+  lazy val announcements: Vector[OracleAnnouncementTLV] = {
+    oracleInfos.flatMap(_.singleOracleInfos.map(_.announcement))
+  }
+
+  lazy val eventIds: Vector[String] = {
+    //is it right to pick this one?
+    announcements.map(_.eventTLV.eventId)
```
nkohen, line 40: ping

core/src/main/scala/org/bitcoins/core/util/EnvUtil.scala
```diff
@@ -14,4 +14,9 @@ object EnvUtil {
   lazy val isCI: Boolean = Properties.envOrNone("CI").contains("true")
 
   def getVersion: String = getClass.getPackage.getImplementationVersion
+
+  def isNativeSecp256k1Disabled: Boolean = {
+    val secpDisabled = System.getenv("DISABLE_SECP256K1")
+    secpDisabled != null && (secpDisabled.toLowerCase == "true" || secpDisabled == "1")
```
nkohen, line 20: don't know what the context is for this but probably also need to check if it `isEnabled` because it could also be disabled due to binaries being missing

nkohen (approved):

core/src/main/scala/org/bitcoins/core/util/EnvUtil.scala
Christewart: This is just checking if the environment variable is set, not if the native libraries are available. This is useful for optimizations on CI machines when we want to reduce the number of test iterations because the flag is set.

PR #3854, opened 2021-11-30
Implemented general payout curves

Christewart:
This replaces #2842

This implements [discreetlogcontracts/dlcspecs#144](https://github.com/discreetlogcontracts/dlcspecs/pull/144)

This PR is problematic as it breaks binary level serialization of contract descriptors at minimum. This means all existing contract descriptors stored in our tables will not be able to be read from the database. When I start this PR with my local wallet that has existing DLCs in it I see these errors. 

To solve this, i implicitly support reading the old protocol format, but not writing it. You will see data structures like `OldTLVPoint` and `OldPayoutFunctionV0TLV` that represent the old serializations. I am open to suggestions for better names. 

Another thing you will see in this PR is the boolean flag `isOldSerialization`. This flag indicates if the bytes were serialized in the 'old' format. This is needed so that our `ValueIterator` works correctly when reading bytes off the wire. The old data formats have a different `byteSize` than the new format. We need this information for parsing.

One thought i've had is turning `isOldSerialization` into an ADT that has three different states

1. Serialization pre-144 ([this PR implements 144](https://github.com/discreetlogcontracts/dlcspecs/pull/144))
2. Serialization pre-163 (#3859 )
3. V0 serialization (what the master dlcspec repo would be after merging 163)

Thoughts?


core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCPayoutCurve.scala
```diff
@@ -1,57 +1,45 @@
...
+    extends TLVSerializable[PayoutFunctionV0TLV] {
+
+  val endpoints: Vector[OutcomePayoutPoint] = {
+    pieces.map(_.leftEndpoint).:+(pieces.last.rightEndpoint)
+  }
+
+  require(pieces.map(_.rightEndpoint) == endpoints.tail,
+          s"Endpoints must line up: $this")
+
+  //should add invariant here to make sure left endpoint is zero?
```
nkohen: go for it :)

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCPayoutCurve.scala
```diff
@@ -83,29 +71,66 @@ case class DLCPayoutCurve(points: Vector[OutcomePayoutPoint]) {
...
-      val payoutWithPrecision =
-        point.value.toLong + (BigDecimal(point.extraPrecision) / (1 << 16))
-      OutcomePayoutPoint(point.outcome, payoutWithPrecision, point.isEndpoint)
-    })
+  def fromPoints(points: Vector[TLVPoint]): DLCPayoutCurve = {
+    //i don't think this is right
+    val curvePiece = DLCPolynomialPayoutCurvePiece(
+      points.map(_.outcomePayoutPoint))
+    val payoutCurve = DLCPayoutCurve(Vector(curvePiece))
+    payoutCurve
```
nkohen: What is the intention here? I think you mean to be using `polynomialInterpolate` or something like that? That is the only situation where points are enough to describe a function?

nkohen: As is, this is definitely wrong as you're putting all the curves into one piece so it is interpolating with some crazy high degree polynomial, instead of straight lines like I assume you want.

Christewart: I want to take a set of points and construct a curve out of it.

Christewart: It's not clear to me how to translate a `Vector[TLVPoint]` into `Vector[PiecewisePolynomialPoint]` that is required for `DLCPayoutCurve.polynomialInterpolate()`. Do i assume that these are all just `PiecewisePolynomialEndpoint` ?

Christewart:
My wallet fires up just fine on 17031a62, this should be ready for review.

Christewart:
Looks like i need to get docs working

benthecarman:
Am able to start up my wallet with this, starting review later

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCPayoutCurve.scala
Christewart: i don't think this is right, can't we have signed numbers? I.e. -100 or something like that

nkohen:
> One thought i've had is turning isOldSerialization into an ADT that has three different states

Probably should just be called `SerializationVersion` or something along those lines but sgtm

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTest.scala
```diff
@@ -2,88 +2,127 @@ package org.bitcoins.core.protocol.dlc
...
     assertThrows[IllegalArgumentException](
       NumericContractDescriptor(func, 2, RoundingIntervals.noRounding))
   }
 
-  it should "correctly create a NumericContractDescriptor" in {
-    val func = DLCPayoutCurve(
+  it should "parse a numeric contract descriptor pre 144" in {
+    //we have to be able to parse old numeric contract descriptors
+    //pre pr 144 as we have old wallets deployed with this
+    //https://github.com/discreetlogcontracts/dlcspecs/pull/144
```
nkohen: nit: might be nice to come up with a name other than pre-144

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTest.scala
```diff
@@ -2,88 +2,127 @@ package org.bitcoins.core.protocol.dlc
...
+                                numDigits = 2,
+                                roundingIntervals =
+                                  RoundingIntervals.noRounding)
+
+    val oldHex =
+      "fda720260002fda7261a0002010000000000000000000000010300000000000000640000fda724020000"
+
+    //i need to write a test case to make sure we can parse both the old serialization
+    //format and the new serialization format
+    //i also need to verify that the data structures above are correct in the diff
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -1001,32 +1041,220 @@ object TLVPoint extends Factory[TLVPoint] {
...
+
+  override val value: ByteVector = u16PrefixedList(points)
+}
+
 /** @see https://github.com/discreetlogcontracts/dlcspecs/blob/8ee4bbe816c9881c832b1ce320b9f14c72e3506f/NumericOutcome.md#curve-serialization */
-case class PayoutFunctionV0TLV(points: Vector[TLVPoint])
+case class PayoutFunctionV0TLV(
+    endpoints: Vector[TLVPoint],
+    pieces: Vector[PayoutCurvePieceTLV],
+    isOldSerialization: Boolean)
```
nkohen: kind of seems (based on the contents of this PR that this should maybe have default value `false`?

core/src/main/scala/org/bitcoins/core/protocol/tlv/ValueIterator.scala
```diff
@@ -33,6 +34,11 @@ case class ValueIterator(value: ByteVector) {
     bytes
   }
 
+  /** Same as [[take]] but doesn't forward the iterator */
+  def takeNoSkip[E <: NetworkElement](factory: Factory[E]): E = {
```
nkohen: this should be called `peek` as I believe that is the convention with data structures like this

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCPayoutCurve.scala
```diff
@@ -1,57 +1,45 @@
...
-
-  /** This Vector contains the function pieces between the endpoints */
-  lazy val functionComponents: Vector[DLCPayoutCurvePiece] = {
-    endpoints.init.zip(endpoints.tail).map { // All pairs of adjacent endpoints
-      case (Indexed(_, index), Indexed(_, nextIndex)) =>
-        DLCPayoutCurvePiece(points.slice(index, nextIndex + 1))
-    }
+case class DLCPayoutCurve(
+    pieces: Vector[DLCPayoutCurvePiece],
+    isOldSerialization: Boolean)
```
nkohen: ditto

nkohen (commented):
Ran `sbt bundle/run` with this branch locally and everything seemed to work fine including loading my one numeric dlc and one enum dlc that I had and previewing the numeric's graph

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTest.scala
Christewart: I'm open to suggestions :sweat_smile:

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: I'd like to table this for now. 

When we get everything in #163 merged, i think its a great idea. For now, while development is still ongoing, i think it's less likely to lead to footguns if the developer has to explicitly state what serialization format they want

core/src/main/scala/org/bitcoins/core/protocol/tlv/DLCSerializationVersion.scala
```diff
@@ -0,0 +1,23 @@
...
+    * on the DLC spec repo. See the diff below
+    * @see [[https://github.com/discreetlogcontracts/dlcspecs/pull/144]]
+    */
+  case object PrePR144 extends DLCSerializationVersion
+
+  /** This represents binary serialization for the case where we have
+    * included support for 144, but not included support for 163 yet
+    * @see [[https://github.com/discreetlogcontracts/dlcspecs/pull/144]]
+    */
+  case object Post144Pre163 extends DLCSerializationVersion
```
benthecarman, line 22: Maybe for betters names can do `Alpha`, `Beta`, then when we have 163 merged can call it `v0.1`

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -1001,32 +1041,224 @@ object TLVPoint extends Factory[TLVPoint] {
...
+      UInt16(extraPrecision).bytes
+  }
+}
+
+object Signed16PTLVNumber extends Factory[Signed16PTLVNumber] {
+
+  override def fromBytes(bytes: ByteVector): Signed16PTLVNumber = {
+    val sign = bytes.head match {
+      case 0 => false
+      case 1 => true
```
benthecarman, line 1116: should use `ValueIterator` here you can can just do `getBool`, if not should do `case FALSE_BYTE` and `case TRUE_BYTE`

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -986,9 +1026,9 @@ case class TLVPoint(
...
 }
 
-object TLVPoint extends Factory[TLVPoint] {
+object OldTLVPoint extends Factory[OldTLVPoint] {
 
-  override def fromBytes(bytes: ByteVector): TLVPoint = {
+  override def fromBytes(bytes: ByteVector): OldTLVPoint = {
     val isEndpoint = bytes.head match {
       case 0 => false
       case 1 => true
```
benthecarman, line 1034: ditto

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCPayoutCurve.scala
```diff
@@ -219,56 +265,205 @@ sealed trait DLCPayoutCurvePiece {
...
+    translatePayout: BigDecimal,
+    a: BigDecimal,
+    b: BigDecimal,
+    c: BigDecimal,
+    d: BigDecimal,
+    leftEndpoint: OutcomePayoutPoint,
+    rightEndpoint: OutcomePayoutPoint)
+    extends DLCPayoutCurvePiece
+    with TLVSerializable[HyperbolaPayoutCurvePieceTLV] {
+  require(a * d != b * c, s"ad cannot equal bc: $this")
```
benthecarman, line 303: nit error string should say `a*d cannot equal b*c` so it is more clear

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTest.scala
```diff
@@ -118,9 +128,6 @@ class ContractDescriptorTest extends BitcoinSUnitTest {
     val oldHex =
       "fda720260002fda7261a0002010000000000000000000000010300000000000000640000fda724020000"
 
-    //i need to write a test case to make sure we can parse both the old serialization
-    //format and the new serialization format
-    //i also need to verify that the data structures above are correct in the diff
     val actual = ContractDescriptorV1TLV.fromHex(oldHex)
```
nkohen, line 131: are there any tests for anything other than `Post144Pre163`?

nkohen (approved):
ACK assuming @benthecarman is happy

core-test/src/test/scala/org/bitcoins/core/protocol/dlc/ContractDescriptorTest.scala
Christewart: The `val oldHex` in this file is a descriptor I created off of master before this was merged

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: i'm not going to touch this code as it is old (`OldTLVPoint`) hopefully we can remove it in the next 3-6 months.

As a separate PR i'm considering something that runs on startup that automatically reads all serialized data structures from the database and writes the new format. After having this for a few releases, we should be able to remove the old data structures all together, at least that is my plan :man_shrugging:

Christewart: i probably won't do this until #3859 is merged though, as the serialization formats included in this PR are still not up to date with the latest DLC protocol.

PR #3851, opened 2021-11-28
2021 11 23 crud action

Christewart:
This PR begins exposing database actions in our `CRUD` trait. There is now a super trait called `CRUDAction`, and has two methods on it `createAllAction`/`createAction`/`deleteAction`/`deleteAllAction`. I'm not going to implement `upsert`/`upsertAll` on this PR.

For more information on how `DBIOAction` please see this slick documentation: https://scala-slick.org/doc/3.3.3/dbio.html

This PR begins to fix #3844 / #3846 / #3119

PR #3872, opened 2021-12-03
Add CRUDAction.{updateAllAction, updateAction}

Christewart:
Related to #3851

This adds `DBIOAction`s for our update methods.

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/CRUDAction.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/CRUDAction.scala
@@ -18,4 +18,43 @@ abstract class CRUDAction[T, PrimaryKeyType](implicit
     createAllAction(Vector(t))
       .map(_.head)
   }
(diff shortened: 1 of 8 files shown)


db-commons/src/main/scala/org/bitcoins/db/CRUDAction.scala
```diff
@@ -18,4 +18,43 @@ abstract class CRUDAction[T, PrimaryKeyType](implicit
     createAllAction(Vector(t))
       .map(_.head)
   }
+
+  def updateAction(t: T): DBIOAction[T, NoStream, Effect.Write] = {
+    updateAllAction(Vector(t)).map(_.head)
```
benthecarman, line 23: Should match on the headOption like we did before so we can throw a more descriptive error

benthecarman (commented):
Looks good to me, just a nit on the error message

PR #3887, opened 2021-12-08
2021 12 08 block parsing bug caused by 65 byte taproot signatures

Christewart:
Fixes bug where we throw an exception if we parse a transaction that spends a taproot transaction with a 65 byte signature. By default, taproot signatures are 64 bytes in length _unless_ you are using a [non default sighash type](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#signature-validation)

Since _technically_ by consensus segwit v0 scripts can have uncompressed public keys ([although they are discouraged by policy!!](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#restrictions-on-public-key-type)), we had this previously. I think it is ok to remove this check for classify uncompressed public keys inside of `v0` witnesses.

This is the tx that created the taproot output [c56e515d8ad9753de2156c0849b9f90cefaca8fa4d0c17884ef0a4e3f6f598d4](https://mempool.space/tx/c56e515d8ad9753de2156c0849b9f90cefaca8fa4d0c17884ef0a4e3f6f598d4)

The transaction that triggered this bug was [66cc4de192001d970244ffe32896282a1994fef80f01d35b216033aeacac1651](https://mempool.space/tx/66cc4de192001d970244ffe32896282a1994fef80f01d35b216033aeacac1651)


testkit-core/.jvm/src/main/resources/logback-test.xml
```diff
@@ -20,7 +20,7 @@
         </encoder>
     </appender>
 
-    <root level="OFF">
+    <root level="DEBUG">
```
benthecarman: nit: can revert

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -180,9 +180,7 @@ object ScriptWitness extends Factory[ScriptWitness] {
     //TODO: eventually only compressed public keys will be allowed in v0 scripts
     //https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#restrictions-on-public-key-type
     val isPubKey = {
-      stack.nonEmpty &&
-      ((stack.head.size == 33 && (stack.head.head == 0x02 || stack.head.head == 0x03))
-        || (stack.head.size == 65 && stack.head.head == 0x04))
+      stack.nonEmpty && (stack.head.size == 33 && (stack.head.head == 0x02 || stack.head.head == 0x03))
```
benthecarman: Will 65 byte pub keys now cause an error?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -180,9 +180,7 @@ object ScriptWitness extends Factory[ScriptWitness] {
...
     }
     if (stack.isEmpty) {
       EmptyScriptWitness
```
benthecarman, line 189: I think we should parse the version, then give indiviudal parsing functions for each version. So we can have `isPubKey` for v0 be different than v1

Christewart: We don't have the version i believe? Thats in the output. This is the witness stack

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
Christewart: It will now get classified as `P2WSHWitnessV0`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
benthecarman: Oh I see

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
benthecarman: Isn't this just switching out one error for another now?

Christewart: Yes, but this thing doesn't throw an exception that kills the process. It would technically be incorrectly typed, if there is a segwit v0 that exists on the blockchain with an uncompressed public key (discouraged by policy)

benthecarman: It'd be better to have a solution that still allows for uncompressed public keys

benthecarman: I think you can accomplish that by having `isPublicKey` check that when `((stack.head.size == 65 && stack.head.head == 0x04))` (uncompressed pubkey) then the stack size should be greater than 1. If it is a taproot sig, it will be the only element, a p2wpkh will have the pubkey and the sig

Christewart: Any suggestions on how to distinguish between a BIP342 schnorr sig and a uncompress public key?

benthecarman: You also could run `ECPublicKey.isFullyValid`

rorp (approved):
LGTM

Christewart:
The commit  15a07ed7a9ae237c7fe66fac898ef256461ab4a0 successfully scans from heights `703,276-713,276` 

>2021-12-08T22:19:26UTC INFO [ScanBitcoind] Count of taproot outputs from height=703276 to endHeight=713276 is 7969. It took 317352ms

benthecarman:
Code looks good, could probably also benefit by doing this as well https://github.com/bitcoin-s/bitcoin-s/pull/3887#discussion_r765255781

Idk if the scan bitcoind is supposed to change but LGTM

Christewart:
I scanned the last 100,000 blocks on 662ac13 and saw no exceptions thrown

>2021-12-08T23:30:19UTC INFO [ScanBitcoind] Count of taproot outputs from height=613277 to endHeight=713277 is 7980. It took 3579401ms

Issue #3891, opened 2021-12-09
app sever takes > 1 minute to start on a raspberry pi


Christewart:
The umbrel platform provides to us 

1. bitcoind
2. tor

It shouldn't take 1 minute to startup when we aren't responsible for managing those services. We should look through the startup code and see what we can parallelize / improve performance on. It seems like we should be able to get this down to ~10 seconds without too much work

https://github.com/bitcoin-s/bitcoin-s/blob/781e77844f733e5474dc716402ee28ad76ca1485/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L186

Christewart:
I ran some benchmarks on this branch: 

https://github.com/Christewart/bitcoin-s-core/commit/10b9cbda2b89019d5c09504de85675083407b920

It seems to take ~6 seconds to startup on my higher end desktop machine using a bitcoind backend


`BitcoinSAppConfig.start()`

>2021-12-10T18:29:33UTC INFO [BitcoinSAppConfig] Done starting BitcoinSAppConfig, it took=2685ms


`BitcoinSServerMain.start()`

>2021-12-10T18:29:38UTC INFO [BitcoinSServerMain] Done start BitcoinSServerMain, it took=6916ms

Christewart:
These are the startup logs from my pi on 174c511c

[pi-bitcoin-s.log](https://github.com/bitcoin-s/bitcoin-s/files/7706544/pi-bitcoin-s.log)

PR #3894, opened 2021-12-10
2021 12 10 issue 3891

Christewart:
Chips away at #3891

This parallelizes the startup logic for our sub modules inside of `BitcoinSAppConfig.start()`

On d060c1c5faa5b6239f81430938fc0a75de5bcf11 it would take this long to startup: 

>2021-12-10T18:29:33UTC INFO [BitcoinSAppConfig] Done starting BitcoinSAppConfig, it took=2685ms

On 003e6a97847f1d186b5f5ee566979397b91baba2 it takes this long to startup:

>2021-12-10T20:04:14UTC INFO [BitcoinSAppConfig] Done starting BitcoinSAppConfig, it took=1618ms

So this cuts a second off of startup time (40% improvement to `BitcoinSAppConfig.start()`). I would assume this would be even larger on the pi, but haven't tested.

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -195,7 +195,7 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
   /** Returns blockchain info, in case of  [[InWarmUp]] exception it retries.
     */
   private def getBlockChainInfo(
-                                 client: BitcoindRpcClient): Future[GetBlockChainInfoResult] = {
+      client: BitcoindRpcClient): Future[GetBlockChainInfoResult] = {
     val promise = Promise[GetBlockChainInfoResult]()
     for {
       _ <- AsyncUtil.retryUntilSatisfiedF(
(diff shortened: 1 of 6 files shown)


benthecarman (approved):

benthecarman (approved):

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -207,6 +212,19 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
   }
 
   def startBitcoindBackend(): Future[Unit] = {
+
+    val bitcoindF = for {
+      client <- bitcoindRpcConf.clientF
```
rorp, line 217: `bitcoindRpcConf` doesn't seem initialized here.

wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -1054,36 +1054,40 @@ object Wallet extends WalletLogger {
...
+      accounts
     }
 
     for {
       _ <- createMasterXpubF
-      _ <- createAccountFutures.flatMap(accounts =>
-        FutureUtil.collect(accounts))
+      actions = createAccountActions
+      accounts <- wallet.accountDAO.safeDatabase.runVec(
+        DBIOAction.sequence(actions))
```
rorp: `DBIOAction.sequence(actions)` should be transactional.

app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
Christewart: This is done here: 

https://github.com/Christewart/bitcoin-s-core/blob/2b4d8a659439b7f376ca97602319c14236a0e594/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L59

Issue #3888, opened 2021-12-08
Single UTXO wallet spending error / DLC accounting error


user411:
**Describe the bug**
I started with a wallet with a single reserved UTXO. I try to create a new DLC Offer, but cannot until the UTXO gets its first confirm at 38:08. Then I make a 7 digit numeric offer at 39:15 and a enum offer at 49:15 before another block clears. At 51:19 a block is received and there is an error processing it with "requirement failed: If we have spent a spendinginfodb, the spendingTxId must be defined.". Then I try to sign an 18 digit Accept message made earlier and get an "Error fetching accounting" at 55:47.

One thing to note is that the 7 digit numeric DLC stayed stuck in Broadcasted state even after 6 confirmations. I had previously seen DLCs stuck in Broadcasted state on initiator and non-initiator wallets.

Doing a `rescan --force` updated the Broadcasted DLC to Confirmed state, but oddly did not change the Last Updated date on the contract or on the 18 digit DLC in Signed state. All other contracts had their Last Updated date overwritten with current time even though they did not ultimately have their state changed by the rescan.

This was all done on a Neutrino connected wallet. The bitcoin-s implementation running was commit d393848cc204b92edc3c200faab05c1f72bf3e75

**Logs**
Attached
[1 utxo wallet, make 7 digit numeric, then make enum, then receive block.txt](https://github.com/bitcoin-s/bitcoin-s/files/7679005/1.utxo.wallet.make.7.digit.numeric.then.make.enum.then.receive.block.txt)

**Platform (please complete the following information):**
 - Ubuntu Linux x64

Christewart:
Other issues that have this same error are #3182 and #2917

>[info] 2021-12-08T18:51:19UTC INFO [DLCWallet$DLCWalletImpl] Processing block=000000000000000fed6e60a14d420f1e35e3a2fc16c393bb3bf00e54c7a8b73e
[info] 2021-12-08T18:51:19UTC ERROR [DLCWallet$DLCWalletImpl] Error processing of block=000000000000000fed6e60a14d420f1e35e3a2fc16c393bb3bf00e54c7a8b73e.
[info] java.lang.IllegalArgumentException: requirement failed: If we have spent a spendinginfodb, the spendingTxId must be defined. Outpoint=TransactionOutPoint(8f00808ae851a78b1c15d91c8694f6cdb99a95baee759080b49a90b68e15b4a6:1)
[...]

PR #3898, opened 2021-12-13
2021 12 12 wallet doublespend tests

### LINKED ISSUE #3890: Investigate impliciations of having `FundTransactionHandling.fundRawTransactionInternal()` `markAsReserved`  param default false
It seems we should have this parameter be default true or not have a default at all so people at the call site have to determine whether to reserve the utxos.

https://github.com/bitcoin-s/bitcoin-s/blob/92ab9faa4536861ad246ddea3c268fea98e4d1e6/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala#L57

By having it default false, that means it is very easy to forget to pass the parameter. Thus you can accidentally double spend utxos intended for something else. I'm not sure if this is the root problem in #3888 , but it does seem like a very easy footgun we have in the codebase.

Christewart:
Related to #3888 and fixes #3890

This PR adds `SpendingInfoDAO.markAsReserved()`. This method should always be used to reserve utxos rather than `SpendingInfoDAO.update()`. 

Previously, we would have a race condition as we didn't check at the _database_ level to see if the utxo was reserved before reserving it. Thus two threads could be trying to reserve utxos, succeed as there wasn't any sychnronization at the databse level.  Now `SpendingInfoDAO.markAsReserved()` transactionally switches a utxo to reserved, and will throw an exception if we couldn't reserve all utxos. 

This PR also removes the default parameter of  `FundTransactionHandling.fundRawTransactionInternal()` `markAsReserved=false`. The caller of `fundRawTransactionInternal()` should be force to specify this.

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/AddressTagIntegrationTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/AddressTagIntegrationTest.scala
@@ -81,7 +81,8 @@ class AddressTagIntegrationTest extends BitcoinSWalletTest {
           .fundRawTransactionInternal(destinations = Vector(output),
                                       feeRate = feeRate,
                                       fromAccount = account,
-                                      fromTagOpt = Some(exampleTag))
+                                      fromTagOpt = Some(exampleTag),
+                                      markAsReserved = true)
       }
       utx = txBuilder.buildTx()
       signedTx = RawTxSigner.sign(utx, utxoInfos, feeRate)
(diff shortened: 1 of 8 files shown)


benthecarman:
With all the functions reserving the utxos now, maybe should add a `recoverWith` that unreserves the utxos. That way if something fails then it won't leave your utxos reserved too.

Christewart:
> With all the functions reserving the utxos now, maybe should add a `recoverWith` that unreserves the utxos. That way if something fails then it won't leave your utxos reserved too.

This should be done in bfea082

Interestingly, the `.transactionally` needs to be the very last thing on the action, otherwise it will not rollback the database transaction if an exception is thrown.

benthecarman:
There you solve if it fails in `fundRawTransactionInternal` but if `finishSend` fails in `sendToOutputs` then the utxos will remain reserved.

Christewart:
> There you solve if it fails in `fundRawTransactionInternal` but if `finishSend` fails in `sendToOutputs` then the utxos will remain reserved.

Addressed in 1cd44cd

PR #3909, opened 2021-12-16
Fix bug where we didn't set spendingTxId when transitioning from `Reserved` -> `PendingConfirmationsSpent`

### LINKED ISSUE #3888: Single UTXO wallet spending error / DLC accounting error

Christewart:
This PR fixes a bug when we try to transition a `Reserved` utxo to `Spent` inside of `TransactionProcessing.processExistingReceivedTxo()`.

The bug was transition a utxo to spent when we are receiving the utxo. A received utxo doesn't have a `spendingTxId`, thus we will violate the invariant that says every utxo that in a spent state must have a `spendingTxId`

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -117,9 +117,6 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
           for {
             _ <- acc
             receivedSpendingInfoDbs <- cachedReceivedF
-            //_ = logger.info(s"received spendingInfoDbs")
-            //_ = receivedSpendingInfoDbs.foreach(u => logger.info(s"u=$u"))
-            // _ = logger.info(s"Done with spendingInfoDbs")
             spentSpendingInfo <- cachedSpentF
             processTxResult <- {
               processTransactionImpl(
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -117,6 +117,9 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
           for {
             _ <- acc
             receivedSpendingInfoDbs <- cachedReceivedF
+            _ = logger.info(s"received spendingInfoDbs")
+            _ = receivedSpendingInfoDbs.foreach(u => logger.info(s"u=$u"))
+            _ = logger.info(s"Done with spendingInfoDbs")
             spentSpendingInfo <- cachedSpentF
             processTxResult <- {
               processTransactionImpl(
@@ -323,7 +326,7 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
       spentSpendingInfoDbsOpt: Option[Vector[SpendingInfoDb]]): Future[
     ProcessTxResult] = {
 
-    logger.debug(
+    logger.info(
       s"Processing transaction=${transaction.txIdBE.hex} with blockHash=${blockHashOpt
         .map(_.hex)}")
 
@@ -356,6 +359,7 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
     for {
       receivedSpendingInfoDbs <- receivedSpendingInfoDbsF
       receivedStart = TimeUtil.currentEpochMs
+      _ = logger.info(s"receivedSpendingInfoDbs=$receivedSpendingInfoDbs")
       incoming <- processReceivedUtxos(transaction = transaction,
                                        blockHashOpt = blockHashOpt,
                                        spendingInfoDbs =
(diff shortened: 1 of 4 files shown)


benthecarman (approved):

benthecarman:
Probably worth adding a test

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -480,7 +482,6 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
           val unreservedTxo = foundTxo.state match {
             case TxoState.Reserved =>
               foundTxo
-                .copyWithSpendingTxId(transaction.txIdBE)
```
benthecarman: why revert this?

Christewart:
> Probably worth adding a test

So i tried to add a test in c2869f8 and can't get it to fail

https://github.com/bitcoin-s/bitcoin-s/blob/c2869f8b626ab1d7aa150d396ad3cea65f6682a1/wallet-test/src/test/scala/org/bitcoins/wallet/UTXOLifeCycleTest.scala#L521

I pushed the code for now as i need to work on something else. I"m not sure why that test case wont fail, i reverted the change to try and verify that the bug fix actually fixes it.

I can't trigger the initial bug in #3888 at all :man_shrugging:

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -461,6 +461,8 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
       transaction: Transaction,
       blockHashOpt: Option[DoubleSha256DigestBE],
       foundTxo: SpendingInfoDb): Future[SpendingInfoDb] = {
+    logger.info(
+      s"processExistingReceivedTxo txId=${transaction.txIdBE.hex} foundingTxo.outPoint=${foundTxo.outPoint}")
```
Christewart: These txids end up being the same, this seems wrong to me?

```
10:53:25.467 INFO Wallet$WalletImpl - processExistingReceivedTxo txId=6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e foundingTxo.outPoint=TransactionOutPoint(6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e:0)
10:53:25.468 INFO Wallet$WalletImpl - processExistingReceivedTxo txId=6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e foundingTxo.outPoint=TransactionOutPoint(6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e:2)
10:53:25.470 INFO Wallet$WalletImpl - processExistingReceivedTxo txId=6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e foundingTxo.outPoint=TransactionOutPoint(6d6e8213a449d5797cc3b629f41ce075b30f626a44478e1f071f10a7c9f3199e:3)
```

You should be able to reproduce with turning the log level = `INFO` and then running

```
 walletTest/testOnly *UTXOLifeCycle* -- -z "mark a utxo as reserved that is still receiving confirmations and not unreserve the utxo"
```

benthecarman: That's right, the tx is the transaction we received in. The outpoint pointing to which output we received with in the transaction. It doesn't have anything to do with spending here

Christewart: This isn't right, how can the `spendingTxId` be the same as the `outpoint` txid? You can't have a self referential transaction like that. If we are _receiving_ funds, how can we have a _spending_ txid?

PR #3921, opened 2021-12-19
Add `getmediantimepast` RPC call

rorp:
It's useful to check if a transaction with a specified lock time can be broadcasted to the network.

Bitcoin Core uses `GetMedianTimePast()` to verify  transaction finality: https://github.com/bitcoin/bitcoin/blob/master/src/validation.cpp#L206

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -107,7 +107,8 @@ class BitcoindRpcClient(override val instance: BitcoindInstance)(implicit
   }
 
   /** Gets the number of compact filters in the database */
-  override def getFilterCount(): Future[Int] = ???
+  override def getFilterCount(): Future[Int] = Future.failed(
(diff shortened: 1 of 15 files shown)


Christewart (commented):
We need to add documentation here: 

https://bitcoin-s.org/docs/applications/server#blockchain

Christewart:
After adding documentation LGTM

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -129,6 +130,13 @@ class BitcoindRpcClient(override val instance: BitcoindInstance)(implicit
   override def epochSecondToBlockHeight(time: Long): Future[Int] =
     Future.successful(0)
 
+  override def getMedianTimePast(): Future[Long] = {
+    for {
+      hash <- getBestBlockHash
+      header <- getBlockHeader(hash)
+    } yield header.mediantime.toLong
```
benthecarman: instead of this you can call `getblockchaininfo` and it has a `mediantime` field

benthecarman (approved):

Christewart (approved):

PR #3906, opened 2021-12-15
2021 12 14 websockets

Christewart:
Add websocket support for events in the wallet. This PR builds off of the existing callbacks we have in the wallet documented here: https://bitcoin-s.org/docs/wallet/wallet-callbacks#callbacks

You can subscribe to the websocket at `ws://localhost:19999/events` on server startup. You should see this log to indicate that the websocket server is bound

>[info] 2021-12-15T13:19:20UTC INFO [Server] Started Bitcoin-S websocket at /127.0.0.1:19999

- [x] tests
- [x] website documentation

### Example Payloads 

Here are some example payloads that our wallet has built in already, you should see these broadcast over the websocket

#### `txprocessed` 

```
{"type":"txprocessed","payload":"0200000000010136aec419a24dff41b3d414a80d7299301f078ae060b30c52a016e03ba7c6f24c0000000000fdffffff02e8030000000000001600144265372b5e4a85657d2fc883f8793beb86592767002c0100000000001600146dcf81f06c4fe9902fd216542278280c3ef2d397024730440220612cc09601a24872c8b9ea0a929cd490eb314ab927f26e4f1ebbe50805c63381022056755f25acfc7517e6cab1c1f3c6cfc5ed93df55d440e5cb665d7a74765c0dc20121035add684b2f0cbd4957debb6c4594579797d5d3ba29cd255de355999e00a7c8922c852000"}
```

#### `reservedutxos`

Reserving the utxos: 
```
[...]

PR #3928, opened 2021-12-20
Remove `BitcoinSServerMain.startedFP`

Christewart:
This was causing me a lot of trouble when testing #3906 and i don't really see what it was necessarily if we just return a `Future[Unit]` from `BitcoinSServerMain.run()`


benthecarman:
This was originally added so the bundle could know when the server was done starting up to launch the gui

Christewart:
> This was originally added so the bundle could know when the server was done starting up to launch the gui

Yes. I'm not sure why we didn't just return a `Future[Unit]` from `BitcoinSServerMain.run()` . This the diff on this PR

To be clear, is there something i am missing? Or does this LGTY

benthecarman:
LGTM but haven't tested

Christewart:
This starts up fine for me on 5c96743b5b24949ef273d57f700c3e7a97d09bee

Screenshot from 2021-12-20 13-36-59

benthecarman (approved):

PR #3946, opened 2021-12-28
2021 12 28 blockprocessed callback

Christewart:
This PR changes the semantics of `blockprocessed` callback introduced in #3906 . 

The reason we needed to do this was the divergence in behavior between `neutrino` and `bitcoind` backends. On master, with a `bitcoind` backened, we will execute callbacks everytime a block comes in on the network, whereas our neutrino wallet will only execute `blockprocessed` when a _compact filter matches_, thus requesting the full block from the network. 

Since `blockprocessed` only contains header information, we use `ChainCallbacks.onBlockHeaderConnected` to sent `blockprocessed` over the websocket.

As a by product of this, a few things had to be changed and reordered. I'll comment on them in the PR.

The payload of `blockprocessed` should not be changed.

### DIFF
--- a/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
+++ b/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
@@ -138,7 +138,7 @@ case class Server(
       .run()
   }
 
-  def walletQueue: SourceQueueWithComplete[Message] = tuple._1
+  def wsQueue: SourceQueueWithComplete[Message] = tuple._1
   def source: Source[Message, NotUsed] = tuple._2
 
   private val eventsRoute = "events"
--- a/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
+++ b/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
@@ -23,7 +23,7 @@ case class Server(
     rpcbindOpt: Option[String],
     rpcport: Int,
     wsConfigOpt: Option[WsServerConfig],
-    source: Source[Message, NotUsed])(implicit system: ActorSystem)
+    wsSource: Source[Message, NotUsed])(implicit system: ActorSystem)
     extends HttpLogger {
 
   import system.dispatcher
@@ -126,7 +126,7 @@ case class Server(
 
   private def wsHandler: Flow[Message, Message, Any] = {
     //we don't allow input, so use Sink.ignore
-    Flow.fromSinkAndSource(Sink.ignore, source)
+    Flow.fromSinkAndSource(Sink.ignore, wsSource)
   }
 
 }
--- a/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
+++ b/app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
@@ -9,15 +9,7 @@ import akka.http.scaladsl.model.ws.Message
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server._
 import akka.http.scaladsl.server.directives.DebuggingDirectives
-import akka.stream.OverflowStrategy
-import akka.stream.scaladsl.{
-  BroadcastHub,
-  Flow,
-  Keep,
-  Sink,
-  Source,
-  SourceQueueWithComplete
-}
+import akka.stream.scaladsl.{Flow, Sink, Source}
 import de.heikoseeberger.akkahttpupickle.UpickleSupport._
 import org.bitcoins.commons.config.AppConfig
 import org.bitcoins.server.util.{ServerBindings, WsServerConfig}
@@ -30,7 +22,8 @@ case class Server(
     handlers: Seq[ServerRoute],
     rpcbindOpt: Option[String],
     rpcport: Int,
-    wsConfigOpt: Option[WsServerConfig])(implicit system: ActorSystem)
+    wsConfigOpt: Option[WsServerConfig],
+    source: Source[Message, NotUsed])(implicit system: ActorSystem)
     extends HttpLogger {
 
   import system.dispatcher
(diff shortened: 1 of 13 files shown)


app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
```diff
@@ -30,7 +22,8 @@ case class Server(
     handlers: Seq[ServerRoute],
     rpcbindOpt: Option[String],
     rpcport: Int,
-    wsConfigOpt: Option[WsServerConfig])(implicit system: ActorSystem)
+    wsConfigOpt: Option[WsServerConfig],
+    source: Source[Message, NotUsed])(implicit system: ActorSystem)
```
Christewart: We now have to construct the SourceQueue outside of `startHttpServer` to be able to pass it into the wallet callbacks, thus we have to take it as a parameter here

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -81,7 +86,7 @@ object BitcoindRpcBackendUtil extends Logging {
 
       val hasFiltersF = bitcoind
         .getFilter(wallet.walletConfig.chain.genesisHashBE)
-        .map(_ => true)
+        .map(_ => false)
```
benthecarman: this should be reverted

benthecarman (commented):
After a quick passover looks good

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -40,9 +44,10 @@ object BitcoindRpcBackendUtil extends Logging {
             _ <- lastConfirmedOpt match {
               case None =>
                 for {
-                  header <- bitcoind.getBestBlockHeader()
-                  _ <- wallet.stateDescriptorDAO.updateSyncHeight(header.hashBE,
-                                                                  header.height)
+                  _ <- doSync(walletHeight = 0,
```
Christewart: This changes an optimization in the wallet where if we had zero txs in the wallet, we would just update our wallet state to bitcoinds height. 

This needed to change to make sure we emit a `blockprocessed` event for every block

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: This skips filter sync when we have a compatible bitcoind node. This should slow down the wallet IBD. We need this so we emit a `blockprocessed` message for every block header, rather than just ones that are compact filter matches.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -230,10 +246,29 @@ object BitcoindRpcBackendUtil extends Logging {
...
-              .foldAsync(wallet) { case (wallet, block) =>
-                wallet.processBlock(block)
+              .foldAsync(wallet) { case (wallet, (block, blockHeaderResult)) =>
+                val blockProcessedF = wallet.processBlock(block)
+                val executeCallbackF: Future[Wallet] = blockProcessedF.flatMap {
+                  wallet =>
+                    chainCallbacksOpt match {
+                      case None => Future.successful(wallet)
+                      case Some(callback) =>
+                        val f = callback
```
Christewart, line 263: where we actually call the `onBlockHeaderConnectedCallback()` for the bitcoind backend. It needs to be done here since we don't have a `chain` module when bitcoind backend is used.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -264,16 +299,21 @@ object BitcoindRpcBackendUtil extends Logging {
...
       ec: ExecutionContext): Future[Cancellable] = {
-    bitcoind.getBlockCount.map { startCount =>
+    val walletSyncStateF = wallet.getSyncState()
+    val resultF: Future[Cancellable] = for {
+      walletSyncState <- walletSyncStateF
+    } yield {
       val numParallelism = Runtime.getRuntime.availableProcessors()
-      val atomicPrevCount: AtomicInteger = new AtomicInteger(startCount)
+      val atomicPrevCount: AtomicInteger = new AtomicInteger(
+        walletSyncState.height)
```
Christewart, line 308: There is a slight change here, we no longer are comparing bitcoind heights, we are comparing wallet height and bitcoind height. Is this a significant change?

user411:
UI in https://github.com/bitcoin-s/bitcoin-s-ts/pull/51 validated against commit e1d65c2

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: why do we need to do this? could we just emit one for the chain tip?

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: this will greatly slow down the sync time

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: this shouldn't be a big change, this is probably a small bug fix. before we would call `startBitcoindBlockPolling` we would make sure we were synced to the api. this should fix an edge case where a block is mined between the syncing being done and the block polling being started.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
benthecarman: IIRC changing from blocks to filters improved sync time by like 5-10x

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: I took your suggestion in  3574e63692c41a0393c0c054fb11eec882f8d48c , I think this is how to implement it? I basically just called `doSync()` on `bitcoind.height - 1`.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
Christewart: I reverted this, but the tradeoff here is that we do not propagate `blockprocessed` events when doing IBD with the bitcoind backend. When our wallet is in steady state, it will execute callbacks for `blockprocessed`.

PR #3950, opened 2021-12-29
Exclude seeds from backup

rorp:


app/server/src/main/scala/org/bitcoins/server/BitcoinSAppConfig.scala
```diff
@@ -259,21 +261,30 @@ object BitcoinSAppConfig extends Logging {
     conf.bitcoindRpcConf
   }
 
-  def zipDatadir(source: Path, target: Path): Path = {
+  def zipDatadir(source: Path, target: Path): Try[Path] = Try {
```
Christewart, line 264: Should we have an option to include the seed? Thoughts @user411 ?

user411: It seems like there may be times where people want to pull a backup from Umbrel for use on desktop. Would extracting the seed damage the security of the Umbrel install at all?

rorp: Extracting the seed could damage security of the wallet: one should never have an electronic copy of the seed. The users should be able to write down the seed instead.

Christewart: > It seems like there may be times where people want to pull a backup from Umbrel for use on desktop. Would extracting the seed damage the security of the Umbrel install at all?

I believe the umbrel environment derives a seed for us based off of the root seed. I don't believe it is possible to reverse engineer the root umbrel seed from a specific app's seed.

Christewart: > Extracting the seed could damage security of the wallet: one should never have an electronic copy of the seed. The users should be able to write down the seed instead.

What if the seed is encrypted? We support encrypted seeds in the `seeds/` directory.

PR #3959, opened 2022-01-04
Patch bug where `zipdatadir` doesn't work if the directory was not created

### LINKED ISSUE #3958: `zipdatadir` needs to create target directory on backups
reported by @user411 , as a hack for now he is creating the `target` directory himself, but this should really be created by the backend.

Here is the stack trace seen from the front end

```
2022-01-04T15:44:33.503Z error: /downloadBackup backupPath is not accessible
ZipDataDir() /Users/user411/.bitcoin-s/backup2/bitcoin-s-backup.zip
SendServerMessage() error Request failed: /Users/ivan/.bitcoin-s/backup2/bitcoin-s-backup.zip (No such file or directory)
2022-01-04T15:44:33.851Z error: uncaught rejection
```

Christewart:
fixes #3958 

We had a bug where if the `target` given as a parameter to `zipdatadir` did not have its parent directories created, we would fail to create the backup. This PR fixes this along with a unit test to verify that the backup occurred. 

Works for me on a0dbdea

Screenshot from 2022-01-04 10-36-12

### DIFF
--- a/app-commons/src/main/scala/org/bitcoins/commons/file/FileUtil.scala
+++ b/app-commons/src/main/scala/org/bitcoins/commons/file/FileUtil.scala
@@ -26,9 +26,6 @@ object FileUtil extends Logging {
     //create directories for target if they DNE
     Files.createDirectories(target.getParent)
 
-    //create the file itself
-    Files.createFile(target)
-
     val zos = new ZipOutputStream(new FileOutputStream(target.toFile))
 
     Files.walkFileTree(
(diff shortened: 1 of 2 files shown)


app-commons/src/main/scala/org/bitcoins/commons/file/FileUtil.scala
```diff
@@ -19,7 +19,18 @@ object FileUtil extends Logging {
...
       fileNameFilter: Vector[Regex] = Vector.empty): Path = {
+    require(
+      !Files.exists(target),
+      s"Cannot overwrite existing target directory=${target.toAbsolutePath}")
+
+    //create directories for target if they DNE
+    Files.createDirectories(target.getParent)
+
+    //create the file itself
+    Files.createFile(target)
```
rorp: This line is not needed.`FileOutputStream` creates the file.

rorp (changes requested):

user411 (approved):

PR #3961, opened 2022-01-04
RPC password authentication

rorp:


Christewart (commented):
Could you add a negative unit test to make sure bad passwords fail?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -2011,6 +2014,8 @@ object ConsoleCli {
         sttp
           .post(uri"http://$host:${config.rpcPort}/")
           .contentType("application/json")
+          .auth
+          .basic("bitcoins", config.rpcPassword)
```
benthecarman, line 2018: why not let the rpc user be set too like most other applications?

Christewart:
Compile failing https://github.com/bitcoin-s/bitcoin-s/runs/4708619274?check_suite_focus=true#step:5:207

benthecarman:
I copied this is another project. It currently has a problem where if you don't send the correct authentication it doesn't send an error back, just an empty result and error.

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
rorp: Because there's only one user. It doest's matter what's its name.

rorp:
> I copied this is another project. It currently has a problem where if you don't send the correct authentication it doesn't send an error back, just an empty result and error.

What do you mean? It responds as RFC7235 requires:

```shell
$ curl -i --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "getinfo", "params": []}' -H "Content-Type: application/json" http://127.0.0.1:9999/
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Basic realm="auth",charset=UTF-8
Server: akka-http/10.2.7
Date: Mon, 10 Jan 2022 19:42:56 GMT
Content-Type: text/plain; charset=UTF-8
Content-Length: 77

The resource requires authentication, which was not supplied with the request
```

```shell
$ curl -i -u user:pwd --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "getinfo", "params": []}' -H "Content-Type: application/json" http://127.0.0.1:9999/
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Basic realm="auth",charset=UTF-8
Server: akka-http/10.2.7
Date: Mon, 10 Jan 2022 19:49:03 GMT
Content-Type: text/plain; charset=UTF-8
Content-Length: 38

The supplied authentication is invalid
```

app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
```diff
@@ -29,6 +29,15 @@ case class Server(
 
   import system.dispatcher
 
+  if (rpcPassword.isEmpty) {
+    if (rpchost == "localhost" || rpchost == "127.0.0.1") {
```
Christewart, line 34: This seems like a really good idea. Best of both worlds IMO :+1:

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
Christewart: > why not let the rpc user be set too like most other applications?

We are copying what eclair does for their backend. I'm not sure how we can correctly map users to access to backend services (for instance, does Alice have access to "bob-wallet") and vice versa

app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
```diff
@@ -65,19 +66,29 @@ case class Server(
     }
   }
 
+  def authenticator(credentials: Credentials): Option[Done] =
```
Christewart: Consider making `private`

app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
```diff
@@ -119,8 +130,10 @@ case class Server(
   private val eventsRoute = "events"
 
   private def wsRoutes: Route = {
-    path(eventsRoute) {
-      Directives.handleWebSocketMessages(wsHandler)
+    authenticateBasic("auth", authenticator) { _ =>
```
Christewart, line 147: Whats the benefit of adding auth on the websocket? Currently we only write things to the websocket and do not support requests sent to us over the websocket.

Christewart (changes requested):
I see a unit test for websocket authorization but not the rpc

app/server-routes/src/main/scala/org/bitcoins/server/routes/Server.scala
rorp: Same benefit as for the HTTP server - it adds access control.

Christewart (approved):

PR #3962, opened 2022-01-05
Address ben's code review on #3854

Christewart:
I got a little merge happy on #3854 and didn't address ben's last code review


benthecarman (changes requested):
Docs failing to compile

otherwise ACK on CI

Issue #3869, opened 2021-12-02
Exception throwing during rescan


Christewart:
When rescanning my wallet on d393848cc204b92edc3c200faab05c1f72bf3e75 with my wallet connected via neutrino node to the bitcoin network. It appears my rescan died on block height `685523` due to the dlc module of the wallet

```
2021-12-02T19:12:33UTC INFO [DLCWallet$DLCWalletImpl] Adding UTXO to wallet: c1e817693548348bdc11ec6ef3a21a34039c14e89209ee94a034641cbfd92d13:2 amt=9822359 sats
2021-12-02T19:12:33UTC INFO [DLCWallet$DLCWalletImpl] Successfully inserted UTXO c1e817693548348bdc11ec6ef3a21a34039c14e89209ee94a034641cbfd92d13:2 into DB
2021-12-02T19:12:33UTC INFO [DLCWallet$DLCWalletImpl] Processing tx 132dd9bf1c6434a094ee0992e8149c03341aa2f36eec11dc8b3448356917e8c1 for 1 DLC(s)
2021-12-02T19:12:33UTC INFO [DLCWallet$DLCWalletImpl] 1 txos are now confirmed!
2021-12-02T19:12:40UTC INFO [DLCWallet$DLCWalletImpl] Finished processing of block=00000000000000000001e6900be9b4987ba898758cf8c58f4626e3528bc6947c. It took 12190ms
2021-12-02T19:12:40UTC INFO [DLCWallet$DLCWalletImpl] Found 1 matches from start=685323 to end=685422
2021-12-02T19:12:40UTC INFO [DLCWallet$DLCWalletImpl] Requesting 0 block(s)
2021-12-02T19:12:40UTC INFO [DLCWallet$DLCWalletImpl] Found 0 matches from start=685423 to end=685522
2021-12-02T19:12:40UTC INFO [DLCWallet$DLCWalletImpl] Found a match in block 685523
2021-12-02T19:12:41UTC INFO [DLCWallet$DLCWalletImpl] Requesting 1 block(s)
2021-12-02T19:12:44UTC INFO [DataMessageHandler] Received block message with hash 000000000000000000085a8452a22250569563f4adbd23cb3c22b116ee934296
2021-12-02T19:12:44UTC INFO [DLCWallet$DLCWalletImpl] Processing block=000000000000000000085a8452a22250569563f4adbd23cb3c22b116ee934296
[...]

PR #3965, opened 2022-01-06
2021 01 06 tlv invariants

Christewart:
Add invariants to various TLV data structures. Mainly around making sure `Vector`s aren't empty.

@nkohen and i discussed this might be a easy way to get to the bottom of #3869 as when we delete adaptor signatures from the database, we might end up have a `CETSignaturesTLV` that is empty. This gives us the opportunity to "fail-fast" and see the root cause of the issue (not having adaptor sigs).

I haven't tested this out yet on #3869, but this seems like a good idea to add regardless.

I had modify the type returned by `DLCDataManagement.executorAndSetupFromDb()` to return an `Option[DLCExecutorWithSetup]`. The `None` case represents when we have pruned the cet signatures from the database to save on disk space.

I had to bubble this `Option` throughout the codebase, and try to appropriately handle it. I left comments in the diff where i had questions and could use review there.


benthecarman (approved):

nkohen (changes requested):
I'm not sure this will catch what we discussed, and also I don't think that it is necessarily good to have exceptions thrown at this level, the invariants should live on the deserialized versions not the serialized ones (for example, we could have places where we don't go through TLV when communicating between db and wallet). For example, an invariant should probably be on our internal DLCAccept and DLCSign types.

Christewart:
> I'm not sure this will catch what we discussed, and also I don't think that it is necessarily good to have exceptions thrown at this level, the invariants should live on the deserialized versions not the serialized ones (for example, we could have places where we don't go through TLV when communicating between db and wallet). For example, an invariant should probably be on our internal DLCAccept and DLCSign types.

That introduces just a few failures on 8186f14 :sweat_smile: 

Screenshot from 2022-01-06 12-39-18

Christewart:
> > I'm not sure this will catch what we discussed, and also I don't think that it is necessarily good to have exceptions thrown at this level, the invariants should live on the deserialized versions not the serialized ones (for example, we could have places where we don't go through TLV when communicating between db and wallet). For example, an invariant should probably be on our internal DLCAccept and DLCSign types.
> 
> That introduces just a few failures on [8186f14](https://github.com/bitcoin-s/bitcoin-s/commit/8186f1409eaac4824883bcb62fd9400e4cf9b2a3) sweat_smile
> 
> Screenshot from 2022-01-06 12-39-18

Commenting out one line gets all the test cases too pass on 8186f14

https://github.com/bitcoin-s/bitcoin-s/blob/4ca586ca462cd0bf029cfe43f698d9e49da4aa6e/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala#L64

My understanding is this line deletes the adaptor signatures in the database. @benthecarman hypothesized this was the root cause of #3869 , still haven't been able to prove it yet but it could be related to this change.

I think i'm going to wrap `Vector[DLCCETSignaturesDb]` -> `Option[Vector[DLCCETSignaturesDb]]` so we have a typesafe way of representing when we do / do not have the cet sigs available to us.

https://github.com/bitcoin-s/bitcoin-s/blob/f71d3567ed21df6bb050ee184d6fe31b4369c654/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala#L317

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1272,15 +1272,57 @@ abstract class DLCWallet
...
-      _ = logger.info(
-        s"Created DLC execution transaction ${tx.txIdBE.hex} for contract ${contractId.toHex}")
+            //lets try to retrieve it from our transactionDAO
+            val dlcDbOptF = dlcDAO.findByContractId(contractId)
 
+            for {
+              dlcDbOpt <- dlcDbOptF
+              _ = require(
+                dlcDbOpt.isDefined,
+                s"Could not find dlc associated with this contractId=${contractId}")
```
benthecarman: nit: call `.hex` on `contractId`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1272,15 +1272,57 @@ abstract class DLCWallet
...
+              dlcDb = dlcDbOpt.get
+              _ = require(
+                dlcDb.closingTxIdOpt.isDefined,
+                s"If we don't have CET signatures, the closing tx must be defined, contractId=$contractId")
```
benthecarman: nit: call `.hex` on `contractId`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1272,15 +1272,57 @@ abstract class DLCWallet
...
+              closingTxId = dlcDb.closingTxIdOpt.get
+              closingTxOpt <- transactionDAO.findByTxId(closingTxId)
+            } yield {
+              require(
+                closingTxOpt.isDefined,
+                s"Could not find closing tx for DLC in db, contactId=$contractId closingTxId=$closingTxId")
```
benthecarman: nit: call `.hex` on `contractId` and `closingTxId`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCDataManagement.scala
```diff
@@ -307,14 +307,20 @@ private[bitcoins] trait DLCDataManagement { self: DLCWallet =>
...
-    } yield (dlcDb,
-             contractData,
-             dlcOffer,
-             dlcAccept,
-             refundSigs.head,
-             contractInfo,
-             fundingInputs,
-             outcomeSigs)
+    } yield {
+      logger.info(s"outcomeSigs=$outcomeSigs dlcId=${dlcId.hex}")
```
benthecarman: `outcomeSigs` could be very large here, maybe make this a trace level log or simplify what it is printing

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -31,50 +32,67 @@ private[bitcoins] trait DLCTransactionProcessing extends TransactionProcessing {
...
+        executorWithSetupOptF.flatMap { case executorWithSetupOpt =>
+          executorWithSetupOpt match {
+            case Some(exeutorWithSetup) =>
+              calculateAndSetStateWithSetupDLC(exeutorWithSetup.setup,
+                                               dlcDb,
+                                               txId)
+            case None =>
+              //what do we do if we don't have the cet sigs?
+              //this means we have already deleted the cet sigs
+              //just return the dlcdb given to us?
```
benthecarman: yeah this sounds correct

benthecarman (commented):
otherwise lgtm

PR #3973, opened 2022-01-11
Modify estimatefee endpoint to return a number rather than a string

### LINKED ISSUE #3803: WalletRoutes 'estimatefee'  returns string instead of number
**Describe the bug**
WalletRoutes 'estimatefee' JSON endpoint returns string instead of number like "1 sats/vbyte". The frontend would like the raw number and know the return is in units of sats/vbyte.

**To Reproduce**
1. Call WalletRoutes 'estimatefee', note it's a string instead of number

**Expected behavior**
A number is returned.

We'll have to add the string unit label part back to the desktop UI where it's necessary.

Christewart:
fixes #3803 

This changes the return type of the `estimatefee` endpoint from a string to a number. The number always has a unit of `sats/vbyte`. 

Response payload now looks like 

```
{"result":1,"error":null}
```

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/feeprovider/FeeRateApi.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/feeprovider/FeeRateApi.scala
@@ -6,6 +6,6 @@ import scala.concurrent.Future
 
 trait FeeRateApi {
 
-  def getFeeRate: Future[FeeUnit]
+  def getFeeRate(): Future[FeeUnit]
 
 }
--- a/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
+++ b/core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
@@ -35,6 +35,13 @@ sealed abstract class FeeUnit {
   def toLong: Long = currencyUnit.satoshis.toLong
 
   override def toString: String = s"$toLong ${factory.unitString}"
+
+  def toSatsPerVByte: SatoshisPerVirtualByte = this match {
+    case s: SatoshisPerVirtualByte => s
+    case s: SatoshisPerByte        => s.toSatsPerVByte
+    case s: SatoshisPerKiloByte    => s.toSatsPerVByte
+    case s: SatoshisPerKW          => s.toSatsPerVByte
+  }
 }
 
 trait FeeUnitFactory[+T <: FeeUnit] {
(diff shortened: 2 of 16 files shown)


core/src/main/scala/org/bitcoins/core/api/feeprovider/FeeRateApi.scala
```diff
@@ -6,6 +6,6 @@ import scala.concurrent.Future
 
 trait FeeRateApi {
 
-  def getFeeRate: Future[FeeUnit]
+  def getFeeRate(): Future[FeeUnit]
```
Christewart, line 9: i had to modify this because i couldn't figure out how to mock a no argument method correctly 🤷‍♂️ . I can revert this if there is a solution 

https://github.com/bitcoin-s/bitcoin-s/pull/3973/files#diff-b5f2265432eb3e3a5f714135702e23e99dc538645bbe2a94ef69a062b553c479R28

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
```diff
@@ -35,6 +35,13 @@ sealed abstract class FeeUnit {
   def toLong: Long = currencyUnit.satoshis.toLong
 
   override def toString: String = s"$toLong ${factory.unitString}"
+
+  def toSatsPerVByte: SatoshisPerVirtualByte = this match {
+    case s: SatoshisPerVirtualByte => s
+    case s: SatoshisPerByte        => s.toSatsPerVByte
+    case s: SatoshisPerKiloByte    => s.toSatsPerVByte
+    case s: SatoshisPerKW          => s.toSatsPerVByte
+  }
```
benthecarman: won't this just infinitely loop

benthecarman:
`BitGoFeeRateProvider` returns in `SatoshisPerKiloByte`, maybe should add a unit to the return so it is more clear, can just call `.factory.unitString`

core/src/main/scala/org/bitcoins/core/wallet/fee/FeeUnit.scala
benthecarman: for example `SatoshisPerKiloByte` doesn't have this implemented

Christewart: i am not a smart man :sweat_smile:

Christewart:
> `BitGoFeeRateProvider` returns in `SatoshisPerKiloByte`, maybe should add a unit to the return so it is more clear, can just call `.factory.unitString`

@user411 's idea for the `estimatefee` endpoint is to always return `sats/vbyte` and handle conversions from various fee providers internally and give a consistent unit on the endpoint. This means the app dev doesn't have to think and knows they always get `sats/vbyte` back.

user411 (approved):

PR #3974, opened 2022-01-12
Create FallbackFeeRateApi

benthecarman:


fee-provider/src/main/scala/org/bitcoins/feeprovider/AggregateFeeRateApi.scala
```diff
@@ -0,0 +1,34 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
+
+case class AggregateFeeRateApi(providers: Vector[FeeRateApi])(implicit
+    ec: ExecutionContext)
+    extends FeeRateApi {
+
+  override def getFeeRate(): Future[FeeUnit] = {
+    val init: Option[FeeUnit] = None
+    val retOptF = FutureUtil.foldLeftAsync(init, providers) {
```
Christewart: Just be aware , this does query all of them. I was looking for a better solution, i think we want something like `Future.foldUntilSuccess()` but that doesn't seem to exist :-(. 


There is a `Future.firstCompletedOf()` but that returns either failures or successes, not just succeses and i believe it starts all the Futures.

benthecarman: That is kinda what this is doing it only attempts the fee rate api until it has a success.

Christewart: Ahh ok. I see. That makes sense to me.

Christewart (approved):

PR #3975, opened 2022-01-12
Implement recovery for when we cannot receive a fee rate from a FeeRateApi

### LINKED ISSUE #3972: `estimatefee` endpoint returns a 500 internal server error in some cases
This was reported by @user411 when he is running on `umbrel-dev`. 

>I don’t know technically, but on regtest umbrel-dev by default “there’s not enough data for fee estimation” per @rorp 

>You are getting this in response to the estimatefee endpoint right? https://bitcoin-s.org/docs/applications/server#util - @Christewart 

>Yes, I get a 500 from that endpoint in this scenario @user411

Maybe there is some sort of IBD going on with bitcoind and we cannot get fee estimates? :man_shrugging:

Christewart:
fixes #3972

Now we return `-1 sats/vbyte` when we cannot estimate a fee rate. We also log the exception internally so we can diagnose why we aren't getting fee rate estimates from the provider. 

I began implementing some test cases for this, but after discussion with @benthecarman i decided to not handle this at the `BitcoindRpcClient` level as this doesn't solve the problem when other fee providers are down.

This along with #3973 should address @user411 's requests.

Screenshot from 2022-01-12 10-47-05

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/rpc/CachedBitcoind.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/rpc/CachedBitcoind.scala
@@ -22,10 +22,7 @@ import scala.concurrent.{Await, Future}
   * This does mean that test cases have to be written in such a way where assertions
   * are not dependent on specific bitcoind state.
   */
-trait CachedBitcoind[T <: BitcoindRpcClient] { _: BitcoinSAkkaAsyncTest => }
-
-trait CachedBitcoindFunded[T <: BitcoindRpcClient] extends CachedBitcoind[T] {
-  _: BitcoinSAkkaAsyncTest =>
+trait CachedBitcoind[T <: BitcoindRpcClient] { _: BitcoinSAkkaAsyncTest =>
 
   /** Flag to indicate if the bitcoind was used
     *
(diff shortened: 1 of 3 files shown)


testkit/src/main/scala/org/bitcoins/testkit/rpc/CachedBitcoind.scala
```diff
@@ -39,6 +36,44 @@ trait CachedBitcoindFunded[T <: BitcoindRpcClient] extends CachedBitcoind[T] {
     */
   protected val isBitcoindUsed: AtomicBoolean = new AtomicBoolean(false)
 
+}
+
+/** A cached bitcoind that has zero blocks in its chainstate. This is useful for
+  * testing at certain times when we need to make sure bitcoind is ONLY initialized, no chain state.
+  */
+trait CachedBitcoindNoFunds[T <: BitcoindRpcClient] extends CachedBitcoind[T] {
```
Christewart, line 44: I began implementing these to try and make a reproducible test case with bitcoind, but after talking with ben i decided to just handle this at `WalletRoutes`. I'm going to keep these fixtures around in case we need them in the future.

user411 (approved):

benthecarman (approved):

PR #3979, opened 2022-01-13
upgrade to scala-js 1.8.0

Christewart:
replaces #3895 and fixes compiler warnings in 3895


Christewart:
Takes care of #3970 too -- these dep bumps needed to be made in tandem

PR #3980, opened 2022-01-13
Make RPC authentication optional for localhost

rorp:
Fixes bug in #3961 where we were _always_ using authentication. The intended behavior is to only require authentication when we are binding to a network interface that _isn't_ `localhost` or `127.0.0.1`.


Christewart (approved):
works for me

PR #3986, opened 2022-01-16
Implement rpc password in GUI

### LINKED ISSUE #3984: Wallet is not showing balance 'The supplied authentication is invalid'
**Describe the bug**
In my dev environment on Mac running on bitcoin-s master e520b4 I currently cannot see my wallet balance with a server password set or not.

I turned off all proxy and frontend projects to make sure they were not hitting the server with any requests while it starts up and this behavior persists.

I have not done any transactions on the wallet since we added authorization. I have started and stopped it with and without a password many times.

**To Reproduce**
1. I don't know how to reproduce, but you might try setting and unsetting the password on the server and running a number of times. I have been running with `bundle/run`.

**Expected behavior**
That I can see my balances.

**Logs**
[cannotSeeBalances.txt](https://github.com/bitcoin-s/bitcoin-s/files/7873953/cannotSeeBalances.txt)
Screen Shot 2022-01-14 at 5 45 49 PM
```
    server {
        # The port we bind our rpc server on
        rpcport = 9999

        # The ip address we bind our server too
        rpcbind = "127.0.0.1"

        password = "password"

        wsport = 19999
        wsbind = "127.0.0.1"
    }
```

**Platform (please complete the following information):**
 - OS: macOS

Christewart:
fixes #3984 

This implements rpc password authentication in the GUI. We do this by reading the password from the application configuration.

### DIFF
--- a/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
+++ b/app/bundle/src/main/scala/org/bitcoins/bundle/gui/BundleGUI.scala
@@ -105,6 +105,10 @@ object BundleGUI extends WalletGUI with BitcoinSAppJFX3 {
         datadirParser.datadir,
         serverArgParser)(system)
 
+    if (appConfig.rpcPassword.nonEmpty) {
+      GlobalData.setPassword(appConfig.rpcPassword)
+    }
+
     validatePreferenceValues()
 
     val landingPane = new LandingPane(glassPane, serverArgParser)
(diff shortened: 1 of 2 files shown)


user411:
Running on top of this with password set or not set works for `bundle/run` for my dev env. It also works to make API calls through the UI. Adding headers to the websocket on the way through the proxy does not work (which has worked previously). Setting the websocket password at the UI does work (just like with Docker env).

This seems to be the unblocking fix I need.

user411 (approved):

PR #4003, opened 2022-01-22
Implement batching of database calls for our chain callback

### LINKED ISSUE #3967: IDB too slow
IBD on the master branch (commit 8857af2b66e8d493814d0bb7b817d8d46579709d) is too slow compared to what it used to be. It stalls randomly and eventually stops.

Attaching log file of current situation with header sync still incomplete after roughly an hour (usually takes 15 minutes for me). I checked that it works as expected on latest release build so I don't think it's a network issue.

[log.txt](https://github.com/bitcoin-s/bitcoin-s/files/7828828/latest_master2.txt)

Christewart:
fixes #3967 

When we merged #3946 I didn't realize that there was a performance impact for IBD. This happens because _for every block header we receive, we make a database call_. We receive block headers in batches of 2,000 headers, which means we have 2,000 database calls!

https://github.com/bitcoin-s/bitcoin-s/commit/d06b064b6bcd7e383dfe75b184454a71ad59c728#diff-dc390950db2105baa88147c5b8d335fa214be7415b77e2005b6a4f5d9e5dbfbcR34

This PR does a few things 

1. Reworks `ChainCallbacks.OnBlockHeaderConnected` to take a `Vector[(Int,BlockHeaderDb)]` rather than just a single `(Int,BlockHeaderDb)`.
2. Adds a few APIs like `ChainApi.getHeaders()` and `BlockHeaderDAO.findByHashes()`. This allows us to more efficiently query block headers from the database.

Here is what my visualvm `chaindb` thread looks like on da193037ed86303658b71bb13b9694053ca3fae8 . This was pinned at 100% usage before da193037ed86303658b71bb13b9694053ca3fae8

Screenshot from 2022-01-22 13-53-20

### DIFF
--- a/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
+++ b/app/server-test/src/test/scala/org/bitcoins/server/RoutesSpec.scala
@@ -251,13 +251,16 @@ class RoutesSpec extends AnyWordSpec with ScalatestRouteTest with MockFactory {
       (mockChainApi
         .getHeader(_: DoubleSha256DigestBE))
         .expects(blockHeader.hashBE)
-        .twice()
         .returning(Future.successful(Some(blockHeaderDb)))
 
+      (mockChainApi.getBestBlockHeader: () => Future[BlockHeaderDb])
+        .expects()
+        .returning(Future.successful(blockHeaderDb))
+
       (mockChainApi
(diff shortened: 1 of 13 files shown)


Christewart:
Shreyansh confirmed that this fixes the issue for him too

Screenshot from 2022-01-23 07-38-23

PR #4009, opened 2022-01-24
Use DatabaseDriver.fromString instead of if statements

benthecarman:


benthecarman:
nevermind, realized that this is needed because the string is something like `org.postgresql.Driver`

PR #3992, opened 2022-01-18
2022 01 18 issue 3969 Add `serialization_version` to `global_dlc_data`

Christewart:
fixes #3969

This PR adds a new column to `global_dlc_data` called `serialization_version`. 

When we fire up with dlc wallet with this change, this PR

1. Adds a new column to `global_dlc_data` called `serialization_version`.
2. Reads all rows inside of `global_dlc_data`
3. Queries for the `ContractInfo` associated with the row in `global_dlc_data`. 
4. Recomputes the `tempContractId` and `contractId` for DLCs
5. Updates them in the database

This is what the log looks like when the contractIds are migrated. This PR works because when I implemented #3854 I supported reading the old format (`Alpha`) and the new format (`Beta`), but we only support writing the new format (`Beta`). Now whenever we serialize something it will have an updated `tempContractId`/`contractId`. I then write the new `tempContractId`/`contractId` to the database. This migration also includes #4012 . All contractIds were broken in our current wallets because we did not include the funding output index in the hash computation.
 
The first few `WARN` logs are for #4001 , we cannot migrate those DLCs. See 4001 for why.

```
[info] 2022-01-26T18:46:07UTC WARN [DLCAppConfig] Corrupted dlcId=48c8fe360f09964920232452a315f7c621d9ebb602d19a22c2787ec6a7350958 state=Accepted, this is likely because of issue 4001 https://github.com/bitcoin-s/bitcoin-s/issues/4001 . This DLC will not have its contractId migrated to DLSerializationVersion.Beta
[info] 2022-01-26T18:46:07UTC WARN [DLCAppConfig] Corrupted dlcId=eafc75a2253898d60bc56900d3048aa774ffa49eb9bcfd0fe84d5740ca8c60c5 state=Accepted, this is likely because of issue 4001 https://github.com/bitcoin-s/bitcoin-s/issues/4001 . This DLC will not have its contractId migrated to DLSerializationVersion.Beta
[info] 2022-01-26T18:46:08UTC INFO [DLCAppConfig] Updating contractId for dlcId=3004f08a0fe2e2c928a78e33fd0f82ea431d6f576ac15004cd952a0802701cde old contractId=Some(b9abdebe933b1e27920cd9a111ef0c13aa9932d8489ca71fc70d912e0fcd27b6) new contractId=1bdb2ef6fd3b93cacf23de41d4e332330e670dc3104bb28e1fcda4b34f0ed402
[...]

PR #4012, opened 2022-01-25
Fix contractid computation

### LINKED ISSUE #4006: Contract id computation is wrong, it does not include the `funding_output_index`
https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#definition-of-contract_id

>Most messages use a contract_id to identify the contract. It's derived from the funding transaction and the offer by combining the funding_txid, the funding_output_index and the temporary_contract_id, using big-endian exclusive-OR (i.e. funding_output_index alters the last 2 bytes of funding_txid XOR temporary_contract_id).

https://github.com/bitcoin-s/bitcoin-s/blob/ac3bae403bb34152c1372211847026bfad67da1c/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala#L17

We should probably tie this into #3992 since that PR is already running migrations to update the contract ids.

H/T to @tibo-lg for finding this bug.

Christewart:
fixes #4006

H/T to @tibo-lg to finding this.

We were not including the funding output index in our contract id calculation, this fixes this.

This is going to require migrations in existing wallets, we should tie this into #3992 before we merge that PR.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/DLCUtilTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/protocol/dlc/DLCUtilTest.scala
@@ -1,0 +1,27 @@
+package org.bitcoins.core.protocol.dlc
+
+import org.bitcoins.core.protocol.dlc.compute.DLCUtil
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.crypto.Sha256Digest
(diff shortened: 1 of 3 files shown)


Christewart:
From @tibo-lg

Screenshot from 2022-01-26 05-28-34

So it sounds like we are good for merge :tada:

PR #4016, opened 2022-01-27
Fix bugs where we were building internally inconsistent SpendingInfoDb

Christewart:
Found when looking through #3909 

We cannot have the `SpendingInfoDb.outPoint.txIdBE` and `SpendingInfoDb.txid` be different. This means we are internally inconsistent with what we are storing inside of `SpendingInfoDb`

This PR doesn't truly fix the issue IMO, we need to remove the in memory `SpendingInfoDb.txid` parameter. We can still provide this method as we can just define it as 

```scala
def txid: DoubleSha256DigestBE = outPoint.txIdBE
```

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/api/wallet/db/SpendingInfoDb.scala
+++ b/core/src/main/scala/org/bitcoins/core/api/wallet/db/SpendingInfoDb.scala
@@ -151,6 +151,10 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** The TXID of the transaction this output was received in */
   def txid: DoubleSha256DigestBE
 
+  require(
+    txid == outPoint.txIdBE,
+    s"Cannot have difference outpoint txId and txIdd outpoint=${outPoint.txIdBE.hex} txid=${txid.hex}")
+
   /** TxId of the transaction that this output was spent by */
   def spendingTxIdOpt: Option[DoubleSha256DigestBE]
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bitcoins/core/api/wallet/db/SpendingInfoDb.scala
```diff
@@ -151,6 +151,10 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** The TXID of the transaction this output was received in */
   def txid: DoubleSha256DigestBE
 
+  require(
+    txid == outPoint.txIdBE,
+    s"Cannot have difference outpoint txId and txIdd outpoint=${outPoint.txIdBE.hex} txid=${txid.hex}")
```
benthecarman: > Cannot have difference outpoint txId and txIdd

Cannot have `different` outpoint txId and `txId`

core/src/main/scala/org/bitcoins/core/api/wallet/db/SpendingInfoDb.scala
```diff
@@ -151,6 +151,10 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** The TXID of the transaction this output was received in */
   def txid: DoubleSha256DigestBE
```
benthecarman, line 152: would be probably easier to just do this so the construction is impossible

`def txid: DoubleSha256DigestBE = outPoint.txIdBE`

Christewart: See #4017 , i didn't want to provide a default implementation because it will just get override by sub classes and make things more confusing IMO

PR #4019, opened 2022-01-27
Add invariant to make sure spendingTxId is different than the txid inside of `SpendingInfoDb`

Christewart:
Follow up on #4016 

This adds test cases and a new invariant to make sure the `spendingTxId` is not the same as `txid`


core/src/main/scala/org/bitcoins/core/api/wallet/db/SpendingInfoDb.scala
```diff
@@ -158,7 +158,13 @@ sealed trait SpendingInfoDb extends DbRowAutoInc[SpendingInfoDb] {
   /** TxId of the transaction that this output was spent by */
   def spendingTxIdOpt: Option[DoubleSha256DigestBE]
 
+  require(
```
Christewart, line 161: If the spendingTxId is not defined this is trivially true

benthecarman (approved):

PR #4022, opened 2022-01-27
Update flyway to latest version

benthecarman:
Was getting these logs so updated the flyway dep

```
Flyway upgrade recommended: PostgreSQL 13.4 is newer than this version of Flyway and support has not been tested. The latest supported version of PostgreSQL is 12.
```


benthecarman:
This doesn't seem to work

PR #4024, opened 2022-01-30
2022 01 30 optimize processblock

Christewart:
Some small optimzations in the hot path of `processBlockCachedUtxos`. This PR does not intend to change any functionality. Related to #2596

1. Remove allocations of `Some` inside of the parameters to `processTransactionImpl()`
2. Move the `.map()` calls on `cachedReceivedOptF` and `spentSpendingInfoDbsF` out of the `block.transactions.foldLeft()`. This means we only need to call these once rather than `n` times avoiding the `.map()` overhead.
3. Calculate `blockInputs` in advance rather than inside of the inner loop. This should reduce the amount of time to search inputs in the block.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
@@ -108,30 +108,32 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
     val spentSpendingInfoDbsF =
       spendingInfoDAO.findOutputsBeingSpent(block.transactions.toVector)
 
-    //we need to keep a cache of spentSpendingInfoDb
-    //for the case where we receive & then spend that
-    //same utxo in the same block
-    var cachedSpentOptF: Future[Option[Vector[SpendingInfoDb]]] = {
-      spentSpendingInfoDbsF
-        .map(Some(_)) //reduce allocations by creating Some here
-    }
-
+    val blockHashOpt = Some(block.blockHeader.hash.flip)
     val resultF: Future[Future[Wallet]] = for {
+      //map on these first so we don't have to call
+      //.map everytime we iterate through a tx
+      //which is costly
       receivedSpendingInfoDbsOpt <- cachedReceivedOptF
+      spentSpendingInfoDbs <- spentSpendingInfoDbsF
     } yield {
+      //we need to keep a cache of spentSpendingInfoDb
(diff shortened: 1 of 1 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -98,56 +98,75 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
...
+        Some(spentSpendingInfoDbs)
+      }
+      val blockInputs = block.transactions.flatMap(_.inputs)
+      val wallet: Future[Wallet] = {
+        block.transactions.foldLeft(Future.successful(this)) {
+          (acc, transaction) =>
+            for {
+              _ <- acc
+              processTxResult <- {
+                processTransactionImpl(
```
benthecarman: this should be `acc.processTransactionImpl` so we using the returned wallet

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
```diff
@@ -98,56 +98,75 @@ private[bitcoins] trait TransactionProcessing extends WalletLogger {
...
-                Future.successful(spentSpendingInfo ++ spentInSameBlock)
+                //add it to the cache
+                val newCachedSpentOpt = {
+                  cachedSpentOpt match {
+                    case Some(spentSpendingInfo) =>
+                      Some(spentSpendingInfo ++ spentInSameBlock)
+                    case None =>
+                      Some(spentInSameBlock)
+                  }
+                }
```
benthecarman, line 158: would this be more effienct?

```scala
val newSpent = cachedSpentOpt.getOrElse(Vector.empty)
val newCachedSpentOpt = cachedSpentOpt.map(_ ++ newSpent)
```

benthecarman (commented):
LGTM, just one question

wallet/src/main/scala/org/bitcoins/wallet/internal/TransactionProcessing.scala
Christewart: I'm not sure to be honest. I think it comes down to what is more efficiently implemented by the Scala language

1. the `match` keyword along with deconstructing `case` statements
2. `Option.map()`and `.getOrElse()`. 

I'm guessing it is a minuscule performance difference either way.

benthecarman (approved):

PR #4026, opened 2022-01-30
Refactor pending receives/spent states into one spot

Christewart:
Simple refactor. This PR also adds `ImmatureCoinbase` to the pending received states. I don't know what implication this has for things like coin selection. Will this select immature coinbase txs (consensus invalid txs) during coin selection?

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
@@ -346,12 +346,7 @@ abstract class Wallet
   }
 
   override def getUnconfirmedBalance(): Future[CurrencyUnit] = {
-    filterThenSum(utxo =>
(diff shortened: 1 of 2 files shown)


wallet/src/main/scala/org/bitcoins/wallet/Wallet.scala
```diff
@@ -348,7 +343,7 @@ abstract class Wallet
     } yield {
       val unconfirmed = allUnspent.filter { utxo =>
         HDAccount.isSameAccount(utxo.privKeyPath.path, account) &&
-        (utxo.state == PendingConfirmationsReceived || utxo.state == BroadcastReceived)
+        TxoState.pendingReceivedStates.contains(utxo.state)
```
Christewart, line 359: will add immature coinbases now?

benthecarman: yeah it will

PR #4029, opened 2022-01-31
[GUI] Fix 'First and last points must be endpoints' exception

### LINKED ISSUE #4027: Cannot create numeric offer with several points in desktop UI
**Describe the bug**
I don't seem to be able to create a numeric offer with more than 2 points (I did not re-test 2 point contracts for this bug report, maybe they don't work either). I am checking all points as endpoints, but seeing an error like attached.

**To Reproduce**
1. Build a numeric offer with 4 endpoints
2. Execute offer and see error popup.

**Expected behavior**
DLC offer is created

**Additional context**
EndpointsError
EndpointsInWebUI

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4027


Christewart (approved):
:man_facepalming: . Oops.

Issue #4032, opened 2022-02-01
Received incorrect outcomes [] for numeric contract outcome


user411:
**Describe the bug**
Doing a numeric DLC between two web UIs, one wallet ended up in a state with the Outcome of the contract listed as half the value of the signed outcome on the Oracle Explorer.

The announcement used was https://test.oracle.suredbits.com/announcement/8d328f8962853255bec0182e3f452a88079e86973b3e06781ba78cfab33744b4

The contract initiator and claimer ended up in a good state with 38208 reported as the outcome, but the non-initiating, non-claiming participant ended up in a state that reports half the value of the contract in the DLC structure returned to the UI. In this case the reported values are

[0, 0, 1, 0, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 0, 0] (supposed to) => 38208
actual 32 + 128 + 512 + 2048 + 16384 = 19104

![BadOutcomeJSON](https://user-images.githubusercontent.com/22351459/151896669-48f53b1e-2049-4ef8-a85f-dd2e1ba32d6b.png)
Non-Initiator:
![DLCNonInitiatorBadOutcome](https://user-images.githubusercontent.com/22351459/151896671-6944c8e7-1fd6-4ceb-b298-16a15df8920e.png)
Initiator:
![DLCInitiatorGoodOutcome](https://user-images.githubusercontent.com/22351459/151896887-786a3b0a-5b74-4001-9cd6-d087538e7e63.png)

**To Reproduce**
1. I don't know if this is reproducible. It is currently in my wallet.

**Expected behavior**
Report correct outcomes[] from backend.

**Platform (please complete the following information):**
 - macos m1 web UI over bundle/run

Christewart:
The difference is the wallet has 17 digits, and the Suredbits Oracle Explorer has 18 digits

![Screenshot from 2022-02-01 13-15-46](https://user-images.githubusercontent.com/3514957/152036015-bde48bc5-e3c2-45fd-99cb-247a9c3ba753.png)
![Screenshot from 2022-02-01 13-05-03](https://user-images.githubusercontent.com/3514957/152036017-868c0202-7084-45ce-851e-188d2f49a316.png)

Christewart:
Still looking into 

1. How did we drop a digit on the floor
2. Do we have invariants s.t. when we input oracle signatures with the incorrect number of digits, we get an exception

user411:
[...]

PR #4037, opened 2022-02-02
2022 02 01 Add invariants to make sure we have correct number of oracle sigs

Christewart:
Add invariants on #4032 to make sure this can't happen again from a bad number of oracle signatures given.

This PR explicitly ignores some MultiOracle DLC test cases. I will be filing an issue to address these. Since multi oracle is not exposed to any end users, i think this is ok for now as a stop gap solution. Likely we will need to refactor the multi oracle test suites to fix the bugs in #4032 that are present there.

These are what the test failures look like for the multi oracle numeric tests if I do not `ignore` them.

Screenshot from 2022-02-02 11-22-10

### DIFF
--- a/dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleNumericExecutionTest.scala
+++ b/dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleNumericExecutionTest.scala
@@ -113,8 +113,6 @@ class DLCMultiOracleNumericExecutionTest
       contractId <- getContractId(wallets._1.wallet)
       status <- getDLCStatus(wallets._1.wallet)
       (sigs, _) = getSigs(status.contractInfo)
-      /*      _ = require(sigs.length == numDigits,
-                  s"sigs.length=${sigs.length} numDigits=$numDigits")*/
       func = (wallet: DLCWallet) => wallet.executeDLC(contractId, sigs)
 
       _ = logger.info(s"Beginning dlcExecutionTest")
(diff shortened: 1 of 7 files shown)


benthecarman (approved):
ACK on CI

I'll try to take a stab at getting it working for multi oracle in the next few days

dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleNumericExecutionTest.scala
```diff
@@ -185,7 +146,31 @@ class DLCMultiOracleNumericExecutionTest
...
-  it must "execute as the recipient" in { wallets =>
+  /** Builds oracle attestments for the given oracle outcome */
+  private def buildAttestments(
+      oracleOutcomes: NumericOracleOutcome): Vector[OracleAttestmentV0TLV] = {
+    privateKeys.zip(kValues).flatMap { case (priv, kValues) =>
+      val outcomeOpt = oracleOutcomes.oraclesAndOutcomes.find(
+        _._1.publicKey == priv.schnorrPublicKey)
+
+      outcomeOpt.map { case (oracleInfo, outcome) =>
+        val sigs = outcome.digits.zip(kValues).map { case (num, kValue) =>
```
nkohen, line 157: if you want to have the invariants then you'll need to extend `outcome.digits` with right padding (say, `0`s, since it doesn't matter what you put there) up to the number of digits specified in the contract info (probably a parameter this `def` will need

nkohen:
Looks like there are still other tests failing, I think the solution in all cases is to add extra sigs of `0`s to attestments generated in tests since we were previously having "just the right number" of signatures generated rather than all of them (as mentioned in my comment).

PR #4042, opened 2022-02-04
2022 02 03 issue 4032

Christewart:
Addresses parts of #4032 

This PR adds a check for `ContractInfo` DLCs to make sure we have the entire set of oracle signatures.

This PR also refactors all places we match oracle announcements against oracle signatures into `DLCUtil`. These are now helper methods that can be used from anywhere in the codebase.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
@@ -206,15 +206,11 @@ object DLCUtil {
       case single: SingleContractInfo =>
         checkSingleContractInfoOracleSigs(single, oracleSigs)
       case disjoint: DisjointUnionContractInfo =>
-        val contractHasMatchingSigs: Vector[Boolean] = {
-          disjoint.contracts.map { single: SingleContractInfo =>
-            checkSingleContractInfoOracleSigs(single, oracleSigs)
-          }
-        }
-
         //at least one disjoint union contract
         //has to have matching signatures
-        contractHasMatchingSigs.exists(_ == true)
+        disjoint.contracts.exists { single: SingleContractInfo =>
+          checkSingleContractInfoOracleSigs(single, oracleSigs)
+        }
     }
   }
 
@@ -276,11 +272,12 @@ object DLCUtil {
             val oracleSig =
               OracleSignatures(SingleOracleInfo(ann), attestment.sigs)
             val isMatch = matchOracleSignaturesForAnnouncements(ann, oracleSig)
-            if (isMatch.isDefined) {
-              acc.:+(isMatch.get)
-            } else {
-              //don't add it, skip it
-              acc
+            isMatch match {
+              case Some(matchedSig) =>
+                acc.:+(matchedSig)
+              case None =>
+                //don't add it, skip it
+                acc
             }
           }.toVector
           r
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
@@ -8,9 +8,14 @@ import org.bitcoins.core.protocol.dlc.models.DLCMessage.{
   DLCAcceptWithoutSigs,
   DLCOffer
 }
-import org.bitcoins.core.protocol.dlc.models.{ContractInfo, OracleOutcome}
+import org.bitcoins.core.protocol.dlc.models._
 import org.bitcoins.core.protocol.script.P2WSHWitnessV0
+import org.bitcoins.core.protocol.tlv.{
+  OracleAnnouncementTLV,
+  OracleAttestmentTLV
+}
 import org.bitcoins.core.protocol.transaction.{Transaction, WitnessTransaction}
+import org.bitcoins.core.util.sorted.OrderedAnnouncements
 import org.bitcoins.crypto._
 import scodec.bits.ByteVector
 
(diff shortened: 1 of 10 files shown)


benthecarman (approved):
Much better way than our previous 2 attempts

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
```diff
@@ -187,4 +192,101 @@ object DLCUtil {
...
+      case disjoint: DisjointUnionContractInfo =>
+        val contractHasMatchingSigs: Vector[Boolean] = {
+          disjoint.contracts.map { single: SingleContractInfo =>
+            checkSingleContractInfoOracleSigs(single, oracleSigs)
+          }
+        }
+
+        //at least one disjoint union contract
+        //has to have matching signatures
+        contractHasMatchingSigs.exists(_ == true)
```
nkohen: `.map` followed by `.exists` should be replaced by just a single `.exists` call to avoid extraneous computation once a match has been found

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
```diff
@@ -187,4 +192,101 @@ object DLCUtil {
...
+          val r: Vector[OracleSignatures] = announcements.flatMap { ann =>
+            val oracleSig =
+              OracleSignatures(SingleOracleInfo(ann), attestment.sigs)
+            val isMatch = matchOracleSignaturesForAnnouncements(ann, oracleSig)
+            if (isMatch.isDefined) {
+              acc.:+(isMatch.get)
+            } else {
+              //don't add it, skip it
+              acc
+            }
```
nkohen: nit: would be more idiomatic to match instead of using `val` with `if`/`else`

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
```diff
@@ -187,4 +192,101 @@ object DLCUtil {
...
+      Vector(signature)
+    )
+  }
+
+  /** Builds a set of oracle signatures from given announcements
+    * and attestations. This method discards attestments
+    * that do not have a matching announcement. Those attestments
+    * are not included in the returned set of [[OracleSignatures]]
+    */
+  def buildOracleSignatures(
```
nkohen, line 264: I'm a little confused as to why we need this?

nkohen: Specifically it seems like it is working too hard, I can't think of a situation where we would pass around a pool of attestments not associated with one another (i.e. all attestments from the same oracle would be together wouldn't they?)

core/src/main/scala/org/bitcoins/core/protocol/dlc/execution/DLCExecutor.scala
```diff
@@ -136,6 +136,9 @@ object DLCExecutor {
       fundingTx: Transaction,
       fundOutputIndex: Int
   ): ExecutedDLCOutcome = {
+    require(
+      DLCUtil.checkOracleSignaturesAgainstContract(contractInfo, oracleSigs),
+      s"Incorrect oracle signatures and contract combination")
```
nkohen, line 141: I think this is fine (the alternative being checking when we receive the attestments, i.e. in the wallet) but just wanted to make note again that this reduces the capabilities of the library meaning that if anyone ever ends up in a weird state where they have enough of the oracle attestments to claim their funds but don't have all of them they'll need a custom build in order to claim their funds

dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleExactNumericExecutionTest.scala
```diff
@@ -242,4 +203,32 @@ class DLCMultiOracleExactNumericExecutionTest extends BitcoinSDualWalletTest {
...
+
+  /** Builds an attestment for the given numeric oracle outcome */
+  private def buildAttestments(
+      outcome: NumericOracleOutcome): Vector[OracleAttestmentTLV] = {
+    privateKeys.zip(kValues).flatMap { case (priv, kValues) =>
+      val outcomeOpt =
+        outcome.oraclesAndOutcomes.find(_._1.publicKey == priv.schnorrPublicKey)
+
+      outcomeOpt.map { case (oracleInfo, outcome) =>
+        val neededPadding = kValues.length - outcome.digits.length
```
nkohen: should use `numDigits` directly instead of `kValues.length`

dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleExactNumericExecutionTest.scala
```diff
@@ -242,4 +203,32 @@ class DLCMultiOracleExactNumericExecutionTest extends BitcoinSDualWalletTest {
...
+        val digitsPadded = outcome.digits ++ Vector.fill(neededPadding)(0)
```
nkohen, line 216: I think there is a `padRight` function for this

dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/DLCMultiOracleNumericExecutionTest.scala
```diff
@@ -249,4 +212,32 @@ class DLCMultiOracleNumericExecutionTest
         aggregateSignature == statusB.oracleSig
     }
   }
+
+  /** Builds an attestment for the given numeric oracle outcome */
+  private def buildAttestments(
```
nkohen, line 217: probably would be nice to have this in one place instead of multiple (also my above comments apply here)

testkit-core/src/main/scala/org/bitcoins/testkitcore/dlc/DLCTest.scala
```diff
@@ -880,8 +880,11 @@ trait DLCTest {
         .get
         ._2
 
+      val neededPadding =
+        singleOracleInfo.announcement.eventTLV.eventDescriptor.noncesNeeded - digitsToSign.digits.length
+      val paddedDigits = digitsToSign.digits ++ Vector.fill(neededPadding)(0)
```
nkohen, line 885: `padRight`

nkohen (commented):
ACK on response and CI

testkit-core/src/main/scala/org/bitcoins/testkitcore/dlc/DLCTest.scala
Christewart: So this doesn't seem to be available on a `Vector`.  I tried using `.padTo` but it didn't seem to work.

core/src/main/scala/org/bitcoins/core/protocol/dlc/compute/DLCUtil.scala
Christewart: Idk, user copies the attestments wrong when highlighting them? 

Seems to me this

1. Consolidates matching of announcements and attestments to one place
2. Validates all user input

core/src/main/scala/org/bitcoins/core/protocol/dlc/execution/DLCExecutor.scala
benthecarman: I don't think this is a huge concern. The `DLCAttestmentTLV` should be the way people deal with signatures, not the actual raw `SchnorrDigitalSignature`s

Issue #4050, opened 2022-02-06
Error no oracles used


Christewart:
When attempting to login to my umbrel wallet this morning i get this error when i put in my password. I cannot proceed any further. This was on 2066447c

Related issues are #3686 #3826 . The notable difference is that this issue was with a `bitcoind` backend, not a `neutrino` backend.

![Screenshot from 2022-02-06 07-34-35](https://user-images.githubusercontent.com/3514957/152683455-e7d10114-3c6d-48a8-acc9-daf48626137b.png)

Stack trace 

```
2022-02-06T13:21:46UTC INFO [DLCWallet$DLCWalletImpl] Finished processing of block=00000000000000000000941e92474ee4430498bf1ff0dea90f4d5d2abd3f3671. It took 19710ms
2022-02-06T13:32:49UTC INFO [Server] Unhandled error in server:
java.lang.IllegalArgumentException: requirement failed: Error, no oracles used
        at scala.Predef$.require(Predef.scala:337)
        at org.bitcoins.dlc.wallet.util.DLCStatusBuilder$.getOracleOutcomeAndSigs(DLCStatusBuilder.scala:221)
        at org.bitcoins.dlc.wallet.util.DLCStatusBuilder$.buildClosedDLCStatus(DLCStatusBuilder.scala:159)
        at org.bitcoins.dlc.wallet.DLCWallet.$anonfun$buildDLCStatus$5(DLCWallet.scala:1670)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
        at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:100)
        at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:49)
        at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:295)
[...]

PR #4054, opened 2022-02-06
2022 02 06 remoteclaimed refactor

Christewart:
This PR does not intend to change functionality.

Related to #4050 

This PR intends to refactor the `RemoteClaimed` flow and it make it easier to read. This asserts invariants about the flow that should hold true, and gives more useful error mesages for cases when these are not true.


benthecarman (approved):

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/DLCTransactionProcessing.scala
```diff
@@ -87,17 +88,19 @@ private[bitcoins] trait DLCTransactionProcessing extends TransactionProcessing {
         val updatedOpt = Some(dlcDb.copy(state = DLCState.Claimed))
         Future.successful(updatedOpt)
       } else {
-        val withState = dlcDb.updateState(DLCState.RemoteClaimed)
         if (dlcDb.state != DLCState.RemoteClaimed) {
+          val withState = dlcDb.updateState(DLCState.RemoteClaimed)
           for {
-            // update so we can calculate correct DLCStatus
-            _ <- dlcDAO.update(withState)
```
Christewart, line 94: Don't do the database update here. This will make the database update for the state atomic with the rest of the writing to the database.

Issue #4049, opened 2022-02-05
getannouncement call fails in fresh wallet with one announcement on umbrel


alexwaltz:
While trying to log in to Krystal Bull on umbrel I got 
_"Error sending message to OracleServer Request failed: requirement failed:
Not enough nonces for this event descriptor"_
I managed to create 1 contract before this happened.

<img width="695" alt="Screenshot 2022-02-05 at 21 27 52" src="https://user-images.githubusercontent.com/92048497/152662002-dccd9937-3f37-4ec7-98ce-9401aa54aada.png">

After clicking OK the round circle thing keeps spinning.
Even though I created an announcement in the past it does not appear in the history.
<img width="1132" alt="Screenshot 2022-02-05 at 22 40 16" src="https://user-images.githubusercontent.com/92048497/152662057-55eedcee-3fe8-40ca-9b73-e7011ef45b1d.png">

**Network log**
<img width="1275" alt="Screenshot 2022-02-05 at 22 56 48" src="https://user-images.githubusercontent.com/92048497/152662214-f71fc0e3-8de8-48e2-95ba-033c79aaef45.png">

The call that is 500ing is:
**Request payload**
`{"method":"getannouncement","params":["resrser"]}`

**Edit:** added log files
- [oracle-server-ui-proxy.log](https://github.com/bitcoin-s/bitcoin-s/files/8010846/oracle-server-ui-proxy.log)
- [bitcoin-s.log](https://github.com/bitcoin-s/bitcoin-s/files/8010847/bitcoin-s.log)
- [oraclesqlite.txt](https://github.com/bitcoin-s/bitcoin-s/files/8011437/oraclesqlite.txt) (changed file type to .txt as github did not allow me to upload a .sqlite file for some reason)

Christewart:
It appears that two sets of nonces got created for the same event

![Screenshot from 2022-02-07 07-20-17](https://user-images.githubusercontent.com/3514957/152795694-78aa9643-0039-4788-93b5-f2f80f3c6c25.png)

Christewart:
This is the event descriptor

>fdd80a0b0002000131000000010008

which parses to

>eventTLV=UnsignedDigitDecompositionEventDescriptor(UInt16Impl(2),UInt16Impl(8),1,Int32Impl(1)) noncesNeeded=8

So we end up with two sets of nonces for one event somehow.

Christewart:
[...]

PR #4057, opened 2022-02-07
Limit number of max-connections on oracleServer to 1

Christewart:
Tries to prevent #4049 from happening again

This PR makes only 1 http connection available to the oracleServer. This is intended to reduce paralellism so we don't have race conditions on the pi as documented here

https://github.com/bitcoin-s/bitcoin-s/issues/4049#issuecomment-1031496481

here is the documentation on this setting from akka

https://doc.akka.io/docs/akka-http/current/server-side/low-level-api.html#controlling-server-parallelism

### DIFF
--- a/app/oracle-server/src/main/resources/reference.conf
+++ b/app/oracle-server/src/main/resources/reference.conf
@@ -15,4 +15,6 @@ akka {
   http.server.request-timeout = 10s
   http.server.parsing.max-content-length = 8m
   http.client.parsing.max-content-length = 8m
+
+  http.server.max-connections = 1
 }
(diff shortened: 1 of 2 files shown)


user411:
So if I opened a second tab aimed at the Krystal Bull UI, it would not work?

Christewart:
> So if I opened a second tab aimed at the Krystal Bull UI, it would not work?

It works when i test it. My understanding is the underlying oracle proxy server will be shared across the two tab instances.

Christewart:
Superseded by https://github.com/bitcoin-s/bitcoin-s-ts/pull/81

PR #4061, opened 2022-02-08
Change `DLCUtil.buildOracleSignatures`

rorp:
Introduced in #4042 , old implementation leads to memory leaks on CI.


Christewart (approved):

benthecarman:
Shouldn't you remove the old impl?

Issue #4030, opened 2022-01-31
Add websocket events for Tor DLC negotiation


user411:
**Describe the bug**
We need websocket events to show the user the state of setting up Tor exchanged DLC contracts to keep them in the loop on longer running interactions. For numeric contracts especially there is longer running interaction between creating signatures and accepting and passing data back for signing. Right now there is no indication to the user that the process is ongoing.

**To Reproduce**
Steps to reproduce the behavior
1. Create a numeric offer
2. Send it to your buddy over tor
3. Wonder what state things are in until they magically click into place.

**Expected behavior**
Some indication at the UI that the tor exchange process is ongoing.

PR #4064, opened 2022-02-08
2022 02 08  dlc accept refactor

Christewart:
Another round of refactors to `DLCWallet.createNewDLCAccept()`. 

This is related to #4048 . This PR does not intend to change any functionality. I need this work to continue on #4030 .

This PR adds a new object `DLCAcceptUtil` where utility methods for building accept messages are moved too.

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -6,8 +6,6 @@ import org.bitcoins.core.api.dlc.wallet.db.DLCDb
 import org.bitcoins.core.api.feeprovider.FeeRateApi
 import org.bitcoins.core.api.node.NodeApi
 import org.bitcoins.core.api.wallet.db._
-import org.bitcoins.core.config.BitcoinNetwork
-import org.bitcoins.core.crypto.ExtPublicKey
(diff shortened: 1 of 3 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -631,105 +568,64 @@ abstract class DLCWallet
...
-                                         BIP32Node(dlc.keyIndex,
-                                                   hardened = false)))
+    val accountF = getDefaultAccountForType(AddressType.SegWit)
 
-      privKeyPath = HDPath.fromString(bip32Path.toString)
-      fundingPrivKey =
-        keyManager.toSign(privKeyPath)
-
-      dlcPubKeys = calcDLCPubKeys(account.xpub, dlc.changeIndex, dlc.keyIndex)
+    val txBuilderAndSpendingInfosF: Future[(
```
rorp, line 577: I don't think we really want to execute these `Future`s in parallel.

rorp (changes requested):

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
Christewart: What would you like it synchronized with?

PR #4065, opened 2022-02-09
DLC Mutual Close

benthecarman:
impl: https://github.com/discreetlogcontracts/dlcspecs/pull/170

This doesn't have the extraInputs implemented on the create mutual close side, that could be done later

still waiting on a number for the type number of `DLCMutualCloseTLV`

### DIFF
--- a/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
+++ b/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
@@ -1397,7 +1397,8 @@ abstract class DLCWallet
     dlcDb.state match {
       case DLCState.Broadcasted | DLCState.Signed => dlcDb
       case state @ (DLCState.Offered | DLCState.Confirmed | DLCState.Accepted |
-          DLCState.Claimed | DLCState.RemoteClaimed | DLCState.Refunded) =>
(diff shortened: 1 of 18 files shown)


dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -1594,6 +1595,90 @@ abstract class DLCWallet
...
+          signer <- dlcDataManagement
+            .signerFromDb(dlcId = dlcId,
+                          transactionDAO = transactionDAO,
+                          fundingUtxoScriptSigParams = scriptSigParams,
+                          keyManager = keyManager)
+            .map(_.get)
+
+          closeTx = signer.completeMutualClose(closeTLV)
+          _ = logger.info(
+            s"Created DLC refund transaction ${closeTx.txIdBE.hex} for contract ${contractId.toHex}")
```
matthewjablack: I'm guessing this should be `close` instead of `refund`

matthewjablack (commented):
Thanks for putting this together @benthecarman 🔥 

One small nit 👇

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
benthecarman: thanks fixed

PR #4068, opened 2022-02-10
Implement `BitcoindRpcClient.epochSecondToBlockHeight`

### LINKED ISSUE #4044: Bitcoind backend rescans do not honor wallet creation time
When we use neutrino as a backend, we have an optimization in place to rescan since the wallet's creation time.

This functionality does not work with a bitcoind backend. We need to implement it.

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4044


Christewart:
Tested on my wallet this morning with a bitcoind backend. Works for me.

PR #4071, opened 2022-02-10
Validate announcement signatures on create/accept offer

rorp:


Christewart (commented):
Can we add a few test cases to make sure we don't introduce this bug again?

benthecarman (approved):

PR #4073, opened 2022-02-11
Create upsertAllAction

benthecarman:


db-commons/src/main/scala/org/bitcoins/db/CRUDAction.scala
```diff
@@ -94,6 +94,26 @@ abstract class CRUDAction[T, PrimaryKeyType](implicit
...
+  }
+
+  /** Upsert all of the given ts.
+    * Returns all ts that were inserted or updated
+    */
+  def upsertAllAction(
+      ts: Vector[T]): DBIOAction[Vector[T], NoStream, Effect.Write] = {
+    val upsertActions: Vector[DBIOAction[T, NoStream, Effect.Write]] = {
+      ts.map { t =>
+        table.insertOrUpdate(t).flatMap { _ =>
```
Christewart: I think we should probably return the thing we did from the database like this:

https://github.com/Christewart/bitcoin-s-core/commit/841fd3b21f94d2efd20eb180ded43bcb15c5ba0c#diff-a6b895857ae8ac824badaded9e2cb25f28accd6ae02761c027512b1f6268a52bR97

Christewart (changes requested):

db-commons/src/main/scala/org/bitcoins/db/CRUDAction.scala
Christewart: Probably add this scaladoc link to the top of the method too

https://scala-slick.org/doc/3.3.3/queries.html#upserting

Christewart: Never mind, this doesn't work on sqlite :cry: 

The implementation you have may lead to bugs in the future for rows that have auto incremented ids. You will have to perform another database read to get the actual database id.

benthecarman: That's what I added

Christewart (approved):

PR #4081, opened 2022-02-12
Implement connection timeout for `dlcNode`

### LINKED ISSUE #4080: Request timeout occurs before connection timeout occurs when trying to conenct with tor peer
This leads to a misleading message on the frontend about the request timing out. IIUC, the tor connection could actually get established after teh request timeouts. WE probably need to set the tor connection timeout < the request timeout on the server.

Screenshot from 2022-02-11 16-11-20

Screenshot from 2022-02-11 16-11-35

Christewart:
fixes #4080 

If we can't establish a connection with our peers node after 45 seconds, fail.

This timeout is set to less than the request timeout so that we don't have background processes happening after the request times out. The request timeout is set to 60 seconds here: 

https://github.com/bitcoin-s/bitcoin-s/blob/08941206fc0c238aa68046172f9ecb060a75d2cb/db-commons/src/main/resources/reference.conf#L226

This is what the error looks like on the UI. 

Screenshot from 2022-02-12 06-39-08

### DIFF
--- a/dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCClient.scala
+++ b/dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCClient.scala
@@ -10,6 +10,7 @@ import org.bitcoins.tor.{Socks5Connection, Socks5ProxyParams}
 
 import java.io.IOException
 import java.net.InetSocketAddress
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{Future, Promise}
 
 class DLCClient(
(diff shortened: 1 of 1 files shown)


user411:
Might consider resolving the Some in the error message so it reads "Connection timeout of 45 seconds expired"

Christewart:
> Might consider resolving the Some in the error message so it reads "Connection timeout of 45 seconds expired"

I think this will probably take more energy than its worth. This exception is coming from deep within the http library we use. I guess i could catch the exception, parse the error message, and detect if it contains `Some` inside the String, but I think my time would be spent better else where for now

user411 (approved):

Issue #3868, opened 2021-12-02
Wallet not accruately tracking utxos


Christewart:
My wallet ended up not tracking states correctly. I'm unsure of why this is occurring at the moment. I believe this corruption was happening slowly over time, but i need to look into the specific utxos that are different from the two files. One reason the utxos sets are different is that the `wallet-utxos.txt` contained reserved utxos that are not included in the [`getutxos`](https://github.com/bitcoin-s/bitcoin-s/blob/31e83245228d207f9c5a3f2d230c9397c6fe1e34/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala#L41) response. I'm attaching two files to this issue, the first file `wallet-utxos.txt` is the result of `getutxos` before a rescan.

The second file `wallet-utxos-2.txt` is the rest of `getutxos` after a rescan.

Balance of the wallet before rescan

```
 ./bitcoin-s-cli getbalances
{
  "confirmed": 0.04385147,
  "unconfirmed": 1.4195E-4,
  "reserved": 0.0859621,
  "total": 0.12995552
}
```

Balance after rescan

```
 ./bitcoin-s-cli getbalances
{
  "confirmed": 1.33076847,
  "unconfirmed": 0,
  "reserved": 0,
  "total": 1.33076847
}
```
The contents of `wallet-utxos.txt`

```
[
  {
    "outpoint": {
      "txid": "c4698bf9afad0714f3cd8704a154985e3193cbfcc8cdaf754a806cacabed7fd8",
      "vout": 1
    },
    "value": 1179
  },
  {
    "outpoint": {
      "txid": "8c137c7f5000e5ce0035f9fb868acf7cc5706891a9da17d0e44e3033dd15c0aa",
      "vout": 91
    },
    "value": 547
  },
  {
    "outpoint": {
      "txid": "8c137c7f5000e5ce0035f9fb868acf7cc5706891a9da17d0e44e3033dd15c0aa",
      "vout": 105
    },
    "value": 547
  },
  {
    "outpoint": {
      "txid": "9974b0465681e3f77a02da23e8ef0a2b0703faf5ca248e5389b9d85afd41ff1f",
      "vout": 0
    },
    "value": 36435
  },
  {
    "outpoint": {
      "txid": "3a12f1154b8394ecc72b2cb7c9ad7d8a4110d7051cbc48955c22914c8d23e3c0",
      "vout": 0
    },
    "value": 3035
  },
  {
[...]

Issue #3864, opened 2021-12-01
Failed to broadcast DLC funding tx


Christewart:
This occurred on 169222a3 with a neutrino backend

I just tried to broadcast a DLC funding tx with a neutrino node and it looks like my peer disconnected me when i attempted to broadcast it via the bitcoin p2p network. It appears that the transaction COULD be broadcast via `bitcoin-cli sendrawtransaction`

> bitcoin-cli sendrawtransaction 02000000000103e0d59865f880cb2d2faad8f4172f3bbd4712fe8f8fafa41f920a5fa928f4a3b50000000000fdffffffddecd8a666a89240745bff559c8b77ef46c1f3577e788bb45047eb1b047e7d380000000000fdffffff55a6653fe15b051a435c7c9e6e16b191692c4341d11be7cd9f85ef2c3e4c84310000000000fdffffff03adcf410000000000160014fe6fcaa038fa8f875a1249c756205e01f715e2a69a8c010000000000220020ad8dfb73175ac9750daff0343ac6cc6636885dbb62edb85735426a26a64930d0e5bb000000000000160014046d4e182705ee59ac772360abe0182cb90020c70247304402203bcab239c1c62d3a6ff00a8fb699cc01d21bef8ef21ddbde7b95c2b2ea5a98c702203025751123ef767fa84f1dfae8ba1d6c3b0cd63a7195be22df1d3932d1ef45720121024d97b3648139442063321cde6f70aef14ff40332ecbfd5caaf261f5c3fad531f0247304402202ebb810524c8626bc47cd09b1d6528b991d7fd1770e9a7ac19273da405b057e50220339b613f1b9c15a10d638fc8b69f9316818a54233409256263049fb649aba340012103c3485d658640f7d4cbff5b2f415863538e8d306996e6901c0c61ae57f0cfa5240247304402204af9b1ddcbeed7066ca495e8ba63182693b8bf8140d44b160d88b636034c3cc302205d5195dec992242a4f6079601f3926e44839e81732639fa0447993471db3e82f012103c17ab7406d77c6fc34eb7de32fc281d1fc0a8687d91940925e8cabcef2c93a9200000000                         
ed65a83747baa4c9028e4b064f1566201499df8c99d1ee4af265b3903ef1d5bd

This is the tx in question

https://mempool.space/tx/ed65a83747baa4c9028e4b064f1566201499df8c99d1ee4af265b3903ef1d5bd

### Logs

[...]

Issue #3842, opened 2021-11-22
Unconfirmed balance not being calculated correctly on my wallet


Christewart:
**Describe the bug**
When opening up my DLC wallet on 90e01d7fc6ffb929ebc2a2c877e8942f0d490b63 I see this 

![Screenshot from 2021-11-22 07-40-46](https://user-images.githubusercontent.com/3514957/142871657-bb8bf6c0-098c-4aad-9662-590ec74db28c.png)

Using `bitcoin-s-cli getutxos` i can find the utxo on a block explorer, and it has a ton of confirmations on it

https://blockstream.info/tx/0d1e8c2572fe5483b53fb711243c4d11893b58b64af759f439c22c5fb34c0f05

![Screenshot from 2021-11-22 07-41-44](https://user-images.githubusercontent.com/3514957/142871762-313a9dad-bd22-425d-b647-b5951f8ef56e.png)

Christewart:
related to #3868 likely

Christewart:
Likely same bug as #4168 so closing

Issue #3812, opened 2021-11-11
Having issue rebroadcasting DLC funding tx


user411:
**Describe the bug**
I have a main wallet and a secondary wallet on the second computer that were interacting. The main wallet started acting weird and not sending transactions. The main wallet offered a DLC to the secondary wallet and the secondary wallet shows the DLC they entered into in Broadcasted state and the main wallet shows Offered. When I try to rebroadcast from the secondary wallet I see logs like

```
[info]   Response: Complete(HttpResponse(200 OK,List(),HttpEntity.Strict(application/json,1042 bytes total),HttpProtocol(HTTP/1.1)))
[info] 2021-11-11T14:59:16UTC INFO [DLCWallet$DLCWalletImpl] Created funding transaction 7c7ab92f0bd68b9858a935af509643670d1f7535ddc436cf066462fbddfd15a4 for contract d2f2d232f535443166cfb296dd3a8f0aa08cc3746ddb89a6fc8e40a538f1efdc
[info] 2021-11-11T14:59:16UTC INFO [DLCWallet$DLCWalletImpl] Broadcasting funding transaction 7c7ab92f0bd68b9858a935af509643670d1f7535ddc436cf066462fbddfd15a4 for contract d2f2d232f535443166cfb296dd3a8f0aa08cc3746ddb89a6fc8e40a538f1efdc
[info] 2021-11-11T14:59:16UTC INFO [NeutrinoNode] Sending out tx message for tx=Vector(7c7ab92f0bd68b9858a935af509643670d1f7535ddc436cf066462fbddfd15a4)
[info] 2021-11-11T14:59:16UTC INFO [DLCWallet$DLCWalletImpl] Processing tx 7c7ab92f0bd68b9858a935af509643670d1f7535ddc436cf066462fbddfd15a4 for 1 DLC(s)
[info] 2021-11-11T14:59:16UTC INFO [ActorSystemImpl] http-rpc-server: Response for
[info]   Request : HttpRequest(HttpMethod(POST),http://localhost:9999/,List(Timeout-Access: <function1>, Accept-Encoding: gzip, deflate, Cache-Control: no-cache, Pragma: no-cache, User-Agent: Java/11.0.11, Host, Accept: text/html, image/gif, image/jpeg, */*;q=0.2, */*;q=0.2, Connection: keep-alive),HttpEntity.Strict(application/json,156 bytes total),HttpProtocol(HTTP/1.1))
[...]

PR #4093, opened 2022-02-14
Segregate updating received utxos and spent utxos

Christewart:
My attempt at fixing #3842 . I wouldn't be surprised if this is also the cause of #3868 #3864 #3812 et al

This PR also enables some reorg capability in the wallet #2179

The meat and potates of this PR is splitting `updateTxoWithConfs` into two methods

1. `updateUtxoReceiveConfirmedStates`
2. `updateUtxoSpentConfirmedStates` 

As we need to do different things depending on if the utxo is received or confirmed.

I still would like to add more unit tests to this PR to verify the behavior is correct.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
@@ -192,6 +192,8 @@ private[wallet] trait UtxoHandling extends WalletLogger {
 
   /** Updates all the given SpendingInfoDbs to the correct state
     * based on how many confirmations they have received
+    * @param spendingInfoDbs the utxos we need to update
+    * @param fn the function used to transition the [[TxoState]] given a utxo and number of confirmations
     */
   private def updateUtxoStates(
       spendingInfoDbs: Vector[SpendingInfoDb],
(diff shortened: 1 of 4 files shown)


wallet/src/main/scala/org/bitcoins/wallet/internal/UtxoHandling.scala
```diff
@@ -379,7 +417,10 @@ private[wallet] trait UtxoHandling extends WalletLogger {
     for {
       infos <- spendingInfoDAO.findAllPendingConfirmation
       _ = logger.debug(s"Updating states of ${infos.size} pending utxos...")
-      updatedInfos <- updateUtxoConfirmedStates(infos)
-    } yield updatedInfos
+      receivedUtxos = infos.filter(_.state.isInstanceOf[ReceivedState])
```
Christewart, line 420: What to do about `TxoState.DoesNotExist`? It is not a `ReceivedState` or `SpentState`, so it gets filtered out here.

benthecarman: We honestly could remove `DoesNotExist`, I don't think we have a use for it

PR #4101, opened 2022-02-16
Add an ability to set custom payout and change addresses

### LINKED ISSUE #4092: Add ability to provide a custom payout address for a DLC
We need to offer the ability to provide a custom payout address via the UI. Users may not want their payouts sent back to the internal bitcoin-s wallet. 

This is required for a demo at bitcoin 2022.

rorp:
fixes #4092 

This PR adds the ability to specify an external payout and change address to send the DLC winnings and funding change output too.


user411:
Is there a need to specify what UTXO / funding address will fund the DLC as well? This would allow the user to send coin to a given address, then the wallet can wait on that value to show up and fund the contract with that UTXO, kick the extra to the change address, and resolve the winnings to the payout address later?

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -939,14 +982,27 @@ object DLCDataFromFile extends ServerJsonModels {
...
         }
       case pathJs :: destJs :: Nil =>
         Try {
           val path = new File(pathJs.str).toPath
           val destJsOpt = nullToOpt(destJs)
           val destOpt = destJsOpt.map(js => new File(js.str).toPath)
-          DLCDataFromFile(path, destOpt)
+          DLCDataFromFile(path, destOpt, None, None)
+        }
+      case pathJs :: destJs :: payoutAddressJs :: changeAddressJs :: Nil =>
```
benthecarman, line 1026: Should also add a case for `case pathJs :: destJs :: payoutAddressJs :: Nil`

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -848,7 +876,18 @@ object AcceptDLCOffer extends ServerJsonModels {
       case offerJs :: Nil =>
         Try {
           val offer = LnMessageFactory(DLCOfferTLV).fromHex(offerJs.str)
-          AcceptDLCOffer(offer)
+          AcceptDLCOffer(offer, None, None)
+        }
+      case offerJs :: payoutAddressJs :: changeAddressJs :: Nil =>
```
benthecarman, line 910: Should also add a case for ` case offerJs :: payoutAddressJs :: Nil`

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
```diff
@@ -279,19 +281,15 @@ abstract class DLCWallet
...
     val feeRateF = determineFeeRate(feeRateOpt).map { fee =>
       SatoshisPerVirtualByte(fee.currencyUnit)
     }
 
     for {
       feeRate <- feeRateF
-      groupedAnnouncements <- groupedAnnouncementsF
+      //hack for now to get around https://github.com/bitcoin-s/bitcoin-s/issues/3127
+      //filter announcements that we already have in the db
+      groupedAnnouncements <- groupByExistingAnnouncements(announcements)
```
benthecarman, line 298: why move this?

user411 (approved):

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
Christewart: If i understand correctly, currently we _have_ to specify the change address. I agree with Ben that that should be optional

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala
rorp: For brevity and clarity: one line > four lines

benthecarman (approved):

Christewart (approved):

PR #4104, opened 2022-02-17
Remove invoice max amount limit

benthecarman:
The invoice in the test is generated from lnd so it should be parsable by bitcoin-s


Christewart:
Looks like there is some test failures that need to be fixed.

PR #4130, opened 2022-02-25
2022 02 24 rm prune addresses after rescan

Christewart:
This addresses 2 bugs

1. make sure we can't run multiple rescans at one time
2. Make sure we don't call `pruneAddresses()` after rescan

This is how to trigger the 2nd bug

0. Alice generates address
1. `rescan` started
3. `rescan` finishes (and calls `pruneUnusedAddresses`)
4. Bob sends a payment to Alice's address

Since there were no funds on rescan, we deleted the address. Bob can still pay to the address after the rescan is finisehd and we need to be able to discover funds.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithWalletTest.scala
@@ -169,9 +169,9 @@ class NeutrinoNodeWithWalletTest extends NodeTestWithCachedBitcoindNewest {
         !rescan &&
         balance == BitcoinSWalletTest.expectedDefaultAmt + TestAmount &&
         utxos.size == 4 &&
-        addresses.map(_.scriptPubKey.hex).sorted == utxos
-          .map(_.output.scriptPubKey.hex)
-          .sorted
+        utxos
+          .map(_.output.scriptPubKey)
+          .forall(spk => addresses.exists(_.scriptPubKey == spk))
       }
     }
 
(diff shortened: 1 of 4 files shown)


benthecarman:
This will really screw up the wallet. This will make it so we increase the gap limit of the wallet by the batch size every rescan. 


I think a better solution would probably to change the wallet to mark addresses as used or unused in the db and then mark addresses as used when they are requested.

Christewart:
> I think a better solution would probably to change the wallet to mark addresses as used or unused in the db and then mark addresses as used when they are requested.

I don't understand how this solves (2) i have on the OP. The address is unused and then gets pruned with `pruneUnusedAddresses()` and then Bob sends a payment. Alice could have served the request to Bob pre-rescan IIUC

Christewart:
More over, its a very unwise idea to prune addresses at all tbh. Once you serve a bitcoin address to someone else, they can theoretically pay to that address forever. This is a problem with large services like exchanges, the list of bitcoin addresses have to monitor is always growing. There is no way to "revoke" a bitcoin address you have issued to a customer

EDIT: Well.. i guess you do have a point. You can't generate infinite address for fear of serving one to a customer, you have to use some intelligent guess.

Christewart:
> This will really screw up the wallet. This will make it so we increase the gap limit of the wallet by the batch size every rescan.

So I'm looking through this again this morning, and I don't understand how the gap limit gets increased. It seems to me that the `addressGapLimit` is always constant (configured with `bitcoin-s.wallet.addressGapLimit=20`)

Christewart:
After doing multiple rescans, the address pool stays static for me on my wallet. I always end up with 400 addresses in my wallet. It doesn't grow with every rescan on c741fda

Screenshot from 2022-02-26 07-42-17

PR #4135, opened 2022-02-27
2022 02 26 dynamic fee gui

Christewart:
Fix bug where our frontend desktop GUI had a hard coded fee rate of 50 sats/vbtyte. This PR calls our backend to determine the appropriate fee rate for network conditions.

### DIFF
--- a/app/gui/src/main/scala/org/bitcoins/gui/GlobalData.scala
+++ b/app/gui/src/main/scala/org/bitcoins/gui/GlobalData.scala
@@ -1,7 +1,9 @@
 package org.bitcoins.gui
 
-import org.bitcoins.cli.Config
+import org.bitcoins.cli.CliCommand.EstimateFee
+import org.bitcoins.cli.{Config, ConsoleCli}
 import org.bitcoins.core.config._
+import org.bitcoins.core.currency.Satoshis
 import org.bitcoins.core.wallet.fee.{FeeUnit, SatoshisPerVirtualByte}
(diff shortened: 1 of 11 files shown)


benthecarman:
It wasn't hard coded it was a `var` that was set on start up

app/gui/src/main/scala/org/bitcoins/gui/GlobalData.scala
```diff
@@ -133,7 +135,16 @@ object GlobalData {
...
-  var feeRate: FeeUnit = SatoshisPerVirtualByte.fromLong(50)
+  def getFeeRate: FeeUnit = {
+    val feeRateT = ConsoleCli
+      .exec(EstimateFee, GlobalData.consoleCliConfig)
+      .map { feeStr =>
+        val feeUnit = SatoshisPerVirtualByte(Satoshis(feeStr.toLong))
+        feeUnit
+      }
+
+    feeRateT.get
```
benthecarman, line 146: maybe should have a cache incase this fails

benthecarman (commented):
This looks better than the previous impl

Christewart:
> This looks better than the previous impl

I think we broke the previous impl in #3973 also. I believe thats why my DLCs are always using 50 sats/vbyte in fees

Somehow the exception was being swallowed

benthecarman:
>  I believe thats why my DLCs are always using 50 sats/vbyte in fees

rekt

PR #4146, opened 2022-03-01
Add the temporaryContractId field to the json rpc response for offers

Christewart:
Screenshot from 2022-03-01 11-04-54

This PR also consistently uses the json key `temporaryContractId` through the codebase to represent temporary contract Ids. Previously we used _both_ `tempContractId` and `temporaryContractId`. We choose `temporaryContractId` as that is what other implementations use.


user411 (approved):

user411:
I will migrate "offerCollateralSatoshis" to "offerCollateral" after this merges.

PR #4153, opened 2022-03-02
`offer-send` RPC

### LINKED ISSUE #4102: Implement offer mailbox
We also currently blindly accept an offer message by the looks of it

https://github.com/bitcoin-s/bitcoin-s/blob/019c9b26442d3ecb1761d3e152b9d79da4bfc9e3/dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCDataHandler.scala#L48

We should remove this functionality and replace it with a "mailbox" concept. We should store offers that our node sees, and then display them on the UI for the user to determine if they want to accept them or not. The user can accept the offer on the UI if they want. We should ideally store the tor address that the node sent the offer from so we can accept it with a button click.

Features needed from the front end: 

- Get mailbox contents 
- websocket event for when a new offer enters the mailbox
- delete mailbox offer

rorp:
fixes #4102

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
@@ -107,10 +107,10 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
 
     case ServerCommand("offer-send", arr) =>
       withValidServerCommand(OfferSend.fromJsArr(arr)) {
-        case OfferSend(address, message, offerTLV) =>
+        case OfferSend(remoteAddress, localAddress, message, offerTLV) =>
           complete {
             dlcNode
-              .sendDLCOffer(address, message, offerTLV)
+              .sendDLCOffer(remoteAddress, localAddress, message, offerTLV)
               .map { _ => Server.httpSuccess(offerTLV.hex) }
           }
       }
--- a/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
@@ -104,5 +104,15 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
           }
         }
       }
+
+    case ServerCommand("offer-send", arr) =>
+      withValidServerCommand(OfferSend.fromJsArr(arr)) {
+        case OfferSend(offerHash, address) =>
+          complete {
+            dlcNode
+              .sendDLCOffer(address, offerHash)
+              .map { _ => Server.httpSuccess(offerHash.hex) }
+          }
+      }
   }
 }
--- a/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
@@ -107,11 +107,11 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
 
     case ServerCommand("offer-send", arr) =>
       withValidServerCommand(OfferSend.fromJsArr(arr)) {
-        case OfferSend(offerHash, address) =>
+        case OfferSend(address, message, offerTLV) =>
           complete {
             dlcNode
-              .sendDLCOffer(address, offerHash)
-              .map { _ => Server.httpSuccess(offerHash.hex) }
+              .sendDLCOffer(address, message, offerTLV)
+              .map { _ => Server.httpSuccess(offerTLV.hex) }
           }
       }
   }
--- a/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
@@ -107,10 +107,10 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
 
     case ServerCommand("offer-send", arr) =>
       withValidServerCommand(OfferSend.fromJsArr(arr)) {
-        case OfferSend(remoteAddress, localAddress, message, offerTLV) =>
+        case OfferSend(remoteAddress, message, offerTLV) =>
           complete {
             dlcNode
-              .sendDLCOffer(remoteAddress, localAddress, message, offerTLV)
+              .sendDLCOffer(remoteAddress, message, offerTLV)
               .map { _ => Server.httpSuccess(offerTLV.hex) }
           }
       }
--- a/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
@@ -107,11 +107,22 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
 
     case ServerCommand("offer-send", arr) =>
       withValidServerCommand(OfferSend.fromJsArr(arr)) {
-        case OfferSend(remoteAddress, message, offerTLV) =>
+        case OfferSend(remoteAddress, message, offerE) =>
           complete {
-            dlcNode
-              .sendDLCOffer(remoteAddress, message, offerTLV)
-              .map { _ => Server.httpSuccess(offerTLV.hex) }
+            offerE match {
+              case Left(offerTLV) =>
+                dlcNode
+                  .sendDLCOffer(remoteAddress, message, offerTLV)
+                  .map { tempContractId =>
+                    Server.httpSuccess(tempContractId.hex)
+                  }
+              case Right(tempContractId) =>
+                dlcNode
+                  .sendDLCOffer(remoteAddress, message, tempContractId)
+                  .map { tempContractId =>
+                    Server.httpSuccess(tempContractId.hex)
+                  }
+            }
           }
       }
   }
--- a/app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
@@ -1413,7 +1413,7 @@ object OfferRemove {
 case class OfferSend(
     remoteAddress: InetSocketAddress,
     message: String,
-    offerTLV: DLCOfferTLV)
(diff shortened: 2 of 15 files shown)


app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
```diff
@@ -104,5 +104,15 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
           }
         }
       }
+
+    case ServerCommand("offer-send", arr) =>
+      withValidServerCommand(OfferSend.fromJsArr(arr)) {
+        case OfferSend(offerHash, address) =>
```
user411, line 110: We need a `message` string field

user411:
@rorp We don't want to send out of our local IncomingOffer list, we want to be able to send on demand with message, peer, and OfferTLV

app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
user411: This should take offerTLV, message, and peerAddress. We do not want to send out of IncomingOffers

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
```diff
@@ -79,4 +73,32 @@ case class DLCNode(wallet: DLCWalletApi)(implicit
       accept
     }
   }
+
+  override def sendDLCOffer(
+      peerAddress: InetSocketAddress,
+      incomingOfferHash: Sha256Digest): Future[Unit] = {
+    for {
+      offer <- wallet.findIncomingDLCOffer(incomingOfferHash)
```
user411: This is not what we want to be doing. We want to create a new one that is not in IncomingOffers.

docs/applications/server.md
```diff
@@ -304,9 +304,10 @@ the `-p 9999:9999` port mapping on the docker container to adjust for this.
     - `offerTLV` - Offer TLV
     - `peer` - Peer URI (optional)
     - `message` - Peer's message or note (optional)
- - `"offer-remove` `hash` - remove an incoming offer from inbox
+ - `"offer-remove` `hash` - Remove an incoming offer from inbox
     - `hash` - Hash of the offer TLV
- - `offers-list` - List all incoming offers from the inbox
+- `offer-send` `hash` `address` - Sends an offer specified by its hash to the IP/Tor address
```
user411: Signature needs message, offerTLV and peer

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1403,7 +1404,34 @@ object OfferRemove {
...
+          val message = messageJs.str
+          val offerTLV =
+            Try(LnMessageFactory(DLCOfferTLV).fromHex(offerJs.str).tlv)
+              .orElse(Try(DLCOfferTLV.fromHex(offerJs.str)))
+              .get
+          OfferSend(peer, message, offerTLV)
+        }
+      case other =>
+        val exn = new IllegalArgumentException(
+          s"Bad number or arguments to offer-send, got=${other.length} expected=1")
```
user411: expected=3

docs/applications/server.md
```diff
@@ -304,9 +304,10 @@ the `-p 9999:9999` port mapping on the docker container to adjust for this.
     - `offerTLV` - Offer TLV
     - `peer` - Peer URI (optional)
     - `message` - Peer's message or note (optional)
- - `"offer-remove` `hash` - remove an incoming offer from inbox
+ - `"offer-remove` `hash` - Remove an incoming offer from inbox
     - `hash` - Hash of the offer TLV
- - `offers-list` - List all incoming offers from the inbox
+- `offer-send` `peerAddress` `message` `offerTLV` - Sends an offer to a peer
```
user411: May want to standardize argument order to match offer-add

dlc-node/src/main/scala/org/bitcoins/dlc/node/DLCNode.scala
```diff
@@ -79,4 +72,33 @@ case class DLCNode(wallet: DLCWalletApi)(implicit
...
+      message: String,
+      offerTLV: DLCOfferTLV): Future[Unit] = {
+    for {
+      handler <- connectToPeer(peerAddress)
+    } yield {
+      val peer = NormalizedString(
+        peerAddress.getHostString + ":" + peerAddress.getPort)
+      val msg = NormalizedString(message)
+      val lnMessage = LnMessage(
+        DLCOfferMessageTLV(peer = peer, message = msg, offer = offerTLV))
```
user411: We want our local Tor DLC Host Address injected here sending the message to our friend (if I'm reading this right) so they can get back to us.

app/server/src/main/scala/org/bitcoins/server/DLCRoutes.scala
```diff
@@ -104,5 +104,15 @@ case class DLCRoutes(dlcNode: DLCNodeApi)(implicit system: ActorSystem)
           }
         }
       }
+
+    case ServerCommand("offer-send", arr) =>
+      withValidServerCommand(OfferSend.fromJsArr(arr)) {
+        case OfferSend(remoteAddress, localAddress, message, offerTLV) =>
```
user411: I shouldn't have to send the Tor DLC Host address back for this unless there is some use case that makes sense not to use the local one (which is known to the backend already)

user411 (approved):

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -1687,6 +1688,38 @@ object FundingSignaturesV0TLV extends TLVFactory[FundingSignaturesV0TLV] {
...
+  override val tpe: BigSizeUInt = DLCOfferMessageTLV.tpe
+
+  override val value: ByteVector = {
+    strBytes(peer) ++ strBytes(message) ++ offer.bytes
+  }
+}
+
+object DLCOfferMessageTLV extends TLVFactory[DLCOfferMessageTLV] {
+
+  override val tpe: BigSizeUInt = BigSizeUInt(65534)
```
Christewart, line 1710: @nkohen is there a good way to pick TLV  numbers?

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1403,7 +1404,35 @@ object OfferRemove {
...
+          s"Bad number or arguments to offer-remove, got=${other.length} expected=1")
+        Failure(exn)
+    }
+  }
+}
+
+case class OfferSend(
+    remoteAddress: InetSocketAddress,
+    message: String,
+    offerTLV: DLCOfferTLV)
```
user411: This needs to take an offer temporaryContractId so we can call it again if it doesn't go through the first time from the Contracts list where we don't have the DLCOfferTLV. The backend needs to look up and pull the DLCOfferTLV from the database.

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
nkohen: I usually use the next even number

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
```diff
@@ -1687,6 +1688,38 @@ object FundingSignaturesV0TLV extends TLVFactory[FundingSignaturesV0TLV] {
 
 sealed trait DLCSetupTLV extends TLV
 
+case class DLCOfferMessageTLV(
```
nkohen: probably worth a scaladoc since the name is so similar to `DLCOfferTLV` which could be confusing

dlc-test/src/test/scala/org/bitcoins/dlc/testgen/DLCParsingTestVector.scala
```diff
@@ -550,7 +550,8 @@ object DLCParsingTestVector extends TestVectorParser[DLCParsingTestVector] {
         )
 
         DLCMessageTestVector(LnMessage(tlv), "oracle_attestment_v0", fields)
-      case _: UnknownTLV | _: ErrorTLV | _: PingTLV | _: PongTLV | _: InitTLV =>
+      case _: UnknownTLV | _: ErrorTLV | _: PingTLV | _: PongTLV | _: InitTLV |
+          _: DLCOfferMessageTLV =>
```
nkohen: Should probably add a parsing test

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
Christewart: next relative to what?

docs/applications/server.md
```diff
@@ -300,13 +300,14 @@ the `-p 9999:9999` port mapping on the docker container to adjust for this.
...
- - `offer-add` `offerTLV` `peer` `message` - Puts an incoming offer into the inbox
+ - `offer-add` `offerTLV` `peerAddress` `message` - Puts an incoming offer into the inbox
     - `offerTLV` - Offer TLV
     - `peer` - Peer URI (optional)
     - `message` - Peer's message or note (optional)
- - `"offer-remove` `hash` - remove an incoming offer from inbox
+ - `"offer-remove` `hash` - Remove an incoming offer from inbox
     - `hash` - Hash of the offer TLV
- - `offers-list` - List all incoming offers from the inbox
+- `offer-send` `offerTLV` `peerAddress` `message` - Sends an offer to a peer
```
user411: This needs to be offer.temporaryContractId instead of offerTLV to support resending from the Contracts view

core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala
rorp: Any suggestions for a better name for this type?

Christewart: `SendOfferTLV` perhaps?

Christewart (approved):

user411 (approved):

Christewart (approved):

PR #4154, opened 2022-03-03
Round `Bitcoins` to nearest `Satoshi`

benthecarman:
This fixes a bug where we would allow having a fraction of a satoshi with `Bitcoins`. This fixes it by rounding it using `setScale` using the default rounding mode `HALF_UP`, curious on thoughts on using `DOWN` so we guarantee there is no extra sats.


Christewart:
> This fixes a bug where we would allow having a fraction of a satoshi with `Bitcoins`. This fixes it by rounding it using `setScale` using the default rounding mode `HALF_UP`, curious on thoughts on using `DOWN` so we guarantee there is no extra sats.

Is there a reason you didn't pick `DOWN`? IMO that does seem like a better default :man_shrugging: . Maybe you can make it a default parameter to that `apply()` method so the caller of the method can specify that

Christewart (approved):

PR #4156, opened 2022-03-03
Add log message for the case where we match a script but do not have an address for it

Christewart:
Helps to debug our problem where we aren't finding funds on rescan


benthecarman:
We add the dlc multisig to the spk table so this will always log those

PR #4166, opened 2022-03-04
`getdlcoffer` RPC

### LINKED ISSUE #4164: Add `getoffer temporaryContractId` endpoint
This allows the frontend to query the backend for an arbitrary `temporaryContractId`. 

The use case is for users to be able to copy the offer tlv _after_ it's initially displayed to the user. Currently if you create an offer via the UI, and then click away without copying the tlv to your clipboard, you have no ability to get the tlv for the offer again.

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4164

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
@@ -256,6 +256,29 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
           }
       }
 
+    case ServerCommand("getdlcoffer", arr) =>
+      GetDLCOffer.fromJsArr(arr) match {
+        case Failure(exception) =>
+          complete(Server.httpBadRequest(exception))
+        case Success(GetDLCOffer(tempContractId)) =>
+          complete {
+            for {
+              dlcOpt <- wallet.findDLCByTemporaryContractId(tempContractId)
+              dlc = dlcOpt.getOrElse(
+                throw new IllegalArgumentException(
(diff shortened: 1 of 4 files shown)


Christewart:
Does not compile

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -256,6 +256,29 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
...
+              dlcOpt <- wallet.findDLCByTemporaryContractId(tempContractId)
+              dlc = dlcOpt.getOrElse(
+                throw new IllegalArgumentException(
+                  s"Cannot find a DLC with temp contact ID $tempContractId"))
+              offerOpt <- wallet.getDLCOffer(dlc.dlcId)
+              offer = offerOpt.getOrElse(
+                throw new IllegalArgumentException(
+                  s"Cannot find an offer with for DLC ID ${dlc.dlcId}"))
+            } yield {
+              val tlv = offer.toTLV
```
Christewart: We return the `LnMessage(tlv)` here. We should probably be consistent and return the same format here too.

https://github.com/bitcoin-s/bitcoin-s/blob/9cfe65f1cca8e3a7df78e0c1d07866bfbe9a28da/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala#L348

Christewart (changes requested):
A unit test would be nice so we can detect future regressions if we ever accidentally break the API

Christewart (approved):

user411 (approved):
This tests well for me

PR #4173, opened 2022-03-08
Mark extarnal payout addresses

### LINKED ISSUE #4125: DLC Contracts with external payout addresses need marked with the addresses
**Describe the bug**
DLCs created with external payout addresses for maker or taker need those address on the DLCContract data structure so the UI can present them to users. Good field names would be 'makerPayoutAddress' and 'takerPayoutAddress' or use relative terms like other fields and do 'localPayoutAddress' and 'remotePayoutAddress'.

**To Reproduce**
1. Make a DLC contract where each side uses external payout addresses
2. Notice they are not on the data for the returned contract.

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4125

Adds `isExternalAddress` field for `DLCAccept` and and `DLCOffer`. 

Also adds optional `payoutAddress` field (which contains the address and the external flag) to `DLCStatus`.

### DIFF
--- a/app-commons-test/src/test/scala/org/bitcoins/commons/DLCStatusTest.scala
+++ b/app-commons-test/src/test/scala/org/bitcoins/commons/DLCStatusTest.scala
@@ -4,7 +4,11 @@ import org.bitcoins.commons.serializers.Picklers
 import org.bitcoins.commons.serializers.Picklers._
 import org.bitcoins.core.currency.{CurrencyUnit, Satoshis}
 import org.bitcoins.core.protocol.dlc.models.DLCMessage._
-import org.bitcoins.core.protocol.dlc.models.{DLCState, DLCStatus}
+import org.bitcoins.core.protocol.dlc.models.{
+  DLCState,
+  DLCStatus,
+  PayoutAddress
+}
 import org.bitcoins.core.util.TimeUtil
 import org.bitcoins.crypto.Sha256Digest
 import org.bitcoins.testkitcore.gen.{CryptoGenerators, NumberGenerator, TLVGen}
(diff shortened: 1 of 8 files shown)


user411:
So the frontend/backend will get to see the `payoutAddress` object (including isExternal) for the counterparty in the Offer serialization? If this is the case, isn't it possible to database it and have both the local and remote payoutAddresses and whether they are marked external in the local database?

Christewart:
> Fixes #4125
> 
> Adds `isExternalAddress` field for `DLCAccept` and and `DLCOffer`.
> 
> Also adds optional `payoutAddress` field (which contains the address and the external flag) to `DLCStatus`.

Responses I see on `getdlcs`. This seems to LGTM

Screenshot from 2022-03-09 14-24-48

Christewart:
> So the frontend/backend will get to see the `payoutAddress` object (including isExternal) for the counterparty in the Offer serialization? If this is the case, isn't it possible to database it and have both the local and remote payoutAddresses and whether they are marked external in the local database?

There is a difference between [`DLCOffer`](https://github.com/bitcoin-s/bitcoin-s/blob/349d7a15ad6ee0a1671c7585630231d0a93a5557/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/DLCMessage.scala#L79) and [`DLCOfferTLV`](https://github.com/bitcoin-s/bitcoin-s/blob/56d0ae68ad19969094aa879014c19ae6c53bef3c/core/src/main/scala/org/bitcoins/core/protocol/tlv/TLV.scala#L1723). If the case class has a `TLV` suffix on the end of it, this is the binary level serialized datastructure that gets sent over the wire. 

If the case class does not have a `TLV` suffix, it is an in memory representation that is meant to be easier to work with in Scala.

user411 (approved):

Christewart (approved):

PR #4175, opened 2022-03-09
2022 03 09 label refactor

Christewart:
This reworks the labeling API to be easier to use. 

This PR adds the following RPCs 

- `getaddresslabel` `address` gets all labels for a given address
- `dropaddresslabel` `address` `label` drops the specific label for a given address

This PR modifies these existing RPCs

- `getaddresslabels`  returns all addresses that have labels, and their associated labels
- `dropaddresslabels` `address` this drops all labels for the given addresses.

Screenshot from 2022-03-09 10-46-51


user411 (approved):
Ideally in a web environment where you just want to be able to update one label you would not have to make a delete and replace call to do so.

PR #4188, opened 2022-03-14
Bind to localhost rather than 127.0.0.1

### LINKED ISSUE #4178: Add ipv6 support to websocket/rpc server
Was working through something with @nkohen this morning and we got hung up due to our software being incompatible with node17. We need to support binding to ipv6 addresses to work with node17/angular on bitcoin-s-ts

Christewart:
fixes #4178 

I can node use `node 17.x` locally and it seems to work. I would like others to test this to confirm it works on their machines.

### DIFF
--- a/db-commons/src/main/resources/reference.conf
+++ b/db-commons/src/main/resources/reference.conf
@@ -155,11 +155,11 @@ bitcoin-s {
     server {
         # The port we bind our rpc server on
         rpcport = 9999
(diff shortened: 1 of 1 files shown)


Christewart:
> fixes #4178
> 
> I can node use `node 17.x` locally and it seems to work. I would like others to test this to confirm it works on their machines.

Actually it looks like this is working regardless of the version of node i'm using, so i don't think this fixes anything.

PR #4205, opened 2022-03-21
Update lots of deps

benthecarman:


Christewart:
What happened to scala-steward ? ☠️ ☠️. I think i would prefer to merge this after we cut 1.9.1 (https://github.com/bitcoin-s/bitcoin-s/projects/9) so we don't introduce any surprises just before releasing.

Is there anything time sensitive for you in here?

benthecarman:
> Is there anything time sensitive for you in here?

Mainly just wanted the akka update

project/Deps.scala
```diff
@@ -5,41 +5,41 @@ object Deps {
...
+    val postgresV = "42.3.3"
     val akkaActorV = akkaStreamv
     val slickV = "3.3.3"
     val sqliteV = "3.36.0.3"
 
     val scalameterV = "0.17"
     val scalamockV = "5.2.0"
     val scalaCollectionCompatV = "2.6.0"
-    val pgEmbeddedV = "0.13.4"
+    val pgEmbeddedV = "1.0.0"
```
Christewart, line 42: This needs to be reverted. They are moving to a docker based postgres environment in the `1.x.x` series. We do not have that setup. 

https://github.com/opentable/otj-pg-embedded/issues/166

testkit/src/main/scala/org/bitcoins/testkit/EmbeddedPg.scala
```diff
@@ -35,7 +35,7 @@ trait EmbeddedPg extends BeforeAndAfterAll { this: Suite =>
   }
 
   def pgUrl(): Option[String] =
-    pg.map(_.getJdbcUrl(userName = "postgres", dbName = "postgres"))
+    pg.map(_.getJdbcUrl("postgres"))
```
Christewart, line 38: i'm guessing this will need to be reverted too?

Christewart (changes requested):

project/Deps.scala
Christewart: https://github.com/bitcoin-s/bitcoin-s/runs/5634927314?check_suite_focus=true#step:6:2672

PR #4223, opened 2022-03-28
Support for Tor Docker

rorp:


Christewart:
I've published the image here and it seems to be working: https://hub.docker.com/repository/docker/bitcoinscala/tor

This does not have automated publishing to dockerhub currently, every time we need to update `tor/Dockerfile` we will have to manually publish the image

Christewart (approved):

PR #4225, opened 2022-03-29
Fix peer configuration for docker-compose

rorp:


Christewart (approved):

Christewart (changes requested):
This does not work for me

Screenshot from 2022-03-29 14-07-05

Christewart (approved):
Apologies, this does work after building the image locally with teh `[]` inside of the `docker-application.conf`

PR #4232, opened 2022-03-31
2022 03 31 collateral invariant

Christewart:
See the spec for more information: https://github.com/discreetlogcontracts/dlcspecs/blob/master/Protocol.md#the-offer_dlc-message

These names were incorrect and very confusing in our codebase.


Christewart:
Test failures are not specific to this PR and are being addressed in #4233

PR #4265, opened 2022-04-16
Remove staking address from codebase

Christewart:
Remove staking address from codebase. We shouldn't serve an address we can't spend from. Its going to take a bit of logic to implement spending from the staking address, and that is a low priority due to the questionability of how to implement staking addresses that can't trivially be spent from before the oracle lies.


Christewart:
This cannot be done now, we have a user that has funds in a staking address, see #4267

PR #4285, opened 2022-04-22
Default createDLCOffer to current block height

nkohen:
Added createDLCOffer endpoint which doesn't take a locktime and uses current block height instead.

Did I miss adding any tests? Not sure where one would go.


benthecarman (approved):
Lgtm

Christewart (changes requested):
The code looks ok to me, but documentation should be updated IMO to use this. That way our examples are closer to working "out of the box". Unfortunately I still the refundLT will have to be manually configured by the user :-(
https://bitcoin-s.org/docs/applications/server#dlc

Screenshot from 2022-04-25 09-47-30

https://bitcoin-s.org/docs/wallet/wallet-election-example#offer

https://bitcoin-s.org/docs/wallet/wallet-price-example#offer

user411 (approved):

Christewart (changes requested):
On ee72683e6939be2d400f6e52ec1a4f2b0b230ffd a API breaking change is introduced. Namely the positions of `refundLT` and `cetLT` are switched in their parameter positions on the backend server. This has the effect of breaking the API. 

Here is an example of the error message we get on the backend when using bitcoin-s-ts

Screenshot from 2022-04-25 12-14-17

IMO, the least painful thing to do is just switch the parameter positions on the `cli`, and leave them the same on the backend. I would suggest reverting the backend changes and then swapping the positioning of the refund locktime and cet locktime when building the request for the cli here: 

https://github.com/bitcoin-s/bitcoin-s/blob/ee72683e6939be2d400f6e52ec1a4f2b0b230ffd/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala#L1895-1896

Christewart:
Looks like there is a compile failure: https://github.com/bitcoin-s/bitcoin-s/runs/6162568443?check_suite_focus=true#step:5:145

Christewart (approved):

PR #4286, opened 2022-04-24
Add router rpc to lnd, use for paying invoices

benthecarman:


lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
```diff
@@ -516,26 +518,48 @@ class LndRpcClient(val instance: LndInstance, binaryOpt: Option[File] = None)(
       }
   }
 
-  def sendPayment(invoice: LnInvoice): Future[SendResponse] = {
-    val request: SendRequest = SendRequest(paymentRequest = invoice.toString)
+  def sendPayment(invoice: LnInvoice, timeout: Int): Future[Payment] = {
```
Christewart: Any reason to not make `timeout` a `FiniteDuration` and then do the appropriate conversion to whatever the required unit required by `lnd` in this method?

benthecarman: Good idea, fixed

Christewart (approved):

Issue #4051, opened 2022-02-06
Implement compression and rolling of `bitcoin-s.log` file


Christewart:
I've got 4,000,000 lines in my `bitcoin-s.log` file due to #3983 

We need to implement rolling and compression of the bitcoin-s.log file. 

See: https://logback.qos.ch/manual/appenders.html#RollingFileAppender

PR #4288, opened 2022-04-24
Fix rolling log file

benthecarman:
Previously the rolling log file was not working for me, this seems to be the fix

### DIFF
--- a/app/oracle-server/src/main/resources/logback.xml
+++ b/app/oracle-server/src/main/resources/logback.xml
@@ -11,7 +11,7 @@
         <appender-ref ref="STDOUT" />
     </appender>
 
-    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+    <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${bitcoins.log.location}/bitcoin-s.log</file>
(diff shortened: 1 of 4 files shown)


Christewart:
#4051

Christewart (changes requested):
[I don't think this works](https://logback.qos.ch/manual/appenders.html#RollingFileAppender)

>To be of any use, a RollingFileAppender must have both a RollingPolicy and a TriggeringPolicy set up. However, if its RollingPolicy also implements the TriggeringPolicy interface, then only the former needs to be specified explicitly. 

For instance, the log file rolls after a certain amount time passes, or a certain amount of data stored in the log file. If you look into this further, it would also be nice to compress the log file we just produced to minimize disk space used by the archived log files ( #4051 )

benthecarman:
Looks like it implements the triggering policy https://logback.qos.ch/manual/appenders.html#SizeAndTimeBasedRollingPolicy

not sure how to do the compression

Christewart:
> Looks like it implements the triggering policy https://logback.qos.ch/manual/appenders.html#SizeAndTimeBasedRollingPolicy
> 
> not sure how to do the compression

Ah, sorry I didn't realize we already had `<rollingPolicy></rollingPolicy>` defined

Christewart (approved):

PR #4295, opened 2022-04-27
LnURL Module

benthecarman:
Creates a new LnURL module

Has support for:

- [LUD 1](https://github.com/fiatjaf/lnurl-rfc/blob/luds/01.md)
- [LUD 3](https://github.com/fiatjaf/lnurl-rfc/blob/luds/03.md)
- [LUD 6](https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md)
- [LUD 9](https://github.com/fiatjaf/lnurl-rfc/blob/luds/09.md)

### DIFF
--- a/.github/workflows/Linux_2.12_App_Chain_Core_Tests.yml
+++ b/.github/workflows/Linux_2.12_App_Chain_Core_Tests.yml
@@ -28,4 +28,4 @@ jobs:
             ~/.bitcoin-s/binaries
           key: ${{ runner.os }}-cache
       - name: run tests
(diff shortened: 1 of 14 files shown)


Christewart (changes requested):
I have not looked at this at all yet, but this needs to be run on CI, thus it needs to be added to some `yml` files so github work flows picks it up

fiatjaf:
Looks good to me, if a lazy review from someone completely foreign to the codebase counts for anything.

PR #4296, opened 2022-04-27
Handle duplicate UTXOs

rorp:


app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
```diff
@@ -518,6 +516,27 @@ class BitcoinSServerMain(override val serverArgParser: ServerArgParser)(implicit
...
+      rescanNeeded <- spendingInfoDAO.hasDuplicates()
+      _ <-
+        if (rescanNeeded) {
+          logger.warn("Found duplicate UTXOs. Rescanning...")
+          wallet
+            .rescanNeutrinoWallet(startOpt = None,
+                                  endOpt = None,
+                                  addressBatchSize = wallet.discoveryBatchSize,
+                                  useCreationTime = true)
+            .recover { case _: Throwable => RescanState.RescanDone }
```
Christewart: What is the possible exception here? Can we log it in case it is something unexpected? Also maybe we want to catch `NonFatal()`

wallet/src/main/scala/org/bitcoins/wallet/models/SpendingInfoDAO.scala
```diff
@@ -565,6 +565,19 @@ case class SpendingInfoDAO()(implicit
       .map(_ => ts.map(_.copyWithState(TxoState.Reserved)))
   }
 
+  def createOutPointsIndexIfNeeded(): Future[Unit] = {
+    val query =
+      sqlu"CREATE UNIQUE INDEX IF NOT EXISTS utxo_outpoints ON txo_spending_info (tx_outpoint)"
```
Christewart, line 570: So eventually (after a few releases?) we will want to move this to be a proper migration, right? 

If so, i'm going to file an issue so that I don't forget to do this in the future.

Christewart:
to loop in @user411 here, this is going to _possibly_ -- if they have duplicate utxos in their wallet -- require a rescan when when a user starts up their wallet. Should we display something to the user or is our user base small enough to just make this happen behind the scenes?

The rescan would likely take 5-10 minutes.

This worked on my wallet with duplicated utxos. I ran this on 26fe336

Christewart (approved):

user411:
> 

My guess is that Tor Starting and Rescanning are states that the UI will need to be aware of and that will lockdown certain actions. That lockdown will need presented to the user as a state the backend is in and it may change the UI polling behavior if we need to proactively keep track of the progress of a Rescan or Tor Starting. Websocket events for completion of Rescan or Tor Startup would be a great way to address and would keep the UI state handling simpler.

We know that it would be nice to track and display rescanning state in general, but I wouldn't let that specifically stand in the way of fixing database issues that could be preventing use of the wallet.

Christewart:
needs rebase

Issue #4297, opened 2022-04-28
`[SQLITE_TOOBIG]  String or BLOB exceeds size limit (statement too long)` when processing block


Christewart:
Seen on 4b964074 when processing a block on my umbrel this morning. I'm not sure what the root cause is

```
 2022-04-28T03:11:57UTC INFO [BitcoindRpcBackendUtil$] Fetching 1 blocks from bitcoind
2022-04-28T03:11:58UTC INFO [DLCWallet$DLCWalletImpl] Processing block=0000000000000000000779b40ecfb60de84f11fc7e598c668fa4623987d25fd2
2022-04-28T03:11:59UTC ERROR [SafeDatabase] Error when executing query flatMap slick.dbio.DBIOAction$$Lambda$1087/0x000000080111a680@51b4f6f2
2022-04-28T03:11:59UTC ERROR [SafeDatabase] org.sqlite.SQLiteException: [SQLITE_TOOBIG]  String or BLOB exceeds size limit (statement too long)
2022-04-28T03:11:59UTC ERROR [BitcoindRpcBackendUtil$] Requesting blocks from bitcoind polling failed, range=[733898, 733899]
org.sqlite.SQLiteException: [SQLITE_TOOBIG]  String or BLOB exceeds size limit (statement too long)
	at org.sqlite.core.DB.newSQLException(DB.java:1030)
	at org.sqlite.core.DB.newSQLException(DB.java:1042)
	at org.sqlite.core.DB.throwex(DB.java:1007)
	at org.sqlite.core.NativeDB.prepare_utf8(Native Method)
	at org.sqlite.core.NativeDB.prepare(NativeDB.java:137)
	at org.sqlite.core.DB.prepare(DB.java:257)
	at org.sqlite.core.CorePreparedStatement.<init>(CorePreparedStatement.java:45)
	at org.sqlite.jdbc3.JDBC3PreparedStatement.<init>(JDBC3PreparedStatement.java:30)
	at org.sqlite.jdbc4.JDBC4PreparedStatement.<init>(JDBC4PreparedStatement.java:25)
	at org.sqlite.jdbc4.JDBC4Connection.prepareStatement(JDBC4Connection.java:35)
	at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:241)
	at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:205)
[...]

PR #4299, opened 2022-04-28
Improve logging around signDLC

Christewart:
More improvements to logging, specifically on `signDLC`


Christewart:
Added some logs around failing to process a transaction so we can hopefully debug #4297

PR #4326, opened 2022-05-09
Persist whether wallet is rescanning in the database

### LINKED ISSUE #4043: Persist whether wallet is rescanning in the database, so we can restart the rescan on next bootup
If a user powers down their wallet during a rescan, on next startup they will end up in a weird wallet state. We need to restart the rescan from the last compact filter seen. 

1. Make sure we can determine if our wallet was rescanning
2. Restart the rescan on boot up if it was.

This happened to a user when they were using a raspberry pi. They restarted the pi for unrelated reasons while the wallet was in the middle of a rescan.

Screenshot from 2022-02-04 09-07-33

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4043


wallet/src/main/resources/postgresql/wallet/migration/V17__create_wallet_state.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE wallet_state (id INT NOT NULL PRIMARY KEY DEFAULT(1), is_rescanning BOOL NOT NULL);
```
Christewart: Maybe we want to consider just adding this as a column to `state_descriptors` table?

https://github.com/bitcoin-s/bitcoin-s/blob/f71d3567ed21df6bb050ee184d6fe31b4369c654/wallet/src/main/scala/org/bitcoins/wallet/models/WalletStateDescriptorDAO.scala#L89

Christewart (commented):
The only suggestion I have is adding the rescanning flag to `state_descriptors` to reduce the tables in the wallet. If you don't want to do that, I think this PR looks good otherwise.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -299,79 +299,82 @@ object BitcoindRpcBackendUtil extends Logging {
...
   }
 
   def startBitcoindMempoolPolling(
+      wallet: WalletApi,
       bitcoind: BitcoindRpcClient,
       interval: FiniteDuration = 10.seconds)(
       processTx: Transaction => Future[Unit])(implicit
       system: ActorSystem,
-      ec: ExecutionContext): Cancellable = {
+      ec: ExecutionContext): Future[Unit] = {
```
Christewart: nit: Return `Future[Cancellable]` so that the user of this method can cancel the scheduled task if they want too.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -383,51 +386,59 @@ object BitcoindRpcBackendUtil extends Logging {
...
-            val processors = Runtime.getRuntime.availableProcessors()
-            //max open requests is 32 in akka, so 1/8 of possible requests
-            //can be used to query the mempool, else just limit it be number of processors
-            //see: https://github.com/bitcoin-s/bitcoin-s/issues/4252
-            Math.min(4, processors)
-          }
+    for {
+      rescanning <- wallet.isRescanning()
+    } yield {
+      if (!rescanning) {
```
Christewart: This means if we start the wallet and we are rescanning, we will never start polling bitcoind after the rescan is finished.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -299,79 +299,82 @@ object BitcoindRpcBackendUtil extends Logging {
...
-      system.scheduler.scheduleWithFixedDelay(0.seconds, interval) { () =>
-        {
-          if (processing.compareAndSet(false, true)) {
-            logger.trace("Polling bitcoind for block count")
-            val res = bitcoind.getBlockCount.flatMap { count =>
-              val prevCount = atomicPrevCount.get()
-              if (prevCount < count) {
-                logger.info(
-                  s"Bitcoind has new block(s), requesting... ${count - prevCount} blocks")
+      if (!rescanning) {
```
Christewart: This means if we start the wallet and we are rescanning, we will never start polling bitcoind after the rescan is finished. Its probably worth adding a unit test case for this as it is a pretty tricky corner case.

Christewart (changes requested):

Christewart (approved):

PR #4332, opened 2022-05-11
Make all uint64 types from lnd a UInt64

benthecarman:
This makes it so any uint64 from the lnd rpc is converted into a bitcoin-s `UInt64` instead of a `Long`. The encoding would still work fine previously (https://github.com/scalapb/ScalaPB/issues/1368) but sometimes the actual number wasn't useful. This also fixes some potential errors where we would call `longExact` on some values and it couldn't be converted into a `Long` and thus throw an error.


benthecarman:
Did the same for uint32

Christewart (approved):

PR #4323, opened 2022-05-08
Fix `ERROR: relation "txo_spending_info" does not exist`

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4321

PR #4336, opened 2022-05-17
Validate `bitcoin-s.wallet.walletName` config option

Christewart:
This came up in #4323 . This makes sure that the walletName is alpha numeric and only contains `-` or `_`

https://github.com/bitcoin-s/bitcoin-s/pull/4323#issuecomment-1121402512

### DIFF
--- a/key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
+++ b/key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
@@ -30,7 +30,10 @@ case class KeyManagerAppConfig(
   lazy val networkParameters: NetworkParameters = chain.network
 
   lazy val walletNameOpt: Option[String] = {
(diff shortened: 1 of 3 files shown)


key-manager/src/main/scala/org/bitcoins/keymanager/config/KeyManagerAppConfig.scala
```diff
@@ -30,7 +30,10 @@ case class KeyManagerAppConfig(
   lazy val networkParameters: NetworkParameters = chain.network
 
   lazy val walletNameOpt: Option[String] = {
-    config.getStringOrNone(s"bitcoin-s.wallet.walletName")
+    val nameOpt = config.getStringOrNone(s"bitcoin-s.wallet.walletName")
+    require(nameOpt.map(KeyManagerAppConfig.validateWalletName).getOrElse(true),
+            s"Invalid wallet name, only alphanumeric with _, got=$nameOpt")
```
benthecarman, line 35: this error message doesn't mention you can include `-`

PR #4346, opened 2022-05-18
DLC <-> contact mapping

rorp:
Creates a one-to-many relation between DLCs and contacts.

Add `dlc-contact-add` and `dlc-contact-remove` RPC calls to add and remove the associations.

`getdlc` and `getdlcs` RPC calls return JSON responses with associated contacts.


user411:
I made an Offer to a peer on my Contact List using this branch (which they then accepted) and the contact field on the resulting DLC came back null. I would expect it to be populated automatically.

Screen Shot 2022-05-23 at 9 27 40 AM

user411:
I notice DLC.contact can be null, a string, or a Contact object. I would prefer it was always the peer address string (contactId) or null. We could potentially call this field 'peer' vs 'contact' as it's the peer we're working with and we call it 'peer' in the IncomingOffer context.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/internal/IncomingDLCOffersHandling.scala
```diff
@@ -58,4 +58,15 @@ trait IncomingDLCOffersHandling { self: DLCWallet =>
   override def findDLCContacts(alias: String): Future[Vector[DLCContactDb]] =
     contactDAO.findByAlias(alias)
 
+  override def addDLCContactMapping(
+      dlcId: Sha256Digest,
+      contcatId: InetSocketAddress): Future[Unit] = {
```
Christewart, line 73: Why not just return the `DLCContactMappingDb` rather than `Unit`? That way we can actually do something with the returned value, especially if we every modify `ContactDb` in the future to be more than `dlcId` and `contcatId` ?

I don't think there is any reason `DLCContactMappingDb` can't be moved to `core`.

dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/models/DLCContactMappingDAO.scala
```diff
@@ -0,0 +1,155 @@
...
+import org.bitcoins.core.api.dlc.wallet.db.{DLCContactDb, DLCDb}
+import org.bitcoins.crypto.Sha256Digest
+import org.bitcoins.db.{CRUD, SlickUtil}
+import org.bitcoins.dlc.wallet.DLCAppConfig
+import slick.lifted.ForeignKeyQuery
+
+import java.net.InetSocketAddress
+import scala.concurrent.{ExecutionContext, Future}
+
+case class DLCContactMapping(dlc: DLCDb, contact: Option[DLCContactDb]) {
```
Christewart: Move `DLCContactMapping` and `DLCContactMappingDb` to core

Christewart (changes requested):
Mostly just code re-organization, but @user411 's requested API level changes need to be addressed too.

user411:
If an IncomingOffer has a peer that is not in the contacts list, when completing the DLC, the DLC.peer field does not come in with a value. We need that field to populate so the user can later make a contact from their DLC. DLC.peer should not be keyed/indexed to the contacts address field. It should just be the Tor address for the counterparty of the DLC, null, or whatever the user has arbitrarily set.

user411:
We should not be creating Contacts for users, we should be leaving the Tor DLC Host Address behind so they can create their own contacts from previous data as they like.

user411:
Running on a261a7b, when I start up I get

```
[info] 2022-06-09T15:51:55UTC INFO [NodeAppConfig] Applied org.flywaydb.core.api.output.MigrateResult@6a0b7489 migrations fro the node project
[info] 2022-06-09T15:51:55UTC ERROR [BitcoinSServerMain] Failed to startup server!
[info] org.flywaydb.core.api.FlywayException: Unable to baseline schema history table "main"."flyway_schema_history" as it already contains migrations
```

Any recommendation on getting around this? I ran this branch at several other points during its history.

user411:
I deleted my dlcdb.sqlite to get past the last issue.

I started up, lost all contacts I had previously and of course DLCs. I made a fake contact then I sent a new offer over from another computer, and accepted it. Then I went to contacts and tried to create a contact for the address the offer came in from. I got the following:
Screen Shot 2022-06-09 at 10 53 49 AM

I previously had this address in use as a contact. Do I need to do more to reset my state? If the database really was destroyed, I wouldn't expect anything to get in the way of this keying - the contact doesn't exist yet according to the list received.

user411 (approved):

Christewart (approved):

PR #4348, opened 2022-05-22
Fix lnd sendToRoute for 0 amount invoices

benthecarman:


Christewart (commented):
can we add a unit test? This seems like a tricky scenario that is easy to forget about

benthecarman:
> can we add a unit test? This seems like a tricky scenario that is easy to forget about

Good idea, done

Christewart:
typo i believe https://github.com/bitcoin-s/bitcoin-s/runs/6546624277?check_suite_focus=true#step:5:213

PR #4356, opened 2022-05-29
Remove Spv code

### LINKED ISSUE #4215: Remove `Spv` code.
I don't think this is necessary anymore and just adds dead weight to the codebase. I dont believe bitcoin core even recommends spv anymore.

solo1g:
Closes #4215.


Christewart (approved):

Christewart (changes requested):
I think these can be removed as well: 

- https://github.com/bitcoin-s/bitcoin-s/blob/8f958bf5d9957438d703819a087d87813f3edd89/core/src/main/scala/org/bitcoins/core/api/wallet/SpvWalletApi.scala#L13
- https://github.com/bitcoin-s/bitcoin-s/blob/5777ec1c316a3e160a3706cf9fd119c2e7436452/testkit/src/main/scala/org/bitcoins/testkit/BitcoinSTestAppConfig.scala#L61

Otherwise this LGTM 🎉

Christewart (approved):

PR #4357, opened 2022-06-01
Seed backup API

### LINKED ISSUE #4339: Add rpc for fetching mnemonic seed
We need the ability to display a mnemonic seed to users when the wallet is initialized so they can properly backup the wallet on desktop environments. this needs to be exposed on both the `oracleServer` and `appServer`s

I believe we have the ability to import a mnemonic seed on the `appServer` but we don't have it on the `oracleServer` 

https://github.com/bitcoin-s/bitcoin-s/blob/525fb2ac0dab126bbb428e2fe1470462fa244ac8/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala#L842

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/4339


key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -149,6 +156,19 @@ object WalletStorage extends KeyManagerLogger {
     }
   }
 
+  private def parseBackupTime(json: Value): Option[Long] = {
+    Try(json(MnemonicJsonKeys.BACKUP_TIME)) match {
+      case Success(js) =>
+        js match {
+          case ujson.Null         => None
+          case value: ujson.Value => Some(Try(value.num.toLong).get)
```
Christewart: nit: I think `Try(value.num.toLong).toOption` works too

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1598,4 +1618,47 @@ trait ServerJsonModels {
...
   }
+
+  def jsToAESPassword(js: Value): Option[AesPassword] = {
+    js match {
+      case Str(str) =>
+        Some(AesPassword.fromString(str))
+      case Null =>
+        None
+      case Arr(_) | False | True | Num(_) | Obj(_) =>
+        throw new IllegalArgumentException("password must be a string or null")
```
Christewart, line 1677: nit: Always nice to add the input to the error message for ease of debugging

i.e.

```scala
      case x @ (Arr(_) | False | True | Num(_) | Obj(_)) =>
        throw new IllegalArgumentException(s"password must be a string or null, got=$x")
```

now the error message will contain the input.

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1598,4 +1618,47 @@ trait ServerJsonModels {
...
+          case walletNameJs :: Nil =>
+            (jsToStringOpt(walletNameJs), None)
+          case Nil =>
+            (None, None)
+          case other =>
+            throw new IllegalArgumentException(
+              s"Bad number of arguments: ${other.length}. Expected: 2")
+        }
+      case _: Value =>
+        throw new IllegalArgumentException(s"Expected json.Arr")
```
Christewart, line 1708: same here, would be nice to put the thing we got (`v: Value`) in the error message.

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -862,6 +863,58 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
           }
       }
 
+    case ServerCommand("exportseed", arr) =>
+      withValidServerCommand(ExportSeed.fromJsArr(arr)) {
+        case ExportSeed(walletNameOpt, passwordOpt) =>
+          complete {
+            val seedPath = kmConf.seedFolder.resolve(
```
Christewart: This pattern is used in a ton of places, can we refactor this to a helper method somewhere?

app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala
```diff
@@ -862,6 +863,58 @@ case class WalletRoutes(wallet: AnyDLCHDWalletApi)(implicit
...
+      withValidServerCommand(MarkSeedAsBackedUp.fromJsArr(arr)) {
+        case MarkSeedAsBackedUp(walletNameOpt, passwordOpt) =>
+          complete {
+            val seedPath = kmConf.seedFolder.resolve(
+              walletNameOpt
+                .map(_ + "-")
+                .getOrElse("") + WalletStorage.ENCRYPTED_SEED_FILE_NAME)
+
+            WalletStorage.markSeedAsBackedUp(seedPath, passwordOpt).get
+            Server.httpSuccess(ujson.Null)
```
Christewart: Why not use the `"ok"` response we use elsewhere?

key-manager/src/main/scala/org/bitcoins/keymanager/WalletStorage.scala
```diff
@@ -364,6 +405,63 @@ object WalletStorage extends KeyManagerLogger {
...
+              decrypted.backupTimeOpt match {
+                case None =>
+                  val toWrite = (passphraseOpt match {
+                    case Some(pass) =>
+                      decrypted.encrypt(pass)
+                    case None =>
+                      decrypted
+                  }).withBackupTime(Instant.now())
+                  val _ = writeSeedState(seedPath, toWrite)
+                case Some(_) => () // ignore
```
Christewart: Maybe a 

```
logger.warn(s"Seed has already been backed up, ignorning new request to mark it as backed up")
``` 

here that the seed has already been backup?

Christewart (changes requested):

Christewart (changes requested):
The `cli` project needs to be updated, it doesn't support the latest rpcs

Screenshot from 2022-06-01 09-28-51

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
rorp: That's a password.

Christewart: Ah, good point.

rorp: No need to log a password even if it's malformed

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
rorp: Same

Christewart:
This does not work with an encrypted seed on  06eaa49793e3c030a24ae623d50724a0d3595f21 This is the error that i get when trying with local encrypted wallet

Screenshot from 2022-06-06 07-37-04

Christewart:
It doesn't seem like the `importseed` rpc is working, or i am using it incorrectly. Here is what i tried

Screen Shot 2022-06-06 at 8 02 03 AM

It appears the key error message here is

Screen Shot 2022-06-06 at 8 03 09 AM

>requirement failed: Number of words must be one of the following: 12, 15, 18, 21, 24, got: 1

rorp:
> It doesn't seem like the `importseed` rpc is working, or i am using it incorrectly. Here is what i tried

This PR didn't touch `importseed` that means it was broken before.

rorp:
> This does not work with an encrypted seed on 

It cannot decrypt the seed because the password is not specified

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1598,4 +1618,47 @@ trait ServerJsonModels {
...
+        throw new IllegalArgumentException("password must be a string or null")
+    }
+  }
+
+  def jsToWalletNameAndPassword(
+      js: Value): (Option[String], Option[AesPassword]) = {
+    js match {
+      case Arr(arr) =>
+        arr.toList match {
+          case walletNameJs :: passJs :: Nil =>
```
Christewart, line 1697: This means the wallet name has to be defined to decrypt. I have a wallet locally that does not have a wallet name, but _does_ have a password.

Christewart:
needs rebase

docs/applications/server.md
```diff
@@ -223,13 +223,22 @@ the `-p 9999:9999` port mapping on the docker container to adjust for this.
...
+     - `words` - Mnemonic seed words, space separated
+     - `passphrase` - Passphrase to encrypt this seed with
+ - `importxprv` `walletname` `xprv` `passphrase` - Imports a mnemonic seed as a new seed file
+     - `walletname` - Name to associate with this seed
+     - `xprv` - base58 encoded extended private key
+     - `passphrase` - Passphrase to encrypt this seed with
+ - `exportseed` `walletname` `passphrase` - Exports the mnemonic seed phrase
     - `walletname` - Name to associate with this seed
-    - `words` - Mnemonic seed words, space separated
     - `passphrase` - Passphrase to encrypt this seed with
```
user411: encrypt -> decrypt

docs/applications/server.md
```diff
@@ -223,13 +223,22 @@ the `-p 9999:9999` port mapping on the docker container to adjust for this.
...
-- `importxprv` `walletname` `xprv` `passphrase` - Imports a mnemonic seed as a new seed file
+ - `markseedasbackedup` `walletname` `passphrase` - Marks the seed as backed up. It prevents `exportseed` from returning the mnemonic phrase
+    - `walletname` - Name to associate with this seed
+    - `passphrase` - Passphrase to encrypt this seed with
+ - `getseedbackuptime` `walletname` `passphrase` - Returns time when the seed was backed up
     - `walletname` - Name to associate with this seed
-    - `xprv` - base58 encoded extended private key
     - `passphrase` - Passphrase to encrypt this seed with
```
user411: encrypt -> decrypt

user411:
When I call the `exportseed` endpoint with no parameters, I am getting "org.bitcoins.keymanager.ReadMnemonicError$DecryptionError$: Could not decrypt mnemonic!"

Christewart (approved):

PR #4364, opened 2022-06-03
Have Satoshis extend Numeric

benthecarman:
Should allow us to do things call things like `.sum` on a `Vector[Satoshis]` without having to convert to a `CurrencyUnit`.

I was going to do it for `Bitcoins` but it felt messy because there are things like `toInt` and `toLong`, which for `CurrencyUnit` convert to `Satoshis` but for `Bitcoins` we would probably wouldn't want to do that, so it would likely just end up with creating foot-guns.

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/currency/CurrencyUnitTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/currency/CurrencyUnitTest.scala
@@ -229,4 +229,11 @@ class CurrencyUnitTest extends BitcoinSUnitTest {
     val btc = Bitcoins(12.3456789)
     assert(btc.toString == "12.34567890 BTC")
   }
+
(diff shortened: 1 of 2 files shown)


Christewart:
> Should allow us to do things call things like `.sum` on a `Vector[Satoshis]` without having to convert to a `CurrencyUnit`.

I don't think this does. This is the test case i tried to write but it does not compile

```scala
  it must "sum" in {
    import org.bitcoins.core.currency.Satoshis._
    val sats: Vector[Satoshis] = Vector.fill(10)(Satoshis.one)
    val sum = sats.sum
    assert(sum == Satoshis(10))
  }
```

benthecarman:
> I don't think this does. This is the test case i tried to write but it does not compile

Added a test case, needed to set the implict because of conflicting imports

Christewart (approved):

PR #4376, opened 2022-06-09
Fix outPointToChannelPoint to use correct endianness

benthecarman:


Christewart (approved):
Would be nice to have unit tests to dect

PR #4384, opened 2022-06-13
Remove carve out for protoc on aarch64 now that the akka grpc plugin has been updated

Christewart:
Reverts #3013 as this is now supported natively.

Would like confirmation from @GreyMcCarthy that his intellij can compile now with this change


GreyMcCarthy:
protoc did not come pre installed on my mac after installed with brew still didn't work, but after using this pr it worked.

PR #4387, opened 2022-06-13
Added data structure for x-only public keys with undetermined parity

nkohen:
This is needed for Taproot script spending verification


crypto/src/main/scala/org/bitcoins/crypto/ECKey.scala
```diff
@@ -392,6 +392,10 @@ case class ECPublicKey(private val _bytes: ByteVector)
   def tweakMultiply(tweak: FieldElement): ECPublicKey = {
     CryptoUtil.tweakMultiply(this, tweak)
   }
+
+  def toXOnly: XOnlyPubKey = XOnlyPubKey(bytes.drop(1))
```
benthecarman, line 396: nit: use `.tail`

benthecarman (approved):

Christewart (approved):

PR #4388, opened 2022-06-13
2022 06 13 taprootspk xonlypubkey

Christewart:
Adds ontop of #4387 

Modifies `TaprootScriptPubKey.pubkey` to return `XOnlyPubKey`

Also adds `ECPrivateKey.toXOnly`

The equivalent data structure in bitcoin core

https://github.com/bitcoin/bitcoin/blob/66e3b16b8b1033414f843058f360e22b725d89c5/src/script/standard.h#L119

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/protocol/script/ScriptPubKeyTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/protocol/script/ScriptPubKeyTest.scala
@@ -41,7 +41,7 @@ class ScriptPubKeyTest extends BitcoinSUnitTest {
   }
 
   it must "construct valid witness spk v1 for taproot" in {
-    val pubKey = CryptoGenerators.schnorrPublicKey.sample.get.toXOnly
+    val pubKey = CryptoGenerators.xOnlyPubKey.sample.get
     val witSPKV1 = TaprootScriptPubKey.fromPubKey(pubKey)
     assert(witSPKV1.pubKey == pubKey)
   }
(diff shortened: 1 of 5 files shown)


core-test/src/test/scala/org/bitcoins/core/protocol/script/ScriptPubKeyTest.scala
```diff
@@ -41,7 +41,7 @@ class ScriptPubKeyTest extends BitcoinSUnitTest {
   }
 
   it must "construct valid witness spk v1 for taproot" in {
-    val pubKey = CryptoGenerators.schnorrPublicKey.sample.get
+    val pubKey = CryptoGenerators.schnorrPublicKey.sample.get.toXOnly
```
benthecarman: Should do `CryptoGenerators.xOnlyPubKey.sampleSome`, as this would only test valid `TaprootScriptPubKey`s with known parity AFIAK

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptPubKey.scala
```diff
@@ -1358,13 +1358,13 @@ object TaprootScriptPubKey extends ScriptFactory[TaprootScriptPubKey] {
                 s"Given asm was not a P2WSHWitnessSPKV0, got $asm")
   }
 
-  def apply(schnorrPubKey: SchnorrPublicKey): TaprootScriptPubKey = {
+  def apply(schnorrPubKey: XOnlyPubKey): TaprootScriptPubKey = {
```
benthecarman: should rename variable

benthecarman: Also would be nice to still have the ability to create from a `SchnorrPublicKey`

nkohen (commented):
Agree with Ben's comments, ow looks good and hopefully unblocks validation code

benthecarman (approved):

PR #4391, opened 2022-06-14
ECPublicKey.tweakMultiply -> multiply and Added CryptoParams.getG

nkohen:


crypto/src/main/scala/org/bitcoins/crypto/FieldElement.scala
```diff
@@ -40,5 +40,5 @@ object FieldElement
   private val G: ECPublicKey = ECPublicKey(
```
Christewart: Where does this `G` come from? Since a field isn't a group, does it have its own generator? If so, what field is this a part of ?

Naively I thought this would be the same as `CryptoParams.getG` but it is not

nkohen: So actually it is the same, I was just using decompressed in one spot and compressed here, decompressed is more efficient so I've changed this to use that as well.

nkohen: In fact, now that this is in `CryptoParams`, I've just delegated to there and deleted this `val` altogether

Christewart (approved):

PR #4392, opened 2022-06-14
Fix lnurl module name to have bitcoin-s prefix

benthecarman:
Something missed in my original PR.


Christewart:
Nice! Right before the 1.9.2 release

PR #4406, opened 2022-06-19
Fix download lnd command

benthecarman:
Before the check to see if it was already downloaded was looking at the wrong directory


Christewart:
Nice catch, hopefully this speeds up builds

PR #4408, opened 2022-06-20
Find and switch peers

solo1g:
Should do the following:
* Find a peer to sync with
* Try syncing with it
* If any issues with that peer (disconnection, too many connection drops, ignored queries etc), then find and use another.

### Key changes
* Added a `PeerFinder` which deals with peer discovery. Moved relevant parts from `PeerManager` to `PeerFinder`.
* Added functionality to query peers and get notified if the query times out. Used to switch peers in case the current one does not work. Adds timeouts to `Initializing` and a new state `Waiting` in `PeerMessageReceiver`.
* Changed `InitializeDisconnect` state to now disconnect when connecting too. Added `StopReconnect` state to `PeerMessageReceiver` to stop reconnect from our side when not connected.
* Some fixes dealing with edge cases in `P2PClient` and `PeerMessageReceiver`.


solo1g:
So this should be working as intended now. I will leave comments on all the changes before taking the draft off. Meanwhile, this could be tested by running with the following config
```
peers = []
maxConnectedPeers = 8
enable-peer-discovery = true
use-default-peers = false
```

core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
```diff
@@ -166,13 +166,27 @@ object GetDataMessage extends Factory[GetDataMessage] {
     GetDataMessage(Seq(inventory))
 }
 
+sealed trait ExpectsResponse {
```
solo1g, line 169: So I needed a way to if the peer failed to respond to a query for eg `getheaders` etc that is used in sync, so that I can switch to a different peer and continue sync from there. For `NetworkPayload` that extends this, an `ExpectReponse` message is sent to the `P2PClient` right after sending the message. The message initiates a timer in `PeerMessageReceiver` that notifies if the message timed out.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
```diff
@@ -37,81 +40,123 @@ class NeutrinoNodeTest extends NodeTestWithCachedBitcoindPair {
...
-
-      val initAssertion = isInitializedF.map(assert(_))
-
-      val disconnF = for {
-        _ <- initAssertion
-        _ <- node.stop()
-        f <- isAllDisconnectedF(node)
-      } yield f
-      disconnF.map(assert(_))
+      def peerManager = node.peerManager
```
solo1g, line 46: Did away with using indexes and added check to ensure we have 2 peers.

node-test/src/test/scala/org/bitcoins/node/UnsyncedNeutrinoNodeTest.scala
```diff
@@ -0,0 +1,110 @@
...
+        bitcoinds.toVector)(system, getFreshConfig)
+      f <- outcome.toFuture
+    } yield f
+    new FutureOutcome(outcomeF)
+  }
+
+  behavior of "Neutrino Node"
+
+  //assuming both peers are in sync with each other
+  it must "be able to sync" in { nodeConnectedWithBitcoinds =>
```
solo1g, line 36: Ig this test is a bit redundant since creating the fixture for `NeutrinoNodeTest` does have an `awaitSync`.

node-test/src/test/scala/org/bitcoins/node/UnsyncedNeutrinoNodeTest.scala
```diff
@@ -0,0 +1,110 @@
...
+      cnt5 <- bitcoinds.head.getBestBlockHash
+      cnt6 <- bitcoinds.head.getBestBlockHash
+    } yield {
+      assert(cnt == cnt2 && cnt == cnt3)
+      assert(cnt4 == cnt5 && cnt4 == cnt6 && cnt4 == bestHash)
+    }
+  }
+
+  //todo: so the disconnection should rather originate from the remote client rather than our side
+  ignore must "sync with another second peer if the first one is disconnected" in {
```
solo1g: Reconnect after stopping was throwing some exception about a closed port. Looking into it, didn't want to block review on minor stuff so marked to ignore test at the moment,

node/src/main/resources/postgresql/node/migration/V3__peer_table.sql
```diff
@@ -1 +1 @@
-create table if not exists "peers" ("address" VARCHAR NOT NULL PRIMARY KEY UNIQUE,"port" INT NOT NULL,"last_seen" TIMESTAMP,"first_seen" TIMESTAMP, "network_id" INT)
+create table if not exists "peers" ("address" VARCHAR NOT NULL, "port" INT NOT NULL, "last_seen" TIMESTAMP,"first_seen" TIMESTAMP, "network_id" INT, "service_bytes" VARCHAR, constraint pk_peers PRIMARY KEY (address,port));
```
solo1g: Now treats peers with same address but different ports as distinct. Needed for multiple nodes on localhost with different ports.

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -26,7 +26,7 @@ case class NeutrinoNode(
     nodeConfig: NodeAppConfig,
     chainConfig: ChainAppConfig,
     actorSystem: ActorSystem,
-    configPeersOverride: Vector[Peer] = Vector.empty)
+    paramPeers: Vector[Peer] = Vector.empty)
```
solo1g, line 30: So we don't override the config ones anymore, rather use both. This was one of the suggestion earlier so added in this.

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -136,8 +146,11 @@ case class NeutrinoNode(
       bestFilterHeader: CompactFilterHeaderDb,
       chainApi: ChainApi,
       bestFilterOpt: Option[CompactFilterDb]): Future[Unit] = {
+    val syncPeer = dataMessageHandler.syncPeer.getOrElse(
+      throw new RuntimeException("Sync peer not set"))
+    val syncPeerMsgSender = peerManager.peerData(syncPeer).peerMessageSender
     val sendCompactFilterHeaderMsgF = {
-      peerManager.randomPeerMsgSenderWithCompactFilters
+      syncPeerMsgSender
```
solo1g, line 153: So unless the peers are all ideal, can't really use a random one, besides without parallel sync, not much point in switching. The previous change too was by me, so fixed it now. Would add this later with properly working parallel sync.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -81,37 +79,8 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
...
-        maxTries = 20,
-        interval = 1.second
-      ) //20 seconds
-    isInitializedF.failed
-      .foreach { err =>
-        logger.error(s"Failed to initialize with peer=$peer with err=$err")
-      }
-    isInitializedF
+  def send(msg: NetworkPayload, peer: Peer): Future[Unit] = {
+    peerManager.peerData(peer).peerMessageSender.sendMsg(msg)
```
solo1g, line 83: Alll of this is now in `PeerManager`

node/src/main/scala/org/bitcoins/node/PeerData.scala
```diff
@@ -26,7 +28,9 @@ case class PeerData(
       context = system,
       peer = peer,
       peerMessageReceiver = peerMessageReceiver,
-      onReconnect = node.sync
+      onReconnect = node.peerManager.onReconnect,
+      onStop = node.peerManager.onP2PClientStopped,
+      maxReconnectionTries = 4
```
solo1g, line 33: Changed max attempts to 4 as this would roughly be the time that we spend on each peer waiting for it to initialize. Ig all of these related constant need to be put together.

node/src/main/scala/org/bitcoins/node/PeerData.scala
```diff
@@ -41,4 +45,15 @@ case class PeerData(
   def setServiceIdentifier(serviceIdentifier: ServiceIdentifier): Unit = {
     _serviceIdentifier = Some(serviceIdentifier)
   }
+
+  private var lastTimedOut: Long = 0
+
+  def updateLastFailureTime(): Unit = {
+    lastTimedOut = System.currentTimeMillis()
+  }
+
```
solo1g, line 54: So that we don't randomly retry the peers that failed just now.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,209 @@
...
+case class PeerFinder(paramPeers: Vector[Peer], skipPeers: () => Vector[Peer])(
+    implicit
+    node: Node,
+    ec: ExecutionContext,
+    system: ActorSystem,
+    nodeAppConfig: NodeAppConfig)
+    extends P2PLogger {
+
+  /** Returns peers by querying each dns seed once. These will be IPv4 addresses. */
+  def getPeersFromDnsSeeds: Vector[Peer] = {
```
solo1g: Moved this from manager to a new `Finder` to separate things.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,330 @@
...
-    if (_peerData.contains(peer)) {
-      val connF = peerData(peer).peerMessageSender.isConnected()
-      val disconnectF = connF.map { conn =>
-        if (conn) peerData(peer).peerMessageSender.disconnect()
-        else Future.unit
+  def removePeer(peer: Peer): Unit = {
+    logger.debug(s"Removing persistent peer $peer")
+    val client = peerData(peer).client
+    _peerData.remove(peer)
+    //so we need to remove if from the map for connected peers so no more request could be sent to it but we before
```
solo1g, line 152: When we don't want a peer as one of our persistent ones, we remove it from the `peerData` map, but still keep it around to check if the client actor for it actually stopped.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -65,19 +69,23 @@ import scala.util._
...
-  private var reconnectHandlerOpt: Option[() => Future[Unit]] = None
-
-  private val maxReconnectionTries = 16
+  private var reconnectHandlerOpt: Option[Peer => Future[Unit]] = None
 
+  //total
   private var reconnectionTry = 0
 
+  //from the last initialization
+  private var curReconnectionTry = 0
```
solo1g, line 87: Keeping both the total and the current since last initialization. Whenever we reconnect successfully only the current is set to 0 (and used for exponential delays in connection). While the total is used for comparing with max tries.

This fixes an issue where a peer intializes but then disconnects us again and again. Earlier this would be an infinite loop.

node/src/main/scala/org/bitcoins/node/networking/peer/ControlMessageHandler.scala
```diff
@@ -84,19 +80,39 @@ case class ControlMessageHandler(node: Node)(implicit ec: ExecutionContext)
...
+        val bytes = addr.bytes
+        val port = addr.port.toInt
+        val inetAddress =
+          NetworkUtil.parseInetSocketAddress(bytes, port)
+        val peer = Peer.fromSocket(socket = inetAddress,
+                                   socks5ProxyParams =
+                                     node.nodeAppConfig.socks5ProxyParams)
+        addr match {
+          case IPv4AddrV2Message(_, _, _, _) | IPv6AddrV2Message(_, _, _, _) =>
+            //todo: ensure finder prioritises ones with filters
```
solo1g: Missed my own todo :sweat_smile: . Shouldn't affect the remaining review. Right now even though can get peers that possibly support filters, the finder does not prioritizes trying them. Will fix this.
Update: This is fixed now.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -60,9 +66,19 @@ class PeerMessageReceiver(
     * peer initialized a disconnection from us
     */
   private[networking] def initializeDisconnect(): PeerMessageReceiver = {
+    logger.debug(s"Initializing disconnect from $peer")
     state match {
-      case bad @ (_: Disconnected | _: InitializedDisconnectDone |
-          Preconnection) =>
+      case good @ (_: Disconnected) =>
```
solo1g, line 70: Case where we disconnected right before sending an intialize disconnect. In this case just don't try to reconnect.

project/CommonSettings.scala
```diff
@@ -141,6 +141,7 @@ object CommonSettings {
     "-Xmax-classfile-name",
     "128",
     "-Ywarn-unused",
+    "-Ywarn-macros:after",
```
solo1g: Added somewhere down the line. Will see what breaks when removing this :sweat_smile:

core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
Christewart: Was this common when testing? This is an interesting blog post about the bitcoin p2p network possibly slowing down: https://blog.lopp.net/is-bitcoin-network-slowing-down/

it details the logic bitcoin core has about detecting peers that are stalling, probably worth a read. It seems the behavior detailed there is what you experienced as well?

node-test/src/test/scala/org/bitcoins/node/UnsyncedNeutrinoNodeTest.scala
Christewart: I think its still valuable because if you introduce a regression, you can see that this specific test case is failing around syncing and then can deduce that is probably what is causing the rest of the test suites to fail.

Essentially it allows you to target the root cause of the regression faster, IMO.

node-test/src/test/scala/org/bitcoins/node/UnsyncedNeutrinoNodeTest.scala
```diff
@@ -0,0 +1,110 @@
...
+    val remotesInSync: Future[Assertion] = for {
+      h1 <- bitcoinds(0).getBestBlockHash
+      h2 <- bitcoinds(1).getBestBlockHash
+    } yield assert(h1 == h2)
+
+    def hashF: Future[DoubleSha256DigestBE] = bitcoinds.head.getNewAddress
+      .flatMap(bitcoinds.head.generateToAddress(1, _))
+      .map(_.head)
+
+    def awaitAllSync(
```
Christewart: Any reason not to move this to `NodeTestUtil`? seems like a useful utility.

node-test/src/test/scala/org/bitcoins/node/networking/P2PClientTest.scala
```diff
@@ -151,56 +154,82 @@ class P2PClientTest
...
       try2
     }
   }
 
-  /** Helper method to connect to the
-    * remote node and bind our local
-    * connection to the specified port
-    */
-  def connectAndDisconnect(peer: Peer): Future[Assertion] = {
+  it must "close actor on disconnect" in {
```
Christewart, line 182: :tada: , good idea to add this test.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,257 @@
...
+    assert(hasPeer(peer), "finder.getData called without any data on peer")
+    _peerData(peer)
+  }
+
+  def addToTry(peers: Vector[Peer], priority: Int = 0): Unit = {
+    _peersToTry.pushAll(peers, priority)
+  }
+}
+
+case class PeerStack() {
```
solo1g, line 219: I wanted a stack like structure that pops based on priority and then recency. PriorityQueue would have been good but needed to remove from the least priority ones too in case a higher priority values is pushed and limit is reached. So settled for a set finally.

node/src/main/resources/postgresql/node/migration/V3__peer_table.sql
Christewart: Won't this break migrations? I believe we have to add a new migration called `V4` and then add the new `service_bytes` column. Modifying existing deployed migrations will cause an exception on startup if the user had started with the `V3__peer_table.sql` migration.

node/src/main/resources/sqlite/node/migration/V3__peer_table.sql
```diff
@@ -1 +1 @@
-create table if not exists "peers" ("address" VARCHAR NOT NULL PRIMARY KEY UNIQUE, "port" INT NOT NULL, "last_seen" TIMESTAMP,"first_seen" TIMESTAMP, "network_id" INT)
\ No newline at end of file
+create table if not exists "peers" ("address" VARCHAR NOT NULL, "port" INT NOT NULL, "last_seen" TIMESTAMP,"first_seen" TIMESTAMP, "network_id" INT, "service_bytes" VARCHAR, constraint pk_peers PRIMARY KEY (address,port));
```
Christewart: same here wrt to migrations

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -36,11 +36,11 @@ case class NeutrinoNode(
 
   implicit override def nodeAppConfig: NodeAppConfig = nodeConfig
 
-  override def chainAppConfig: ChainAppConfig = chainConfig
+  implicit override def chainAppConfig: ChainAppConfig = chainConfig
 
-  val controlMessageHandler: ControlMessageHandler = ControlMessageHandler(this)
+  implicit def node: NeutrinoNode = this
```
Christewart: Lets not get crazy with `implicit`. Its easy to overuse `implicit` and then your code becomes very unreadable quickly.

Our general policy is its ok to use `implicit` for things like `ActorSystem`, `AppConfig` (generally any configuration related code), but not for normal data structures like `NeutrinoNode` or `Wallet` etc.

node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -196,16 +136,21 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
     * @return
     */
   def sync(): Future[Unit] = {
+    logger.debug("Started Node.sync")
     val blockchainsF =
       BlockHeaderDAO()(executionContext, chainAppConfig).getBlockchains()
+    val syncPeer = peerManager.peers.head
```
Christewart: IIUC, this means we are always choosing the first peer to sync with (`.head`), is this what we want?

node/src/main/scala/org/bitcoins/node/PeerData.scala
Christewart: Can you add this comment on github as a scaladoc, and can you add a scaladoc to `isDeferred` as well with the intention of that method

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,257 @@
...
+import org.bitcoins.node.models.{Peer, PeerDAO, PeerDb}
+
+import java.net.{InetAddress, UnknownHostException}
+import scala.collection.mutable
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.concurrent.{ExecutionContext, Future}
+import scala.io.Source
+import scala.util.Random
+
+object PeerSource extends Enumeration {
```
Christewart: can we make this an ADT in Scala ? That way we can pattern match. 

IIUC, this would be

```scala
sealed trait PeerSource
object PeerSource {
  case object Config extends PeerSource
  ...
  case object Param extends PeerSource
}
```
you might want to `extend StringFactory` too so you can use `fromString()`

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  //for the peers we try
+  private val _peerData: mutable.Map[Peer, PeerData] = mutable.Map.empty
+
+  private val _peersToTry: PeerStack = PeerStack()
+
+  val maxPeerSearchCount: Int = 1000
+
+  val timeoutForTrying: FiniteDuration = 12.second
```
Christewart: maybe configurable via `bitcoin-s.conf`?

core/src/main/scala/org/bitcoins/core/p2p/NetworkPayload.scala
solo1g: Interesting :eyes: .
Not very common but did see this a few times. Instead of being connected and not responding, what I found more common was for the peer to straight up disconnect us after initialization. Reconnection would be successful followed by the same behavior.

Current behavior is for timeouts to complete instantly and requery another peer (same if that's all we have) in case of a disconnection and wait for the whole duration in case its still connected. This covers both cases.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,257 @@
...
+          val peers = (for { _ <- 1 to 32 } yield _peersToTry.pop()).distinct
+            .filterNot(skipPeers().contains(_))
+
+          logger.debug(s"Trying next set of peers $peers")
+          peers.foreach(tryPeer)
+        }
+      }
+    }
+
+  def start: Future[Unit] = {
```
Christewart: nit: maybe use `StaratStopAsync` interface?

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -0,0 +1,257 @@
...
+      } yield {
+        _peersToTry.pushAll(getPeersFromDnsSeeds)
+        _peersToTry.pushAll(getPeersFromResources)
+        _peersToTry.pushAll(dbNonCf)
+        _peersToTry.pushAll(dbCf, priority = 1)
+        peerConnectionScheduler //start scheduler
+        ()
+      }
+    } else {
+      logger.debug("Peer discovery disabled")
```
Christewart: probably worth a `info` or maybe even `warn` log

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
     nodeAppConfig: NodeAppConfig)
     extends P2PLogger {
 
-  private val _peerData: mutable.Map[Peer, PeerData] =
-    mutable.Map.empty
+  private val _peerData: mutable.Map[Peer, PeerData] = mutable.Map.empty
+  private val _waitingForDeletion: mutable.Set[Peer] = mutable.Set.empty
 
-  def peerData: Map[Peer, PeerData] = _peerData.toMap
+  def waitingForDeletion: Set[Peer] = _waitingForDeletion.toSet
```
Christewart, line 39: scaladoc on this would be nice, what are we deleting it from?

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
+
+  val finder: PeerFinder = PeerFinder(paramPeers, skipPeers = () => peers)
+
+  def connectedPeerCount: Int = _peerData.size
+
+  def addPeer(peer: Peer): Future[Unit] = {
+    require(finder.hasPeer(peer), "Unknown peer marked as usable")
```
Christewart: nit, add the peer to the error message, i.e.

```scala
require(finder.hasPeer(peer), "Unknown peer marked as usable, peer=$peer")
```

that way we can see the input that caused the exception.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
+    logger.debug(s"Waiting for peer connection. ${_peerData.keys}")
+    val ret = AsyncUtil
+      .retryUntilSatisfied({
+                             _peerData.exists(x => f(x._2.serviceIdentifier))
+                           },
+                           interval = 1.seconds,
+                           maxTries = timeout.getSeconds.toInt)
+      .recover {
+        case _: AsyncUtil.RpcRetryException =>
+          throw new RuntimeException("No supported peers found!")
```
Christewart: Can we add the service identifier to this exception somehow?

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
+    _waitingForDeletion.add(peer)
+    //now send request to stop actor which will be completed some time in future
+    client.close()
+  }
+
+  def isReconnection(peer: Peer): Boolean = {
+    peerData.contains(peer)
+  }
+
+  def start: Future[Unit] = {
```
Christewart: same comment wrt to `StartStopAsync` interface

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
+    logger.debug(s"Starting PeerManager")
+    finder.start.map(_ => logger.info("Done starting PeerManager"))
+  }
+
+  def peerData: Map[Peer, PeerData] = _peerData.toMap
+
+  def stop: Future[Unit] = {
+    logger.debug(s"Stopping PeerManager")
```
Christewart: `info` IMO

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -1,199 +1,325 @@
...
+
+    val managerStopF = AsyncUtil.retryUntilSatisfied(
+      _peerData.isEmpty && waitingForDeletion.isEmpty,
+      interval = 1.seconds,
+      maxTries = 10)
+
+    for {
+      _ <- finderStopF
+      _ <- managerStopF
+    } yield ()
```
Christewart: another log indicating that it fully stopped 
```scala
   for {
      _ <- finderStopF
      _ <- managerStopF
    } yield {
 logger.info(s"Stopped peer manager")
}
```
a timer might be useful here to to see how long it took to stop

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
```diff
@@ -143,6 +143,25 @@ case class NodeAppConfig(baseDatadir: Path, configOverrides: Vector[Config])(
...
+  lazy val peerDiscoveryTimeout: Duration = {
+    if (config.hasPath("bitcoin-s.node.peer-discovery-timeout"))
+      config.getDuration("bitcoin-s.node.peer-discovery-timeout")
+    else Duration.ofMinutes(10)
+  }
+
+  lazy val useDefaultPeers: Boolean = {
+    if (config.hasPath("bitcoin-s.node.use-default-peers"))
+      config.getBoolean("bitcoin-s.node.use-default-peers")
+    else true
```
Christewart, line 163: so by default if we merge this PR, we will obtain the original behavior of the node because this is `true`, correct? 

When we want to "turn on" this PR by default, we will adjust this setting to `false`?

Christewart (changes requested):
Looking good, i'll test this out locally after the migrations are fixed.

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
solo1g: This PR being "on" is independent of this particular value. You have this `true` you get the suredbits neutrino node as a peer, if not you don't.
This was added for two reasons, one to turn off the default peers introduced in #4404 as so that it does not connect to that. Second is in ci which runs with regtest mode, querying for config peers throws an exception so this is a way around that too. https://github.com/bitcoin-s/bitcoin-s/pull/4404/files#diff-d15b5841b444ff367c6ef6c9a6089315aee633ea64f95a97ab6b93ca6b52b496

node/src/main/resources/sqlite/node/migration/V4__new_peer_table.sql
```diff
@@ -0,0 +1,3 @@
+drop table "peers";
```
solo1g, line 1: Cannot know the `service_bytes` value for existing peers so dropping that table seems the only option.

solo1g:
@Christewart fixed the migrations so should be good for testing now.

node/src/main/scala/org/bitcoins/node/Node.scala
solo1g: Good catch. No, we don't want this. In any case, this `sync` function in `Node` is overriden by `NeutrinoNode` so there's no place where this is used right now.

node/src/main/scala/org/bitcoins/node/PeerData.scala
solo1g: Changed the name to `hasFailedRecently` to make the behavior clear and added a scaladoc.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
solo1g: Removed this for now as it's not used anywhere at the moment. Was doing some changes on top of this but they are not in this PR, will add when in another PR with those changes done.

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
Christewart: FWIW, i think we can remove #4404 when this PR gets merged.

The problem i had on #4404 was in docker environments and resolving the configuration relative to our `bitcoin-s.network` inside of `docker-application.conf`

There was no way to configure a hard coded peer based on network in `docker-application.conf` and I needed a way to get a user unblocked, thus #4404

Christewart: It would be helpful to have some documentation on the configuration options you added and what they do. Can you add that to the example configuration file we have in `configuration.md` ? 

https://bitcoin-s.org/docs/config/configuration#example-configuration-file

node/src/main/scala/org/bitcoins/node/NeutrinoNode.scala
```diff
@@ -78,13 +74,25 @@ case class NeutrinoNode(
       BlockHeaderDAO()(executionContext, chainConfig).getBlockchains()
     for {
       chainApi <- chainApiFromDb()
+      _ <- chainApi.getBestBlockHash()
+
+      syncPeer <- peerManager.randomPeerWithService(_.nodeCompactFilters)
+      _ = logger.info(s"Starting sync with $syncPeer")
```
Christewart: Lets change this log to `Syncing with $syncPeer`, its a little confusing to look at logs and see that we keep `Starting sync with peer` and I immediately thought, why does it keep restarting?! 

It turns out it is just querying data? Maybe thats better term? :man_shrugging: 

Screenshot from 2022-06-24 07-47-06

Christewart:
I was able to sync correctly on c5f0f9cf0376e5ac4a6028382d544dd4f6794a92 with the default out of the box settings :tada: . I'm going to start playing with the various settings introduced in this PR but wanted to document that things do seem to work out of the box.

Screenshot from 2022-06-24 08-41-51

Christewart:
Sync worked correctly for me on c5f0f9cf0376e5ac4a6028382d544dd4f6794a92

with these settings 

```
bitcoin-s.node.use-default-peers = false
bitcoin-s.node.enable-peer-discovery = true
```

Christewart:
Things work on c5f0f9cf0376e5ac4a6028382d544dd4f6794a92 for me with IBD and running a few hours with the following settings

```
bitcoin-s.node.use-default-peers = false
bitcoin-s.node.enable-peer-discovery = true
bitcoin-s.node.maxConnectedPeers = 8
```

IIUC what the logs means, it should be reworded to something about disconnecting. Also does this mean _they_ disconnected from us or does it mean we disconnected from them?

>2022-06-24T16:49:33UTC WARN [PeerManager] onP2PClientStopped called for unknown Peer(51.38.11.15:8333)

solo1g:
> 2022-06-24T16:49:33UTC WARN [PeerManager] onP2PClientStopped called for unknown Peer(51.38.11.15:8333)

Was this common? Ideally, you should not be seeing this. This basically means somewhere down the line a peer got "lost" from the `PeerManager` before it's ciient actor was stopped.

Edit: Fixed this. This happened when one of the peers while it still is initalizing is received again in an addr message and another actor is created for it while still having only one entry in peerData.

node/src/main/scala/org/bitcoins/node/config/NodeAppConfig.scala
solo1g: Added this and some other configs.

project/CommonSettings.scala
solo1g: Removed

PR #3769, opened 2021-10-19
Add static test vectors for Taproot

Christewart:
This does not implement anything in Taproot, it just adds the test harness for Taproot static test vectors found here: https://raw.githubusercontent.com/bitcoin-core/qa-assets/main/unit_test_data/script_assets_test.json

This PR intentionally does not add these to scalajs yet, i just want to get things working on the JVM before we cross that bridge

PR #4409, opened 2022-06-21
Implement BIP341

Christewart:
This PR (partially) implements [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki).

This pulls over work on #3769 . The sighash hex in `TransactionSignatureSerializerTest` are obtained from bitcoin core on this branch: https://github.com/christewart/bitcoin/tree/2022-06-12-taproot-sig-serialization

This PR does not exhaustively test all possible combinations of SIGHASH flags, rather it only tests sighash flags encountered in the first ~10 static test vectors in [`script_assets_test.json`](https://github.com/bitcoin-s/bitcoin-s/pull/3769/files#diff-a40719b6d8b23275dfcefc0dcb2a60500facae6fba1dae6bc1621e177b46c26f). 

Since this PR does not exhaustively test everything, there may be future bug fixes made to this. 

I wanted to get this PR opened and hopefully merged as #3769 is growing in size and this seems like an independent piece of work we could get merged before merging all the interpreter logic that will have to be added in #3769.

I also believe this piece of work should be enough to get wallet signing operations working for basic sighash schemes with taproot if someone wants to pick that up.


core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -74,73 +118,107 @@ trait TransactionSignatureChecker {
...
-      val sigWithoutHashType = stripHashType(signature)
-      val isValid = pubKey.verify(hashForSignature, sigWithoutHashType)
-      if (isValid) SignatureValidationSuccess
-      else
-        nullFailCheck(Seq(signature),
-                      SignatureValidationErrorIncorrectSignatures,
-                      flags)
+  def checkSigTapscript(
+      txSignatureComponent: TxSigComponent,
+      pubKey: XOnlyPubKey,
```
nkohen: I think this is supposed to be a `SchnorrPublicKey`

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -47,32 +35,52 @@ sealed abstract class TransactionSignatureSerializer {
...
+                              t.outputs,
+                              script,
+                              txSigComponent.sigVersion,
+                              taprootOptions)
+    }
+
   }
 
-  def serializeForSignature(
+  private def serializeForSignature(
```
nkohen: what's the reason for making this private?

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -227,6 +243,133 @@ sealed abstract class TransactionSignatureSerializer {
...
+
+        val spendType: Byte = ((extFlag << 1) + annexByte).toByte
+
+        val inputIndexBytes = inputIndex.bytes.reverse
+
+        //          spendingTransaction.version.bytes.reverse ++ outPointHash ++ sequenceHash ++
+        //            i.previousOutput.bytes ++ CompactSizeUInt.calc(scriptBytes).bytes ++
+        //            scriptBytes ++ amount.bytes ++ i.sequence.bytes.reverse ++
+        //            outputHash ++ spendingTransaction.lockTime.bytes.reverse ++ Int32(
+        //              hashType.num).bytes.reverse
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -253,8 +397,17 @@ sealed abstract class TransactionSignatureSerializer {
...
       val serializedTxForSignature =
-        serializeForSignature(txSigComponent, hashType)
-      CryptoUtil.doubleSHA256(serializedTxForSignature)
+        serializeForSignature(txSigComponent = txSigComponent,
+                              hashType = hashType,
+                              taprootOptions = taprootOptions)
+
+      val hash = txSigComponent.sigVersion match {
+        case _: SigVersionTaproot =>
+          CryptoUtil.taggedSha256(serializedTxForSignature, "TapSighash")
```
nkohen, line 396: this should probably be given a method in `CryptoRuntime.scala` so that the tag doesn't have to be re-computed every time

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -180,8 +180,11 @@ object TxSigComponent {
       case _: WitnessScriptPubKey =>
         transaction match {
           case _: NonWitnessTransaction =>
-            throw new IllegalArgumentException(
-              s"Cannot spend from segwit output ($output) with a base transaction ($transaction)")
+            /*            throw new IllegalArgumentException(
+              s"Cannot spend from segwit output ($output) with a base transaction ($transaction)")*/
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -342,12 +354,74 @@ sealed abstract class WitnessTxSigComponentRebuilt extends TxSigComponent {
...
+    outputMap: Map[TransactionOutPoint, TransactionOutput],
+    flags: Seq[ScriptFlag])
+    extends WitnessTxSigComponent {
+  require(
+    scriptPubKey.isInstanceOf[TaprootScriptPubKey],
+    s"Can only spend taproot spks with TaprootTxSigComponent, got=$scriptPubKey")
+//  require(
+//    transaction.witness(inputIndex.toInt).isInstanceOf[TaprootWitness],
+//    s"Can use taproot witnesses to spend taproot spk, got=${transaction.witness(inputIndex.toInt)}"
+//  )
```
nkohen: ping

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -403,6 +403,7 @@ trait BitcoinScriptUtil {
         //BIP143 removes requirement for calling FindAndDelete
         //https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#no-findanddelete
         scriptForChecking
+      case SigVersionTaprootKeySpend | SigVersionTapscript => scriptForChecking
```
nkohen: nit: could be merged with case above

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
Christewart: #4410

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
Christewart: Its `XOnlyPubKey` in core

https://github.com/bitcoin/bitcoin/blob/34869114a72435b9d8364385abaefed70d703fa8/src/script/interpreter.cpp#L1684

https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L1093-L1096

nkohen: To the best of my knowledge core does not have distinctions between XOnlyPublicKey, SchnorrPublicKey, and SchnorrNonce, but we do and XOnly is meant for x-coords representing arbitrary keys while SchnorrPublicKey is for those to be interpreted as having even y coordinate

Christewart: They parse it as XOnly on the first link

```c++
XOnlyPubKey pubkey{pubkey_in};
```

nkohen: In general XOnly keys on which you only call .schnorrPublicKey and never explicitly pass a parity should be SchnorrPublicKey (or SchnorrNonce in the rare case that it is an R value and not a pubkey)

nkohen: Sure, but the use XOnly as schnorr because they don't have a different type for it

Christewart: > In general XOnly keys on which you only call .schnorrPublicKey and never explicitly pass a parity should be SchnorrPublicKey (or SchnorrNonce in the rare case that it is an R value and not a pubkey)

Is this the line of code that implicitly gives even parity? 

https://github.com/bitcoin-s/bitcoin-s/blob/bf3d724ee9baafb8c4d856b475dd8844a1a352a7/crypto/src/main/scala/org/bitcoins/crypto/SchnorrPublicKey.scala#L64

If so, I think I agree and can modify, but would like to modify that line in `SchnorrPublicKey.scala` to use our new `EvenParity` type so that there are less magic values (`2` in this case)

nkohen: Yup, sounds good to me

app-commons/src/main/scala/org/bitcoins/commons/serializers/JsonWriters.scala
```diff
@@ -24,6 +24,7 @@ object JsonWriters {
 
     override def writes(hash: HashType): JsValue =
       hash match {
+        case SIGHASH_DEFAULT                => JsString("DEFAULT")
```
benthecarman, line 27: Should start using this instead of `ALL` in a lot of places I think. Probably fine to do later though

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -47,32 +35,52 @@ sealed abstract class TransactionSignatureSerializer {
...
       spendingTransaction: Transaction,
       inputIndex: UInt32,
       hashType: HashType,
-      amount: CurrencyUnit,
+      outputs: Vector[TransactionOutput],
       script: Seq[ScriptToken],
-      sigVersion: SignatureVersion): ByteVector = {
+      sigVersion: SignatureVersion,
+      taprootOptions: TaprootSerializationOptions): ByteVector = {
+    val keyVersion: Byte = 0.toByte
```
benthecarman, line 77: Shouldn't we get this from the pubkey / a param?

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -128,6 +136,9 @@ sealed abstract class TransactionSignatureSerializer {
         val sigHashBytes = Int32(hashType.num).bytes.reverse
 
         hashType match {
+          case SIGHASH_DEFAULT =>
+            sys.error(
+              s"SIGHASH_DEFAULT is only available in taproot signature serialization, got=${sigVersion}")
```
benthecarman, line 141: Correct me if I'm wrong but I thought this mapped to `SIGHASH_ALL` in non-taproot signing

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
```diff
@@ -227,6 +243,127 @@ sealed abstract class TransactionSignatureSerializer {
...
+        val outPointHash: ByteVector = if (isNotAnyoneCanPay) {
+          val prevOuts = spendingTransaction.inputs.map(_.previousOutput)
+          val bytes: ByteVector = BytesUtil.toByteVector(prevOuts)
+          CryptoUtil.sha256(bytes).bytes
+        } else if (!isNotAnyoneCanPay) {
+          val b = spendingTransaction.inputs(inputIndex.toInt).previousOutput
+          b.bytes
+        } else {
+          emptyHash.bytes
+        }
```
benthecarman: This should be unreachable?

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
```diff
@@ -342,12 +352,70 @@ sealed abstract class WitnessTxSigComponentRebuilt extends TxSigComponent {
...
+  */
+case class TaprootTxSigComponent(
+    transaction: WitnessTransaction,
+    inputIndex: UInt32,
+    outputMap: Map[TransactionOutPoint, TransactionOutput],
+    flags: Seq[ScriptFlag])
+    extends WitnessTxSigComponent {
+  require(
+    scriptPubKey.isInstanceOf[TaprootScriptPubKey],
+    s"Can only spend taproot spks with TaprootTxSigComponent, got=$scriptPubKey")
```
benthecarman, line 392: Would it instead make it better to make `WitnessTxSigComponent` to `WitnessTxSigComponent[T <: WitnessScriptPubKey]`?

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -301,6 +301,18 @@ case class TaprootScriptPath(stack: Vector[ByteVector]) extends TaprootWitness {
...
     */
   def p: XOnlyPubKey = controlBlock.p
+
+  /** As per bip341
+    *  the SHA256 of (compact_size(size of annex) || annex), where annex includes the mandatory 0x50 prefix.
+    *  @see https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#signature-validation-rules
+    */
+  def annexHashOpt: Option[Sha256Digest] = {
+
+    annexOpt.map { annex =>
```
benthecarman: nit: odd spacing

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -301,6 +301,18 @@ case class TaprootScriptPath(stack: Vector[ByteVector]) extends TaprootWitness {
   /** Let p = c[1:33] and let P = lift_x(int(p)) where lift_x and [:] are defined as in BIP340. Fail if this point is not on the curve.
     */
   def p: XOnlyPubKey = controlBlock.p
+
+  /** As per bip341
+    *  the SHA256 of (compact_size(size of annex) || annex), where annex includes the mandatory 0x50 prefix.
```
benthecarman, line 305: I dont see any checks of the `mandatory 0x50 prefix.`

core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
```diff
@@ -62,19 +62,44 @@ case object WitnessVersion0 extends WitnessVersion {
...
+            case w @ (EmptyScriptWitness | _: P2WPKHWitnessV0 |
+                _: P2WSHWitnessV0) =>
+              sys.error(
+                s"Cannot rebuild witnessv1 with a non v1 witness, got=$w")
+          }
+          rebuiltSPK
+        }
+      case _ =>
+        //witness version 1 programs need to be 32 bytes in size
+        Left(ScriptErrorWitnessProgramWrongLength)
```
benthecarman: I thought this was undefined?

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
Christewart: Maybe in the future, but not now:

https://github.com/bitcoin/bitcoin/blob/a085a554913ae8f4ed83afac830ce6dc39c9cc65/src/script/interpreter.cpp#L1492

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureSerializer.scala
Christewart: Its unclear to me how the `0x00` sighash byte is treated in a pre-taproot world. I'm guessing the static test vectors i'm running will test this eventually.

Here is the pre-taproot sighash impl https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L1568

core/src/main/scala/org/bitcoins/core/crypto/TxSigComponent.scala
Christewart: I tried doing this, this ended up being a huge diff on its own and ran into problems of having to use `WitnessTxSigComponent[_]` everywhere. I think this is the best path forward for now unless we go to Scala 3 :-)

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
Christewart: That is what `annexOpt` does. It returns `None` if the prefix is _not_ `0x50`

core/src/main/scala/org/bitcoins/core/protocol/script/WitnessVersion.scala
Christewart: This is dependent on policy. Unfortunately we don't have access to our policy flags in this spot, so what I did for now is adjust the failure type to `ScriptErrorDiscourageUpgradeableWitnessProgram` and I will have to come back to this on #3769 to make it behave correctly with the static test vectors.

benthecarman (approved):

PR #4415, opened 2022-06-22
Fixes support for bitcoind taproot descriptor wallets

benthecarman:


bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v23/BitcoindV23RpcClientTest.scala
```diff
@@ -29,24 +27,11 @@ class BitcoindV23RpcClientTest extends BitcoindFixturesFundedCachedV23 {
       }
   }
 
-  it should "be able to decode a reedem script" in {
```
benthecarman, line 32: removed this test as it doesn't work with descriptor wallets and it was just a copy from v22 tests

PR #4418, opened 2022-06-23
Musig2 Implementation

nkohen:
Implements https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki

TODO:
- [x] Add missing safety checks
- [x] Add static test vectors
- [x] Implement and test partial signature verification
- [x] Test with more signers
- [x] Implement tweaking support
- [x] Refactor types
- [x] Support js
- [ ] Test against secp256k1-zkp implementation of MuSig2

Edit: Probably won't bind to and test against the secp256k1-zkp implementation in this PR


crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
```diff
@@ -0,0 +1,225 @@
...
+
+  def genMultiNonce(
+      privKeyOpt: Option[ECPrivateKey] = None,
+      aggPubKeyOpt: Option[SchnorrPublicKey] = None,
+      msgOpt: Option[ByteVector] = None,
+      extraInOpt: Option[ByteVector] = None): (
+      MultiNoncePub,
+      MultiNoncePriv) = {
+    msgOpt.foreach(msg => require(msg.length == 32))
+    extraInOpt.foreach(extraIn => require(extraIn.length <= 4294967295L))
```
benthecarman: nit: more readable imo if you do `require(x.forAll(...))`

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
```diff
@@ -0,0 +1,225 @@
...
+
+    val preRand =
+      ECPrivateKey.freshPrivateKey.bytes // TODO this needs to draw 32 bytes uniformily at random
```
benthecarman: is that not what this does?

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
```diff
@@ -0,0 +1,225 @@
...
+      gacc: Boolean) {
+    def getPubKey: ByteVector = aggPubKey.bytes.tail
+    def getPubKey33: ByteVector = aggPubKey.bytes
+  }
+
+  case class KeySet(keys: Vector[SchnorrPublicKey]) {
+    keys.init.zip(keys.tail).foreach { case (key1, key2) =>
+      require(key1.hex.compareTo(key2.hex) <= 0,
+              "Keys must be sorted lexicographically")
+    }
```
benthecarman: might be eaiser to make a `Ordering[SchnorrPublicKey]` so you can just call `.sorted`

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
nkohen: If I'm not mistaken it almost is, but this throws away a random draw if it is bigger than the field order (which is just slightly less than 2^32), so I need to rewrite without the throw away code

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
nkohen: Perhaps we should just have a lexicographic ordering defined for all NetworkElements. Now that I say this I vaguely recall already doing this but maybe it's false deja vu, I'll look for it

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
nkohen: Turns out we already have `CryptoUtil.randomBytes`, so I use this :)

crypto/src/main/scala/org/bitcoins/crypto/MuSig2Util.scala
```diff
@@ -92,21 +81,35 @@ object MuSig2Util {
         .reduce(_.add(_))
     }
 
-    lazy val secondKey: SchnorrPublicKey = keys.tail.head
+    // In truth this represents the first key different from the head key
+    lazy val secondKeyOpt: Option[SchnorrPublicKey] = {
+      keys.find(_ != keys.head)
```
benthecarman: why not do `keys.tail.headOption`

nkohen: As I discovered while implementing test vectors, if the same pubkey is used in the 0,1 spots then you have to take the next one

crypto/src/main/scala/org/bitcoins/crypto/CryptoRuntime.scala
```diff
@@ -211,6 +211,17 @@ trait CryptoRuntime {
 
   def tweakMultiply(publicKey: ECPublicKey, tweak: FieldElement): ECPublicKey
 
+  def tweakMultiply(point: SecpPoint, tweak: FieldElement): SecpPoint = {
+    point match {
+      case SecpPointInfinity => SecpPointInfinity
+      case pt: SecpPointFinite =>
+        Try(tweakMultiply(pt.toPublicKey, tweak)) match {
+          case Failure(_)       => SecpPointInfinity
```
nkohen: This really should double check by multiplying by tweak.subtract(Field element.one) and checking they're opposite

crypto/src/main/scala/org/bitcoins/crypto/CryptoBytesUtil.scala
```diff
@@ -99,6 +99,18 @@ trait CryptoBytesUtil {
   final def toByteVector[T <: NetworkElement](h: Seq[T]): ByteVector = {
     ByteVector.concat(h.map(_.bytes))
   }
+
+  def splitEvery(bytes: ByteVector, intervalLen: Int): Vector[ByteVector] = {
```
benthecarman: this already exists, call `.grouped`

crypto/src/main/scala/org/bitcoins/crypto/musig/MuSigNoncePub.scala
```diff
@@ -0,0 +1,62 @@
...
+      case SecpPointInfinity  => CryptoParams.getG
+      case p: SecpPointFinite => p.toPublicKey
+    }
+  }
+}
+
+object MuSigNoncePub extends Factory[MuSigNoncePub] {
+
+  /** In the BIP, the point at infinity is serialized as 33 0x00 bytes */
+  val infPtBytes: ByteVector = ByteVector.fill(33)(0)
```
benthecarman: nit: `ByteVector.low(33)`

Christewart (commented):
Can you add a `musig.md` or something with an example?

Christewart (approved):

PR #4424, opened 2022-06-24
V22 RPC Support Update Continued

GreyMcCarthy:
- [x] The getpeerinfo RPC returns two new boolean fields, bip152_hb_to and bip152_hb_from, that respectively indicate whether we selected a peer to be in compact blocks high-bandwidth mode or whether a peer selected us as a compact blocks high-bandwidth peer. High-bandwidth peers send new block announcements via a cmpctblock message rather than the usual inv/headers announcements. See BIP 152 for more details. (#19776)

- [x] getpeerinfo no longer returns the following fields: addnode, banscore, and whitelisted, which were previously deprecated in 0.21. Instead of addnode, the connection_type field returns manual. Instead of whitelisted, the permissions field indicates if the peer has special privileges. The banscore field has simply been removed. (#20755)

- [x] The following RPCs: gettxout, getrawtransaction, decoderawtransaction, decodescript, gettransaction, and REST endpoints: /rest/tx, /rest/getutxos, /rest/block deprecated the following fields (which are no longer returned in the responses by default): addresses, reqSigs. The -deprecatedrpc=addresses flag must be passed for these fields to be included in the RPC response. This flag/option will be available only for this major release, after which the deprecation will be removed entirely. Note that these fields are attributes of the scriptPubKey object returned in the RPC response. However, in the response of decodescript these fields are top-level attributes, and included again as attributes of the scriptPubKey object. (#20286)

- [x] When creating a hex-encoded bitcoin transaction using the bitcoin-tx utility with the -json option set, the following fields: addresses, reqSigs are no longer returned in the tx output of the response. (#20286)

- [x] The listbanned RPC now returns two new numeric fields: ban_duration and time_remaining. Respectively, these new fields indicate the duration of a ban and the time remaining until a ban expires, both in seconds. Additionally, the ban_created field is repositioned to come before banned_until. (#21602)

- [x] The setban RPC can ban onion addresses again. This fixes a regression introduced in version 0.21.0. (#20852

- [x] The getnodeaddresses RPC now returns a "network" field indicating the network type (ipv4, ipv6, onion, or i2p) for each address. (#21594)

- [x] getnodeaddresses now also accepts a "network" argument (ipv4, ipv6, onion, or i2p) to return only addresses of the specified network. (#21843)

- [x] The testmempoolaccept RPC now accepts multiple transactions (still experimental at the moment, API may be unstable). This is intended for testing transaction packages with dependency relationships; it is not recommended for batch-validating independent transactions. In addition to mempool policy, package policies apply: the list cannot contain more than 25 transactions or have a total size exceeding 101K virtual bytes, and cannot conflict with (spend the same inputs as) each other or the mempool, even if it would be a valid BIP125 replace-by-fee. There are some known limitations to the accuracy of the test accept: it's possible for testmempoolaccept to return "allowed"=True for a group of transactions, but "too-long-mempool-chain" if they are actually submitted. (#20833)

- [x] addmultisigaddress and createmultisig now support up to 20 keys for Segwit addresses. (#20867)

- [x] External signers such as hardware wallets can now be used through the new RPC methods enumeratesigners and displayaddress. Support is also added to the send RPC call. This feature is experimental. See external-signer.md for details. (#16546)

- [x] A new listdescriptors RPC is available to inspect the contents of descriptor-enabled wallets. The RPC returns public versions of all imported descriptors, including their timestamp and flags. For ranged descriptors, it also returns the range boundaries and the next index to generate addresses from. (#20226)

- [x] The bumpfee RPC is not available with wallets that have private keys disabled. psbtbumpfee can be used instead. (#20891)

- [x] The fundrawtransaction, send and walletcreatefundedpsbt RPCs now support an include_unsafe option that when true allows using unsafe inputs to fund the transaction. Note that the resulting transaction may become invalid if one of the unsafe inputs disappears. If that happens, the transaction must be funded with different inputs and republished. (#21359)

- [x] We now support up to 20 keys in multi() and sortedmulti() descriptors under wsh(). (#20867)

Link to previous Pr which this comes from https://github.com/bitcoin-s/bitcoin-s/pull/3821


app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/RawTransactionResult.scala
```diff
@@ -12,39 +12,96 @@ import scala.concurrent.duration.FiniteDuration
...
+    extends RpcTransactionOutput
+
+case class RpcTransactionOutputV22(
+    value: Bitcoins,
+    n: Int,
+    scriptPubKey: RpcScriptPubKeyV22)
+    extends RpcTransactionOutput
 
-case class RpcScriptPubKey(
+sealed abstract trait RpcScriptPubKey extends RawTransactionResult {
```
Christewart: just do `sealed trait`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/RawTransactionResult.scala
```diff
@@ -12,39 +12,96 @@ import scala.concurrent.duration.FiniteDuration
...
     scriptType: ScriptType,
     addresses: Option[Vector[BitcoinAddress]])
-    extends RawTransactionResult
+    extends RpcScriptPubKey
+
+case class RpcScriptPubKeyV22(asm: String, hex: String, scriptType: ScriptType)
+    extends RpcScriptPubKey
 
-case class DecodeScriptResult(
+sealed abstract trait DecodeScriptResult extends RawTransactionResult {
```
Christewart: `sealed trait` rather than `sealed abstract trait`

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/RawTransactionResult.scala
```diff
@@ -57,7 +114,24 @@ case class SignRawTransactionWithWalletResult(
     complete: Boolean
 )
 
-case class GetRawTransactionResult(
+sealed abstract trait GetRawTransactionResult extends RawTransactionResult {
```
Christewart: same here

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/WalletResult.scala
```diff
@@ -437,6 +470,20 @@ object AddressInfoResultPostV21 {
   }
 }
 
+case class listDescriptorsResult(
```
Christewart: Class names follow camel casing format. 

Class names should look like `ListDescriptorsResult` 

where as a `val` name is 

```scala
val descriptorResult = ListDescriptorsResult()
```

app-commons/src/main/scala/org/bitcoins/commons/jsonmodels/bitcoind/WalletResult.scala
```diff
@@ -437,6 +470,20 @@ object AddressInfoResultPostV21 {
...
+    wallet_name: String,
+    descriptors: Vector[descriptorsClass]
+) extends WalletResult
+
+case class descriptorsClass(
```
Christewart: Same here, `DescriptorsResult` or something like that. I would remove the `Class` suffix and replace it with `Result`.

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/common/MiningRpcTest.scala
```diff
@@ -42,7 +42,7 @@ class MiningRpcTest extends BitcoindRpcTest {
     } yield assert(info.chain == "regtest")
   }
 
-  it should "be able to generate blocks to an address" in {
+  /* it should "be able to generate blocks to an address" in {
```
Christewart: does this need to be uncommented?

bitcoind-rpc-test/src/test/scala/org/bitcoins/rpc/v22/BitcoindV22RpcClientTest.scala
```diff
@@ -1,52 +1,327 @@
...
+      _ <- client.getNewAddress.flatMap(
+        client.generateToAddress(100, _)
+      ) // Can't spend until depth 100
+
+      mempoolAccept <- client.testMempoolAccept(
+        Vector(signedTransactionOne.hex, signedTransactionTwo.hex))
+      _ = println(mempoolAccept)
+    } yield {
+      val mempooltxid: Int = mempoolAccept.length
+      println(s"mempoolAcceptTxid=$mempooltxid")
```
Christewart: remove this `println`

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/v22/TestMempoolAcceptRpc.scala
```diff
@@ -0,0 +1,68 @@
...
+}
+import org.bitcoins.core.protocol.transaction.Transaction
+import org.bitcoins.rpc.client.common.Client
+import play.api.libs.json.Json
+
+import scala.concurrent.Future
+
+trait TestMempoolAcceptRpc { self: Client =>
+
+  /**  def testMempoolAccept(
```
Christewart: Does this need to be here anymore? Or can all this commented code be deleted?

Christewart (changes requested):

Christewart (approved):
LGTM, great job @GreyMcCarthy !

PR #4427, opened 2022-06-27
lnd: Add needed param for taproot compute input script

benthecarman:
Because of the new signing algo we need `prevOuts: Vector[TransactionOutput]`


benthecarman:
Adding a test first

PR #4428, opened 2022-06-27
lnd: Add needed param for taproot compute input script

benthecarman:
Because of the new signing algo we need `prevOuts: Vector[TransactionOutput]`


Christewart (approved):

Christewart:
Looks like test case fails: https://github.com/bitcoin-s/bitcoin-s/runs/7068371120?check_suite_focus=true#step:5:783

benthecarman:
Turns out this isn't supported by LND yet :/

https://github.com/lightningnetwork/lnd/issues/6678

Issue #4429, opened 2022-06-27
Update documentation for UI to use build script


Christewart:
https://bitcoin-s.org/docs/ui-setup#building-and-starting-the-frontend

when this PR is merged: https://github.com/bitcoin-s/bitcoin-s-ts/pull/155

Christewart:
Done in #4430

PR #4430, opened 2022-06-27
 Update documentation for UI to use build script Advanced Setup 

GreyMcCarthy:
Updated the instructions for setting up the Suredbits wallet app, now using the build script.
Closes this PR   https://github.com/bitcoin-s/bitcoin-s/issues/4429

### DIFF
--- a/docs/ui-setup.md
+++ b/docs/ui-setup.md
@@ -29,67 +29,77 @@ https://youtu.be/oR0I0aHxNMM?t=219
 
 This requires command line skills.
 
-### Starting the backend
 
-From the image above, download the `bitcoin-s-server` artifact.
-
-![Example banner](/img/installers.png)
-
-After unzipping, you will need to run `chmod +x ./bin/bitcoin-s-server` from the terminal make the file executable.
-
-After making the file executable, you can start the server with 
-
-```
-unzip bitcoin-s-server.zip
-cd bitcoin-s-server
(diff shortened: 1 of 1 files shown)


docs/ui-setup.md
```diff
@@ -29,67 +29,77 @@ https://youtu.be/oR0I0aHxNMM?t=219
...
+cd wallet-electron-ts && npm run start"
+
 ```
+###Instill desktop application (Optional)
+
+If you would like to use the desktop application then run
 
-Now if you navigate to your web browser, you should see a page like this at `http://localhost:3002` 
+```
+npm make
```
Christewart: is this right? or is it `npm run make`? Also isn't this run in `./build-wallet-electron.sh`

GreyMcCarthy: Yeah it should be `npm run make`, my bad. I took the `npm run make` out of the code. Would you like me to put it back in so that the installer downloads and then give instructions on how to install i.e the part after `npm run make`?

Christewart: I don't understand why this is here if we put `npm run make` in `build-wallet-electron.sh`, won't that get run inside of that script?

GreyMcCarthy: Yeah it will get run inside `build-wallet-electron.sh` but `npm run make` is not currently in the bash script would you like me to add it? Then I can take that step out of the instructions

Christewart (approved):

PR #4439, opened 2022-06-30
2022 06 30 `OP_CODESEPARATOR` impl compatible with taproot

Christewart:
Pulled from @benthecarman 's branch to implement `OP_CODESEPARATOR` in  way that is compatible with taproot. I thought this was worth pulling out, and we might want to figure out better variable names. `lastCodeSeparator` and `codeSeparatorIdx` aren't very descriptive, but i don't really have great proposals either

https://github.com/Christewart/bitcoin-s-core/pull/35


benthecarman (approved):
Yeah not sure of better name, maybe just prefixing the new variable with Taproot or tapscript would be better

benthecarman:
Also could alternatively reuse the old variable and set it based on the sig version

benthecarman (approved):

PR #4109, opened 2022-02-19
2022 02 18 sync since creationtime pt2

Christewart:
fixes #1731

replaces #3136 

This syncs compact filters from our key managers `creationTime`. The benefit of this is to reduce initial block download time and save on disk space.

This PR syncs

1. All block headers
2. All compact filter headers
3. Compact filters _after_ `creationTime`. 

This does _not_ sync old compact filters from before the wallet creation time after syncing to tip. We may want to implement this in a separate PR.

### Syncing an existing wallet
When testing on my own local wallet with IBD, ` "creationTime" : 1558799128`  (`May 25, 2019`) it took me ~50 minutes to sync. 

```
2022-02-19T13:25:53UTC INFO [BitcoinSServerMain] version=1.9.0-7-a073f0c1-20220219-0721-SNAPSHOT
...
2022-02-19T14:19:49UTC INFO [DataMessageHandler] We are synced
```

### Syncing a new wallet

When syncing it a fresh wallet with a new seed it took me ~30 minutes on my desktop. This 30 minutes is all block headers + filter header sync.

```
2022-02-19T14:57:33UTC INFO [BitcoinSServerMain] version=1.9.0-7-a073f0c1-20220219-0721-SNAPSHOT
...
2022-02-19T15:31:50UTC INFO [PeerMessageSender] Requesting next compact filter headers from FilterSyncMarker(startHeight=724000, stopBlockHash=00000000000000000004e4ab92fe65a3e98dca855737f4d94084763cb0af97de)
2022-02-19T15:31:51UTC INFO [DataMessageHandler] Done syncing filter headers, beginning to sync filters from startHeightOpt=Some(724044)
2022-02-19T15:31:51UTC INFO [PeerMessageSender] Requesting compact filters from FilterSyncMarker(startHeight=724045, stopBlockHash=00000000000000000004e4ab92fe65a3e98dca855737f4d94084763cb0af97de)
2022-02-19T15:31:51UTC INFO [DataMessageHandler] We are synced
2022-02-19T15:31:51UTC INFO [DataMessageHandler] Processing 1 filters
```

### Disk space savings

The wallet in question was created on `1620073871` (`5/3/2021`) 
The disk space savings with prune filters is 13GB! This wallet was created within the last year, so old wallets won't have as much disk savings. 

Screen Shot 2022-02-20 at 5 52 29 PM

This second wallet was created on ` "creationTime" : 1558799128`  (`May 25, 2019`). This saves 10GB in disk space.

[...]

PR #4441, opened 2022-07-01
Fix sync issues post #4408

solo1g:
* #4408 had an issue with changing state of shared fixtures (a shared bitcoind instance was stopped). Changed the test that required it to not use shared fixtures.
* `PeerMessageReceiverTest`: The `Peer` argument does not correspond to the client. Works as the test is independent of what the `Peer` is but may cause confusion. Was introduced in #4408 so changed here.
* Using a `walletCreationTime` while creating `DataMessageHandler` may lead to sync issues in fixtures sometimes. We sync filters post the wallet creation time. It may sometimes happen that the creation time is so recent that the `startHeight` is one more than the tip, in which case it won't sync any filters at all. This is a problem only for test fixtures since we don't get more headers that would allow filters to sync.


Christewart:
`walletCreationTime` bug was likely introduced in #4109

Christewart (approved):

PR #4444, opened 2022-07-02
Fix TaprootKeyPath.isValid for 64 byte sigs

benthecarman:
checking the size of the stack not the element


Christewart:
:man_facepalming:

PR #4445, opened 2022-07-03
Allow creation of TaprootTxSigComponent

benthecarman:
For verifying a taproot txs we need a `Map[TransactionOutPoint, TransactionOutput]`. This adds it to the all the `TxSigComponent` creation functions


nkohen (commented):
Everything looks good to me, but I would ask for the Map to get a case class extending MapWrapper since it has so many uses (I think Neutrino might someday use such a map too?), And it looks like there are some common constructors from other things you could put in the companion object

benthecarman:
> Everything looks good to me, but I would ask for the Map to get a case class extending MapWrapper since it has so many uses (I think Neutrino might someday use such a map too?), And it looks like there are some common constructors from other things you could put in the companion object

Good idea, done

Christewart (approved):

Christewart (approved):

PR #4449, opened 2022-07-05
2022 07 05 `UnknownControlBlock`

Christewart:
Adds new data structures 

1. `UnknownControlBlock` - needed to represent a control block that does not have a known leaf version
2. `TapscriptControlBlock` - represents a taproot script path with a leaf version defined in [BIP341/BIP342 (`0xc0`,`0xc1`)](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#rationale)
3. `TaprootUnknownPath` - a taproot script path that corresponds to `UnknownControlBlock`

The reason this is needed is because the static test vectors in #3769 have unknown leaf versions to make sure we maintain future soft fork compatability.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
@@ -3,6 +3,8 @@ package org.bitcoins.core.protocol.script
 import org.bitcoins.crypto.{Factory, NetworkElement, XOnlyPubKey}
 import scodec.bits.ByteVector
 
+import scala.util.Try
+
 /** Control block as defined by BIP341
   *
   * The last stack element is called the control block c, and must have length 33 + 32m,
@@ -10,12 +12,8 @@ import scodec.bits.ByteVector
   *
   * @see https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules
   */
-case class ControlBlock(bytes: ByteVector) extends NetworkElement {
-  //invariants from: https://github.com/bitcoin/bitcoin/blob/37633d2f61697fc719390767aae740ece978b074/src/script/interpreter.cpp#L1835
-  require(bytes.size >= TaprootScriptPath.TAPROOT_CONTROL_BASE_SIZE)
-  require(bytes.size <= TaprootScriptPath.TAPROOT_CONTROL_MAX_SIZE)
-  require(
-    (bytes.size - TaprootScriptPath.TAPROOT_CONTROL_BASE_SIZE) % TaprootScriptPath.TAPROOT_CONTROL_NODE_SIZE == 0)
+sealed trait ControlBlock extends NetworkElement {
+  require(ControlBlock.isValid(bytes), s"Bytes for control block are not valid")
 
   /** Let p = c[1:33] and let P = lift_x(int(p)) where lift_x and [:] are defined as in BIP340. Fail if this point is not on the curve.
     */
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
```diff
@@ -10,12 +10,8 @@ import scodec.bits.ByteVector
...
   * @see https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#script-validation-rules
   */
-case class ControlBlock(bytes: ByteVector) extends NetworkElement {
-  //invariants from: https://github.com/bitcoin/bitcoin/blob/37633d2f61697fc719390767aae740ece978b074/src/script/interpreter.cpp#L1835
-  require(bytes.size >= TaprootScriptPath.TAPROOT_CONTROL_BASE_SIZE)
-  require(bytes.size <= TaprootScriptPath.TAPROOT_CONTROL_MAX_SIZE)
-  require(
-    (bytes.size - TaprootScriptPath.TAPROOT_CONTROL_BASE_SIZE) % TaprootScriptPath.TAPROOT_CONTROL_NODE_SIZE == 0)
+sealed trait ControlBlock extends NetworkElement {
+  require(ControlBlock.isValid(bytes), s"Bytes for control block are not valid")
```
benthecarman: would this be better as an `abstract class` because we have this `require`

benthecarman: Seems like we repeat it in the case classes as welll, which would be ran first?

benthecarman (commented):
otherwise lgtm

benthecarman (changes requested):
Looks like `BlockTest` is failing

core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
```diff
@@ -31,9 +27,60 @@ case class ControlBlock(bytes: ByteVector) extends NetworkElement {
...
+  * This is needed for future soft fork compatability where we introduce new leaf versions
+  * to correspond to new spending rules
+  */
+case class UnknownControlBlock(bytes: ByteVector) extends ControlBlock
+
 object ControlBlock extends Factory[ControlBlock] {
 
   override def fromBytes(bytes: ByteVector): ControlBlock = {
-    new ControlBlock(bytes)
+    TapscriptControlBlock(bytes)
```
benthecarman, line 44: This should cover both cases, known and unknown

```scala
    if (TapscriptControlBlock.isValid(bytes)) {
      TapscriptControlBlock(bytes)
    } else {
      UnknownControlBlock(bytes)
    }
```

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -181,13 +181,18 @@ object ScriptWitness extends Factory[ScriptWitness] {
...
+      P2WPKHWitnessV0(pubKey)
+    } else if (TaprootScriptPath.isValid(stack.toVector)) {
+      TaprootScriptPath.fromStack(stack.toVector)
     } else if (isPubKey && stack.size == 2) {
       val pubKey = ECPublicKeyBytes(stack.head)
       val sig = ECDigitalSignature(stack(1))
       P2WPKHWitnessV0(pubKey, sig)
-    } else if (isPubKey && stack.size == 1) {
-      val pubKey = ECPublicKeyBytes(stack.head)
-      P2WPKHWitnessV0(pubKey)
```
benthecarman: These changes seem to have broke the lnd tests where we fund a psbt. Seems like we can't properly parse the witness it is filling in. Looking into it now

core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
Christewart: We don't want to do this because there are certain cases where a `P2WSHWitnessV0` is a valid control block. I've been battling with this on #3769 (serialization symmetry in `ScriptWitnessSpec`) unfortunately some witnesses can have the correct length (`33 + 32m` bytes in length).

I think the only way we can properly support this is if we _know_ we are spending a `TaprootScriptPubKey`, else we will get false positives on `UnknownControlBlock`

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -441,10 +440,37 @@ object TaprootScriptPath {
 
   /** Checks the witness stack has an annex in it */
   def hasAnnex(stack: Vector[ByteVector]): Boolean = {
-    stack.headOption.map(_.head) == annexOpt
+    stack.headOption
+      .map(_.headOption == annexOpt)
+      .getOrElse(false)
```
benthecarman, line 447: nit using `exists(...)` would have the same functionality

core/src/main/scala/org/bitcoins/core/protocol/script/ControlBlock.scala
benthecarman: Yeah maybe a parent type could be made that encapsulates `P2WSHWitnessV0` and `TaprootScriptPath` that we parse to and when evaluating we cast to the correct one based on the spk

benthecarman: `UnresolvedWitnessScript`?

Christewart: Idk, we can get it right 99% of the time just not 100% of the time. I think it will be very rare in production that the heurstics are incorrect, we just test a lot of corner cases in our property based tests. I think we leave it for now and revisit when i implement failure cases

Christewart: Is this the root cause of the `lndRpcTest/test` failures?

benthecarman: > Is this the root cause of the lndRpcTest/test failures?

I believe so

benthecarman: > Is this the root cause of the lndRpcTest/test failures?

I am not too sure, adding the ability to parse these witnesses breaks the tests

Christewart: This seems to pass on 4e643be85331e8dddc112ee2dd8bafe119715e69 so I must have accidentally copied it over wrong, i'll look closer after I exercise

core/src/main/scala/org/bitcoins/core/protocol/script/ScriptWitness.scala
```diff
@@ -279,7 +281,7 @@ object TaprootKeyPath {
       //means SIGHASH_ALL is implicitly encoded
```
benthecarman: nit: fix this comment

PR #4456, opened 2022-07-07
Header sync validation

solo1g:
Multiple changes related to header sync in case of multiple peers:
- [x] Post header sync, query all peers for headers once and sync if a better chain exists.
- [x] Test for ^
- [x] Re-query in case invalid headers are sent
- [x] Test for ^
- [x] Disconnect and use another peer if current one keeps sending invalid headers.
- [x] Test for ^ 
- [ ] ~Start syncing headers from non filter peers while waiting for filter peers.~ The test suite peers show support for `NODE_COMPACT_FILTERS` but not `NODE_NETWORK` so seeing where the issue lies at.
- [ ] ~Test for ^~

Put this up for some review. Personally I think this should be merged after the existing issues wrt node tests are dealt with.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithUncachedBitcoindTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithUncachedBitcoindTest.scala
@@ -1,7 +1,8 @@
 package org.bitcoins.node
 
+import com.typesafe.config.ConfigFactory
 import org.bitcoins.asyncutil.AsyncUtil
-import org.bitcoins.core.p2p.GetHeadersMessage
+import org.bitcoins.core.p2p.{GetHeadersMessage, HeadersMessage}
 import org.bitcoins.node.models.Peer
 import org.bitcoins.node.networking.P2PClient.ExpectResponseCommand
 import org.bitcoins.server.BitcoinSAppConfig
@@ -20,7 +21,9 @@ import scala.concurrent.{Await, Future}
 class NeutrinoNodeWithUncachedBitcoindTest extends NodeUnitTest with CachedTor {
 
   lazy val bitcoindsF =
-    BitcoindRpcTestUtil.createNodePair().map(p => Vector(p._1, p._2))
+    BitcoindRpcTestUtil
+      .createUnconnectedNodePairWithBlocks()
+      .map(p => Vector(p._1, p._2))
 
   lazy val bitcoinPeersF: Future[Vector[Peer]] = {
     bitcoindsF.flatMap { bitcoinds =>
@@ -30,7 +33,11 @@ class NeutrinoNodeWithUncachedBitcoindTest extends NodeUnitTest with CachedTor {
   }
 
   override protected def getFreshConfig: BitcoinSAppConfig = {
-    BitcoinSTestAppConfig.getMultiPeerNeutrinoWithEmbeddedDbTestConfig(pgUrl)
+    val configs = ConfigFactory.parseString("""
+                                              | peer-discovery-timeout = 5s
+                                              |""".stripMargin)
+    BitcoinSTestAppConfig.getMultiPeerNeutrinoWithEmbeddedDbTestConfig(pgUrl,
+                                                                       configs)
   }
 
   override type FixtureParam = NeutrinoNodeConnectedWithBitcoinds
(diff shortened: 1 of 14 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -205,6 +212,52 @@ case class DataMessageHandler(
           s"Received headers=${headers.map(_.hashBE.hex).mkString("[", ",", "]")}")
         val chainApiF = chainApi.processHeaders(headers)
 
+        chainApiF.failed.map {
+          case r: RuntimeException
+              if r.getMessage.startsWith(
+                "Failed to connect any headers to our internal chain state") =>
```
Christewart: Do we want to do something more intelligent from the chain project on this? Perhaps rather than returning an exception in this case, return some type that allows you to pattern match on and handle the failure gracefully?

This also seems like it would be useful in the context of multiple peers if we have _seen_ a header before. So perhaps we would have an ADT

```scala

sealed trait ChainHeaderResult
object ChainHeaderResult {
  //we've successfully connect to header to our blockchain
  case object Success extends ChainHeaderResult
  //we've seen this header before
  case object Duplicate extends ChainHeaderResult
  //catch all for invalid headers, however we could break this down further to
  //bad POW, bad previous block hash, etc 
  case object Invalid extends ChainHeaderResult 
}
```

Christewart:
Fails CI

shreyanshyad:
Yikes! I wonder how it worked locally for me :open_mouth:.

Christewart:
rebased to fix a few merge conflicts, i'm hopeful this PR will fix #4630

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandlerState.scala
```diff
@@ -0,0 +1,22 @@
...
+import scala.collection.mutable
+
+sealed abstract class DataMessageHandlerState
+
+object DataMessageHandlerState {
+
+  final case object HeaderSync extends DataMessageHandlerState
+
+  case class ValidatingHeaders(
+      inSyncWith: mutable.Set[Peer],
```
Christewart: Can this be refactored to remove direct access to mutable collections?

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -27,19 +30,21 @@ import scala.util.control.NonFatal
...
     walletCreationTimeOpt: Option[Instant],
+    node: Node,
     initialSyncDone: Option[Promise[Done]] = None,
     currentFilterBatch: Vector[CompactFilterMessage] = Vector.empty,
     filterHeaderHeightOpt: Option[Int] = None,
     filterHeightOpt: Option[Int] = None,
     syncing: Boolean = false,
-    syncPeer: Option[Peer] = None)(implicit
+    syncPeer: Option[Peer] = None,
+    state: DataMessageHandlerState = HeaderSync)(implicit
```
Christewart: i think its probably a good idea to not provide a default argument here. I think the creator `DataMesageHandler`  should have to explicitly pass the `state` it wants? 

Seems easy to forget to pass the state since it won't give a compiler error.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithUncachedBitcoindTest.scala
```diff
@@ -30,7 +33,11 @@ class NeutrinoNodeWithUncachedBitcoindTest extends NodeUnitTest with CachedTor {
   }
 
   override protected def getFreshConfig: BitcoinSAppConfig = {
-    BitcoinSTestAppConfig.getMultiPeerNeutrinoWithEmbeddedDbTestConfig(pgUrl)
+    val configs = ConfigFactory.parseString("""
+                                              | peer-discovery-timeout = 5s
```
Christewart: I believe this has to be`bitcoin-s.node.peer-discovery-timeout = 5s`

Christewart (changes requested):
Overall looks good to me, mostly small changes / nits.

Christewart (approved):
Thanks for getting this done @shreyanshyad :tada:

PR #4458, opened 2022-07-07
Have `FutureUtil.makeAsync` handle thrown exceptions

benthecarman:
Previously if `func()` failed  `FutureUtil.makeAsync` would just hang forever. This will now return the expected failed future


Christewart:
Unit test on this would be nice

benthecarman:
> Unit test on this would be nice

Added

Christewart (approved):

PR #4459, opened 2022-07-07
Add `imported` flag to the seed JSON

rorp:
All seeds created using `importseed` RPC will have this flag set to `true`. Automatically genarted seeds (ex. the default wallet) will have it set to `false`.


Christewart:
This doesn't seem to return the status of the keymanager via the `walletinfo` endpoint. It should probably be inside the `keymanager` json object IMO

Screenshot from 2022-07-08 09-04-31

user411:
Does the new `imported` flag return on the WalletInfo or on the nested Wallet object in the WalletInfo call?

Christewart:
> Does the new `imported` flag return on the WalletInfo or on the nested Wallet object in the WalletInfo call?

Looks like it isn't inside of the `keymanager` object. I'm not sure why @rorp didn't put it on `keymanager` as the code changes are in the keymanager module

Screenshot from 2022-07-08 12-29-31

user411:
I will just throw it away at the UI if it's not on the wallet object proper.

Christewart (approved):

PR #4461, opened 2022-07-07
tapscript/sigopsratio_3 failing correctly

benthecarman:


core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
```diff
@@ -81,60 +81,69 @@ sealed abstract class CryptoInterpreter {
...
-              ) {
-                //trivially pass signature validation as required by BIP342
-                //when the public key type is not known and the
-                //SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_PUBKEYTYPE is NOT set
-                handleSignatureValidation(program = program,
-                                          result = SignatureValidationSuccess,
+      val sigBytes = program.stack.tail.head.bytes
+
+      if (
+        sigBytes.nonEmpty && program.validationWeightRemaining.forall(_ >= 50)
```
Christewart: `.forall(_ > 0)`

core-test/.jvm/src/test/scala/org/bitcoins/core/protocol/transaction/TaprootTxTests.scala
```diff
@@ -80,7 +81,7 @@ class TaprootTxTests extends BitcoinSJvmTest {
     }
   }
 
-  it must "run the failure test cases through the script interpreter" ignore {
+  it must "run the failure test cases through the script interpreter" in {
```
Christewart: can you re-ignore this?

Christewart (approved):
LGTM, just re-ignore that test case and can merge

Christewart (commented):
Hmm, guess it looks like there are some bugs here according to CI

core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
benthecarman: This needs to be `>= 50` because we haven't decremented the budget yet

benthecarman: Woops, needed to be `.exists(_ < 0)` because this is the failure case

Christewart (changes requested):
Still fails for me on f9d306debc6adccc44100cc6e3024555decbf025 

Screenshot from 2022-07-08 07-02-21

benthecarman:
Forgot to implement for `OP_CHECKSIGADD`, passes now

benthecarman:
Dang this broke the success cases will take a look later

Christewart (approved):

PR #4463, opened 2022-07-08
Fix filter sync if headers received while syncing

solo1g:
If headers are received when the node is still syncing filters or filter headers, then no queries to fetch filers are made for the received headers.

In tests, when node receives header messages one after the other, it makes a request for filters for the first header message and sets `syncing` as true which is back again to false after filters are received. Any headers received in between them are processed but query for filter would not be made.
Haven't verified it, but this should also happen for any headers received during the node IBD on mainnet as the filters for the headers received while node is syncing filter headers and filters are not queried for. In such a case, we would still have a fully synced chain once we receive any header post IBD.

This adds a simple check after a compact filter sync to ensure that the filters and filter headers are actually in sync with headers, and if not, it tries to sync.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
@@ -13,7 +13,7 @@ import org.bitcoins.testkit.node.{
   NodeTestWithCachedBitcoindPair,
   NodeUnitTest
 }
-import org.bitcoins.testkit.util.{AkkaUtil, TorUtil}
+import org.bitcoins.testkit.util.TorUtil
 import org.scalatest.{Assertion, FutureOutcome, Outcome}
 
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


Christewart:
All tests seem to pass on 7530d527040671ad9fb800c78be578af85e92260 locally for me _besides_ the `ReConnectionTest`. That one seems to hang consistently on mac and infrequently on my linux machine.

node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala
```diff
@@ -44,7 +44,7 @@ class NeutrinoNodeTest extends NodeTestWithCachedBitcoindPair {
 
   behavior of "NeutrinoNode"
 
-  it must "be able to sync" in { nodeConnectedWithBitcoinds =>
+  ignore must "be able to sync" in { nodeConnectedWithBitcoinds =>
```
Christewart: are these ignored intentionally or did you forget to revert?

solo1g: Oops! Forgot this :sweat_smile:.

Christewart (approved):

Issue #4465, opened 2022-07-08
Unable to restart `BitcoindRpcClient`


solo1g:
Found while debugging some node test failures post #4408 where `ReConnectionTest` would fail because `bitcoind.start()` future called after stopping the bitcoind client never completes. Note that no exceptions were thrown, it just hangs at that. Looking into `start`, possible issue may be with `tryPing()` future and the http request it sends as that is the only part that does not have a timeout.

This was found after #4408, in which the only relevant thing I can think of is that another test was added which also restarts bitcoind. Could be related, not sure. Can be reproduced on 5df5bf674165eb2af0497fe4ec54b7789acceaca.

PR #4468, opened 2022-07-08
Change tests to use disconnectnode rpc instead of restarting bitcoind

solo1g:
Workaround for #4465.


Christewart (approved):

Christewart:
Looks like there is a new bug introduced in your helper method

Screenshot from 2022-07-08 13-27-51

solo1g:
I was using `addr_local` being defined for filtering out which is our node and which are other bitcoind instances but that being defined requires the peer to be initialized, not just connected. Verified this by not sending a version message and querying for peer info. Changed the condition so it should work fine now.

Christewart (approved):

PR #4476, opened 2022-07-08
Implement `torstarted` websocket callback

### LINKED ISSUE #4472: `torstarted` websocket event
request from @user411

Christewart:
fixes #4472 

Adds a `torstarted` websocket event that looks like 

```
{"type":"torstarted","payload":null}
```


user411:
Can tor ever stop? If so, maybe we want a true/false if we can send the stop event.

Christewart:
> Can tor ever stop? If so, maybe we want a true/false if we can send the stop event.

Tor only stops on shutdown currently, or if the underlying OS kills the process for some reason (unlikely)

user411 (approved):

PR #4481, opened 2022-07-10
Restrict hash types allowed by taproot signature verification

Christewart:
[From BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message)

```
The following restrictions apply, which cause validation failure if violated:
Using any undefined hash_type (not 0x00, 0x01, 0x02, 0x03, 0x81, 0x82, or 0x83[[13](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-13)]).
```


core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -95,8 +96,16 @@ trait TransactionSignatureChecker {
       TransactionSignatureSerializer.hashForSignature(txSigComponent,
                                                       hashType,
                                                       taprootOptions)
-    val result = pubKey.verify(hash, schnorrSignature)
-    if (result) ScriptOk else ScriptErrorSchnorrSig
+    //bip341 restricts valid hash types: https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message
+    val invalidHashType = {
+      !(hashType.byte <= 3 || (hashType.byte >= 81.toByte && hashType.byte <= 0x83.toByte))
```
benthecarman: Shouldn't this be `0x81` not `81`, can do the same with `3`

https://github.com/sipa/bitcoin/blob/0e2a5e448f426219a6464b9aaadcc715534114e6/src/script/interpreter.cpp#L1523

benthecarman (approved):

core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -95,8 +96,16 @@ trait TransactionSignatureChecker {
       TransactionSignatureSerializer.hashForSignature(txSigComponent,
                                                       hashType,
                                                       taprootOptions)
-    val result = pubKey.verify(hash, schnorrSignature)
-    if (result) ScriptOk else ScriptErrorSchnorrSig
+    //bip341 restricts valid hash types: https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message
+    val invalidHashType = {
+      !(hashType.byte <= 0x03.toByte || (hashType.byte >= 0x81.toByte && hashType.byte <= 0x83.toByte))
+    }
```
benthecarman, line 99: nit: it'd be more effienct if we calculated the hash after doing this check

benthecarman (approved):

PR #4482, opened 2022-07-10
Schnorr sig parsing checks

benthecarman:
Was able to refactor `getSignatureAndHashType` to just take the `sigBytes` as we already had this in scope when we call it. We also already did the trivial check for the empty signature in a previous if statement.

This will now correctly handle to fail schnorr signatures that fail the invariants we have on our `SchnorrDigitalSignature` type


Christewart:
This seems to fix all failure cases of the flavor `siglen/empty_cs_neg`, does this sound right to you?

benthecarman:
> This seems to fix all failure cases of the flavor `siglen/empty_cs_neg`, does this sound right to you?

Yes, should have mentioned that in the comment

Christewart (approved):

PR #4483, opened 2022-07-10
Validate max ScriptNumber bytes in OP_CHECKSIGADD

benthecarman:
Gets the `tapscript/checksigaddoversize` failure cases passing.

From BIP 341

> For OP_CHECKSIGADD, the public key (top element), a CScriptNum n (second to top element), and a signature (third to top element) are popped from the stack.
> - If fewer than 3 elements are on the stack, the script MUST fail and terminate immediately.
> - If n is larger than 4 bytes, the script MUST fail and terminate immediately.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
+++ b/core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
@@ -451,43 +451,46 @@ sealed abstract class CryptoInterpreter {
           val flags = updatedProgram.flags
           val requireMinimal =
             ScriptFlagUtil.requireMinimalData(updatedProgram.flags)
+
           val numT = ScriptNumber(updatedProgram.stack(1).bytes, requireMinimal)
 
-          if (numT.isFailure) {
-            throw numT.failed.get
-          }
-          val restOfStack =
-            updatedProgram.stack.tail.tail.tail //remove signature, num, pubkey
+          if (numT.isFailure || numT.get.byteSize > 4) {
+            // not sure what error to throw here
+            updatedProgram.failExecution(ScriptErrorInvalidStackOperation)
+          } else {
+            val restOfStack =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala
```diff
@@ -451,43 +451,46 @@ sealed abstract class CryptoInterpreter {
...
-              if (
-                err == ScriptErrorDiscourageUpgradablePubkeyType && !ScriptFlagUtil
-                  .discourageUpgradablePublicKey(flags)
-              ) {
-                //trivially pass signature validation as required by BIP342
-                //when the public key type is not known and the
-                //SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_PUBKEYTYPE is NOT set
+          if (numT.isFailure || numT.get.byteSize > 4) {
+            // not sure what error to throw here
+            updatedProgram.failExecution(ScriptErrorInvalidStackOperation)
```
benthecarman: Not sure what error to throw here, this seemed the most logical

Christewart: Non minimally encoded numbers get `ScriptErrorUnknownError` since an actual exception throw in the interpreter. I don't make the rules 🤷‍♂️ 

Here is an example of the error that gets set if we have a non minimally encoded number in OP_CMS
https://github.com/bitcoin-s/bitcoin-s/blob/0397ca6c78ed307c25be90edf4306ecfc47824f1/core/src/main/scala/org/bitcoins/core/script/crypto/CryptoInterpreter.scala#L336

benthecarman:
Fixed another issue where we were getting the sigBytes before checking the size of the stack, this could cause a potential issue if there wasn't enough elements on the stack

Christewart (approved):

PR #4486, opened 2022-07-10
Enforce MAX_PUSH_SIZE for taproot

benthecarman:
Fixes `tapscript/inputmaxlimit` failure cases. Enforces the 520 byte max push size limit


Christewart:
Looks like this introduced a regression in the success unit test cases

benthecarman:
> Looks like this introduced a regression in the success unit test cases

Was doing the check against the control block & script so it was failing, should be fixed now

Christewart (approved):

PR #4496, opened 2022-07-12
Refactor coin selection to be not be bitcoin-s specific

benthecarman:
Previously coin selection used `SpendingInfoDb` which is specific to the bitcoin-s wallet. To make it more general I created a `CoinSelectorUtxo` which just has the necessary data required for coin selection.

Also added size calculations for taproot scripts


Christewart (approved):
This all looks fine to me, one small request is to make a `SpendingInfoDb.toCoinSelectorUtxo`

Issue #4460, opened 2022-07-07
Figure out what endpoints are not safe to call while syncing


Christewart:
We made #4452 to allow the frontend to be able to determine when we are syncing so we can avoid calling certain endpoints when we are syncing.

As discussed with @user411 this is really a workdaround and doesn't address the root cause. We shouldn't have endpoints that can fail at any point rather they should return something intelligent even when we are syncing. Our problematic endpoint in this case was `estimatefee`. 

We should figure out which endpoints are not safe to call when we are syncing and figure out what safe default should be to return in the case we are syncing.

user411:
`getdlchostaddress` occassionally returns a message that it is not a valid method to call on the backend during startup.

user411:
UI startup call timings. The one that doesn't return in ms is `estimatefee`.
![Screen Shot 2022-07-12 at 1 08 38 PM](https://user-images.githubusercontent.com/22351459/178575892-a6acc985-5ee7-4dad-9184-3c7116e06584.png)

Christewart:
> UI startup call timings. The one that doesn't return in ms is `estimatefee`. ![Screen Shot 2022-07-12 at 1 08 38 PM](https://user-images.githubusercontent.com/22351459/178575892-a6acc985-5ee7-4dad-9184-3c7116e06584.png)

Some details on how `estimatefee` works. 

#### neutrino backend

We make a request against the mempool.space API (by default, this is configurable). We use tor if the user has tor enabled. A source of latency here could be waiting for tor to startup


#### bitcoind backend
We use bitcoind's internal fee estimator by default if we are using a bitcoind backend. 

In the implementation, we do return `-1` if the request cannot be completed: 


https://github.com/bitcoin-s/bitcoin-s/blob/414557effb6697335f1c47de950c01101951a4a1/app/server/src/main/scala/org/bitcoins/server/WalletRoutes.scala#L981

However _my guess_ is the request against mempool.space isn't completing in a timely fashion. I have not been able to reproduce this locally. Perhaps we could add some sort of timeout so that if mempool.space isn't performing (or getting DDOS'd or something) we don't hang forever.

PR #4503, opened 2022-07-12
Implement a timeout on fee rate requests

Christewart:
This fixes the variable latency from retrieving a fee rate by introducing a 1 second timeout. If the request hasn't completed in 1 second, we timeout the request and return -1. 

Since we cache fee rates internally, the next time you call `estimatefee` you should get the correctly cached fee rate (assuming the original request completed).

Discussion: 
https://github.com/bitcoin-s/bitcoin-s/issues/4460#issuecomment-1182325014

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala
@@ -31,6 +31,7 @@ import org.bitcoins.core.wallet.rescan.RescanState
 import org.bitcoins.dlc.node.DLCNode
 import org.bitcoins.dlc.node.config.DLCNodeAppConfig
 import org.bitcoins.dlc.wallet._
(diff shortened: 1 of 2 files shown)


user411 (approved):
It's fine with me however you solve this. The frontend would be fine with a -1 immediate return with a later websocket message with the accurate fee rate as well.

user411:
Are you thinking the frontend will call again later if it gets a -1, or will you supply the value async when it returns?

Christewart:
> Are you thinking the frontend will call again later if it gets a -1, or will you supply the value async when it returns?

The frontend should call again later. 

We do cache fee rate values for 5 minutes on the backend after we have initially retrieved a fee rate, so after we have the first one fetched we should get instant responses.

user411:
In this "solution", the frontend may never get a response. Please send it async.

I would rather wait for the accurate value than do this, though it delays user login.

PR #4505, opened 2022-07-13
Remove assertion after syncing in NodeUnitTest

solo1g:
Was seeing this assertion fail occasionally.
* Removed assertion to check not syncing status after sync as this may not necessarily be true in a parallel setting. Example: after node is synced, some other test using the same cached bitcoind may produce another block right at that time which may change status to syncing again and hence this assertion fails.
* Added this check separately in another test that would not be affected by parallel execution.

### DIFF
--- a/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithUncachedBitcoindTest.scala
+++ b/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeWithUncachedBitcoindTest.scala
@@ -50,6 +50,21 @@ class NeutrinoNodeWithUncachedBitcoindTest extends NodeUnitTest with CachedTor {
 
   behavior of "NeutrinoNode"
 
+  it must "have not syncing status after sync" in {
+    nodeConnectedWithBitcoinds =>
+      val node = nodeConnectedWithBitcoinds.node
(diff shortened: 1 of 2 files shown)


Christewart:
> * Removed assertion to check not syncing status after sync as this may not necessarily be true in a parallel setting. Example: after node is synced, some other test using the same cached bitcoind may produce another block right at that time which may change status to syncing again and hence this assertion fails.

Each test suite should have it's own **unique** bitcoind pair. 

Each test within a suite should be executed synchronously (i.e. test `2` cannot start until test `1` is completed). 

Also we should be shutting down bitcoinds after the test suite is finisehd completing. This is done in the `afterAll` method

I made a little test case to prove this, if you look at the uuids in the image, they are different. However both `PeerMessageHandlerTest` and `PeerMessageReceiverTest` share the same parent trait called `NodeTestWithCachedBitcoindNewest`

Run 

>nodeTest/testOnly *PeerMessage*

Screenshot from 2022-07-13 10-39-21


Here is the branch if you would like to test yourself: https://github.com/Christewart/bitcoin-s-core/tree/2022-07-13-test-shared-trait

TLDR: 

The `parallelExecution` flag should not matter for `bitcoind` state. This is because test suites do not share bitcoind instances, rather they create their own unique bitcoind instance and share that instance for all the unit tests _within_ a test suite.

PR #4452, opened 2022-07-06
Add `sync` JSON field to the `getinfo` endpoint output

rorp:
Fixes https://github.com/bitcoin-s/bitcoin-s/issues/3899

PR #4506, opened 2022-07-13
Set syncing flag to true on startup

Christewart:
Follow up on #4452 to set the `syncing` flag to true on start by default. This occur before the http server is bound so that there is not a race condition.


user411:
Works well for me.

PR #4509, opened 2022-07-14
Add P2PClientSupervisor

solo1g:
-  Fixes logging for errors thrown in `P2PClientActor`.
 - The way P2PClient awaits response for a message like getheaders is that I send an ExpectResponseCommand to the node. So, in the case of headers, 2 messages would be sent consecutively, first getheaders then an ExpectResponseCommand that schedules a timeout function to be executed sometime in the future. Problem here was that sometimes a response from the peer may be received in between these 2 messages that we send. I noticed that, especially in the case of tests, sometimes response for getheaders is sent before the ExpectResponseCommand is sent. This is fixed now, the timer is started before even sending a getheaders so no messages are missed.
 - Second was with disconnecting peers, so that now we ignore messages once we have started disconnecting a peer.
 - Third would be the use of Await at some places. I had a quite big misunderstanding wrt foreach that cleared up with your earlier comments on this. Now I'm following the function signatures properly. So Await was introduced in postStop of Actor as that returned a Unit and schedulers or runnables that I'm using as they too expect a function with a Unit return type.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
+++ b/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
@@ -1,6 +1,6 @@
 package org.bitcoins.node
 
-import akka.actor.{ActorSystem, Cancellable}
+import akka.actor.{ActorRef, ActorSystem, Cancellable}
 import org.bitcoins.asyncutil.AsyncUtil
 import org.bitcoins.core.p2p.ServiceIdentifier
 import org.bitcoins.core.util.{NetworkUtil, StartStopAsync}
@@ -17,7 +17,8 @@ import scala.util.Random
 case class PeerFinder(
     paramPeers: Vector[Peer],
     node: NeutrinoNode,
-    skipPeers: () => Vector[Peer])(implicit
+    skipPeers: () => Vector[Peer],
+    supervisor: ActorRef)(implicit
     ec: ExecutionContext,
     system: ActorSystem,
     nodeAppConfig: NodeAppConfig)
@@ -177,7 +178,7 @@ case class PeerFinder(
 
   /** creates and initialises a new test peer */
   def tryPeer(peer: Peer): Unit = {
-    _peerData.put(peer, PeerData(peer, node))
+    _peerData.put(peer, PeerData(peer, node, supervisor))
     _peerData(peer).peerMessageSender.connect()
   }
 
--- a/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
+++ b/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
@@ -10,7 +10,7 @@ import org.bitcoins.node.models.{Peer, PeerDAO, PeerDb}
 import java.net.{InetAddress, UnknownHostException}
 import scala.collection.mutable
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.io.Source
 import scala.util.Random
 
@@ -131,7 +131,9 @@ case class PeerFinder(
             .filterNot(p => skipPeers().contains(p) || _peerData.contains(p))
 
           logger.debug(s"Trying next set of peers $peers")
-          peers.foreach(tryPeer)
+          val peersF = Future.sequence(peers.map(tryPeer))
+          Await.result(peersF, 10.seconds)
+          ()
         }
       }
     }
@@ -139,9 +141,10 @@ case class PeerFinder(
   override def start(): Future[PeerFinder] = {
     logger.debug(s"Starting PeerFinder")
 
-    (getPeersFromParam ++ getPeersFromConfig).distinct.foreach(tryPeer)
+    val initPeerF = Future.sequence(
+      (getPeersFromParam ++ getPeersFromConfig).distinct.map(tryPeer))
 
-    if (nodeAppConfig.enablePeerDiscovery) {
+    val peerDiscoveryF = if (nodeAppConfig.enablePeerDiscovery) {
       val startedF = for {
         (dbNonCf, dbCf) <- getPeersFromDb
       } yield {
@@ -159,6 +162,8 @@ case class PeerFinder(
       logger.info("Peer discovery disabled.")
       Future.successful(this)
     }
+
+    initPeerF.flatMap(_ => peerDiscoveryF)
   }
 
   override def stop(): Future[PeerFinder] = {
@@ -167,19 +172,21 @@ case class PeerFinder(
     //delete try queue
     _peersToTry.clear()
 
-    _peerData.foreach(_._2.client.close())
+    val closeF = Future.sequence(_peerData.map(_._2.client.map(_.close())))
 
-    AsyncUtil
+    val waitStopF = AsyncUtil
       .retryUntilSatisfied(_peerData.isEmpty,
                            interval = 1.seconds,
-                           maxTries = 10)
+                           maxTries = 30)
       .map(_ => this)
+
+    closeF.flatMap(_ => waitStopF)
   }
 
   /** creates and initialises a new test peer */
-  def tryPeer(peer: Peer): Unit = {
+  def tryPeer(peer: Peer): Future[Unit] = {
     _peerData.put(peer, PeerData(peer, node, supervisor))
-    _peerData(peer).peerMessageSender.connect()
+    _peerData(peer).peerMessageSender.map(_.connect())
   }
 
   def removePeer(peer: Peer): Unit = {
--- a/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
+++ b/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
@@ -1,6 +1,7 @@
 package org.bitcoins.node
 
 import akka.actor.{ActorRef, ActorSystem, Cancellable}
+import monix.execution.atomic.AtomicBoolean
 import org.bitcoins.asyncutil.AsyncUtil
 import org.bitcoins.core.p2p.ServiceIdentifier
 import org.bitcoins.core.util.{NetworkUtil, StartStopAsync}
@@ -12,7 +13,7 @@ import scala.collection.mutable
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.io.Source
-import scala.util.Random
+import scala.util.{Failure, Random, Success}
 
 case class PeerFinder(
     paramPeers: Vector[Peer],
@@ -117,12 +118,17 @@ case class PeerFinder(
     else nodeAppConfig.tryNextPeersInterval
   }
 
+  private val isConnectionSchedulerRunning = AtomicBoolean(false)
+
   private lazy val peerConnectionScheduler: Cancellable =
     system.scheduler.scheduleWithFixedDelay(
       initialDelay = initialDelay,
-      delay = nodeAppConfig.tryNextPeersInterval) {
-      new Runnable() {
-        override def run(): Unit = {
+      delay = nodeAppConfig.tryNextPeersInterval) { () =>
+      {
+        if (
+          isConnectionSchedulerRunning.compareAndSet(expect = false,
+                                                     update = true)
+        ) {
           logger.debug(s"Cache size: ${_peerData.size}. ${_peerData.keys}")
           if (_peersToTry.size < 32)
             _peersToTry.pushAll(getPeersFromDnsSeeds)
@@ -132,9 +138,16 @@ case class PeerFinder(
 
           logger.debug(s"Trying next set of peers $peers")
           val peersF = Future.sequence(peers.map(tryPeer))
-          peersF.failed.foreach(err =>
-            logger.error(s"Failed to connect to peers", err))
-          ()
+          peersF.onComplete {
+            case Success(_) =>
+              isConnectionSchedulerRunning.set(false)
+            case Failure(err) =>
+              isConnectionSchedulerRunning.set(false)
+              logger.error(s"Failed to connect to peers", err)
+          }
+        } else {
+          logger.warn(
+            s"Previous connection scheduler is still running, skipping this run, it will run again in ${nodeAppConfig.tryNextPeersInterval}")
         }
       }
     }
--- a/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
+++ b/node/src/main/scala/org/bitcoins/node/PeerFinder.scala
@@ -10,7 +10,7 @@ import org.bitcoins.node.models.{Peer, PeerDAO, PeerDb}
 import java.net.{InetAddress, UnknownHostException}
 import scala.collection.mutable
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
-import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.{ExecutionContext, Future}
 import scala.io.Source
 import scala.util.Random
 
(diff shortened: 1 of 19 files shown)


node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -559,20 +559,23 @@ object P2PClient extends P2PLogger {
...
+
+    val clientProps = props(peer,
+                            peerMessageReceiver,
+                            onReconnect,
+                            onStop,
+                            maxReconnectionTries)
+
+    implicit val timeout: Timeout = Timeout(10.second)
+    val actorRefF = supervisor ? clientProps
+    val actorRef =
```
Christewart: Lets avoid blocking here, can we just return a `Future[P2PClient]` from this method?

Christewart (changes requested):

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -166,7 +167,7 @@ case class PeerFinder(
     //delete try queue
     _peersToTry.clear()
 
-    _peerData.foreach(_._2.client.close())
+    _peerData.foreach(_._2.client.foreach(_.close()))
```
Christewart: This is not a good idea. This is dropping a `Future` on the floor here. 

Don't use `Future.foreach()`. If an error occurs in the future it will be swallowed and not propgated upwards.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -154,7 +162,7 @@ case class PeerManager(
     //leading to a memory leak may happen
     _waitingForDeletion.add(peer)
     //now send request to stop actor which will be completed some time in future
-    client.close()
+    client.foreach(_.close())
```
Christewart: You are dropping a `Future` on the floor here. 

This introduces race conditions as the `removePeer()` method returns `Unit`, however the computation inside of `removePeer` is done asynchronously. 

Part of the bugs you are experiencing in the code you've introduced is the fact that you aren't being consistent with handling async code. A nightmare of async code is the combination of 

1. mutability (specifically globally accessible mutability, which you are attempting to introduce here: https://github.com/bitcoin-s/bitcoin-s/pull/4456/files#diff-61413f0a707d1e3e49431ad34fcb1af04b4fa36148e75c7abc692bebc07822b7R36)
2. APIs that don't reflect asynchronous computation that occurs within a method.

This is an example of `2`. Your method return type is `Unit`, when in reality the method itself is async (`client`). 

While it may seem "easier" in the short term to use these tricks, (not reflecting async computations in the API type signature, using mutability liberally), it ends in tears on machines that have multi-core processors.

Thankfully we aren't working in C which doesn't have built in abstractions (`Future`) to handle this. We are using Scala, which has the capability to express this behavior.

Lets use the tools available to us.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -212,11 +220,11 @@ case class PeerManager(
 
     if (finder.hasPeer(peer)) {
       //one of the peers that we tried, failed to init within time, disconnect
-      finder.getData(peer).client.close()
+      finder.getData(peer).client.foreach(_.close())
```
Christewart: This is a `Future` that you are dropping on the floor.

Use `.map`, don't use `foreach` with `Future` as it encourages dropping `Future`'s on the floor.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -212,11 +220,11 @@ case class PeerManager(
...
     } else if (peerData.contains(peer)) {
       //this is one of our persistent peers which must have been initialized earlier, this can happen in case of
       //a reconnection attempt, meaning it got connected but failed to initialize, disconnect
-      peerData(peer).client.close()
+      peerData(peer).client.foreach(_.close())
```
Christewart: This is a `Future` that you are dropping on the floor.

Use `.map`, don't use `foreach` with `Future` as it encourages dropping `Future`'s on the floor.

node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -260,7 +268,7 @@ case class PeerManager(
               .map { _ =>
                 //could have already been deleted in case of connection issues
                 if (finder.hasPeer(peer))
-                  finder.getData(peer).client.close()
+                  finder.getData(peer).client.foreach(_.close())
```
Christewart: Use `.map`, don't use `foreach` with `Future` as it encourages dropping `Future`'s on the floor.

Christewart (changes requested):

solo1g:
I'll be looking into the failed compact filter header sync mac test and fix it in another PR. This PR, concerned with adding a supervisor, should be ready now.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -126,11 +139,30 @@ case class P2PClientActor(
       case metaMsg: P2PClient.MetaMsg =>
         sender() ! handleMetaMsg(metaMsg)
       case ExpectResponseCommand(msg) =>
-        handleExpectResponse(msg)
+        Await.result(handleExpectResponse(msg), timeout)
       case Terminated(actor) if actor == peerConnection =>
         reconnect()
     }
 
+  private def ignoreNetworkMessages(
```
Christewart, line 150: Can you add a scaladoc of why we would want to ignore certain network messages?

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -454,20 +491,25 @@ case class P2PClientActor(
...
               currentPeerMsgHandlerRecv.stopReconnect()
             context.stop(self)
         }
     }
   }
 
-  def handleExpectResponse(msg: NetworkPayload): Unit = {
-    currentPeerMsgHandlerRecv =
-      currentPeerMsgHandlerRecv.handleExpectResponse(msg)
+  def handleExpectResponse(msg: NetworkPayload): Future[Unit] = {
```
Christewart, line 518: Can you add a scaladoc to this with an example of an expected response, and what we do when we did not expect a response (I presume call `ignoreNetworkMessages`)

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
Christewart: So what happens if we receive a normal network message when we were expecting a _specific_ message.

For instance, 

1. I send an inventory message to request a bitcoin transaction 
2. A bitcoin block gets mined, and I receive a spontaneous `blockmsg`
3. My peer sends me the `txmsg` with the transaction i requested in (1). 

Does `2` (the block message) get dropped on the floor here?

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -46,7 +47,7 @@ class PeerMessageReceiver(
 
         val timeout =
```
Christewart: Can you rename this `timeout` to be someting like `initializationTimeoutChecker` or something like that?  

I that that name is much more representative of whats going on. You should change in `Initializing` too.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -271,21 +276,22 @@ class PeerMessageReceiver(
...
+        Future.unit
     }
   }
 
-  def handleExpectResponse(msg: NetworkPayload): PeerMessageReceiver = {
+  def handleExpectResponse(msg: NetworkPayload): Future[PeerMessageReceiver] = {
     state match {
       case good: Normal =>
         logger.debug(s"Handling expected response for ${msg.commandName}")
         val timeout =
```
Christewart: Please give a better name thatn `timeout`. Maybe `expectedResponseCancellable`?

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -130,17 +131,20 @@ case class PeerFinder(
             .filterNot(p => skipPeers().contains(p) || _peerData.contains(p))
 
           logger.debug(s"Trying next set of peers $peers")
-          peers.foreach(tryPeer)
+          val peersF = Future.sequence(peers.map(tryPeer))
+          Await.result(peersF, 10.seconds)
```
Christewart: Is there any benefit here to using `Await.result` compared to just mapping on `peersF` like so 

```scala 
peersF.failed.foreach(err => logger.error(s"Failed to connect to all peers,err"))
```

What is the significance of the `10.seconds` timeout? 

In my solution it doesn't block the thread at least, and I believe there is anything dependent on `peersF` being completed in 10 seconds?

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -166,19 +172,21 @@ case class PeerFinder(
     //delete try queue
     _peersToTry.clear()
 
-    _peerData.foreach(_._2.client.close())
+    val closeF = Future.sequence(_peerData.map(_._2.client.map(_.close())))
```
Christewart: nit: Sometimes it is better to just add a new line as it makes things more readable IMO. 

```scala
val closed = _peerData.map(_._2.client.map(_.close()))
val closedF = Future.sequence(closed)
```

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
```diff
@@ -166,19 +172,21 @@ case class PeerFinder(
...
 
-    AsyncUtil
+    val waitStopF = AsyncUtil
       .retryUntilSatisfied(_peerData.isEmpty,
```
Christewart, line 193: Why wait for `_peerData.isEmpty`, is it not good enough for `closeF` future to be completed?

Christewart (changes requested):
Overall good job, this is looking much more async safe! :tada:

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
solo1g: Not all requests expect a response, it's only for `getheaders`, `getcfilters`, `getcfheaders` atm. But I get what you want to ask, in such a case both are processed. But only a matching response cancels the timer, so suppose I asked for headers and got a block, it will process the block and keep on waiting for headers.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
solo1g: Nothing special about `10.seconds`, `Await` needed a duration and `10.second` is a generous one. But you are correct, there's nothing dependent on it.

As for why `Await`, the idea was that otherwise this block of code would be completed and scheduler will start the next countdown while things are still happening. With `Await`, the next countdown would only start once the future is done. ~~`Runnable` run in their own thread~~ so I felt using `Await` is good enough.

I'm, not really sure what would be better and to be fair, I think it doesn't really matter, both will do just fine. I will make this change, so that we are clearly sure that any `Await` that we have in our code base as far as `node` is concerned, is in an actor.

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
solo1g: So `peerData` map is such that the keys correspond to a running actor. The `close` called here has a return type of `Unit` and just sends a message to the client to stop the actor along with properly changing the state of `PeerMessageReceiver`. The actor actually stops sometime in the future, and the key-value pair is deleted from the map in the `postStop` callback of actors, so this is just an additional check to ensure that all actors have actually stopped.

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
solo1g: As for why and when we ignore messages, this only happens when we have started a disconnection from the peer. `initializeDisconnect` does not stop the actor in one actor message, it first sends a `Tcp.Close` to the `peerConnection` which when complete is intercepted by the actor and then the actor is stopped. The relevant part is that actor may process a message if its in the queue even while its disconnecting and that was causing some issues. We do not at all ignore any message in the usual operation.

Christewart:
These CI failures will be fixed in #4565 

https://github.com/bitcoin-s/bitcoin-s/runs/7672367084?check_suite_focus=true#step:5:3640

node/src/main/scala/org/bitcoins/node/networking/P2PClient.scala
```diff
@@ -454,20 +494,34 @@ case class P2PClientActor(
...
-      currentPeerMsgHandlerRecv.handleExpectResponse(msg)
+  /** For any [[org.bitcoins.core.p2p.NetworkPayload]], if it a subtype of [[ExpectsResponse]], starts a
+    * cancellable timer and changes state to [[PeerMessageReceiverState.Waiting]]. Note that any messages, received
+    * while waiting for a particular message are still processed, they just won't cancel the scheduled query timeout.
+    * Can only wait for one query at a time, other messages that are [[ExpectsResponse]] and received while in
+    * [[PeerMessageReceiverState.Waiting]] are still sent, but no query timeout would be executed for those.
+    * Currently, such a situation is not meant to happen.
+    */
+  def handleExpectResponse(msg: NetworkPayload): Future[Unit] = {
+    require(msg.isInstanceOf[ExpectsResponse],
```
Christewart: nit: It's always nice to put the unexpected thing you receive in the error message, i.e. 

```scala
    require(msg.isInstanceOf[ExpectsResponse],
            "Tried to wait for response to message which is not a query, got=$msg")
```

node/src/main/scala/org/bitcoins/node/PeerFinder.scala
Christewart: >As for why Await, the idea was that otherwise this block of code would be completed and scheduler will start the next countdown while things are still happening. With Await, the next countdown would only start once the future is done. Runnable run in their own thread so I felt using Await is good enough.

This is a good point. Maybe we should add 

```scala
    val isConnectionSchedulerRunning = new AtomicBoolean(false)
    system.scheduler.scheduleWithFixedDelay(
      initialDelay = initialDelay,
      delay = nodeAppConfig.tryNextPeersInterval) { () =>
          if (isConnectionSchedulerRunning.compareAndSet(false,true)) {
            
          logger.debug(s"Cache size: ${_peerData.size}. ${_peerData.keys}")
          if (_peersToTry.size < 32)
            _peersToTry.pushAll(getPeersFromDnsSeeds)

          val peers = (for { _ <- 1 to 32 } yield _peersToTry.pop()).distinct
            .filterNot(p => skipPeers().contains(p) || _peerData.contains(p))

          logger.debug(s"Trying next set of peers $peers")
          val peersF = Future.sequence(peers.map(tryPeer))
          peersF.onComplete {
            case Success(_) => 
              isConnectionSchedulerRunning.set(false)
            case Failure(err) => 
              isConnectionSchedulerRunning.set(false)
              logger.error(s"Failed to connect to peers", err)
          }
        } else {
            logger.warn(s"Previous connection scheduler is still running, skipping this run, it will run again in ${nodeAppConfig.tryNextPeersInterval}")
          }
    }
```

Note, the Scala compiler interprets anonymous functions as a `Runnable` if the type signature `() => Unit` so you don't have to do the overhead of `new Runnable { ... }`

Christewart (changes requested):

Christewart (approved):
Great job @shreyanshyad :tada:

PR #4524, opened 2022-07-20
Expose DBIOActions for various models in wallet, remove address queue

Christewart:
This PR does not intend to change functionality, only refactor and expose useful methods.

This exposes various `DBIOAction` for models in the wallet.

This also removes the address queue we implemented in #1009

I'm planning on introducing akka streams into the wallet and will implement the queue with that stream. 

Also now since the `generateScriptPubKey` call is using actions, the addresses are all generated in 1 database call rather than `n`. These are run sequentially against the database ensuring uniquness of the index of bip32 derivation path.

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
@@ -45,7 +45,10 @@ case class AddressDAO()(implicit
       ts: Vector[AddressRecord]): Future[Vector[AddressRecord]] =
     createAllNoAutoInc(ts, safeDatabase)
 
-  def create(addressDb: AddressDb): Future[AddressDb] = {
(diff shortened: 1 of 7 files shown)


wallet/src/main/scala/org/bitcoins/wallet/models/AddressDAO.scala
```diff
@@ -71,15 +74,18 @@ case class AddressDAO()(implicit
           .headOption
     } yield (addr, spk)
 
+    actions.map {
+      case (Some(addr), Some(spk)) => addr.toAddressDb(spk.scriptPubKey)
+      case _ =>
+        throw new SQLException(
+          s"Unexpected result: Cannot create either a address or a SPK record for $addressDb")
```
benthecarman, line 81: We should probably state if one was created and one wasn't

benthecarman (approved):
Otherwise Lgtm

Issue #4499, opened 2022-07-12
`loadwallet` should be logically identical to starting up the server with a wallet


Christewart:
Right now in #4417 we do not enapsulate all wallet startup logic. For instance, in #4326 we persist whether the wallet was rescanning when last shutdown. With `loadwallet`, if we load a wallet and find it was rescanning when we last shut it down, we should start rescanning again.

PR #4526, opened 2022-07-21
2022 07 21 Modify helper methods in `BitcoindRpcBackendUtil` to not materialize streams eagerly.

Christewart:
This re-implements internals helper methods of `BitcoindRpcBackendUtil` to allow us to connect stream pieces together rather than materializing streams inside of the helper method.

This gives will give us more control over eventually materializing streams and completing the stream as is needed in #4499

### DIFF
--- a/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
+++ b/app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
@@ -50,31 +50,7 @@ object BitcoindRpcBackendUtil extends Logging {
       heightRange <- {
         walletStateOpt match {
           case None =>
-            for {
-              txDbs <- wallet.listTransactions()
-              lastConfirmedOpt = txDbs
-                .filter(_.blockHashOpt.isDefined)
-                .lastOption
-              range <- lastConfirmedOpt match {
-                case None =>
-                  val range = (bitcoindHeight - 1).to(bitcoindHeight)
-                  Future.successful(range)
-                case Some(txDb) =>
-                  for {
-                    heightOpt <- bitcoind.getBlockHeight(txDb.blockHashOpt.get)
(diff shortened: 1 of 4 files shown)


app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -217,32 +206,27 @@ object BitcoindRpcBackendUtil extends Logging {
     pairedWallet
   }
 
-  private def filterSync(
-      blockHashes: Vector[DoubleSha256Digest],
+  private def filterSyncSink(
       bitcoindRpcClient: V19BlockFilterRpc,
-      wallet: Wallet)(implicit system: ActorSystem): Future[Unit] = {
+      wallet: Wallet)(implicit
+      system: ActorSystem): Sink[DoubleSha256Digest, Future[Wallet]] = {
```
Christewart: note: Returning `Sink` here now rather than the materialized stream represented by `Future[Unit]`. This allows us to connect this `Sink` to other stream pieces.

app/server/src/main/scala/org/bitcoins/server/BitcoindRpcBackendUtil.scala
```diff
@@ -86,49 +74,50 @@ object BitcoindRpcBackendUtil extends Logging {
...
-      val hashFs = Source(blockRange)
-        .mapAsync(numParallelism) {
-          bitcoind
-            .getBlockHash(_)
-            .map(_.flip)
-        }
-        .toMat(Sink.seq)(Keep.right)
-        .run()
+      wallet: Wallet)(implicit
+      system: ActorSystem): Future[Sink[Int, Future[Wallet]]] = {
```
Christewart: Same concept here, return the `Sink` rather than the materialized stream represented by `Future[Wallet]`

Issue #4528, opened 2022-07-21
Refactor duplicate data structures in `cli` and `app-server` into a common module `app-commons`


Christewart:
Here is an example of the duplicate data structure: `Rescan` 

We shouldn't have this defined in two different places! This is confusing and ripe for problems where we update code in one place, but forget to update the code in another place!

https://github.com/bitcoin-s/bitcoin-s/blob/a7aad4693409538d41dd2c7c2157461e45ff465f/app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala#L2619

https://github.com/bitcoin-s/bitcoin-s/blob/e90e372e54cf40f846cad3f4e6fc1932d4e765d9/app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala#L559

This should be moved into a module that both `appServer` and `cli` depend on. `app-commons` would make the most sense, specifically moving it into this file: https://github.com/bitcoin-s/bitcoin-s/blob/fdf281b46969bc56d8a71c7ce2de76f12868ad22/app-commons/src/main/scala/org/bitcoins/commons/rpc/Command.scala#L9

Christewart:
Done in #4541

PR #4541, opened 2022-07-25
Moving duplicate dataStructures in `cli` and `app-server` into `app-commons`

GreyMcCarthy:
Comes from this PR
https://github.com/bitcoin-s/bitcoin-s/issues/4528


app-commons/src/main/scala/org/bitcoins/commons/rpc/Command.scala
```diff
@@ -20,20 +39,1537 @@ trait AppServerCliCommand extends CliCommand {
...
+  override def defaultPort: Int = 9999
+}
+
+trait AcceptDLCCliCommand extends AppServerCliCommand{
+  override def defaultPort: Int = 9999
+}
+
+trait SendCliCommand extends AppServerCliCommand{
+  override def defaultPort: Int = 9999
+}
```
benthecarman: These should all get `defaultPort` from `AppServerCliCommand` so you dont need to override here

app-commons/src/main/scala/org/bitcoins/commons/rpc/Command.scala
```diff
@@ -20,20 +40,1603 @@ trait AppServerCliCommand extends CliCommand {
   override def defaultPort: Int = 9999
 }
 
+trait Broadcastable extends CliCommand {
+  override def defaultPort: Int = 9999
+}
+trait SignDLCCliCommand extends AppServerCliCommand {}
```
Christewart: You can remove the `{}` on these lines

app-commons/src/main/scala/org/bitcoins/commons/rpc/Command.scala
```diff
@@ -20,20 +40,1603 @@ trait AppServerCliCommand extends CliCommand {
...
+    Try(FinalizePSBT(jsToPSBT(jsArr.arr.head)))
+  }
+}
+
+case class ExtractFromPSBT(psbt: PSBT)
+    extends CliCommand
+    with AppServerCliCommand
+
+object ExtractFromPSBT
+    extends CliCommand
```
Christewart: All objects are extending too much, for instance, this is what this was `extending` before this PR

https://github.com/bitcoin-s/bitcoin-s/blob/e90e372e54cf40f846cad3f4e6fc1932d4e765d9/app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala#L469

Note, there is a difference between `case class` and `object`. 

The only objects in this file need to `extends ServerJsonModels`, that gives us access to methods like 

- `jsToPSBT`
- `jsToTx`
- `nullToOpt` 

etc

Christewart (changes requested):

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -2425,41 +2414,23 @@ object CliCommand {
...
-
-  object CreateContractInfo {
-
-    lazy val empty: CreateContractInfo = {
-      CreateContractInfo(announcementTLV = OracleAnnouncementV0TLV.dummy,
-                         totalCollateral = Satoshis.zero,
-                         contractDescriptor = ujson.Null)
-    }
-  }
+  /**  case class CreateContractInfo(
```
Christewart: Can be removed?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -2359,9 +2378,12 @@ object Config {
 
 object CliCommand {
 
-  trait Broadcastable {
-    def noBroadcast: Boolean
-  }
+  /**  trait Broadcastable {
```
Christewart: Can be removed?

app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -2469,79 +2440,14 @@ object CliCommand {
 
   case class RemoveDLCOffer(offerHash: Sha256Digest) extends AppServerCliCommand
 
-  sealed trait SendCliCommand extends AppServerCliCommand {
-    def destination: BitcoinAddress
-  }
+  /**  sealed trait SendCliCommand extends AppServerCliCommand {
```
Christewart: Can be removed?

app/server/src/main/scala/org/bitcoins/server/ServerJsonModels.scala
```diff
@@ -1,1102 +1,104 @@
...
-import scodec.bits.ByteVector
 import ujson._
 
 import java.io.File
 import java.net.{InetSocketAddress, URI}
 import java.nio.file.Path
 import scala.util._
 
-case class GetNewAddress(labelOpt: Option[AddressLabelTag])
+/** trait Broadcastable {
```
Christewart: Can be removed?

Christewart (changes requested):

Issue #4539, opened 2022-07-25
`appServerTest/testOnly *WebsocketTest` doesn't clean up database threads after running


Christewart:
This is what I see after running `appServerTest/testOnly *Websocket*` 

It seems like this was introduced in #4527 . If i run on 6b9d2db35072e620247cb1ef51d480f8719e04ab then things work fine and there are no threads left over after a test is run.

![Screenshot from 2022-07-25 06-51-25](https://user-images.githubusercontent.com/3514957/180771694-04f41a4d-f773-4ecd-b1c5-c78f980b593b.png)

PR #4544, opened 2022-07-27
Create a helper case class called `FundRawTxHelper`

Christewart:
The purpose of this class is to encapsulate the values returned by `fundRawTransactionInternal`. This makes it easier to work with `fundRawTransactionInternal` when refactoring code for #4539


core/src/main/scala/org/bitcoins/core/wallet/builder/FundRawTxHelper.scala
```diff
@@ -0,0 +1,19 @@
...
+    scriptSigParams: Vector[ScriptSignatureParams[InputInfo]],
+    feeRate: FeeUnit) {
+
+  /** Produces the unsigned transaction built by fundrawtransaction */
+  def unsignedTx: Transaction = txBuilderWithFinalizer.buildTx()
+
+  /** Produces a signed bitcoin transaction with the given fee rate */
+  def signedTx: Transaction = {
+    RawTxSigner.sign(unsignedTx, scriptSigParams, feeRate)
+  }
```
benthecarman: should make this a `lazy val` because it can be computationally expensive

PR #4547, opened 2022-07-28
Add branch and bound coin selection

### LINKED ISSUE #1650: Implement Branch and Bound `CoinSelectionAlgo`
Branch and bound is considered one of the best coin selection algorithms for privacy and block space efficiency. Would be great to have for the wallet.

http://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf

https://github.com/bitcoin/bitcoin/pull/10637

benthecarman:
Closes #1650

This required a small refactor because I found a bug in our utxo size approximation which would only return in bytes instead of the unit we needed for the `FeeUnit`

The implementation of branch of bound is a translation from the implementation in BDK. It could be made to be more scala-y but left that as a `// todo`

### DIFF
--- a/wallet-test/src/test/scala/org/bitcoins/wallet/ProcessTransactionTest.scala
+++ b/wallet-test/src/test/scala/org/bitcoins/wallet/ProcessTransactionTest.scala
@@ -150,7 +150,7 @@ class ProcessTransactionTest extends BitcoinSWalletTest {
       val receivingAddressF = wallet.getNewAddress()
       val amount = Bitcoins.one
 
-      val amtWithFee = amount + Satoshis(175) //for fee
+      val amtWithFee = amount + Satoshis(200) //for fee
 
       //build funding tx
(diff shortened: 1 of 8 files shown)


wallet-test/src/test/scala/org/bitcoins/wallet/ProcessTransactionTest.scala
```diff
@@ -150,7 +150,7 @@ class ProcessTransactionTest extends BitcoinSWalletTest {
       val receivingAddressF = wallet.getNewAddress()
       val amount = Bitcoins.one
 
-      val amtWithFee = amount + Satoshis(175) //for fee
+      val amtWithFee = amount + Satoshis(200) //for fee
```
benthecarman, line 153: We were previously under estimating the fee for this transaction

benthecarman:
Ready for review :)

PR #4548, opened 2022-07-28
Refactor so we don't create multiple WalletAppConfig in testkit

### LINKED ISSUE #4539: `appServerTest/testOnly *WebsocketTest` doesn't clean up database threads after running

Christewart:
Make sure we don't create multiple `WalletAppConfig` in the testkit harness so we avoid resource leaks. This is needed to fix #4539 

a6868ce - this commit removes the `bip39PasswordOpt` parameter everywhere in test fixtures and consolidates it to be set into one place -- `BaseWalletTest.getFreshConfig`.

### DIFF
--- a/testkit/src/main/scala/org/bitcoins/testkit/wallet/BaseWalletTest.scala
+++ b/testkit/src/main/scala/org/bitcoins/testkit/wallet/BaseWalletTest.scala
@@ -25,8 +25,12 @@ trait BaseWalletTest extends EmbeddedPg { _: Suite with BitcoinSAkkaAsyncTest =>
   }
 
   /** Wallet config with data directory set to user temp directory */
-  protected def getFreshConfig: BitcoinSAppConfig =
-    BaseWalletTest.getFreshConfig(pgUrl, Vector.empty)
+  protected def getFreshConfig: BitcoinSAppConfig = {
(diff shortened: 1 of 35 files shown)


testkit/src/main/scala/org/bitcoins/testkit/wallet/BaseWalletTest.scala
```diff
@@ -25,8 +25,12 @@ trait BaseWalletTest extends EmbeddedPg { _: Suite with BitcoinSAkkaAsyncTest =>
   }
 
   /** Wallet config with data directory set to user temp directory */
-  protected def getFreshConfig: BitcoinSAppConfig =
-    BaseWalletTest.getFreshConfig(pgUrl, Vector.empty)
+  protected def getFreshConfig: BitcoinSAppConfig = {
+    val bipPasswordOpt = KeyManagerTestUtil.bip39PasswordOpt
```
Christewart, line 29: there is where it get set now rather than being passed as a parameter everywhere

PR #4550, opened 2022-07-28
WIP: Fix rescan batching

rorp:
The current rescan logic process only first batch of addresses.  Here is a fix.

Needs changes to make the unit tests pass


Christewart:
Fixed in #4556

PR #4551, opened 2022-07-29
Bump lnd start up timeout

benthecarman:
Noticed that this was timing out on CI


Christewart:
Now it seems like it just times out the entire CI job? I'm not sure this is desirable, if its a bug in lnd no matter how long we wait for it to startup is going to matter

https://github.com/bitcoin-s/bitcoin-s/runs/7571747726?check_suite_focus=true

PR #4553, opened 2022-07-29
Improve DB disconnection handling

rorp:
`SafeDatabase` does not submit new actions to the underlying database after disconnect.


Christewart:
CI fails: https://github.com/bitcoin-s/bitcoin-s/runs/7583100291?check_suite_focus=true

PR #4555, opened 2022-07-30
Make DLCWallet.listDLCs use DBIOActions

### LINKED ISSUE #4552: Optimize `getdlcs` endpoint
It takes 9.2 seconds to complete on my umbrel. This wallet has > 30 DLCs in it.

Screenshot from 2022-07-29 04-49-18

benthecarman:
Fixes #4552

This should make this much more efficient. Ideally everything should be converted to use DBIOActions


benthecarman:
Ah, getting errors because some calls are in the DLC db while some are in the wallet db and we can't have a single db action across the two of them :/

benthecarman:
Looks like the 2 things we get from the wallet db is if it's an external address and the closing transaction. Probably best way to handle is to have an intermediary data type and then we read from the wallet db to get finish the DLCStatus

benthecarman:
Should be ready for review

Christewart (approved):

PR #4556, opened 2022-07-31
2022 07 29 fix multiple batch test

Christewart:
Fix broken unit test case that didn't test for discovering funds across multiple batches and then fix a bug introduced in #4211 where our `addressCount` would always be non zero.


Christewart:
Merging this with the understanding that #4563 is still an issue

PR #4562, opened 2022-08-02
Refactor to not use filterThenSum() when calculating balances

Christewart:
Refactor `getBalance()` to use `getBalance(account`) so our implementation is internally consistent rather than `getBalance()` having a different implementation than `getBalance(account)`. This should reduce the possibility for bugs.

On another note, this seems to fix `CallBackUtilTest` intermittent test failures, although i'm not sure why.


benthecarman (approved):
ACK on CI

PR #4572, opened 2022-08-04
Use DBIOActions to speed up processing transactions

benthecarman:
Every transaction requires at least one Future because we need to call the `ChainQueryApi` to get block hashes or heights but this should speed up processing.


Christewart (approved):
Nit: 

Just make a field for `safeDatabase` somewhere and re-use it across `TransactionProcessing`

```scala
private lazy val safeDatabase: SafeDatabase = transactionDAO.safeDatabase
``` 

somewhere

PR #4575, opened 2022-08-05
Make fundRawTransactionInternal use DBIOActions

benthecarman:
Changes `fundRawTransactionInternal` to use a single DBIOAction. This should improve speed and reliability. One notable improvement this does make is help prevent failures like we previously needed to here:

```scala
    resultF.recoverWith { case NonFatal(error) =>
      logger.error(
        s"Failed to reserve utxos for amount=${amt} feeRate=$feeRate, unreserving the selected utxos")
      // un-reserve utxos since we failed to create valid spending infos
      if (markAsReserved) {
        for {
          utxos <- selectedUtxosF
          _ <- unmarkUTXOsAsReserved(utxos.map(_._1))
        } yield error
      } else Future.failed(error)
    }
```

However, this does remove is a call to `walletCallbacks.executeOnReservedUtxos(logger, utxos)` when reserving through this. This could be added by instead making `fundRawTransactionInternalAction`  return a `DBIOAction[(FundRawTxHelper[ShufflingNonInteractiveFinalizer], Future[Unit])]` this is kind of an ugly way around it, otherwise not sure the best course of action (maybe just drop the future on the floor?)

Also made `safeDatabase` `protected` so it required a small refactor

### DIFF
--- a/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
+++ b/wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
@@ -78,7 +78,11 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
                                                   coinSelectionAlgo,
                                                   fromTagOpt,
                                                   markAsReserved)
-    safeDatabase.run(action)
+
+    for {
+      (txHelper, callbackF) <- safeDatabase.run(action)
+      _ <- callbackF
+    } yield txHelper
   }
 
   private[bitcoins] def fundRawTransactionInternalAction(
@@ -88,7 +92,7 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
       coinSelectionAlgo: CoinSelectionAlgo = CoinSelectionAlgo.LeastWaste,
       fromTagOpt: Option[AddressTag],
       markAsReserved: Boolean): DBIOAction[
-    FundRawTxHelper[ShufflingNonInteractiveFinalizer],
+    (FundRawTxHelper[ShufflingNonInteractiveFinalizer], Future[Unit]),
     NoStream,
     Effect.Read with Effect.Write with Effect.Transactional] = {
     val amts = destinations.map(_.value)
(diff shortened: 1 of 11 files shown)


Christewart:
>this does remove is a call to walletCallbacks.executeOnReservedUtxos(logger, utxos) when reserving through this

We definitely need this, we display the funds as reserved at the UI level while a user is waiting for their counterparty to accept a DLC.

benthecarman:
> > this does remove is a call to walletCallbacks.executeOnReservedUtxos(logger, utxos) when reserving through this
> 
> We definitely need this, we display the funds as reserved at the UI level while a user is waiting for their counterparty to accept a DLC.


Fixed to correctly use the callback

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
```diff
@@ -71,24 +72,52 @@ trait FundTransactionHandling extends WalletLogger { self: Wallet =>
...
+  }
+
+  private[bitcoins] def fundRawTransactionInternalAction(
+      destinations: Vector[TransactionOutput],
+      feeRate: FeeUnit,
+      fromAccount: AccountDb,
+      coinSelectionAlgo: CoinSelectionAlgo = CoinSelectionAlgo.LeastWaste,
+      fromTagOpt: Option[AddressTag],
+      markAsReserved: Boolean): DBIOAction[
+    (FundRawTxHelper[ShufflingNonInteractiveFinalizer], Future[Unit]),
```
Christewart: Lets just move the `Future[Unit]` into the `FundRawTxHelper`? That way it can have a descriptive named like `reservedUtxoCallbackF` or something like that.

Christewart (commented):
Otherwise LGTM, lets use the helper case class `FundRawTxHelper` now that we have it to consoldiate returns from the API

wallet/src/main/scala/org/bitcoins/wallet/internal/FundTransactionHandling.scala
benthecarman: done

Christewart (approved):

PR #4611, opened 2022-08-15
Add min & max functions for CurrencyUnits

benthecarman:
I find myself using Math.min or Math.max with CurrencyUnits and having to turn it into a double, and then back into a Satoshis which is pretty ugly, this should allow some cleanup


Christewart:
Doesn't seem to compile, maybe we should drop `2.12.x` in the `2.0.0` release :man_shrugging: . I do think this utility is valuable though as I believe i've needed the same thing.

benthecarman:
Seems we can already do this because we have `Ordering[CurrencyUnit]` defined

PR #4617, opened 2022-08-16
Implement `exportstakingaddresswif`

### LINKED ISSUE #4267: Implement the ability to spend from the staking address
We had a user deposit funds into this address and we don't allow spending from it currently. Idk we want to continue with #4265 or not if we have the spend functionality implemented :man_shrugging:

Christewart:
fixes #4267

Now a user can take this WIF, import it into another wallet, and recover funds in the staking address.

Example: 

```
 curl --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "exportstakingaddresswif", "params": []}' -H "Content-Type: application/json" http://127.0.0.1:9998/
{"result":"5Kcx1CqW6soAFAvDd32T4932BPZPUqYHmXzSxydhfVBKHoWxF1C","error":null}
```


benthecarman (approved):
Lgtm, probably worth adding a test that it is the correct private key

PR #4628, opened 2022-08-20
Add more JSON serializers

benthecarman:


app-commons/src/main/scala/org/bitcoins/commons/serializers/JsonWriters.scala
```diff
@@ -84,6 +115,14 @@ object JsonWriters {
             ("sequence", JsNumber(o.sequence.toLong))))
   }
 
+  implicit object TransactionOutPointWrites
+      extends OWrites[TransactionOutPoint] {
+
+    override def writes(o: TransactionOutPoint): JsObject = {
+      Json.obj(PicklerKeys.txIdKey -> o.txIdBE.hex, PicklerKeys.voutKey -> o.vout.toLong)
```
Christewart: Trying to use consistent json field names across the codebase, to avoid small discrepancies like `txid` and `txId`.

PR #4633, opened 2022-08-20
Get rid of standalone scalajs, mdoc CI matrix. Consolidate with other CI jobs

Christewart:
This should help clogging CI. Merge these two standalone jobs with existing ones so we don't have to go through VM overhead for these small jobs.


.github/workflows/Compile.yml
```diff
@@ -22,4 +22,4 @@ jobs:
           java-version: '17'
           cache: 'sbt'
       - name: Compile and Check Formatting
-        run: sbt -J-Xmx2g +test:compile scalafmtCheckAll
+        run: sbt -J-Xmx2g +test:compile scalafmtCheckAll docs/mdoc
```
benthecarman, line 24: should remove the `if: "! contains(github.event.pull_request.labels.*.name, 'documentation')"` clause

benthecarman (changes requested):

PR #4653, opened 2022-08-23
Adds logic to make sure we create network directories

Christewart:
When we changed out docker logic in #4601 it appears our startup logic will no longer create the directories we are specifying in the docker entrypoint with the `--datadir` flag. Now the USER we are running under doesn't necessarily correspond to the path we are specifying in `--datadir`.


benthecarman:
should this be handled by migrating the database?

https://github.com/bitcoin-s/bitcoin-s/blob/7e214da6012a8b7fe04a2912ab75e0ebfafa80f3/db-commons/src/main/scala/org/bitcoins/db/SQLiteUtil.scala#L74

Issue #4645, opened 2022-08-23
Node syncing fails `Failing because buffer is full and overflowStrategy is: [Fail] in stream [queueSource]`


Christewart:
This occurred on e65346f5

It seems to me our downstream consumers of the queue are not fast enough for the amount data we are fetching, leading to failure here: 

https://github.com/bitcoin-s/bitcoin-s/blob/2c2e03b2799d1a6807eb8f3e513c0acbb4eb45fc/node/src/main/scala/org/bitcoins/node/PeerManager.scala#L378

```
2022-08-23T11:55:36UTC INFO [DLCWallet$DLCWalletImpl] Processing 1000 block filters for 12 utxos and 4067 scripts
2022-08-23T11:55:41UTC INFO [DataMessageHandler] Processing 1000 filters
2022-08-23T11:55:44UTC INFO [DLCWallet$DLCWalletImpl] Finished processing of block=00000000000000000000286b871b7a20160e80490416ad1d7eea6ce4fa204f2e. It took 20472ms
2022-08-23T11:55:48UTC INFO [DLCWallet$DLCWalletImpl] Processing block=000000000000000000065433f3a67c2d15f56c6947ca358f887f98580c3f27cd heightOpt=Some(712904)
2022-08-23T11:55:48UTC INFO [ChainHandler] Processed filters from height=715724 to 716723. Best filter.blockHash=000000000000000000036f0aff5c19a7ce4d177fb427cbf60b37582c3d43b760
2022-08-23T11:55:48UTC INFO [DataMessageHandler] Received maximum amount of filters in one batch. This means we are not synced, requesting more
2022-08-23T11:55:48UTC INFO [PeerMessageSender] Requesting compact filters from FilterSyncMarker(startHeight=716724, stopBlockHash=00000000000000000004697dd8b94025f5df4ac9bb5122f27e399bb8b0266e74)
2022-08-23T11:55:48UTC INFO [DataMessageHandler] Received block message with hash 000000000000000000097923ab89a63cfe440f54a5b79c17d35ce1d03f94c358
2022-08-23T11:55:48UTC INFO [DataMessageHandler] Received block message with hash 0000000000000000000acecfa1ce6bfb7b1e2fcbfe7e809ee06ebc9db2c5ef9e
2022-08-23T11:55:48UTC INFO [DLCWallet$DLCWalletImpl] Processing 1000 block filters for 12 utxos and 4067 scripts
[...]

PR #4654, opened 2022-08-23
Change stream to use `OverflowStrategy.backpressure`

solo1g:
* Decreases buffer size to 1500 from 10,000 as that was causing `OutOfMemoryError`
* Stream now uses backpressure as overflow strategy.


Christewart:
Related to #4645

Christewart (approved):

PR #4667, opened 2022-08-25
Fix infinite invalid header loop

solo1g:
@Christewart @nkohen 

How it re-queries headers is that we make a request, and ask again if we don't get a response in 10 seconds.
I have tested the laptop sleep issue and it works for that on my end.

A new issue that I noticed while testing this is that if the internet goes down, every 10 second we send a `getheaders` and when we get back online, all of those messages are sent at once.

So the first response that we send would give us the valid headers, for the remaining we will get a duplicate exception, and rest should proceed fine. This behavior I suppose needs to change, instead of continuous request, maybe try `ping` . As for this PR, I believe everything here is a strict improvement.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -384,11 +384,11 @@ case class DataMessageHandler(
 
         getHeadersF.recoverWith {
           case _: DuplicateHeaders =>
-            logger.debug(
+            logger.warn(
               s"Received duplicate headers from ${syncPeer.get} in state=$state")
             Future.successful(this)
           case _: InvalidBlockHeader =>
-            logger.debug(
+            logger.warn(
               s"Invalid headers of count $count sent from ${syncPeer.get} in state=$state")
             recoverInvalidHeader(peer, peerMsgSender)
           case e: Throwable => throw e
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -1,7 +1,7 @@
 package org.bitcoins.node.networking.peer
 
 import akka.Done
-import org.bitcoins.chain.blockchain.InvalidBlockHeader
+import org.bitcoins.chain.blockchain.{DuplicateHeaders, InvalidBlockHeader}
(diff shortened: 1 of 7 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -391,6 +382,18 @@ case class DataMessageHandler(
               }
             }
 
+        getHeadersF.recoverWith {
```
solo1g, line 385: For any invalid header, `chainApi` recovers and sends re-queries for headers. This future is then chained to to `getHeadersF` which again asks for headers. Two similar header requests are sent, the first one is processed the second one is invalid, again causing 2 requests.

Now the failure propagates all the way to end of chain.

node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageReceiver.scala
```diff
@@ -294,7 +303,8 @@ class PeerMessageReceiver(
         logger.debug(s"Handling expected response for ${msg.commandName}")
         val expectedResponseCancellable =
           system.scheduler.scheduleOnce(nodeAppConfig.queryWaitTime)(
-            Await.result(onResponseTimeout(msg), 10.seconds))
+            Await.result(node.peerManager.sendResponseTimeout(peer, msg),
```
solo1g, line 306: Needed as this future needs to be called with the then (after 10s) state. Not sure if there's a better way.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
Christewart: A bug introduced by yours truely, sorry about that.

solo1g: Not really :sweat_smile:, this was present before your refactor too.

node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -391,6 +382,18 @@ case class DataMessageHandler(
...
+          case _: DuplicateHeaders =>
+            logger.debug(
```
Christewart: It might be worth `warn` here, as we don't want to see this happen too much right?

Christewart (commented):
Besides the logging comment, i think this LGTM. I'm testing across environments now.

PR #4676, opened 2022-08-26
Fix the CET calculator to support hyperbolic functions

rorp:
First outcome of a hyperbolic curve or many first outcomes, depending on `translate_outcome`) can throw a division by zero exception. This PR changes `CETCalculator` to let it find the first valid outcome and use it as a min value for CET calculations.


nkohen:
nack: I don't understand what the goal is here, if you don't have defined values in the beginning of a DLCPayoutCurve then the problem is with the DLCPayoutCurve and a better one needed to be constructed. I'm not opposed to better error messages in this case but changing it to just omit things for early values would mean that the money gets stuck in these cases and furthermore is not spec compliant l.

Perhaps what you meant to do is add a require at the top of this function to check that the curve is defined as 0 and throw a nice error if it isn't?

PR #4677, opened 2022-08-28
Add `BITCOIN_S_UID`, explicit volume in `docker-compose.yml`

Christewart:
This makes our `docker-compose.yml` flexible with the `uid` that the docker containers get run as. Now the `uid` on the host machine will be what runs the containers. You can run the `docker-compose.yml` with 

```
APP_PASSWORD=topsecret BITCOIN_S_UID="$(id -u):$(id -g)" docker-compose up
```

Currently this PR does not write to the old volume location in `data/wallet`. For some reason i can't seem to create child directories for volumes inside of `$(pwd)`. I'm hoping someone answers this question: https://stackoverflow.com/questions/73520753/failure-to-mount-local-volume-when-the-volume-is-a-child-directory-of-pwd\

If you ran our `docker-compose.yml` before this PR is merged, you will need to run 

```
id -u
sudo chown -R [id-output-from-id-u] data/
```
to give permissions over the `data/` directory to the correct user.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     restart: on-failure
     stop_grace_period: 1m
     volumes:
-      - bitcoin-s-datadir:/bitcoin-s
+      - bitcoin-s-frontend-log:/bitcoin-s
     environment:
       LOG_PATH: "/bitcoin-s/log/"
       BITCOIN_S_HOME: "/bitcoin-s/"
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -48,3 +48,13 @@ services:
...
+
+volumes:
+  bitcoin-s-datadir:
+    name: "bitcoin-s"
+    driver: local
+    external: false
+    driver_opts:
+      o: bind
+      type: none
+      device: ${PWD}
```
Christewart: For backwards compatability, this should be `${PWD}/data`, but i can't figure out how to do it: https://stackoverflow.com/questions/73520753/failure-to-mount-local-volume-when-the-volume-is-a-child-directory-of-pwd/73528609#73528609

Christewart:
Superseded by #4680

PR #4679, opened 2022-08-29
use real docker volumes rather than bind mounts

Christewart:
This is an alternative to #4677 

This uses real docker volumes rather than bind mounts for our `docker-compose.yml`. Here are the advantages of using volumes [according to docker's own site](https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose).

Screenshot from 2022-08-29 13-35-33

This makes us less sensitive to host specific configurations. 

### Migrating an existing bind mount into a volume

This is only needed if you have an existing datadir you would like to migrate. If you are a new user this is not necessary.

```
 $ docker volume create bitcoin-s_bitcoin-s
bitcoin-s_bitcoin-s
#copy contents of existing bind mount into volume 
$ sudo cp -R data/wallet/* /var/lib/docker/volumes/bitcoin-s_bitcoin-s/_data/
# get your user ids, so we can switch permissions correctly. My uid is 1001. Yours might be different.
$ id -u 
1001
$ id -g 
1001
# use ids from your machine rather than my machines id of 1001:1001
sudo chown -R 1001:1001  /var/lib/docker/volumes/bitcoin-s/_data/
```

and now startup your service with 

>APP_PASSWORD=topsecret BITCOIN_S_UID="$(id -u):$(id -g)" docker-compose up

and all your old information should show up in your new docker volume! :tada:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,14 +3,13 @@ version: "3.7"
 services:
   web:
     image: bitcoinscala/wallet-server-ui:latest
-    user: "0:1000"
+    user: 1000:1000
     restart: on-failure
     stop_grace_period: 1m
     volumes:
-      - ./data/wallet:/bitcoin-s
-      - ./data/log:/log
+      - bitcoin-s:/bitcoin-s
     environment:
-      LOG_PATH: "/log/"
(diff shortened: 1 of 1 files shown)


Christewart:
Superseded by #4680

PR #3497, opened 2021-08-06
WIP: 2021 06 30 scala3

Christewart:
Bored work on the airplane

The entire project does not compile, but `crypto`, `core` do compile

PR #4704, opened 2022-09-02
Drop support for scalac `2.12.x`

Christewart:
Only support scalac `2.13.x`. Scala `2.12.x` is old and i'm not aware of anyone using the 2.12 version of this library. This will also make future migrations to scala 3 easier as well (#3497). 

This has a side benefit of allowing us to use Scala `2.13.x` APIs.


benthecarman:
There are some scala 2.12 specific files you'll need remove too

example:
https://github.com/bitcoin-s/bitcoin-s/blob/master/core/src/main/scala-2.12/org/bitcoins/core/compat/CompactPackage.scala

PR #4724, opened 2022-09-06
Update akka deps

benthecarman:
https://akka.io/blog/news/2022/09/06/akka-2.6.20-released


Christewart:
I think this along with #4620 should mean most of our scala deps are on Scala `2.13.7+` which is android friendly!

Christewart (approved):

PR #4728, opened 2022-09-07
Fix bug where no exception was thrown if the acceptor did not have enough money

### LINKED ISSUE #4727: Give error message to UI if a user tries to accept an offer and doesn't have enough money
I tried to do a DLC with a user on 1.9.3 this morning. I sent them an offer, and they tried to accept my offer multiple times. 

I saw their connection attempts show up in my logs, but no data came thorugh.

The root cause of this was because the user did not deposit sats into their wallet.

I was suprised that the endpoint for accepting an offer sent to you didn't get an error message. We need to do this.

Screenshot from 2022-09-07 10-27-00

Finally after depositing sats, things worked as expected

Screenshot from 2022-09-07 10-32-23

Christewart:
fixes #4727

This fixes an interesting bug in the wallet for enum contracts.

Previously, we would require that the `max` value in the contract be the most amount of paid to one counterparty. 

This doesn't make sense, because we should always have the total collateral assigned in the enum contract. I.e.

```
total_collateral = offer_collateral + accept_collateral
```

before this PR, for enum contracts, we had

```
max_payout = offer_collateral + accept_collateral
```

but `max_payout == total_collateral` is not necessarily true. For instance, imagine i'm doing a coinflip bet with an enum contract for `1BTC` where the payouts are

- Heads Alice receives `0.75BTC`, Bob receives `0.25BTC`
- Tails Bob receives `0.75BTC`, Alice receives `0.25BTC`.

The `max_payout=0.75BTC`, but the total collateral is `1BTC`.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
@@ -42,7 +42,7 @@ sealed trait ContractInfo extends TLVSerializable[ContractInfoTLV] {
   def contractDescriptors: Vector[ContractDescriptor]
 
   /** Returns the maximum payout the offerer could win from this contract */
-  def max: Satoshis
+  def maxOffererPayout: Satoshis
 
   /** Computes the CET set and their corresponding payouts using CETCalculator. */
   def allOutcomesAndPayouts: Vector[(OracleOutcome, Satoshis)]
@@ -220,7 +220,7 @@ case class SingleContractInfo(
   }
 
   /** @inheritdoc */
-  override val max: Satoshis = {
+  override val maxOffererPayout: Satoshis = {
     contractDescriptor match {
       case descriptor: EnumContractDescriptor =>
         descriptor.values.maxBy(_.toLong)
@@ -406,7 +406,8 @@ case class DisjointUnionContractInfo(contracts: Vector[SingleContractInfo])
   }
 
   /** @inheritdoc */
-  override val max: Satoshis = contracts.map(_.max).max
+  override val maxOffererPayout: Satoshis =
+    contracts.map(_.maxOffererPayout).max
 
   /** @inheritdoc */
   override lazy val allOutcomesAndPayouts: Vector[(OracleOutcome, Satoshis)] = {
--- a/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
@@ -220,7 +220,14 @@ case class SingleContractInfo(
   }
 
   /** @inheritdoc */
-  override val max: Satoshis = totalCollateral
+  override val max: Satoshis = {
+    contractDescriptor match {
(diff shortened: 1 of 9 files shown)


dlc-wallet-test/src/test/scala/org/bitcoins/dlc/wallet/WalletDLCSetupTest.scala
```diff
@@ -982,6 +982,30 @@ class WalletDLCSetupTest extends BitcoinSDualWalletTest {
...
   }
 
+  it must "fail to accept an offer when you do not have enough money in the wallet" in {
+    wallets =>
+      val walletA = wallets._1.wallet
+      val walletB = wallets._2.wallet
+
+      val offerData: DLCOffer =
+        DLCWalletUtil.buildDLCOffer(totalCollateral = Bitcoins(100))
+      //val walletBBalanceF = walletB.getBalance()
```
benthecarman: nit can remove comment

core/src/main/scala/org/bitcoins/core/protocol/dlc/models/ContractInfo.scala
```diff
@@ -220,11 +220,7 @@ case class SingleContractInfo(
   }
 
   /** @inheritdoc */
-  override val max: Satoshis = contractDescriptor match {
-    case descriptor: EnumContractDescriptor =>
-      descriptor.values.maxBy(_.toLong)
-    case _: NumericContractDescriptor => totalCollateral
-  }
+  override val max: Satoshis = totalCollateral
```
Christewart: So now that i'm looking at the scaladoc for this, it appears that this might have been implemented correctly, but we used it in teh wrong place: 

https://github.com/Christewart/bitcoin-s-core/blob/a33e8b0ad0a3fcff1783dc7e9fd3be5d77891f2d/dlc-wallet/src/main/scala/org/bitcoins/dlc/wallet/DLCWallet.scala#L695

This should be

```scala
    val collateral = offer.contractInfo.totalCollateral - offer.collateral
```

thoughts @benthecarman @nkohen ?

benthecarman: doesn't this go against the whole PR's description of changing it not to be `max_payout = offer_collateral + accept_collateral`

benthecarman: yeah i think that code above is correct and that line you linked needs to be fixed

Christewart: I can revert this change if thats the case and change `DLCWallet.scalaL695`

benthecarman: should do a pass of places we call `contractInfo.max` and see if it should be totalCollateral instead

Christewart: Maybe rename `max` -> `maxOffererPayout` ? Although it doesn't really work for numeric contracts :/

benthecarman (approved):

PR #4729, opened 2022-09-08
Add handling so test lnd clients can be reached by docker

benthecarman:
To access the lnd rpc from a docker container we need to be bound to `0.0.0.0`, this makes it so our default test ones do that, as well as adds the tls domain to allow requests from it.

### DIFF
--- a/lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
+++ b/lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
@@ -131,8 +131,12 @@ class LndRpcClient(val instance: LndInstance, binaryOpt: Option[File] = None)(
       case None         => None
     }
 
+    val host =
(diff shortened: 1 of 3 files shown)


lnd-rpc/src/main/scala/org/bitcoins/lnd/rpc/LndRpcClient.scala
```diff
@@ -131,8 +131,12 @@ class LndRpcClient(val instance: LndInstance, binaryOpt: Option[File] = None)(
       case None         => None
     }
 
+    val host =
+      if (instance.rpcUri.getHost == "0.0.0.0") "localhost"
```
Christewart: So it will never be able to bind to `0.0.0.0`? 

IIUC this correctly it will substitute `0.0.0.0` -> `localhost`?

benthecarman: This is if it is bound to 0.0.0.0, we don't attempt to make requests to it, we instead make it to local host

Issue #2371, opened 2020-12-16
bitcoin-s server runs database migrations twice on startup


rorp:
Probably, the server should do it just once before everything else.

See the log.
```
2020-12-16T18:17:12UTC INFO [VersionPrinter] Flyway Community Edition 6.4.2 by Redgate
2020-12-16T18:17:12UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/Users/roman/.bitcoin-s/mainnet/chaindb.sqlite (SQLite 3.34)
2020-12-16T18:17:12UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/Users/roman/.bitcoin-s/mainnet/walletdb.sqlite (SQLite 3.34)
2020-12-16T18:17:12UTC INFO [DatabaseFactory] Database: jdbc:sqlite:/Users/roman/.bitcoin-s/mainnet/nodedb.sqlite (SQLite 3.34)
2020-12-16T18:17:12UTC INFO [DbValidate] Successfully validated 2 migrations (execution time 00:00.013s)
2020-12-16T18:17:12UTC INFO [DbValidate] Successfully validated 5 migrations (execution time 00:00.013s)
2020-12-16T18:17:12UTC INFO [DbValidate] Successfully validated 9 migrations (execution time 00:00.015s)
2020-12-16T18:17:12UTC INFO [DbMigrate] Current version of schema "main": 5
2020-12-16T18:17:12UTC INFO [DbMigrate] Current version of schema "main": 2
2020-12-16T18:17:12UTC INFO [DbMigrate] Current version of schema "main": 9
2020-12-16T18:17:12UTC INFO [DbMigrate] Schema "main" is up to date. No migration necessary.
2020-12-16T18:17:12UTC INFO [DbMigrate] Schema "main" is up to date. No migration necessary.
2020-12-16T18:17:12UTC INFO [DbMigrate] Schema "main" is up to date. No migration necessary.
2020-12-16T18:17:13UTC INFO [BitcoinSServerMain] Chain work already calculated
2020-12-16T18:17:13UTC INFO [BitcoinSServerMain] Initialized chain api
2020-12-16T18:17:13UTC INFO [BitcoinSServerMain] Using fee provider: MempoolSpaceProvider(HourFeeTarget)
2020-12-16T18:17:13UTC INFO [BitcoinSServerMain] Done configuring wallet
2020-12-16T18:17:13UTC INFO [BitcoinSServerMain] Done configuring node
[...]

Issue #4606, opened 2022-08-15
Fail to startup first time running when selecting wallet name descriptor


Christewart:
Seen on  d0dadfa423d21f21b5a199147a692265e03c1fbd , when i first started with `bitcoins-s.node.mode=neutrino` I receivd this error message. It seems to happen everytime i attempt to start and can't get past it. This bug does not appear with `bitcoin-s.node.mode=bitcoind`

```
2022-08-15T13:02:23UTC ERROR [SafeDatabase] Error when executing query delete [delete from "state_descriptors" where "state_descriptors"."type" = 'WalletName']
2022-08-15T13:02:23UTC ERROR [SafeDatabase] org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: state_descriptors)
2022-08-15T13:02:23UTC ERROR [BitcoinSServerMain] Failed to startup server!
org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (no such table: state_descriptors)
        at org.sqlite.core.DB.newSQLException(DB.java:1030)
        at org.sqlite.core.DB.newSQLException(DB.java:1042)
        at org.sqlite.core.DB.throwex(DB.java:1007)
        at org.sqlite.core.NativeDB.prepare_utf8(Native Method)
        at org.sqlite.core.NativeDB.prepare(NativeDB.java:137)
        at org.sqlite.core.DB.prepare(DB.java:257)
        at org.sqlite.core.CorePreparedStatement.<init>(CorePreparedStatement.java:45)
        at org.sqlite.jdbc3.JDBC3PreparedStatement.<init>(JDBC3PreparedStatement.java:30)
        at org.sqlite.jdbc4.JDBC4PreparedStatement.<init>(JDBC4PreparedStatement.java:25)
        at org.sqlite.jdbc4.JDBC4Connection.prepareStatement(JDBC4Connection.java:35)
        at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:241)
        at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:205)
        at com.zaxxer.hikari.pool.ProxyConnection.prepareStatement(ProxyConnection.java:311)
        at com.zaxxer.hikari.pool.HikariProxyConnection.prepareStatement(HikariProxyConnection.java)
[...]

PR #4764, opened 2022-09-11
Attempt to reduce redundant work on startup to speed up start up time

Christewart:
This PR attempts to reduce redundant work on startup to decrease startup time.

This does this by reducing redundant calls to `.migrate()`. Now we just call `.migrate()` on `*AppConfig`'s that are dependent on tor. This is because when we return from `BitcoinSAppConfig.start()`, the configs that are guaranteed to be fully started are non tor configs (`WalletAppConfig`, `KeyManagerAppConfig`, `ChainAppConfig`).

Related to #4606 / #4607


benthecarman:
This might be worth looking into as well

#2371

Christewart:
Didn't know about that, i think this maybe fixes #2371 . I'll look closer

PR #4780, opened 2022-09-13
Add support for mac m1 lnd rpc

### LINKED ISSUE #3878: `lndRpcTest` hangs on mac
I'm running on 9c9a0a618f09523e0637b701e59a890f4e2cbdff and `lndRpcTest/test` seems to hang

Screen Shot 2021-12-06 at 5 27 54 PM

it does seem like binaries are started on the underlying OS

Screen Shot 2021-12-06 at 5 28 37 PM

benthecarman:
Fixes #3878
Previously we were downloading the wrong lnd for m1 macs. I don't have an m1 mac so if someone could test that it previously didn't work (`sbt lndRpcTest/test`) but now does, that would be greatly appreciated

### DIFF
--- a/lnd-rpc/lnd-rpc.sbt
+++ b/lnd-rpc/lnd-rpc.sbt
@@ -43,6 +43,8 @@ TaskKeys.downloadLnd := {
 
   val (platform, suffix) =
     if (Properties.isLinux) ("linux-amd64", "tar.gz")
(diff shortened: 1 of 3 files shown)


Christewart:
fixes #3878 :tada:

Christewart (approved):

PR #4781, opened 2022-09-15
Set db auth in slick db config

benthecarman:
This lets us set the db credentials from a source (aws secrets) other than the actual config by overriding `dbUsername` and `dbPassword`.

### DIFF
--- a/db-commons/src/main/scala/org/bitcoins/db/DbAppConfig.scala
+++ b/db-commons/src/main/scala/org/bitcoins/db/DbAppConfig.scala
@@ -19,6 +19,12 @@ abstract class DbAppConfig extends AppConfig {
     Future.successful(slickDbConfig.db.close())
   }
 
+  lazy val dbUsername: String =
+    config.getString(s"bitcoin-s.$moduleName.db.user")
+
+  lazy val dbPassword: String =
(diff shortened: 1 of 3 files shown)


Christewart (commented):
Can any of this stuff be removed if we add this change? 

https://github.com/bitcoin-s/bitcoin-s/blob/4f7b6422ea3390f276733ee3c0fb21d426bde751/db-commons/src/main/scala/org/bitcoins/db/DbAppConfig.scala#L92

benthecarman:
> Can any of this stuff be removed if we add this change? 
> 
> 
> 
> https://github.com/bitcoin-s/bitcoin-s/blob/4f7b6422ea3390f276733ee3c0fb21d426bde751/db-commons/src/main/scala/org/bitcoins/db/DbAppConfig.scala#L92

I don't see the change you're talking about?

Christewart:
> > Can any of this stuff be removed if we add this change?
> > https://github.com/bitcoin-s/bitcoin-s/blob/4f7b6422ea3390f276733ee3c0fb21d426bde751/db-commons/src/main/scala/org/bitcoins/db/DbAppConfig.scala#L92
> 
> I don't see the change you're talking about?

Err.... i don't either 😅 . Don't code review before ☕  in the morning 😂

PR #4791, opened 2022-09-20
Have `EmbeddedPg` check `PG_ENABLED` value

benthecarman:
This way you can set `PG_ENABLED=false` and it'll actually disabled postgres


rorp (approved):
👍

Christewart (approved):

PR #4814, opened 2022-10-03
Use `ByteVector.compare` rather than rolling our own

Christewart:
The old implementation doesn't work for anything past the first byte.


benthecarman (approved):
I think the previous version would work for past the first byte with the last conditional, however ACK for just using scodec's version

PR #4844, opened 2022-10-15
Update to lnd 0.15.2

Christewart:
https://github.com/lightningnetwork/lnd/releases/tag/v0.15.2-beta


benthecarman:
This is going to have the same issue as #4687 because the bug with fee estimation is still there.

They are about to cut a release for 0.15.3 that should fix this, could be worth just waiting until then

PR #4852, opened 2022-10-19
Add generate helper function for bitcoind

benthecarman:
we do `bitcoind.getNewAddress.flatMap(bitcoind.generateToAddress(1, _))` everywhere, this makes a helper function for this

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -275,6 +275,13 @@ class BitcoindRpcClient(override val instance: BitcoindInstance)(implicit
     }
   }
 
+  def generate(numBlocks: Int): Future[Vector[DoubleSha256DigestBE]] = {
+    for {
+      addr <- getNewAddress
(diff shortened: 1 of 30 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -275,6 +275,13 @@ class BitcoindRpcClient(override val instance: BitcoindInstance)(implicit
     }
   }
 
+  def generate(numBlocks: Int): Future[Vector[DoubleSha256DigestBE]] = {
```
Christewart, line 278: Might want to provide an optional address here if you want to fund a wallet

```scala
def generateToAddress(numBlocks: Int, addressOpt: Option[BitcoinAddress] = None): Futur[Vector[DoubleSha256DigestBE]]
```

Christewart: Eh, actually if you keep the old `generateToAddress` around maybe it isn't worth it 🤷‍♂️

benthecarman: Yeah that's what generate to address is for

PR #4865, opened 2022-10-30
Upgrade sbt ci release to 1.5.11

Christewart:
Supersedes #4856 , it seems that `sbt-git` is no longer bundled with sbt ci release so I needed to remove usage of `SbtGit` keys.


benthecarman:
Looks like it was just updated and the package name is different now: https://github.com/sbt/sbt-ci-release/pull/236/files

PR #4868, opened 2022-11-01
Add test for another tx that broke lnd/btcd

benthecarman:
https://mempool.space/tx/73be398c4bdc43709db7398106609eea2a7841aaf3a4fa2000dc18184faa2a7e


Christewart (approved):

Christewart:
I suppose one of these days a weirdly crafted tx will get our parsing logic to fail, but today is not that day! 🎉

PR #4869, opened 2022-11-01
Handle failed getting version of bitcoind

benthecarman:
@tompro reported that he was getting failures when trying to get the bitcoind version. We should just default to the newest version in this case.

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/config/BitcoindInstance.scala
@@ -13,7 +13,7 @@ import java.io.File
 import java.net.URI
 import java.nio.file.{Files, Path, Paths}
 import scala.sys.process._
-import scala.util.Properties
+import scala.util.{Failure, Properties, Success, Try}
 
 /** Created by chris on 4/29/17.
   */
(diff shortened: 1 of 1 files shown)


benthecarman:
getting this when trying to run
image

seems this is not our fault, however, I still think this would be a useful change

Christewart:
> getting this when trying to run image
> 
> seems this is not our fault, however, I still think this would be a useful change

@tompro Can you try closing your terminal and re-opening it? Sometimes on mac downloaded binaries are not valid in the current terminal session. If you create a new terminal, it sometimes just works :tm: . I'm not sure why.

tompro:
@Christewart I already did open new terminals, also added the permission exception in MacOS.

Christewart:
> @Christewart I already did open new terminals, also added the permission exception in MacOS.

What did you have to run ? Was it something similar to this? 

https://github.com/bitcoin-s/bitcoin-s/blob/2001e86ab94019dd468aaa658695620626e19e98/app/server/src/universal/wallet-server-extra-startup-script.sh#L18

tompro:
You can add the exception in Privacy & Security settings UI right after you launch the bin. The message above is not from a launch via SBT but directly launching the bitcoind bin (freshly downloaded from https://bitcoincore.org/bin/bitcoin-core-23.0/bitcoin-23.0-arm64-apple-darwin.tar.gz) . I was just verifying if bitcoind is executable at all. It behaves the same as the bin downloaded via SBT.

Christewart (commented):
So did @tompro build bitcoin from source resulting ina version of that was something like `23.99`? I'm confused otherwise how he got an invalid version from bitcoincore.org

tompro:
Nope I just ran the bin as downloaded. Reason might be m2 or recent MacOs update:
image

Christewart:
> Nope I just ran the bin as downloaded. Reason might be m2 or recent MacOs update: image

oooh! Nice, i hear the m2 chip has a ~30% performance improvement over the m1 chip (!!!). 

I only have an m1, so you may experience other weird things. Please report bugs if you find them :-)

Christewart (approved):

PR #4870, opened 2022-11-01
Remove legacy GUI, this fixes CI merges to master

Christewart:
We've talked about removing this now since we have an electron app, as a by product this fixes our CI builds that are failing when we notarize the bundle with apple. 

https://github.com/bitcoin-s/bitcoin-s/actions/runs/3371392670/jobs/5593552937#step:7:1065


benthecarman:
RIP

PR #4900, opened 2022-11-26
Fix isLocalhost for 127.0.0.1

benthecarman:
Curious on @rorp's thoughts here as he made this I believe


rorp:
Good catch! That was a typo, apparently.

Christewart (approved):

PR #4902, opened 2022-11-27
Add bitcoind v24

benthecarman:
https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-24.0.md

### DIFF
--- a/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
+++ b/bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
@@ -370,7 +370,7 @@ sealed trait BitcoindVersion
 object BitcoindVersion extends StringFactory[BitcoindVersion] with Logging {
 
   /** The newest version of `bitcoind` we support */
-  val newest: BitcoindVersion = V24
+  val newest: BitcoindVersion = V23 // todo make v24
 
   val standard: Vector[BitcoindVersion] =
     Vector(V19, V20, V21, V22, V23, V24)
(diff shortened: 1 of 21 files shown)


bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
```diff
@@ -368,10 +370,10 @@ sealed trait BitcoindVersion
 object BitcoindVersion extends StringFactory[BitcoindVersion] with Logging {
 
   /** The newest version of `bitcoind` we support */
-  val newest: BitcoindVersion = V23
+  val newest: BitcoindVersion = V24
```
Christewart: I'm seeing some test failures in `nodeTest`. I'm _guessing_ that its related to bumping this version. 

If we wanted to merge this PR, i would suggest reverting this to `V23` and I can investigate why our `nodeTest` does not seem to be playing nicely with `V24`. 

I believe that will unblock you so you can use v24 while I investigate the bug in our `node` impl. If we take this path, we will need to make sure we file an issue to bump this version and fix the bug in `node`. 

Thoughts?

testkit/src/main/scala/org/bitcoins/testkit/rpc/BitcoindRpcTestUtil.scala
```diff
@@ -110,6 +111,7 @@ trait BitcoindRpcTestUtil extends Logging {
                   |port=${uri.getPort}
                   |debug=$isDebug
                   |walletbroadcast=1
+                  |mempoolfullrbf=1
```
Christewart, line 114: 👀 👀 😂

bitcoind-rpc/src/main/scala/org/bitcoins/rpc/client/common/BitcoindRpcClient.scala
Christewart: bug is most likely related to this: https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-24.0.md#p2p-and-network-changes

benthecarman: I tried debugging for a little but can't figure it out. Went ahead and reverted to v23, if you wanna take a stab at it that'd be great

Christewart (approved):

PR #4913, opened 2022-12-07
2022 12 07 Preliminary support for wallet descriptors

Christewart:
Add preliminary support for descriptors. 

https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki

This is not intended to be a complete implementation of BIP 380. This is only intended to implements parts needed to parse BitMEX's descriptors they use in proof of reserves. An example can be found in the test case.

https://blog.bitmex.com/bitmex-provides-snapshot-update-to-proof-of-reserves-proof-of-liabilities/

I filed an issue to add full descriptor support #4914


Christewart:
Superseded by #5525

PR #4920, opened 2022-12-12
Fix synced flag but in `DataMessageHandler` bug

Christewart:
This was introduced in #1969

https://github.com/bitcoin-s/bitcoin-s/pull/1969/files#diff-61413f0a707d1e3e49431ad34fcb1af04b4fa36148e75c7abc692bebc07822b7R72

It seems we renamed the local scoped variable from `syncing` -> `synced`. We didn't replace the return value.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
+++ b/node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
@@ -107,7 +107,7 @@ case class DataMessageHandler(
                   peerMsgSender,
                   startHeightOpt).map { synced =>
                   if (!synced) logger.info("We are synced")
-                  syncing
+                  synced
                 }
               } yield (syncing, startHeightOpt)
             }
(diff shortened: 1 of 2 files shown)


node/src/main/scala/org/bitcoins/node/networking/peer/DataMessageHandler.scala
```diff
@@ -107,7 +107,7 @@ case class DataMessageHandler(
                   peerMsgSender,
                   startHeightOpt).map { synced =>
                   if (!synced) logger.info("We are synced")
-                  syncing
+                  synced
```
benthecarman: should this be `!synced`? `!synced` would imply to me that we are `syncing`

Christewart: The correct variable name should be `syncing` so i will rename it. It's unfortunate that we also have a `syncing` value passed as a parameter to `DataMessageHandler` too which can cause confusion. Hopefully I can get rid of that in a future PR.

This is where the `syncing` value comes from.

https://github.com/bitcoin-s/bitcoin-s/blob/c4d358061a20312cd30b252c853b4536e51a20d3/node/src/main/scala/org/bitcoins/node/networking/peer/PeerMessageSender.scala#L213

benthecarman (approved):

PR #4522, opened 2022-07-19
Fix block fetch issue #4513

solo1g:
The logic for fetching blocks that I added was wrong. Closes #4513

PR #4924, opened 2022-12-14
Download blocks from random peer if we aren't in IBD

Christewart:
This fixes this comment on #4522 https://github.com/bitcoin-s/bitcoin-s/pull/4522/files#r924895166

In #4627 we added a flag to track IBD state. We use this flag to select what peer to request blocks from. If we are in IBD, we only request from the peer we are syncing with. If we aren't in IBD, we select a random peer.

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/Node.scala
+++ b/node/src/main/scala/org/bitcoins/node/Node.scala
@@ -183,6 +183,21 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
     if (blockHashes.isEmpty) {
       Future.unit
     } else {
+      for {
+        chainApi <- chainApiFromDb()
+        isIBD <- chainApi.isIBD()
+        _ <- downloadBlocksBasedOnIBD(isIBD, blockHashes)
+      } yield ()
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


node/src/main/scala/org/bitcoins/node/Node.scala
```diff
@@ -183,6 +183,21 @@ trait Node extends NodeApi with ChainQueryApi with P2PLogger {
     if (blockHashes.isEmpty) {
       Future.unit
     } else {
+      for {
+        chainApi <- chainApiFromDb()
+        isIBD <- chainApi.isIBD()
```
benthecarman, line 188: might be good to have a version of `downloadBlocks` that has the `isIBD` so we don't always need to call `chainApiFromDb` as this does some database reads I believe

Christewart: I don't think the performance hit here is significant. Network latency of transmitting a block will be the limiting factor here.

rorp: OTOH the fewer disk I/O operations the longer the storage device lives.

Issue #1098, opened 2020-02-06
Generating many blocks on regtest bitcoind node peered with neutrino corrupts chaindb


Christewart:
If you peer your bitcoind node with bitcoin-s and then run `bitcoin-cli generatetoaddress 100 $(bitcoin-cli getnewaddress)` you will end up with a corrupted chain state, I believe we are receiving blocks out of order. 

If you generate blocks 1 by 1 with `bitcoin-cli generatetoaddress 1 $(bitcoin-cli getnewaddress)` you will stay in sync with the bitcoind node.

Also if you turn off your neutrino node, and then do `bitcoin-cli generatetoaddress 100 $(bitcoin-cli getnewaddress)` and then restart your node, you will be able to sync fine. 

So the basic issue here is when we are peered with `bitcoind` in regtest and are generating a large amount of blocks. We must have an async bug in the node/chain processing logic. 

I do not have a stack trace at the moment. 

This should be easy to replicate in a unit test, specifically by modifying this unit test: 

https://github.com/bitcoin-s/bitcoin-s/blob/master/node-test/src/test/scala/org/bitcoins/node/NeutrinoNodeTest.scala#L85

The difference between this test case and the one linked above is that the test case above generates the blocks BEFORE starting the neutrino node, this bug occurs while the neutrino node is peered with bitcoind, and then bitcoind generates a burst of blocks. 

This was originally reported by @nkohen

rorp:
This issues is not specific for neutrino mode. 

For some reason, when one generates a large number of blocks (more than 10) using `generatetoaddress` CLI command, `bitcoind` does not send `headers` messages, but instead it sends `inv` messages for the generated blocks. However, the number of `inv`messages is doesn't match the number of the generated blocks. That's why chain database becomes "corrupted". In fact, not corrupted, but rather incomplete, because the `bitcoin-s` node does not get notified about the generated blocks.

benthecarman:
https://github.com/bitcoin/bitcoin/issues/17451#issuecomment-575898689

Christewart:
So it seems we need to start IBD again when this happens? Bitcoin core gets overloaded with blocks, so it stops announcing them. If we cannot connect the blocks we need to send a `getheaders` message to our peer?

PR #5009, opened 2023-03-07
2023 03 03 rm `DataMessageHandler` filter params

Christewart:
Simplify `DataMessageHandler` by removing the filter header / filter parameters. These can be access via `ChainApi`. 

This is a follow up on #4925


benthecarman (commented):
We originally did this to speed up syncing because getting from the database the new heights was costly.

Christewart:
some test failures being caused seem to be related to #1098 .

Christewart:
> We originally did this to speed up syncing because getting from the database the new heights was costly.

In #4109 we sync since wallet creation time which makes syncing compact filters less of a performance concern.

PR #5065, opened 2023-05-05
Add `MisbehavingPeer` state

Christewart:
Adds `MisbehavingPeer` state. This state is used when a peer is misbehaving. 

This purpose for this is to give us a return value to indicate the peer is misbehaving, with the goal of decoupling `PeerManager` and `DataMessageHandler`. Now that we have a state to represent a misbehaving peer, we can move logic for syncing with a new peer from `DataMessageHandler` to `PeerManager`. 

This is the logic that is moved to `PeerManager` now.

https://github.com/bitcoin-s/bitcoin-s/compare/master...Christewart:2023-05-05-misbehaving-peer-state?expand=1#diff-61413f0a707d1e3e49431ad34fcb1af04b4fa36148e75c7abc692bebc07822b7L474

This PR works towards a higher level goal of using akka streams to create a pipeline for p2p message processing. Related to #5062

### DIFF
--- a/node/src/main/scala/org/bitcoins/node/PeerManager.scala
+++ b/node/src/main/scala/org/bitcoins/node/PeerManager.scala
@@ -440,9 +440,6 @@ case class PeerManager(
     Future.unit
   }
 
-  def syncFromNewPeer(): Future[DataMessageHandler] =
-    node.syncFromNewPeer().map(_ => getDataMessageHandler)
-
   private def onHeaderRequestTimeout(
       peer: Peer,
       state: DataMessageHandlerState): Future[DataMessageHandler] = {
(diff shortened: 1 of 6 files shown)


node/src/main/scala/org/bitcoins/node/PeerManager.scala
```diff
@@ -485,9 +486,20 @@ case class PeerManager(
...
         getDataMessageHandler
           .handleDataPayload(payload, peerMsgSender, peer)
-          .map { newDmh =>
-            updateDataMessageHandler(newDmh)
-            msg
+          .flatMap { newDmh =>
+            newDmh.state match {
+              case m: MisbehavingPeer =>
+                updateDataMessageHandler(newDmh)
+                //disconnect the misbehaving peer
```
Christewart, line 490: Here is where we are now syncing with a new peer based on the return value from `DataMessageHandler.handleDataPayload()`

PR #5217, opened 2023-08-30
2023 08 30 refactor `ConsoleCli`

Christewart:
- Cleans up implementation inside of `ConsoleCli`
- Adds support for logging in `bitcoin-s-cli`


app/cli/src/main/scala/org/bitcoins/cli/ConsoleCli.scala
```diff
@@ -2303,8 +2408,9 @@ object ConsoleCli {
         RequestParam("getdlcwalletaccounting")
       case GetVersion =>
         // skip sending to server and just return version number of cli
-        return Success(EnvUtil.getVersion)
-
+        //comeback and implement version serverside
```
Christewart: need to come back and implement this server side

Issue #5216, opened 2023-08-29
`OutOfMemory` errors happening during IBD with wallet


Christewart:
Observed on 8ab9718a

The error occurs when we are matching compact filters with utxos in our wallet and then requesting blocks on the p2p network. It seems that our upstream producer of blocks (the p2p network, our compact filter matching logic) is faster than the downstream consumer of bitcoin blocks (`WalletApi.processBlock()`). 

This might be related to #5212 and #5206 . #5206 specifically increased upstream performance by reducing the costliness of sql queries on `chaindb` 

The command I am running with sets the heap size to `-J-Xmx2g` 

```
./bin/bitcoin-s-server --datadir /tmp/test-sync -J-Xmx2g 
```

You can see in this excerpt of the logs that requesting filters from the p2p network is at block height `637,723`, whereas we are processing block `622,966` which means we have queued a bunch of blocks to be processed.

```
2023-08-25 21:11:11,644UTC INFO [DLCWallet$DLCWalletImpl] Processing block=00000000000000000001d3ae7ccb6d0ff918d94a378b29c41e5926c0df2a8298 heightOpt=Some(622966)
2023-08-25 21:11:11,645UTC INFO [PeerManager$] Requesting compact filters from FilterSyncMarker(startHeight=637723, stopBlockHash=0000000000000000000d7cd6c226e543120bd0c6d6259953218a5cec60a01535) with peer=Peer(neutrino.suredbits.com:8333)
```

Full Logs: 

```
2023-08-25 21:10:54,585UTC INFO [DLCWallet$DLCWalletImpl] Processing block=000000000000000000007bb0d973c7f1bf4dd2dab3feb95852f465bf8491afb6 heightOpt=Some(622399)
2023-08-25 21:10:59,153UTC INFO [DLCWallet$DLCWalletImpl] Finished processing of block=000000000000000000007bb0d973c7f1bf4dd2dab3feb95852f465bf8491afb6. It took 4571ms
2023-08-25 21:10:59,161UTC INFO [DLCWallet$DLCWalletImpl] Processing block=0000000000000000000fc2062d710c55903b775adfea7623c57fe554879477f9 heightOpt=Some(622755)
2023-08-25 21:11:04,475UTC INFO [DLCWallet$DLCWalletImpl] Finished processing of block=0000000000000000000fc2062d710c55903b775adfea7623c57fe554879477f9. It took 5321ms
[...]

Issue #5212, opened 2023-08-28
Callback system can drop data in the case where upstream producers are faster than downstream consumers


Christewart:
Introduced in #3946

We can drop elements from our callback system

This is because we specify `OverfloflowStrategy.dropHead`

https://github.com/bitcoin-s/bitcoin-s/blob/08e780a8844f860c512bb6a0ba82631b709c102a/app/server/src/main/scala/org/bitcoins/server/BitcoinSServerMain.scala#L674

Christewart:
This issue doesn't really make sense. This logic controls the websocket queue. We don't want to infinitely queue things waiting for a websocket connection. It makes sense to use `OverflowStrategy.dropHead` as we want to send the _latest_ data to a new connection to the websocket rather than the oldest data.

The intended behavior fix was addressed in #5229

PR #5222, opened 2023-09-05
2023 09 04 wssource backpressure

Christewart:
Related to #5212 but doesn't necessarily fix it. This specifies a bufferSize for `BroadcastHub.sink()` that is the same as the queue buffer size. 

This also changes the OverflowStrategy from `dropNew` -> `backpressure`. This could lead to `OutOfMemory` errors as talked about in #5216 , but it doesn't seem to make things any worse than they already are. We don't want to drop data for our internal callback system as our wallet depends on this to make sure we are synced with the network correctly.


Christewart:
See: https://github.com/bitcoin-s/bitcoin-s/issues/5212#issuecomment-1740953273

PR #5427, opened 2024-02-27
Add delay to wait for zmq to get setup in chainTest

Christewart:
Attempt to fix this CI failure:

https://github.com/bitcoin-s/bitcoin-s/actions/runs/8066023192/job/22033148799?pr=5425#step:5:1129


Christewart:
Reverts #5410 as it seems there is an internal bug causing us to fail on CI now.

Issue #5672, opened 2024-09-17
Scala 3 release


kamilkloch:
Hello,

it would be fantastic if bitcoin-s contained a cross-release for Scala 2.13 / 3 :)

Christewart:
Aren't 2.13 libraries consumable from 3?

kamilkloch:
Thank you for the response, I will give it a try with 
```scala
("org.bitcoin-s" %% "bitcoin-s-core" % bitcoinsVersion).cross(CrossVersion.for3Use2_13),
```
It might work, assuming bitcoin-s does not use macros, compiler plugins, or reflection.

kamilkloch:
Update: it ends up in a conflict - http4s libs depend on `scodec-bits` for Scala 3. Probably solvable with a bunch of manual additions / exclusions, but not that straightforward. Also, forced lib versions  / exclusions tend to shoot in the back...

PR #5713, opened 2024-10-15
2024 05 11 scala3 crossbuild

Christewart:
Get codebase compiling with scala3

Supersedes #3497 and related to #5672


Christewart:
Seems like this is stuck on a bug in scala3 and jdeps: https://stackoverflow.com/questions/79636833/fix-missing-transnitive-dependencies-with-jlink-and-scala3-scala-quoted-scal/79646280

PR #5961, opened 2025-03-15
2025 03 15 Remove `data.length == 32` requirement from secp256k1jni

### LINKED ISSUE #5436: Bug: unnecessary message length check in BIP-340 Schnorr signature verification
BIP-340 Schnorr message can be of arbitrary length, but you enforce message to be 32 bytes which is wrong:

[BIP-340 Spec](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#messages-of-arbitrary-size)

> The signature scheme specified in this BIP accepts byte strings of arbitrary size as input messages.

Your implementation:
https://github.com/bitcoin-s/bitcoin-s/blob/87003eb70303a0c58dd82d5b1afd7ba06860f97d/secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java#L682

Suggested fix: just remove the data size check, should work.

Christewart:
fixes #5436 

@nau I could have swore at some point that BIP340 stated that the data length must be 32 bytes, but either I imagined that or it was removed from the spec. I'm too lazy to do the git history to find out if it existed at any point.

Also I thought I tried this before and ran into testing issues, we will see if CI finds any failed tests.

Sorry for the delay on this, my bad.

### DIFF
--- a/secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
+++ b/secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
@@ -17,13 +17,13 @@
 
 package org.bitcoin;
 
+import java.math.BigInteger;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-
-import java.math.BigInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 import static org.bitcoin.NativeSecp256k1Util.*;
 
 /**
@@ -571,7 +571,7 @@ public class NativeSecp256k1 {
      * @param secKey key to sign with
      */
     public static byte[] schnorrSign(byte[] data, byte[] secKey, byte[] auxRand) throws AssertFailException {
-        checkArgument(data.length == 32 && secKey.length == 32 && auxRand.length == 32);
+        checkArgument(secKey.length == 32 && auxRand.length == 32);
 
         ByteBuffer byteBuff = nativeECDSABuffer.get();
         if (byteBuff == null || byteBuff.capacity() < 32 + 32 + 32) {
@@ -608,7 +608,7 @@ public class NativeSecp256k1 {
      * @param nonce the nonce (k value) used in signing
      */
     public static byte[] schnorrSignWithNonce(byte[] data, byte[] secKey, byte[] nonce) throws AssertFailException {
-        checkArgument(data.length == 32 && secKey.length == 32 && nonce.length == 32);
+        checkArgument(secKey.length == 32 && nonce.length == 32);
 
         ByteBuffer byteBuff = nativeECDSABuffer.get();
         if (byteBuff == null || byteBuff.capacity() < 32 + 32 + 32) {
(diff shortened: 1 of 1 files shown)


secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -52,7 +52,7 @@ public class NativeSecp256k1 {
      * @param pub The public key which did the signing
      */
     public static boolean verify(byte[] data, byte[] signature, byte[] pub) {
-        checkArgument(data.length == 32 && signature.length <= 520 && pub.length <= 520);
+        checkArgument(signature.length <= 520 && pub.length <= 520);
```
nau: I believe the ECDSA expects 32 byte message. It's only the Schnorr signing/verify doesn't restrict the message length: https://github.com/bitcoin-core/secp256k1/blob/70f149b9a1bf4ed3266f97774d0ae9577534bf40/src/secp256k1.c#L566

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -82,7 +82,7 @@ public static boolean verify(byte[] data, byte[] signature, byte[] pub) {
      * @return sig byte array of signature
      */
     public static byte[] sign(byte[] data, byte[] seckey) throws AssertFailException{
-        checkArgument(data.length == 32 && seckey.length <= 32);
```
nau: same as above

secp256k1jni/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -128,7 +128,7 @@ public static byte[] sign(byte[] data, byte[] seckey) throws AssertFailException
      * @return sig byte array of signature
      */
     public static byte[] signWithEntropy(byte[] data, byte[] seckey, byte[] entropy) throws AssertFailException{
-        checkArgument(data.length == 32 && seckey.length == 32 && entropy.length == 32);
```
nau: same as above

nau:
No worries Chris. The BIP340 literally says it used to be the case: 

> Earlier revisions of this BIP required messages to be exactly 32 bytes. This restriction puts a burden on callers who typically need to perform pre-hashing of the actual input message by feeding it through SHA256 (or another collision-resistant cryptographic hash function) to create a 32-byte digest which can be passed to signing or verification (as for example done in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki).)

We use this library in our [Scalus](https://github.com/nau/scalus) project that implements Cardano smart contract builtins for verifying Bitcoin Secp256k1 ECDSA and Schnorr signatures. Thanks for your work!

Christewart:
> No worries Chris. The BIP340 literally says it used to be the case:
> 
> > Earlier revisions of this BIP required messages to be exactly 32 bytes. This restriction puts a burden on callers who typically need to perform pre-hashing of the actual input message by feeding it through SHA256 (or another collision-resistant cryptographic hash function) to create a 32-byte digest which can be passed to signing or verification (as for example done in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki).)
> 
> We use this library in our [Scalus](https://github.com/nau/scalus) project that implements Cardano smart contract builtins for verifying Bitcoin Secp256k1 ECDSA and Schnorr signatures. Thanks for your work!

Glad you are finding it useful!

PR #5974, opened 2025-04-02
chore: remove redundant words

luozexuan:
remove redundant words


luozexuan:
Thanks for your review!

PR #6065, opened 2025-09-09
core: Make `ReservedOperation`s dynamic based on allocated opcodes

Christewart:
This makes adding new opcodes simpler, you no longer have to remember to remove the byte that is the new opcode in 

https://github.com/bitcoin-s/bitcoin-s/blob/654d4086b993c35d4cd343a2009daed40c27d7b0/core/src/main/scala/org/bitcoins/core/script/reserved/ReservedOperations.scala#L93

Tested on #6064 to make sure it works as expected


Christewart:
This PR doesn't address the OP_SUCCESS semantics with newly assigned opcodes. We determine if an opcode is OP_SUCCESSx [here](https://github.com/bitcoin-s/bitcoin-s/blob/c999084d970a0d0ad50ced33a9d318459c43b064/core/src/main/scala/org/bitcoins/core/script/interpreter/ScriptInterpreter.scala#L609) - it isn't clear to me how to tie this nicely into existing allocated opcodes pre-tapscript. 

When a new opcode is added, we will have to remove the assigned byte from `opSuccessBytes`

PR #6129, opened 2025-10-29
Fix Taproot fundingOutputs ordering to comply with BIP-341

liam-safulet:
The previous implementation used outputMap.values.toVector which does not guarantee ordering. BIP-341 requires that outputs be ordered according to the transaction inputs order for correct sighash calculation.

This bug only affects Taproot transactions with multiple inputs that have different amounts, causing signature verification to fail.


liam-safulet:
Can i ask when will  1.9.12 release?

Christewart:
hi @liam-safulet . I'm not sure yet. Probably towards the end of the year. In the mean time, you should be able to use SNAPSHOT releases if you need any behavior in master since then. You can publish the library locally too with `sbt publishLocal`

PR #6143, opened 2025-11-17
2025 11 06 processtransaction opt

Christewart:
Attempt to optimize wallet block processing by using `DBIOAction` exclusively (no `Future`) in `processBlockCachedUtxos`

This makes things much slower as of now, not sure why. Perhaps an index needs to be added to the wallet to make `spendingInfoDAO.findTxsAction()` faster? 

It seems that commit 710a264336b19a63243c0b361085108fe39a61c8 makes things exponentially slower

and commit a622aba2774e4434a4821c64d2091fd9f16a4709 makes things maybe 2x slower?


Christewart:
The performance issues this PR intended to help were fixed by #6242

PR #6155, opened 2025-12-04
Add BIP47 Reusable Payment Codes Support

MightyMercurian:
## Summary
Implements BIP47 (Reusable Payment Codes) for bitcoin-s core, enabling privacy-preserving payment address generation between two parties.

### PaymentCode
- 80-byte payload encoding/decoding with Base58Check (0x47 version prefix)
- XOR masking for notification transaction blinding
- V1/V2 version support

### SecretPoint
- ECDH shared secret computation
- Native secp256k1 with pure Scala fallback

### PaymentAddress
- Ephemeral address derivation for send/receive operations
- SHA256-based key tweaking from ECDH secrets
- SegWit address generation

### BIP47Account
- HD account at derivation path m/47'/coin_type'/account'
- Factory methods: fromSeed, fromExtendedKey, fromPaymentCode
- WatchOnlyBIP47Account for watch-only wallet support

## Test plan
- [x] Payment code creation from seed
- [x] Payment code serialization symmetry (bytes → PaymentCode → bytes)
- [x] ECDH shared secret symmetry (Alice·Bob == Bob·Alice)
- [x] Masking/unmasking roundtrip
- [x] Mutual address derivation (Alice sends to Bob == Bob receives from Alice)
- [x] SegWit address derivation
- [x] Key derivation at correct indices
- [x] Watch-only account creation from payment code

### Wallets
- Stack Wallet
- Sparrow Wallet
- Blue Wallet
- Ashigaru Wallet
- Samourai Wallet

### Mining Pools
- Lincoin (for payouts)

### BIP47 Specification
- [BIP47 - Bitcoin Wiki](https://en.bitcoin.it/wiki/BIP_0047)

### Technical Deep Dives
- [BIP47, The Ugly Duckling](https://21ideas.org/en/bip47-the-ugly-duckling/)
- [How BIP47 Works - Medium](https://medium.com/@ottosch/how-bip47-works-ee641cc14bf3)
- [How BIP47 Enriches Bitcoin UX - Medium](https://medium.com/billion-crypto-stories/how-bip47-reusable-payment-codes-enrich-bitcoin-and-overall-cryptocurrency-user-experience-6f929c87a61b)
- [BIP47, The Ugly Duckling](https://21ideas.org/en/bip47-the-ugly-duckling/)

### PayNym Guides
- [PayNym Directory](https://paynym.is)
- [Bitcoin PayNyms Explained - Bitcoin Manual](https://thebitcoinmanual.com/articles/bitcoin-paynyms/)
- [PayNym Guide - Bitcoiner.guide](https://bitcoiner.guide/paynym/)
- [Guide to PayNyms with BIP47 - D-Central](https://d-central.tech/guide-to-paynyms-enhancing-bitcoin-privacy-with-bip-47/)

### Auth47 (PayNym Authentication)
- [Auth47 PayNym Accounts - Bitcoin Manual](https://thebitcoinmanual.com/articles/auth47-paynym-accounts/)
- [Auth47 PayNyms - RoninDojo Blog](https://blog.ronindojo.io/auth47-paynyms/)

### Video Tutorials
- [What are PayNyms - Andreas Antonopoulos](https://www.youtube.com/watch?v=Qmc1ubPVhnQ)
- [PayNyms: Enhanced Bitcoin Privacy With Public IDs](https://www.youtube.com/watch?v=w6o_3sUcHHY)
- [PayNyms in Sparrow Wallet - Privacy Preserving Public Bitcoin IDs](https://www.youtube.com/watch?v=o-iDeLZ4BiE)
- [Enhancing Privacy and Security: Reusable Payment Codes for Bitcoin Transactions](https://www.youtube.com/watch?v=PTGp4HCrFfc)

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
+++ b/core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
@@ -1,0 +1,204 @@
+package org.bitcoins.core.bip47
+
+import org.bitcoins.core.config.{MainNet, NetworkParameters}
+import org.bitcoins.core.crypto._
+import org.bitcoins.core.crypto.ExtKeyVersion.LegacyMainNetPriv
+import org.bitcoins.core.crypto.ExtKeyVersion.LegacyTestNet3Priv
+import org.bitcoins.core.crypto.ExtKeyPubVersion.LegacyMainNetPub
+import org.bitcoins.core.crypto.ExtKeyPubVersion.LegacyTestNet3Pub
+import org.bitcoins.core.number.{UInt32, UInt8}
+import org.bitcoins.core.protocol.{Bech32Address, P2PKHAddress}
+import org.bitcoins.core.protocol.script.P2WPKHWitnessSPKV0
+import org.bitcoins.crypto.{CryptoUtil, ECPrivateKey, ECPublicKey}
+import scodec.bits.ByteVector
+
+import scala.util.Try
+
+/** Represents a BIP47 account at derivation path m/47'/coin_type'/account'.
+  * @see
+  *   [[https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki BIP47]]
+  */
+sealed abstract class BIP47Account {
+
+  def extPrivKey: ExtPrivateKey
+
+  def network: NetworkParameters
+
+  def accountIndex: Int
+
+  def paymentCode: PaymentCode = {
+    val accountKey = extPrivKey
+    PaymentCode(accountKey.key.publicKey, accountKey.chainCode)
+  }
+
+  def notificationKey: ExtPrivateKey = deriveKey(0)
+
+  def notificationAddress: P2PKHAddress = {
+    val pubKey = notificationKey.key.publicKey
+    val hash = CryptoUtil.sha256Hash160(pubKey.bytes)
+    P2PKHAddress(hash, network)
+  }
+
+  def notificationSegwitAddress: Bech32Address = {
+    val pubKey = notificationKey.key.publicKey
+    val spk = P2WPKHWitnessSPKV0(pubKey)
+    Bech32Address(spk, network)
+  }
+
+  def deriveKey(index: Int): ExtPrivateKey = {
+    extPrivKey.deriveChildPrivKey(UInt32(index))
+  }
+
+  def privateKeyAt(index: Int): ECPrivateKey = deriveKey(index).key
+
+  def publicKeyAt(index: Int): ECPublicKey = privateKeyAt(index).publicKey
+
+  def addressAt(index: Int): Bech32Address = {
+    val pubKey = publicKeyAt(index)
+    val spk = P2WPKHWitnessSPKV0(pubKey)
+    Bech32Address(spk, network)
+  }
+
+  /** Creates a PaymentAddress for sending to a remote party. */
+  def paymentAddressForSending(
+      remotePaymentCode: PaymentCode,
+      index: Int): PaymentAddress = {
+    PaymentAddress.forSending(
+      localPrivKey = notificationKey.key,
+      remotePaymentCode = remotePaymentCode,
+      index = index,
+      network = network
+    )
+  }
+
+  /** Creates a PaymentAddress for receiving from a remote party. */
+  def paymentAddressForReceiving(
+      remotePaymentCode: PaymentCode,
+      index: Int): PaymentAddress = {
+    PaymentAddress.forReceiving(
+      localPaymentCode = paymentCode,
+      localPrivKeyAtIndex = privateKeyAt(index),
+      remotePaymentCode = remotePaymentCode,
+      index = index,
+      network = network
+    )
+  }
+}
+
+object BIP47Account {
+  val Purpose: Int = 47
+  val HardenedOffset: Long = 0x80000000L
+
+  private case class BIP47AccountImpl(
+      extPrivKey: ExtPrivateKey,
+      network: NetworkParameters,
+      accountIndex: Int)
+      extends BIP47Account
+
+  def apply(
+      extPrivKey: ExtPrivateKey,
+      network: NetworkParameters,
+      accountIndex: Int): BIP47Account = {
+    BIP47AccountImpl(extPrivKey, network, accountIndex)
+  }
+
+  /** Creates a BIP47Account from a seed at the standard derivation path. */
+  def fromSeed(
+      seed: ByteVector,
+      network: NetworkParameters,
+      accountIndex: Int = 0): BIP47Account = {
+    val version = network match {
+      case _: MainNet => LegacyMainNetPriv
+      case _          => LegacyTestNet3Priv
+    }
+
+    val master = ExtPrivateKey(version, Some(seed))
+
+    val purposeKey =
+      master.deriveChildPrivKey(UInt32(Purpose + HardenedOffset))
+    val coinTypeKey =
+      purposeKey.deriveChildPrivKey(UInt32(coinType(network) + HardenedOffset))
+    val accountKey =
+      coinTypeKey.deriveChildPrivKey(UInt32(accountIndex + HardenedOffset))
+
+    BIP47AccountImpl(accountKey, network, accountIndex)
+  }
+
+  def fromExtendedKey(
+      accountKey: ExtPrivateKey,
+      network: NetworkParameters,
+      accountIndex: Int): BIP47Account = {
+    BIP47AccountImpl(accountKey, network, accountIndex)
+  }
+
+  def fromPaymentCode(
+      paymentCode: PaymentCode,
+      network: NetworkParameters): WatchOnlyBIP47Account = {
+    WatchOnlyBIP47Account(paymentCode, network)
+  }
+
+  private def coinType(network: NetworkParameters): Long = network match {
+    case _: MainNet => 0L
+    case _          => 1L
+  }
+}
+
+/** Watch-only BIP47 account created from a payment code (no private keys). */
+sealed abstract class WatchOnlyBIP47Account {
+
+  def paymentCode: PaymentCode
+
+  def network: NetworkParameters
+
+  def publicKeyAt(index: Int): ECPublicKey = {
+    val masterKey = ExtPublicKey(
+      extKeyVersion(network),
+      UInt8.zero,
+      ByteVector.fill(4)(0),
+      UInt32.zero,
+      paymentCode.chainCode,
+      paymentCode.pubKey
+    )
+    masterKey.deriveChildPubKey(index).get.key
+  }
+
+  def addressAt(index: Int): Bech32Address = {
+    val pubKey = publicKeyAt(index)
+    val spk = P2WPKHWitnessSPKV0(pubKey)
+    Bech32Address(spk, network)
+  }
+
+  def notificationAddress: P2PKHAddress = {
+    val pubKey = publicKeyAt(0)
+    val hash = CryptoUtil.sha256Hash160(pubKey.bytes)
+    P2PKHAddress(hash, network)
+  }
+
+  private def extKeyVersion(network: NetworkParameters): ExtKeyPubVersion =
+    network match {
+      case _: MainNet => LegacyMainNetPub
+      case _          => LegacyTestNet3Pub
+    }
+}
+
+object WatchOnlyBIP47Account {
+
+  private case class WatchOnlyBIP47AccountImpl(
+      paymentCode: PaymentCode,
+      network: NetworkParameters)
+      extends WatchOnlyBIP47Account
+
+  def apply(
+      paymentCode: PaymentCode,
+      network: NetworkParameters): WatchOnlyBIP47Account = {
+    WatchOnlyBIP47AccountImpl(paymentCode, network)
+  }
+
+  def fromBase58(
+      base58: String,
+      network: NetworkParameters): Try[WatchOnlyBIP47Account] = {
+    PaymentCode.fromBase58(base58).map { pc =>
+      WatchOnlyBIP47AccountImpl(pc, network)
+    }
+  }
+}
--- a/core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
+++ b/core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
@@ -1,15 +1,12 @@
 package org.bitcoins.core.bip47
 
-import org.bitcoins.core.config.{MainNet, NetworkParameters}
+import org.bitcoins.core.config._
 import org.bitcoins.core.crypto._
-import org.bitcoins.core.crypto.ExtKeyVersion.LegacyMainNetPriv
-import org.bitcoins.core.crypto.ExtKeyVersion.LegacyTestNet3Priv
-import org.bitcoins.core.crypto.ExtKeyPubVersion.LegacyMainNetPub
-import org.bitcoins.core.crypto.ExtKeyPubVersion.LegacyTestNet3Pub
+import org.bitcoins.core.hd.HDCoinType
 import org.bitcoins.core.number.{UInt32, UInt8}
 import org.bitcoins.core.protocol.{Bech32Address, P2PKHAddress}
 import org.bitcoins.core.protocol.script.P2WPKHWitnessSPKV0
-import org.bitcoins.crypto.{CryptoUtil, ECPrivateKey, ECPublicKey}
+import org.bitcoins.crypto.{ECPrivateKey, ECPublicKey}
 import scodec.bits.ByteVector
 
 import scala.util.Try
@@ -34,9 +31,7 @@ sealed abstract class BIP47Account {
   def notificationKey: ExtPrivateKey = deriveKey(0)
 
   def notificationAddress: P2PKHAddress = {
-    val pubKey = notificationKey.key.publicKey
-    val hash = CryptoUtil.sha256Hash160(pubKey.bytes)
-    P2PKHAddress(hash, network)
+    P2PKHAddress(notificationKey.key.publicKey, network)
   }
 
   def notificationSegwitAddress: Bech32Address = {
@@ -87,7 +82,6 @@ sealed abstract class BIP47Account {
 
 object BIP47Account {
   val Purpose: Int = 47
-  val HardenedOffset: Long = 0x80000000L
 
   private case class BIP47AccountImpl(
       extPrivKey: ExtPrivateKey,
@@ -107,19 +101,18 @@ object BIP47Account {
       seed: ByteVector,
       network: NetworkParameters,
       accountIndex: Int = 0): BIP47Account = {
-    val version = network match {
-      case _: MainNet => LegacyMainNetPriv
-      case _          => LegacyTestNet3Priv
-    }
-
+    val version = ExtKeyPrivVersion.legacyFromNetwork(network)
     val master = ExtPrivateKey(version, Some(seed))
+    val hardenedIdx = ExtKey.hardenedIdx
+
+    val coinType = HDCoinType.fromNetwork(network).toInt
 
     val purposeKey =
-      master.deriveChildPrivKey(UInt32(Purpose + HardenedOffset))
+      master.deriveChildPrivKey(hardenedIdx + UInt32(Purpose))
     val coinTypeKey =
-      purposeKey.deriveChildPrivKey(UInt32(coinType(network) + HardenedOffset))
+      purposeKey.deriveChildPrivKey(hardenedIdx + UInt32(coinType))
     val accountKey =
-      coinTypeKey.deriveChildPrivKey(UInt32(accountIndex + HardenedOffset))
+      coinTypeKey.deriveChildPrivKey(hardenedIdx + UInt32(accountIndex))
 
     BIP47AccountImpl(accountKey, network, accountIndex)
   }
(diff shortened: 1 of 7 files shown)


MightyMercurian:
@Christewart would love to get your thoughts on this

Christewart:
Hi @MightyMercurian . I'll try to get review done on this within a week. I'm not familiar with BIP47 so i'll take a deeper look at that as well :+1:

MightyMercurian:
@Christewart Awesome thank you! No rush at all. BIP 47 gets a little technical under the hood so take your time. Added a few resources that might be helpful. Cheers.

MightyMercurian:
Will keep an eye on the checks too and fix the failures

Christewart:
Hi @MightyMercurian 

Still getting up to speed on BIP47 before reviewing, but thought I would ask if you've followed any of the Silent Payments development? https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki

Admittedly i'm not super familiar with Silent Payments either but it seems like Silent Payments is a successor (without the adoption ? 😅 ) to BIP47.

Of course there is room for both primitives in the core library, but thought I'd see what your thoughts were (if any) on Silent Payments.

MightyMercurian:
Hey @Christewart

Great question. I've been diving deep into both protocols while working on this.

Silent Payments (BIP352) and BIP47 solve similar problems but with different tradeoffs. You're right that SP is newer and often positioned as a "successor," but after researching I'd actually argue BIP47 is the more practical choice today for a few reasons:

Maturity & Wallet Support
- BIP47 has been around since 2015 with polished implementations in Stack Wallet, Sparrow, Blue Wallet, Ashigaru Wallet, Samourai Wallet, Lincoin Mining Pool, etc.
- Silent Payments is much newer and the few implementations that exist are buggy and incomplete

Light Client Limitations
- [Per Electrum developers](https://github.com/spesmilo/electrum/pull/9900#issuecomment-3302980342): "there does not seem to be a realistic way to implement both sending and receiving silent payments in a light client"
- SP is incompatible with standard Electrum server protocol - requires entirely new infrastructure
- Even mature wallets like [Sparrow](https://github.com/sparrowwallet/sparrow/issues/1434) and Electrum only implemented send-only support, deliberately avoiding receiving complexity

Operational Complexity
- SP has RBF fund loss risks - if same seed is on two devices and you bump fee with changed inputs on device B, funds are lost
- SP breaks standard invoice tracking (scriptpubkeys aren't known until tx creation)
- SP requires scanning every Bitcoin transaction with ECDH calculations to detect payments
- Wallet recovery requires scanning entire blockchain history

BIP47's Notification Transaction
The notification tx is often cited as BIP47's downside, but it actually provides benefits:
- Deterministic recovery point (no full chain scan needed)
- Anti-sybil cost barrier
- Helps identify dust attacks
- One-time cost (~$1-5), all subsequent payments have zero overhead

That said, I'd love to see bitcoin-s support both eventually! BIP47 is just the more mature starting point. Think it would be cool to add Silent Payments next and hopefully as the ecosystem continues to stabilize.

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+import org.bitcoins.crypto.{CryptoUtil, ECPrivateKey, ECPublicKey}
+import scodec.bits.ByteVector
+
+import scala.util.Try
+
+/** Represents a BIP47 account at derivation path m/47'/coin_type'/account'.
+  * @see
+  *   [[https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki BIP47]]
+  */
+sealed abstract class BIP47Account {
```
Christewart, line 18: I think we want to `extend Bip32Path` here at minimum.

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+    extPrivKey.deriveChildPrivKey(UInt32(index))
+  }
+
+  def privateKeyAt(index: Int): ECPrivateKey = deriveKey(index).key
+
+  def publicKeyAt(index: Int): ECPublicKey = privateKeyAt(index).publicKey
+
+  def addressAt(index: Int): Bech32Address = {
+    val pubKey = publicKeyAt(index)
+    val spk = P2WPKHWitnessSPKV0(pubKey)
```
Christewart, line 53: It doesn't seem that BIP47 officially supports `P2WPKH`? It seems ti only support `P2PKH` AFAICT. https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki#versions

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /** Creates a BIP47Account from a seed at the standard derivation path. */
+  def fromSeed(
+      seed: ByteVector,
+      network: NetworkParameters,
+      accountIndex: Int = 0): BIP47Account = {
+    val version = network match {
+      case _: MainNet => LegacyMainNetPriv
+      case _          => LegacyTestNet3Priv
```
Christewart: Please enumerate the subclasses of `NetworkParameters` here and elsewhere. I.e. 

```scala
    val version = network match {
      case _: MainNet => LegacyMainNetPriv
      case _: TestNet3 | _: TestNet4 | _: RegTest | _: SigNet =>
        LegacyTestNet3Priv
    }
```

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+    BIP47AccountImpl(accountKey, network, accountIndex)
+  }
+
+  def fromPaymentCode(
+      paymentCode: PaymentCode,
+      network: NetworkParameters): WatchOnlyBIP47Account = {
+    WatchOnlyBIP47Account(paymentCode, network)
+  }
+
+  private def coinType(network: NetworkParameters): Long = network match {
```
Christewart:
```scala
HDCoinType.fromNetwork(network).toInt
```

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  def addressAt(index: Int): Bech32Address = {
+    val pubKey = publicKeyAt(index)
+    val spk = P2WPKHWitnessSPKV0(pubKey)
+    Bech32Address(spk, network)
+  }
+
+  def notificationAddress: P2PKHAddress = {
+    val pubKey = publicKeyAt(0)
+    val hash = CryptoUtil.sha256Hash160(pubKey.bytes)
```
Christewart: Shouldn't need to do the hash manually, just use the constructor that takes a public key

```scala
P2PKHAddress(pubKey, network)
```

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  override def toString: String = {
+    val checksum = CryptoUtil.doubleSHA256(bytes).bytes.take(4)
+    Base58.encode(bytes ++ checksum)
+  }
+
+  def isValid: Boolean = {
+    pubKey.bytes.size == 33 &&
+    (pubKey.bytes.head == 0x02 || pubKey.bytes.head == 0x03) &&
+    chainCode.bytes.size == 32
```
Christewart: This is redundant I believe: https://github.com/bitcoin-s/bitcoin-s/blob/654d4086b993c35d4cd343a2009daed40c27d7b0/core/src/main/scala/org/bitcoins/core/crypto/ChainCode.scala#L7

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
```diff
@@ -0,0 +1,165 @@
...
+  override def bytes: ByteVector =
+    ByteVector(PaymentCode.VersionByte) ++ payload
+
+  override def toString: String = {
+    val checksum = CryptoUtil.doubleSHA256(bytes).bytes.take(4)
+    Base58.encode(bytes ++ checksum)
+  }
+
+  def isValid: Boolean = {
+    pubKey.bytes.size == 33 &&
```
Christewart, line 43: From glancing through BIP47, I don't see a requirement that the public key is in a compressed format? I guess if we are upgrading the protocol for segwit, it must be?

Kinda why i'm suspicious of supporting segwit receives/spends when the protocol doesn't seem to explicitly support segwit - unless I'm missing something.

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
```diff
@@ -0,0 +1,165 @@
...
+  def apply(pubKey: ECPublicKey, chainCode: ChainCode): PaymentCode = {
+    apply(PaymentCodeVersion.V1, pubKey, chainCode, 0x00.toByte)
+  }
+
+  def apply(
+      version: PaymentCodeVersion,
+      pubKey: ECPublicKey,
+      chainCode: ChainCode,
+      features: Byte): PaymentCode = {
+    require(pubKey.isCompressed,
```
Christewart, line 76: Some concern here wrt to compressed public keys, where is this required at in the protocol?

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
Christewart: Even better if you create a helper method somewhere inside of `ExtKeyVersion` to do this :-)

core/src/main/scala/org/bitcoins/core/bip47/PaymentAddress.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  def secretPoint: SecretPoint = SecretPoint(localPrivKey, remotePublicKey)
+
+  def sharedSecretHash: ByteVector =
+    CryptoUtil.sha256(secretPoint.ecdhSecret).bytes
+
+  def tweakFieldElement: FieldElement = {
+    val s = new BigInteger(1, sharedSecretHash.toArray)
+    val n = CryptoParams.getN
+    if (s.compareTo(BigInteger.ONE) <= 0 || s.compareTo(n) >= 0) {
```
Christewart: I believe the `s.compareTo(n) >=0` is covered by this invariant: https://github.com/bitcoin-s/bitcoin-s/blob/3af204e74da3073ecd9541ef018e49b186e0f097/crypto/src/main/scala/org/bitcoins/crypto/FiniteField.scala#L14

However i'm not sure if `s.compareTo(BigInteger.ONE` is covered by an invariant in `FiniteFieldMember`... perhaps something that should be added there in a separate PR?

Christewart (commented):
Here is a first round of review.

MightyMercurian:
Awesome thank you. Going through addressing all of these.

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: Could you clarify what you're hoping to gain from extending `BIP32Path`?

I considered integrating with the `HDPurpose` → `HDCoin` → `HDAccount` hierarchy, but BIP47 accounts are conceptually different from BIP44-style accounts in a few ways:

1. Different lifecycle: A BIP47 account's primary artifact is the `PaymentCode` (derived from the account key's pubkey + chaincode), which gets shared via a notification transaction. The account itself is more of a "payment code generator" than a key derivation endpoint.
2. Non-standard child derivation: After the account level (`m/47'/coin'/account'`), BIP47 derives keys at index 0 for the notification key, then uses ECDH with the remote party's payment code to derive actual payment addresses - it doesn't follow the BIP44 chain/address_index pattern.
3. Bidirectional key material: The same account is used for both sending (using notification key + remote payment code) and receiving (using local keys + remote payment code), which doesn't map cleanly to HDChain's external/internal model.

That said, I can see value in having path serialization for display/debugging. Would adding a def` bip32Path: BIP32Path `method that returns `m/47'/{coinType}'/{accountIndex}'` satisfy the use case? Or is there a specific integration point with the HD infrastructure you have in mind?

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: You're correct that the BIP47 spec only formally defines P2PKH for version 1. However, the derived keys themselves are just EC keypairs - the address type is an implementation choice for the actual payment.

In practice, wallets like Samourai/Sparrow and the [ExtLibJ](https://github.com/Archive-Samourai-Wallet/extlibj/tree/feature/dexwp/java/com/samourai/wallet/bip47) reference implementation use SegWit (P2WPKH) for the derived payment addresses since it reduces fees and is the modern standard. The **notificationAddress** is still P2PKH per spec (needed for the OP_RETURN notification transaction), but once two parties have exchanged payment codes, they can agree to use any address type for subsequent payments.

That said, I can see the argument for being strict to spec. Would you prefer I:
1. Keep as-is (pragmatic, matches real-world implementations)
2. Change addressAt() to return P2PKH and add a separate segwitAddressAt() method
3. Remove SegWit address methods entirely and stick to P2PKH only

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
```diff
@@ -0,0 +1,204 @@
...
+      remotePaymentCode = remotePaymentCode,
+      index = index,
+      network = network
+    )
+  }
+}
+
+object BIP47Account {
+  val Purpose: Int = 47
+  val HardenedOffset: Long = 0x80000000L
```
MightyMercurian: Done - removed `HardenedOffset` and using `ExtKey.hardenedIdx` directly

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: Done - added `ExtKeyPrivVersion.legacyFromNetwork(network)` and `ExtKeyPubVersion.legacyFromNetwork(network)` helpers and updated BIP47Account to use them.

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: Done - using `HDCoinType.fromNetwork(network).toInt`

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: Done - using `P2PKHAddress(pubKey, network)` factory

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
MightyMercurian: Done - removed redundant check

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
MightyMercurian: The compressed public key requirement comes from the BIP47 payload format itself. The payment code is exactly 80 bytes with a fixed structure:

byte 0: version (0x01 or 0x02)
byte 1: features
bytes 2-34: compressed public key (33 bytes)
bytes 35-66: chain code (32 bytes)
bytes 67-79: reserved (13 bytes of zeros)

The spec allocates exactly 33 bytes for the public key, which only fits a compressed key (uncompressed would be 65 bytes). (https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki#binary-serialization)

Re: SegWit - you're right that the BIP47 spec only formally defines P2PKH. However, the derived keys are just EC keypairs, so the address type for actual payments is an implementation choice. Wallets like Samourai/Sparrow use SegWit for derived payment addresses to reduce fees. The notification address must stay P2PKH per spec (for the OP_RETURN notification tx), which is why notificationAddress returns P2PKHAddress. The segwit methods (addressAt, notificationSegwitAddress) are convenience methods for payment addresses where both parties can agree on address type.

I could rename them to make the distinction clearer (e.g., segwitAddressAt)?

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
MightyMercurian: Answered in the above comment - the 80-byte payload format only allocates 33 bytes for the pubkey, requiring compressed format.

core/src/main/scala/org/bitcoins/core/bip47/PaymentAddress.scala
MightyMercurian: Good catch. Removed the redundant s.compareTo(n) >= 0 check since FiniteFieldMember already enforces < fieldOrder.

The s.compareTo(BigInteger.ONE) <= 0 check remains to reject 0 and 1 - both are technically valid field elements but would be insecure/degenerate as EC scalars (0 produces infinity, 1 produces the known generator point G).

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
Christewart: >I considered integrating with the HDPurpose → HDCoin → HDAccount hierarchy, but BIP47 accounts are conceptually different from BIP44-style accounts in a few ways:

I agree with this.

>That said, I can see value in having path serialization for display/debugging. Would adding a defbip32Path: BIP32Pathmethod that returns m/47'/{coinType}'/{accountIndex}' satisfy the use case?

I think extending it would just make more sense - as this is literally a BIP32Path. I.e. `m/47'/coin'/account'`.

I tried to this here: https://github.com/Christewart/bitcoin-s-core/commit/87d2587b4faf4fd87b3f8ea6622896d06ab1cdff

but I ran into an issue. You are putting the `xprv`/`xpub` inside of `BIP47Account` rather than letting the data structure just be a derivation path used on key that is provided else where. For instance, [`HDAccount`](https://github.com/bitcoin-s/bitcoin-s/blob/9feabc1c2b95141c67cf934fd24072bf810fdab2/core/src/main/scala/org/bitcoins/core/hd/HDAccount.scala#L14) doesn't have an `xpub` or `xprv` embedded in the data structure. 

In the wallet, we store the `BIP32Path` along with an `xpub` in a data structure called [`AccountDb`](https://github.com/bitcoin-s/bitcoin-s/blob/afddf73c48d12515d44d11e959ac6d995fdbe61f/core/src/main/scala/org/bitcoins/core/api/wallet/db/AccountDb.scala#L18). When we want to generate a _new_ address, we find the most recent address index used in the database for that account, and [then derive a new address associated with that index](https://github.com/bitcoin-s/bitcoin-s/blob/01854bed91d0bb7b86e61730dcca986859a2741a/wallet/src/main/scala/org/bitcoins/wallet/internal/AccountHandling.scala#L446)

Let me think about this a bit more for what the best path forward is.

core/src/main/scala/org/bitcoins/core/bip47/PaymentCode.scala
Christewart: I think this is fine. My main concern was attempting to use an uncompressed public key is highly discouraged in segwit v0: https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#restrictions-on-public-key-type

As you pointed out, BIP47 only allows compressed keys so this concern is moot. Thanks for pointing that out to me.

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
Christewart: Thanks for the context. I think your choices make sense now that I have more context on the BIP47 spec and understanding of how other implementations work. Thanks for the helpful info!

core/src/main/scala/org/bitcoins/core/bip47/BIP47Account.scala
MightyMercurian: Thanks for digging into this and sharing your refactoring attempt - really helpful to see where the friction points are.

I took a look at [ExtLibJ](https://github.com/Archive-Samourai-Wallet/extlibj/tree/feature/dexwp) (the Samourai Java reference implementation) to see how they handled this. Interestingly, their `BIP47Account` extends `HD_Account`, which does embed the key material (`DeterministicKey aKey`) directly in the account class. So the pattern we're using matches the reference implementation.

I think the embedded key approach makes sense for BIP47 specifically because the primary operations (`paymentAddressForSending`, `paymentAddressForReceiving`) require private key access for ECDH shared secret computation. Having the key embedded keeps the API ergonomic for those use cases.

That said, I totally understand the value of consistency with bitcoin-s's existing wallet architecture and the `HDAccount`/`AccountDb` pattern. A few options I can think of:

1. Add a `bip32Path`: `BIP32Path` getter - Gives you path serialization for display/storage without changing the core structure. Simple addition.
2. Create a separate `BIP47Path extends BIP32Path` - A lightweight path-only class for wallet DB integration, keeping `BIP47Account` as-is for the key-holding use case.
3. Full refactor to match `HDAccount` pattern - Happy to do this if you think it's the right long-term direction, though it would be a larger change.

I'm flexible here and want to make sure this integrates well with the rest of bitcoin-s. Any thoughts on what approach feels right to you?

Christewart: I'm leaning towards just merging as is - and if refactors are required in the future we can tackle that with more application specific knowledge. I'll take one more look over this tomorrow and hopefully we can get it merged this week.

Christewart (approved):

Christewart:
Thanks for the PR @MightyMercurian !

MightyMercurian:
Awesome! Thanks for your help @Christewart

If you think of anything else to follow up with based on our discussions, feel free to let me know!

PR #6170, opened 2026-01-05
2025 11 27 bip54 tests

Christewart:
Implement test vectors found for BIP54 found here: https://github.com/bitcoin/bips/pull/2015


core/src/main/scala/org/bitcoins/core/policy/Policy.scala
```diff
@@ -70,6 +83,39 @@ sealed abstract class Policy {
   def maxFee: CurrencyUnit = Satoshis(10) * CurrencyUnits.oneMBTC
 
   def sequence: UInt32 = UInt32.zero
+
+  /** Limit of sigops per block as per BIP54
```
darosior, line 87: per transaction*

Issue #6192, opened 2026-02-03
Query timed out with in state... when node running in steady state (not IBD)


Christewart:
Observed on 444f390e . This keeps occurring against a locally running bitcoind until we disconnect the peer. Possibly related to https://github.com/bitcoin/bitcoin/issues/27085 #4976 #5802 

```
2026-02-03 01:55:53,418UTC INFO [TransactionProcessing] Finished processing of block=00000000000000000000cbe49fc0be4f9c60a3d78110b80898660bcb276026c8. It took 3451ms
2026-02-03 01:56:38,782UTC INFO [ChainHandler] Processed filters headers from height=934786 to 934786. Best filterheader.blockHash=00000000000000000000cbe49fc0be4f9c60a3d78110b80898660bcb276026c8
2026-02-03 01:58:38,955UTC INFO [DataMessageHandler] Query timed out with in state=FilterSync(syncPeer=Peer(127.0.0.1:8333),peers=Set(Peer(127.0.0.1:8333)),waitingForDisconnection=Set(),filterBatchCache.size=0), received payload=tx
2026-02-03 01:59:45,965UTC INFO [ChainHandler] Processed filters from height=934786 to 934786. Best filter.blockHash=00000000000000000000cbe49fc0be4f9c60a3d78110b80898660bcb276026c8
2026-02-03 01:59:45,971UTC INFO [DataMessageHandler] We are synced
```

```
2026-02-03 02:23:58,899UTC INFO [DataMessageHandler] Received block=000000000000000000007223a7d56886662a32c3616828801e34454919b211d7 from peer=Peer(127.0.0.1:8333) processing block's header... state=DoneSyncing(peers=Set(Peer(127.0.0.1:8333)),waitingForDisconnection=Set())
2026-02-03 02:23:58,994UTC INFO [ChainHandler] Processed headers from height=934786 to 934787. Best hash=000000000000000000007223a7d56886662a32c3616828801e34454919b211d7
2026-02-03 02:24:02,532UTC INFO [TransactionProcessing] Finished processing of block=000000000000000000007223a7d56886662a32c3616828801e34454919b211d7. It took 2222ms
2026-02-03 02:25:59,122UTC INFO [DataMessageHandler] Query timed out with in state=FilterHeaderSync(syncPeer=Peer(127.0.0.1:8333),peers=Set(Peer(127.0.0.1:8333)),waitingForDisconnection=Set()), received payload=tx
[...]

PR #6210, opened 2026-02-08
node: Optimize receiving packets in the `PeerConnection` stream

Christewart:
Part of #6192 

Optimize parsing of messages by batching bytes into bigger chunks to avoid overhead of parsing when we likely wouldn't have all the packets as part of a `NetworkMessage`. 

We now emit `ByteStrings` downstream when either

- 10 milliseconds have passed
- We have accumulated 64 `ByteString`'s


Christewart:
The original problem (#6192) is fixed by #6242

PR #6453, opened 2026-08-22
crypto: fix 8 lax/unsafe parsing findings in the crypto module

Christewart:
## Summary

Fixes all 8 findings reported in `CryptoParsingSecurityTest` (see the security reproduction report this branch was built against). Each commit pairs one production fix with the reproduction test that proves it — the reproduction tests were written to assert the *correct* (safe) behavior, so each failed until its corresponding fix landed.

- `DERSignatureUtil.parseDERLax` mis-parsed long-form DER length bytes due to signed-`Byte` arithmetic (`0x81 - 0x80` computed as `-255` instead of Bitcoin Core's `lenbyte & 0x7f` result of `1`).
- `SchnorrPublicKey`/`XOnlyPubKey`/`SchnorrNonce.fromBytes` silently left-padded short inputs and silently dropped the first byte of 33-byte inputs instead of requiring exactly 32 bytes. This uncovered a latent bug in `ECKey.scala` where several `schnorrPublicKey`/`schnorrNonce`/`toXOnly` accessors relied on that leniency to extract the x-coordinate from a compressed public key; fixed those to explicitly use `compressedBytes.tail`.
- A 65-byte `SchnorrDigitalSignature` with an explicit `0x00` (SIGHASH_DEFAULT) trailing byte was accepted, violating BIP340/341.
- Garbage bytes fed into `ECDigitalSignature` silently "decoded" to r=s=0 via `DERSignatureUtil.decodeSignature`'s fallback, which made `isLowS`/`hashTypeOpt` misreport undecodable bytes as valid. Fixed those two accessors to correctly report "not a valid signature" — `ECDigitalSignature.fromBytes` itself intentionally stays lenient (doesn't throw), since it's called directly on untrusted scriptSig bytes in `CryptoInterpreter.opCheckSig`/`opCheckMultiSig` with no `Try` wrapper, and throwing there would turn a clean script failure into a crash.
- `HashType.isDefinedHashtypeSignature` wrongly counted `0x00` as a defined ECDSA hashtype (it's taproot/schnorr-only per Core). This uncovered two related `testkit-core` generator bugs producing invalid signatures with an explicit `0x00` byte, previously masked by the same crypto bug.
- `ECDigitalSignature.fromFrontOfBytesWithSigHash` threw a raw `NoSuchElementException` instead of a clean `IllegalArgumentException` when no sighash byte was present.
- `Sha1Digest`/`Sha1DigestBE.fromBytes` had no length validation, unlike every sibling hash digest type.
- `CryptoBytesUtil.isHex` rejected uppercase hex while `decodeHex` accepted it — now consistent.

## Test plan

- [x] `sbt cryptoTestJVM/test` — 213/213 passing
- [x] `sbt cryptoTestJS/test` — 154/154 passing
- [x] `sbt coreTestJVM/test` script/crypto/protocol/serializer suites — all passing, including the full BIP script interpreter test-vector suite
- [x] `sbt dlcOracleTest/test` — 58/58 passing (exercises schnorr nonce/key derivation end-to-end)

https://claude.ai/code/session_013yT4qDB1VycMnpcNMooCR5

### DIFF
--- a/testkit-core/src/main/scala/org/bitcoins/testkitcore/gen/CryptoGenerators.scala
+++ b/testkit-core/src/main/scala/org/bitcoins/testkitcore/gen/CryptoGenerators.scala
@@ -228,7 +228,7 @@ sealed abstract class CryptoGenerators {
   def digitalSignatureWithSigHash: Gen[ECDigitalSignature] = {
     for {
       sig <- digitalSignature
-      sigHash <- hashType
+      sigHash <- preTaprootHashType
     } yield {
       ECDigitalSignature(sig.bytes :+ sigHash.byte)
     }
@@ -253,7 +253,14 @@ sealed abstract class CryptoGenerators {
       sigNoHashType = privKey.schnorrSign(hash.bytes)
       hashType <- taprootHashType
     } yield {
-      sigNoHashType.appendHashType(hashType)
+      // BIP341 states that if we use default sighash we don't append the
+      // hash type byte (see Sign.schnorrSignWithHashType); a 65 byte
+      // signature with an explicit 0x00 trailing byte is invalid
+      if (hashType == HashType.sigHashDefault) {
+        sigNoHashType
+      } else {
+        sigNoHashType.appendHashType(hashType)
+      }
     }
   }
   def schnorrDigitalSignature: Gen[SchnorrDigitalSignature] = {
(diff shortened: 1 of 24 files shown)


testkit-core/src/main/scala/org/bitcoins/testkitcore/gen/CryptoGenerators.scala
```diff
@@ -309,8 +316,15 @@ sealed abstract class CryptoGenerators {
...
-  /** Generates a random [[HashType HashType]] */
-  def hashType: Gen[HashType] = Gen.oneOf(HashType.hashTypes)
+  /** Generates a random [[HashType HashType]] to append to a legacy/segwit v0
+    * ECDSA signature. Excludes SIGHASH_DEFAULT: that hashtype byte is only
+    * meaningful for taproot/schnorr signatures (see [[taprootHashType]]) and
+    * ECDigitalSignature.appendHashType always appends the byte verbatim, so
+    * including it here would generate ECDSA signatures with an explicit 0x00
+    * trailing byte that Bitcoin Core rejects as an undefined hashtype.
+    */
+  def hashType: Gen[HashType] =
```
Christewart: Can you modify the name of this generator to be something like `preTaprootHashType`

PR #6454, opened 2026-08-22
Fix core module security reproduction findings (41 bugs, 1 commit per bug)

Christewart:
## Summary

Fixes 41 of the 43 security findings reproduced for the `core` module in `SECURITY_REPRODUCTIONS.md`, each as its own commit with a paired reproduction test that fails before the fix and passes after. Companion branch/PR to the `crypto` module fixes (#6453).

Findings by file:
- `TlvParsingSecurityTest`: 1/1 fixed
- `TransactionParsingSecurityTest`: 7/8 fixed (finding 7, out-of-money-range `TransactionOutput` values, intentionally skipped — `TransactionOutput` deliberately supports negative values for general data modeling, matching existing `EmptyTransactionOutput`/generator design and Core's own split between raw deserialization and `CheckTransaction`-layer validation)
- `P2pParsingSecurityTest`: 7/7 fixed
- `MerkleConsensusSecurityTest`: 6/7 fixed (finding 2, CVE-2012-2459 duplicate-leaf mutation detection, intentionally skipped — an unconditional duplicate-sibling check broke a real historical 20-tx block fixture with a legitimate duplicate leaf pair; no confident, verified distinguishing rule was found)
- `ScriptArithmeticSecurityTest`: 8/8 fixed
- `SignatureCheckingSecurityTest`: 12/12 fixed

## Notable fixes

- **Taproot SPK classification**: a v1 32-byte witness program with an invalid x-only coordinate was misclassified as an unassigned (anyone-can-spend) witness version instead of taproot. Fixed by making classification purely structural (matching Core) and making `TaprootScriptPubKey.pubKey` lazy so parsing real on-chain data with such outputs doesn't crash.
- **BIP143 sighash / OP_CODESEPARATOR**: the segwit v0 sighash silently ignored an executed `OP_CODESEPARATOR`, always committing to the full witness script instead of the post-separator scriptCode.
- **BIP342 tapscript NULLFAIL**: an invalid non-empty tapscript signature only hard-failed when `SCRIPT_VERIFY_NULLFAIL` was explicitly set, instead of unconditionally as BIP342 requires.
- **OP_CHECKMULTISIG NULLFAIL**: only checked remaining (unconsumed) signatures on failure, missing signatures that matched earlier and were consumed.
- Several parser hardening fixes (CompactSizeUInt canonical encoding, witness stack/flag validation, P2P message checksum/length checks, bloom filter size limits) mirroring Bitcoin Core's actual deserialization behavior.

## Test plan

[...]

PR #6457, opened 2026-08-23
core: fix 8 script interpreter findings in locktime/arithmetic/stack opcodes

Christewart:
## Summary

Fixes 8 of the findings reported against the script interpreter (see the security reproduction report this branch was built against, and split out of #6454 into a smaller, independently-reviewable PR). Each commit pairs one production fix with the reproduction test that proves it, relocated into the relevant existing test suite rather than a new dedicated test file — the tests were written to assert the *correct* (Bitcoin Core compatible) behavior, so each failed until its corresponding fix landed.

- `LockTimeInterpreter.opCheckSequenceVerify` only rejected the exact value `ScriptNumber.negativeOne` with `SCRIPT_ERR_NEGATIVE_LOCKTIME`; any other negative operand (e.g. -2) has the BIP68 disable flag set in its two's-complement bit pattern and fell through to the "treat as NOP" branch, bypassing BIP112's negative-locktime check entirely.
- `ArithmeticInterpreter.performUnaryArithmeticOperation`'s `ScriptConstant` branch re-derived the operand via `ScriptNumber(ScriptNumberUtil.toLong(s.hex))`, silently minimizing a non-minimal 5-byte encoding down to ≤4 bytes before the `isLargerThan4Bytes` consensus check ran, bypassing it whenever `SCRIPT_VERIFY_MINIMALDATA` was not set.
- `ScriptNumber` construction from a `ScriptConstant` eagerly parses the operand's hex with `java.lang.Long.parseLong`; an operand wider than 8 bytes overflowed that parse and threw an uncaught `NumberFormatException` instead of a script evaluation error, across every `ScriptConstant`-interpreting branch in both `ArithmeticInterpreter` and `LockTimeInterpreter`. Added a shared `interpretNumber` helper in each interpreter that wraps the construction in a `Try` and fails execution with `ScriptErrorUnknownError`.
- Opcode bytes 187-254 parsed to `ReservedOperation.UndefinedOP_NOP`, a generic catch-all. Per BIP342 (Tapscript) these are individually named `OP_SUCCESS187`-`OP_SUCCESS254` opcodes; added a new `OP_SUCCESSx` operation type and wired it into opcode parsing (0xff/255 remains the only genuinely unassigned byte). The base interpreter loop doesn't yet evaluate Tapscript-aware "succeed immediately" semantics, so execution behavior outside that context is unchanged — only the parsed opcode identity changes.
- `LockTimeInterpreter.opCheckLockTimeVerify` never checked `ScriptFlagUtil.requireMinimalData`/`isShortestEncoding` on its operand, unlike `opCheckSequenceVerify` which already enforces this for CSV; Core's `CScriptNum` constructor enforces `fRequireMinimal` identically for both CLTV and CSV.
- `StackInterpreter.opIfDup` only compared the stack top against `ScriptNumber.zero` (the canonical empty-vector encoding), so a non-canonical zero like `"00"` fell through and got duplicated; switched to `BitcoinScriptUtil.castToBool`, which correctly treats every all-zero byte encoding as false.
- `StackInterpreter.opRoll` removed the rolled element with `stack.tail.diff(List(newStackTop))`, which removes the first element equal to `newStackTop` by *value*, not the element at depth n by *position* — corrupting stack order when duplicate values are present. Fixed by splitting the stack at the target index and reassembling it.
- `BitcoinScriptUtil.isPushOnly` rejected `OP_16` itself via a strict `scriptOp.opCode < OP_16.opCode` comparison; Core's `IsPushOnly` allows all opcodes up to and including `OP_16`, since it pushes the number 16 just like `OP_1`-`OP_15`.

## Test plan

- [x] Targeted test suites for each fix (`LockTimeInterpreterTest`, `ArithmeticInterpreterTest`, `StackInterpreterTest`, `ScriptParserTest`, `ReservedOperationsFactoryTest`, `BitcoinScriptUtilTest`) — all passing
- [x] `ScriptInterpreterTest` (full `script_tests.json` BIP test-vector suite) — passing after every commit
- [x] `sbt coreTestJVM/test` — 1717/1717 passing
- [x] `sbt scalafmtCheckAll` — clean

https://claude.ai/code/session_013yT4qDB1VycMnpcNMooCR5

### DIFF
--- a/core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
@@ -86,6 +86,26 @@ class ArithmeticInterpreterTest extends BitcoinSUnitTest {
       .error must be(Some(ScriptErrorUnknownError))
   }
 
+  it must "fail gracefully with a ScriptError on numeric operands larger than 8 bytes instead of throwing NumberFormatException" in {
+    // ScriptNumber construction eagerly derives a Long value via ScriptNumberUtil.toLong, which
+    // parses the operand's hex with java.lang.Long.parseLong -- a >8-byte operand overflows that
+    // parse and throws NumberFormatException. That must be caught and turned into a script error
+    // rather than escaping the interpreter as an uncaught crash.
+    val stack = List(ScriptConstant("ffffffffffffffff7f")) // 9-byte operand
+    val script = List(OP_1ADD)
+    val t = buildTxSigComponent(Seq(ScriptVerifyNone))
+    val program = PreExecutionScriptProgram(t).toExecutionInProgress
+      .updateStackAndScript(stack, script)
+    val newProgram = AI.op1Add(program)
+    newProgram.isInstanceOf[ExecutedScriptProgram] must be(true)
+    // matches Bitcoin Core: CScriptNum parsing failures are caught and
+    // mapped to SCRIPT_ERR_UNKNOWN_ERROR, same as the oversized-operand
+    // case above
+    newProgram
+      .asInstanceOf[ExecutedScriptProgram]
+      .error must be(Some(ScriptErrorUnknownError))
+  }
+
   private def buildTxSigComponent(flags: Seq[ScriptFlag]): TxSigComponent = {
     BaseTxSigComponent(
       transaction = TestUtil.transaction,
--- a/core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
+++ b/core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
@@ -1,10 +1,15 @@
 package org.bitcoins.core.script.arithmetic
 
+import org.bitcoins.core.crypto.{BaseTxSigComponent, TxSigComponent}
+import org.bitcoins.core.currency.CurrencyUnits
+import org.bitcoins.core.protocol.transaction.TransactionOutput
 import org.bitcoins.core.script.constant._
+import org.bitcoins.core.script.flag.{ScriptFlag, ScriptVerifyNone}
 import org.bitcoins.core.script.result._
 import org.bitcoins.core.script.{
   ExecutedScriptProgram,
-  ExecutionInProgressScriptProgram
+  ExecutionInProgressScriptProgram,
+  PreExecutionScriptProgram
 }
 import org.bitcoins.core.util.ScriptProgramTestUtil
 import org.bitcoins.testkitcore.util.TestUtil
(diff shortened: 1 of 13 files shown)


core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
```diff
@@ -60,6 +65,35 @@ class ArithmeticInterpreterTest extends BitcoinSUnitTest {
...
+    // check is bypassed whenever SCRIPT_VERIFY_MINIMALDATA is not set
+    val stack =
+      List(ScriptConstant("ff00000000")) // 5-byte non-minimal encoding of 255
+    val script = List(OP_1ADD)
+    val t = buildTxSigComponent(Seq(ScriptVerifyNone))
+    val program = PreExecutionScriptProgram(t).toExecutionInProgress
+      .updateStackAndScript(stack, script)
+    val newProgram = AI.op1Add(program)
+    newProgram.isInstanceOf[ExecutedScriptProgram] must be(true)
+    newProgram.asInstanceOf[ExecutedScriptProgram].error.isDefined must be(
```
Christewart: Probably should assert the exact error and verify it is the same as bitcoin cores

core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
```diff
@@ -82,6 +82,23 @@ class ArithmeticInterpreterTest extends BitcoinSUnitTest {
...
+    // parse and throws NumberFormatException. That must be caught and turned into a script error
+    // rather than escaping the interpreter as an uncaught crash.
+    val stack = List(ScriptConstant("ffffffffffffffff7f")) // 9-byte operand
+    val script = List(OP_1ADD)
+    val t = buildTxSigComponent(Seq(ScriptVerifyNone))
+    val program = PreExecutionScriptProgram(t).toExecutionInProgress
+      .updateStackAndScript(stack, script)
+    val newProgram = AI.op1Add(program)
+    newProgram.isInstanceOf[ExecutedScriptProgram] must be(true)
+    newProgram.asInstanceOf[ExecutedScriptProgram].error.isDefined must be(
```
Christewart: assert the exact error to make sure we are identical to bitcoin core

core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
Christewart: Good catch. Both assertions now check the exact `ScriptErrorUnknownError` rather than just `isDefined`, matching Bitcoin Core's behavior: the CScriptNum constructor throws `scriptnum_error` for an oversized/overflowed operand, which `interpreter.cpp` catches and maps to `SCRIPT_ERR_UNKNOWN_ERROR` (see `isLargerThan4Bytes`/`interpretNumber` in ArithmeticInterpreter.scala, which already documented this mapping). Pushed as fixups into the two original commits.

core-test/src/test/scala/org/bitcoins/core/script/arithmetic/ArithmeticInterpreterTest.scala
Christewart: Same fix as the comment above — both assertions now check the exact `ScriptErrorUnknownError`.

PR #6459, opened 2026-08-23
core: fix 8 transaction/TLV parsing security findings

Christewart:
## Summary
- Fixes 8 security-reproduction findings in the core module's transaction/TLV/block parsing paths (unbounded allocation from attacker-controlled counts, malformed witness tx reinterpretation, trailing-garbage/truncated-field acceptance, short-input crashes, non-canonical CompactSizeUInt varints, oversized witness stack elements, invalid witness flag bytes, and Long-overflow in fee-rate multiplication).
- Each fix is its own commit (1 bug per commit) with its reproduction test relocated into an existing test suite (`TxUtilTest`, `FeeUnitTest`, and others touched by the fixes) rather than a new test file.
- One additional finding from the original investigation (out-of-money-range transaction output values accepted at parse time) was intentionally left unfixed — see the last commit's message for the rationale (mirrors Bitcoin Core's own separation between raw deserialization and `CheckTransaction`-level consensus validation).

## Test plan
- [x] Each commit's relocated test passes individually
- [x] Full `coreTestJVM/test`: 1722/1722 passed
- [x] `scalafmtCheckAll`: clean
- [x] No new test files added — confirmed no `core-test/.../security/` paths in this branch's history

This is part of a 5-PR split of #6454 (closed in favor of these smaller, logically-grouped PRs) — this one covers transaction/TLV parsing. The other four cover P2P networking (#6458), merkle/consensus, script arithmetic (#6457), and signature checking.

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
@@ -180,6 +180,27 @@ object Transaction extends Factory[Transaction] {
     tx
   }
 
+  /** Note: unlike [[fromBytes]], this requires `hex` to decode to EXACTLY one
+    * transaction with no trailing bytes left over. [[fromBytes]] itself cannot
+    * enforce that: it also serves as the per-element parser when deserializing
+    * a sequence of transactions sharing one buffer (e.g. a block's transaction
+    * list in RawBlockSerializer), where bytes trailing a single parsed
+    * transaction are the start of the next one, not garbage. A hex string, by
+    * contrast, is always meant to represent exactly one transaction --
+    * mirroring Bitcoin Core's own split between its stream-based tx
+    * deserializer (never checks for leftover bytes) and DecodeHexTx (which
+    * explicitly checks the stream is empty afterward).
+    */
+  override def fromHex(hex: String): Transaction = {
+    val bytes = org.bitcoins.crypto.CryptoBytesUtil.decodeHex(hex)
+    val tx = fromBytes(bytes)
+    require(
+      tx.bytes.length == bytes.length,
+      s"${bytes.length - tx.bytes.length} trailing bytes after a complete transaction"
+    )
+    tx
+  }
+
   /** This allows us to accurately type transaction input's
     * [[org.bitcoins.core.protocol.script.ScriptSignature]] as we have the
     * corresponding [[org.bitcoins.core.protocol.script.ScriptPubKey]] the input
@@ -241,6 +262,15 @@ object BaseTransaction extends Factory[BaseTransaction] {
       BytesUtil.parseCmpctSizeUIntSeq(txInputBytes, TransactionInput)
     val (outputs, lockTimeBytes) =
       BytesUtil.parseCmpctSizeUIntSeq(outputBytes, TransactionOutput)
+    // intentionally >= rather than ==: fromBytes is used to parse a single
+    // transaction out of a longer byte string that may have more data
+    // trailing the locktime (e.g. the next transaction in a block), and is
+    // not itself responsible for rejecting trailing garbage -- see the
+    // Transaction.fromHex doc comment above, which is the entry point that
+    // enforces "exactly one transaction, no trailing bytes"
+    require(
+      lockTimeBytes.length >= 4,
+      s"Must have at least 4 bytes for locktime, got=${lockTimeBytes.length}")
     val lockTime = UInt32(lockTimeBytes.take(4).reverse)
 
     BaseTransaction(version, inputs, outputs, lockTime)
--- a/core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
+++ b/core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
@@ -405,8 +405,8 @@ object WitnessTransaction extends Factory[WitnessTransaction] {
       "Incorrect marker for witness transaction, the marker MUST be 0 for the marker according to BIP141, got: " + marker)
     val flag = bytes(5)
     require(
-      flag.toInt != 0,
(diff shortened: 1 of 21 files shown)


core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -132,11 +132,37 @@ object Transaction extends Factory[Transaction] {
...
+        // vector has parsed successfully, the bytes are unambiguously meant
+        // to be a witness transaction, and any later failure (e.g. a
+        // malformed witness stack or truncated locktime) is a genuine parse
+        // error that must not be silently reinterpreted as an unrelated
+        // base transaction.
+        scala.util.Try(WitnessTransaction.parseInputsAndOutputs(bytes)) match {
+          case scala.util.Failure(_) =>
             BaseTransaction.fromBytes(bytes)
+          case scala.util.Success((version, inputs, outputs, witnessBytes))
+              if inputs.isEmpty =>
```
Christewart, line 161: How does this work if our parsed inputs is empty? Doesn't this mean we have no outpoints or sequence numbers?

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
```diff
@@ -232,6 +253,9 @@ object BaseTransaction extends Factory[BaseTransaction] {
       BytesUtil.parseCmpctSizeUIntSeq(txInputBytes, TransactionInput)
     val (outputs, lockTimeBytes) =
       BytesUtil.parseCmpctSizeUIntSeq(outputBytes, TransactionOutput)
+    require(
+      lockTimeBytes.length >= 4,
```
Christewart, line 272: Shouldn't `locktimeBytes.length == 4`

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
```diff
@@ -658,16 +658,30 @@ trait BitcoinScriptUtil {
...
+    *   false: some internal callers (e.g. TransactionSignatureSerializer's
+    *   legacy sighash algorithm, which mirrors Bitcoin Core's CScript handling)
+    *   deliberately construct a script from just a length-prefix with no bytes
+    *   following it as a placeholder, and are not parsing untrusted wire data.
+    */
   def parseScript[T <: Script](
       bytes: ByteVector,
-      f: Vector[ScriptToken] => T): T = {
+      f: Vector[ScriptToken] => T,
+      strict: Boolean = false): T = {
```
Christewart: Shouldn't this be `true` by default ?

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
Christewart: Good question — added a clarifying comment inline. Short answer: it's fine. `inputs.isEmpty` means there are zero outpoints/sequence numbers to build a witness for, which is trivially consistent (`TransactionWitness(witnessBytes, inputs.size = 0)` just consumes zero witness stacks). This branch exists purely for the marker/flag ambiguity described above it: retry the witness/locktime parse in case this really is a valid (if unusual) zero-input witness tx, and only fall back to the base-tx reinterpretation if that retry itself fails.

core/src/main/scala/org/bitcoins/core/protocol/transaction/Transaction.scala
Christewart: No — this has to stay `>=`, not `==`. `BaseTransaction.fromBytes` doubles as the per-element parser when deserializing a sequence of transactions sharing one buffer (e.g. a block's transaction list in RawBlockSerializer), where bytes trailing this tx's locktime are the start of the next transaction, not garbage. The place that actually enforces "exactly one transaction, no trailing bytes" is `Transaction.fromHex` (see its doc comment a few lines up), which requires the parsed tx's own re-serialization to consume every byte. Added a comment pointing at that so it's not confusing in isolation.

core/src/main/scala/org/bitcoins/core/util/BitcoinScriptUtil.scala
Christewart: Yes, you're right — flipped the default to `strict = true`. Traced the one caller that actually needed the lenient behavior: `TransactionSignatureSerializer`'s legacy (SigVersionBase) sighash algorithm constructs a placeholder `NonStandardScriptSignature` from just a `compactSizeUInt` length-prefix with no script bytes following it (mirroring Bitcoin Core's `CScript` retaining only its compact size while clearing the actual asm for OP_CODESEPARATOR handling) — that went through `ScriptFactory.fromBytes`'s now-strict default and would throw.

Threaded a `strict` param through `ScriptFactory.fromBytes` so that one call site can opt out explicitly (`NonStandardScriptSignature.fromBytes(bytes, strict = false)`) with a comment explaining why, while every other caller — including the many `ScriptPubKey`/`ScriptSignature` subtypes that go through the shared `ScriptFactory` machinery on untrusted wire data — now gets strict-by-default parsing.

Running the full suite with this flipped surfaced one genuinely broken test fixture in `RawTxFinalizerTest` that was constructing a P2PKH `ScriptPubKey` from a raw-bytes hex string via `ScriptPubKey(hex)` (which expects a CompactSizeUInt length prefix) instead of `ScriptPubKey.fromAsmHex(hex)`. It was silently misparsing (dropping the leading OP_DUP byte, since `0x76` was consumed as a bogus declared length of 118) instead of erroring, and coincidentally still produced correct BIP69 sort-order test results. Fixed to use `fromAsmHex`. Full `coreTestJVM/test` (1722/1722) and `scalafmtCheckAll` both pass with the flip.

PR #6460, opened 2026-08-23
core: fix 12 sighash/taproot signature-checking security findings

Christewart:
## Summary
Fixes 12 security-reproduction findings in the core module's signature-checking and sighash paths (taproot invalid x-coordinate classification, BIP143/BIP341/BIP342 sighash computation, NULLFAIL/LOW_S/DER strictness semantics, FindAndDelete, and STRICTENC hashtype handling):

- `9c973a6b04` — don't treat a v1 witness program with an invalid x coordinate as anyone-can-spend
- `2751feda66` — commit the segwit v0 sighash to the scriptCode after OP_CODESEPARATOR
- `79dcf8a9bb` — reject a 65-byte tapscript signature with an explicit SIGHASH_DEFAULT byte
- `334cc6f1e4` — fail validation for taproot SIGHASH_SINGLE with no corresponding output
- `4b3ffa3058` — fail a tapscript OP_CHECKSIG immediately on an invalid non-empty signature
- `42827fd22a` — don't fail an empty signature with the LOW_S check
- `c82d7e1a37` — apply NULLFAIL to all OP_CHECKMULTISIG signatures, including consumed ones
- `66fd85b7e3` — require strict DER encoding when only the LOW_S flag is set
- `11db1bf9e4` — don't enforce WITNESS_UNEXPECTED without the ScriptVerifyWitness flag
- `8283bac411` — remove all occurrences of a signature with legacy FindAndDelete
- `bb04cc5e47` — don't crash on a segwit v0 sighash with an out of range input index
- `8a1e1d1df4` — don't count 0x00 as a defined hashtype byte for STRICTENC

Each fix is its own commit (1 bug per commit) with its reproduction test relocated into an existing test suite (`TransactionSignatureCheckerTest`, `TransactionSignatureSerializerTest`, `ScriptInterpreterTest`, `ScriptPubKeyTest`, `ScriptFlagUtilTest`, `HashTypeTest`, and others touched by each fix) rather than a new test file.

## Test plan
- [x] Each commit's relocated test passes individually
- [x] Full `coreTestJVM/test`: 1720/1720 passed
- [x] `scalafmtCheckAll`: clean
- [x] No new test files added — confirmed no `core-test/.../security/` paths in this branch's history

This is part of a 5-PR split of #6454 (closed in favor of these smaller, logically-grouped PRs) — this one covers signature checking. The other four cover transaction/TLV parsing (#6459), P2P networking (#6458), merkle/consensus, and script arithmetic (#6457).

### DIFF
--- a/core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
+++ b/core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
@@ -242,19 +242,13 @@ trait TransactionSignatureChecker {
                                                       taprootOptions)
     ) match {
       case Failure(_) =>
-        nullFailCheckSchnorrSig(sigs = Vector(signature),
-                                result =
-                                  SignatureValidationErrorIncorrectSignatures,
-                                flags = flags)
+        tapscriptNullFailCheck(signature)
       case Success(hash) =>
         val result = pubKey.verify(hash, signature)
         if (result) {
           SignatureValidationSuccess
         } else {
-          nullFailCheckSchnorrSig(sigs = Vector(signature),
-                                  result =
-                                    SignatureValidationErrorIncorrectSignatures,
-                                  flags = flags)
+          tapscriptNullFailCheck(signature)
         }
     }
   }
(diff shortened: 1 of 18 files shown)


core/src/main/scala/org/bitcoins/core/crypto/TransactionSignatureChecker.scala
```diff
@@ -281,18 +281,25 @@ trait TransactionSignatureChecker {
       sigs: List[ECDigitalSignature],
       pubKeys: List[ECPublicKeyBytes],
       flags: Seq[ScriptFlag],
-      requiredSigs: Long): TransactionSignatureCheckerResult = {
+      requiredSigs: Long,
+      originalSigs: List[ECDigitalSignature] = null)
```
Christewart: Don't use `null` and don't use a default parameter. require the  `originalSigs` to be passed everywhere explicity (`List.empty` in the case there are no sigs).

Christewart: Fixed. Removed the `= null` default — `originalSigs` is now a required `List[ECDigitalSignature]` param at every call site:

- the tail-recursive internal calls already threaded it explicitly (just dropped the now-redundant `if (originalSigs == null) sigs else originalSigs` local val and use `originalSigs` directly)
- the one production call site (CryptoInterpreter.scala) now passes `originalSigs = signatures` explicitly
- all 8 test call sites in TransactionSignatureCheckerTest.scala now pass `originalSigs` explicitly too (matching their `sigs` argument, since at the top-level call the two coincide)

Full `coreTestJVM/test` (1729/1729) and `scalafmtCheckAll` pass. Pushed as a fixup into the original NULLFAIL commit.

